PR #2, opened 2020-09-29
Updating docs to indicate correct Authorization Header for Signed URL requests

kellydunn:
While attempting to integrate our backend with Cloudflare Stream, I was consistently receiving the following error when supplying the `"X-Auth-Key: ${API-KEY}"` header as the docs indicate:

```
{"success":false,"errors":[{"code":10000,"message":"Authentication error"}]}
```

However, when supplying the same `API_KEY` as an `Authorization: Bearer ${API_KEY}` header, I'm able to get a successful request.

This PR updates the docs to reflect this.

### DIFF
--- a/products/stream/src/content/security/signed-urls.md
+++ b/products/stream/src/content/security/signed-urls.md
@@ -19,7 +19,7 @@ You can [revoke a key](#revoking-keys) anytime for any reason.
 Upon creation you will get a RSA private key in PEM and JWK formats. Keys are created, used and deleted independently of videos. Every key can sign any of your videos.
 
 ```javascript
-// curl -X POST -H "X-Auth-Email: ${EMAIL}" -H "X-Auth-Key: ${API-KEY}"  "https://api.cloudflare.com/client/v4/accounts/{account_id}/stream/keys"
+// curl -X POST -H "X-Auth-Email: ${EMAIL}" -H "Authorization: Bearer ${API-KEY}"  "https://api.cloudflare.com/client/v4/accounts/{account_id}/stream/keys"
 
 {
   "result": {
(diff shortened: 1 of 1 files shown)


renandincer:
Hey @kellydunn: Cloudflare API has two authentication methods: API Keys and API Tokens.

You either need to specify `X-Auth-Email` and `X-Auth-Key` headers where auth key is your _global_ API key, or a `Authorization` header with a API Token. The reason you might be getting authentication errors could be because you are using an API Token where a Global API Key is expected.

Here's the page where some differences between tokens and keys are documented: https://support.cloudflare.com/hc/en-us/articles/200167836-Managing-API-Tokens-and-Keys

New developers on Cloudflare seem to use API Tokens rather than Keys so I think we should move all examples in the Stream docs to use Tokens.

adamschwartz:
@ispivey @renandincer should this be closed?

kellydunn:
>New developers on Cloudflare seem to use API Tokens rather than Keys so I think we should move all examples in the Stream docs to use Tokens.

Word, yeah -- we currently enjoy segmenting access to third-party APIs in our SOA/microservice Architecture by creating Service Specific Access Tokens, not just for cloudflare, but for other services as well.  I'd imagine folks would definitely want to follow similar patterns if they wanted to separate access on a service by service level.

renandincer:
@adamschwartz I'll push up a change making all examples use auth tokens, rather than auth keys to avoid confusion. Please keep this open for a few days until I get to it.

renandincer:
Opened https://github.com/cloudflare/cloudflare-docs/pull/149 to make this change across the entire Stream docs

adamschwartz:
@renandincer Should we close this then?

renandincer:
Yep! Let's discuss on #149!

PR #3, opened 2020-09-29
Updating go example for signed urls

kellydunn:
When attempting to integrate our backend systems to use Cloudflare Stream, we encountered a compiler issue when attempting to use the examples currently in the documentation.

This PR updates the golang implementation so that it can compile.  

Cheers!


renandincer:
Thank you!

PR #44, opened 2020-10-04
Make style consistent in 'Using Durable Objects'

mappum:
I had noticed there were style inconsistencies, so I fixed the following:
- Use of semicolons
- Indentation (2 vs 4 spaces)
- Space before opening bracket of methods

Also, would be able to give great feedback if accepted into the Durable Objects private beta :pleading_face::point_right::point_left:.


rita3ko:
Looks good to me!

PR #59, opened 2020-10-09
fix typo webpack.staging.js

outloudvi:
The `webpack.production.js` seems to be a typo for `webpack.staging.js`.


rita3ko (approved):
Looks good! Thanks @outloudvi !

PR #63, opened 2020-10-10
update access, tunnel, resolver docs to current. rename _index to index

TownLake:
**Change log**
1. Brought Access, Argo Tunnel, 1.1.1.1 sections to current from the snapshot taken when the docs project began. Will consider this repo authoritative going forward. This is a significant change because we recently rewrote all of Access and Tunnel - although this PR mostly just lifts and drops those changes.

2. Updated `_index.md` files to `index.md`

I did not update the Welcome pages yet. I want to port over the current content before tweaking those.

**Notes to contributors (and self)**

* Had to manually run `npm install cf-gatsby-theme` to get `develop` to work. Otherwise an SSR error complains about looking for the gatsby theme.
* If you don't rename `index.md` files, nothing will display in sidebar.


TownLake:
cc @adamschwartz @dknecht and @abracchi-tw

adamschwartz (changes requested):
This is so awesome. Thank you for putting this together!

Re your two bullets:

- You _should not_ `npm install cf-gatsby-theme`. Instead you should remove all instances of `import {Notice} from 'cf-gatsby-theme'` and replace all `<Notice/>` components with the best equivalent [`<Aside/>`, documented here](https://developers.cloudflare.com/docs-engine/reference/markdown#asides) (either naked or `type="warning"`).

  - You should remove `"cf-gatsby-theme": "^0.3.10",` from the `package.json` files.

- Yes, prefixing a Markdown file name with an `_` will create a [“partial”, as documented here](https://developers.cloudflare.com/docs-engine/reference/pages#partials). For future readers, you can run this command to speed up the process:

    ```bash
    find . -type f -name _index.md -exec sh -c 'folder="$(dirname "{}")"; filename="$(basename "{}")"; mv "$folder/$filename" "$folder/index.md"' \;
    ```

Other important things to note:

- Pages must have an `h1`. The previous docs used the `title` frontmatter property as an `h1`. This was backwards and led to us having hundreds of invalid Markdown files. If a page has only a `title` frontmatter property, you should use that to set an `h1` and remove that if from the frontmatter if you want the title in the sidebar to be the same. Only use the `title` frontmatter property when you want the title in the sidebar to be something different from the `h1`. This is useful, e.g. on the docs homepage, where you want something like “Home” or “Welcome” in the sidebar, but “Cloudflare Access documentation” as the `h1` (e.g.). Details of [how the title works are documented here](https://developers.cloudflare.com/docs-engine/reference/pages#title).

- The `weight` frontmatter property should be changed to `order`.

- All instances of frontmatter `hidden: false` should be removed. It’s so bizarre to me that this was necessary before. If you want to hide a page from the sidebar, use `hidden: true`. If you want a page’s children to be hidden (e.g. the [Cloudflare Workers docs Examples section](https://developers.cloudflare.com/workers/examples)), use `hiddenChildren: true`. [These properties are documented here](https://developers.cloudflare.com/docs-engine/reference/pages#hidden-and-hidechildren).

- All instances of frontmatter `showNew: false` and `alwaysopen: true` can be removed. They do nothing.

- All internal links need to lose their respective ([`pathPrefix` documented here](https://developers.cloudflare.com/docs-engine/reference/configuration#properties)). In other words, within the `/access` docs, a link to `/access/about` (like `checkout out our [about](/access/about) page` in Markdown) needs to be just `/about` (so `checkout out our [about](/about) page`). When linking from one docs site to another, use the full URL (so within the Access docs, a link to the Workers tutorials page should be `https://developers.cloudflare.com/workers/tutorials/` in full).

Note: I worked through all of these yesterday in a video call with @abracchi-tw, so she may already have a better branch in progress for `/access`.

-------------------

I also wanted to add, re:

> I did not update the Welcome pages yet. I want to port over the current content before tweaking those.

This is very smart 👍

adamschwartz:
I also wanted to add, re:

> I did not update the Welcome pages yet. I want to port over the current content before tweaking those.

This is very smart 👍

TownLake:
Updated the PR with the front matter changes, checked for links as well. Unless there's other issues, we're ready to merge this over to staging and then can discuss cutting over developers.cloudflare. Thanks again for the help!

PR #69, opened 2020-10-12
Only deploy products with changed files

kristianfreeman:
This PR improves the workflow process by using the `paths-filter` workflow to conditionally build jobs only if a path has changed related to that product.

It also adds `.github/workflows/build.yml`, which runs on each commit to the repo — similar to deploy, it will only run for a product if a change has been found.

Finally, I've bumped the version of https://github.com/actions/checkout to v2 — this doesn't seem to cause any breaking changes and is required for use with `paths-filter`.

For proof of this working, check out the following commits:

- 99c252c is a configuration update to the workflows themselves. Because no changes were made inside of `products/**`, the workflow finishes in 40 seconds and skips every product workflow.
- 46b6146 is a test commit (removed from this branch history, but still available in the GitHub UI) inside of `products/1.1.1.1`. In the workflow, the `1.1.1.1` docs are built, while the rest of the products are skipped. This can be confirmed by viewing the [Checks page for this commit](https://github.com/cloudflare/cloudflare-docs/pull/69/checks?sha=46b6146)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,638 @@
+name: Build
+
+on:
+  - push
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      onedot: ${{ steps.filter.outputs.onedot }}
+      access: ${{ steps.filter.outputs.access }}
+      analytics: ${{ steps.filter.outputs.analytics }}
+      argo-tunnel: ${{ steps.filter.outputs.argo-tunnel }}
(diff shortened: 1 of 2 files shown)


adamschwartz:
Wow this is so cool. Does it work? (Have you tested on a test repo?)

signalnerve:
> Wow this is so cool. Does it work? (Have you tested on a test repo?)

yep, it does! I just removed the draft status 👍 

@adamschwartz gave me the OK to merge this so I'm going to do that shortly!

PR #74, opened 2020-10-13
fix orders and int

TownLake:


TownLake:
I think this fixes the last issue @adamschwartz and @dknecht

adamschwartz (approved):

PR #87, opened 2020-10-14
Update README for Teams products

TownLake:


adamschwartz:
Need to add prod links as well.

TownLake:
I was going to wait to add those until we migrate these to developers.cloudflare, but we can just hold until that's done then.

PR #100, opened 2020-10-16
Add DNS over Discord to 1.1.1.1 Fun with DNS 

MattIPv4:
👋 The 1.1.1.1 Discord bot launched a while back but had been held back from being added to the docs due to this new version launching.

I've worked in the past with Matthew Gall & assorted folks from support/marketing to make sure we're on the same page with this -- I hope everything is still okay and this can be added.

I wasn't sure about whether to link to the open-source code for the bot (https://github.com/MattIPv4/1.1.1.1-Discord) as it felt a bit too self-promotion-y, so I haven't included a link for now.

Also, two other quick fixes in separate commits :)


TownLake (approved):
We keep this section fairly light-hearted and this fits within that; thanks for submitting the example.

PR #118, opened 2020-10-20
Added warning + awareness about wrangler secret

IgnorantSapient:
Some users might think about adding environment variables to hold their secrets and might store it on their .toml file. 
A warning will add a layer of protection and it will also redirect people towards the right way to store secrets.

### DIFF
--- a/products/workers/src/content/platform/environments.md
+++ b/products/workers/src/content/platform/environments.md
@@ -263,7 +263,7 @@ vars = {FOO = "some text"}
 
 <Aside type="warning">
 
-\* __Warning:__ We do not recommend using text variables to store secrets. If possible use the `wrangler secret put` command instead.
+\* __Warning:__ We do not recommend using text variables to store secrets. If possible use the [`wrangler secret put`](/cli-wrangler/commands#secret) command instead.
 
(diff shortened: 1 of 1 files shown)


products/workers/src/content/platform/environments.md
```diff
@@ -261,6 +261,12 @@ kv_namespaces = [
 vars = {FOO = "some text"}
 ```
 
+<Aside type="warning">
+
+\* __Warning:__ Do **not** use text varibales to hold secrets like API keys, tokens etcetra, use the [`wrangler secret put`](/cli-wrangler/commands#secret) command instead.
```
rita3ko, line 266: Hi @123aswin123, thank you for the contribution! Would you mind updating the language to make it more of a recommendation (e.g. "We do not recommend using text variable to store secrets.... If possible use the wrangler secret put` command instead.")? Otherwise, looks good! 

Another good example here: https://developers.cloudflare.com/workers/cli-wrangler/authentication and here

123aswin123: Thanks for your recommendation, I have fixed the language. around this.

PR #143, opened 2020-10-22
Update Tunnel docs on service mode in Named Tunnels

TownLake:


products/argo-tunnel/src/content/create-tunnel/run-as-service.md
```diff
@@ -4,14 +4,19 @@ order: 30
...
-Argo Tunnel can install itself as a system service on Linux and Windows and as a launch agent on macOS.
+|Argument|Description|
+|---|---|
+|`url`|The destination for proxied traffic in your environment|
+|`tunnel`|The name or UUID of your Tunnel
+|`credentials-file`|The location of the credentials file for your Tunnel|
 
-By default, Argo Tunnel expects all of the configuration to exist in the `cloudflared/config.yml` configuration file. The available options are documented on the [configuration file reference](/configuration/config/), but at a minimum you should set `hostname` and `url`.
+<Aside>
+The arguments above are required for pre-configured Argo Tunnel deployments. If you are using legacy Argo Tunnel, without names, you can append the `--legacy` flag when running cloudflared tunnel install command.
```
adamschwartz, line 18: There needs to be whitespace above this line (and for symmetry, we recommend below as well), or it won’t be parsed as Markdown. (So e.g. `--legacy` won’t be properly turned into an inline code block.)

See: https://developers.cloudflare.com/docs-engine/reference/markdown#asides

adamschwartz (changes requested):
One small issue with an `<Aside/>` (see below). Otherwise LGTM! Thanks

products/argo-tunnel/src/content/create-tunnel/run-as-service.md
TownLake: Fixed!

PR #182, opened 2020-10-27
Fix tabulation in "fit" properties

maguayo:
Fix the tabulation in "FIT" property inside the "Image Resizing" docs.


adamschwartz:
For future reference: https://github.com/cloudflare/cloudflare-docs/commit/9616fba92f391ff5b1bc1e8b85628789340daf5a

PR #235, opened 2020-11-04
Announce 2020-11-04 WAF changes

gyx-cf:


migueldemoura (changes requested):
This one should follow the "emergency" release template.

migueldemoura (approved):

PR #236, opened 2020-11-04
Update/access gsuite identify provider

fniko:
I have updated workflow for configuring Access for GSuite (now Google Workspaces) Identity Provider.
Google made some changes and I was unable to reproduce steps.

So I have updated text and add new images.


abracchi-tw:
@fniko closing this, thanks so much for your work!

PR #253, opened 2020-11-07
Fix Access-Control-Allow-Origin on final response

handlebauer:
I may be off-base here but I propose using `*` instead of `url.origin` because `url.origin` represents the worker's origin, not the user's. This means the user's origin won't match the response's, etc., which is contrary to the whole point of things, if the point is enabling your local machine's browser to fetch data from a CORS-restricted API.

### DIFF
--- a/products/workers/src/content/examples/cors-header-proxy.md
+++ b/products/workers/src/content/examples/cors-header-proxy.md
@@ -111,7 +111,7 @@ async function handleRequest(request) {
   response = new Response(response.body, response)
 
   // Set CORS headers
-  response.headers.set("Access-Control-Allow-Origin", url.origin)
+  response.headers.set("Access-Control-Allow-Origin", '*')
 
   // Append to/Add Vary header so browser will cache response correctly
   response.headers.append("Vary", "Origin")
--- a/products/workers/src/content/examples/cors-header-proxy.md
+++ b/products/workers/src/content/examples/cors-header-proxy.md
@@ -112,6 +112,8 @@ async function handleRequest(request) {
 
   // Set CORS headers
   response.headers.set("Access-Control-Allow-Origin", '*')
+  // Note: If you wish to serve the worker in front of an existing origin, 
+  // you may wish to change the above value fron `*` to `url.origin`
 
   // Append to/Add Vary header so browser will cache response correctly
   response.headers.append("Vary", "Origin")


kentonv:
This seems right to me but I didn't write the example so I'm not sure what it was going for exactly. @signalnerve do you know who wrote this example?

Another option would have been to set it to the value of the incoming `Origin` header on the request. But given that the `OPTIONS` request handler sets ACAO to `*`, it seems like that's what the regular request handler should do as well.

bulk88:
I sort of agree with the change, but read this to the end. '*' the wildcard vs explicit URL origin only matters for XMLHttpRequest.withCredentials (send Cookies). Since this simple cors proxy demo DOES NOT authenticate/filter referrers or origins, and does not send Access-Control-Allow-Credentials=true. It should use the wildcard * origin for consistency with other CORS proxies from frameworks on github since this CFW CORS proxy demo does not implement ACAC=true. If the demo did ACAC=true, the `url.origin` parameter is still probably incorrect unless the CFW CORS proxy's domain is a "router" and does display "navigate" able HTML pages. CORS spec requires 

`
response.headers.set('Access-Control-Allow-Origin', request.headers.get("Origin"))
`

url.origin will be `https://thissite.thisuser.workers.dev`, while chances are the browser is on "http://mystaticsite.com" and mystaticsite.com is using the CFW CORS proxy at thissite.thisuser.workers.dev to scrape legacythirdparty.com, and thissite.thisuser.workers.dev will return ACAO=thissite.thisuser.workers.dev to mystaticsite.com's AJAX req and the browser will turn mystaticsite.com's AJAX req into xhr.statusCode=0/timeout CORS violation in browser response object. because ACAO is not * and is not mystaticsite.com. On the other hand, perhaps keeping `url.origin` in the CFW CORS demo, is an intentional typo, so any dev that creates a private instance has to do 5 mins of work and recognize what the code actually does before creating an open proxy instance 😈

The CORS proxy demo as written today ONLY works if the proxy code/worker is a route for a orange cloud domain that also serves navigable HTML and the only user of the proxy is that orange cloud domain. Anti-open proxy. If a dev wants it to make the demo be an open proxy he needs to spend 5-10 mins reading the CORS spec and change the line to

`
let clientOrigin = request.headers.get("Origin");
//create new request object that can be modified
request.headers.delete("Origin");
...
await fetch(request)
...
response.headers.set('Access-Control-Allow-Origin',clientOrigin)
`

or 

`
response.headers.set('Access-Control-Allow-Origin','*')
`

I'd suggest keeping the demo with `url.origin` in the code, and a `//` comment saying to change it to `*` if the dev really needs an open proxy, anti-copy paste by minimally trained devs/IT folks.

Theoretically, a blackhat could try to find some customer's CFW open proxy instance on CF's network, and drain that customer's 10M req/mo quota with reqs to a target victim server. I'm not sure if the attack's point is to DOS the middle customer by $ draining quota, or CPU/bandwidth exhaust origin victim server. And both middle proxy victim and origin server victim are probably CF orange anyways, and a CFW AFAIK, on its `fetch()` still will go 1 more time through WAF to origin (is this skipped if CFW & origin are same "zone" on CF) only question is which will block/go offline/WAF trigger first, the CFW proxy or origin. CFW proxy will go offline first, since orange cloud origin servers usually have no pay-per-req on their CF plans.

kristianfreeman:
Hey all 👋 

> @signalnerve do you know who wrote this example?

This has probably been ported through a few iterations of our docs.

> The CORS proxy demo as written today ONLY works if the proxy code/worker is a route for a orange cloud domain that also serves navigable HTML and the only user of the proxy is that orange cloud domain.

This is correct. This CORS proxy configuration is very specific for a certain kind of website -- it's common to see CORS proxy setups use `*` for `Access-Control-Allow-Origin` to just enable CORS more generally, but if I had to guess the history behind this example, the idea is to build on top of Workers living on the edge and letting you build a CORS proxy _in front_ of an existing site.

> I'd suggest keeping the demo with `url.origin` in the code, and a `//` comment saying to change it to `*` if the dev really needs an open proxy, anti-copy paste by minimally trained devs/IT folks.

This is interesting because I think I'm favor of the swapped version of this -- moving to `*` as seen in this example, and adding some documentation/commented-out code to say "hey, you can also use `url.origin` if you're serving this in front of an existing origin". 

Because of that, I think this PR is on the right track (and apologies for the delay!) -- cc @duncanlettuce can you add a comment below the line you changed to indicate that `url.origin` is an option and what purpose it serves? Thanks!

handlebauer:
👋

@signalnerve, that sufficient? I'm not familiar with the very specific kind of context/website you mention, so I kept the comment brief and just reiterated what you said.

kentonv:
If I understand correctly, setting `Access-Control-Allow-Origin` to `url.origin` will always be equivalent to not setting the header at all, so suggesting that in a comment doesn't seem right. But setting it to `request.headers.get("Origin")` could make sense.

lukeed:
Generally you only want to use a `url.origin` or a `req.headers.get('Origin')` value _after_ doing some allow/block logic. Having it directly echo serves no CORS benefit.

Also, with this change to using `"*"`, the `Vary, Origin` line below it should be removed, as it's no longer a varying/relevant value: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#cors_and_caching

PR #286, opened 2020-11-13
Stream: Clarify usage of `name` in the TUS upload metadata

renandincer:


adamschwartz (approved):

products/stream/src/content/uploading-videos/upload-video-file.md
```diff
@@ -55,7 +55,13 @@ The tus protocol allows you to add optional parameters [in the `Upload-Metadata`
 
 ### Supported options in "Upload-Metadata"
 
+Setting arbitrary metadata values in the `Upload-Metadata` header sets values the [meta key in Stream API](https://api.cloudflare.com/#stream-videos-properties).
+
 <Definitions>
+	
+- `name`
```
Everlag, line 62: In dash, I see `filename c21hbGwubXA0` in `Upload-Metadata` when I upload a video; I think this is supposed to be `filename`?

Everlag (changes requested):

products/stream/src/content/uploading-videos/upload-video-file.md
Everlag: Actually, while looking at unrelated work, I see `name` in `Upload-Metadata` as well. Uppy seems to send both and the UI picks to show `name`.

So, you're right here, some of the internal code I was looking at is wrong.

adamschwartz:
@renandincer Is this OK to merge?

PR #291, opened 2020-11-16
PCX-963 Updated session affinity

aaron-ruby-df:
## User story

Update load-balancing/understand-basics/session-affinity so that clients can easily understand that

* TCP is not a valid method for session affinity
* the ip-cookie is available and add link to API ref


aaron-ruby-df:
@adamschwartz Added you, since I can't find an appropriate CODEOWNER who is in Github.

adamschwartz (approved):

PR #298, opened 2020-11-17
Fix typo

delroh:
This or maybe "retrieve and put data in the cloudflare cache" ? not sure.

### DIFF
--- a/products/workers/src/content/learning/getting-started.md
+++ b/products/workers/src/content/learning/getting-started.md
@@ -172,7 +172,7 @@ This starter is used in the tutorial for [building a Slack Bot](/tutorials/build
 
 ### 4c. Make use of runtime APIs
 
-The example outlined in this guide is just a starting point. There are many Workers [runtime APIs](/runtime-apis) available to manipulate requests and generate responses. For example, you can use [HTMLRewriter](/runtime-apis/html-rewriter) to parse and transform HTML on the fly, use the [Cache](/runtime-apis/cache) to retrieve data from and put data into from [the Cloudflare cache](/learning/how-the-cache-works), compute a custom response right from the edge, redirect the request to another service, and so much more.
+The example outlined in this guide is just a starting point. There are many Workers [runtime APIs](/runtime-apis) available to manipulate requests and generate responses. For example, you can use [HTMLRewriter](/runtime-apis/html-rewriter) to parse and transform HTML on the fly, use [Cache](/runtime-apis/cache) to retrieve data from and put data into  [the Cloudflare cache](/learning/how-the-cache-works), compute a custom response right from the edge, redirect the request to another service, and so much more.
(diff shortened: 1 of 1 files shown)


products/workers/src/content/learning/getting-started.md
```diff
@@ -172,7 +172,7 @@ This starter is used in the tutorial for [building a Slack Bot](/tutorials/build
 
 ### 4c. Make use of runtime APIs
 
-The example outlined in this guide is just a starting point. There are many Workers [runtime APIs](/runtime-apis) available to manipulate requests and generate responses. For example, you can use [HTMLRewriter](/runtime-apis/html-rewriter) to parse and transform HTML on the fly, use the [Cache](/runtime-apis/cache) to retrieve data from and put data into from [the Cloudflare cache](/learning/how-the-cache-works), compute a custom response right from the edge, redirect the request to another service, and so much more.
+The example outlined in this guide is just a starting point. There are many Workers [runtime APIs](/runtime-apis) available to manipulate requests and generate responses. For example, you can use [HTMLRewriter](/runtime-apis/html-rewriter) to parse and transform HTML on the fly, use [Cache](/runtime-apis/cache) to retrieve data from and put data into  [the Cloudflare cache](/learning/how-the-cache-works), compute a custom response right from the edge, redirect the request to another service, and so much more.
```
kristianfreeman:
```suggestion
The example outlined in this guide is just a starting point. There are many Workers [runtime APIs](/runtime-apis) available to manipulate requests and generate responses. For example, you can use [HTMLRewriter](/runtime-apis/html-rewriter) to parse and transform HTML on the fly, use the [Cache API](/runtime-apis/cache) to retrieve data from and put data into [the Cloudflare cache](/learning/how-the-cache-works), compute a custom response right from the edge, redirect the request to another service, and so much more.
```

kristianfreeman (changes requested):
thanks for removing that extra word, little suggestion for a diff part of the paragraph

products/workers/src/content/learning/getting-started.md
delroh: Great, will do! Should "API" also be added to HTMLRewriter? Initially I added "API" to Cache but then removed it because I was not sure if it was intentionally omitted in HTMLRewriter, given that it is talking about available runtime APIs and therefor all subsequently mentioned are APIs. Got a little tangled there on should add or not so just followed the omitting like on HMLTRewriter.

kristianfreeman: oh good question - sure, let's go with "use the HTMLRewriter API". thanks!!

delroh: Almost forgot about this. Added the suggested changes. 😁

adamschwartz (approved):

PR #304, opened 2020-11-17
You can do an argo tunnel for the Apex domain

kcantrel:
Updated the FAQ to reflect the fact that you can now do an Argo tunnel for the apex domain, since we allow CNAMEs at the apex, and it can point to the named tunnel. I tested this and it works just fine.

Also created a typo where there was one too many "#" for the "How can origin servers be secured when using Tunnel?" line.

- Keith Cantrell/SE
- keith@cloudflare.com


abracchi-tw:
Hey @kcantrel sorry for overlooking this - opened #828 to apply these changes to the new Teams docs. Thanks!!

PR #307, opened 2020-11-17
Remove Workers Bundled requirement for using Sites

kristianfreeman:
Because KV is now free, I made a quick pass through the Workers Sites docs to confirm that we don't mention needing Workers Bundled to use it.


adamschwartz (approved):
Awesome. Thank you!

adamschwartz:
We should update https://workers.cloudflare.com/sites as well.

PR #312, opened 2020-11-18
[Gateway] security threats and content categories docs

abracchi-tw:
- Added step-by-step guides on how to select categories and test policies
- Added a learning section as reference, with categories + definitions


TownLake (approved):
These match what we publish in the UI today.

PR #359, opened 2020-11-25
Add note saying that NameCheap doesn't allow you to expedite transfers

Moire9:
The email they send you only contains a link to *cancel* the transfer, and not one to expedite the transfer. You have to wait 5 days.


Moire9:
WTF WHY DID EVERYTHING GET CHANGED

Moire9:
Well then, let's just ignore that that happened...

KimJ15:
Commit addressed by PR referenced above.

PR #360, opened 2020-11-25
Remove query parameters added by tus uploader

alexfedosov:
Tus uploader might add query parameters which should be removed when parsing mediaId from response url


renandincer:
A better solution could be to use the `Stream-Media-ID` header in the initial TUS request.

renandincer:
https://github.com/cloudflare/cloudflare-docs/pull/445

PR #364, opened 2020-11-25
Fix docs to include that Workers KV is now also included in the free plan

yvessavoy:
Was a small change to say that it's included in both the free and the bundled plan. Wasn't sure about adding the numbers included in the free tier so I just excluded them for now.


greg-mckeon:
Thanks for the catch @yvessavoy!

greg-mckeon (approved):

PR #366, opened 2020-11-25
Added stronger callouts for ScheduledEvent listener

sgraden:
Added an aside in Cron Triggers to create a stronger callout of this requirement that is easier to spot and reference than the background description.

Added small change to ScheduledEvent listener to callout earlier that it works with Cron Triggers.

### DIFF
--- a/products/workers/src/content/runtime-apis/scheduled-event.md
+++ b/products/workers/src/content/runtime-apis/scheduled-event.md
@@ -2,7 +2,7 @@
 
 ## Background
 
-The event type for Scheduled requests to a Worker. The `Object` passed through as `event` in [`addEventListener()`](/runtime-apis/add-event-listener).
+A ScheduledEvent is an event type invoked by a [Cron Trigger](/platform/cron-triggers). `Scheduled` is the `Object` passed through as `event` in [`addEventListener()`](/runtime-apis/add-event-listener).
 
 ## Context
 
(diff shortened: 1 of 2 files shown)


products/workers/src/content/runtime-apis/scheduled-event.md
```diff
@@ -2,7 +2,7 @@
 
 ## Background
 
-The event type for Scheduled requests to a Worker. The `Object` passed through as `event` in [`addEventListener()`](/runtime-apis/add-event-listener).
+A ScheduledEvent is an event type invoked by a [Cron Trigger](/platform/cron-triggers). `Scheduled` is the `Object` passed through as `event` in [`addEventListener()`](/runtime-apis/add-event-listener).
```
kristianfreeman: What is `Scheduled` in reference to? It seems like the `event` object here is a `ScheduledEvent` type. The phrasing of "`Scheduled` is the `Object` passed through as `event`" is a little confusing to me...

sgraden: Yeah same here a bit actually. Is this more clear?


> A ScheduledEvent is the event type for Scheduled requests to a Worker. It is the `Object` passed through as the 'event' when a Workers `scheduled` event listener is invoked by a Workers Cron Trigger. 


My goal here is to just mention "Cron Triggers" in this passage so they are more immediately linked to each other. Making the background more clear isn't a bad idea either.

kristianfreeman (approved):

hui37210:
ok

hui37210:
ok

PR #377, opened 2020-11-27
Add link to IATA code for `colo` in request docs

12932:
A bit nit-picky but less ambiguous than "three letter airport code"


adaptive:
That information is incorrect, ICAO code is 4 letters.

12932:
Oh you're right, I have changed it to IATA code instead

kristianfreeman:
it seems like this is an IATA code, but i don't know for sure! cc @rita3ko who may know definitively

haneefmubarak (approved):
LGTM, squashing and merging.

PR #405, opened 2020-12-02
PCX-645 Update Logs docs to mention firewall_events dataset for Logpush API

soheiokamoto:
This updates `Logs` docs to mention `firewall_events` dataset for Logpush API.


tlozoot (approved):
Thanks for doing this!

PR #418, opened 2020-12-03
updates links to correct "production"-branch

frusznyak:
GitHub is (rightfully) discouraging the use of the Term `master`, this repo no longer uses them anyway, so we can change some links to the correct `production` branch without loosing anything

### DIFF
--- a/products/docs-engine/src/content/index.md
+++ b/products/docs-engine/src/content/index.md
@@ -25,16 +25,16 @@ It’s __open-source__ and [available on GitHub](https://github.com/cloudflare/c
 
 <TableWrap>
 
-| Docs site                                                             | GitHub                                                                                                                        |
-|-----------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------|
-| [Minimal example](https://docs-engine-example.adam.workers.dev)       | [@adamschwartz/docs-engine-example](https://github.com/adamschwartz/docs-engine-example)                                      |
-| [Docs Engine](https://docs-engine-docs.adam.workers.dev) — This site. | [@cloudflare/cloudflare-docs/.../docs-engine](https://github.com/cloudflare/cloudflare-docs/tree/master/products/docs-engine) |
-| [Access](https://developers.cloudflare.com/access)                    | [@cloudflare/cloudflare-docs/.../access](https://github.com/cloudflare/cloudflare-docs/tree/master/products/access)           |
-| [Images](https://developers.cloudflare.com/images)                    | [@cloudflare/cloudflare-docs/.../images](https://github.com/cloudflare/cloudflare-docs/tree/master/products/images)           |
-| [Stream](https://developers.cloudflare.com/stream)                    | [@cloudflare/cloudflare-docs/.../stream](https://github.com/cloudflare/cloudflare-docs/tree/master/products/stream)           |
-| [Terraform](https://developers.cloudflare.com/terraform)              | [@cloudflare/cloudflare-docs/.../terraform](https://github.com/cloudflare/cloudflare-docs/tree/master/products/terraform)     |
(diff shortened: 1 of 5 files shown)


products/docs-engine/src/content/reference/configuration.md
```diff
@@ -118,7 +118,7 @@ module.exports = {
 
 --------------------------------
 
-Here’s the `docs-config.js` file for [these docs](https://github.com/cloudflare/cloudflare-docs/tree/master/products/docs-engine):
+Here’s the `docs-config.js` file for [these docs](https://github.com/cloudflare/cloudflare-docs/production/master/products/docs-engine):
```
adamschwartz: Looks like you changed `tree` here by accident (should be `tree/production/...`).

products/docs-engine/src/content/index.md
```diff
@@ -28,12 +28,12 @@ It’s __open-source__ and [available on GitHub](https://github.com/cloudflare/c
...
+| [Access](https://developers.cloudflare.com/access)                    | [@cloudflare/cloudflare-docs/.../access](https://github.com/cloudflare/cloudflare-docs/tree/production/products/access)           |
+| [Images](https://developers.cloudflare.com/images)                    | [@cloudflare/cloudflare-docs/.../images](https://github.com/cloudflare/cloudflare-docs/tree/production/products/images)           |
+| [Stream](https://developers.cloudflare.com/stream)                    | [@cloudflare/cloudflare-docs/.../stream](https://github.com/cloudflare/cloudflare-docs/tree/production/products/stream)           |
+| [Terraform](https://developers.cloudflare.com/terraform)              | [@cloudflare/cloudflare-docs/.../terraform](https://github.com/cloudflare/cloudflare-docs/tree/production/products/terraform)     |
+| [Workers](https://developers.cloudflare.com/workers)                  | [@cloudflare/cloudflare-docs/.../workers](https://github.com/cloudflare/cloudflare-docs/tree/production/products/workers)         |
```
adamschwartz: Please clean up the table spacing here to accommodate the change.

adamschwartz (changes requested):
Thanks for doing this! Please correct the issues below.

products/docs-engine/src/content/reference/configuration.md
frusznyak: Absolutely right, sorry for that oversight!

products/docs-engine/src/content/index.md
frusznyak: fixed! 🙂

frusznyak:
Thanks for the fast review!

Additionally one could also argue, that the automatically generated "Edit on GitHub" link should/could also be updated, see the [corresponding line](https://github.com/cloudflare/cloudflare-docs-engine/blob/64ac0c5fa58c4a33b1b29ec830c7824accee131e/src/components/docs-footer.js#L15), but i cannot judge if that is a worthwhile endeavour since an additional variable (e.g. `${contentRepoBranch}`) would likely have to be introduced to the docs and docs-engine.
This is currently not a real issue (since `master` links will [automatically redirect](https://github.com/github/renaming#whats-already-changed) to the default branch) but any additional legacy-redirect that can be eliminated is a good reason to go the extra mile.
If you find this to be of value, i'd love to provide a PR for that as well.

adamschwartz:
@frusznyak It’s a great point. Funny you should mention too because @adaptive just submitted a PR for this a few days ago: https://github.com/cloudflare/cloudflare-docs-engine/pull/308.

The reason I haven’t merged it yet is precisely for the point you mentioned. I’d like to tweak the PR to introduce a configuration for docs config, so I’d like to take the time to think that through and document it properly. Perhaps it can default to `main` (see https://github.com/github/renaming) if not specified in `docs-config.js`.

adamschwartz (approved):

PR #423, opened 2020-12-04
[Teams] add one-tutorials component

abracchi-tw:
@adamschwartz I added a new component to `cloudflare-one` (`one-tutorials.js`). 
Had to tweak `gatsby-node.js` to add a field ( `category: String` ) to `type Frontmatter`. Everything works locally, but of course this change is outside `src/content` so it doesn't show up when I commit/push. Something might break?

### DIFF
--- a/products/cloudflare-one/src/content/components/code-block.js
+++ b/products/cloudflare-one/src/content/components/code-block.js
@@ -1,0 +1,121 @@
+import React from "react"
+
+import * as frontMatterParser from "gray-matter"
+
+import Highlight, { defaultProps } from "prism-react-renderer"
+
+import { languageMappings, prismLanguages, transformToken } from "./custom-syntax-highlighting"
+
+// Additional language support (See https://git.io/JJuZX)
+import Prism from "prism-react-renderer/prism"
+(typeof global !== "undefined" ? global : window).Prism = Prism
+require("prismjs/components/prism-toml")
+require("prismjs/components/prism-rust")
+Prism.languages.sh = prismLanguages.sh
+
+const codeBlockClassName = theme => {
+  const themeClassName = theme === "dark" ? "" : " CodeBlock-is-light-in-light-theme"
+  return `CodeBlock CodeBlock-with-rows CodeBlock-scrolls-horizontally${themeClassName}`
+}
+
+const addNewlineToEmptyLine = line => {
+  if (line && line.length === 1 && line[0].empty) {
+    // Improves copy/paste behavior
+    line[0].content = "\n"
+  }
+
+  return line
+}
+
+const CodeBlock = props => {
+  const { className, children } = props.children.props
+
+  if (props.className) {
+    return (<pre {...props}/>)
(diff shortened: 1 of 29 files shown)


TownLake (approved):

TownLake (approved):

products/cloudflare-one/src/content/components/code-block.js
```diff
@@ -0,0 +1,121 @@
...
+              </span>
+            </span>
+          </code>
+        </pre>
+      )}
+    </Highlight>
+  )
+}
+
+export default CodeBlock
```
adamschwartz, line 121: Why is this file being included? It would be better to import it from https://github.com/cloudflare/cloudflare-docs-engine/blob/master/src/components/mdx/code-block.js if possible so we’re not duplicating code.

products/cloudflare-one/src/content/components/one-tutorials.js
```diff
@@ -0,0 +1,116 @@
...
+            </div>
+            <div className="OneTutorials--column" data-column="category">{tutorial.category}</div>
+          </div>
+        ))}
+      </div>
+    </div>
+  )
+}
+
+export default OneTutorials
```
adamschwartz, line 116: Rather than build a custom component for Cloudflare One tutorials, I think we would be better off making `category` option for any tutorials list, building that into https://github.com/cloudflare/cloudflare-docs-engine/blob/master/src/components/docs-tutorials.js so that it’s reusable by Workers and One and more products over time. I can help with this next week. (This will also address the `gatsby-node.js` issue you mentioned by the way.)

adamschwartz (changes requested):
Thanks @abracchi-tw for putting this together! I’m really excited to see this when it’s done.

Please see my comments below.

I can help refactor this to make use of the engine’s `docs-tutorials.js` directly, augmenting that component with additional optional table columns as needed for Cloudflare One.

products/cloudflare-one/src/content/components/one-tutorials.js
adamschwartz: `category` was the only addition I spotted, but please let me know what else you’d like added.

products/cloudflare-one/src/content/components/code-block.js
abracchi-tw: yeah, I think I left it there by mistake, no point in duplicating it

products/cloudflare-one/src/content/components/one-tutorials.js
abracchi-tw: @adamschwartz thanks for your input! sure, let's build `category` into `docs-tutorials.js`. I just have a question so that I know what to do moving forward: when do you think it makes sense to add custom components to `src/content/components` and when should we think cross-product instead? personally, I'm all for adding new components to a centralized place, so that all products can use them when needed (I just don't think I know how to do that, but that's another story :D)

PR #428, opened 2020-12-04
Update publish-durable-object comment about chicken-and-egg problem

a-robinson:
Now that you can actually create a durable object namespace without a
script assigned to it. I'm not bothering to modify the code, though,
since we still need 3 requests either way and it would make updating an
existing Worker using the script slightly more difficult.

Also fix a minor typo while I'm here.

@greg-mckeon @bretthoerner


bretthoerner (approved):

a-robinson:
@greg-mckeon I don't have merge permissions here, so can you merge it when you get a chance?

greg-mckeon (approved):

PR #440, opened 2020-12-07
WAF announcement 2020-12-07

ggaborCF:


helloandre (approved):

products/waf/src/content/change-log/scheduled-changes.md
```diff
@@ -32,5 +32,40 @@ order: 1
         </tr>
     </thead>
     <tbody>
+        <tr>
```
migueldemoura, line 35: There should only be one `<tr>` here, you can preview this to see how it looks.

PR #442, opened 2020-12-07
[1.1.1.1] Fix cloudflared proxy-dns system service instructions

Lekensteyn:
`cloudflared service install` does not work on Linux (TUN-3520) when
trying to configure a proxy-dns service. Aside from that, the extra
auto-updater service installed by that command might not be desirable
either. Suggest the installation of an unprivileged systemd service with
the correct dependencies to ensure that it starts as early as possible
to avoid DNS resolution failures, or fallback to other nameservers in
/etc/resolv.conf.

Also suggest modifying `/etc/resolv.conf`, that should almost always
work. Specific Linux distributions can provide instructions for their
platforms.

### DIFF
--- a/products/1.1.1.1/README.md
+++ b/products/1.1.1.1/README.md
@@ -2,4 +2,4 @@
 
 [View docs →](https://developers.cloudflare.com/1.1.1.1)
 
-[Read contributor guidelines →](https://developers.cloudflare.com/docs-engine/contributing/content-framework)
+[Read contributor guidelines →](https://developers.cloudflare.com/docs-engine/contributing/content)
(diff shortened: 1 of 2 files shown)


tanyav2 (commented):
What behavior is expected if DoH queries are failing? You can check this by stopping `cloudflared` and then sending `dig` requests to @127.0.0.1. On my computer, the requests still go through. (This is with the default port)

Lekensteyn:
> What behavior is expected if DoH queries are failing? You can check this by stopping `cloudflared` and then sending `dig` requests to @127.0.0.1. On my computer, the requests still go through. (This is with the default port)

There are several reasons why `dig @127.0.0.1 example.com` succeeds even when `cloudflared proxy-dns` is not running:

- Another service is already running on the same port, check with `sudo ss -tlpn | grep :53`
- A firewall rule is installed to forward to another local service, check with `sudo iptables -t nat -vnL`, look REDIRECT rules. These could have been added by some (VPN) software that try to avoid "DNS leaks" by redirecting all outgoing DNS traffic to a single service.

abracchi-tw (approved):

PR #443, opened 2020-12-07
Add Sunder Cloudflare Workers starter template

gzuidhof:
Hi Cloudflare 👋 

This PR adds a non-barebones starter template for Cloudflare workers using modern tech.  

I created it mostly to scratch my own itch as setting up a Workers project that goes beyond hello world and is actually test-friendly gets tiring after 5 times :). I think others will also benefit from this.

Here's an [**example deployment**](https://sunder-starter.sunderjs.com/).

### DIFF
--- a/products/workers/src/content/starters/index.md
+++ b/products/workers/src/content/starters/index.md
@@ -56,6 +56,12 @@ $ wrangler generate <new-project-name> <github-repo-url>
 />
 
 <WorkerStarter
+  title="Sunder Starter (Typescript)"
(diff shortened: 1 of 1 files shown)


kristianfreeman:
this is neat! tbh we get less of these submissions than you might think - i'm inclined to merge and then if we continue to get more community-contributed stuff, we can reconsider how/where to show this on the page. cc @rita3ko

gzuidhof:
Sounds good to me :)

kristianfreeman:
@gzuidhof would you mind rebasing this PR, we should merge this!

PR #460, opened 2020-12-09
Fix spelling

jbampton:


adamschwartz (approved):
Thank you!!! @jbampton These updates are always so appreciated.

PR #464, opened 2020-12-09
Fix Durable Object Counter example for concurrent use.

bretthoerner:
This probably deserves an explainer, I just wanted to get this up before I forgot.

Before this change, concurrent requests to the Actor could easily return some other request's incr/decr'd `this.value`, after `await`ing their own `storage.put`.

### DIFF
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -113,7 +113,7 @@ export class DurableObjectExample {
 }
 ```
 
-Transactions operate at a [serializable isolation level](https://en.wikipedia.org/wiki/Isolation_(database_systems)#Serializable).  This means transactions can fail if they conflict with a concurrent transaction being run by the same Durable Object.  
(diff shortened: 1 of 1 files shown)


greg-mckeon (approved):

greg-mckeon:
can we do an update to the docs as well w/ the explainer?

bretthoerner:
Yeah, I'll flesh one out after I finish this other task.

a-robinson (approved):

kentonv:
Heh, I guess whether or not this is a bug depends on the use case. If you're trying to assign a unique number to each caller, it's a bug. If you're trying to measure how many times a thing has happened, it's not a bug -- though your fix doesn't hurt.

bretthoerner:
Yeah, mostly it's that a user tried to load test it and was confused by seeing duplicates on the client side.

bretthoerner:
@greg-mckeon I added a little explainer, feel free to merge if it works for you.

PR #465, opened 2020-12-09
Fix case of Python and fix spelling

jbampton:


renandincer (approved):
@jbampton  thanks for the contribution!

PR #467, opened 2020-12-09
Fix case of iOS

jbampton:
Changed `ios` to `iOS`


renandincer (approved):
Thank you for this one as well, @jbampton!

PR #468, opened 2020-12-09
Fix case of Android

jbampton:


renandincer (approved):

renandincer:
Thanks for all these fixes, @jbampton!

PR #495, opened 2020-12-14
feat(CI): 🍓 add a GitHub Action to lint the YAML

jbampton:
Add a `.yamllint` config file
Run on pull request only

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,22 @@
+name: ❄️ Lint
+
+on: [pull_request]
+
+jobs:
+  yamllint:
+    name: 🍺 YAML
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'  # Version range or exact version of a Python version to use, using SemVer's version range syntax
+          architecture: 'x64'  # optional x64 or x86. Defaults to x64 if not specified
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install yamllint
+      - name: 🧹 YAML Lint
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,12 @@
...
+on: [pull_request]
+
+jobs:
+  yamllint:
+    name: 🍺 YAML
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: 🧹 YAML Lint
+        uses: ibiqlik/action-yamllint@v3
```
jbampton: This GitHub Action needs to be removed since it is a third party Action.  There are more security risks using third party actions.

Official -> "The individual jobs in a workflow can interact with (and compromise) other jobs. For example, a job querying the environment variables used by a later job, writing files to a shared directory that a later job processes, or even more directly by interacting with the Docker socket and inspecting other running containers and executing commands in them.

This means that a compromise of a single action within a workflow can be very significant, as that compromised action would have access to all secrets configured on your repository, and can use the GITHUB_TOKEN to write to the repository. Consequently, there is significant risk in sourcing actions from third-party repositories on GitHub."

I will need to use the official Python certified GitHub Action and install `yamlllint` manually with `pip`.

https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions
https://github.com/marketplace/actions/setup-python
https://yamllint.readthedocs.io/en/stable/
https://github.com/pypa/pip
https://github.com/marketplace/actions/yaml-lint

KimJ15:
👋 @jbampton Thank you for the feedback. We are hard at work on many things related to the deploy and build process. Going forward, we are only taking care of doc contributions in this repo for developers.cloudflare.com!

PR #501, opened 2020-12-15
Update mentions to Argo Smart Routing in Tunnels now that it is no longer a requirement

nmldiegues:


nmldiegues:
It's not clear if this is the direction we're heading. Let's leave it as is.

PR #526, opened 2020-12-18
Fix repeated word

samrobbins85:
The doc says the branch should be main or main, I suspect one of these should be master


kristianfreeman:
hey thanks! we just fixed this with #529 but nice catch :)

PR #527, opened 2020-12-18
Add a guide for Next.js on Cloudflare Pages

samrobbins85:
Next.js is a very popular react framework, so the guide would be useful


kristianfreeman:
this is SO awesome, thanks so much for doing this! i gotta test this at the beginning of next week, but generally looks good to me after a quick readthrough. such an awesome addition :) appreciate it!

rita3ko (approved):

PR #532, opened 2020-12-19
create-react-app -> vue

ghost:


MannyDiera:
Thanks for clarifying, I was reading this excited but that React mention caught me off guard.

kristianfreeman (approved):

PR #540, opened 2020-12-21
Update fields.md

stumblefiend:
PCX-1148


bensol1998 (approved):
Looks good. Should hold until after the change freeze just to be extra extra safe.

bensol1998 (approved):

PR #545, opened 2020-12-22
Doc: Access: Kubectl: add AKS-specific info

xens:
This commit adds more info on how to access Azure AKS clusters that are exposed through Cloudflare Access.

### DIFF
--- a/products/access/src/content/protocols-and-connections/kubectl.md
+++ b/products/access/src/content/protocols-and-connections/kubectl.md
@@ -86,3 +86,9 @@ With this service running, you can run a `kubectl` command and `cloudflared` wil
 ```sh
 $ alias kubeone="env HTTPS_PROXY=socks5://127.0.0.1:1234 kubectl"
 ```
+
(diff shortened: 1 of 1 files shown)


products/access/src/content/protocols-and-connections/kubectl.md
```diff
@@ -86,3 +86,9 @@ With this service running, you can run a `kubectl` command and `cloudflared` wil
 ```sh
 $ alias kubeone="env HTTPS_PROXY=socks5://127.0.0.1:1234 kubectl"
 ```
+
+If you want to use Cloudflare Access to protect an Azure AKS cluster you'll need to exclude the [login + refresh-token endpoints](https://docs.microsoft.com/en-us/azure/aks/concepts-identity#webhook-and-api-server) from the `HTTPS_PROXY` otherwise the login + refresh mechanisms will fail as they'll go to through the proxy.
```
abracchi-tw: If you want to use Cloudflare Access to protect an Azure AKS cluster, exclude the [login and refresh_token endpoints](https://docs.microsoft.com/en-us/azure/aks/concepts-identity#webhook-and-api-server) from the `HTTPS_PROXY` first. This will prevent the login and refresh mechanisms from going through the proxy and failing.

abracchi-tw (changes requested):

products/access/src/content/protocols-and-connections/kubectl.md
xens: agree better phrasing, corrected !

abracchi-tw (approved):

PR #569, opened 2020-12-31
workers: available request.cf properties to plans

### LINKED ISSUE #67: Update Request documentation to indicate which properties are restricted by plan level
Some of the properties in `RequestInitCfProperties` only work if the account owner has a certain minimum plan level, and some of the properties in `IncomingRequestCfProperties` are only present if the account owner has a certain minimum plan level. 

We previously documented this, but in the content migration to the new docs engine, we dropped it. We should label properties on that page as appropriate.

adaptive:
BIZ and ENT have these extra properties available:
`["longitude", "continent", "isEUCountry", "timezone", "postalCode", "regionCode", "region", "city", "latitude"]`

### DIFF
--- a/products/workers/src/content/runtime-apis/request.md
+++ b/products/workers/src/content/runtime-apis/request.md
@@ -197,13 +197,21 @@ All plans have access to:
 
   - The three-letter airport code of the data center that the request hit, e.g. `"DFW"`.
 
-- `tlsCipher` <Type>string</Type>
-
-  - The cipher for the connection to Cloudflare, e.g. `"AEAD-AES128-GCM-SHA256"`.
-
 - `country` <Type>string | null</Type>
 
   - Country of the incoming request. The two-letter country code in the request. This is the same value as that provided in the `CF-IPCountry` header, e.g. `"US"`.
+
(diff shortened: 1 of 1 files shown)


products/workers/src/content/runtime-apis/request.md
```diff
@@ -211,6 +213,12 @@ In addition to the properties on the standard [`Request`](https://developer.mozi
...
 
+</Definitions>
+
+Business and Enterprise scripts have access to:
+
+<Definitions>
+
 - `requestPriority` <Type>string | null</Type>
 
   - The browser-requested prioritization information in the request object, e.g. `"weight=192;exclusive=0;group=3;group-weight=127"`.
```
jclee: FWIW, I believe `requestPriority` and `httpProtocol` are available to all plans, so they should probably go in the earlier list.

adaptive:
@jclee corrected, migrated from the archive, but missed in my array. cc @signalnerve

jclee:
Nice!  Looks good to me, for what it's worth. :)  Thanks for fixing this!

jclee (approved):

Electroid (approved):

PR #602, opened 2021-01-05
Remove backslashes from DNS over Discord content

MattIPv4:
https://github.com/cloudflare/cloudflare-docs-engine/issues/309 -- backslashes don't behave as I had expected, removing them for now cause they look rather odd.


MattIPv4:
Superseded by #674

PR #666, opened 2021-01-20
[Tenant] add tenant docs to homepage

garrettgalow:
One thing I'm not sure about is if there is already a product icon in the system


garrettgalow:
found the brand assets and confirmed the tenant image already exists there

dcpena (approved):

PR #679, opened 2021-01-25
[workers] Add type/external URL to tutorials

kristianfreeman:
Adds the type column in tutorials. In addition, since we're experimenting with adding the new "Video" category (linking to our egghead course), this PR also includes support for setting a URL for a Markdown file, which will cause it to link directly to external URLs.

tuts


rita3ko (approved):

kristianfreeman:
Updated from category->type via @rita3ko's request

PR #699, opened 2021-01-28
Fixed header levels

hannesb0:
The hostname verification has 4 methods, so changing the header levels to make that clear

### DIFF
--- a/products/ssl/src/content/ssl-for-saas/hostname-verification.md
+++ b/products/ssl/src/content/ssl-for-saas/hostname-verification.md
@@ -4,7 +4,7 @@ order: 4
 
 # Hostname verification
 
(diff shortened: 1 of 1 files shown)


kodster28:
Howdy @hannesb0—forgive my ignorance—but you're saying that Apex proxying is not a sub-bullet under CNAME verification, but a completely different verification type?

If so, we'd also want to change the verbiage at the top of the article to say there are 4 verification methods.

hannesb0:
Hey @kodster28 you are right, apex verification is an optional and different verification method. It can be enabled by the account team for saas providers who want to allow their customers to proxy their apex domain.

I just updated the description to reflect this as well. Thanks for calling that out.

kodster28 (approved):

PR #733, opened 2021-02-04
Update to reflect currently supported frameworks

rita3ko:


haneefmubarak (approved):

haneefmubarak:
@rita3ko could you please resolve the merge conflict?

kristianfreeman:
@rita3ko I merged #772 last week, I think we can keep Jekyll in if we document that requirement. What do you think?

PR #738, opened 2021-02-04
Update firewall use cases pcx 1275

kodster28:
Addresses PCX-1275

@bensol1998, here's a screenshot of the output so you don't have to download and run yourself.

image

### DIFF
--- a/products/firewall/src/content/recipes/update-rules-customers-partners.md
+++ b/products/firewall/src/content/recipes/update-rules-customers-partners.md
@@ -1,6 +1,6 @@
 # Update firewall rules for customers or partners
 
-You may want to adjust your firewall rules to provide additional access to customers or partners.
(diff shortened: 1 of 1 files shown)


bensol1998:
Two things I think we should change before publishing:

1. Bot Management banner should say Enterprise plan _with Bot Management_
2. Can you remove the first "potential example" at the top of the page? Not sure internal systems should be protected this way

bensol1998 (changes requested):
Two things I think we should change before publishing:

Bot Management banner should say Enterprise plan with Bot Management
Can you remove the first "potential example" at the top of the page? Not sure internal systems should be protected this way

stumblefiend (approved):

bensol1998 (approved):

PR #768, opened 2021-02-11
[Logs] Updating Logpush to Splunk instructions and tutorial

crwaters16:


crwaters16:
Sending separate review to SE to validate the tutorial steps as well.

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -11,11 +11,11 @@ This tutorial explains how to analyze [Cloudflare Logs](https://www.cloudflare.c
 
 If you haven’t used Cloudflare Logs before, visit our [Logs documentation](/) for more details. Contact your Cloudflare Customer Account Team to enable logs for your account.
```
tlozoot: Seems like we can remove this given that the customer is already in the Cloudflare Logs documentation?

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -27,107 +27,111 @@ Before sending your Cloudflare log data to Splunk, make sure that you:
 Cloudflare logs are HTTP/HTTPS request logs in JSON format and are gathered from our 200+ data centers globally. By default, timestamps are returned as Unix nanosecond integers. We recommend using the RFC 3339 format for sending logs to Splunk.
```
tlozoot: We support many more types of logs beyond "HTTP/HTTPS request logs.

I would just remove this line; not sure what purpose it serves here. If there's a suggested timestamp format, we should specify that elsewhere...

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -27,107 +27,111 @@ Before sending your Cloudflare log data to Splunk, make sure that you:
...
 You can enable or disable acceleration after the initial configuration by accessing the app Set up page by clicking the **Apps** dropdown, then **Manage Apps** > **Cloudflare Set Up**.
 
 ![Splunk search macros](../../static/images/splunk/screenshots/splunk-apps-manage-apps-cloudflare-set-up-enable-data-model-acceleration.png)
 
-You can also manually configure Data Models by going to **Settings** > **Data models**. Learn more about data model acceleration [here](https://docs.splunk.com/Documentation/Splunk/latest/Knowledge/Acceleratedatamodels).
+You can also manually configure Data Models by going to **Settings** > **Data models**. Learn more about data model acceleration in the [Splunk documentation](https://docs.splunk.com/Documentation/Splunk/latest/Knowledge/Acceleratedatamodels).
+
+## Task 2 - Make the API call to create the Logpush job
+
+The Logpush API call should look like the following:
```
tlozoot: I wonder if we should just link out to the other page that we're adding, which contains more detail? This section seems a bit light/confusing confusing to me.

tlozoot: To clarify: for this whole "task", we can just link folks over to /logpush/splunk/index.md

tlozoot (commented):
Thanks Claire! Just a few minor comments.

products/logs/src/content/logpush/splunk/index.md
```diff
@@ -0,0 +1,145 @@
...
+  * `<SOURCE-TYPE>`: The Splunk sourcetype. Example: `cloudflare:json`
+  * `<SPLUNK-AUTH-TOKEN>`: The Splunk authorization token that’s URL-encoded. Example: `Splunk%20e6d94e8c-5792-4ad1-be3c-29bcaee0197d`   
+
+```bash
+"splunk://<SPLUNK-ENDPOINT-URL>?channel=<SPLUNK-CHANNEL-ID>&insecure-skip-verify=<INSECURE-SKIP-VERIFY>&sourcetype=<SOURCE-TYPE>&header_Authorization=<SPLUNK-AUTH-TOKEN>"
+```
+
+* `dataset` - The category of logs you want to receive, which is either `http_requests` (default), `spectrum_events`, or `firewall_events`.
+
+* `logpull_options` (optional) - To configure fields, sample rate, and timestamp format, see [Logpush API options](https://developers.cloudflare.com/logs/logpush/logpush-configuration-api/understanding-logpush-api#options).
```
tlozoot: A few other places in this doc suggest that it's best to use the option `timestamps=rfc3339 ` with Splunk. Maybe we should call that out explicitly here.

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -11,11 +11,11 @@ This tutorial explains how to analyze [Cloudflare Logs](https://www.cloudflare.c
 
 If you haven’t used Cloudflare Logs before, visit our [Logs documentation](/) for more details. Contact your Cloudflare Customer Account Team to enable logs for your account.
 
-This tutorial describes how to get Cloudflare logs from Amazon S3 into Splunk using the [Splunk Add-on for Amazon Web Services](https://splunkbase.splunk.com/app/1876/). To learn how to use Logpush to send logs to AWS S3, refer to the [Logpush documentation](/logpush/). Alternatively, you can use [Logpull](/logpull-api/) to get logs to your [Splunk HTTP Event Collector](http://dev.splunk.com/view/event-collector/SP-CAAAE6M) directly and skip Task 1.
+This tutorial describes how to send HTTP request lots directly to Splunk.
```
soheiokamoto: `logs`, I believe.
Also, same applies here that this is not only for HTTP request logs.

products/logs/src/content/analytics-integrations/splunk/index.md
soheiokamoto: +1

Also, I don't know if we want to remove `Connect AWS S3 to Splunk`. Wouldn't some customers still do that?

products/logs/src/content/analytics-integrations/splunk/index.md
crwaters16: This reference preexisted the tutorial update, which is why I left it in on the tutorial page. The Analytics integration tutorials are slightly separated from the logs docs in the TOC, but you are still in the doc set, so...I'm good either way.

products/logs/src/content/analytics-integrations/splunk/index.md
crwaters16: Good question, @soheiokamoto - if we expect some customers to still connect that way, I think I'd rather separate out the two pages (have one for Logpush to Splunk and one for Logpush to Splunk via AWS S3). They're pretty lengthy pages as-is already.

tlozoot: @soheiokamoto what do you think is the use case for S3 --> Splunk? I would like to strongly discourage folks from doing that.

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -27,107 +27,111 @@ Before sending your Cloudflare log data to Splunk, make sure that you:
...
+"Email="<Your Email Address Associated with your Global Key>"
+GlobalKey="<Your Global Key>"
+ZoneID="<Your ZoneID>"
+Hostname="<The Hostname Of Your Splunk Instance>"
+HEC_Token="<The Token presented after creating the HEC>"
+SelfSingedCert="true"
+Port=8088
+# Create a Channel ID. It just needs to be a unique string. It is recommended to use an UUID. The uuidgen program on Linux creates a UUID.
+Channel=$(uuidgen)
+curl -s -X POST -H "X-Auth-Email: $Email" -H "X-Auth-Key: $GlobalKey" -H "Content-Type: application/json" https://api.cloudflare.com/client/v4/zones/$ZoneID/logpush/jobs -d "{\"name\":\"Splunk\",\"destination_conf\":\"Splunk://${Hostname}:${Port}/services/collector/raw?channel=${Channel}&insecure-skip-verify=${SelfSignedCert}&sourcetype=_json&header_Authorization=Splunk%20${HEC_Token}\",\"logpull_options\":\"fields=RayID&timestamps=rfc3339\",\"dataset\": \"http_requests\"}"
```
bharatnc: This can be formatting better, why are we escaping  everywhere (i.e `\"name\"...`).

products/logs/src/content/analytics-integrations/splunk/index.md
bharatnc: > Good question, @soheiokamoto - if we expect some customers to still connect that way, I think I'd rather separate out the two pages (have one for Logpush to Splunk and one for Logpush to Splunk via AWS S3). They're pretty lengthy pages as-is already.

Those two should be separate. Also since we push to Splunk directly, can we just archive the old way of doing this integration (though S3) and then just recommend people to use Logpush to Splunk directly ?

soheiokamoto: @tlozoot I don't know. If you want to discourage it, then sure we can remove it.

products/logs/src/content/analytics-integrations/splunk/index.md
crwaters16: I copied this text over from an example and didn't parse it for things like that. Noted for future reviews since we'll be removing this section and referring customers to separate documentation for this task.

products/logs/src/content/analytics-integrations/splunk/index.md
tlozoot: Yeah... let's remove that part entirely. This isn't a case where having more options is a good thing. :)

crwaters16:
Updated the files with review feedback, lmk what you think about the new task 2!

abracchi-tw (approved):

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -7,127 +7,92 @@ order: 103
...
+## Task 2 - Make the API call to create the Logpush job
 
-The Cloudflare App for Splunk comes with a custom Cloudflare Data Model which has an acceleration time frame of 1 day but is not accelerated by default. If you enable [Data Model acceleration](https://docs.splunk.com/Documentation/Splunk/7.2.6/Knowledge/Acceleratedatamodels), we recommend that the Data Model is only accelerated for 1 or 7 days to ensure there are no adverse effects within your Splunk environment.
+Create the Logpush job by following the instructions on [Enable Logpush to Splunk](https://developers.cloudflare.com/logs/logpush/splunk).
 
-You can enable or disable acceleration after the initial configuration by accessing the app Set up page by clicking the **Apps** dropdown, then **Manage Apps** > **Cloudflare Set Up**.
+This API call creates a Logpush job but does not enable it. You will need to enable the Logpush job after setting the desired fields to push. Enable the job through the API by following the instructions on [Enable Logpush to Splunk](https://developers.cloudflare.com/logs/logpush/splunk) or through the Cloudflare dashboard by following these instructions:
```
tlozoot: Might be good to add a sentence to that effect of "You can't yet _create_ a Splunk Logpush job in the dashboard. However, after creating the job, you can visit the dashboard to _enable_ it or edit the fields used..."

tlozoot (approved):
Looks great! Left one minor comment.

In a future change, we should also take a look at the rest of the Splunk instructions. We probably want to add some more fields than just the ones recommended here.

tlozoot:
I'd like Keith to review this (based on his comments in wiki/jira) but then I think this is good to merge.

crwaters16:
> I'd like Keith to review this (based on his comments in wiki/jira) but then I think this is good to merge.

I sent him a PDF version of the steps to review separately and will wait to merge until I hear back from him!

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -7,127 +7,92 @@ order: 103
...
+3. Name the index **cloudflare**, which is the default index that the Cloudflare App will use. 
 
-![Splunk add-on search](../../static/images/splunk/screenshots/splunk-cloudflare-app-for-splunk-search-logs.png)
+After you've created an index, set up the HTTP Event Collector (HEC) on Splunk. To create an HEC:
 
-## Task 2 - Install and Configure the Cloudflare App for Splunk
+1. Click the **Settings** dropdown and select **Data inputs**. 
+2. Click **+Add new** and follow the wizard. When prompted, submit the following responses:
+    * Name: Cloudflare
+    * Source Type: Select > "_json"
```
soheiokamoto: `cloudflare:json`, I believe.

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -7,127 +7,92 @@ order: 103
...
+At the end of the wizard you will see a **Token Value**. This token authorizes the Cloudflare Logpush job to send data to your Splunk instance. If you forget to copy it now, Splunk allows you to get the value at any time. 
 
-1. Login to your Splunk instance.
-2. Under **Apps** > **Find More Apps** search for _Cloudflare App for Splunk._
-3. Click **Install**.
+The final information you need is the endpoint to use to send the data to. The endpoint should be:
 
-![Splunk find Cloudflare app](../../static/images/splunk/screenshots/splunk-cloudflare-app-for-splunk.png)
+```bash
+"<protocol>://input-<host>:<port>/<endpoint>" or "<protocol>://http-inputs-<host>:<port>/<endpoint>"
```
soheiokamoto, line 54: There seems to be confusion on the prefix (in some case there is no prefix), though I have not been able to find clear and up-to-date Splunk docs.

products/logs/src/content/logpush/splunk/index.md
```diff
@@ -0,0 +1,145 @@
...
+* `name` (optional) - Use your domain name as the job name.
+* `destination_conf` - A log destination consisting of an endpoint URL, channel id, insecure-skip-verify flag, sourcetype, authorization header in the string format below. 
+
+  * `<SPLUNK-ENDPOINT-URL>`: The Splunk raw HTTP Event Collector URL with port. Example: `splunk.cf-analytics.com:8088/services/collector/raw`
+      * Cloudflare expects the HEC network port to be configured to :443 or :8088. 
+      * Cloudflare expects the Splunk endpoint to be /services/collector/raw while configuring and setting up the Logpush job.  
+  * `<SPLUNK-CHANNEL-ID>`: A unique channel ID. This is a random GUID that you can generate by:
+      * Using an online tool like the [GUID generator](https://www.guidgenerator.com/) 
+      * Using command line.  Example: `python -c 'import uuid; print(uuid.uuid4())'` 
+  * `<INSECURE-SKIP-VERIFY>`: Boolean value. If set to true, makes insecure connection to Splunk. Setting this value to true is equivalent to using the `-k` option with curl as shown in Splunk examples and is NOT recommended.
```
soheiokamoto: There is in FAQ belowm but I think it would be good if we can tell when a customer needs this, or how they can check to know around here (i.e., I believe this needs to be turned on when HEC uses self-signed certificate).

products/logs/src/content/logpush/splunk/index.md
```diff
@@ -0,0 +1,145 @@
...
+</Aside>
+
+## 1. Create a job
+
+To create a job, make a `POST` request to the Logpush jobs endpoint with the following fields:
+
+* `name` (optional) - Use your domain name as the job name.
+* `destination_conf` - A log destination consisting of an endpoint URL, channel id, insecure-skip-verify flag, sourcetype, authorization header in the string format below. 
+
+  * `<SPLUNK-ENDPOINT-URL>`: The Splunk raw HTTP Event Collector URL with port. Example: `splunk.cf-analytics.com:8088/services/collector/raw`
```
soheiokamoto: It would be probably good to mention to have HTTP Event Collector (HEC) on Splunk as a prerequisite, and point to the section how to setup in `analytics-integrations/splunk/index.md` as an example with Cloudflare app.

products/logs/src/content/analytics-integrations/splunk/index.md
```diff
@@ -7,127 +7,92 @@ order: 103
...
+    * App Context: Cloudflare App for Splunk (cloudflare)
+    * Index: cloudflare
 
-To install the [Cloudflare App for Splunk](https://splunkbase.splunk.com/app/4501/):
+At the end of the wizard you will see a **Token Value**. This token authorizes the Cloudflare Logpush job to send data to your Splunk instance. If you forget to copy it now, Splunk allows you to get the value at any time. 
 
-1. Login to your Splunk instance.
-2. Under **Apps** > **Find More Apps** search for _Cloudflare App for Splunk._
-3. Click **Install**.
+The final information you need is the endpoint to use to send the data to. The endpoint should be:
```
soheiokamoto: I think it would be helpful to mention to check whether it is using self-signed certificate or not around here, as it relates to one of the parameter in Logpush job for Splunk.

products/logs/src/content/logpush/splunk/index.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+### What's the insecure-skip-verify parameter?
+This flag, if set to true, makes an insecure connection to Splunk. Setting this value to true is equivalent to using the `-k` option with curl as shown in Splunk examples and is NOT recommended. Cloudflare highly recommends setting this flag to false when using the `insecure-skip-verify` parameter.
+
+### Why do we have the insecure-skip-verify parameter if it's not recommended?
+Certificates generated by Splunk/third party certificates should have the Common Name field in the certificate match the Splunk server’s domain name. Otherwise you may see errors like: `x509: certificate is valid for SplunkServerDefaultCert, not <your-instance>.splunkcloud.com.` 
+
+During our testing, we discovered that it was common to come across a Splunk instance with this misconfiguration, and pushes will never succeed unless the certificates are fixed. This happens especially with those default certificates generated by Splunk on startup.
```
soheiokamoto: nit: Feels like too much detail here on the first sentence. Maybe we can just say `In such case, pushes will never succeed unless the certificate is fixed`.

products/logs/src/content/analytics-integrations/splunk/index.md
crwaters16: Keith recommended including his example here, so for now let's address this confusion by including that example here and linking out to Splunk's documentation.

products/logs/src/content/analytics-integrations/splunk/index.md
crwaters16: We try to avoid framing things from a negative standpoint (you can't do XYZ) and try to avoid hinting at future functionality because even if it'll be enabled next week, there might be a current-state user who's trying to follow these instructions. Slightly tweaking this section to clarify the process and highlight the dash more prominently.

products/logs/src/content/analytics-integrations/splunk/index.md
crwaters16: Sohei, I addressed this by adding a new task for verifying whether it's using a self-signed certificate. I don't have steps for how to do that, though, if you happen to have that info!

products/logs/src/content/analytics-integrations/splunk/index.md
soheiokamoto: Sure. I would suggest something like below, but it's fine either way:
```
"<protocol>://<input-prefix><host>:<port>/<endpoint>"
```
And:
* `input-prefix`: input prefix based on your instance, such as `input-` or `http-inputs-`. 

@bharatnc may have some feedback on this.

bharatnc: Sounds good, one thing to note is that when we push logs directly to Splunk we only use `https` and don't support `http`.  So the `protocol` looks confusing to me.

products/logs/src/content/logpush/splunk/index.md
bharatnc: May be even can move to a separate section called Troubleshooting.

products/logs/src/content/analytics-integrations/splunk/index.md
soheiokamoto: Looks good.
I think it is based on one of the config, like `Enable SSL` checkbox, similar to this I think: https://docs.splunk.com/Documentation/Splunk/latest/Data/UsetheHTTPEventCollector

CC: @bharatnc

products/logs/src/content/logpush/splunk/index.md
soheiokamoto: The way it is in the latest looks good to me.

products/logs/src/content/analytics-integrations/splunk/index.md
bharatnc: Ok.

bharatnc (approved):
LGTM.

soheiokamoto (approved):

tlozoot:
Woohoo! @crwaters16 ready to merge?

crwaters16:
> Woohoo! @crwaters16 ready to merge?

Three days later thanks to TX power outages, but good to go!!

PR #780, opened 2021-02-12
Add bharatnc to CODEOWNERS for Logs

soheiokamoto:
This adds `bharatnc` to `CODEOWNERS` for `Logs`.


soheiokamoto:
CC: @bharatnc

PR #798, opened 2021-02-16
[Durable Objects] fix formatting bug

haneefmubarak:


a-robinson (commented):
I suppose I'm happy enough with this if it fixes the formatting, although it's weird that it's needed. If the extra blank line is needed in general, should we also add it to the other list entries here or will that break them?

a-robinson (approved):

haneefmubarak:
Well so IMO it's a parser / spec annoyance, because it seems to only be an issue if the preceding line has `<code>` and/or `<type>` in it (I haven't played around to actually figure out which or whether it's both). I'm not entirely sure whether adding it to the other lines will break it, but my strategy here is not to fix what isn't broken.

The file isn't exactly a pleasant raw Markdown file (given all the ampersand escaped characters and the HTML tags littered all over), so I'm not too concerned with the raw input as opposed to the output on this one.

haneefmubarak:
(apologies for the delayed reply, hadn't seen the GH notification for this until today)

PR #811, opened 2021-02-19
Put `state` in class variable in DO examples, not `state.storage`

a-robinson:
Because there are other useful fields within `state` as well, including
`id` and the `waitUntil` function. And if you do the naive thing:

constructor(state, env) {
  this.storage = state.storage;
  this.waitUntil = state.waitUntil;
}

...then your `this.waitUntil()` calls will fail with an "Illegal
invocation" error because the wrong object will be passed as `this` into
`waitUntil()`.

By just storing `state` as a member variable in our examples instead,
it's less likely that folks will do the wrong thing with waitUntil.


haneefmubarak:
Ah that makes sense. LGTM.

haneefmubarak (approved):

PR #821, opened 2021-02-22
Update scheduled-changes.md

umahasit:
Scheduled change for next monday waf

### DIFF
--- a/products/waf/src/content/change-log/scheduled-changes.md
+++ b/products/waf/src/content/change-log/scheduled-changes.md
@@ -31,6 +31,30 @@ order: 1
             </th>
         </tr>
     </thead>
+        <tbody>
+        <tr>
+            <td>
+                2021-02-22
+            </td>
+            <td>
+                2021-03-01
+            </td>
+            <td>
+                Cloudflare Specials
+            </td>
+            <td>
+                100005
+            </td>
+            <td>
+                Improve LFI Log Injection/Disclourse coverage.
+            </td>
+            <td>
+                simulate
+            </td>
+            <td>
+                simulate
+            </td>
+        </tr>
     <tbody>
     </tbody>
 </table></TableWrap>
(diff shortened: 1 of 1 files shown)


products/waf/src/content/change-log/scheduled-changes.md
```diff
@@ -31,6 +31,30 @@ order: 1
...
+                Cloudflare Specials
+            </td>
+            <td>
+                100005
+            </td>
+            <td>
+                Improve LFI Log Injection/Disclourse coverage.
+            </td>
+            <td>
+                simulate
```
helloandre: is this a new rule? in which case this would be "n/a"
```suggestion
                n/a
```

umahasit: 100005 is a currently existing rule, this is an update to that rule(rule modification)

products/waf/src/content/change-log/scheduled-changes.md
```diff
@@ -31,6 +31,30 @@ order: 1
...
+            <td>
+                Cloudflare Specials
+            </td>
+            <td>
+                100005
+            </td>
+            <td>
+                Improve LFI Log Injection/Disclourse coverage.
+            </td>
+            <td>
```
migueldemoura, line 51: `100005`'s current default is `drop` and I presume we're keeping it as `drop`? If so, both of these fields should be `drop`.

migueldemoura, line 51: `100005`'s current default is `block` and I presume we're keeping it as `block`? If so, both of these fields should be `Block`.

umahasit: Good spot, thanks will update!

umahasit: updated

migueldemoura (approved):

PR #836, opened 2021-02-23
Fix s-maxage expiry header

vladkosarev:
According to https://support.cloudflare.com/hc/en-us/articles/115003206852-Understanding-Origin-Cache-Control the correct directive is s-maxage, not s-max-age. I thought maybe it was an alias but s-max-age does not work at all.


haneefmubarak (approved):

haneefmubarak:
Thanks for the PR!

PR #854, opened 2021-02-26
[Teams] cert pinning + regex docs

abracchi-tw:
we're not going to add the cert pinning hostnames table for now.
Just adding regex docs


ptzimmerman (changes requested):
Let's wait on this for now since this list of hostnames is going to change

We should also keep a few example rules of how to bypass

abracchi-tw:
> Let's wait on this for now since this list of hostnames is going to change

@ptzimmerman could we merge this and then update the list when it changes?

abracchi-tw:
@ptzimmerman added the table with bypass rules back

TownLake (approved):

ptzimmerman (changes requested):
@abracchi-tw we made updates to the hostname list:
https://bitbucket.cfdata.org/projects/DOPE/repos/applications/browse/definitions/cert-pinning.yaml

ptzimmerman (approved):

PR #856, opened 2021-02-28
[pages] Fixed section id in pricing URL

arunesh90:
This fixes the section id that's in the pricing URL, so clicking on it now will show the pricing section immediately.


haneefmubarak (approved):

haneefmubarak:
Thanks for the fix!

PR #866, opened 2021-03-02
Update run-as-service.md

phatmandrake:
`sc start cloudflared tunnel run` and `Start-Service cloudflared tunnel run` are not equivalent.

https://stackoverflow.com/questions/37337718/powershell-start-windows-service-with-a-parameter

Although it may also be pertinent to note that neither ensures that the parameters are passed each time.


TownLake (approved):
Thanks for the edit!

PR #874, opened 2021-03-03
some people were confused by destructuring, be more explicit

koeninger:
@greg-mckeon

### DIFF
--- a/products/workers/src/content/runtime-apis/kv.md
+++ b/products/workers/src/content/runtime-apis/kv.md
@@ -143,10 +143,11 @@ For large values, the choice of `type` can have a noticeable effect on latency a
 
 You can get the metadata associated with a key-value pair alongside its value by calling the `getWithMetadata` method on a namespace you’ve bound in your script:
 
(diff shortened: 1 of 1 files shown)


products/workers/src/content/runtime-apis/kv.md
```diff
@@ -143,7 +143,10 @@ For large values, the choice of `type` can have a noticeable effect on latency a
 
 You can get the metadata associated with a key-value pair alongside its value by calling the `getWithMetadata` method on a namespace you’ve bound in your script:
 
-`const {value, metadata} = await NAMESPACE.getWithMetadata(key)`
+`const valueAndMetadata = await NAMESPACE.getWithMetadata(key)
```
a-robinson: I'm not sure if you checked out how this renders, but I'd expect for a multi-line code snippet you'd want to format it like:

\`\`\`js
const valueAndMetadata = await NAMESPACE.getWithMetadata(key)
const value = valueAndMetadata.value
const metadata = valueAndMetadata.metadata
\`\`\`

a-robinson (approved):

PR #896, opened 2021-03-06
Document HUGO_VERSION env var

barnybug:


rita3ko (approved):
looks good! thanks so much!

PR #910, opened 2021-03-08
Fix some promise-sharing bugs in the Durable Object counter example

bretthoerner:
Developers need to `waitUntil` on shared promises, otherwise the promise
may never settle (if, for example, the client disconnects) and other
requests will `await` indefinitely -- effectively breaking the object
until its isolate is recreated.

Additionally, developers should always catch exceptions in
initialization promises or else they could be left with a shared promise
that will throw forever ... when they probably want to retry
initialization instead.

### DIFF
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -399,8 +399,16 @@ export class Counter {
     }
 
     async initialize() {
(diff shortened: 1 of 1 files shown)


a-robinson (approved):
Thanks for jumping on this quickly!

eidam:
Thank you for this! :)

PR #916, opened 2021-03-09
Update log values

migueldemoura:
Firewall Actions:
`simulate` -> `log`
`drop` -> `block`
Old `log` is removed.

Firewall Source:
`sanityCheck` -> `validation`

IP Class:
`whitelist` -> `allowlist`

Deprecates `WAFFlags` and `WAFMatchedVar`.


pedrosousa (approved):

umahasit (approved):

soheiokamoto:
Any changes under `products/logs/src/content/log-fields/_partials/` must be done on the internal `logshare-api` repo, otherwise any change here will be lost (as the contents of these files are based on the templates there).

I think you have updated some there, but make sure all are up-to-date.

PR #917, opened 2021-03-09
Update connection-details.md

tenaciousdlg:
Updated SNI description to account for SSL Modes with Custom Origins due to the SNI header not matching the Custom Hostname.


maxnystrom:
This change is not accurate and should be declined.

kodster28:
Basing this of Max's comment.

PR #918, opened 2021-03-10
Add Hugo environment variable

samrobbins85:
The Hugo version can be set using an environment table, so updating the table to reflect this


KimJ15:
Confirmed that this update has been made to the docs through another PR. Closing as complete.

PR #973, opened 2021-03-20
Some changes to "Installing Hugo"

fu-sen:
- Align "Installing Hugo" with Zola
- Unification of headings
- Add "Windows (Scoop)"


KimJ15:
Confirmed that many of these updates were made through different PRs and the doc is updated. Closing PR.

PR #977, opened 2021-03-22
Create 2020-03-22.md

umahasit:


products/waf/src/content/change-log/2020-03-22.md
```diff
@@ -0,0 +1,47 @@
...
+                100185
+            </td>
+            <td>
+                Deprecated Short PHP Open Tag (&lt;?) Mitigation
+            </td>
+            <td>
+                N/A
+            </td>
+            <td>
+                Simulate
```
migueldemoura: Should this be "Block"?

umahasit: As its a new rule, as per the SOP, this should be simulate, then we can push out a scheduled change for next monday for default block.

migueldemoura: Judging by the file name, this would be a change log entry, not a scheduled change.

I think you meant to add this rule here: https://github.com/cloudflare/cloudflare-docs/blob/production/products/waf/src/content/change-log/scheduled-changes.md.
The `New Action` field there would also contain the final action after the 1 week "staging" period, so it'd be `Block`.

migueldemoura: Talked offline: we'll use `Disable` instead so we can make a new change log entry straight away.

migueldemoura (approved):

PR #1000, opened 2021-03-24
include zola/gutenberg env vars in Pages > Build Configuration

1000hz:
This PR updates the Pages > Build Configuration page to include the environment variables needed for Gutenberg and Zola. These tools are not installed by default and are only available if their corresponding environment variables are specified in your project's build settings.


rita3ko (changes requested):
@1000hz would you mind adding the latest version number instead of `(requires environment variable` for consistency with the others?

1000hz:
> @1000hz would you mind adding the latest version number instead of `(requires environment variable` for consistency with the others?

The reason for that bit of text is that the build image doesn't install these two tools _unless_ you have those environment variables set. There are no "default" versions included. Putting the latest version of the tools in that column would be misleading, as a user might see it and think they are safe to omit the env variable because they don't need an older version.

rita3ko:
ah i see what you're saying. we started out having that column be named "latest version" but it felt a bit silly since that's what the platform should just use (pending pages-11). 
but that makes sense, thanks for that clarification

rita3ko (approved):

YoutacRSVARs (approved):

PR #1005, opened 2021-03-25
Update Schema Validation rule to use regular expression

pedrosousa:


products/firewall/src/content/cf-dashboard/configure-schema-validation.md
```diff
@@ -42,11 +42,13 @@ To configure Schema Validation in the Cloudflare dashboard, follow these steps:
 
 1. Configure the expression for the API Shield using the available request fields.
 
-    For example, if your API is available at `http://api.example.com/v1`, the expression must include a check for the _hostname_ field (which must be equal to `api.example.com`) and a check for the _URI path_ field (which must be equal to `/v1`).
+    For example, if your API is available at `http://api.example.com/v1`, the expression must include a check for the _Hostname_ field (which must be equal to `api.example.com`) and a check for the _URI Path_ field using a regular expression (which must match regex `^/v1.*`).
```
migueldemoura:
```suggestion
    For example, if your API is available at `http://api.example.com/v1`, the expression must include a check for the _Hostname_ field (which must be equal to `api.example.com`) and a check for the _URI Path_ field using a regular expression (which must match the regex `^/v1`).
```

pedrosousa: Hi @migueldemoura! Thank you for updating the regex. Regarding the "the", I followed the _matches regex_ label in the UI for the **Operator** field, but in fact it doesn't have to be exactly the same.

pedrosousa: I'll also update the regex in the screenshot.

migueldemoura: Sounds good!

migueldemoura (approved):

kodster28 (approved):

PR #1010, opened 2021-03-25
Update installation.md to fix copy/paste of commands

tenaciousdlg:
Fixed a copy/paste issue related to the syntax descriptor of code blocks. If you use `sh` you need to have an account indicator ahead of your code (i.e. `$ code` vs `code`). If you use a descriptor of `bash` this is not an issue and you can just use `code` without the user account type descriptor.


TownLake (approved):

TownLake:
Thanks for the catch and fix! We're updating this globally based on your report here. Apologies for the inconvenience.

PR #1011, opened 2021-03-25
replace-sh-with-bash

TownLake:


products/1.1.1.1/src/content/setting-up-1.1.1.1/linux.md
```diff
@@ -38,7 +38,7 @@ Step 8: Click Apply.
 
 Step 1: In the command line, type:
 
-```sh
+```bash
 $ sudo vim /etc/resolv.conf
```
pedrosousa: @TownLake We can keep these as `sh` because the line inside the code block starts with a `$` prompt.
Everything after the `$` can be copied without issues, and the CSS styling (on hover) makes it clear that the `$` is not part of the command itself.

TownLake: good point; fixed if `$` present in next line

KimJ15:
@TownLake if i'm remembering correctly, I think a fix for this was found. Is this a PR we can close?

TownLake:
@KimJ15 yes we can close this out

PR #934, opened 2021-03-15
[Logs] Fix a typo in Datadog's ingestion webhook attributes description

bakhtin:
[As per Datadog docs](https://docs.datadoghq.com/api/latest/logs/#send-logs):
- `host` should be `hostname`
- `dtags` should be `ddtags`

PR #1052, opened 2021-04-03
YoutacRS

ytbigvinny27:


tlozoot (approved):
Thanks for the fix. @soheiokamoto can you confirm that this is correct before we merge?

crwaters16:
> Thanks for the fix. @soheiokamoto can you confirm that this is correct before we merge?

These were both addressed in https://github.com/cloudflare/cloudflare-docs/pull/934, FYI.

crwaters16:
Closing PR because the issues were addressed in #934. Let us know if you have any questions!

PR #1125, opened 2021-04-13
Update WebCrypto conformance table

vlovich:
Use `-` to indicate function is invalid for that algorithm.
Use `✘` to indicate when the function isn't implemented at all.
For partial implementation, document the caveats.
Add missing `importKey` column.

Document that the following additions have been made to the runtime:
* AES-CTR
* AES import/export for JWK format
* HMAC JWK import
* ECDH (generateKey, import, export, & deriveBits/deriveKey). JWK import not implemented.
* ECDSA generateKey added. Fully supported except for missing JWK import.

### DIFF
--- a/products/workers/src/content/runtime-apis/web-crypto.md
+++ b/products/workers/src/content/runtime-apis/web-crypto.md
@@ -271,30 +271,46 @@ These methods are all accessed via `crypto.subtle`, which is also [documented in
 
 ### Supported algorithms
 
-Workers implements a subset of the most common cryptographic algorithms, as shown in the following table. We are happy to add support for more algorithms — [let us know about your use case](https://community.cloudflare.com/c/developers/workers).
+Workers implements all operation of the [WebCrypto standard](https://www.w3.org/TR/WebCryptoAPI), as shown in the following table.
+We are happy to add support for more algorithms — [let us know about your use case](https://community.cloudflare.com/c/developers/workers).
+
+A checkmark (✓) indicates that this feature is believed to be fully supported according to the spec.
+[//]: #  An x (✘) indicates that this feature is part of the specification but not implemented.
+[//]: #  If a feature only implements the operation partially, details are listed.
 
 <TableWrap>
 
-| Algorithm                                 | sign()<br/>verify() | encrypt()<br/>decrypt() | digest() | deriveBits()<br/>deriveKey() | generateKey() | wrapKey()<br/>unwrapKey() | exportKey() |
-| :---------------------------------------- | :------------------ | :---------------------- | :------- | :--------------------------- | :------------ | :------------------------ | ----------- |
-| RSASSA-PKCS1-v1_5                         | ✓                   |                         |          |                              | ✓             |                           | ✓           |
-| RSA-PSS                                   | ✓                   |                         |          |                              | ✓             |                           | ✓           |
-| ECDSA                                     | ✓                   |                         |          |                              |               |                           | ✓           |
-| HMAC                                      | ✓                   |                         |          |                              | ✓             |                           |             |
-| AES-CBC                                   |                     | ✓                       |          |                              |               | ✓                         |             |
-| AES-GCM                                   |                     | ✓                       |          |                              | ✓             | ✓                         |             |
-| SHA-1                                     |                     |                         | ✓        |                              |               |                           |             |
-| SHA-256                                   |                     |                         | ✓        |                              |               |                           |             |
-| SHA-384                                   |                     |                         | ✓        |                              |               |                           |             |
-| SHA-512                                   |                     |                         | ✓        |                              |               |                           |             |
-| MD5<sup><a href="#footnote-1">1</a></sup> |                     |                         | ✓        |                              |               |                           |             |
-| PBKDF2                                    |                     |                         |          | ✓                            |               |                           |             |
+| Algorithm                                         | sign()<br/>verify()  | encrypt()<br/>decrypt()   | digest() | deriveBits()<br/>deriveKey() | generateKey() | wrapKey()<br/>unwrapKey() | exportKey() | importKey() |
+| :------------------------------------------------ | :------------------- | :------------------------ | :------- | :--------------------------- | :------------ | :------------------------ | :---------- | :---------- |
+| RSASSA PKCS1 v1.5                                 | ✓                    |                           |          |                              | ✓             |                           | ✓           | ✓           |
+| RSA PSS                                           | ✓                    |                           |          |                              | ✓             |                           | ✓           | ✓           |
+| RSA OAEP                                          |                      | ✓                         |          |                              | ✓             | ✓                         | ✓           | ✓           |
+| ECDSA                                             | ✓                    |                           |          |                              | ✓             |                           | ✓           | ✓           |
+| ECDH                                              |                      |                           |          | ✓                            | ✓             |                           | ✓           | ✓           |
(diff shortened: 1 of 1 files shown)


vlovich:
cc @harrishancock @mkevanz

products/workers/src/content/runtime-apis/web-crypto.md
```diff
@@ -271,31 +271,37 @@ Workers implements a subset of the most common cryptographic algorithms, as show
...
+| SHA-1                                     | -                                        | -                         | ✓        | -                            | -             | -                         | -           | -           |
+| SHA-256                                   | -                                        | -                         | ✓        | -                            | -             | -                         | -           | -           |
+| SHA-384                                   | -                                        | -                         | ✓        | -                            | -             | -                         | -           | -           |
+| SHA-512                                   | -                                        | -                         | ✓        | -                            | -             | -                         | -           | -           |
+| MD5<sup><a href="#footnote-2">1</a></sup> | -                                        | -                         | ✓        | -                            | -             | -                         | -           | -           |
```
harrishancock: The new footnote anchor is correct, but the `1` text needs to be `2` now.

products/workers/src/content/runtime-apis/web-crypto.md
```diff
@@ -271,31 +271,37 @@ Workers implements a subset of the most common cryptographic algorithms, as show
...
+| Algorithm                                 | sign()<br/>verify()                      | encrypt()<br/>decrypt()   | digest() | deriveBits()<br/>deriveKey() | generateKey() | wrapKey()<br/>unwrapKey() | exportKey() | importKey() |
+| :---------------------------------------- | :--------------------------------------- | :------------------------ | :------- | :--------------------------- | :------------ | :------------------------ | :---------- | :---------- |
+| RSASSA-PKCS1-v1_5                         | ✓                                        | -                         | -        | -                            | ✓             | -                         | not jwk     | ✓           |
+| RSA-PSS                                   | ✓                                        | -                         | -        | -                            | ✓             | -                         | not jwk     | ✓           |
+| RSA-OAEP                                  | ✘ <sup><a href="#footnote-1">1</a></sup> | -                         | -        | -                            | ✘             | -                         | ✘           | ✓           |
```
harrishancock: I think we don't support RSA-OAEP importKey() yet.

products/workers/src/content/runtime-apis/web-crypto.md
```diff
@@ -271,31 +271,37 @@ Workers implements a subset of the most common cryptographic algorithms, as show
...
+| HMAC                                      | ✓                                        | -                         | -        | -                            | ✓             | -                         | raw only    | ✓           |
+| AES-CBC                                   | -                                        | ✓                         | -        | -                            | ✓             | ✘                         | raw only    | ✓           |
+| AES-GCM                                   | -                                        | ✓                         | -        | -                            | ✓             | ✘                         | raw only    | ✓           |
+| AES-CTR                                   | -                                        | ✘                         | -        | -                            | ✓             | ✘                         | raw only    | ✓           |
+| AES-KW                                    | -                                        | -                         | -        | -                            | ✘             | ✘                         | raw only    | ✓           |
```
harrishancock: I think we _do_ support AES-KW generateKey() (not that it's useful yet).

products/workers/src/content/runtime-apis/web-crypto.md
vlovich: Perhaps you're aware of this already (or maybe our docs use a slightly different renderer), but Markdown renderers typically ignore any numbers you actually use & numbers the list contiguously for you. Generally I prefer to keep markdown lists to always have `1.` prefix so that it's clear that that's what I'm relying on & it makes it easier to rearrange the list without needing to worry about also renumbering. I do see both styles both within the docs more broadly although workers does seem to use 1/2/3. However, it's also possible I've been misunderstanding the difference in usage & the existing code uses the mix of them intentionally for specific reasons so I'd love to learn more.

For example:
[1, 2, 3 numbering](https://raw.githubusercontent.com/cloudflare/cloudflare-docs/production/products/workers/src/content/cli-wrangler/install-update.md)
[1, 1, 1 numbering](https://raw.githubusercontent.com/cloudflare/cloudflare-docs/production/products/magic-transit/src/content/set-up/onboarding.md)

harrishancock: To be clear, Markdown will render `<a href="#footnote-2">1</a>` as a hyperlinked numeral `2`? Note that GitHub's preview renders it as a `1`, which is why this stuck out to me.

vlovich: Ah shoot. I misread the comment as applying down below where the footer actually is. My bad. Will be fixed in next push.

products/workers/src/content/runtime-apis/web-crypto.md
vlovich: It is now, right? :).

products/workers/src/content/runtime-apis/web-crypto.md
vlovich: You're right.

YoutacRSVARs (approved):

products/workers/src/content/runtime-apis/web-crypto.md
```diff
@@ -267,35 +267,49 @@ These methods are all accessed via `crypto.subtle`, which is also [documented in
...
+| PBKDF2                                            |                      |                           |          | ✓                            |               |                           |             | ✓           |
 
 </TableWrap>
 
 __Footnotes:__
 
 1. <a name="footnote-1"></a> MD5 is not part of the WebCrypto standard, but is supported in Cloudflare Workers for interacting with legacy systems that require MD5. MD5 is considered a weak algorithm. Do not rely upon MD5 for security.
+2. <a name="footnote-2"></a> Non-standard EdDSA is supported for the Ed25519 curve using <Code>NODE-ED25519</Code> as the algorithm and namedCurve parameters. X25519 support is not implemented. Unlike NodeJS, import in "raw" format of private
+keys will not be supported unless the WebCrypto spec relaxes ECDSA/ECDH restrictions similarly. Since this algorithm is non-standard, we cannot guarantee stabilty of the implementation although we will strive to provide some kind of
+compatability path.
```
vlovich: @rita3ko could use some help word-smithing this.

YoutacRSVARs (approved):

products/workers/src/content/runtime-apis/web-crypto.md
rita3ko: I would break it up into bullet points to make it easier to parse. Something like: 

```
Non-standard EdDSA is supported for the Ed25519 curve. Since this algorithm is non-standard, and based on the Node implementation, a few things to keep in mind while using it: 
- Use <Code>NODE-ED25519</Code> as the algorithm and namedCurve parameters. 
- Unlike NodeJS, Workers does not support "raw" import of private
keys. 
- Since this algorithm is not available as a part of the WebCrypto standard API today, we cannot guarantee the stability of the implementation. If there are changes here, we will strive to make them backward compatible, whilst maintaining an option for compatibility with NodeJS's implementation.
```

vlovich: That works for me. I adjusted the wording slightly but hopefully ready for approval now (I want to squash the fixup commit before merging once we're settled on the wording)

rita3ko (approved):

PR #1127, opened 2021-04-13
Fix root wildcard example

nihaals:
I'm unsure if this is a bug or is intended, but the documented behaviour isn't accurate. I noticed this when trying to have an application for `/` and an application for `/*`, and the `/*` policy still took precedence. After removing the `/` application I noticed the application name used for `/*` was still being shown on the login page (for `/`) and the policy was still being used.

## Steps to reproduce

1. Create an orange cloud record for `test.example.com`
2. Create an application for `test.example.com/*`
    - Name it `Test Application`
    - Disable instant auth
3. Go to `test.example.com/` and notice you are taken to the identity provider selection page and `Test Application` is shown


kennyj42 (approved):
Good catch, thank you. This is the intended behavior for wildcards in a path. We'll merge this change in.

PR #1154, opened 2021-04-16
[Stream] Fix typo in FAQ

gabriel-kaam:
"We use to few different" -> "We use a few different"


dcpena (commented):
Can we add "a" to make the sentence "We use a few different adaptive streaming..."?

dcpena (approved):

Issue #841, opened 2021-02-24
Missing docs on how to add custom headers to Pages


lawzava:
Either the docs are missing or the feature itself but in that case, I think it should be explicitly stated that adding custom HTTP headers is not supported.

theamanbhargava:
Is there any update on this?

ghost:
Any Updates on this?

kristianfreeman:
We're working on a guide to show how you can build custom header support using Workers for your Pages project - more on that hopefully soon!

ghost:
> We're working on a guide to show how you can build custom header support using Workers for your Pages project - more on that hopefully soon!

When will it be done without Workers?

RafidMuhymin:
This feature should be available without workers. Because it's something that everyone must have to use.

dustensalinas:
Documentation is incorrect for the referenced github page.

deadlypants1973:
> Documentation is incorrect for the referenced github page.

@dustensalinas Can you please drop the incorrect section in the docs you are referencing so we can review?

PR #1157, opened 2021-04-17
Clarify which netlify features are supported

johtso:
So it's not currently mentioned that `_headers` custom header rules aren't supported (this _is_ correct, right?).

Thought it might make sense to mention that, and communicate the level of feature parity more clearly.

Feel free to throw this away and make a different edit if there's a better way for this to be communicated!


deadlypants1973:
Closing this with a pending update to https://github.com/cloudflare/cloudflare-docs/issues/841

PR #1162, opened 2021-04-18
[Workers] fix type in `event.respondWith()` parameter in `runtime-apis/fetch-event.md`

myakura:
it should be `Response` or `Promise<Response>`


Electroid (approved):
Thanks!!

PR #1221, opened 2021-04-26
Fix signed url script snippet

third774:


renandincer (approved):

products/stream/src/content/viewing-videos/securing-your-stream.md
```diff
@@ -69,8 +69,8 @@ After creating a key, you can use it to sign unique signed tokens. These tokens
 
 You can sign to assert these optional constraints on the token:
 
-- `exp` - expiration; a unix epoch timestamp **after** which the token will not be accepted.
-- `nbf` - notBefore; a unix epoch timestamp **before** which the token will not be accepted.
+- `exp` - expiration; a unix epoch timestamp **after** which the token will not be accepted. Defaults to one hour after issuing.
```
Everlag: The default is part of the signing utility, not part of the verification logic AFAIK.

As-is, we SHOULD be allowing infinitely long tokens as that is a customer decision.

Everlag (changes requested):

products/stream/src/content/viewing-videos/securing-your-stream.md
third774: Ah, thanks! Updated to move the info to that section.

Everlag (approved):

PR #1232, opened 2021-04-28
Added content type flags to bots pages

kodster28:


crwaters16 (approved):
I'm the capitalization enforcer today, apparently ;) guessing we need to be consistent in cap-ing/not for the eventual reporting?

PR #1248, opened 2021-04-29
Updated content type to prevent false positives

kodster28:


products/bots/src/content/about/plans/biz-and-ent.md
```diff
@@ -2,7 +2,7 @@
 title: Business
 order: 2
 type: document
-content-type: Reference
+pcx-content-type: Reference
```
pedrosousa: I'd use the same capitalization throughout the docs, probably `reference` (and also `faq`).

pedrosousa (approved):

PR #1251, opened 2021-04-29
pages: add private-registry how-to

### LINKED ISSUE #942: Add instructions for using private npm registries with Pages
@signalnerve 

1. Create and check in a `.npmrc` file with your registry ([details](https://docs.npmjs.com/cli/v7/configuring-npm/npmrc))
2. If using an environment variable (which are automatically substituted in `.npmrc` files), set that variable in your deployment settings (or wherever the tab is for setting environment variables)

e.g.

.npmrc:
```
@myscope:registry=https://npm.myprivateregistry.com/
//npm.myprivateregistry.com/:_authToken=${MY_AUTH_TOKEN}
```

environment variable config:
| Key           | Value        |
|---------------|--------------|
| MY_AUTH_TOKEN | XXX-XXX-XX-X |

lukeed:
Couldn't really decide on a URL and H1 title 🤷 But I did verify that all this works with a real Pages site using private registries.

Closes #942

### DIFF
--- a/products/pages/src/content/how-to/npm-private-registry.md
+++ b/products/pages/src/content/how-to/npm-private-registry.md
@@ -1,0 +1,146 @@
+# Install Private Packages
+
+Pages supports custom package registries, allowing you to include private dependencies in your application. While this walkthrough may focus specifically on [npm](https://www.npmjs.com/), the Node package manager and registry, the same approach can be applied to other registry tools.
+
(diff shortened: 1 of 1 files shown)


kristianfreeman (approved):

miguelcobain:
@lukeed how can the build command `mv .npmrc.pages .npmrc && npm run build` work, if the build command runs **after** `npm install` itself?

By the time the `mv .npmrc.pages .npmrc` runs, the build already tried to npm install before and fails because there's no `.npmrc`.

lukeed:
@miguelcobain Thanks, last night I noticed I had a dangling `preinstall` script locally 🙃 Sending in a follow-up PR

miguelcobain:
@lukeed @signalnerve I think for the time being we should remove this how to, right?
It's leading people to a dead end...

lukeed:
@miguelcobain There is a follow up to this w/ the fixes #1260

PR #1262, opened 2021-05-02
optimize query

drhino:
no loop

### DIFF
--- a/products/workers/src/content/examples/extract-cookie-value.md
+++ b/products/workers/src/content/examples/extract-cookie-value.md
@@ -21,27 +21,26 @@ const COOKIE_NAME = "__uid"
  * Returns a cookie value or null.
  * @param {Request} request incoming request
  * @param {string}  key of the cookie to get
- * @returns {string} value of the cookie or null
+ * @returns {string|void} value of the cookie if found
  */
 function getCookie(request, key) {
-  // No cookies were found in the request headers.
-  if ( ! request.headers.has('Cookie') )
-    return null
+  const cookie = request.headers.get('Cookie')
+  
+  // No cookie found
+  if (!cookie) return
 
   // Search for the cookie key in the header.
-  const str = request.headers.get('Cookie'),
-     search = `${key}=`,
-     starts = str.indexOf(search)
+  const search = `${key}=`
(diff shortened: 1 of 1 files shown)


products/workers/src/content/examples/extract-cookie-value.md
```diff
@@ -18,25 +18,30 @@ tags:
...
+
+  // The cookie could not be found.
+  if (starts === -1)
+    return null
+
+  // Parse the cookie value.
+  const val = str.substr(starts + search.length, str.length),
+        end = val.indexOf(';')
+
+  return (end === -1) ? val : val.substr(0, end)
```
lukeed:
```suggestion
function getCookie(request, key) {
  const cookie = request.headers.get('Cookie')
  
  // No cookie found
  if (!cookie) return

  // Search for the cookie key in the header.
  const search = `${key}=`
  const starts = cookie.indexOf(search)

  // The cookie could not be found.
  if (starts === -1) return

  // Parse the cookie value.
  const value = cookie.substring(starts + search.length, cookie.length)
  const end = value.indexOf(';')

  return end === -1 ? value : value.substring(0, end)
```

products/workers/src/content/examples/extract-cookie-value.md
```diff
@@ -18,25 +18,30 @@ tags:
 const COOKIE_NAME = "__uid"
 
 /**
- * Gets the cookie with the name from the request headers
- * @param {Request} request incoming Request
- * @param {string} name of the cookie to get
+ * Returns a cookie value or null.
+ * @param {Request} request incoming request
+ * @param {string}  key of the cookie to get
+ * @returns {string} value of the cookie or null
```
lukeed:
```suggestion
 * @returns {string|void} value of the cookie if found
```

lukeed (commented):
Slight refactor to the code, main difference is checking for the header value once & returning `void` instead of `null` in the not-found conditions

kristianfreeman (approved):

drhino:
@lukeed first sentence of comment block:
"Returns a cookie value or null." — still has the "null".

maybe: "Returns the value of a Cookie."

PR #1287, opened 2021-05-06
Added content type to 3 doc repos

kodster28:
Bots edits
Load balancing
Waiting Room
Page Shield

### DIFF
--- a/products/load-balancing/src/content/understand-basics/traffic-steering.md
+++ b/products/load-balancing/src/content/understand-basics/traffic-steering.md
@@ -1,6 +1,7 @@
 ---
 title: Traffic steering
 order: 12
+pcx-content-type: reference
 ---
 
 import RegionMapping from "../_partials/_region-mapping.md"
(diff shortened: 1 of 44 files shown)


products/waiting-room/src/content/how-to/index.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: Work with Cloudflare Waiting Rooms
-alwaysopen: true
-weight: 150
-hidden: false
+order: 2
+pcx-content-type: tutorial
```
kodster28, line 4: Not the best example of a `tutorial`, but some upcoming changes should make this a better example.

products/load-balancing/src/content/understand-basics/traffic-steering.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Traffic steering
 order: 12
+pcx-content-type: reference
```
kodster28, line 4: I'd imagine that a common use for `reference` is when we're providing more detail about a specific dropdown field in the UI.

dcpena (approved):

PR #1328, opened 2021-05-12
WAF: Add content type information to topics

pedrosousa:
These changes have no impact in the rendered docs. It's for internal purposes only.

### DIFF
--- a/products/waf/src/content/exposed-credentials-check/test-configuration.md
+++ b/products/waf/src/content/exposed-credentials-check/test-configuration.md
@@ -1,6 +1,6 @@
 ---
 title: Test your configuration
-pcx-content-type: concept
+pcx-content-type: reference
 order: 4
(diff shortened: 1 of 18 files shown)


products/waf/src/content/exposed-credentials-check/test-configuration.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Test your configuration
+pcx-content-type: concept
```
kodster28: To me, this feels more like a `reference`.

kodster28 (approved):

products/waf/src/content/exposed-credentials-check/test-configuration.md
pedrosousa: Had some questions about this one... but I agree.

PR #1363, opened 2021-05-17
[Workers] Change WebsocketPair to WebSocketPair in runtime-apis/websockets.md

Mooshua:
- Based on trial/error & other documentation, it seems WebSocketPair is correct and attempting to use WebsocketPair causes an exception.
- Replaced instances of WebsocketPair with WebSocketPair in all code examples
- Replaced one instance of WebsocketPair in the body of the documentation to keep consistency.


lukeed (approved):
Thanks!

PR #1365, opened 2021-05-17
[Stream] Add percentComplete to example download responses

bkrebsbach:
Add percentComplete to the example MP4 download API responses.


Everlag (approved):
👍

PR #1369, opened 2021-05-17
Test waiting room

kodster28:
Addresses PCX-1639.

Preview image for ease of reviewing:
image


kcantrel:
Here are my suggestions:

1. In the "Before you begin" section you might add a little more explanation as to why you might have to modify the sample script if they don't use the default responses like: "If you use a custom response template, you might have to update the sample script to ensure it picks up on when a "simulated user" is in the waiting room, and when they are released to the origin site."
2. In the "Download sample script" section, since "Phase 1" is as load increases, you might change the order and say "It divides traffic into two phases (Phase 1 and Phase 2) so you can test wait times as the load increases and then decreases.
3. In the Run sample script section, for the "-s" option explanation, there is either a missing "(" or an extra ")" at the end.
4. After the example run, the first sentence needs to be rewritten. Maybe something like "As the script runs, it will output characters. Each character represents a user session that advanced past the waiting room. Each letter corresponds to the amount of time the user waited in the wait room."
5. Also, the letters don't actually represent a range of time. It is: a = 10 seconds, b = 20 seconds, since the script only checks to see if the user can be passed through every 10 seconds. Also, you might include "0 = 530 seconds, 1 = 540 seconds ..."
6. In the Analyze results section. For "job:" Maybe change the second sentence to "The fixed string will be either main for phase 1 or post for phase 2."
7. My final suggestion might be to get of the graph at the bottom, since it clearly shows a problem with the Waiting Room in that it releases people earlier than it should. I'll run my test again and see if I get a better result.

kodster28:
slightly revised screenshot for PCX. **Note**, still need the updated screenshot from Keith.

image

products/waiting-room/src/content/how-to/create-waiting-room/create-waiting-room-dashboard/configure-settings.md
```diff
@@ -1,5 +1,6 @@
 ---
 order: 0
+type: overview
 pcx-content-type: configuration
```
crwaters16, line 4: File has both a "type" and a "pcx-content-type"

products/waiting-room/src/content/how-to/test-waiting-room.md
```diff
@@ -0,0 +1,114 @@
+---
+order: 5
+pcx-content-type: tutorial
+---
+
+# Test a Waiting Room
+
+Follow this tutorial to make sure your Waiting Room queues and admits users as expected.
```
crwaters16: Consider "ensure" for conciseness

products/waiting-room/src/content/how-to/test-waiting-room.md
```diff
@@ -0,0 +1,114 @@
...
+
+## 1. Download sample script
+
+First, download the [sample script](https://github.com/kcantrel/test-cf-waitingroom/blob/master/simulate_requests) from GitHub.
+
+This script simulates users entering a waiting room. It divides traffic into two phases (`Phase 1` and `Phase 2`) so you can test wait times as load increases and then decreases.
+
+## 2. Run sample script
+
+Once you have downloaded the script, run it with the following command-line arguments:
```
crwaters16, line 33: This bulleted lists ends with periods and the list under "As the script runs..." (line 66) does not. I don't think we have style guide guidance on which to go with, so I'd just pick one for consistency's sake within this tutorial.

products/waiting-room/src/content/how-to/test-waiting-room.md
```diff
@@ -0,0 +1,114 @@
...
+
+## 3. Analyze results
+
+Once the script finishes running, it creates a CSV file with the following fields:
+
+<details>
+  <summary>Fields in CSV file</summary>
+  <div>
+    <ul>
+        <li><strong>job</strong>: The fixed string will either be <strong>main</strong> for phase 1 or <strong>post</strong> for phase 2.</li>
```
crwaters16: Same comment here about periods at the end of each list item.

crwaters16 (approved):
Small comments - looks good!

products/waiting-room/src/content/how-to/create-waiting-room/create-waiting-room-dashboard/configure-settings.md
kodster28: That's fine! The `type` value controls the page layout + I set up the reporting stuff to ignore anything that doesn't match the full `pcx-content-type` string.

PR #1383, opened 2021-05-20
Correct an incorrect TOML Format

phily245:
A direct copy/paste of these results in an invalid TOML error, but the correct format is used elsewhere and is valid.

We ran across this earlier today and this was the fix.

Issue: https://github.com/cloudflare/cloudflare-docs/issues/1384

### DIFF
--- a/products/workers/src/content/cli-wrangler/configuration.md
+++ b/products/workers/src/content/cli-wrangler/configuration.md
@@ -330,15 +330,14 @@ Usage:
 
 ```toml
 [build]
-upload.format = "service-worker"
 command = "npm install && npm run build"
+[build.upload]
+format = "service-worker"
 ```
 
+#### `[build]`
+
 <Definitions>
-
-- `upload.format` <PropMeta>required</PropMeta>
-
-  - The format of the Worker script, must be "service-worker"
 
 - `command` <PropMeta>optional</PropMeta>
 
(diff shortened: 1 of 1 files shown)


Electroid:
I think it's _technically_ valid, but popular editors like VSCode are marking it incorrectly. I think we should approve this.

phily245:
> I think it's _technically_ valid, but popular editors like VSCode are marking it incorrectly. I think we should approve this.

It may be, I'm by far no TOML expert!

Some background for this PR:

I submitted it because was failing in nodejs jest integration tests when using https://www.npmjs.com/package/toml to parse the file to extract variables, which then lets me seed the KV store data. The error was:

```
SyntaxError: Expected "=", [ \t] or [A-Za-z0-9_\-] but "." found.
```
I double checked on https://www.toml-lint.com/ to try to do my due diligence before my PR

I made a code sandbox to demonstrate my use case: https://codesandbox.io/s/admiring-austin-1oq5g

xortive (approved):

xortive:
dotted keys were added in toml 0.5, which is surprisingly unsupported by many toml parsers... Let's just standardize on 0.4, which we've already done in our templates and the majority of the docs

Electroid (approved):

phily245:
> dotted keys were added in toml 0.5, which is surprisingly unsupported by many toml parsers... Let's just standardize on 0.4, which we've already done in our templates and the majority of the docs

Thanks for the clarification! :smile:

PR #1391, opened 2021-05-21
[Teams] add new block page docs

abracchi-tw:
+ other fixes


TownLake (approved):

products/cloudflare-one/src/content/policies/filtering/configuring-block-page.md
```diff
@@ -0,0 +1,93 @@
+---
+order: 5
+---
+
+# Custom block page
```
ptzimmerman: Let's call this the "Block Page" and add a section "Customize the Block Page"

ptzimmerman: All the references in this section should be to a "block page"

products/cloudflare-one/src/content/policies/filtering/configuring-block-page.md
```diff
@@ -0,0 +1,93 @@
...
+
+Follow the instructions below to finish configuring the block page:
+
+1. Enter **about:config** in the address bar.
+1. Click on **Accept the risk!** if you see a prompt from Firefox.
+1. Set `security.enterprise_roots.enabled` to `true`.
+
+## Enable the block page
+
+The Custom Block Page is automatically enabled for all HTTP policies. For DNS policies:
```
ptzimmerman: "The Block Page is automatically enabled for all HTTP policies. When a user is blocked at the HTTP layer, the block page is displayed." You can configure a message explaining why the block occurred to a user on a per-policy basis.

ptzimmerman: Separate line: "For DNS policies:..."

ptzimmerman: We're trying to say that for HTTP, they don't need to do anything because the block page will always be presented. If they choose, they can input the block reason text in each HTTP policy to inform users why the block occurred. 

For DNS, they need to enable the block page in each DNS policy since Gateway responds to blocked queries with 0.0.0.0 by default. When they choose to enable the block page, an input field for blocked reason will appear (and look and behave the same as HTTP policies).

products/cloudflare-one/src/content/policies/filtering/configuring-block-page.md
```diff
@@ -0,0 +1,93 @@
...
+
+<Aside header='Warning: Potential Security Risk'>
+
+If your users are seeing the following error when navigating to a blocked page, double-check that you have downloaded the Cloudflare certificate and you have added it correctly to your system.
+
+![Block page](../../static/documentation/policies/https-browser-error.png)
+
+</Aside>
+
+## Customize the block page
```
ptzimmerman, line 86: We should expand this to say what they're customizing on the block page (what input fields correspond to which parts of the block page. 

Then we have a separate paragraph explaining where the "block reason" text gets input on the custom block page (it goes below the template text and above the rule ID at the bottom"

ptzimmerman: We should also mention that the rule ID will always be displayed on the page to make it easier for admins to debug which rule matched.

products/cloudflare-one/src/content/policies/filtering/configuring-block-page.md
abracchi-tw: makes sense. at this point we may need to explicitly call out that the customizations in the Settings page apply across DNS and HTTP block pages. will add a paragraph

dcpena (approved):
Looks good to me! The only feedback I have is to potentially eliminate some of the screenshots. I think some are worth including, like the image of the error message, but the screenshot that shows the Download Certificate button could likely be taken out.

PR #1395, opened 2021-05-21
point out that caches.open is async unlike caches.default

koeninger:


koeninger:
@rita3ko or @greg-mckeon minor doc clarification that would have saved me some time today :)

products/workers/src/content/learning/how-the-cache-works.md
```diff
@@ -53,7 +53,8 @@ The Cache API can be thought of as an ephemeral key-value store, whereby the `Re
 There are two types of cache namespaces available to the Cloudflare Cache:
 
 - __`caches.default`__ – You can access the default cache (the same cache shared with `fetch` requests) by accessing `caches.default`. This is useful when needing to override content that is already cached, after receiving the response.
-- __New namespaces__ – When writing new content to the cache, for example after running a more compute heavy operation such as parsing HTML, or running a computation, you can use `caches.open(CACHE_NAME)` to store them locally in the colo, and readily access them on the following request, rather having to rerun the same operation.
+- __New namespaces__ – You can access a namespaced cache (separate from the cache shared with `fetch` requests) using `let cache = await caches.open(CACHE_NAME)`. This can be useful when writing new content to the cache, for example after running a more compute heavy operation such as parsing HTML or running a computation, to store them locally in the colo and readily access them on the following request, rather having to rerun the same operation. Note that `caches.open` is an async function, unlike `caches.default`.
```
lukeed:
```suggestion
- __New namespaces__ – You can access a namespaced cache (separate from the cache shared with `fetch` requests) using `let cache = await caches.open(CACHE_NAME)`. This can be useful when writing new content to the cache, for example after running a more compute heavy operation such as parsing HTML or running a computation, to store them locally in the colo and readily access them on the following request, rather having to rerun the same operation. Note that [`caches.open`](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage/open) is an async function, unlike `caches.default`.
```

Maybe worth pointing to the [MDN documentation page](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage/open)?

kristianfreeman:
i had no idea that `caches.open` exists! should we document it on our [cache](https://developers.cloudflare.com/workers/runtime-apis/cache) page in the runtime apis section? cc @rita3ko @deadlypants1973

rita3ko:
@signalnerve yes! make a new issue?

greg-mckeon:
This LGTM

greg-mckeon (approved):

PR #1408, opened 2021-05-24
Initial api security docs. 

kodster28:
Addresses PCX-1705.

PM is okay with temporarily using the same product icon as page shield.

Also relates to https://github.com/cloudflare/cloudflare-docs-engine/pull/336.

### DIFF
--- a/products/api-security/README.md
+++ b/products/api-security/README.md
@@ -1,0 +1,5 @@
+# Cloudflare API Security docs
+ 
+[View docs →](https://developers.cloudflare.com/api-security)
+ 
+[Read contributor guidelines →](https://developers.cloudflare.com/docs-engine/contributing/content-framework)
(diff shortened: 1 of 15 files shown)


products/api-security/README.md
```diff
@@ -0,0 +1,5 @@
+# Cloudflare API Security docs
+ 
+[View docs →](https://developers.cloudflare.com/api-security)
+ 
+[Read contributor guidelines →](https://developers.cloudflare.com/docs-engine/contributing/content-framework)
```
dcpena: Something I thought about in my last PR was that we point to the contribution guidelines for the docs-engine, but maybe we should be pointing to the contribution guidelines in the repo? 
https://github.com/cloudflare/cloudflare-docs/blob/10189c62728443109c98c82cff9d6900ceb185dc/CONTRIBUTING.md

We don't have to make the update now though. Might be worth addressing all at once in a separate ticket!

dcpena (approved):

PR #1411, opened 2021-05-25
[SSL client cert] add Troubleshooting page as troubleshooting.md and add Android client cert example in configure-your-mobile-app-or-iot-device.md

zechu-cf:
- Add Troubleshooting page for Client Certificates
- Add an example of embedding a client certificate in an Android app
- Change ordering of subpages so that "Enable mTLS" comes before "Configure your mobile app or IoT device" since it may be more intuitive that way


pedrosousa:
> * Change ordering of subpages so that "Enable mTLS" comes before "Configure your mobile app or IoT device" since it may be more intuitive that way

I believe the order is "the less intuitive one" because it's the one that doesn't cause any downtime. 

Also, in "Configure your mobile app or IoT device" there's a [link to the next step](https://github.com/cloudflare/cloudflare-docs/blob/29a973e37b31216d5f30bc6053c740df88fda945/products/ssl/src/content/client-certificates/configure-your-mobile-app-or-iot-device.md#enable-mtls), which should probably be changed if we're reordering the steps.

zechu-cf:
> I believe the order is "the less intuitive one" because it's the one that doesn't cause any downtime.

That makes sense. I'll revert to the previous ordering.

kodster28:
Howdy @zechu-cf, I updated a few small things. 

Biggest thing I changed was to remove two of the images on the `Troubleshooting` page. There's a maintenance cost associated with each image and those didn't provide a lot of additional troubleshooting value.

kodster28 (approved):

PR #1419, opened 2021-05-25
Added notes about ruleset engine to SBFM docs

kodster28:
Addresses PCX-1738.


products/bots/src/content/_partials/_bfm_ruleset_engine.md
```diff
@@ -0,0 +1 @@
+Super Bot Fight Mode runs during the `firewall_managed phase` of the [Ruleset Engine](https://developers.cloudflare.com/firewall/cf-rulesets#phases).
```
pedrosousa: I'd use the full phase name (`http_request_firewall_managed`) and leave "phase" out of the monospaced formatting.

Also, I'd use hyphens in the filename instead of underscores, except for the first underscore (following the filenames of other partials in the same folder).

pedrosousa (approved):

PR #1431, opened 2021-05-26
fix(workflow): normalize indentation

lukeed:
Noticed that the [Build Action](https://github.com/cloudflare/cloudflare-docs/actions/workflows/build.yml) has been failing. Looks like indentation got funky – YAML things


.github/workflows/build.yml
```diff
@@ -649,25 +649,24 @@ jobs:
...
-    - uses: actions/checkout@v2
-    - uses: actions/setup-node@v1
-      with:
-        node-version: 14.2.0
-    - run: |
-        npm install
-        mkdir .docs
-        npm run ghactionsbootstrap
-        npm run build
+  build-railgun:
```
lukeed, line 652: Specifically this block – was top-level

kristianfreeman (approved):

PR #1442, opened 2021-05-27
Fixed code bug

olawanlejoel:
I was following this article to develop something robust but unfortunately, I ran into this bug for over an hour until I later discovered the slash which made the query not to run!


kristianfreeman:
thanks! what's the error message with the slash? this is a little surprising to me!

olawanlejoel:
It was not displaying the data in the dynamic route, I ran all forms of text until I discovered the slash was not performing any function but rather blocking the query from running.

PR #1453, opened 2021-05-28
Update small issues in TF tutorial docs

garrettgalow:
list of fixes:
hello world
	1. fix provider setup
	2. fix ability to copy the initial config
	3. add instructions to create api token
source control
	1. fix sed replaces
	2. update head command
zone settings
	1. update var.domain to var.zone_id

### DIFF
--- a/products/terraform/src/content/tutorial/hello-world.md
+++ b/products/terraform/src/content/tutorial/hello-world.md
@@ -7,17 +7,26 @@ order: 1
 
 Let’s say you have a web server for your domain that’s accessible on 203.0.113.10. You just signed up your domain, example.com, on Cloudflare and want to manage everything in Terraform.
 
-This tutorial step shows you how to get started. Before you do so, make sure you’ve [installed Terraform](/installing).
+This tutorial step shows you how to get started. Before you do so, make sure you’ve [installed Terraform](/installing). You will need to [create an API Token](https://developers.cloudflare.com/api/tokens/create) with permissions to edit resources for this tutorial.
 
 ## 1. Defining your first Terraform config file
 
 First we’ll create a initial Terraform config file. Any files ending in `.tf` will be processed by Terraform. As you configuration gets more complex you’ll want to split the config into separate files and modules, but for now we’ll proceed with a single file:
 
-```sh
+```
 $ cat > cloudflare.tf <<'EOF'
+terraform { 
+  required_providers {
+    cloudflare = {
+      source = "cloudflare/cloudflare"
+      version = "~> 2.0"
+    }
+  }
+}
+
 provider "cloudflare" {
-  email   = "you@example.com"
-  api_key = "your-api-key"
+      email = "you@example.com"
+      api_token = "your-api-token"
 }
 
 variable "zone_id" {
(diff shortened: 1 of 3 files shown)


garrettgalow:
this fixes internal ticket PCX-967

products/terraform/src/content/tutorial/hello-world.md
```diff
@@ -7,17 +7,26 @@ order: 1
...
 Let’s say you have a web server for your domain that’s accessible on 203.0.113.10. You just signed up your domain, example.com, on Cloudflare and want to manage everything in Terraform.
 
-This tutorial step shows you how to get started. Before you do so, make sure you’ve [installed Terraform](/installing).
+This tutorial step shows you how to get started. Before you do so, make sure you’ve [installed Terraform](/installing). You will need to [create an API Token](https://developers.cloudflare.com/api/tokens/create) with permissions to edit resources for this tutorial.
 
 ## 1. Defining your first Terraform config file
 
 First we’ll create a initial Terraform config file. Any files ending in `.tf` will be processed by Terraform. As you configuration gets more complex you’ll want to split the config into separate files and modules, but for now we’ll proceed with a single file:
 
-```sh
```
deadlypants1973, line 16: sh should be retained in ```sh

We follow this syntax in the docs because it makes text highlightable.

deadlypants1973 (changes requested):
Need to retain sh after ```sh because this makes the text highlightable.

products/terraform/src/content/tutorial/hello-world.md
```diff
@@ -7,17 +7,26 @@ order: 1
...
+```
```
deadlypants1973:
```suggestion
```sh
```

adding sh back in so that code in the docs remains highlightable.

garrettgalow: sh blocks being able to copy all of the text window, since this is all input. The current docs have this issue. I think removing the SH will keep it as regular code block which will allow for copying the whole 'code' block.

lukeed: I agree that it's unfortunate our syntax-highlighting kills text-highlighting (lol) and removes the ability to copy-paste the entire snippet. That said, i feel like it might be better to keep things consistent and then remove the `user-select: none` CSS from the docs-engine/site assets directly. Just my take, wdyt?

lukeed: Specifically, this is the CSS that kills text selection (for `sh`, at least):

```css
[language=sh] .CodeBlock--token-plain, [language=sh] .CodeBlock--token-unselectable {
    -webkit-user-select: none; /** << THIS **/
    -ms-user-select: none; /** << THIS **/
    user-select: none; /** << THIS **/
    pointer-events: none;
    text-decoration: none;
    transition: opacity .25s ease;
}
```

pedrosousa: When left empty, the default syntax highlighting is `javascript`... You probably want to use `bash` here, which allows readers to copy the text block. To disable syntax highlighting, use `txt`.

lukeed: Ah ok cool, thanks for the clarification @pedrosousa – seems like `bash` would be the best bet here imo

garrettgalow: yeah I didn't realize now keyword would default to js. That's not ideal. bash is probably right. let me test it locally to confirm and I'll update

garrettgalow: updated it to `bash`. I checked locally and it allows for full text copying while having pretty good syntax highlighting.

deadlypants1973 (approved):

PR #1456, opened 2021-05-30
Update docs to aviod using "simple"

Ekwuno:
I think it's great to avoid using words like “easy”, “simple” and “basic” in docs, especially when Introducing a new method of a thing, because if users have a hard time completing the task that is supposedly “easy,” they will question their abilities. And we want people to feel comfortable learning.

### DIFF
--- a/products/workers/src/content/learning/playground.md
+++ b/products/workers/src/content/learning/playground.md
@@ -4,7 +4,7 @@ order: 1
 
 # Playground
 
-The quickest way to experiment with Cloudflare Workers is in the [Playground](https://cloudflareworkers.com/#36ebe026bf3510a2e5acace89c09829f:about:blank). It doesn’t require _any_ setup. It’s just a simple, instant way to preview and test a Workers script directly in the browser against any site.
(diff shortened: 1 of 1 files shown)


greg-mckeon:
Hey @Ekwuno! Thanks for the PR.

I agree with you we don't want to make developers feel uncomfortable.  In this case, these sentences aren't referring to tasks a user is completing.  

The first is calling the sandbox simple, which is true - it doesn't have robust configuration options available, which are available when configuring a script via Wrangler.

The second is referring to the Workers script we've defaulted to showing there, which is indeed the least complex script you can write with a worker.  Simple serves to tell developers that there are more complex Workers they could write in the sandbox, and that this is a starting point for them.

If you have other examples where we're implying that a task is actually easy, let me know!

Ekwuno:
Hey, @greg-mckeon I see what you mean, but I was trying to point out that using "simple" in docs can be a little off-putting, especially when the developer reading it doesn’t understand what just happened. 

Don’t you think that if someone saw the "simplest" worker function and couldn’t understand it because maybe they didn't understand for example, what the respondWith runtime API was doing. It wouldn’t encourage them to even look at the complex ones?

greg-mckeon:
Yep, that's a good point! I agree we should make our description of what's happening here a bit better, since it's a starting point for the platform (and especially since it re-uses the word essentially twice :) ).

greg-mckeon (approved):

Ekwuno:
Awesome! Let's get more people comfortable with the docs. Thanks, @greg-mckeon for taking the time to review this and meeting me in the middle.

PR #1459, opened 2021-05-31
Correct “Gigabytes per sec” to “Gigabyte-seconds”

DuncanMacWeb:
Gb-s is appropriate for this, as the meaning is “the amount of memory used, multiplied by (or _over_) the total duration.” Thus in the example given, the memory used can be correctly expressed as *0.125 GB/s* (gigabytes per second) because the average amount of memory used in each second is 0.125GB. By contrast, what is intended is to communicate not the memory used per second, but instead the cumulative memory used, over the whole execution time, adding each second of time together. Gigabyte-seconds expresses this concept.

### DIFF
--- a/products/workers/src/content/platform/limits.md
+++ b/products/workers/src/content/platform/limits.md
@@ -156,7 +156,7 @@ There is no limit on the real runtime for a Workers script. As long as the clien
 
 ## Duration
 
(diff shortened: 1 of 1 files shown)


greg-mckeon (approved):

greg-mckeon:
Thanks for the fix, @DuncanMacWeb!

PR #1466, opened 2021-06-01
Update configuration.md

xortive:


products/workers/src/content/cli-wrangler/configuration.md
```diff
@@ -385,10 +385,9 @@ Cloudflare Workers now supports uploading scripts as a collection of modules, in
 import html from './index.html'
 
 export default {
-  // request is almost the same as `event.request` from the service worker format, except for:
+  // * request is the same as `event.request` from the service worker format
   // * waitUntil() and passThroughOnException() are accessible from ctx instead
```
a-robinson: If you don't mind making another iteration on this, it isn't obvious from context what "instead" refers to here. Maybe just remove that word? Or clarify that it's instead of on the `event` object, but then we have to explain what the `event object is.

a-robinson (approved):
Thanks for finding this!

products/workers/src/content/cli-wrangler/configuration.md
a-robinson: ping? Or do you just want me to merge?

lukeed: Not sure that need to explain `event` here. Also might be worth having a side-by-side comparison of the same Worker in SW vs ESM formats – could then link out to it here instead of having an inline list.

```suggestion
  // * waitUntil() and passThroughOnException() are accessible from `ctx` instead of `event` from the service worker format
```

xortive: Sorry, missed this in my notifications. I like what @lukeed suggested below, I'll use that

xortive: Yeah, once we have separate pages for modules we can do that. This whole section is a bit of a mess anyways, these docs deserve their own page not to hang out in the wrangler configuration section...

a-robinson (approved):
👍

PR #1495, opened 2021-06-05
[Durable Objects] remove note regarding request.cf availability in durable-objects.md

nicgene:
- It appears that request.cf is now available in Durable Objects.


a-robinson (approved):
Thanks!

PR #1496, opened 2021-06-07
[Cloudflare for Teams] Add to Tunnel docs and edit IdP integration

zechu-cf:
- Small edits to SAML and OIDC integration
- Added common cloudflared commands to Tunnel docs
- Added hosting requirements for Tunnels
- Added "secure the server" section to Tunnel docs
- Added troubleshooting guide for Tunnels


products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
```diff
@@ -0,0 +1,69 @@
+---
+order: 8
+---
+
+# Common `cloudflared` Commands
+
+Here are some of the commonly used `cloudflared` commands that you may find useful when building your own Tunnel. You may refer to the full list via `cloudflared --help`
+
+## `cloudflared tunnel run`
+Start running a Cloudflare Tunnel with the given UUID or name. If no UUID or name is given, the Tunnel specified in your `config.yml` file will be run. You may specify optional flags in the following format:
```
nmldiegues: > the Tunnel specified in your `config.yml` file will be run

maybe you can say which properties of the config are picked up in that case

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
```diff
@@ -0,0 +1,69 @@
...
+```sh
+$ cloudflared tunnel [FLAG(s)] run <tunnel UUID or NAME>
+```
+
+## `cloudflared update`
+Update `cloudflared` to the latest version.
+
+## `cloudflared proxy-dns`
+Run a [DNS over HTTPS proxy server](https://developers.cloudflare.com/1.1.1.1/dns-over-https/cloudflared-proxy).
```
nmldiegues: This is a tricky one. If we talk about this, then why don't we talk about `cloudflared access`?
But if we start talking about those, then this is no longer about Tunnel anymore.

In fact, you can think that cloudflared is a tool for 3 purposes:
 - Tunnel (origin)
 - Access (eyeballs)
 - DNS proxying (dns on eyeball)

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
```diff
@@ -0,0 +1,69 @@
...
+| `--no-tls-verify` | Disables TLS verification of the certificate presented by your origin. It will allow any certificate from the origin to be accepted. Note: The connection from your machine to Cloudflare's Edge is still encrypted. (default: false) |
+| `--loglevel value` | Application logging level {`debug`, `info`, `warn`, `error`, `fatal`}. At `debug` level `cloudflared` will log request URL, method, protocol, content length, as well as all request and response headers. This is a useful debugging tool when running a Tunnel: `cloudflared tunnel --loglevel debug run`. However, please note that this can expose sensitive information in your logs. (default: `info`) |
+| `--transport-loglevel value` | Transport logging level (previously called protocol logging level) {`debug`, `info`, `warn`, `error`, `fatal`} (default: `info`) |
+| `--logfile value` | Saves application log to this file for reporting issues. |
+| `--log-directory value` | Saves application log to this directory for reporting issues. |
+| `--trace-output value` | Name of trace output file, generated when `cloudflared` stops. |
+
+</TableWrap>
```
nmldiegues: I am trying to wrap my head around the value of this page.

On the one hand, it is cool to have this summary of useful commands in a single page that I can scroll through quickly.

On the other hand, all this information should already be in other pages around the same docs --- maybe some bit is missing, and it should be added, but otherwise I expect all of this to already be documented. So we're adding redundancy, which I'm not sure brings any value (and it certainly can contribute to some parts becoming stale as we evolve cloudflared  and update 1 place but forget the other).

Thoughts?

products/cloudflare-one/src/content/connections/connect-apps/hosting-requirements.md
```diff
@@ -0,0 +1,11 @@
+---
+order: 7
+---
+
+# Tunnel Hosting Requirements
+
+Running `cloudflared` instances on a server and proxying traffic through it require computing resources such as CPU and memory on the server. The actual amount of resource usage depends on many variables, including the number of requests per second, bandwidth, network latency and hardware environment.
+
+Although there is a physical limit on the amount of traffic that a single instance of `cloudflared` can proxy on a single server, you may easily scale up the capacity of your Tunnel by running many instances of `cloudflared` [over the same Tunnel multiple times](https://blog.cloudflare.com/highly-available-and-highly-scalable-cloudflare-tunnels/) on different machines or network cards, regardless of the baseline performance of one `cloudflared` instance.
```
nmldiegues: I would not phrase this in this way:

> there is a physical limit on the amount of traffic that a single instance of `cloudflared` can proxy on a single server

I would instead turn it around:

"The limit to how much traffic a single `cloudflared` can proxy will be bound to the hardware capabilities of the underlying machine (including network-wise)."

`cloudflared` per se has no limits in the code, it is embarrassingly parallel, and so it is really up to what the hardware can deliver. I think writing it in this way makes it more clear.

products/cloudflare-one/src/content/connections/connect-apps/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
+---
+order: 9
+pcx-content-type: faq
+---
+
+# Troubleshooting
```
nmldiegues: There's already a FAQ at https://developers.cloudflare.com/cloudflare-one/faq

Should this be moved there?

nmldiegues:
Thanks for these contributions @zechu-cf 

I'm curious how did you land on all these "scattered" improvements. Were you overall trying to use/employ Tunnel in solutions for customers, and accumulated these suggestions over time?

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
joliveirinha: I agree. Particularly with this last table. I have mixed feelings because I do think that a summary is a nice thing to have, but it may difficult to guarantee that all places are in sync.

That said, a good compromise may be to remove this last table since we are not adding much more value.

zechu-cf: My original intention was to create a quick summary for the useful commands so that it is easier to know what commands are available and how to use them. But you raised a good point about redundancy and consistency (when there're changes in the future) which I didn't really think about. If this page doesn't add too much value we can remove it from the docs.

products/cloudflare-one/src/content/connections/connect-apps/hosting-requirements.md
zechu-cf: Thanks for the rephrasing. I've edited it in the latest commit.

products/cloudflare-one/src/content/connections/connect-apps/troubleshooting.md
zechu-cf: Yes, that works too. I've edited it in the latest commit.

zechu-cf:
Hi @nmldiegues , thanks for reviewing the PR. These were mainly from my experience of using Tunnel + suggestions from my manager based on his experience with the customers.

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
abracchi-tw: I would avoid redundancy, as the maintenance cost of duplicate information, as @nmldiegues brings up, is quite high. If that's ok, I propose we merge this PR without this page, and in a separate ticket we can think of ways to optimize the existing docs to address the need for a more digestible summary of commands. cc @abelinkinbio

products/cloudflare-one/src/content/faq/cloudflare-tunnels-faq.md
```diff
@@ -109,4 +109,31 @@ To log external visitor IPs, you will need to [configure an alternative method](
 
 ## Why does the name "warp" appear in some legacy materials?
 
-Cloudflare Tunnel was previously named Warp during the beta phase. As Warp was added to the Argo product family, we changed the name to match.
\ No newline at end of file
+Cloudflare Tunnel was previously named Warp during the beta phase. As Warp was added to the Argo product family, we changed the name to match.
+
```
zechu-cf, line 113: By the way, on Line 102 of this file: 

> Tunnels that use Cloudflare's Load Balancer use CNAME records. Tunnels that do not use the Load Balancer product will create AAAA records.

Is this information up to date?

nmldiegues: Does not look like it.

You can say that:
 - Named Tunnels can be routed via DNS records, in which case we use CNAME records to point to the uuid.cfargotunnel.com
 - Or as Load Balancer origins, which also point to uuid.cfargotunnel.com

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
nmldiegues: I think you should explicitly name the YAML property (`tunnel:`) that is picked up from the config

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
nmldiegues: Re-iterating on this, since it was not really addressed. My opinion is that we should focus this on `Common cloudflared commands for Cloudflare Tunnel`.

Otherwise, as I comment above, you would have to mention access commands as well.

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
```diff
@@ -0,0 +1,69 @@
...
+## `cloudflared proxy-dns`
+Run a [DNS over HTTPS proxy server](https://developers.cloudflare.com/1.1.1.1/dns-over-https/cloudflared-proxy).
+
+## `cloudflared service install`
+If you have already logged in and have a configuration file in `~/.cloudflared/`, this command will [run a Tunnel as a system service](/connections/connect-apps/run-tunnel/run-as-service).
+
+## `cloudflared service uninstall`
+Uninstall `cloudflared` as a system service.
+
+## `cloudflared route`
```
nmldiegues: There's also `route ip` for routing to private networks.

products/cloudflare-one/src/content/faq/cloudflare-tunnels-faq.md
```diff
@@ -109,4 +109,31 @@ To log external visitor IPs, you will need to [configure an alternative method](
...
+When the encryption mode is set to **Off (not secure)**, you may encounter connection issues when running a Tunnel.
+
+### Check location of credentials file
+If you encouter the following error when running a Tunnel, double check your `config.yml` file and ensure that the `credentials-file` points to the correct location. You may need to change `/root/` to your home directory.
+```sh
+$ cloudflared tunnel run
+2021-06-04T06:21:16Z INF Starting tunnel tunnelID=928655cc-7f95-43f2-8539-2aba6cf3592d
+Tunnel credentials file '/root/.cloudflared/928655cc-7f95-43f2-8539-2aba6cf3592d.json' doesn't exist or is not a file
+```
```
nmldiegues, line 139: While you are at it, it would be valuable to add this to the FAQ: https://jira.cfops.it/browse/TEAM-593


### cloudflared access has error "websocket: bad handshake"

This means that your cloudflared access client is unable to reach your cloudflared tunnel origin. 
To diagnose this, you should look at the cloudflared tunnel logs. A very often root cause is that the cloudflared tunnel is unable to proxy to your origin (e.g. because the ingress is mis-configured, or the origin is down, or because the origin HTTPS certificate cannot be validated by cloudflared tunnel).
If cloudflared tunnel has no logs, it means Cloudflare Edge is not even able to route the websocket traffic to it. There are two possible root causes there: 1) your cloudflared tunnel is not running or connected to Cloudflare Edge; or 2) you need to make sure that WebSockets are enabled in your dash.cloudflare.com (under the Network tab).

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
zechu-cf: In the latest commit, I've already removed this page as suggested by the comments below, so this was not updated.

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
zechu-cf: I've removed this page in the commit `remove common cloudflared commands page`, thus this was not updated.

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
zechu-cf: Thanks everyone for the feedback. I've removed this page in the latest commit.

products/cloudflare-one/src/content/faq/cloudflare-tunnels-faq.md
zechu-cf: Ok! will change it.

zechu-cf: The question prompt is `What is the difference between Tunnel creating a CNAME or AAAA record in the hostname's DNS setting?`. Does the question need to be changed as well? Since we don't use AAAA record for named Tunnels.

products/cloudflare-one/src/content/faq/cloudflare-tunnels-faq.md
zechu-cf: Ok! Added in the latest commit.

products/cloudflare-one/src/content/connections/connect-apps/common-cloudflared-commands.md
zechu-cf: I've removed this page based on the comments, so I didn't update changes related to the file `common-cloudflared-commands.md`.

products/cloudflare-one/src/content/faq/cloudflare-tunnels-faq.md
nmldiegues: Yeah that's a "legacy tunnels" question indeed.

You can maybe change it to:  What records are created for routing to a Named Tunnel's hostname?

zechu-cf: Alright. Changed it in the latest commit.

nmldiegues (approved):

abracchi-tw (approved):

PR #1501, opened 2021-06-07
Fix example value formatting

crstauf:


abracchi-tw:
hi @crstauf, this specific example has been causing a bit of confusion, will remove it until we have something clearer. Thanks for flagging!

abracchi-tw (approved):

PR #1510, opened 2021-06-07
Http3 redirects

kodster28:
Initial use case --> likely will be easier + quicker than fixing a link on the Cloudflare quic homepage
Additional use case --> Will be helpful if we do any content re-orgs in the future


DaoWays369:
Hf

PR #1521, opened 2021-06-08
[STREAM] Update signing utility documentation

bkrebsbach:
Update Stream documentation to reference new token signing API.

### DIFF
--- a/products/stream/src/content/viewing-videos/securing-your-stream.md
+++ b/products/stream/src/content/viewing-videos/securing-your-stream.md
@@ -14,7 +14,7 @@ Signed URLs are controlled by you and can be set to expire after a set time peri
 To implement signed URLs
 
 1. Create a key
-1. Make a video require signed URLS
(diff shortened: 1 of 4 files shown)


bkrebsbach:
@zaidf I'll hold off on merging this until I get the thumbs up from you.

products/stream/src/content/viewing-videos/securing-your-stream.md
```diff
@@ -14,7 +14,7 @@ Signed URLs are controlled by you and can be set to expire after a set time peri
 To implement signed URLs
 
 1. Create a key
```
zaidf, line 16: This step becomes optional if you are using the /token endpoint right? If so, I'd consider noting something like "(not required if you use the /token endpoint to generate the token)"

bkrebsbach: That's correct

Everlag (approved):

PR #1526, opened 2021-06-08
[MT] Added PCX content types to files

dcpena:
Added metadata used for PCX purposes to docs

Addresses [PCX-1665](https://jira.cfops.it/browse/PCX-1665)

### DIFF
--- a/products/magic-transit/src/content/magic-firewall/examples.md
+++ b/products/magic-transit/src/content/magic-firewall/examples.md
@@ -1,7 +1,7 @@
 ---
 title: Examples
 order:  5
-pcx-content-type: concept
+pcx-content-type: configuration
 ---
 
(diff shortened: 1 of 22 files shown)


products/magic-transit/src/content/magic-firewall/examples.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Examples
 order:  5
+pcx-content-type: concept
```
kodster28:
```suggestion
pcx-content-type: configuration
```

kodster28: Believe this is more specific than a concept, it's about applying a concept by setting specific values.

kodster28 (approved):

PR #1549, opened 2021-06-10
Added additional requirement for before you begin

kodster28:


products/registrar/src/content/_partials/_before-you-begin.md
```diff
@@ -1,4 +1,7 @@
-Before transferring a domain to or from Cloudflare:
+Before transferring a domain to Cloudflare:
+- Make sure Cloudflare is active on your domain, meaning that you:
```
dcpena: I think you could add this line to the previous line which would eliminate the consecutive colons.

dcpena (approved):

PR #1578, opened 2021-06-15
Add search to Dev Docs home page

pedrosousa:
- Added Algolia config with no `facetFilters` value to `developers.cloudflare.com/docs-config.js` to enable global search.
- Created a `<SiteSearch>` component specific for `developers.cloudflare.com`, very similar to the `<DocsSearch>` one in the docs engine, with the following differences:
    - Had to run `useStaticQuery` locally due to a bug that causes issues on relative imports, just like in the `<SEO>` component for `developers.cloudflare.com`.
    - Clicking a search result is handled differently. It navigates to the correct page (full page refresh), since the destination is outside the current GatsbyJS project.
    - Copied and adapted the original CSS so that the `<SiteSearch>` component fits the top header in `developers.cloudflare.com`. Also, [disabled the blur effect](https://github.com/cloudflare/cloudflare-docs/pull/1578/files#diff-63acc0317f184f4ef6d7f58a62e799c1170f2dc84cb4ecd92cc7053df7846950L53) in the site header because it interfered with the blur effect in the search results.
- Added the new `<SiteSearch>` component to the page header.
- Navigation links in the page header ("Docs" / "Sponsorships" / "Open Source" / "Newsletter") are now hidden when the width is below 1024px to make room for the search component.


jrf0110:
@gganch can you pull this branch down and test locally?

developers.cloudflare.com/docs-config.js
```diff
@@ -9,8 +9,8 @@ module.exports = {
     url: "https://cloudflare.com"
   }],
   search: {
-    indexName: "",
-    apiKey: "",
+    indexName: "developers-cloudflare",
+    apiKey: "b23088ab4d346409f9d3ece6606344c3",
```
jrf0110, line 13: This is the _search only_ api key, right?

developers.cloudflare.com/src/components/site-search.js
```diff
@@ -0,0 +1,167 @@
...
+  const enableSearch = indexName && apiKey && algoliaOptions
+
+  useEffect(() => {
+    let frames = 0
+    const init = () => {
+      frames += 1
+
+      // Sadly this is needed because of the way Helmet works in local development
+      if (!window.docsearch && frames < 60) {
+        return requestAnimationFrame(() => init())
```
jrf0110: This is _almost certainly_ not what we want. At worst, this will call the `window.docsearch` function 60 times

jrf0110: SO! I'm guessing that we just need some time to coordinate when certain elements are loading. I wonder if there's some dependency/ref we can put in the useEffect dependencies list

developers.cloudflare.com/src/components/site-search.js
```diff
@@ -0,0 +1,167 @@
...
+
+        // https://docsearch.algolia.com/docs/behavior
+        handleSelected: (input, event, suggestion, datasetNumber, context) => {
+          const url = new URL(suggestion.url)
+
+          // Don’t scroll to hash when it’s just the h1.
+          if (suggestion.isLvl0) {
+            url.hash = ""
+          }
+          window.location.href = url
```
jrf0110: `location.href` is supposed to be a string whereas `url` is an instance of `URL`. I'm assuming the browser is implicitly casting the `URL` to a string via the built in `toString()` method. In general, it's better to be explicit about this casting.

I suggest the following:

```suggestion
          window.location.href = url.toString()
```

gganch:
@pedrosousa @jrf0110 I just tested it locally and everything seems to be working perfectly. Love that there is a global search now, pretty cool! 

But on mobile it is a little hard to know that there is a search option 

Screen Shot 2021-06-17 at 5 31 52 PM

can we put the 'search docs...' or something similar to let users know that you can search??

Pedro would you mind updating the description of this PR to give a little more context what you implemented?

pedrosousa:
> But on mobile it is a little hard to know that there is a search option
> 
> Screen Shot 2021-06-17 at 5 31 52 PM
> 
> can we put the 'search docs...' or something similar to let users know that you can search??

Regarding the mobile version, I followed the current implementation for each product's documentation, which only displays the icon until you click it:

image
image

I'll see what I can do about this.

KimJ15:
@pedrosousa I think we can iterate and improve the mobile experience as a different PR given all search experiences would need to be updated.

pedrosousa:
Yes, I agree.
@gganch I'll create a separate ticket for addressing this improvement. It will affect all product documentation areas, since they all share the same UI issue you identified.

pedrosousa:
@jrf0110 Since GG is OOO, can you approve this PR?

jrf0110 (commented):
Sorry, @pedrosousa - I never submitted this review. I added some comments a while back and just let them sit in pending.

developers.cloudflare.com/docs-config.js
pedrosousa: Yes. By Algolia's own documentation, it can be made public.

developers.cloudflare.com/src/components/site-search.js
pedrosousa: Unfortunately, I don't have the details on this specific implementation. I followed what [the original `DocsSearch` component](https://github.com/cloudflare/cloudflare-docs-engine/blob/master/src/components/docs-search.js#L55) does.
Not sure if something changed in the meanwhile and we can have a different implementation now.

pedrosousa: Updated the code with a `useEffect` dependency. Had to use a separate `useEffect` (with an empty dependencies array) to load the `docsearch.js` script in a browser context, and define an `onload` handler that updates the state when the script is loaded.

gganch:
it looks pretty good to me! I am trying to approve but I cant it seems like...

developers.cloudflare.com/src/components/site-search.js
```diff
@@ -0,0 +1,178 @@
...
+      const input = search.input[0]
+      const wrapper = input.closest(".SiteSearch")
+
+      input.addEventListener("focus", () => {
+        wrapper.setAttribute("is-focused", "")
+      })
+
+      input.addEventListener("blur", () => {
+        wrapper.removeAttribute("is-focused")
+      })
```
jrf0110: Manually manipulating the DOM in this way is very _anti-react_. These sorts of things should be handled via state and attributes on the input element in question

pedrosousa: Updated — now using state for handling focus/blur events on the input element.

I don't think we can use this new approach for the autocomplete wrapper (`autocomplete:shown` and `autocomplete:closed` events), since that element is created dynamically by the Algolia script.

KimJ15: @jrf0110 @gganch is this something we can move forward? Had another stakeholder with a question about it :)

jrf0110 (approved):
Tested in Chrome, Firefox, and Edge. LGTM

KimJ15 (approved):

PR #1580, opened 2021-06-15
pages: adjust `redirects` phrasing

lukeed:
A customer was confused by some of the [Redirects](https://developers.cloudflare.com/pages/platform/redirects) documentation – specifically the limits. I noticed that the formatting demonstration wasn't rendering correctly:

Screen Shot 2021-06-15 at 3 23 39 PM

### DIFF
--- a/products/pages/src/content/platform/redirects.md
+++ b/products/pages/src/content/platform/redirects.md
@@ -3,9 +3,13 @@
 ## Creating redirects
 To use redirects on Cloudflare Pages, declare your redirects in a `_redirects` plain text file in the output folder of your project. The [build output folder](https://developers.cloudflare.com/pages/platform/build-configuration) is project-specific so the `_redirects` file should not always be in the root directory of the repo. Changes to redirects will be updated to your website at build time so make sure you commit and push the file to trigger a new build each time you update redirects.
 
-Redirects can be stated as such with one redirect per line in the form: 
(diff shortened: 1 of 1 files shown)


products/pages/src/content/platform/redirects.md
```diff
@@ -14,12 +18,12 @@ filename: _redirects
 /home301 / 301
 /home302 / 302
 /querystrings /?query=string 301
-/twitch https://twitch.tv
+/twitch https://twitch.tv 301
```
lukeed: We don't mention if omitting the `code` is allowed – and if it is, we don't say if it defaults to a `301` or `302` status code

kristianfreeman: we use this w/o status code [here](https://github.com/cloudflare/pages.cloudflare.com/blob/aaa012eaec97ae790e27a5db3e85ebe7d51ae9fa/static/_redirects#L1) in pages landing page - it defaults to a 302 according to my network tab

kristianfreeman: i think we should mention 302 is default and that the code is optional

lukeed: OK, if that's the case, then yes I agree

lukeed: I added a quick callout for this.

kristianfreeman (approved):

PR #1581, opened 2021-06-15
pages: add rollbacks page

lukeed:
/cc @nprogers @Albertozhao @signalnerve I'm not sure if you wanted to include info about pausing automatic deployments? If so, I'm not sure of how that works / how automatic deployments can be toggled on/off


products/pages/src/content/platform/rollbacks.md
```diff
@@ -0,0 +1,13 @@
+# Rollbacks
+
+Rollbacks allow you to instantly revert your project to a previous production deployment.
+
+Any production deployment that has been successfully built is a valid rollback target. When your project has rolled back to a previous deployment, you may still rollback to deployments that are _newer_ than your current version. Please note that preview deployments are not valid rollback targets.
```
kristianfreeman, line 5: > When your project has rolled back to a previous deployment, you may still rollback to deployments that are _newer_ than your current version.

products/pages/src/content/platform/rollbacks.md
```diff
@@ -0,0 +1,13 @@
...
+
+In order to perform a rollback, select a Pages project and ensure its Deployments tab is activated. Then browse the **All deployments** list and click on the actions menu for the desired target. A confirmation window will appear after clicking on the "Rollback to this deployment" option. When confirmed, your project's production deployment will change instantly.
```
kristianfreeman, line 7: is it instant or does it need to do a new deploy? that's neat if it is instant

kristianfreeman (approved):

products/pages/src/content/platform/rollbacks.md
lukeed: It's instant – or at least near-instant :D Testing it out showed that the rollback happened in the time it took me to refresh the page

rita3ko:
cc @nevikashah

nevikashah:
I think we should leave pausing automatic deployments out of the rollbacks section of docs. But I do think we could add a  how to pause deployments in the "Getting Started" section. I can create a separate issue for that though. Let me know your thoughts.

PR #1602, opened 2021-06-17
Waiting room interval -- time change

kodster28:


products/waiting-room/src/content/about/index.md
```diff
@@ -8,28 +8,31 @@ pcx-content-type: concept
...
 * Because waiting rooms support dynamic inflow and [outflow](/reference/configuration-settings#session-duration), new spots appear more quickly and estimated wait times are lower and more accurate.
 
 ## Availability
 
-The following Cloudflare customers have access to waiting room:
+The following customers have access to Cloudflare Waiting Room:
 - Those qualified under [Project Fair Shot](https://www.cloudflare.com/fair-shot/)
 - Business customers
 
+Access to certain features depends on a customer's [plan type](plans).
```
abracchi-tw, line 34: should this be a path pointing to another section in the dev docs? if so, it may be missing a slash in the beginning `[plan type](/plans)`

products/waiting-room/src/content/about/index.md
```diff
@@ -8,28 +8,31 @@ pcx-content-type: concept
 
 Cloudflare Waiting Room protects websites from surges in legitimate traffic that may otherwise bring an application down:
 - If a page is not experiencing heavy traffic, a visitor goes straight to the page.
```
abracchi-tw: a voice/tone question, but one that also touches the semantics of how we talk about certain processes within a product. I'm wondering about the use of the verb `go`, which at least in my mind entails a very physical motion. should we say `goes straight to the page` or should we go for something like `accesses the page directly`? Same a couple lines below,  should `a visitor goes to a virtual waiting room` be something like `a visitor enters a virtual waiting room`? I leave this 100% up to you, just something I was wondering about :)

abracchi-tw (approved):

products/waiting-room/src/content/about/index.md
kodster28: It resolves correctly since `plans` is a page within the `about` section. Confusing, I know. Only believe it b/c I previewed it locally.

PR #1610, opened 2021-06-18
[Teams] custom block page note

abracchi-tw:


abracchi-tw:
note is not necessary anymore, closing this out

PR #1682, opened 2021-06-29
WAF: Add notice about Managed Rulesets running at most once per request

pedrosousa:
Addresses https://jira.cfops.it/browse/PCX-1912.


rboulton-cf (changes requested):
This content is entirely accurate, but I would like to add a note to it that we are intending to change this behaviour, such that running the same ruleset multiple times will work as customers would expect.

pedrosousa:
@rboulton-cf Updated notice: rewrote first paragraph and added a new one at the end.

rboulton-cf (approved):

crwaters16 (approved):

PR #1685, opened 2021-06-29
addEventListener code snippet update

deadlypants1973:
'return' portion of code deleted


lukeed (approved):
It wasn't invalid, but ya – we omit the `return` everywhere else

PR #1699, opened 2021-06-30
Lb load shedding

kodster28:
Addresses PCX-1452.

Screenshot (for reviewing):
image


products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Load shedding
+
+Use load shedding to prevent an at-risk origin from becoming [unhealthy](/understand-basics/monitors) and starting the failover process. 
+
+Once you configure load shedding on a pool, that pool will begin diverting traffic to other pools according to your load shedding settings and the load balancer's [steering policy](/understand-basics/traffic-steering).
+
+## Step 1 — Identify at-risk origins
+
+Using your internal metrics, identify origins at risk of reaching their failure threshold:
```
dcpena: This is a nitpick, sorry, but I think you can end this with a period instead of a colon.

products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,133 @@
...
+A *Random* policy:
+
+- Randomly sheds the percentage of requests specified in the *Shed %*.
+- Distributes traffic more accurately because it sheds at the request level.
+- May cause requests from the same IP to hit different origins, potentially leading to cache misses, inconsistent latency, or session disruption for [DNS-only load balancers](/understand-basics/proxy-modes#dns-only-mode)).
+
+An *IP hash* policy:
+
+- Sheds the percentage of IP addresses specified in the *Shed %*.
+- Ensures requests from the same IP will hit the same origin, promoting cache hits, consistent latency, and session preservation.
```
dcpena: I had to read this line a few times because I didn't realize initially that promoting cache hits, consistent latency, and session preservation were describing the first half of the sentence. 

Maybe saying, "...will hit the same origin, which promote cache hits, consistent latency...".

dcpena (approved):
Approved with two, small non-blocking comments!

products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,134 @@
...
+
+<details>
+<summary>Example request</summary>
+<div>
+
+```json
+---
+header: Request
+---
+curl -X PUT "https://api.cloudflare.com/client/v4/accounts/{account-id}/load_balancers/pools/{pool-id}" \
```
gofish: This should be  `PATCH` request.

products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,134 @@
...
+     -H "X-Auth-Email: user@example.com" \
+     -H "X-Auth-Key: REDACTED" \
+     -H "Content-Type: application/json" \
+     --data '{
```
gofish: I think you want `--data-binary`.

products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Shedding policies
+
+For **Default traffic**, you have two choices for shedding policy.
+
+A *Random* policy:
+
+- Randomly sheds the percentage of requests specified in the *Shed %*.
+- Distributes traffic more accurately because it sheds at the request level.
+- May cause requests from the same IP to hit different origins, potentially leading to cache misses, inconsistent latency, or session disruption for [DNS-only load balancers](/understand-basics/proxy-modes#dns-only-mode)).
```
gofish: Extra close-paren at EOL (or missing open-paren).

products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,134 @@
...
+
+An *IP hash* policy:
+
+- Sheds the percentage of IP addresses specified in the *Shed %*.
```
gofish: Possibly more accurate:

> Sheds the percentage of **IP address hash space** specified in the *Shed %*

products/load-balancing/src/content/understand-basics/load-shedding.md
```diff
@@ -0,0 +1,134 @@
...
+- Ensures requests from the same IP will hit the same origin, which will increase cache hits, provide consistent latency, and preserve sessions.
+- Can over- or under-shed requests, since individual IPs may be responsible for different percentages of your requests. For example, setting a *Shed %* of *50%* may shed between **40%** to **60%** of pool requests.
```
gofish: The example here is potentially misleading: 50% could potentially shed less than 40% or more than 60% of pool requests, in extreme cases where traffic is heavily skewed. I don't think we should provide any raw numbers and suggest

> Can over- or under-shed requests, since hashing does not guarantee a perfectly even IP distribution and individual IPs may be responsible for different percentages of your requests.

gofish (changes requested):
Looks great! Love the details here.

products/load-balancing/src/content/understand-basics/load-shedding.md
gofish: But could also be more confusing.

products/load-balancing/src/content/understand-basics/load-shedding.md
kodster28: Howdy @gofish... sorry for the multiple edits, okay. Answered the question by doing some research on my own.

Thanks for giving me an opportunity to learn about `curl` today!

kodster28:
@gofish, should have addressed all your changes. Let me know if a) that didn't happen or b) you see anything else that needs changes.

gofish (approved):

PR #1704, opened 2021-07-01
Route specificity update to Workers docs

### LINKED ISSUE #1212: How to measure specificity in  route pattern
### Expected Behavior
https://github.com/cloudflare/cloudflare-docs/blob/production/products/workers/src/content/platform/routes.md#matching-behavior

Section _When more than one route pattern could match a request URL, the most specific route pattern wins. For example, the pattern www.example.com/* would take precedence over *.example.com/* when matching a request for https://www.example.com/._

Could we please clarify how specificity is calculated? If it is the number of wildcards and/or also the number of URL segments? Will `https://www.example.com/path1/*` win over `https://www.example.com/*`? It's a struggle that several workers cannot be applied in order, such as Java Servlets, and defining this better would really help. We can just guess by testing several times what are rules and what is random coincidences.

### Actual Behavior
Expect documentation to describe a deterministic behaviour for more scenarios.

The section about specificity under "Matching BEhaviour" as linked above.

### Additional information
-

deadlypants1973:
Closes https://github.com/cloudflare/cloudflare-docs/issues/1212

Needs rebase before merge due to #1703 

Screenshot 2021-07-01 at 21 28 06

### DIFF
--- a/products/workers/src/content/platform/known-issues.md
+++ b/products/workers/src/content/platform/known-issues.md
@@ -5,12 +5,14 @@ pcx-content-type: concept
 
 # Known issues
 
-Below are some known bugs and issues to be aware of when using Cloudflare Workers:
+Below are some known bugs and issues to be aware of when using Cloudflare Workers.
+
+## Route specificity 
 
 - When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
 
     1) `example.com/images/*`
 
-    2) example.com/images*
+    2) `example.com/images*`
 
-for requests like `example.com/images/hello` to route successfully, the route pattern path must be set to option 2 (`example.com/images*`) without the forward slash. If the route pattern path is set to option 1 (`example.com/images/*`) requests that include a suffix, like `example.com/images/world`, will not route as intended. 
+For requests like `example.com/images/hello` to route successfully, the route pattern path must be set to option 2 (`example.com/images*`) without the forward slash. If the route pattern path is set to option 1 (`example.com/images/*`) requests that include a suffix, like `example.com/images/world`, will not route as intended. 
--- a/products/workers/src/content/platform/known-issues.md
+++ b/products/workers/src/content/platform/known-issues.md
@@ -9,10 +9,20 @@ Below are some known bugs and issues to be aware of when using Cloudflare Worker
 
 ## Route specificity 
 
-- When defining route specificity, a trailing `/*` in your pattern may not act as expected:
+- When defining route specificity, a trailing `/*` in your pattern may not act as expected.
 
-    1) `example.com/images/*`
+Consider two different Workers, each deployed to the same zone. Worker A is assigned the `example.com/images/*` route and Worker B is given the `example.com/images*` route pattern. With these in place, here are how the following URLs will be resolved:
 
-    2) `example.com/images*`
+```
+// (A) example.com/images/*
+// (B) example.com/images*
 
-For requests like `example.com/images/hello` to route successfully, the route pattern path must be set to option 2 (`example.com/images*`) without the forward slash. If the route pattern path is set to option 1 (`example.com/images/*`) requests that include a suffix, like `example.com/images/world`, will not route as intended. 
+"example.com/images" 
+// -> B
+"example.com/images123" 
+// -> B
+"example.com/images/hello"
+// -> B
+```
+
+You will notice that all examples trigger Worker B. This includes the final example, which exemplifies the unexpected behavior.
--- a/products/workers/src/content/platform/known-issues.md
+++ b/products/workers/src/content/platform/known-issues.md
@@ -7,7 +7,8 @@ pcx-content-type: concept
 
 Below are some known bugs and issues to be aware of when using Cloudflare Workers:
 
-- When defining route specifcity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
+- When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
+
 
     1) `example.com/images/*`
 
--- a/products/workers/src/content/platform/known-issues.md
+++ b/products/workers/src/content/platform/known-issues.md
@@ -9,7 +9,7 @@ Below are some known bugs and issues to be aware of when using Cloudflare Worker
 
 ## Route specificity 
 
-- When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
+- When defining route specificity, a trailing `/*` in your pattern may not act as expected:
 
     1) `example.com/images/*`
 
--- a/products/workers/src/content/platform/known-issues.md
+++ b/products/workers/src/content/platform/known-issues.md
@@ -9,7 +9,6 @@ Below are some known bugs and issues to be aware of when using Cloudflare Worker
 
 - When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
 
-
     1) `example.com/images/*`
 
     2) example.com/images*
--- a/products/workers/src/content/platform/known-issues.md
+++ b/products/workers/src/content/platform/known-issues.md
@@ -5,7 +5,8 @@ pcx-content-type: concept
 
 # Known issues
 
- Below are some known bugs and issues to be aware of when using Cloudflare Workers:
+Below are some known bugs and issues to be aware of when using Cloudflare Workers:
+
 - When defining route specifcity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
 
     1) `example.com/images/*`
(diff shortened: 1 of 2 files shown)


products/workers/src/content/platform/routes.md
```diff
@@ -102,7 +102,7 @@ If a route pattern path ends with `*`, then it matches all suffixes of that path
 
 - `https://example.com/path*` matches `https://example.com/path` _and_ `https://example.com/path2` _and_ `https://example.com/path/readme.txt`
 
-- `https://example.com/path/*` matches `https://example.com/path/readme.txt` but _not_ `https://example.com/path2`.
+- There is a well known bug associated with path matching concerning wildcards and forward slashes that is documented in the [Known Issues section.](/workers/platform/known-issues)
```
deadlypants1973:
```suggestion
- There is a well-known bug associated with path matching concerning wildcards and forward slashes that is documented in the [Known issues section.](/workers/platform/known-issues)
```

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,17 @@
+---
+order: 5
+pcx-content-type: concept
+---
+
+# Known issues
+
+ Below are some known bugs and issues to be aware of when using Cloudflare Workers:
+- When defining route specifcity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
```
kristianfreeman:
```suggestion
- When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:

```

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,17 @@
+---
+order: 5
+pcx-content-type: concept
+---
+
+# Known issues
+
+ Below are some known bugs and issues to be aware of when using Cloudflare Workers:
```
kristianfreeman:
```suggestion
Below are some known bugs and issues to be aware of when using Cloudflare Workers:

```

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,17 @@
...
+- When defining route specifcity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
+
+    1) example.com/images/*
```
kristianfreeman:
```suggestion
    1) `example.com/images/*`
```

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,17 @@
...
+
+    2) example.com/images*
```
kristianfreeman:
```suggestion
    2) `example.com/images*`
```

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,17 @@
...
+
+for requests like `example.com/images/hello` to route successfully, the route pattern path must be set to option 2 (`example.com/images*`) without the forward slash. If the route pattern path is set to option 1 (`example.com/images/*`) requests that include a suffix, like `example.com/images/world`, will not route as intended.
```
kristianfreeman:
```suggestion
For requests like `example.com/images/hello` to route successfully, the route pattern path must be set to option 2 (`example.com/images*`) without the forward slash. If the route pattern path is set to option 1 (`example.com/images/*`) requests that include a suffix, like `example.com/images/world`, will not route as intended. 
```

products/workers/src/content/platform/routes.md
```diff
@@ -102,7 +102,7 @@ If a route pattern path ends with `*`, then it matches all suffixes of that path
 
 - `https://example.com/path*` matches `https://example.com/path` _and_ `https://example.com/path2` _and_ `https://example.com/path/readme.txt`
 
-- `https://example.com/path/*` matches `https://example.com/path/readme.txt` but _not_ `https://example.com/path2`.
+- There is a well-known bug associated with path matching concerning wildcards and forward slashes that is documented in the [Known issues section.](/workers/platform/known-issues)
```
kristianfreeman: Maybe put this in an Aside instead?

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,17 @@
...
+order: 5
+pcx-content-type: concept
+---
+
+# Known issues
+
+Below are some known bugs and issues to be aware of when using Cloudflare Workers:
+
+- When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
+
```
deadlypants1973, line 11:
```suggestion
```

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,18 @@
...
+## Route specificity 
+
+- When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
+
+    1) `example.com/images/*`
+
+    2) `example.com/images*`
+
+For requests like `example.com/images/hello` to route successfully, the route pattern path must be set to option 2 (`example.com/images*`) without the forward slash. If the route pattern path is set to option 1 (`example.com/images/*`) requests that include a suffix, like `example.com/images/world`, will not route as intended.
```
lukeed: I think "speaking it out" here isn't super effective at explaining what's going on TBH :S I think it'd be more effective if we just had a JS/pseudo-code snippet that showed input -> output & do our best to explain why (1) or (2) matched. Something roughly like:

---

Consider two different Workers, each deployed to the same zone. Worker A is assigned the `example.com/images/*` route and Worker B is given the `example.com/images*` route pattern. With these in place, here are how the following URLs will be resolved:

```js
// (A) example.com/images/*
// (B) example.com/images*

"example.com/images" 
// -> B
"example.com/images123" 
// -> B
"example.com/images/hello"
// -> B
```

You will notice that all examples trigger Worker B. This includes the final example, which exemplifies the unexpected behavior.

---

Looks like including a codeblock in the suggestion broke GitHub, so removing the suggestion part haha.

products/workers/src/content/platform/known-issues.md
```diff
@@ -0,0 +1,18 @@
...
+pcx-content-type: concept
+---
+
+# Known issues
+
+Below are some known bugs and issues to be aware of when using Cloudflare Workers.
+
+## Route specificity 
+
+- When defining route specificity, avoid the use of forward slashes before a wildcard to ensure accuracy when matching suffixes. For example, consider the below path examples:
```
deadlypants1973:
```suggestion
- When defining route specificity, a trailing `/*` in your pattern may not act as expected:
```

lukeed:
Rebased & force pushed for synchronization issue

dcpena (approved):

products/workers/src/content/platform/routes.md
```diff
@@ -50,13 +50,13 @@ While they look similar to a [regex](https://en.wikipedia.org/wiki/Regular_expre
 
 - Route patterns may not contain infix wildcards or query parameters, e.g. neither `example.com/*.jpg` nor `example.com/?foo=*` are valid route patterns.
 
-- When more than one route pattern could match a request URL, the most specific route pattern wins. For example, the pattern `www.example.com/*` would take precedence over `*.example.com/*` when matching a request for `https://www.example.com/`.
+- When more than one route pattern could match a request URL, the most specific route pattern wins. For example, the pattern `www.example.com/*` would take precedence over `*.example.com/*` when matching a request for `https://www.example.com/`. The pattern `example.com/hello/*` would take precedence over `example.com/*` when matching a request for `example.com/hello/world`.
```
kristianfreeman, line 53: correct me if i'm wrong but i think that this actually is an example of what _wouldn't_ happen given the route specificity bug? i think that the more specific route _wouldn't_ match unless we changed it to `/hello*` instead of `/hello/*` -- cc @lukeed to double-check my work here

products/workers/src/content/platform/routes.md
```diff
@@ -99,7 +99,11 @@ If a route pattern path ends with `*`, then it matches all suffixes of that path
 
 - `https://example.com/path*` matches `https://example.com/path` _and_ `https://example.com/path2` _and_ `https://example.com/path/readme.txt`
 
-- `https://example.com/path/*` matches `https://example.com/path/readme.txt` but _not_ `https://example.com/path2`.
+<Aside>
```
kristianfreeman:
```suggestion
<Aside type="warning">
```

products/workers/src/content/platform/routes.md
lukeed: In this case, the changes are correct (verified to be sure).
I think it's because the `/hello/*` has an extra layer of static detail that it's able to make that resolution. 

However, if it were `/he*` vs `/hello/*`, then `"/hello/world"` would match `/he*`
Similarly, with `/hello*` vs `/hello/*`, then `"/hello/world"` would match `/hello*`

deadlypants1973: I think the reason I wrote it this way is that given the length specificity, this is all correct normal behavior. But then right below that, is the aside warning that there is a bug, specifically related to forward slashes and wildcards, that people should refer to before proceeding to write paths. Are we okay with proceeding on this PR? If there are no other objections, I will merge. But if we need to reword, can do that too :) @signalnerve @lukeed

PR #1707, opened 2021-07-01
Add API instructions for configuring payload logging

pedrosousa:
Addresses https://jira.cfops.it/browse/PCX-1797.

### DIFF
--- a/products/waf/src/content/managed-rulesets/payload-logging/configure-api.md
+++ b/products/waf/src/content/managed-rulesets/payload-logging/configure-api.md
@@ -10,7 +10,7 @@ You can use the [Rulesets API](https://api.cloudflare.com/) to configure payload
 
 ## Configure and enable payload logging
 
-Do the following:
+To configure:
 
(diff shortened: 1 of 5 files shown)


migueldemoura (approved):

products/waf/src/content/managed-rulesets/payload-logging/configure-api.md
```diff
@@ -0,0 +1,128 @@
...
+order: 3
+---
+
+# Configure payload logging for a Managed Ruleset via API
+
+You can use the [Rulesets API](https://api.cloudflare.com/) to configure payload logging for a Managed Ruleset.
+
+## Configure and enable payload logging
+
+Do the following:
```
nayaabahsan: This might just be personal preference so ignore if you don't agree, but this sounds a bit awkward to me. Although I understand why it would be repetitive to say "To configure payload logging..." when it's been said twice above. Possibly "To configure:" here instead?

nayaabahsan (approved):

products/waf/src/content/managed-rulesets/payload-logging/configure-api.md
pedrosousa: Makes sense! Thanks

PR #1708, opened 2021-07-01
WAF Docs: release 2021-07-01

ggaborCF:


products/waf/src/content/change-log/2021-07-01---emergency-release.md
```diff
@@ -0,0 +1,65 @@
+---
+title: "2021-07-01 – Emergency"
+type: table
+order: 941
```
umahasit: Should be 937

umahasit (approved):

PR #1711, opened 2021-07-02
improvements for Workers example — HTTP "Basic" Authentication

drhino:
- Better validation when the Authorization header does not contain a space.

- Normalizes the decoded string once. Both the user and password must be normalized.
-- Strictly define NFC encoding as per RFC: 3.3.2.4 & 4.2.2.4 (even though it's the default):
   "Normalization Rule: Unicode Normalization Form C (NFC) MUST be applied to all characters."

- Figured out what control characters are: https://tools.ietf.org/html/rfc5234#appendix-B.1
Left column and bottom right value of (Hx column):
-- http://www.asciitable.com/index/asciifull.gif
-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes
    ( "\0" equals "\x00" )

( try catch fails when decoded.match(...) does not return an array )

### DIFF
--- a/products/workers/src/content/examples/basic-auth.md
+++ b/products/workers/src/content/examples/basic-auth.md
@@ -112,20 +112,29 @@ function basicAuthentication(request) {
   const [scheme, encoded] = Authorization.split(' ')
 
   // The Authorization header must start with "Basic", followed by a space.
-  if (!encoded || scheme !== 'Basic')
+  if (!encoded || scheme !== 'Basic') {
     throw new BadRequestException('Malformed authorization header.')
+  }
 
   // Decodes the base64 value and performs unicode normalization.
   // @see https://datatracker.ietf.org/doc/html/rfc7613#section-3.3.2 (and #section-4.2.2)
   // @see https://dev.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
-  const decoded = atob(encoded).normalize('NFC')
+  const decoded = atob(encoded).normalize()
+  
+  // The username & password are split by the first colon.
+  //=> example: "username:password"
+  const index = decoded.indexOf(':')
 
   // The user & password are split by the first colon and MUST NOT contain control characters.
   // @see https://tools.ietf.org/html/rfc5234#appendix-B.1 (=> "CTL = %x00-1F / %x7F")
-  try {
-    const [matched, user, pass] = decoded.match(/^([^\0-\x1F\x7F:]*):([^\0-\x1F\x7F]*)$/)
-    return { user, pass }
-  } catch { throw new BadRequestException('Invalid decoded "user:password" value.') }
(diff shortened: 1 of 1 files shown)


products/workers/src/content/examples/basic-auth.md
```diff
@@ -114,24 +111,21 @@ function basicAuthentication(request) {
...
-    pass: decoded.substring(seperatorPosition + 1).normalize(),
-  }
+  const decoded = atob(encoded).normalize('NFC')
+
+  // The user & password are split by the first colon and MUST NOT contain control characters.
+  // @see https://tools.ietf.org/html/rfc5234#appendix-B.1 (=> "CTL = %x00-1F / %x7F")
+  try {
+    const [matched, user, pass] = decoded.match(/^([^\0-\x1F\x7F:]*):([^\0-\x1F\x7F]*)$/)
+    return { user, pass }
+  } catch { throw new BadRequestException('Invalid decoded "user:password" value.') }
```
lukeed: - NFC is the default normalization. Best to hide it since it's not needed & may cause more confusion.
- Testing for control characters first makes for easier split

```suggestion
  const decoded = atob(encoded).normalize()
  
  // The username & password are split by the first colon.
  //=> example: "username:password"
  const index = decoded.indexOf(':')

  // The user & password are split by the first colon and MUST NOT contain control characters.
  // @see https://tools.ietf.org/html/rfc5234#appendix-B.1 (=> "CTL = %x00-1F / %x7F")
  if (index === -1 || /[\0-\x1F\x7F]/.test(decoded)) {
    throw new BadRequestException('Invalid authorization value.')
  }
  
  return { 
    user: decoded.substring(0, index),
    pass: decoded.substring(index + 1),
  }
```

products/workers/src/content/examples/basic-auth.md
```diff
@@ -114,24 +111,21 @@ function basicAuthentication(request) {
...
   const [scheme, encoded] = Authorization.split(' ')
 
-  // The Authorization header must look like "Basic user:encoded".
-  if (scheme !== 'Basic') throw new BadRequestException('Malformed authorization header.')
-
-  // Decode the base64 value.
-  const decoded = atob(encoded)
+  // The Authorization header must start with "Basic", followed by a space.
+  if (!encoded || scheme !== 'Basic')
+    throw new BadRequestException('Malformed authorization header.')
```
lukeed:
```suggestion
  if (!encoded || scheme !== 'Basic') {
    throw new BadRequestException('Malformed authorization header.')
  }
```

lukeed (approved):

kristianfreeman:
@lukeed you're good to merge if you're +1 on this!

lukeed:
Thanks @drhino~!

drhino:
Very clever changing the regex test. Much better.

PR #1781, opened 2021-07-14
[Workers] Headers class documentation

### LINKED ISSUE #17: Document the Headers class
The Headers class is pretty straightforward, but its behavior is the source of some frequently asked questions. We should document its behavior in at least these regards:

- The comma-concatenation behavior of Headers.get(). If a Headers object is constructed from a message with multiple instances of the same header, Headers.get() will return a single string (not an array) with each of the multiple header values concatenated together with intervening `", "`.
- The comma-concatenation described above is a valid transformation for any well-behaved header, with one infamous exception: Set-Cookie. For this reason, we implement the old, removed `Headers.getAll()` function, but only for the Set-Cookie header (it throws if you try to access anything else). You can read the saga behind this here: https://github.com/whatwg/fetch/issues/973
- The Cloudflare Workers implementation of Headers.get() returns a USVString (or null, of course), not a ByteString as required by the spec. In English, this means we interpret the header bytes on the wire as UTF-8, decode them, and expose them as Unicode strings in JavaScript. We once tried to "fix" this by switching to ByteStrings, but ByteStrings are difficult to use correctly, and in practice it is common to set a header from the contents of URLSearchParams.get(), which returns a USVString, and would thus produce a corrupt header.
- Related to the above, we warn in the preview if a script _does_ try to add a ByteString to a Headers object. Example: https://cloudflareworkers.com/#2b9b84913d7147de5592b6306404e128:https://tutorial.cloudflareworkers.com

/cc @ispivey

deadlypants1973:
Closes https://github.com/cloudflare/cloudflare-docs/issues/17

### DIFF
--- a/products/workers/src/content/runtime-apis/headers.md
+++ b/products/workers/src/content/runtime-apis/headers.md
@@ -1,0 +1,33 @@
+---
+pcx-content-type: configuration
+---
+
+# Headers
+
+## Background
+
+All HTTP request and response headers are available through the [Headers API](https://developer.mozilla.org/en-US/docs/Web/API/Headers).
+
+When a header name possesses multiple values, those values will be concatenated as a single, comma-delimited string value. This means that `Headers.get` will always return a string or a `null` value. This applies to all header names except for `Set-Cookie`, which requires `Headers.getAll`. This is documented below in Differences. 
+
+```js
+let headers = new Headers;
+
+headers.get('x-foo'); //=> null
+
+headers.set('x-foo', '123');
+headers.get('x-foo'); //=> "123"
+
+headers.set('x-foo', 'hello');
+headers.get('x-foo'); //=> "hello"
+
+headers.append('x-foo', 'world');
+headers.get('x-foo'); //=> "hello, world"
+```
+
+## Differences
+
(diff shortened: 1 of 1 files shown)


lukeed (approved):

products/workers/src/content/runtime-apis/headers.md
```diff
@@ -0,0 +1,33 @@
...
+pcx-content-type: configuration
+---
+
+# Headers
+
+## Background
+
+All HTTP request and response headers are available through the [Headers API](https://developer.mozilla.org/en-US/docs/Web/API/Headers).
+
+When a header name possesses multiple values, those values will be concatenated as a single, comma-delimited string value. This means that `Headers.get` will always return a string or a `null` value. This applies to all header names except for `Set-Cookie`, which requires `Headers.getAll`. This is documented below in Differences.
```
abracchi-tw: I know it's literally right below but I usually hyperlink to sections when I mention them elsewhere in the page
 
```suggestion
When a header name possesses multiple values, those values will be concatenated as a single, comma-delimited string value. This means that `Headers.get` will always return a string or a `null` value. This applies to all header names except for `Set-Cookie`, which requires `Headers.getAll`. This is documented below in [Differences](#differences). 
```

abracchi-tw (approved):

PR #1795, opened 2021-07-16
Fix a routing issue in create-react-app guide

treboryx:
`<Post path="/:id" />` => `<Post path="/posts/:id" />` otherwise it doesn't work


kristianfreeman (approved):
thanks!

PR #1800, opened 2021-07-16
Added alternative to Sentry

toddhgardner:
TrackJS is an alternative that is more focused on front-end concerns.


kristianfreeman:
hey @toddhgardner - TrackJS sounds neat! Sentry is pretty ubiquitous so we're going to keep this as-is but as best of luck on the project: if you make a Workers integration, let us know!

toddhgardner:
@signalnerve. Thank you, but this isn't a new project. We've been around just as long as Sentry, but we're more niche. We have customers using us on your workers today, who asked why we weren't referenced here.

PR #1815, opened 2021-07-19
Wr small clarifications

kodster28:
Addresses PCX-1965

### DIFF
--- a/products/waiting-room/src/content/about/index.md
+++ b/products/waiting-room/src/content/about/index.md
@@ -18,7 +18,7 @@ Cloudflare Waiting Room benefits:
     - Showing estimated wait times that are continuously updated.
     - Opening up new spots more quickly by tracking dynamic inflow and [outflow](/reference/configuration-settings#session-duration).
     - Remembering each visitor's status to prevent someone from losing their place in line or having to re-queue if they leave your site.
-    - Appearing in your own [branding and style](../how-to/customize-waiting-room), which enhances trust and lets you provide more information as needed.
+    - Appearing in your own [branding and style](../how-to/customize-waiting-room), which enhances trust and lets you provide additional information as needed.
 
 ## How it works
 
--- a/products/waiting-room/src/content/about/index.md
+++ b/products/waiting-room/src/content/about/index.md
@@ -6,24 +6,32 @@ pcx-content-type: concept
 
 # About Cloudflare Waiting Room
 
-Cloudflare Waiting Room protects websites from surges in legitimate traffic that may otherwise bring an application down:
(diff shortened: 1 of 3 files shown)


products/waiting-room/src/content/about/index.md
```diff
@@ -6,25 +6,33 @@ pcx-content-type: concept
...
 
-Cloudflare Waiting Room protects websites from surges in legitimate traffic that may otherwise bring an application down:
+Cloudflare Waiting Room protects websites from surges in legitimate traffic that may otherwise bring an application down.
+
+## Benefits
+
+Cloudflare Waiting Room benefits:
+
+- **Your origin server** by preventing surges in legitimate traffic that may overload your origin.
+- **Your visitors** by:
```
dcpena: I think the bulleted list for this section makes sense, but I'd consider revising the first half because typically bulleted lists are used for lists longer than two points. Maybe make the intro text and first bullet one sentence but then keep the second set about visitor benefits.

products/waiting-room/src/content/about/index.md
```diff
@@ -6,25 +6,33 @@ pcx-content-type: concept
...
+    - Keeping your application online and preventing them from reaching error pages.
+    - Showing estimated wait times that are continuously updated.
+    - Opening up new spots more quickly by tracking dynamic inflow and [outflow](/reference/configuration-settings#session-duration).
+    - Remembering each visitor's status to prevent someone from losing their place in line or having to re-queue if they leave your site.
+    - Appearing in your own [branding and style](../how-to/customize-waiting-room), which enhances trust and lets you provide additional information as needed.
+
+## How it works
+
+Once you have [created and activated](get-started) a waiting room for a specific application page:
```
dcpena: I'll let you choose, but I think it could be helpful to include "a waiting room" as part of the link to text to make the action stand out a little better because the focus is on creating and activating the waiting room and not just creating and activating if that makes sense.

products/waiting-room/src/content/get-started/index.md
```diff
@@ -26,28 +26,58 @@ Which page will you cover with a waiting room? You can only have one waiting roo
...
-### Access
+### Access method
 
 Do you want visitors to reach your high-traffic page:
 - Directly (via URL)
 - Indirectly (via a [Page Rule redirect](https://support.cloudflare.com/hc/articles/200172286))
 
+### Queue method
+
+When you [activate your waiting room](#step-3--activate-your-waiting-room), do you want:
```
dcpena: Because the second half starts "do you want," I expected this to end with a question. Maybe reword so that you're not posing a question but presenting options. Maybe something like, "After you activate your waiting room, choose whether:"

dcpena (approved):
Approved with non-blocking suggestions

PR #1819, opened 2021-07-19
Docs for Durable Objects tagged migrations/migrations in wrangler.toml

xortive:
The following steps need to be checked off before this should be merged:
- [ ] The corresponding [wrangler PR](https://github.com/cloudflare/wrangler/pull/1992) needs to be merged and released
- [ ] Our DO templates should all be updated to use migrations in wrangler.toml

### DIFF
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -230,7 +230,7 @@ In the above example, we used a string-derived object ID by calling the `idFromN
 
 <Aside type="warning" header="Custom Wrangler installation instructions">
 
-You must use [Wrangler version 1.17 or greater](https://developers.cloudflare.com/workers/cli-wrangler/install-update) in order to manage Durable Objects.
+You must use [Wrangler version 1.19 or greater](https://developers.cloudflare.com/workers/cli-wrangler/install-update) in order to manage Durable Objects.
 
 </Aside>
 
@@ -245,11 +245,7 @@ This will create a directory for your project with basic configuration and a sin
 * [Durable Objects Rollup ES Modules template](https://github.com/cloudflare/durable-objects-rollup-esm)
 * [Durable Objects Webpack CommonJS template](https://github.com/cloudflare/durable-objects-webpack-commonjs)
 
-The following sections will cover how to customize the configuration, but if you'd like you can immediately publish the generated project using this command:
-
-```sh
-$ wrangler publish --new-class Counter
-```
+The following sections will cover how to customize the configuration, but if you'd like you can immediately publish the generated project using `wrangler publish`.
 
 ### Specifying the main module
 
@@ -265,39 +261,72 @@ bindings = [
   { name = "EXAMPLE_CLASS", class_name = "DurableObjectExample" } # Binding to our DurableObjectExample class
 ]
 ```
+
 The `[durable_objects]` section has 1 subsection:
 
 - `bindings` - An array of tables, each table can contain the below fields.
   - `name` - Required, The binding name to use within your worker.
   - `class_name` - Required, The class name you wish to bind to.
-  - `script_name` - Optional, Defaults to the current project's script.
+  - `script_name` - Optional, Defaults to the current environment's script.
 
-### Publishing Durable Object classes
+### Durable Object migrations
 
-Normally when you want to publish a Worker using Wrangler, you just run `wrangler publish`. However, when you export a new Durable Objects class from your script, you must tell the Workers platform about it
-before you can create and access Durable Objects associated with that class. This process is called a "migration", and is currently performed by providing extra options to `wrangler publish`.
+When you export a new Durable Objects class from your script, you must tell the Workers platform about it before you can create and access Durable Objects associated with that class from a Durable Object binding. This process is called a "migration".
 
-To allow creation of Durable Objects associated with an exported class, specify `--new-class`:
-
-```sh
-$ wrangler publish --new-class DurableObjectExample
-```
-
-Note that after you've run `--new-class` for a given class name once, you do not need to include the migration on subsequent uploads of the Worker. You'd just run `wrangler publish` with no additional flags.
-
-If you want to delete the Durable Objects associated with an exported class, remove the corresponding binding from wrangler.toml, then use `--delete-class`:
-
-```sh
-$ wrangler publish --delete-class DurableObjectExample
-```
+The migration mechanism is also used to:
+* Transfer stored objects between two Durable Object classes in the same script, called a rename migration
+* Transfer stored objects between two Durable Object classes in different scripts, called a transfer migration.
+* Delete stored objects for a given Durable Object class
 
 <Aside type="warning" header="Important">
 
-Running a `--delete-class` migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
+Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
+
+A similar warning applies to rename and transfer migrations -- make sure you are not relying on the source Durable Object classes before applying a rename or transfer migration.
 
 </Aside>
 
-These are basic examples -- you can use multiple of these options in a single `wrangler publish` call if you'd like, one class per option. Future versions of Wrangler will also include migration directives for renaming a class or transferring a class from one file to another.
+Migrations can be performed in two different ways: An "ad hoc" migration using extra options on `wrangler publish`, or using the migration list in `wrangler.toml`. Migrations specified in `wrangler.toml` have the additional requirement of a "migration tag", which is used to determine
+which migrations have already been applied, and therefore can be skipped for a given script upload. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
+
+The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments, and cannot be overridden by a specific environment. Like "ad hoc" migrations, they are applied when you run `wrangler publish` on the selected environment's script. For our `DurableObjectExample` class, we need a migration list like this:
+
+```toml
+[[migrations]]
+tag = "v1" # should be unique for each entry
+new_classes = ["DurableObjectExample"] # array of new classes
+```
+
+Each migration in the list can have multiple directives, and multiple migrations can be specified as your project grows in complexity.
+
+```toml
+[[migrations]]
+tag = "v1" # should be unique for each entry
+new_classes = ["DurableObjectExample"] # array of new classes
+deleted_classes = ["DeprecatedClass"] # array of deleted classes
+renamed_classes = [{from: "OldClass", to: "NewClass"}] # array of rename directives
+transferred_classes = [{from_script: "old_script", from: "ExampleClass", to: "ExampleClass"}] # array of transfer directives
+```
+
+Ad hoc migrations are performed using extra options on `wrangler publish`. Several options can be passed at once in a single `wrangler publish` call. You can also specify `--old-tag <tag>` and `--new-tag <tag>` if your script has a migration tag associated with it but you still
+want to do an ad hoc migration.
+
+The list of ad hoc migration options that can be added to `wrangler publish` is as follows:
+
+```sh
+--old-tag <tag name> # old-tag is optional if your script doesn't have a migration tag set yet
+--new-tag <tag name> # new-tag and old-tag are optional if you only use ad hoc migrations.
+
+# Each of the migration directives can be specified multiple times if you are
+# creating/deleting/renaming/transferring multiple classes at once.
+--new-class <class name>
+--delete-class <class name>
+--rename-class <from class> <to class>
+--transfer-class <from script> <from class> <to class>
+```
+
+
+### Test your Durable Objects project
 
 At this point, we're done! If you copy the `DurableObjectExample` and fetch handler code from above into a generated Wrangler project, publish it using a `--new-class` migration, and make a request to it, you'll see that your request was stored in a Durable Object:
 
@@ -437,7 +466,7 @@ While using Wrangler is strongly recommended, if you would really rather not use
 
 ## Troubleshooting
 
-### Debugging 
+### Debugging
 `wrangler dev` does not currently support Durable Objects.
 
 To help with debugging, you may use [`wrangler tail`](/cli-wrangler/commands#tail) to troubleshoot your Durable Object script. `wrangler tail` displays a live feed of console and exception logs for each request your Worker receives. After doing a `wrangler publish`, you can use `wrangler tail` in the root directory of your Worker project and visit your Worker URL to see console and error logs in your terminal.
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -333,9 +333,15 @@ in the surrounding inline array are acceptable.
 
 ### Durable Object migrations through Wrangler CLI
 
-Ad hoc migrations are performed using extra options on `wrangler publish`. Several options can be passed at once in a single `wrangler publish` call. You can also specify `--old-tag <tag>` and `--new-tag <tag>` if your script has a migration tag associated with it but you still want to do an ad hoc migration.
+<Aside type="warning" header="Deprecation Notice">
+    
+While CLI migrations initially served a way to quickly migrate Durable Objects, this method is now deprecated and will be removed in a future release. This is because this approach makes it easy to forget or lose track of previous migrations, thus making it very difficult to mentally track the current infrastructure state and/or coordinate with teammates.
 
-When an ad hoc migration is performed, any migrations listed in `wrangler.toml` are ignored.
+</Aside>
+
+It is possible to define a migration purely through extra arguments to the `wrangler publish` command. When taking this route, any migrations listed in the `wrangler.toml` configuration file are ignored. 
+    
+You should provide an `--old-tag` value whenever possible. This value should be the name of the migration tag that you believe to be most recently active. Your `wrangler publish` command will throw an error if your `--old-tag` expectation does not align with Cloudflare's value. 
 
 The list of CLI migration arguments that can be added to `wrangler publish` is as follows:
 
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -281,7 +281,7 @@ The destination class (the class that stored objects are being transferred to) f
 
 <Aside type="warning" header="Important">
 
-After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class.
+After a rename or transfer migration, requests to the destination Durable Object class will have access to the source Durable Object's stored data. 
 
 After a migration, any existing bindings to the original Durable Object class (e.g., from other Workers) will automatically forward to the updated destination class. However, any Worker scripts bound to the updated Durable Object class must update their `[durable_objects]` configuration in the `wrangler.toml` file for their next deployment.
 
@@ -331,17 +331,17 @@ in the surrounding inline array are acceptable.
 
 </Aside>
 
-### Ad hoc Durable Object migrations
+### Durable Object migrations through Wrangler CLI
 
 Ad hoc migrations are performed using extra options on `wrangler publish`. Several options can be passed at once in a single `wrangler publish` call. You can also specify `--old-tag <tag>` and `--new-tag <tag>` if your script has a migration tag associated with it but you still want to do an ad hoc migration.
 
 When an ad hoc migration is performed, any migrations listed in `wrangler.toml` are ignored.
 
-The list of ad hoc migration options that can be added to `wrangler publish` is as follows:
+The list of CLI migration arguments that can be added to `wrangler publish` is as follows:
 
 ```sh
---old-tag <tag name> # old-tag is optional if your script doesn't have a migration tag set yet
---new-tag <tag name> # new-tag and old-tag are optional if you only use ad hoc migrations.
+--old-tag <tag name> # Optional if your script does not have a migration tag set yet.
+--new-tag <tag name> # new-tag and old-tag are optional if you only use CLI migrations.
 
 # Each of the migration directives can be specified multiple times if you are
 # creating/deleting/renaming/transferring multiple classes at once.
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -271,51 +271,57 @@ The `[durable_objects]` section has 1 subsection:
 
 ### Configuring Durable Object classes with migrations
 
-When you export a new Durable Objects class from your script, you must tell the Workers platform about it before you can create and access Durable Objects associated with that class from a Durable Object binding. This process is called a "migration".
+When you make changes to your list of Durable Objects classes, you must initiate a migration process. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
 
 Migrations can also be used for transferring stored data between two Durable Object classes:
 * Rename migrations are used to transfer stored objects between two Durable Object classes in the same script.
 * Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts.
 
-The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
+The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the deployed script.
 
 <Aside type="warning" header="Important">
 
 After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class.
 
-Existing bindings to the source Durable Object class will continue to work, now pointing to the destination Durable Object class. Future script uploads must update the class name for the binding in the `[durable_objects]` section of `wrangler.toml`.
+After a migration, any existing bindings to the original Durable Object class (e.g., from other Workers) will automatically forward to the updated destination class. However, any Worker scripts bound to the updated Durable Object class must update their `[durable_objects]` configuration in the `wrangler.toml` file for their next deployment.
 
 </Aside>
 
-Migrations can also be used to delete a given Durable Object class along with its stored objects.
+Migrations can also be used to delete a Durable Object class and its stored objects.
 
 <Aside type="warning" header="Important">
 
-Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Do not run a delete migration on a class without first ensuring that you are not relying on the Durable Objects within that class anymore. Copy any important data to some other location before deleting. 
+Running a delete migration will delete all Durable Object instances associated with the deleted class, including all of their stored data. Do not run a delete migration on a class without first ensuring that you are not relying on the Durable Objects within that class anymore. Copy any important data to some other location before deleting. 
 
 </Aside>
 
-Migrations can be performed in two different ways: by listing them in `wrangler.toml` or by manually providing ad hoc migrations using extra options on `wrangler publish`. Migrations specified in `wrangler.toml` have the additional requirement of a **migration tag**, which is used to determine which migrations have already been applied, and therefore can be skipped for a given script upload. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
+Migrations can be performed in two different ways: through the `[[migrations]]` configurations key in your `wrangler.toml` file or through additional CLI arguments during a `wrangler publish` command. Migrations specified in `wrangler.toml` have the additional requirement of a migration tag, which is defined by the **tag** property in each migration entry. Migration tags are treated like unique names and are used to determine which migrations have already been applied. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
 
 ### Durable Object migrations in `wrangler.toml`
 
-The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments, and cannot be overridden by a specific environment. Like ad hoc migrations, they are applied when you run `wrangler publish` on the selected environment's script. For our `DurableObjectExample` class, you need a migration list like this:
+The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments and cannot be overridden by a specific environment. 
+
+All migrations are applied at deployment. This is true for all migrations, whether initiated through `wrangler.toml` configurations or `wrangler publish` command arguments. Each migration can only be applied once per [environment](/platform/environments). 
+
+To illustrate an example migrations workflow, the `DurableObjectExample` class can be initially defined with:
 
 ```toml
 [[migrations]]
-tag = "v1" # should be unique for each entry
-new_classes = ["DurableObjectExample"] # array of new classes
+tag = "v1" # Should be unique for each entry
+new_classes = ["DurableObjectExample"] # Array of new classes
 ```
 
-Each migration in the list can have multiple directives, and multiple migrations can be specified as your project grows in complexity.
+Each migration in the list can have multiple directives, and multiple migrations can be specified as your project grows in complexity. For example, you may want to rename the `DurableObjectExample` class to `UpdatedName` and delete an outdated `DeprecatedClass` entirely.
 
 ```toml
 [[migrations]]
-tag = "v1" # should be unique for each entry
-new_classes = ["DurableObjectExample"] # array of new classes
-deleted_classes = ["DeprecatedClass"] # array of deleted classes
-renamed_classes = [{from: "OldClass", to: "NewClass"}] # array of rename directives
-transferred_classes = [{from_script: "old_script", from: "ExampleClass", to: "ExampleClass"}] # array of transfer directives
+tag = "v1" # Should be unique for each entry
+new_classes = ["DurableObjectExample"] # Array of new classes
+
+[[migrations]]
+tag = "v2"
+renamed_classes = [{from: "DurableObjectExample", to: "UpdatedName" }] # Array of rename directives
+deleted_classes = ["DeprecatedClass"] # Array of deleted class names
 ```
 
 <Aside type="note">
--- a/products/workers/src/content/learning/using-durable-objects.md
+++ b/products/workers/src/content/learning/using-durable-objects.md
@@ -245,7 +245,7 @@ This will create a directory for your project with basic configuration and a sin
 * [Durable Objects Rollup ES Modules template](https://github.com/cloudflare/durable-objects-rollup-esm)
 * [Durable Objects Webpack CommonJS template](https://github.com/cloudflare/durable-objects-webpack-commonjs)
 
-The following sections will cover how to customize the configuration, but if you'd like you can immediately publish the generated project using `wrangler publish`.
+The following sections will cover how to customize the configuration, but you can also immediately publish the generated project using `wrangler publish`.
 
 ### Specifying the main module
 
@@ -265,7 +265,7 @@ bindings = [
 The `[durable_objects]` section has 1 subsection:
 
 - `bindings` - An array of tables, each table can contain the below fields.
-  - `name` - Required, The binding name to use within your worker.
+  - `name` - Required, The binding name to use within your Worker.
   - `class_name` - Required, The class name you wish to bind to.
   - `script_name` - Optional, Defaults to the current environment's script.
 
@@ -274,10 +274,10 @@ The `[durable_objects]` section has 1 subsection:
 When you export a new Durable Objects class from your script, you must tell the Workers platform about it before you can create and access Durable Objects associated with that class from a Durable Object binding. This process is called a "migration".
 
 Migrations can also be used for transferring stored data between two Durable Object classes:
-* Rename migrations are used to transfer stored objects between two Durable Object classes in the same script
-* Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts
+* Rename migrations are used to transfer stored objects between two Durable Object classes in the same script.
+* Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts.
 
-The destination class (the one that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
+The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
 
 <Aside type="warning" header="Important">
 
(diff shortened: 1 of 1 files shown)


products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,72 @@ bindings = [
...
-```sh
-$ wrangler publish --new-class DurableObjectExample
-```
+<Aside type="warning" header="Important">
 
-Note that after you've run `--new-class` for a given class name once, you do not need to include the migration on subsequent uploads of the Worker. You'd just run `wrangler publish` with no additional flags.
+Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
 
-If you want to delete the Durable Objects associated with an exported class, remove the corresponding binding from wrangler.toml, then use `--delete-class`:
+A similar warning applies to rename and transfer migrations -- make sure you are not relying on the source Durable Object classes before applying a rename or transfer migration.
```
bretthoerner: I get what this is saying, but if we added something along the lines of "After the migration, requests to Durable Objects will be served by the new class, existing stored data will remain inside each object." we might avoid some minor confusion.

bretthoerner (approved):

products/workers/src/content/learning/using-durable-objects.md
a-robinson: Yeah, this would be best as a separate warning that isn't called out as "similar", since it's a very different concern.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,72 @@ bindings = [
...
 
-```sh
-$ wrangler publish --delete-class DurableObjectExample
+</Aside>
+
+Migrations can be performed in two different ways: An "ad hoc" migration using extra options on `wrangler publish`, or using the migration list in `wrangler.toml`. Migrations specified in `wrangler.toml` have the additional requirement of a "migration tag", which is used to determine
```
a-robinson: Slight rewording for clarity:

> Migrations can be performed in two different ways: by listing them in wrangler.toml or by manually providing "ad hoc" migrations using extra options on `wrangler publish`.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,72 @@ bindings = [
...
+which migrations have already been applied, and therefore can be skipped for a given script upload. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
+
+The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments, and cannot be overridden by a specific environment. Like "ad hoc" migrations, they are applied when you run `wrangler publish` on the selected environment's script. For our `DurableObjectExample` class, we need a migration list like this:
```
a-robinson: Should we have separate subsections for wrangler.toml migrations vs ad hoc migrations? It seems pretty separable and then each could be deep linked.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,72 @@ bindings = [
...
 The `[durable_objects]` section has 1 subsection:
 
 - `bindings` - An array of tables, each table can contain the below fields.
   - `name` - Required, The binding name to use within your worker.
   - `class_name` - Required, The class name you wish to bind to.
-  - `script_name` - Optional, Defaults to the current project's script.
+  - `script_name` - Optional, Defaults to the current environment's script.
 
-### Publishing Durable Object classes
+### Durable Object migrations
```
a-robinson: Minor, but it'd still be nice to have a section called something like "Publishing Durable Object classes" since it's likely what someone looking at the table of contents to figure out the most important section will choose. It won't be clear to anyone what "migrations" are until they after read this section, which makes it a bit less useful for newcomers.

But I also don't know what could be split out of this section other than to say "Run `wrangler publish` with any desired migrations, as documented above/below".

Maybe it'd work to just name this section "Configuring Durable Object classes"? Or "Configuring Durable Object classes with migrations"?

a-robinson (approved):

products/workers/src/content/learning/using-durable-objects.md
xortive: I like "Configuring Durable Object classes with migrations"

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -269,25 +269,33 @@ The `[durable_objects]` section has 1 subsection:
...
+Migrations can also be used for transferring stored data between two Durable Object classes:
+* Rename migrations are used to transfer stored objects between two Durable Object classes in the same script
+* Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts
+
+The destination class (the one that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
 
 <Aside type="warning" header="Important">
 
-Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
+After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class. Don't do this without first ensuring that you aren't relying on the source Durable Object class having access to its stored data.
```
a-robinson: > Don't do this without first ensuring that you aren't relying on the source Durable Object class having access to its stored data.

The "having access to its stored data" bit still kind of gets me here. After a rename/transfer migration, wouldn't it be access to the Durable Object class/instances themselves that would be lost, not just their data?

But actually, more importantly, in practice I don't think you lose any access to the Durable Objects owned by the class when doing a rename/transfer -- any bindings to the source use the namespace ID under the hood, and that ID will remain valid and continue to work after the rename/transfer. It'll just point at different implementation code now.

kristianfreeman:
cc @deadlypants1973 for style review

@xortive we should also have a section in https://developers.cloudflare.com/workers/cli-wrangler/configuration for these new changes, too

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -245,11 +245,7 @@ This will create a directory for your project with basic configuration and a sin
 * [Durable Objects Rollup ES Modules template](https://github.com/cloudflare/durable-objects-rollup-esm)
 * [Durable Objects Webpack CommonJS template](https://github.com/cloudflare/durable-objects-webpack-commonjs)
 
-The following sections will cover how to customize the configuration, but if you'd like you can immediately publish the generated project using this command:
-
-```sh
-$ wrangler publish --new-class Counter
-```
+The following sections will cover how to customize the configuration, but if you'd like you can immediately publish the generated project using `wrangler publish`.
```
deadlypants1973:
```suggestion
The following sections will cover how to customize the configuration, but you can also immediately publish the generated project using `wrangler publish`.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
   { name = "EXAMPLE_CLASS", class_name = "DurableObjectExample" } # Binding to our DurableObjectExample class
 ]
 ```
+
 The `[durable_objects]` section has 1 subsection:
 
 - `bindings` - An array of tables, each table can contain the below fields.
   - `name` - Required, The binding name to use within your worker.
```
deadlypants1973:
```suggestion
  - `name` - Required, The binding name to use within your Worker.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
+### Configuring Durable Object classes with migrations
 
-Normally when you want to publish a Worker using Wrangler, you just run `wrangler publish`. However, when you export a new Durable Objects class from your script, you must tell the Workers platform about it
-before you can create and access Durable Objects associated with that class. This process is called a "migration", and is currently performed by providing extra options to `wrangler publish`.
+When you export a new Durable Objects class from your script, you must tell the Workers platform about it before you can create and access Durable Objects associated with that class from a Durable Object binding. This process is called a "migration".
 
-To allow creation of Durable Objects associated with an exported class, specify `--new-class`:
+Migrations can also be used for transferring stored data between two Durable Object classes:
+* Rename migrations are used to transfer stored objects between two Durable Object classes in the same script
+* Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts
```
deadlypants1973:
```suggestion
* Rename migrations are used to transfer stored objects between two Durable Object classes in the same script.
* Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
 
-```sh
-$ wrangler publish --new-class DurableObjectExample
-```
+The destination class (the one that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
```
deadlypants1973:
```suggestion
The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
 
-Note that after you've run `--new-class` for a given class name once, you do not need to include the migration on subsequent uploads of the Worker. You'd just run `wrangler publish` with no additional flags.
+<Aside type="warning" header="Important">
 
-If you want to delete the Durable Objects associated with an exported class, remove the corresponding binding from wrangler.toml, then use `--delete-class`:
+After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class. Don't do this without first ensuring that you aren't relying on the source Durable Object class having access to its stored data.
```
deadlypants1973:
```suggestion
After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class. Do not run a rename or transfer migration without first ensuring that you are not relying on the source Durable Object class having access to its stored data.
```

Can we change the "this" in this sentence to be a specific action? "Do not ______ without..." I inserted my suggestion above.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
-$ wrangler publish --delete-class DurableObjectExample
-```
+</Aside>
+
+Migrations can also be used to delete a given Durable Object class along with its stored objects.
 
 <Aside type="warning" header="Important">
 
-Running a `--delete-class` migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
+Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
```
deadlypants1973:
```suggestion
Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Do not run a delete migration on a class without first ensuring that you are not relying on the Durable Objects within that class anymore. Copy any important data to some other location before deleting. 
```

Need to have some technical eyes on this for accuracy in wording.

products/workers/src/content/learning/using-durable-objects.md
xortive: Bah, my memory is stale here and this is one of those things I struggle to explain correctly without mentioning namespaces :/ I see what you mean, there's no "source class" after the migration runs.

I think what's missing here is a mention that rename and transfer migrations mean you also have to update the binding in the `[durable_objects]` section for future uploads, but I'm not actually sure what happens if you rename class A to class B in script A, and script B is bound to (script A, class A) -- does the binding continue to work, now pointing to class B, and future uploads need to update the class name? I think this is the case, but it's a bit confusing....

Here's try 3, assuming the above is correct.

>
>After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class.
>
>Existing bindings to the source Durable Object class will continue to work, now pointing to the destination Durable Object class. Future script uploads must update the class name for the binding in the `[durable_objects]` section of `wrangler.toml`.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
+
+</Aside>
+
+Migrations can be performed in two different ways: by listing them in `wrangler.toml` or by manually providing "ad hoc" migrations using extra options on wrangler publish. Migrations specified in `wrangler.toml` have the additional requirement of a "migration tag", which is used to determine which migrations have already been applied, and therefore can be skipped for a given script upload. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
```
deadlypants1973:
```suggestion
Migrations can be performed in two different ways: by listing them in `wrangler.toml` or by manually providing ad hoc migrations using extra options on `wrangler publish`. Migrations specified in `wrangler.toml` have the additional requirement of a **migration tag**, which is used to determine which migrations have already been applied, and therefore can be skipped for a given script upload. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
```

Is "migration tag" a field in wrangler.toml files? If so, we should bold it.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
+
+### Durable Object migrations in `wrangler.toml`
+
+The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments, and cannot be overridden by a specific environment. Like "ad hoc" migrations, they are applied when you run `wrangler publish` on the selected environment's script. For our `DurableObjectExample` class, we need a migration list like this:
```
deadlypants1973:
```suggestion
The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments, and cannot be overridden by a specific environment. Like ad hoc migrations, they are applied when you run `wrangler publish` on the selected environment's script. For our `DurableObjectExample` class, you need a migration list like this:
```

What is the 'they' in 'they are applied...'?

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,90 @@ bindings = [
...
+new_classes = ["DurableObjectExample"] # array of new classes
+deleted_classes = ["DeprecatedClass"] # array of deleted classes
+renamed_classes = [{from: "OldClass", to: "NewClass"}] # array of rename directives
+transferred_classes = [{from_script: "old_script", from: "ExampleClass", to: "ExampleClass"}] # array of transfer directives
+```
+
+<Aside>
+
+Note that toml doesn't allow line breaks in inline tables (the `{key: "value"}` syntax), but line breaks
+in the surrounding inline array are OK.
```
deadlypants1973:
```suggestion
<Aside type="note">

Note that `.toml` files do not allow line breaks in inline tables (the `{key: "value"}` syntax), but line breaks
in the surrounding inline array are acceptable.
```

deadlypants1973 (commented):
Review round 1 - taking a second look in a few hours.

greg-mckeon (approved):

products/workers/src/content/learning/using-durable-objects.md
a-robinson: > if you rename class A to class B in script A, and script B is bound to (script A, class A) -- does the binding continue to work, now pointing to class B, and future uploads need to update the class name? I think this is the case, but it's a bit confusing....

Yes, that's what I'd expect. Your updated text looks good to me.

products/workers/src/content/learning/using-durable-objects.md
xortive: Looks good to me

products/workers/src/content/learning/using-durable-objects.md
xortive: yep!

products/workers/src/content/learning/using-durable-objects.md
xortive: migrations in the migration list

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
   - `class_name` - Required, The class name you wish to bind to.
-  - `script_name` - Optional, Defaults to the current project's script.
+  - `script_name` - Optional, Defaults to the current environment's script.
 
-### Publishing Durable Object classes
+### Configuring Durable Object classes with migrations
 
-Normally when you want to publish a Worker using Wrangler, you just run `wrangler publish`. However, when you export a new Durable Objects class from your script, you must tell the Workers platform about it
-before you can create and access Durable Objects associated with that class. This process is called a "migration", and is currently performed by providing extra options to `wrangler publish`.
+When you export a new Durable Objects class from your script, you must tell the Workers platform about it before you can create and access Durable Objects associated with that class from a Durable Object binding. This process is called a "migration".
```
deadlypants1973:
```suggestion
When you make changes to your list of Durable Objects classes, you must initiate a migration process. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
```

deadlypants1973 (commented):
review round 2

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
+Existing bindings to the source Durable Object class will continue to work, now pointing to the destination Durable Object class. Future script uploads must update the class name for the binding in the `[durable_objects]` section of `wrangler.toml`.
+
+</Aside>
+
+Migrations can also be used to delete a given Durable Object class along with its stored objects.
 
 <Aside type="warning" header="Important">
 
-Running a `--delete-class` migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
+Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Do not run a delete migration on a class without first ensuring that you are not relying on the Durable Objects within that class anymore. Copy any important data to some other location before deleting.
```
deadlypants1973:
```suggestion
Running a delete migration will delete all Durable Object instances associated with the deleted class, including all of their stored data. Do not run a delete migration on a class without first ensuring that you are not relying on the Durable Objects within that class anymore. Copy any important data to some other location before deleting. 
```

xortive: curious, are we consistent about `Durable Objects` vs `Durable Object instances` in our docs today? I don't feel super strongly since it's annoying that the product name is also used here for the first option, and the second is similar to `ATM machine` which also bothers me :p I'd rather stick to whatever we use elsewhere

lukeed: This was my suggestion. There's already common confusion about "Durable Object" referring to the class definition vs the live copies/instances. The _what_ that "Durable Object(s)" actually refers to is implied by the rest of the sentence.

Agree we need/want to be consistent, but as it is we have 1 term for 2 slightly different things. I've heard "instances" (and actors) in meetings so I thought it'd be okay to use it here for clarification

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
 
-To allow creation of Durable Objects associated with an exported class, specify `--new-class`:
+Migrations can also be used for transferring stored data between two Durable Object classes:
+* Rename migrations are used to transfer stored objects between two Durable Object classes in the same script.
+* Transfer migrations are used to transfer stored objects between two Durable Object classes in different scripts.
 
-```sh
-$ wrangler publish --new-class DurableObjectExample
-```
+The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the script.
```
deadlypants1973:
```suggestion
The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the deployed script.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
 
-Note that after you've run `--new-class` for a given class name once, you do not need to include the migration on subsequent uploads of the Worker. You'd just run `wrangler publish` with no additional flags.
+<Aside type="warning" header="Important">
 
-If you want to delete the Durable Objects associated with an exported class, remove the corresponding binding from wrangler.toml, then use `--delete-class`:
+After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class.
```
deadlypants1973: @xortive Can you verify whether 'stored objects' refers to the live instances or the data within them?

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
 
-```sh
-$ wrangler publish --delete-class DurableObjectExample
-```
+Existing bindings to the source Durable Object class will continue to work, now pointing to the destination Durable Object class. Future script uploads must update the class name for the binding in the `[durable_objects]` section of `wrangler.toml`.
```
deadlypants1973:
```suggestion
After a migration, any existing bindings to the original Durable Object class (e.g., from other Workers) will automatically forward to the updated destination class. However, any Worker scripts bound to the updated Durable Object class must update their `[durable_objects]` configuration in the `wrangler.toml` file for their next deployment.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
+
+</Aside>
+
+Migrations can also be used to delete a given Durable Object class along with its stored objects.
```
deadlypants1973:
```suggestion
Migrations can also be used to delete a Durable Object class and its stored objects.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
 
 <Aside type="warning" header="Important">
 
-Running a `--delete-class` migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Don't do this without first ensuring that you aren't relying on the Durable Objects anymore and have copied any important data to some other location.
+Running a delete migration will delete all Durable Objects associated with the deleted class, including all of their stored data. Do not run a delete migration on a class without first ensuring that you are not relying on the Durable Objects within that class anymore. Copy any important data to some other location before deleting. 
+
+</Aside>
+
+Migrations can be performed in two different ways: by listing them in `wrangler.toml` or by manually providing ad hoc migrations using extra options on `wrangler publish`. Migrations specified in `wrangler.toml` have the additional requirement of a **migration tag**, which is used to determine which migrations have already been applied, and therefore can be skipped for a given script upload. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
```
deadlypants1973:
```suggestion
Migrations can be performed in two different ways: through the `[[migrations]]` configurations key in your `wrangler.toml` file or through additional CLI arguments during a `wrangler publish` command. Migrations specified in `wrangler.toml` have the additional requirement of a migration tag, which is defined by the **tag** property in each migration entry. Migration tags are treated like unique names and are used to determine which migrations have already been applied. Once a given script has a migration tag set on it, all future script uploads must include a migration tag.
```

deadlypants1973: @xortive can you please explain what requirement this is trying to state? "Once a given script has a migration tag set on it, all future script uploads must include a migration tag."

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
+
+### Durable Object migrations in `wrangler.toml`
+
+The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments, and cannot be overridden by a specific environment. Like ad hoc migrations, they are applied when you run `wrangler publish` on the selected environment's script. For our `DurableObjectExample` class, you need a migration list like this:
```
deadlypants1973:
```suggestion
The migration list (added in `wrangler 1.19.0`) is an array of tables, specified as a top-level key in your `wrangler.toml`. The migration list is inherited by all environments and cannot be overridden by a specific environment. 

All migrations are applied at deployment. This is true for all migrations, whether initiated through `wrangler.toml` configurations or `wrangler publish` command arguments. Each migration can only be applied once per [environment](/platform/environments). 

To illustrate an example migrations workflow, the `DurableObjectExample` class can be initially defined with:
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
+
+```toml
+[[migrations]]
+tag = "v1" # should be unique for each entry
+new_classes = ["DurableObjectExample"] # array of new classes
```
deadlypants1973:
```suggestion
tag = "v1" # Should be unique for each entry
new_classes = ["DurableObjectExample"] # Array of new classes
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
+```
+
+Each migration in the list can have multiple directives, and multiple migrations can be specified as your project grows in complexity.
```
deadlypants1973:
```suggestion
Each migration in the list can have multiple directives, and multiple migrations can be specified as your project grows in complexity. For example, you may want to rename the `DurableObjectExample` class to `UpdatedName` and delete an outdated `DeprecatedClass` entirely.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,92 @@ bindings = [
...
+
+```toml
+[[migrations]]
+tag = "v1" # should be unique for each entry
+new_classes = ["DurableObjectExample"] # array of new classes
+deleted_classes = ["DeprecatedClass"] # array of deleted classes
+renamed_classes = [{from: "OldClass", to: "NewClass"}] # array of rename directives
+transferred_classes = [{from_script: "old_script", from: "ExampleClass", to: "ExampleClass"}] # array of transfer directives
```
deadlypants1973:
```suggestion
[[migrations]]
tag = "v1" # Should be unique for each entry
new_classes = ["DurableObjectExample"] # Array of new classes

[[migrations]]
tag = "v2"
renamed_classes = [{from: "DurableObjectExample", to: "UpdatedName" }] # Array of rename directives
deleted_classes = ["DeprecatedClass"] # Array of deleted class names
```

deadlypants1973 (commented):
Review round 3

products/workers/src/content/learning/using-durable-objects.md
xortive: The clarity argument works for me. I'll use `Durable Object instances`

products/workers/src/content/learning/using-durable-objects.md
xortive: The data within them

products/workers/src/content/learning/using-durable-objects.md
xortive: This is much clearer, thank you!

products/workers/src/content/learning/using-durable-objects.md
xortive: Sorry, this was not clear at all on second read :/ The API requirement here is "All future script uploads that include a migration, must include the old migration tag (equal to the current migration tag of the script) and a new migration tag."

If you use `[[migrations]]`, this is done for you automatically as all migrations in the migrations list will have a tag.

If you use ad hoc migrations after a migration tag has been set on a script, you need to use `--old-tag` and `--new-tag` for all future adhoc migrations.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
-```sh
-$ wrangler publish --new-class DurableObjectExample
-```
+The destination class (the class that stored objects are being transferred to) for a rename or transfer migration must be exported by the deployed script.
 
-Note that after you've run `--new-class` for a given class name once, you do not need to include the migration on subsequent uploads of the Worker. You'd just run `wrangler publish` with no additional flags.
+<Aside type="warning" header="Important">
 
-If you want to delete the Durable Objects associated with an exported class, remove the corresponding binding from wrangler.toml, then use `--delete-class`:
+After a rename or transfer migration, requests to the destination Durable Object class will have access to the stored objects of the source Durable Object class.
```
deadlypants1973:
```suggestion
After a rename or transfer migration, requests to the destination Durable Object class will have access to the source Durable Object's stored data. 
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
+
+<Aside type="note">
+
+Note that `.toml` files do not allow line breaks in inline tables (the `{key: "value"}` syntax), but line breaks
+in the surrounding inline array are acceptable.
 
 </Aside>
 
-These are basic examples -- you can use multiple of these options in a single `wrangler publish` call if you'd like, one class per option. Future versions of Wrangler will also include migration directives for renaming a class or transferring a class from one file to another.
+### Ad hoc Durable Object migrations
```
deadlypants1973:
```suggestion
### Durable Object migrations through Wrangler CLI
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
+
+Ad hoc migrations are performed using extra options on `wrangler publish`. Several options can be passed at once in a single `wrangler publish` call. You can also specify `--old-tag <tag>` and `--new-tag <tag>` if your script has a migration tag associated with it but you still want to do an ad hoc migration.
```
deadlypants1973:
```suggestion
As an alternative, migrations can be defined through extra arguments to the `wrangler publish` command. If you would like to reference migrations defined in your `wrangler.toml` file, you can use the `--old-tag <tag>` and `--new-tag <tag>` arguments to reference those migrations by name.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
+
+When an ad hoc migration is performed, any migrations listed in `wrangler.toml` are ignored.
```
deadlypants1973:
```suggestion
When defining a CLI migration, and not using the `--new-tag` and `--old-tag` approach, any migrations listed in `wrangler.toml` are ignored. 
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
+
+The list of ad hoc migration options that can be added to `wrangler publish` is as follows:
```
deadlypants1973:
```suggestion
The list of CLI migration arguments that can be added to `wrangler publish` is as follows:
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
+```sh
+--old-tag <tag name> # old-tag is optional if your script doesn't have a migration tag set yet
+--new-tag <tag name> # new-tag and old-tag are optional if you only use ad hoc migrations.
+
+# Each of the migration directives can be specified multiple times if you are
+# creating/deleting/renaming/transferring multiple classes at once.
+--new-class <class name>
+--delete-class <class name>
+--rename-class <from class> <to class>
+--transfer-class <from script> <from class> <to class>
```
deadlypants1973:
```suggestion
--old-tag <tag name> # Optional if your script does not have a migration tag set yet.
--new-tag <tag name> # new-tag and old-tag are optional if you only use CLI migrations.

# Each of the migration directives can be specified multiple times if you are
# creating/deleting/renaming/transferring multiple classes at once.
--new-class <class name>
--delete-class <class name>
--rename-class <from class> <to class>
--transfer-class <from script> <from class> <to class>
```

deadlypants1973 (commented):
Review round 4

products/workers/src/content/learning/using-durable-objects.md
xortive: This isn't quite right:

The API doesn't keep a history of migration tags, only the script's current migration tag (the uniqueness constraint for migration tags is only about the migrations currently in wrangler.toml)

`--old-tag` is a safety measure to ensure the migration being attempted is actually written against the script in it's current state, not some old state.

products/workers/src/content/learning/using-durable-objects.md
xortive: Same comment as above

products/workers/src/content/learning/using-durable-objects.md
lukeed: The changes weren't suggesting that an API knew/remembered anything – just that the old & new tags had to exist in the `wrangler.toml`

But are you saying that `old-tag` can always only be the currently-active migration? If that's the case, then the API has to remember the most recently applied migration tag in order for it to confirm. Also, what about "ad hoc migrations" that were done purely thru `--new-class`/`--delete-class` arguments? Those have no tag names, and it seems that the landscape of possible approaches would make it very very easy for a user to get out of sync – and/or have no record of what they did.

xortive: Sorry, I glossed over some details in my earlier response.

>  just that the old & new tags had to exist in the wrangler.toml

To be exact, this part is what's not true.

`--old-tag` and `--new-tag` are used in addition with `--new-class/--delete-class/etc`, to perform an ad-hoc migration that adds or changes a migration tag. `--old-tag` is a safety measure to ensure that if you are migrating a script that has a tag set, you're aware of the current tag (and hopefully it's state) as you provided it under `--old-tag`. `--new-tag` doesn't refer to an unpublished migration in the `wrangler.toml` list, it directly sets the new migration tag on the script for that publish command.

> Also, what about "ad hoc migrations" that were done purely thru --new-class/--delete-class arguments? Those have no tag names, and it seems that the landscape of possible approaches would make it very very easy for a user to get out of sync – and/or have no record of what they did.

This is why we introduced tagged migrations in `wrangler.toml`, CLI migrations suck to use for long-term projects. They're nice for one-off changes and small projects though. Most projects should pick one and stick with it

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,98 @@ bindings = [
...
+
+<Aside type="note">
+
+Note that `.toml` files do not allow line breaks in inline tables (the `{key: "value"}` syntax), but line breaks
+in the surrounding inline array are acceptable.
 
 </Aside>
 
-These are basic examples -- you can use multiple of these options in a single `wrangler publish` call if you'd like, one class per option. Future versions of Wrangler will also include migration directives for renaming a class or transferring a class from one file to another.
+### Durable Object migrations through Wrangler CLI
```
xortive, line 334: This heading is somewhat ambiguous for me now -- I think it's important to include `ad hoc` in it to distinguish from migrations in `wrangler.toml`

products/workers/src/content/learning/using-durable-objects.md
lukeed: Got it, thanks. Couldn't we just disallow CLI migrations outright? Personally I don't think we should be opening the door for something to get so out-of-hand so quickly. 

Ok, so if I'm understanding correctly, the CLI migration would still require the `--new-tag` to be used to name the new state. If that is the case, then this name is potentially lost forever, making the _next_ migration fail since you forgot the name & cannot pass it thru as the `--old-tag` the next time.

If `--new-tag` isn't required for the CLI migration, then the burden shifts to you remembering the new state, still making it very possible that your next migration fails due to mismatched state.

While I understand that CLI migrations can be appealing for quick-fixes, the _lack_ of a paper trail is very concerning and what will set up a disconnect/unknown state. We'd have disappearing migrations in the same way that we have disappearing `wrangler secret` values, but with the likelihood of being more confusing. It's arguably still quick & easy to define the 3-4 lines of `wrangler.toml` migration code and always have it there for reference.

In that world, if you define a `wrangler.toml` migration, then `wrangler publish` locally, and never bother to git-commit the new migration, that forgotten/detached state is directly your fault for not having kept the record.

xortive: > Couldn't we just disallow CLI migrations outright

@greg-mckeon Do you have thoughts here? CLI/ad hoc migrations made it into a stable release of wrangler, it's a breaking change to remove them, but I agree that using solely `wrangler.toml` migrations is a better experience for majority of users

koeninger: IMHO release as is because it'd be a breaking change and ad-hoc migrations are useful.  If we're concerned about it add an explicit note in the "Migrations can be performed in two different ways" paragraph that you're better off sticking with one or the other.

lukeed: But isn't this what beta stages are for? Breaking changes shouldn't be done freely, but if they need to happen, it is far more permissible as a beta than as a GA product. Once it's GA we're stuck w/ this forever

koeninger: Since it's a beta we can get user feedback on this change as well.  Worst case scenario if it proves to be a problem we make the error message say what the current expected old tag is, right?

lukeed: Sure, but it would maybe have to print the entire "state" for all Durable Objects too? The name alone might not be enough to remind you what happened. Maybe a team member blew away a few DOs or renamed them, and so your new migration has to work with those changes as its new starting point – or you realize that it's no longer needed/applicable.

xortive: >  the error message say what the current expected old tag is, right?

this is what my PR does already

greg-mckeon: can we deprecate and have CLI migrations print what you need to add to your wrangler.toml to do the tagged migration?  then we can consider removing after we get feedback.

greg-mckeon: I'd much rather everything be in the wrangler.toml, since that's the direction we're heading.

lukeed:
Will make an additional couple of edits before merging this through. Need to aligning on the direction/deprecation notice for the CLI/ad-hoc path.

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -265,39 +261,104 @@ bindings = [
...
+
+<Aside type="warning" header="Deprecation Notice">
+    
+While CLI migrations initially served a way to quickly migrate Durable Objects, this method is now deprecated and will be removed in a future release. This is because this approach makes it easy to forget or lose track of previous migrations, thus making it very difficult to mentally track the current infrastructure state and/or coordinate with teammates.
```
lukeed:
```suggestion
While CLI migrations initially served a way to quickly migrate Durable Objects, this method is now deprecated and will be removed in a future release.
```

products/workers/src/content/learning/using-durable-objects.md
lukeed: Included a deprecation notice in this docs draft, as I hear @xortive is taking care of deprecating it separately (🙏)
thanks!

lukeed (approved):

PR #1858, opened 2021-07-23
Rules language - Add section on Arrays

pedrosousa:
Addresses https://jira.cfops.it/browse/PCX-1895.

### DIFF
--- a/products/firewall/src/content/cf-firewall-language/functions.md
+++ b/products/firewall/src/content/cf-firewall-language/functions.md
@@ -22,7 +22,7 @@ In the expression below, the`lower()` function transforms `http.host` values to 
 lower(http.host) == "www.cloudflare.com"
 ```
 
-Transformation functions that take arrays as an argument type require the `[*]` index operator, which returns an array of values for each field.
+Transformation functions that take arrays as an argument type require the `[*]` special index notation. Refer to [Arrays](/cf-firewall-language/values#arrays) for more information.
 
 The Cloudflare Firewall Rules language supports these transformation functions:
 
--- a/products/firewall/src/content/cf-firewall-language/functions.md
+++ b/products/firewall/src/content/cf-firewall-language/functions.md
@@ -22,7 +22,7 @@ In the expression below, the`lower()` function transforms `http.host` values to 
 lower(http.host) == "www.cloudflare.com"
 ```
 
-Transformation functions that take arrays as an argument type require the `[*]` special index notation. Refer to [Arrays](/cf-firewall-language/values#arrays) for more information.
+Transformation functions that do not take arrays as an argument type require the `[*]` special index notation. Refer to [Arrays](/cf-firewall-language/values#arrays) for more information.
 
 The Cloudflare Firewall Rules language supports these transformation functions:
 
--- a/products/firewall/src/content/cf-firewall-language/values.md
+++ b/products/firewall/src/content/cf-firewall-language/values.md
@@ -53,15 +53,15 @@ The Cloudflare Firewall Rules language includes [fields](/cf-firewall-language/f
 
 You can access individual array elements using an index (a non-negative value) between square brackets (`[]`). Array indexes start at `0` (zero).
 
-Functions like `all()` and `any()` work with arguments of `Array` type. Use the `[*]` index/unpacking operator when specifying an expression that should be evaluated in the context of each array element.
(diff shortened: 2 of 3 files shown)


products/firewall/src/content/cf-firewall-language/values.md
```diff
@@ -47,6 +47,48 @@ To match requests where `ssl` is `false`, use the boolean `not` operator :
 not ssl
 ```
 
+## Arrays
+
+The Cloudflare Firewall Rules language includes [fields](/cf-firewall-language/fields) of `Array` type and [functions](/cf-firewall-language/functions) with `Array` arguments and return values.
+
+You can access individual array elements using an index (a non-negative value) between square brackets (`[]`). Array indexes start at `0` (zero).
+
+Use the special notation `[*]` when specifying an expression that will be evaluated for each array element. This special index notation will unpack the array, call the enclosing function for all its elements individually, and return a new array containing all the individual return values.
```
marmeladema: I would add a link to this wiki page as "reference": https://en.wikipedia.org/wiki/Map_(higher-order_function)

products/firewall/src/content/cf-firewall-language/functions.md
```diff
@@ -22,7 +22,7 @@ In the expression below, the`lower()` function transforms `http.host` values to
 lower(http.host) == "www.cloudflare.com"
 ```
 
-Transformation functions that take arrays as an argument type require the `[*]` index operator, which returns an array of values for each field.
+Transformation functions that take arrays as an argument type require the `[*]` special index notation. Refer to [Arrays](/cf-firewall-language/values#arrays) for more information.
```
marmeladema: I think I understand what you mean but it's actually the opposite: you need to use the `[*]` notation for transformations functions that do *not* take arrays as argument, otherwise you can just specify the argument as-is like for the `any` function.

products/firewall/src/content/cf-firewall-language/values.md
```diff
@@ -47,6 +47,48 @@ To match requests where `ssl` is `false`, use the boolean `not` operator :
...
+* Check if the first array element is equal to `Content-Type` (case sensitive):<br/>
+  `http.request.headers.names[0] == "Content-Type"`
+
+* Check if any array element is equal to `Content-Type` (case sensitive):<br/>
+  `any(http.request.headers.names[*] == "Content-Type")`
+
+* Check if any array element is equal to `Content-Type`, ignoring the case:<br/>
+  `any(lower(http.request.headers.names[*])[*] == "content-type")`
+
+In the last example, the `lower()` function includes the `[*]` notation so that the function is evaluated for each array element. This function, used along `[*]`, returns a new array where each element of the input array is converted to lowercase. Then, the `any()` function also uses `[*]` so that the full expression provided to `any()` is evaluated in the context of each element of the array returned by `lower()`.
```
marmeladema: In this example, the `any` function does not "use" the `[*]` notation but the string comparison does.

marmeladema (approved):

products/firewall/src/content/cf-firewall-language/values.md
pedrosousa: Should I say the following instead?
> Then, the string comparison uses `[*]` to transform the array resulting from applying `lower()` to each header name into an array of boolean values. Finally, `any()` evaluates to true if at least one of these array elements is true.

products/firewall/src/content/cf-firewall-language/functions.md
pedrosousa: You're right, I just added the link but that sentence is not correct. Fixed to say the opposite.

products/firewall/src/content/cf-firewall-language/values.md
pedrosousa: We usually don't link to Wikipedia, but in this case it makes sense because most external pages are about a specific language.

pedrosousa: Added link.

products/firewall/src/content/cf-firewall-language/values.md
pedrosousa: Updated.

marciocloudflare (approved):

products/firewall/src/content/cf-firewall-language/values.md
marmeladema: Great! Thanks!

PR #1869, opened 2021-07-26
Update scheduled-changes.md

umahasit:


products/waf/src/content/change-log/scheduled-changes.md
```diff
@@ -29,29 +29,4 @@ order: 1
             </th>
         </tr>
     </thead>
-    <tbody>
```
pedrosousa, line 32: We should probably keep the `<tbody>` (line 32) / `</tbody>` (line 56) tags when deleting all rows.

umahasit: Thanks! added tbody.  Also forgot to add the new table header Rule ID + Legacy Rule ID, to account for both WAFs

pedrosousa (approved):

PR #1870, opened 2021-07-26
Create 2021-07-26.md

umahasit:


products/waf/src/content/change-log/2021-07-26.md
```diff
@@ -0,0 +1,113 @@
...
+                OWASP XSS (URI)
+            </td>
+            <td>
+                N/A
+            </td>
+            <td>
+                uri-973345
+            </td>
+            <td>
+                Improve Rule Performance
```
pedrosousa:
```suggestion
                Improve Rule performance
```
(To make it consistent.)

products/waf/src/content/change-log/2021-07-26.md
```diff
@@ -0,0 +1,113 @@
...
+            <td>
+                Block
+            </td>
+        </tr>
+        <tr>
+            <td>
+                Specials
+            </td>
+            <td>
+                fe9ace..
```
pedrosousa: Not sure if this line is incomplete or truncated on purpose.

umahasit: I've truncated this instead of showing the whole 32 character UUID, more so for visual reasons.  Though happy to include the whole UUID if deemed more suitable (this is the first time we are adding the new waf ID's, so happy to adjust to see what looks better etc).  Also worth noting that partial UUID searching in WAF dashboard will return the relevant rule.

products/waf/src/content/change-log/2021-07-26.md
umahasit: Changed all instances to `Improve rule performance`

products/waf/src/content/change-log/2021-07-26.md
pedrosousa: Oh, that's why I couldn't find a previous occurrence — this one is for the new WAF.

I think it's okay to proceed with the truncated value, as long as it matches a single rule. However, I noticed that in the UI we display the _last_ characters of the rule ID and not the _first_ characters when truncating the rule ID, which may cause some confusion:
image

umahasit: Good spot, edited to reflect the same representation in dashboard

umahasit:
> (Last) Nitpick: Can we also use a three-dot prefix (`...`) like in the UI?

additional . added

pedrosousa (approved):

PR #1893, opened 2021-07-28
Update example for client certificate in an Android app

frvade:
The method `OkHttpClient.Builder().sslSocketFactory` has been updated with new signature, so the example should be updated as well: https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/ssl-socket-factory/


kodster28:
Addressed by #4183

PR #1898, opened 2021-07-28
Updated documentation to reflect correct options.

rquadling:
Without these changes (using `--resource-type <resource_type>` and the resource types being prefixed with `cloudflare_`), you'll get ...
```
$ cf-terraforming --email $CLOUDFLARE_EMAIL --key $CLOUDFLARE_API_KEY --zone $CLOUDFLARE_ZONE_ID --verbose generate record
DEBU[0000] initializing cloudflare-go                    account_id= email=********** zone_id=**********
DEBU[0004] initialising Terraform in .                  
DEBU[0005] reading Terraform schema for Cloudflare provider 
DEBU[0006] beginning to read and build  resources       
"" is not yet supported for automatic generation
```

This is using `cf-terraforming v0.2.0`.

### DIFF
--- a/products/terraform/src/content/advanced-topics/importing-cloudflare-resources.md
+++ b/products/terraform/src/content/advanced-topics/importing-cloudflare-resources.md
@@ -29,26 +29,26 @@ You can use `cf-terraforming` or `cf-terraforming -h` to view the help file, but
 
 The list of supported resources currently are:
 
-* [access_application](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/access_application)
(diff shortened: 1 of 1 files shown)


KimJ15 (approved):

KimJ15:
@garrettgalow approved from a PCX perspective. Can you confirm/approve this is a change we should make?

garrettgalow (approved):
the command update looks correct.

Keeping the list of resources will be almost impossible to keep accurate. Doesn't have to be in this PR, but we should just link to https://github.com/cloudflare/cf-terraforming#supported-resources directly instead

kodster28:
Thanks for the contribution @rquadling, sorry it took a while to get to it!

Addressed these changes in another PR.

PR #1861, opened 2021-07-23
[Logs] PCX-1815: Add account-scoped data sets

soheiokamoto:
This adds account-scoped data sets:
 - Audit logs
 - Gateway DNS
 - Gateway HTTP

It adds sub navigation for zone-scoped and account-scoped data sets.

It also adds a line `The descriptions below detail the fields available for <dataset>.` on each data set page.

PR #1899, opened 2021-07-28
[Logs] PCX-1815: Add account-scoped example for Logpush API fields endpoint

soheiokamoto:
This adds account-scoped example for Logpush API `fields` endpoint, since account-scoped data sets are added here: https://github.com/cloudflare/cloudflare-docs/pull/1861


crwaters16 (approved):
Added a small tweak (made data sets plural and omitted comma after "zone-scoped data sets"). Thanks, @soheiokamoto!

PR #1901, opened 2021-07-28
Add .NET to list of language support/frameworks

Swimburger:
We noticed that the build environment now comes with .NET out of the box, but the documentation does not list it.
I added .NET to the table with the version I saw from running `dotnet --info` during Cloudflare pages build.
Not sure if .NET should be at the top or bottom when sorting alphabetically, but I put it at the bottom.

### DIFF
--- a/products/pages/src/content/platform/build-configuration.md
+++ b/products/pages/src/content/platform/build-configuration.md
@@ -79,6 +79,7 @@ Here are the pinned versions for tools included in the Cloudflare Workers build 
 | PHP       | 5.6             | `PHP_VERSION`        |                           |
 | Python    | 2.7             | `PYTHON_VERSION`     | `runtime.txt`, `Pipfile`  |
 | Ruby      | 2.6.2           | `RUBY_VERSION`       | `.ruby-version`           |
+| .NET      | 3.1.302         |                      |                           |
 
 Many common tools have been pre-installed as well. The environment variable available for overriding the pinned version is specified, as available:
 


products/pages/src/content/platform/build-configuration.md
```diff
@@ -79,6 +79,7 @@ Here are the pinned versions for tools included in the Cloudflare Workers build
 | PHP       | 5.6             | `PHP_VERSION`        |                           |
 | Python    | 2.7             | `PYTHON_VERSION`     | `runtime.txt`, `Pipfile`  |
 | Ruby      | 2.6.2           | `RUBY_VERSION`       | `.ruby-version`           |
+| .NET      | 3.1.302         |                      |                           |
```
kristianfreeman, line 82: Any idea if the version can be configured via files or env vars?

Swimburger: I do not know how the .NET SDK was installed on the machine and I did not find a version number when printing the env variable inside a build.
I usually use the [.NET install scripts ](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) which takes arguments which you can use to install specific versions of .NET.

lukeed: Doesnt look like it's (currently) possible. Just looked thru the dotnet install script & its documentation, and it seems to only be configurable via a `--version` flag, which isn't in use.

kristianfreeman (approved):

PR #1911, opened 2021-07-29
fix(images): include validation in example worker

lukeed:
Ensures that the `?image` value is acceptable & considered to be safe. Mostly is a call-out to the reader to show where/how to these types of assertions.


ganzcf (approved):

products/images/src/content/resizing-with-workers.md
```diff
@@ -227,11 +227,30 @@ async function handleRequest(request) {
   // You could adjust the URL here, e.g. prefix it with a fixed address of your server,
```
deadlypants1973:
```suggestion
  // You could adjust the URL here, e.g., prefix it with a fixed address of your server,
```

deadlypants1973 (approved):
Approved with short style guide fix.

PR #1914, opened 2021-07-30
Changed example to restrict only resizing images from origin or host

evanderkoogh:
As mentioned in https://chat.google.com/room/AAAAGX7FDAU/U0WF41wf64I, one of our customers mentioned that the code is an XSS vulnerability when straight copy & pasted. I don't agree that it is a full XSS exploit, but it would certainly make it easier to get things wrong with a straight copy and paste of the code.

The example will now only take an absolute pathname and will make the relative to either the origin if one is present and the hostname if not. This should make the example more secure by default.

I haven't been able to test this yet, so it would be great if someone could test it. I just wanted to get a PR in so we wouldn't forget.

### DIFF
--- a/products/images/src/content/resizing-with-workers.md
+++ b/products/images/src/content/resizing-with-workers.md
@@ -198,7 +198,7 @@ if (response.ok) {
 
 ### An example worker
 
-Assuming you [set up a worker](https://developers.cloudflare.com/workers/learning/getting-started) on `https://example.com/image-resizing` to handle URLs like this: `https://example.com/image-resizing?width=80&image=https://example.com/uploads/avatar1.jpg`
+Assuming you [set up a worker](https://developers.cloudflare.com/workers/learning/getting-started) on `https://example.com/image-resizing` to handle URLs like this: `https://example.com/image-resizing?width=80&image=/uploads/avatar1.jpg`
 
 ```js
 addEventListener("fetch", event => {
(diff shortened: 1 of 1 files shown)


ganzcf (approved):

products/images/src/content/resizing-with-workers.md
```diff
@@ -198,7 +198,7 @@ if (response.ok) {
 
 ### An example worker
 
-Assuming you [set up a worker](https://developers.cloudflare.com/workers/learning/getting-started) on `https://example.com/image-resizing` to handle URLs like this: `https://example.com/image-resizing?width=80&image=https://example.com/uploads/avatar1.jpg`
+Assuming you [set up a worker](https://developers.cloudflare.com/workers/learning/getting-started) on `https://example.com/image-resizing` to handle URLs like this: `https://example.com/image-resizing?width=80&image=/uploads/avatar1.jpg`
```
deadlypants1973:
```suggestion
Assuming you [set up a Worker](https://developers.cloudflare.com/workers/get-started/guide) on `https://example.com/image-resizing` to handle URLs like this: `https://example.com/image-resizing?width=80&image=/uploads/avatar1.jpg`
```

deadlypants1973 (commented):
Updated URL for Workers get started guide.

deadlypants1973 (approved):

lukeed:
There is a syntax error in these changes. Also @evanderkoogh I updated the example to address the XSS fix (#1911). In speaking to @ganzcf, adding a `hostname` check was enough to patch the vuln.

deadlypants1973:
> There is a syntax error in these changes. Also @evanderkoogh I updated the example to address the XSS fix (#1911). In speaking to @ganzcf, adding a `hostname` check was enough to patch the vuln.

Can we close this then?

PR #1927, opened 2021-08-02
Update with proper link.

AnilSeervi:
- Previously the link did point to `JSON.stringify`'s Description section but with the ID of **#Description** whereas the ID of the section is **#description**.
- There's a section within [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#symbols_and_json.stringify) Docs which explains the same thing so I've linked to that instead.


kristianfreeman (approved):
thanks!

PR #1935, opened 2021-08-02
[Workers] Partials style guide fix

deadlypants1973:


products/workers/src/content/_partials/_tutorials-before-you-start.md
```diff
@@ -1,3 +1,3 @@
 ## Before you start
 
-All of the tutorials assume you’ve already gone through [Getting started](/get-started/guide), which gets you set up with a Cloudflare Workers account, and the Workers CLI tool, [Wrangler](https://github.com/cloudflare/wrangler).
+All of the tutorials assume you have already completed the [Get started guide](/get-started/guide), which gets you set up with a Cloudflare Workers account, and the Workers CLI tool, [Wrangler](https://github.com/cloudflare/wrangler).
```
lukeed, line 3: we should be more consistent with this – I know this is already your intention, but this is `Get started` and I've seen `Get Started` where. Here I also think the link text should not include `guide` – `[Get Started](...) guide`

deadlypants1973: It should be 'Get started guide' because that is the official capitalization on the docs UI.

crwaters16 (approved):

PR #1977, opened 2021-08-05
[Tunnel] Fixed incorrect windows "installation" instructions

sudarshan-reddy:
The current documentation provides a mix of install and download instructions for an exe treating it as if it were an installer. This PR corrects this assumption. It also fixes an ambiguous instruction that leads users to assume they have to simply run `cloudflared.exe` after downloading.

### DIFF
--- a/products/cloudflare-one/src/content/connections/connect-apps/install-and-setup/installation.md
+++ b/products/cloudflare-one/src/content/connections/connect-apps/install-and-setup/installation.md
@@ -79,9 +79,9 @@ Type   | 32-bit | 64-bit |
 Executable | [Download](https://github.com/cloudflare/cloudflared/releases/latest/download/cloudflared-windows-386.exe) | [Download](https://github.com/cloudflare/cloudflared/releases/latest/download/cloudflared-windows-amd64.exe) |
 
 1. Open PowerShell.
-2. Change to your Downloads directory.
-3. Optional: Rename the executable to `cloudflared.exe`.
-3. Run .\cloudflared.exe --version --- it should output the version of cloudflared. (Note: `cloudflared.exe could be cloudflared-windows-amd64.exe or cloudflared-windows-386.exe if you haven't renamed it`).
+1. Change to your Downloads directory.
+1. (Optional) Rename the executable to `cloudflared.exe`.
+1. Run `.\cloudflared.exe --version`. It should output the version of `cloudflared`. Note that `cloudflared.exe` could be `cloudflared-windows-amd64.exe` or `cloudflared-windows-386.exe` if you haven't renamed it.
 
 <Aside>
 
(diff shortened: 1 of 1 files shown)


products/cloudflare-one/src/content/connections/connect-apps/install-and-setup/installation.md
```diff
@@ -78,12 +78,11 @@ Type   | 32-bit | 64-bit |
 -------|----------------|-----|
 ZIP | [Download](https://github.com/cloudflare/cloudflared/releases/latest/download/cloudflared-windows-386.exe) | [Download](https://github.com/cloudflare/cloudflared/releases/latest/download/cloudflared-windows-amd64.exe) |
 
-Once `cloudflared` is installed:
+Once `cloudflared` is downloaded:
 1. Navigate to the **Downloads** folder.
```
abracchi-tw: why do we have two steps related to the `Downloads` folder? can we do without the first one?

products/cloudflare-one/src/content/connections/connect-apps/install-and-setup/installation.md
```diff
@@ -78,12 +78,11 @@ Type   | 32-bit | 64-bit |
...
-2. Right-click on the ZIP folder and select `Extract All` to extract the executable.
-3. Next, open PowerShell.
-4. Navigate to the same Downloads folder.
-5. Run the `cloudflared.exe` executable as an administrator to confirm the installation, replacing the path in the example below with the specifics of your directory:
+2. Next, open PowerShell.
+3. Navigate to the same Downloads folder.
+4. Run the command below as an administrator to confirm the installation, replacing the path below with the specifics of your directory:
```
nmldiegues: A few remarks:
 - The links right above are downloading executables from github. It's not clear to me why below we have a command that shows up as if we have a folder `cloudflared-stable-amd64`
 - Point 1 says to navigate to the Downloads folder, then point 3 says to do it again?
 - I have downloaded the links above on Windows and I could just execute the obtained binary

I think this should be just:

Once `cloudflared` is downloaded:
 1. Open PowerShell.
 2. Change to your Downloads directory
 3. Run `.\cloudflared.exe --version` --- it should output the version of `cloudflared`.

(and remove mentions to "install" because there's nothing really installed, it's just a binary)

sudarshan-reddy: This is good feedback. I also changed `zip` to `executable` to reflect reality.

nmldiegues (approved):

abracchi-tw (approved):

products/cloudflare-one/src/content/connections/connect-apps/install-and-setup/installation.md
sudarshan-reddy: Makes sense to me. Looks like you already fixed this. Thanks!

PR #1983, opened 2021-08-06
Update deploy-hooks.md

nevikashah:
adding security considerations sub-header


lukeed (approved):
Ah, I noticed that was removed last time but figured it was intentional 🙃  sorry

PR #1996, opened 2021-08-06
Update durable-objects.md

greg-mckeon:
Clarified that waitUntil is available, but unnecessary.

### DIFF
--- a/products/workers/src/content/runtime-apis/durable-objects.md
+++ b/products/workers/src/content/runtime-apis/durable-objects.md
@@ -43,7 +43,7 @@ export class DurableObject {
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
 - `state.waitUntil(<Type>Promise</Type>)`
-  - While waitUntil is available within a Durable Object, it has no effect. Async I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
+  - While `waitUntil` is available within a Durable Object, it has no effect. Asynchronous I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
 
 - `state.storage`
   - Contains methods for accessing persistent storage via the transactional storage API. See [Transactional Storage API](#transactional-storage-api) for a detailed reference.
--- a/products/workers/src/content/runtime-apis/durable-objects.md
+++ b/products/workers/src/content/runtime-apis/durable-objects.md
@@ -42,8 +42,8 @@ export class DurableObject {
 - `state.id` <Type>DurableObjectId</Type>
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
-- `state.waitUntil(<Type>Promise</Type>)`
-  - While `waitUntil` is available within a Durable Object, it has no effect. Asynchronous I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
+- `state.waitUntil`
+  - While `waitUntil` is available within a Durable Object, it has no effect. Refer to [Durable Object Lifespan](#durable-object-lifespan) for more information.
 
 - `state.storage`
   - Contains methods for accessing persistent storage via the transactional storage API. See [Transactional Storage API](#transactional-storage-api) for a detailed reference.
@@ -52,6 +52,63 @@ export class DurableObject {
   - Contains environment bindings configured for the Worker script, such as KV namespaces, secrets, and other Durable Object namespaces. Note that in traditional Workers not using Modules syntax, these same "bindings" appear as global variables within the script. Scripts that export Durable Object classes always use the Modules syntax, and have bindings delivered to the constructor rather than placed in global variables.
 
 </Definitions>
+
+
+### Durable Object Lifespan
+
+A Durable Object remains active until all asynchronous I/O, including Promises, within the Durable Object has resolved. This is true for all HTTP and/or WebSocket connections, except for failure scenarios, which may include unhandled runtime exceptions or exceeding the [CPU limit](/platform/limits#durable-objects).
+
+From a Workers perspective, this is similar to enqueuing tasks with [`FetchEvent.waitUntil`](/runtime-apis/fetch-event). For example, in order to send a POST request without delaying a `Response`, a Worker script may include the following code:
+
+```js
+---
+filename: worker.mjs
+highlight: [5]
+---
+export default {
+  fetch(req, env, ctx)  {
+    // Send a non-blocking POST request.
+    // ~> Completes before the Worker exits.
+    ctx.waitUntil(
+      fetch('https://.../logs', {
+        method: 'POST',
+        body: JSON.stringify({
+          url: req.url,
+          // ...
+        })
+      })
+    );
+
+    return new Response('OK');
+  }
+}
+```
+
+The same functionality can be achieved in a Durable Object, simply by omitting the `await` for the POST request. The request will complete before the Durable Object exits:
+
+```js
+---
+filename: durable.mjs
+highlight: [6]
+---
+export class Example {
+  fetch(req) {
+    // NOTE: Omits `await` intentionally.
+    // ~> Does not block `Response` output
+    // ~> Will still wait for POST to complete
+    fetch('https://.../logs', {
+      method: 'POST',
+      body: JSON.stringify({
+        url: req.url,
+        // ...
+      })
+    });
+
+    return new Response('OK');
+  }
+}
+```
+
 
 ### Transactional Storage API
 
--- a/products/workers/src/content/runtime-apis/durable-objects.md
+++ b/products/workers/src/content/runtime-apis/durable-objects.md
@@ -42,6 +42,9 @@ export class DurableObject {
 - `state.id` <Type>DurableObjectId</Type>
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
+- `state.waitUntil(<Type>Promise</Type>)`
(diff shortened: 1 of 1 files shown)


xortive:
I don't think we should mention the existence of `state.waitUntil`, only that it is not necessary

xortive:
> Async I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.

I believe there's still a timeout for no incoming requests/websocket messages that will kill a DO even if it has a promise/io waiting to complete

lukeed:
Could this also be something that's deprecated now & removed for GA?

If it literally doesn't do anything at all (anymore?) then it probably shouldn't exist.

a-robinson:
We may not need to document `state.waitUntil()`, but we do at least need to call out that/why `waitUntil` isn't needed in the place(s) that someone would look for how to use `waitUntil` in a DO.

kristianfreeman:
> We may not need to document `state.waitUntil()`, but we do at least need to call out that/why `waitUntil` isn't needed in the place(s) that someone would look for how to use `waitUntil` in a DO.

out of curiosity, do we know that people _are_ trying to use `waitUntil` in a DO? if we find that people are trying to use it, I'm in agreement of documenting that the `waitUntil` pattern isn't relevant here as a bare minimum.

I also agree with @lukeed that if it literally is a no-op, it would be preferable to just remove it from the API. it's an extra thing someone may hold onto as part of their mental model around Durable Objects, which is confusing if it actually doesn't do anything.

> Async I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.

just to clarify - creating a promise at all keeps the DO open? or is only when we `await` a promise to have it resolve?

a-robinson:
> out of curiosity, do we know that people are trying to use waitUntil in a DO? if we find that people are trying to use it, I'm in agreement of documenting that the waitUntil pattern isn't relevant here as a bare minimum.

I can't think of a single internal team that hasn't asked about it when onboarding to DOs. The most recent internal team building on us even said "I think I'll call state.waitUntil anyways. Depending on implicit, undocumented behavior makes me nervous 🙂". Documenting the behavior here will obviously remove half of that concern, but the main point is that yes folks are trying to use waitUntil in a DO.

> I also agree with @lukeed that if it literally is a no-op, it would be preferable to just remove it from the API. it's an extra thing someone may hold onto as part of their mental model around Durable Objects, which is confusing if it actually doesn't do anything.

That's true, but it was mentioned to people and publicly documented for a while (before it was made implicit), so removing it will require a process of logging all uses of it and reaching out to everyone who's using it, asking them to change their code. Which is doable if there aren't a ton of people, but it's never a fun (or fast) process.

> just to clarify - creating a promise at all keeps the DO open? or is only when we await a promise to have it resolve?

Creating a promise doing something like `new Promise()` doesn't keep the DO open, only promises created by runtime APIs (e.g. `fetch`, or KV operations, or storage operations, or things like `setTimeout`) keep the DO open. But they don't have to be awaited.

lukeed:
I think we're fine with documenting "Hey, this thing exists but don't use it" – so long as there's a valid reason why that's there. If it follows up with a "This is deprecated and will be removed in a future release" – that's fine and digestible. Having "This thing exists, but don't use it because it literally does not do anything. We also don't plan on removing it." is not a good look.

So, for me, the blocker here is (a) what's the future-facing game plan here; which will answer (b) which variant of the message are we posting.

> but the main point is that yes folks are trying to use waitUntil in a DO

We can address this in other documentation to say how/where users should be doing `waitUntil`-type things. If that's not a satisfactory answer, then perhaps we go with Option 3, which is `waitUntil` _should_ actually do something – specifically, what people are expecting it to do.

a-robinson:
> Having "This thing exists, but don't use it because it literally does not do anything. We also don't plan on removing it." is not a good look.

I'm definitely not advocating for documenting that it exists. We shouldn't be saying "this thing exists but don't use it" to anyone. We should document that there is no `waitUntil` in Durable Objects because all I/O is implicitly waitUntil'ed.

As for whether we actually remove the current no-op implementation (breaking anyone using it today), that's really just a question of how much we care about removing the tech debt of leaving it around. But it shouldn't be part of anyone's mental model except possibly for those who were here early enough to have already added it to theirs.

kristianfreeman:
> Creating a promise doing something like new Promise() doesn't keep the DO open, only promises created by runtime APIs (e.g. fetch, or KV operations, or storage operations, or things like setTimeout) keep the DO open. But they don't have to be awaited.

thanks, that's helpful!

thinking on this more, maybe a solution in this case would be using one of our Aside components (e.g. something like the ["Unique IDs perform best"](https://developers.cloudflare.com/workers/runtime-apis/durable-objects#:~:text=unique%20ids%20perform%20best) section in this same doc)?

the content could mention that `waitUntil` exists in the API (_update after reading more recent comments - maybe let's not do that!_), but isn't used and is a no-op, and may be removed in the future (I hear you on the difficulty of finding those users using it in existing code and removing it - though that's probably an OK compromise on their part if they agreed to use something in beta, right?)

at a high-level, one thing i actually really like about the way that these Aside components were used in other parts of the page is that they add additional context on how DOs work and how devs can use them effectively. they aren't crucial "main" content, but they do help devs gain a deeper understanding of the product and make it feel more approachable, even though they're still looking at reference docs 😄 

by positioning the `waitUntil` as something that a developer might already know of, and as you mentioned, are reaching for currently but _don't actually need_, we can re-frame this as "it's awesome you know a lot about Workers, but here's something different you should know!"

in doing so, we can also de-emphasize the function as something available in the DO API and as a remnant of the alpha/beta version of DO, and hopefully move towards removing it in the future. what do you think?

koeninger:
@lukeed anyone who has used waitUntil in the original service worker syntax is going to be confused unless we explain why waitUntil isn't needed in a durable object.  That's true regardless of plans for deprecating or documenting state.waitUntil.

lukeed:
Sorry @a-robinson, I thought you were wanting to document it, as per the original PR changes.

@a-robinson @koeninger Agree, never planned to _not_ explain why `waitUntil` isn't necessary. Documenting it was always a must-have in my book. I like @signalnerve's idea of using a callout to draw attention for additional context, which could perhaps link out of to a long-form explainer.

A question for my knowledge:

With `waitUntil` "gone" – how do you reliably queue up task(s) to complete after a `Response` has been sent? AKA, the DO handles the request and returns a response early, queuing up logs to be posted so that the response could get out the door asap. Is it just `setTimeout`?

a-robinson:
> Sorry @a-robinson, I thought you were wanting to document it, as per the original PR changes.

Sorry, I had completely missed what the PR we were commenting on was doing :/ I can see why we were talking past each other.

> With waitUntil "gone" – how do you reliably queue up task(s) to complete after a Response has been sent? AKA, the DO handles the request and returns a response early, queuing up logs to be posted so that the response could get out the door asap. Is it just setTimeout?

Just call `fetch()` or whatever else you want to do prior to returning a response. For example, you'd replace:

```
  event.waitUntil(fetch('sentry.com/...'));
  return new Response('ok');
```

with just

```
  fetch('sentry.com/...');
  return new Response('ok');
```

You could put it behind a `setTimeout` if you want, but there's no need to do so.

Now you might say that looks weird and magic, which is reasonable feedback and is likely why the internal user I mentioned said they'd use `state.waitUntil` even though it's a no-op. Maybe we should do something to ease the awkwardness of it, but it really is that easy.

lukeed:
Is there a guarantee that the task will complete, in the same way that a `FetchEvent['waitUntil']` could guarantee completion? Does this extend to _any_ non-awaited `Promise` or is it only for `fetch()` and storage operations? And I imagine that any synchronous tasks (like `this.count++`) would have to be wrapped in an arbitrary Promise or `setTimeout` to defer it?

Apologies for the questions, but asking them here makes sense since we're sorta on the topic now :D 

Trying to understand if this is happening by monitoring the JS call stack or if it's a time-window thing. Because if it's based on call-stack, then forgetting to `await` something could lead to spooky timeout (?) errors, which I think will be important to call out in the docs writeup. (Whereas in a Worker, IIRC, anything missing an `await` that is outside of `waitUntil` and before the `return` will not keep the window open)

---

> _EDIT: Got my answers elsewhere :) Hiding my comment & marking as resolved._

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -42,6 +42,9 @@ export class DurableObject {
 - `state.id` <Type>DurableObjectId</Type>
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
+- `state.waitUntil(<Type>Promise</Type>)`
+  - While waitUntil is available within a Durable Object, it has no effect. Async I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
```
```suggestion
  - While `waitUntil` is available within a Durable Object, it has no effect. Asynchronous I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
```
Not sure where the technical conversation on this is at right now but from a tech writing standpoint, I believe backticks are required for waitUntil. Is 'async I/O' defined previously in this doc? If not, it should be spelled out as "asynchronous input/out (I/O)" for accessibility

kristianfreeman:
@a-robinson @koeninger,  ok with @lukeed updating this PR with my suggestion to use an aside?

koeninger:
> @a-robinson @koeninger, ok with @lukeed updating this PR with my suggestion to use an aside?

It's @greg-mckeon 's PR but sounds fine to me.  As long as it includes the phrase waitUntil so that people searching for how to do the equivalent in a D.O. are more likely to find it.

a-robinson:
Ditto.

lukeed (approved):
Approving & merging on behalf of #2197 reviewers

PR #1998, opened 2021-08-07
Edit reference to Service to match Service Manager

helloimalastair:
Currently, the documentation references `Cloudflare Tunnel agent` for the Service created by *cloudflared*, which is in actuality named the `Argo Tunnel agent`. The change reflects proper terminology.


KimJ15:
Thank you for the feedback about how we can make improvements with developers.cloudflare.com documentation. Argo Tunnel was renamed to Cloudflare Tunnel earlier this year, so the content is accurate. Will pass along the feedback that the name change is confusing.

PR #2004, opened 2021-08-09
Update integration-with-frameworks.md

frostmark:
Added info about integrations `Cloudflare Image Resizing` with Ruby on Rails


KimJ15:
Thank you for the feedback about how we can make improvements with developers.cloudflare.com documentation. This doesn't quite align with the purpose of this page, but your feedback has been noted for future consideration. Thanks!

PR #2049, opened 2021-08-12
'datetimeMinute' doesn't exist but 'datetime'

blackjack4494:
There is a minor error in this sample. `datetimeMinute` doesn't actually exist so the server will respond with an error. But `datetime` exists and in the sample response that is the key used as well.
For some reason someone used a wrong key in the sample request.

Documentation link
https://developers.cloudflare.com/analytics/graphql-api/tutorials/querying-workers-metrics

Error
image

Success (partial view of response to hide sensitive data)
image


tlozoot (approved):
Thanks for the fix!

PR #2074, opened 2021-08-15
[workers] fix incorrect getCookie function in extract-cookie-value.md 

talismanium:
fix for correct cookie find. for example, if cookie values are aaaa=triple; aa=double;
call getCookie(request, 'aa')   will return first occurance of "aa=", which leads to wrong value of "triple"

I fix searching phrase by leading space, and to all cookie string add leading space


VARsbyTOP (approved):

lukeed:
Thanks for this! It spawned some discussion.

While technically this is more correct, it's also a little bit weird & still brittle for a production setting. We decided to update this example to use a third-party package so that it's robust and ready-to-go (#2101). Updating this example to be production-ready would just mean copy-pasting library code as an example, which isn't very ideal.

Thanks again~!

PR #2093, opened 2021-08-17
pages: update `nextjs` framework guide

lukeed:
The base `create-next-app` no longer works with `next export` as of v11.0 – I'm not sure what exactly, but it requires additional configuration changes. I figure it's easier to direct users to the official Next.js example for static/export builds, which I found by trying to find the configuration changes anyway.

### DIFF
--- a/products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
+++ b/products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
@@ -8,11 +8,15 @@ Next.js is an open-source React framework for creating websites and apps. In thi
 
 ## Creating a new project
 
-Create a new project using `npx` by running the following command in your terminal:
+Create a new project using `npx` or `yarn` by running either of the following commands in your terminal:
 
 ```sh
-$ npx create-next-app
+$ npx create-next-app --example with-static-export my-app
+# or
+$ yarn create next-app --example with-static-export my-app
 ```
+
+Once complete, a new `my-app` directory will be created using the official [`with-static-export`](https://github.com/vercel/next.js/tree/canary/examples/with-static-export) example as a template.
 
 ## Creating a GitHub repository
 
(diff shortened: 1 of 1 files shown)


products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -32,31 +36,21 @@ Deploy your site to Pages by logging into the [Cloudflare dashboard](https://das
 
 | Configuration option | Value           |
 | -------------------- | ----------------|
-| Production branch    | `main`          |
-| Build command        | `npm run build` |
-| Build directory      | `out`           |
+| Production branch    | `main`           |
+| Build command        | `npm run export` |
```
kristianfreeman, line 40: if we start using the `with-static-export` template, is there a way we can make this table work for both new (using that template) and old (bringing their own code) users? btw, this also may introduce a change to our UI via the framework `<select>` when a user is creating a project - cc @1000hz

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -8,12 +8,16 @@ Next.js is an open-source React framework for creating websites and apps. In thi
...
+Create a new project using `npx` or `yarn` by running either of the following commands in your terminal:
 
 ```sh
-$ npx create-next-app
+$ npx create-next-app --example with-static-export my-app
+# or
+$ yarn create next-app --example with-static-export my-app
 ```
 
+Once complete, a new `my-app` directory will be created using the official [`with-static-export`](https://github.com/vercel/next.js/tree/canary/examples/with-static-export) example as a template.
```
deadlypants1973:
```suggestion
After creating your project, a new `my-app` directory will be generated using the official [`with-static-export`](https://github.com/vercel/next.js/tree/canary/examples/with-static-export) example as a template.
```

deadlypants1973 (approved):
Approved but Kristian's comment left unaddressed.

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
lukeed: I don't know what the current next preset is off hand, but it should always have `next export` involved. 

A recent version changed jt from `next export` alone to `next build && next export`, which this template does via npm scripts.

I think using this template is safer in the long run since they'll be more inclined to keep this working as is, whereas the default template will always be changing. In order to get it to work now, you need to hard code a bunch of paths into the config file which is (a) messy and (b) not sure if it's responsibility to teach people how to use Next. These are quick starts.

Not sure how this interfaces with the dashboard or how it affects existing users; I'm sure I'm missing something

kristianfreeman: > Not sure how this interfaces with the dashboard or how it affects existing users; I'm sure I'm missing something

ignore me! i went and double checked, the framework preset when creating a new pages project does `next build && next export`, so it's not coupled to this in any way. i think those instructions were derived from this guide at one point, but we should be ok to update this

kristianfreeman (approved):

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
lukeed: Ah gotcha. I'm really only updating this because our current guide is broken as of next@11. @nevikashah discovered this when trying to follow it & I was able to reproduce the same

PR #2101, opened 2021-08-18
chore: defer cookie parsing to external package;

lukeed:
@signalnerve Since our examples are each in a single file right now, this `<Aside>` seemed like the best alternative since there's no way to include a `package.json` for this specific example. (We can do this in the rework, of course)

Unless you meant that I should include a snippet showing `package.json` contents, like this:

```js
// package.json
{
  "dependencies": {
    "cookie": "0.4.1",
    "worktop": "0.7.1"
  }
}
```

---

Closes #2074


kristianfreeman (approved):
good with it like this!

PR #2111, opened 2021-08-19
[LB] Cleanup and a bit of re-org

kodster28:
Lots of shifting and consolidating content, so it looks a lot larger than it is.

Addresses PCX-2144.


products/load-balancing/src/content/additional-options/load-balancing-rules/index.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Custom load balancing rules
+order: 17
+pcx-content-type: concept
```
kodster28, line 4: Same as existing page, just appears new b/c of new location.

products/load-balancing/src/content/additional-options/pagerduty-integration.md
```diff
@@ -0,0 +1,16 @@
+---
+order: 40
+pcx-content-type: how-to
+---
+
+# Integrate with PagerDuty
```
kodster28, line 6: Pulled from another page, content is the same, but just appears new b/c of new location.

dcpena (approved):

PR #2115, opened 2021-08-19
2021 grid icon update

sgraden:
Updated the developers.cloudflare.com products page to use new icons from brand and to use icons coming from per product folders instead of the brand-assets repo (cc and thank you @gganch and @dcpena).

Products currently with placeholders:
- API security
- BYOIP
- Firewall Rules
- Magic Firewall
- Waiting Room

Old:
image

New:
image


third774 (approved):

Everlag (approved):

products/events/src/content/icons/terraform.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
gganch: why is this a js file? shouldnt it be a svg

sgraden: Ah that is an old file I missed when I was cleaning up. Removed

products/partners/docs-config.js
```diff
@@ -1,9 +1,13 @@
+const fs = require('fs')
+
 module.exports = {
   product: "Partners",
   pathPrefix: "/partners",
+  productIconKey: "partners",
   productLogoPathD: require('./src/content/icons/partners').pathD,
```
pedrosousa, line 7: The file you're requiring here is being deleted in the same PR. I'd keep the file (for now) because the current version of the docs engine still uses `productLogoPathD`.

gganch: good catch adding that back in!

pedrosousa: Note: I didn't add the file back, I was just noting that I should be brought back. :) Can you please add it again?

KimJ15:
I can't find the PR, but the icons were updated to the most recent version just a month or two ago. Closing this as a duplicate of completed work.

PR #2133, opened 2021-08-23
Add description for http.request.timestamp.msec

jahands:


WalshyDev:
Maybe add that you can combine this into a whole timestamp with `.sec`. Should be pretty obvious but always worth mentioning

pedrosousa:
Hi @jahands 
This PR fell out of my radar... sorry for taking so long, and thank you for your contribution!

marciocloudflare (approved):

PR #2135, opened 2021-08-23
[Magic Transit] Added info on tunnel failover

dcpena:
Added information to explain what happens when data centers and tunnels are down. 

Addresses [PCX-1592](https://jira.cfops.it/browse/PCX-1592).

### DIFF
--- a/products/magic-transit/src/content/about/probe-construction.md
+++ b/products/magic-transit/src/content/about/probe-construction.md
@@ -61,7 +61,7 @@ Applying a penalty instead of removing the route altogether preserves redundancy
 
 ### Cloudflare data centers and tunnels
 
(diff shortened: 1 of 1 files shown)


products/magic-transit/src/content/about/probe-construction.md
```diff
@@ -59,6 +59,10 @@ The values for failure penalties are intentionally extreme so that they always e
 
 Applying a penalty instead of removing the route altogether preserves redundancy and maintains options for customers with only one tunnel. Penalties also support the case when multiple tunnels are unhealthy.
 
+### Cloudflare data centers and tunnels
+
+In the event a Cloudflare data center is down, Cloudflare's edge network does not advertise your prefixes, and your packets are routed to the next closest data center. If tunnels are down on Cloudflare's side, traffic is routed based on your original GRE route priorities to prevent "blackholing traffic," which occurs when traffic is routed to a null route and dropped from the network.
```
annikagarbers: I think we only really need the first sentence of this paragraph - the second one is not necessarily wrong but could be confusing. the main reason a tunnel would be down on our side is because a data center is offline (eg. for maintenance and/or having issues)

+ maybe add a link to https://www.cloudflarestatus.com/ for reference?

annikagarbers (approved):

nayaabahsan (approved):

PR #2162, opened 2021-08-26
[SSL] Removed "three" since there are 4 pre-val methods

cascadingstyletrees:
image

Alternative, instead of removing `three`, replace `three` with `four`. 
I just find that removing is a bit more future-proof, and writing out distinct counts over three may sound odd.


kodster28 (approved):

kodster28:
Thanks for the catch! Merging this in since it's a pretty small change.

PR #2170, opened 2021-08-30
Update name of integration in description

Triloworld:
Name in content need to be same as actual service.


kodster28 (approved):
Confirmed with https://support.google.com/a/answer/10075155?hl=en.

kodster28:
Howdy @Triloworld, going to go ahead and merge since it's a small change.

PR #2171, opened 2021-08-30
[Workers] CacheTTL revision

### LINKED ISSUE #2119: Clarify that `cf: { cacheTtl: -1 }` is invalid
### Expected Behavior
The "TTL interpretation" section [at the bottom of this page](https://developers.cloudflare.com/workers/examples/cache-using-fetch#ttl-interpretation) should make clear that only `cacheTtlByStatus` supports negative values.

`cacheTtl: -1` (or any negative number) has no defined meaning, and is currently ignored by Cloudflare's infrastructure (i.e., it does not even imply `cacheEverything`).

### Actual Behavior
It is unclear that the "TTL interpretation" section linked above only applies to `cacheTtlByStatus`. Many customers assume `cacheTtl: -1` has a defined meaning. A common assumption is that it means "bypass cache", which is not the case -- it in fact does nothing.

In addition to [the "TTL Interpretation section"](https://developers.cloudflare.com/workers/examples/cache-using-fetch#ttl-interpretation) linked above, we should probably document the valid values here as well: https://developers.cloudflare.com/workers/runtime-apis/request#requestinitcfproperties

deadlypants1973:
closes #2119

### DIFF
--- a/products/workers/src/content/examples/cache-using-fetch.md
+++ b/products/workers/src/content/examples/cache-using-fetch.md
@@ -113,3 +113,5 @@ fetch(request, {
 ```
 
 This option is a version of the `cacheTtl` feature which chooses a TTL based on the response's status code and does not automatically set `cacheEverything: true`. If the response to this request has a status code that matches, Cloudflare will cache for the instructed time, and override cache directives sent by the origin. You can review [details on the `cacheTtl` feature on the Request page](/runtime-apis/request#requestinitcfproperties).
+
+</ContentColumn>
--- a/products/workers/src/content/examples/cache-using-fetch.md
+++ b/products/workers/src/content/examples/cache-using-fetch.md
@@ -73,7 +73,7 @@ A request's cache key is what determines if two requests are "the same" for cach
 fetch(event.request, { cf: { cacheKey: "some-string" } })
 ```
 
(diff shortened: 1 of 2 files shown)


lukeed (approved):

dcpena (approved):

products/workers/src/content/examples/cache-using-fetch.md
```diff
@@ -112,21 +112,4 @@ fetch(request, {
...
-- __`0`__
-  - The asset will get cached but expire immediately (revalidate from origin every time).
-
-- __`-1` or any negative value__
-  - will instruct Cloudflare not to cache at all.
-
-</Definitions>
-
-</ContentColumn>
+This option is a version of the `cacheTtl` feature which chooses a TTL based on the response's status code and does not automatically set `cacheEverything: true`. If the response to this request has a status code that matches, Cloudflare will cache for the instructed time, and override cache directives sent by the origin. You can review [details on the `cacheTtl` feature on the Request page](/runtime-apis/request#requestinitcfproperties).
```
lukeed: Looks like build is failing cuz this is missing

```suggestion
This option is a version of the `cacheTtl` feature which chooses a TTL based on the response's status code and does not automatically set `cacheEverything: true`. If the response to this request has a status code that matches, Cloudflare will cache for the instructed time, and override cache directives sent by the origin. You can review [details on the `cacheTtl` feature on the Request page](/runtime-apis/request#requestinitcfproperties).

</ContentColumn>
```

PR #2176, opened 2021-08-31
Clarify caching behavior for Cloudflare Pages

Albertozhao:
This describes what to do in zone settings if Pages users see stale assets served after a new build. Page Rules and other cache settings for a custom domain can lead to unexpected behavior after new builds complete.


Albertozhao:
Ugh I need to update my repo locally because it's pushing changes that aren't related.

PR #2180, opened 2021-08-31
Update Firewall matched data docs

migueldemoura:
Updates the documentation to the new v0.5.0 version:
https://github.com/cloudflare/matched-data-cli/releases/tag/v0.5.0

### DIFF
--- a/products/waf/src/content/managed-rulesets/payload-logging/command-line/decrypt-payload.md
+++ b/products/waf/src/content/managed-rulesets/payload-logging/command-line/decrypt-payload.md
@@ -29,6 +29,8 @@ Use the `matched-data-cli` tool to decrypt a payload in the command line.
 
     Replace `<PRIVATE_KEY>` with your private key and `<ENCRYPTED_PAYLOAD>` with the encrypted payload.
 
+    Note: The first `printf` command will make your private key visible in your command history.
+
 1. Run the following command to decrypt the payload:
 
     ```sh
--- a/products/waf/src/content/managed-rulesets/payload-logging/command-line/decrypt-payload.md
+++ b/products/waf/src/content/managed-rulesets/payload-logging/command-line/decrypt-payload.md
@@ -19,12 +19,12 @@ Use the `matched-data-cli` tool to decrypt a payload in the command line.
     ~ $ cd matched-data-cli
     ```
 
-1. Define two environment variables: one with your private key and another one with the encrypted payload:
+1. Create two files: one with your private key and another one with the encrypted payload:
 
     ```sh
-    ~/matched-data-cli $ PRIVATE_KEY=<PRIVATE_KEY>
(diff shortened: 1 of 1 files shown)


products/waf/src/content/managed-rulesets/payload-logging/command-line/decrypt-payload.md
```diff
@@ -19,37 +19,39 @@ Use the `matched-data-cli` tool to decrypt a payload in the command line.
     ~ $ cd matched-data-cli
     ```
 
-1. Define two environment variables: one with your private key and another one with the encrypted payload:
+1. Create two files: one with your private key and another one with the encrypted payload:
 
     ```sh
-    ~/matched-data-cli $ PRIVATE_KEY=<PRIVATE_KEY>
+    ~/matched-data-cli $ unset HISTFILE
```
pedrosousa: We should probably add a note below this code block about this specific command (disable command history so that the private key and the payloads are not logged).

migueldemoura: What note would you suggest we add?
Ideally we would ask the user to create the file themselves, but having a set of commands to copy/paste instead seems easier.

pedrosousa: The commands look great! I would just explain what this particular command does, since many people won't understand the command we're asking them to run. Something like the following (added after line 32, for example):

> Note: The `unset HISTFILE` disables command history so that the private key and the payloads are not logged.

Alternatively, we could add comments inside this code block, but that would feel less shell-like.

pedrosousa: I can apply these suggestions myself, but I wanted your opinion on this.

migueldemoura: Was thinking about this and I think it's best we don't include `unset HISTFILE` at all. I've amended the patch.

There are several ways to prevent the history from being persisted, but afaik they either depend on the shell you're using or have side-effects beyond just the command we're trying to "hide" (e.g., `HISTFILE` is only checked when actually exiting the shell). We distribute binaries for a range of platforms, so assuming a POSIX shell may also backfire.

Feel free to add a note about the key potentially being visible in the history. Tangential, but a future version of the cli should output the keys to files so this whole issue can be avoided altogether.

pedrosousa (approved):
Also updated the example description (it was still mentioning the previously used environment variables).

products/waf/src/content/managed-rulesets/payload-logging/command-line/decrypt-payload.md
pedrosousa: Added a note.

PR #2192, opened 2021-09-01
Update spectrum.md

Savemech:
404 to valid link

### DIFF
--- a/products/load-balancing/src/content/additional-options/spectrum.md
+++ b/products/load-balancing/src/content/additional-options/spectrum.md
@@ -8,4 +8,4 @@ title: Spectrum
 
 You can configure a load balancer to provide TCP healthchecks, failover, and traffic steering to bring resiliency to your Spectrum applications.
 
-For more details, refer to [Load balancing in Spectrum applications](https://developers.cloudflare.com/spectrum/load-balancer).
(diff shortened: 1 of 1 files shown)


products/load-balancing/src/content/additional-options/spectrum.md
```diff
@@ -8,4 +8,4 @@ title: Spectrum
 
 You can configure a load balancer to provide TCP healthchecks, failover, and traffic steering to bring resiliency to your Spectrum applications.
 
-For more details, refer to [Load balancing in Spectrum applications](https://developers.cloudflare.com/spectrum/load-balancer).
+For more details, refer to [Create a Load Balancer and Spectrum Application](https://developers.cloudflare.com/spectrum/how-to/create-load-balancer).
```
kodster28, line 11:
```suggestion
For more details, refer to [Create a Load Balancer and Spectrum Application](/about/load-balancer).
```

kodster28: We'll want to use the guide in the relative repo. The doc you're linking to is just linking back to this page.

kodster28 (changes requested):

products/load-balancing/src/content/additional-options/spectrum.md
kodster28: I'd prefer linking to the about page since it provides a bit more context as to how they interact, but this is also fine.

kodster28 (approved):

PR #2197, opened 2021-09-01
chore(do): add waituntil section

lukeed:
This adds the section we landed on via the #1996 discussion.

Targeting that branch, with plans to merge Greg's PR into main branch

### DIFF
--- a/products/workers/src/content/runtime-apis/durable-objects.md
+++ b/products/workers/src/content/runtime-apis/durable-objects.md
@@ -43,7 +43,7 @@ export class DurableObject {
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
 - `state.waitUntil`
-  - While `waitUntil` is available within a Durable Object, it has no effect. Please refer to [Durable Object Lifespan](#durable-object-lifespan) for more information.
+  - While `waitUntil` is available within a Durable Object, it has no effect. Refer to [Durable Object Lifespan](#durable-object-lifespan) for more information.
 
 - `state.storage`
   - Contains methods for accessing persistent storage via the transactional storage API. See [Transactional Storage API](#transactional-storage-api) for a detailed reference.
--- a/products/workers/src/content/runtime-apis/durable-objects.md
+++ b/products/workers/src/content/runtime-apis/durable-objects.md
@@ -42,8 +42,8 @@ export class DurableObject {
 - `state.id` <Type>DurableObjectId</Type>
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
-- `state.waitUntil(<Type>Promise</Type>)`
-  - While `waitUntil` is available within a Durable Object, it has no effect. Asynchronous I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
+- `state.waitUntil`
+  - While `waitUntil` is available within a Durable Object, it has no effect. Please refer to [Durable Object Lifespan](#durable-object-lifespan) for more information.
 
 - `state.storage`
   - Contains methods for accessing persistent storage via the transactional storage API. See [Transactional Storage API](#transactional-storage-api) for a detailed reference.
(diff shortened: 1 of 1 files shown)


products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -42,8 +42,8 @@ export class DurableObject {
 - `state.id` <Type>DurableObjectId</Type>
   - The ID of this Durable Object. It can be converted into a hex string using its `.toString()` method.
 
-- `state.waitUntil(<Type>Promise</Type>)`
-  - While `waitUntil` is available within a Durable Object, it has no effect. Asynchronous I/O, like Promises, made from within a Durable Objects always keep the Durable Object alive until they complete, excepting failure scenarios.
+- `state.waitUntil`
+  - While `waitUntil` is available within a Durable Object, it has no effect. Please refer to [Durable Object Lifespan](#durable-object-lifespan) for more information.
```
a-robinson: I thought we were actually going to remove this from the docs? Based on where @signalnerve wrote "update after reading more recent comments - maybe let's not do that!"

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -53,6 +53,63 @@ export class DurableObject {
 
 </Definitions>
 
+
+### Durable Object Lifespan
+
+A Durable Object remains active until all asynchronous I/O, including Promises, within the Durable Object has resolved. This is true for all HTTP and/or WebSocket connections, except for failure scenarios, which may include unhandled runtime exceptions or exceeding the [CPU limit](/platform/limits#durable-objects).
```
a-robinson, line 59: > A Durable Object remains active until all asynchronous I/O, including Promises, within the Durable Object has resolved.

This isn't strictly true. If an object has responded to all HTTP requests and has no open websocket connections, it will be terminated after some amount of time (30 seconds) if no new requests/connections arrive. It can't just keep running forever if no one is talking to it.

Note that this is the same as the behavior of `waitUntil` in a normal worker -- it gets up to 30 seconds after the response has been returned before getting terminated.

products/workers/src/content/runtime-apis/durable-objects.md
lukeed: Not sure what to do with it TBH. Kept it as per @koeninger's request to have "waitUntil" searchable on the page

a-robinson: Isn't it now searchable further down the page in the new section you added? Maybe we should have an `aside` up here too, though.

products/workers/src/content/runtime-apis/durable-objects.md
a-robinson: We can keep this as is, though, since it's similar to how `waitUntil` is documented for workers as a whole.

products/workers/src/content/runtime-apis/durable-objects.md
kristianfreeman: > Kept it as per @koeninger's request to have "waitUntil" searchable on the page

I think that makes sense!

kristianfreeman (approved):
really love the way this turned out! +1

products/workers/src/content/runtime-apis/durable-objects.md
lukeed: I started off with an `Aside` here & it looked "off" visually and had the same "refer to [section]" anyway. It didn't make sense to randomly link to a lifecycle section without indicating that `waitUntil` was there.

There are two options here; I'm happy to go either way but will defer to you all:

1. keep as is – "technically it's here" with immediate link to fuller context.
2. remove the Definition entirely, including any Aside/link, and rely on the Lifespan section's positioning and `"waitUntil"` mentions

a-robinson: Huh, I thought a big part of the debate on https://github.com/cloudflare/cloudflare-docs/pull/1996 was that it was undesirable to document a method that doesn't do anything. Is the conclusion just that it's better to do that than to add an awkward "aside" that doesn't really fit into the flow of the page?

lukeed: Yeah it was :) But that was under the assumption that it literally did nothing & with the hope that it'd disappear entirely. But after talking to you and to Matt I found out that there was (kinda) a case for it, or at least some user inclination to want to use it. Since people are using it (or will discover it) and it's not likely to be removed, should add a quick mention of it and divert to the preferred path.

But ya, personally I'd prefer waitUntil disappeared entirely. If that ever happened, it'd be simple to remove the Definition entry and keep the full section anyway.

```suggestion
  - While `waitUntil` is available within a Durable Object, it has no effect. Refer to [Durable Object Lifespan](#durable-object-lifespan) for more information.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -53,6 +53,63 @@ export class DurableObject {
 
 </Definitions>
 
+
+### Durable Object Lifespan
```
a-robinson, line 57: This seems like a better fit for a little further down the page, perhaps right after the `fetch() handler method` section.

products/workers/src/content/runtime-apis/durable-objects.md
a-robinson: It doesn't seem to me like an `aside` about `waitUntil` just below the bullet point for `env` would look weird -- we already have a bunch of asides down in the storage API section.

If we document `waitUntil` as existing now, when are we going to remove it? Once you document it publicly, it gets much harder to remove. And if the long-term goal is to not document it, it seems best not to reintroduce documentation saying that it exists.

But if @koeninger has any thoughts here, I'm interested in them.

koeninger: We've already publicly "documented" it for quite a while in e.g. workers-types. I'm agnostic about whether to document it it here, as long as people can figure out from the docs what they need to do instead of waituntil in a durable object.

a-robinson: Alright  ¯\\_(ツ)_/¯

a-robinson (approved):

koeninger (approved):

lukeed:
Happy to address anything else @a-robinson, but if I'm understanding correctly, there doesn't seem to be any definitive asks/suggestions here. To borrow Cody's phrase, I'm (now) agnostic to both comments and will defer to you here.

In the interest of not delaying this any further, and having both approvals, let's ship this and #1996 & revisit this content if needed.

Thanks!

PR #2240, opened 2021-09-07
Update @adaptivelink/pops to version 0.4.0

CryptoGodfatherVA:


KimJ15:
@jrf0110 or @gganch - Do you approve this change?

jrf0110:
I don't think we're actually using this dependency. @gganch can you open a PR to remove it?

PR #2251, opened 2021-09-09
fix http code on "Request entity too large"

nabbisen:
This commit replaces 403 with 413 that I actually met.


kodster28 (approved):

kodster28:
Thanks for the find, @nabbisen! Going to go ahead and merge since it's a small change.

PR #2288, opened 2021-09-14
Fix typo in deploy-a-nextjs-site.md

hanford:
I think my proposed changes are right, feel free to close if it's not!


lukeed:
Thank you, but it's correct as is. Instructions match those from the official Next.js site: https://nextjs.org/docs/api-reference/create-next-app

hanford:
My bad @lukeed!

PR #2316, opened 2021-09-16
Add capability for watching products directories and files

kevin-wynn:
## Why?

When working in individual product directories whenever we make changes to the `.md` files we have to stop our development server and restart to see changes. Or alternatively, make changes directly to the `.md` file in the `.docs` directory.

## What

This PR adds `concurrently` and `chokidar` to the root directory so that we can pass the product name to a develop script. This will kick off the `watcher.js` file that watches the product directory for any changes to the `.md` files and immediately copies the new version into the respective `.docs/src/content` directory. Gatsby will then see the file change and properly hot reload.

## How

After `npm install` and `npm run bootstrap` in a product directory. From the root directory, `npm install` and run `npm run develop -- $product` to start gatsby and the watch service


pedrosousa:
@kevin-wynn I updated the proposed docs. Can you please review before we merge?

develop.js
```diff
@@ -0,0 +1,33 @@
+const concurrently = require("concurrently");
+const { exec } = require("child_process");
+const args = process.argv.slice(2);
+const product = args[0];
+
+concurrently(
+  [
+    {
+      command: `products/${product}/node_modules/cloudflare-docs-engine/bin/commands.sh develop`,
```
pedrosousa: Nit: Had to add `bash ` before the current command to make this work on Windows (shebang lines in Bash scripts are not enough, unfortunately).

pedrosousa:
While testing this on Windows, I noticed that stopping the execution of the new command with `Ctrl+C` only stops `watch.js` — `gatsby` keeps running indefinitely.

Output:
image

Here's `node.exe` still running:
image

Not sure if this happens in other platforms, or if this is a Windows bug.

When I manually kill the `node.exe` process in Task Manager, I get this output:
image

kevin-wynn:
@pedrosousa I'll take a look at these issues today 👍

pedrosousa:
@kevin-wynn Tried again with your latest changes:
1) Starting the watcher and gatsby works!
2) I still can't stop all processes gracefully -- tried using Git Bash (shell included with the Git installation) and Cmder, a command-line prompt on steroids, but it didn't work. A couple of sub-processes do terminate, but the `node` ones that serve the docs and the GraphiQL tool keep running. Don't worry too much about this — it seems to be a Windows-only thing, and I can kill any remaining processes once I stop the new script.

pedrosousa (approved):

kevin-wynn:
@pedrosousa Interesting... I wonder if its a gastby thing? Do you ever run into issues with it not stopping on windows outside of the concurrently? I'd like to look back into this later but we can merge this in now and circle back to it. I'll put a ticket on my teams backlog for me to revisit later. But it might not come up since I know we want to overhaul how these docs get generated in general

pedrosousa:
@kevin-wynn No, I've never had issues with `Ctrl+C` not stopping gatsby's development server outside concurrently.

PR #2336, opened 2021-09-20
Clarify action execution with firewall rules having the same priority

pedrosousa:
Addresses https://jira.cfops.it/browse/PCX-973.


rboulton-cf (commented):
This isn't quite correct. The bypass action also works the same way - it doesn't terminate the processing of the request.

pedrosousa:
> This isn't quite correct. The bypass action also works the same way - it doesn't terminate the processing of the request.

Thank you for the information, @rboulton-cf. Besides mentioning the _Bypass_ action here, I've updated the "Support Actions" initial paragraph, which didn't match the information in the table.

rboulton-cf (approved):

pedrosousa:
@rboulton-cf Removed the sentence about log entries, as discussed.

marciocloudflare (approved):

PR #2338, opened 2021-09-20
[SSL] Additional options for edge certs

kodster28:
Trying to plug a hole in the docs where we don't have any references to a lot of the toggles in dev docs. Also, lays groundwork for porting over these KB articles as we have time by setting up partials.

I'm aware the formatting of the page isn't the best, but figure it's good as an interim solution and we can figure out formatting later.

Main page screenshot:
image


products/ssl/src/content/edge-certificates/universal-ssl/enable-universal-ssl.md
```diff
@@ -3,18 +3,58 @@ order: 1
...
     - Immediately if you add [Domain Control Validation (DCV)](../changing-dcv-method) records to your authoritative DNS.
-    - After a brief period of downtime if you **do not** add DCV records (once your traffic is proxied).
\ No newline at end of file
+    - After a brief period of downtime if you **do not** add DCV records (once your traffic is proxied).
+
+## Disable Universal SSL
+
+Some customers may need to manage their own SSL certificates or rely on specific Certificate Authorities.
+
+If you disable your domain's Universal SSL certificate, we remove that certificate from our network and will not order or renew any additional Universal SSL certificates.
```
dcpena: Should "we" be "Cloudflare?"

dcpena (approved):
One non-blocking comment

PR #2347, opened 2021-09-22
Fix SvelteKit pages docs to use ESM config

Deebster:
The [current instructions](https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-site#sveltekit-configuration) need updating for the current SvelteKit. The main reason is that SvelteKit changed the config to using ESM. See https://kit.svelte.dev/docs#configuration for the new style.

Also, the adapter-node is not included by default so cannot be removed.

This patch fixes both.

### DIFF
--- a/products/pages/src/content/framework-guides/deploy-a-svelte-site.md
+++ b/products/pages/src/content/framework-guides/deploy-a-svelte-site.md
@@ -79,28 +79,32 @@ Optionally, you can customize the **Project name** field. It defaults to the Git
 
 By default, SvelteKit prepares our project with the assumption that it will deployed to a Node.js server. This is not appropriate for this tutorial, but luckily SvelteKit is flexible and has a ready-made "adapter" for your needs. A few, easy changes have to be made.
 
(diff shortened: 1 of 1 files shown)


Deebster:
I would also like to move the "SvelteKit Configuration" section to before "Deploying with Cloudflare Pages" since it's not obvious that you can't deploy after reading the "​Deploying with Cloudflare Pages" section.

lukeed (approved):
Thanks!

products/pages/src/content/framework-guides/deploy-a-svelte-site.md
```diff
@@ -79,7 +79,7 @@ Optionally, you can customize the **Project name** field. It defaults to the Git
 
 By default, SvelteKit prepares our project with the assumption that it will deployed to a Node.js server. This is not appropriate for this tutorial, but luckily SvelteKit is flexible and has a ready-made "adapter" for your needs. A few, easy changes have to be made.
 
-First, install the [`@sveltejs/adapter-static`](https://www.npmjs.com/package/@sveltejs/adapter-static) package instead:
```
Deebster: 🤦‍♂️

lukeed: happens :)

PR #2392, opened 2021-09-28
Downgrade to node v12 to be in sync with docs-engine

### LINKED ISSUE #2391: Github actions using different node version than docs engine
We can see that our build.yaml specifies node version `14.2.0` https://github.com/cloudflare/cloudflare-docs/blob/production/.github/workflows/build.yml#L168

However, the cloudflare docs engine is potentially using 12x https://github.com/cloudflare/cloudflare-docs-engine/blob/master/package.json#L52

In attempt to fix our current build failures, I'm lowering the node engine down to 12x

jrf0110:
Closes #2391

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -165,7 +165,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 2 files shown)


jrf0110:
Unneeded

Issue #2395, opened 2021-09-28
IPFS SSL Input Section has been removed.


pagreczner:
### Expected Behavior
https://web.archive.org/web/20210922160400/https://www.cloudflare.com/distributed-web-gateway/
I expect there to be an input field at the bottom of this page that allows me to register ssl for a cloudflare dns ipfs link.

### Actual Behavior
https://www.cloudflare.com/distributed-web-gateway/
The above input field referenced in the archive link is no longer there. But the docs still point to this page: https://developers.cloudflare.com/distributed-web/ipfs-gateway/connecting-website#make-it-all-secure
```
Cloudflare will issue you a free SSL certificate for your.website, which allows users to load https://your.website. All you have to do is go to cloudflare-ipfs.com and scroll to the bottom, where it says "Connecting Your Website." Enter your domain into the text box and click Submit.
```

### Section that requires update
The above snippet requires update or the page it is pointing too needs to have the input field added back.

### Additional information
n/a

haleycode:
This was resolved in the linked PR.

PR #2397, opened 2021-09-28
Update DWeb gateway X.509 provisioning URLs

thibmeu:
We are making changes to the way X.509 certificates are provisioned on the IPFS and Ethereum gateway. In the meantime, we want to point people to ask-research@ which is a link that works.


Lekensteyn (approved):

thibmeu:
Fixes #2395

PR #2399, opened 2021-09-28
CNAME manual validation is actually a post method

eco-JoshuaJackson:
Documentation said that it should be a patch but actual method mirrors the others and is a post.


kodster28 (approved):

kodster28:
Thanks for the catch, @JoshuaJackson-jobvite!

PR #2401, opened 2021-09-28
Update traffic-steering.md

kcantrel:
Just fixed a typo.


kodster28:
Thanks @kcantrel, but looks like someone else also caught this in #2403. Closing this out since the typo was previously adjusted.

Issue #1601, opened 2021-06-17
export-graphql-to-csv doesn't work


codesse:
### Expected Behavior
https://github.com/cloudflare/cloudflare-docs/blob/production/products/analytics/src/content/graphql-api/tutorials/export-graphql-to-csv/index.md

### Actual Behavior
After inserting the values in the 'complete script' : 
api_email = '[your email here]'
api_token = '[your API token here]'
api_account = '[your account ID here]'

I try the execute the program and get : 

`ValueError: Invalid header name b'Authorization: Bearer'`

It seems that ':' isn't liked by the 'requests' module. If I delete the ':' then I get : 

`{"success":false,"errors":[{"code":9106,"message":"Missing X-Auth-Key, X-Auth-Email or Authorization headers"}]}`

### Section that requires update
https://github.com/cloudflare/cloudflare-docs/blob/production/products/analytics/src/content/graphql-api/tutorials/export-graphql-to-csv/index.md

### Additional information
No more info needed

nnmalex:
Fixing the authentication issue in this tutorial leaves the query still not working with `KeyError: 'viewer'` error.

victor-perov:
Thanks for the report, I will have a look.

victor-perov:
I copied the script from **The final script** [part](https://github.com/cloudflare/cloudflare-docs/blob/65db1d09aa4c53b6501125b7764603fa0a60817c/content/analytics/graphql-api/tutorials/export-graphql-to-csv/index.md?plain=1#L229-L360) of the related tutorial as gql2csv.py (see below), replaced creds, and was able to successfully execute it (Darwin, was also able to run it on Debian):
```zh
viktor@machine cloudflare-docs % python3 gql2csv.py 
Successfully exported to network-analytics-2022-08-01-2022-08-08.csv
```

However, I can imagine the script does not catch all errors therefore `KeyError: 'viewer'` might appear. I will check that case later.

victor-perov:
I can't reproduce this issue. Please let us know if you are still seeing it.

PR #2420, opened 2021-09-30
[analytics] fix header to be "Authorization: Bearer {key}" #1601

westonplatter:


victor-perov:
Hey, @westonplatter! Thanks for your contribution and your willingness to make our docs better. I really appreciate it!

We have two different ways to authenticate the request:

1. via  `Authorization` header, if you have a pre-generated API Token with analytics:read permission,
2. by providing  `X-Auth-Key` and `X-Auth-Email`.

The example in this file doesn't describe it, and also confuses me with providing both `X-Auth-Email` and `Authorization`. I think the problem is the documentation could be misleading. The best way would be to describe this point and provide two different sets of headers.

I will come back with my version of fixing this issue.

Once again, thanks for your finding and proposed remediation!

victor-perov:
Please see #2494

westonplatter:
@victor-perov thanks! I see the code changes show up in your PR so closing this one.

PR #2422, opened 2021-10-01
Update tunnel limits

nmldiegues:


nmldiegues:
This is not yet live, but will be soon. No harm in having the limits documented before we're live with it.

abracchi-tw (approved):

PR #2444, opened 2021-10-05
Updated POST command

avattuonecf:
Correct command to use auth token:

$ export GITHUB_URL=$(curl -H "Authorization: token $GITHUB_TOKEN" -d '{"name": "cf-config", "private": true}' "https://api.github.com/user/repos" 2> /dev/null | jq -r .ssh_url)


kodster28 (approved):

kodster28:
Thanks for the fix, @avattuonecf! Going to go ahead and merge this in.

PR #2445, opened 2021-10-05
2021 icon update part 1

gganch:
Part 1 of the process of moving to a new way to display and use icons.

In this PR: Still using the old process of requiring in and using cloudflare-brand-assets but with the new svgs included in all of the product folders in order to prep for step 2

After merging this piece of code, part 2 will be to merge in new code for clooudflare-docs-engine.

Part 3, 4 will be going back to the 2 repos and removing the old way of displaying icons and adding in the new way officially.


pedrosousa:
@gganch 
I'm not sure when this PR is moving forward, but we have some new documentation tiles that don't have an SVG file. This PR will need to be updated with the latest `production` changes and you must add the missing SVG files.

kodster28:
@gganch, I believe this is outdated and am going to close. Feel free to re-open if needed!

PR #2446, opened 2021-10-05
update product icons, implement new process to use icons

gganch:
Part 1 of the process of moving to a new way to display and use icons.

In this PR: Still using the old process of requiring in and using cloudflare-brand-assets but with the new svgs included in all of the product folders in order to prep for step 2

After merging this piece of code, part 2 will be to merge in new code for clooudflare-docs-engine.

Part 3, 4 will be going back to the 2 repos and removing the old way of displaying icons and adding in the new way officially.


kodster28 (approved):
Looks good from PCX's point of view. 

Tested two product folders + the main developers page. Also tested deleting and re-running `npm install` and `npm run bootstrap` prior to serving locally, since that was an issue we ran into last time.

PR #2461, opened 2021-10-06
[WR] Additional functionality

kodster28:
Addresses PCX-824

### DIFF
--- a/products/waiting-room/src/content/additional-options/create-events.md
+++ b/products/waiting-room/src/content/additional-options/create-events.md
@@ -49,8 +49,8 @@ Users who reach your application **during the prequeue period** are [randomly as
 
 To set up a "lottery", include the [following parameters](#properties) in your API request:
 
-- `prequeue_start_time`: ISO 8601 timestamp that marks when to begin queueing all users before the event starts. Must occur at least **5 minutes before** `event_start_time`.
-- `shuffle_at_event_start`: If **true** and `prequeue_start_time` is not null, users in the prequeue will be shuffled randomly at the `event_start_time`
+- `prequeue_start_time`
+- `shuffle_at_event_start`
 
 ## Preview an event configuration
 
--- a/products/waiting-room/src/content/additional-options/create-events.md
+++ b/products/waiting-room/src/content/additional-options/create-events.md
@@ -5,7 +5,7 @@ pcx-content-type: how-to
 
 # Create scheduled events
 
-When you want to customize the behavior of a waiting room — updating the queuing method, total active users, or any [other property](/reference/configuration-settings) — for a specific period of time, set up a scheduled event.
+When you want to customize the behavior of a waiting room for a specific period of time — queuing method, total active users, or any [other property](/reference/configuration-settings) — set up a scheduled event.
 
 Any properties set on the event will override the default property on the waiting room for the duration of the event.
 
@@ -17,15 +17,11 @@ Only certain customers can support scheduled events with their waiting rooms. Fo
 
 ## Create an event
 
-At the moment, you can only create an event [using the API](https://api.cloudflare.com/#waiting-room-create-event).
+At the moment, you can only create an event [using the API](https://api.cloudflare.com/#waiting-room-create-event). Any properties set on the event will override the default property on the waiting room for the duration of the event.
 
-Any properties set on the event will override the default property on the waiting room for the duration of the event.
+If you create an event, you also have access to additional fields for [creating a custom template](../customize-waiting-room#custom-waiting-room).
 
-<Aside type="note">
-
-A waiting room cannot support overlapping events.
-
-</Aside>
+A waiting room cannot support multiple events happening at the same time. To create overlapping events, use multiple waiting rooms.
 
 ### Unique properties
 
@@ -33,23 +29,38 @@ Though most properties are identical to those on a [regular waiting room](https:
 
 - `event_start_time` (required): ISO 8601 timestamp that marks the start of the event. At this time, queued users will be processed with the event's configuration. Must occur at least 1 minute before `event_end_time`.
 - `event_end_time` (required): ISO 8601 timestamp that marks the end of the event.
-- `shuffle_at_event_start`: If `true` and `prequeue_start_time` is not null, users in the prequeue will be shuffled randomly at the `event_start_time`. Commonly used to ensure fairness if your event is using a [**FIFO** queueing method](/reference/queueing-methods#first-in-first-out-fifo).
-- `prequeue_start_time`: ISO 8601 timestamp that marks when to begin queueing all users before the event starts. Must occur at least 5 minutes before `event_start_time`.
-- `suspended`: If `true`, the event is ignored and traffic is handled based on the regular configuration.
+- `shuffle_at_event_start`: If **true** and `prequeue_start_time` is not null, users in the prequeue will be shuffled randomly at the `event_start_time`. Commonly used to ensure fairness if your event is using a [**FIFO** queueing method](#set-up-a-lottery).
+- `prequeue_start_time`: ISO 8601 timestamp that marks when to begin queueing all users before the event starts. Must occur at least **5 minutes before** `event_start_time`.
+- `suspended`: If **true**, the event is ignored and traffic is handled based on the waiting room's normal configuration.
+
+## Set up a "lottery"
+
+Set up a "lottery" system to reward all users who enter into the queue prior to your event start time. 
+
+Users who reach your application **during the prequeue** period are [randomly assigned](/reference/queueing-methods#random) a place in line when the event starts. Users who reach your application **after the prequeue period** but still during the event are assigned places based on [FIFO ordering](/reference/queueing-methods##first-in-first-out-fifo), meaning they will be behind anyone from the prequeue.
+
+To set up a "lottery", include the [following parameters](#unique-properties) in your API request:
+
+- `prequeue_start_time`
+- `shuffle_at_event_start`
 
 ## Preview an event configuration
 
-Since properties set on an event will override the default property of a waiting room for the duration of an event, you can also use the API to [preview an event configuration](https://api.cloudflare.com/#waiting-room-preview-active-event-details).
+Since properties set on an event will override the default property of a waiting room for the duration of an event, you should use the API to [preview an event configuration](https://api.cloudflare.com/#waiting-room-preview-active-event-details) before it begins.
 
 This command shows you the event's configuration as if it were active, meaning that inherited fields from the waiting room will display as their current values.
-
-Use this command to review your event before it reaches its **Start Time**.
 
 ## Edit an event
 
 To edit an event, use a [PATCH request](https://api.cloudflare.com/#waiting-room-patch-event).
 
-By default, you cannot change the [**queueing method**](/reference/queueing-methods) within five minutes of the `prequeue_start_time` or the `event_start_time`. You have to update these values first and then update the queueing method.
+By default, you cannot change the [queueing method](/reference/queueing-methods) **within five minutes** of the `prequeue_start_time` or the `event_start_time`. You have to update these values first and then update the queueing method.
+
+## Disable events
+
+You can disable an event by setting the `suspended` parameter to `true`.
+
+Additionally, events will not become active if a waiting room itself is **Disabled** or if the waiting room has been set to **Queue All**. 
 
 ## Other API commands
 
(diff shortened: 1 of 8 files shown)


products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
+---
+order: 3
+pcx-content-type: how-to
+---
+
+# Create scheduled events
+
+When you want to customize the behavior of a waiting room for a specific period of time — queuing method, total active users, or any [other property](/reference/configuration-settings) — set up a scheduled event.
```
tc80: I'd change `queuing` to `queueing`.

Also, there are some properties we currently do not support, such as cookie_attributes, so maybe it is not best to say "any other property". How about something like:

`When you want to customize the behavior of a waiting room for a specific period of time, such as using a particular queueing method or increasing the total active users, you can schedule an event.`

tc80: Also, the section that includes the waiting room properties might need to be updated soon as well — we have a lot of new properties! 😄

tc80: Also, you say 

```
Any properties set on the event will override the default property on the waiting room for the duration of the event.
```

This is only true for the fields:
```
"new_users_per_minute",
"total_active_users",
"session_duration",
"custom_page_html",
"disable_session_renewal",
"queueing_method"
```

There are other fields that are shared by waiting rooms and events but are unique to each... for example "name", "description", or "suspended". If you suspend an event, you are not suspending a waiting room, just ignoring this particular event.

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+
+</Aside>
+
+## Create an event
+
+At the moment, you can only create an event using the API.
+
+To create an event:
+
+1. (optional) Depending on the type of event, update the [queueing method](/reference/queueing-methods#event-specific-queueing-methods) for your waiting room.
```
tc80: I'm happy to leave this as-is, but just keep in mind that any of these fields can be overrided, not just queueing_method:

```
"new_users_per_minute",
"total_active_users",
"session_duration",
"custom_page_html",
"disable_session_renewal",
"queueing_method"
```

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+
+<Aside type="note">
+
+If you need to create overlapping events, use different waiting rooms.
+
+</Aside>
+
+### Unique properties
+
+Though most properties are identical to those on a [regular waiting room](https://api.cloudflare.com/#waiting-room-properties), there are a few unique to [creating an event](https://api.cloudflare.com/#waiting-room-create-event):
```
tc80, line 34: Like I mentioned in another comment, `suspended` is a field that is present in both the waiting room and events, but it has different meaning when enabled for an event.

Maybe it is best to be explicit about all of the event properties besides the ones shared with the waiting room... these properties would also include "name" and "description". Note that "name" is a required field (that must be unique), just like how it is required (and unique) when creating a waiting room

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+- `event_end_time` (required): ISO 8601 timestamp that marks the end of the event.
+- `shuffle_at_event_start`: If **true** and `prequeue_start_time` is not null, users in the prequeue will be shuffled randomly at the `event_start_time`. Commonly used to ensure fairness if your event is using a [**FIFO** queueing method](#set-up-a-lottery).
+- `prequeue_start_time`: ISO 8601 timestamp that marks when to begin queueing all users before the event starts. Must occur at least **5 minutes before** `event_start_time`.
+- `suspended`: If **true**, the event is ignored and traffic is handled based on the waiting room's normal configuration.
+
+## Set up a "lottery"
+
+Set up a "lottery" system to reward all users who enter into the queue prior to your event start time. 
+
+Users who reach your application **during the prequeue** period are [randomly assigned](/reference/queueing-methods#random) a place in line when the event starts. Users who reach your application **after the prequeue period** but still during the event are assigned places based on [FIFO ordering](/reference/queueing-methods##first-in-first-out-fifo), meaning they will be behind anyone from the prequeue.
```
tc80: change to **during the prequeue period** for consistency

tc80: The sentence:
```
Users who reach your application after the prequeue period but still during the event are assigned places based on FIFO ordering, meaning they will be behind anyone from the prequeue.
```
Is not correct because you can have `shuffle_at_event_start` enabled but use a queueing_method other than fifo. It doesn't really make sense to use a method other than FIFO, so we strongly recommend that FIFO is used so that ordering can be enforced. However, it is not required... maybe a better explanation would be:
```
If the event uses FIFO ordering, users who reach your application after the prequeue period are assigned places after the users from the prequeue.
```

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+
+To set up a "lottery", include the [following parameters](#unique-properties) in your API request:
+
+- `prequeue_start_time`
```
tc80, line 52: For consistency, maybe add descriptions here

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+- `shuffle_at_event_start`
+
+## Preview an event configuration
+
+Since properties set on an event will override the default property of a waiting room for the duration of an event, you should use the API to [preview an event configuration](https://api.cloudflare.com/#waiting-room-preview-active-event-details) before it begins.
```
tc80: Maybe change to 

--> Since *some* properties...
--> display *their* current ...

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+
+This command shows you the event's configuration as if it were active, meaning that inherited fields from the waiting room will display as their current values.
+
+## Edit an event
+
+To edit an event, use a [PATCH request](https://api.cloudflare.com/#waiting-room-patch-event).
+
+By default, you cannot change the [queueing method](/reference/queueing-methods) **within five minutes** of the `prequeue_start_time` or the `event_start_time`. You have to update these values first and then update the queueing method.
```
tc80: This isn't true anymore. You can edit an event whenever you want — it doesn't matter if it is in the past, current occurring, or in the future

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Disable events
+
+You can disable an event by setting the `suspended` parameter to `true`.
```
tc80: `setting the` => `setting its`

(just be clear since waiting room has `suspended` too)

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,79 @@
...
+
+Additionally, events will not become active if a waiting room itself is **Disabled** or if the waiting room has been set to **Queue All**.
```
tc80: if queue all is set, an event is still active, but no users will be let in (they will just see a waiting room). As a result, I'd remove that last bit of the statement.

For example, a customer could put in their template:
```
{{queueAll}}
   {{isEventActive}}An event is active during queue all!{{/isEventActive}}
{{/queueAll}}
```
which would trigger if an event is active during queue all

products/waiting-room/src/content/reference/queueing-methods.md
```diff
@@ -32,6 +32,24 @@ When your application has open spots, your waiting room chooses visitors at rand
 
 Use this method when you want to distribute products or services more equitably. Earlier users have a better chance of exiting the waiting room before the estimated wait time because they have more chances to be selected.
 
+## Event-specific queueing methods
```
tc80: These methods will be for waiting rooms too.

You can mention that these are really useful as events, but we shouldn't say they are event-only.

products/waiting-room/src/content/reference/queueing-methods.md
```diff
@@ -32,6 +32,24 @@ When your application has open spots, your waiting room chooses visitors at rand
...
+
+<Aside type="note" header="Note:">
+
+If your waiting room is set to **Queue All**, all visitors will be queued, regardless of the chosen method.
+
+</Aside>
+
+### Passthrough
+
+Allow all traffic to pass through your waiting room and into your application by setting `queueing_method` to **passthrough**. Use this setup when you only want to use your waiting room for events — where you can update the queueing method — and otherwise avoid queuing during low-traffic hours.
```
tc80: Change `queuing` => `queueing`

Maybe be even more clear with:
```
Allow all traffic to pass immediately through your waiting room and into your application by setting `queueing_method` to **passthrough**. If **Queue All** is enabled or an event is prequeueing, users will be queued in the waiting room.
```

products/waiting-room/src/content/reference/queueing-methods.md
```diff
@@ -32,6 +32,24 @@ When your application has open spots, your waiting room chooses visitors at rand
...
+
+### Reject
+
+Prevent any traffic from reaching your application unless an event is active by setting `queueing_method` to **reject**. Use this setup for event-only endpoints or to perform application maintenance.
```
tc80: `reject` is used to reject all traffic — so basically if this is enabled then you will see the waiting room page 100% of the time. It is basically the opposite of `passthrough`.

When Queue All is enabled or an event is prequeueing, (I believe) users will still be rejected (this is still in review) and the page will not refresh....

**NOTE** the behavior for passthrough and reject is not completely finalized/released, so I'd wait to release this bit until the cf-api docs are updated

products/waiting-room/src/content/additional-options/create-events.md
kodster28: Gotcha! I actually deleted this step as it was a reference to `what I thought were` event-only queueing methods.

products/waiting-room/src/content/reference/queueing-methods.md
kodster28: Raising this piece up to BB.

kodster28:
Howdy @tc80, I think I've addressed all your changes. Let me know if I missed anything!

dcpena (approved):

products/waiting-room/src/content/additional-options/create-events.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Set up a "lottery"
+
+Set up a "lottery" system to reward all users who enter into the queue prior to your event start time. 
+
+Users who reach your application **during the prequeue period** are [randomly assigned](/reference/queueing-methods#random) a place in line when the event starts. If the event uses [FIFO ordering](/reference/queueing-methods#first-in-first-out-fifo), users who reach your application **after the prequeue period** are assigned places after users from the prequeue.
+
+To set up a "lottery", include the [following parameters](#properties) in your API request:
+
+- `prequeue_start_time`: ISO 8601 timestamp that marks when to begin queueing all users before the event starts. Must occur at least **5 minutes before** `event_start_time`.
```
tc80: Sorry, I realize now that these fields are previously explained. I think it's fine to leave their descriptions here, but feel free to remove them if you don't want the redundancy 🤷 

(Sorry for the hypocrisy, I know I made you add these in!)

products/waiting-room/src/content/reference/queueing-methods.md
```diff
@@ -32,6 +32,28 @@ When your application has open spots, your waiting room chooses visitors at rand
...
+
+### Passthrough
+
+Allow all traffic to pass immediately through your waiting room and into your application by setting its `queueing_method` to **passthrough**. If **Queue All** is enabled or an event is prequeueing, users will be queued in the waiting room.
+
+Use this setup when you only want to use your waiting room for events — where you can update the queueing method — and otherwise avoid queueing during low-traffic hours.
+
+### Reject
+
+Prevent any traffic from reaching your application by setting its `queueing_method` to **reject**. If **Queue All** is enabled or an event is prequeueing, users will be queued in the waiting room (but never reach your application).
```
tc80: We just confirmed this behavior to actually `reject` users during queue all / prequeue. You can change the description to:

```
Prevent any traffic from reaching your application by setting its `queueing_method` to **reject**. If **Queue All** is enabled or an event is prequeueing, users will be **still be rejected** and see a static web page that does not refresh.
```

PR #2481, opened 2021-10-12
Call a space a space

neerolyte:
I don't see any reason to describe `%20`, which is just a space as "white".


kodster28 (approved):

kodster28:
Thanks for the find, @neerolyte. Going ahead and merging it now.

PR #2545, opened 2021-10-21
Update all Workers examples to esmodule format

jkup:
Here are the examples we have so far. We still need to finish a few of them.

cc @lukeed @signalnerve


lukeed (approved):
Looks good! For others passing by: this is the TS -> JS output from another repo we've working on. That JS (shown here) is passed thru `prettier` and so even though the code style _is_ different than what was here originally, it _is_ consistent. Also, the loss of newlines is coming from the TS-to-JS conversion step.

lukeed (approved):
🎉

tobsn:
Hey! Just added the /read-post.md example into a worker and get the following for the export default at the bottom:

`Uncaught__ SyntaxError: Unexpected token 'export' at line 64 (Code: 10021)`

ref: https://developers.cloudflare.com/workers/examples/read-post

lukeed:
Hello, this worker is in the Module Worker format, which utilizes the ESM syntax. Please update your wrangler config so that [it matches this](https://developers.cloudflare.com/workers/cli-wrangler/configuration#modules)

tobsn:
@lukeed simple worker, I don't use any local tools packaging tools, I work directly off of testing simple scripts locally with miniflare and then copy paste (they are very small projects for private) - I think for the example section there should be a vanilla version avail?

lukeed:
The Cloudflare dashboard's editor doesn't currently support Module Worker syntax but is coming very soon. However, initial support is already available in the `cloudflareworkers.com` playground & the example can be seen [here](https://cloudflareworkers.com/#323b15f2fd1efa91a487adc6d76262c6:https://tutorial.cloudflareworkers.com/).

The examples have migrated to the Module Worker format because, going forward, that's the preferred way to author Workers. Service Worker syntax (using `addEventListener('fetch', ...)`) will continue being supported but not encouraged as the primary format.

You may traverse the git history for the old example code or [visit the docs](https://developers.cloudflare.com/workers/runtime-apis/fetch-event) to learn how to go to-and-from the different formats. Or ask questions here/in Discord – happy to help!

tobsn:
@lukeed sorry, now I get it!
I did not realize the addEventListener is missing in the example ;)

Thanks!

PR #2562, opened 2021-10-25
await request.text()

pew:


lukeed:
Hey, so this is actually not necessary and was removed a part of #2258 
Since there's no try/catch here inside this function, there's no benefit to `await`ing here.

Thank you though!

PR #2565, opened 2021-10-25
[Cache] fix vary_header_present description in common-cf-published-statuses.md

cloudsucked:
- Fix the logic in the description of the vary_header_present Polish status
- Only responses where there is no Vary header, or when there is a Vary header with ONLY accept-encoding will be processed by Polish.


dcpena:
@cloudsucked thanks for the contribution! This section actually just got moved to the Knowledgebase, so I'm going to close this PR, but I did make the update for you in the KB. [Here's a link](https://support.cloudflare.com/hc/en-us/articles/4412244347917-Troubleshoot-common-Cf-Polished-statuses) to the article for reference.

PR #2591, opened 2021-10-28
Use zone endpoint instead of account for L7 DDoS MR overrides

pedrosousa:
Update example to use the correct endpoint (zone) for HTTP DDoS Attack Protection Managed Ruleset overrides.


marciocloudflare (approved):

jdesgats (approved):
Looks good. Thanks!

PR #2596, opened 2021-10-28
Adding a secondary note to better describe the example parameters

cdrubin:
In the user interface of the load balancer page (`/traffic/load-balancing`) the suffix of a load balancer name is never shown even though that column is titled 'hostname'. In the CF API call to list load balancers the zone suffix *is* included in the "name" key, and to get the command on this page to work the full hostname is required. For example, for load balancer "hello" in zone "you.com", we need to use "hello.you.com" as the load balancer name in the `cloudflared tunnel route lb <tunnel ID or NAME> <load balancer name> <load balancer pool>` for it to work.


ranbel:
addressed in #5448

PR #2599, opened 2021-10-28
Add canonical tags across pages

adesege:
# Why?
Google sees www.developers.cloudflare.com/bots/about and www.developers.cloudflare.com/bots/about/ as two different pages which can affect the page rank.

# What?
This PR adds canonical tags across pages for urls ending with slash so as to improve our page rank. Query strings from urls are also stripped. `gatsby-plugin-canonical-urls` is used to achieve this.

# How?
1. cd into `developers.cloudflare.com`
1. Install dependencies and run the project using `npm run start`.
1. Visit your browser and you should see a link rel="canonical" tag in the devtools source code.
1. To test other products, pull the changes from this PR https://github.com/cloudflare/cloudflare-docs-engine/pull/365
1. Point `cloudflare-docs-engine` dependency in `products/images/package.json` to the local version of https://github.com/cloudflare/cloudflare-docs-engine/pull/365.
1. Run `npm install` then `npm run bootstrap`.
1. Run `npm run develop` to build and start the application.
1. Visit http://localhost:8000/cloudflare-images/resize-images/ and check the source code for the appropriate link rel="canonical" tag.
1. Repeat `steps 5-8`. For step 8, remember to change the url to point to the appropriate product you are testing for.


KimJ15:
Hi there! We actually have an internal ticket for this work so i'm going to close this PR as a duplicate. Thanks!

kevin-wynn (approved):

kodster28:
Howdy @kevin-wynn, thanks for this! I should be able to get in and test this late next week.

Sorry that it took me a bit to find this!

kodster28:
From my local testing, noticed the following issues:
1. When I'm on a product landing page (`https://developers.cloudflare.com/ssl/`), the canonical tag says `https://developers.cloudflare.com/`). Would that be updated when we deploy it live?

Screen Shot 2021-11-22 at 4 51 25 PM

2. When I'm on a page within a product, the canonical tag just says whatever the url of the page itself is. My understanding is that we'd always want the canonical tag to use the slashed version of the page (unless we're planning on removing the non-slashed versions entirely and setting up redirects?
Screen Shot 2021-11-22 at 4 52 06 PM
)

Screen Shot 2021-11-22 at 4 50 57 PM
.

kodster28 (changes requested):
See the comment above about observed behavior.

adesege:
Thanks for the feedback, @kodster28.

1. To test this locally, you will need to install the latest version of cloudflare-docs-engine from github then point the `cloudflare-docs-engine` in the product you want to test to the locally cloned version. You can also run `npm install` instead of cloning cloudflare-docs-engine locally. 
2. In prod, non-slashed urls will be redirected to their slashed counterpart. For example, `/get-started` will always be `/get-started/`.

Thanks

pedrosousa:
Since we're also using `gatsby-plugin-remove-trailing-slashes`, only the root page in a product tile will have a canonical URL ending with a slash. Fortunately, this matches what users will see in the browser address bar.

Example (from the live docs, since the PR in the `cloudflare-docs-engine` repo was already merged):

* https://developers.cloudflare.com/1.1.1.1/ - Canonical tag:
`<link rel="canonical" href="https://developers.cloudflare.com/1.1.1.1/" data-baseprotocol="https:" data-basehost="developers.cloudflare.com">`

* https://developers.cloudflare.com/1.1.1.1/1.1.1.1-for-families - Canonical tag:
`<link rel="canonical" href="https://developers.cloudflare.com/1.1.1.1/1.1.1.1-for-families" data-baseprotocol="https:" data-basehost="developers.cloudflare.com">`

adesege:
It appears some product pages do not redirect to a path with slash. Should we remove `gatsby-plugin-remove-trailing-slashes` and set up redirects so that we are always redirected to the path with trailing slash?

Or I should just append trailing slash to all canonical urls?

kodster28:
From looking at the [comments on the ticket](https://jira.cfops.it/browse/MRK-13498), it looks like the best approach would be to:
1. Set up canonical tags pointing to the slashed versions.
2. Remove the `gatsby-plugin-remove-trailing-slashes`.
3. Remove all non-slashed pages in general (not sure if this is covered by ^^^).
4. Set up redirects so we'll always be directed to the page with the trailing slash.

adesege:
That's correct, Kody. Number 3 will be solved by Number 4 since 301 redirect means the page has permanently moved to the new url and search engines will respect that.

I will work on the changes as soon as I'm done with my current task.

PR #2614, opened 2021-11-02
Remove meetup.com 404s

s-crypt:
Only the London link leads to an active group, all the rest say "this group doesn't exist". Removed the dead links.


kodster28 (approved):

kodster28:
Thanks for the catch, @s-crypt. Going ahead and merging this in.

PR #2621, opened 2021-11-02
MFW-458: Add Rules List example for Magic Firewall

arges:


tejaswi-cf (approved):

dcpena:
This reads to me as more of a procedure/how to use account-level rules lists, so I'd suggest adding this under the How to section. Content looks good overall though!

arges:
> This reads to me as more of a procedure/how to use account-level rules lists, so I'd suggest adding this under the How to section. Content looks good overall though!

@dcpena addressed comments

dcpena (approved):

PR #2622, opened 2021-11-02
[Workers] Fix handler parameters in module workers examples

mrbbot:
According to https://developers.cloudflare.com/workers/runtime-apis/fetch-event#syntax-module-worker and https://developers.cloudflare.com/workers/runtime-apis/scheduled-event, the signature for `fetch` and `scheduled` handlers is `fetch(request, env, ctx) { ... }` and `scheduled(controller, env, ctx) { ... }`.

This PR fixes the handler signatures used by some of the new module worker examples, so `ctx` is correct in each case.

### DIFF
--- a/products/workers/src/content/examples/cache-api.md
+++ b/products/workers/src/content/examples/cache-api.md
@@ -42,7 +42,7 @@ async function handleRequest(request, ctx) {
   return response;
 }
 export default {
-  fetch(request, ctx) {
+  fetch(request, env, ctx) {
     try {
       return handleRequest(request, ctx);
     } catch (e) {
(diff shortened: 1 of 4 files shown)


Electroid (approved):

deadlypants1973 (approved):

lukeed (approved):

lukeed:
Approved but waiting for the examples to move back to Module format before fixing conflicts here & merging

lukeed:
The examples are currently deployed using the SW syntax ([Cache API example](https://developers.cloudflare.com/workers/examples/cache-api/)) and so trying to resolve the conflicts here (post Hugo migration) isn't going to be applicable anymore anyway.

We have these changes reflected in the Module Worker examples ready to go though.
Thanks~!

PR #2630, opened 2021-11-03
Adding a note for private network/warp routing

kennyj42:


abelinkinbio (approved):

abracchi-tw:
@kennyj42 @abelinkinbio I dont think we need three separate scenarios there. We should just provide a snippet for `application` and one for `network`. I updated the one for network to include the `warp-routing` rule.

abracchi-tw (approved):

PR #2658, opened 2021-11-08
[Cache] Updates to Early Hints

dcpena:
Update to Early Hints that are covered in [PCX-2896](https://jira.cfops.it/browse/PCX-2896).

### DIFF
--- a/products/cache/src/content/about/early-hints.md
+++ b/products/cache/src/content/about/early-hints.md
@@ -21,11 +21,17 @@ After you sign up, you will be added to a list of beta users, and the feature wi
 
 ## Generating Early Hints
 
-Early Hints are only generated and cached for URL paths on responses that:
+Early Hints are only generated and cached:
 
-- Return 200, 301, or 302
-- Have a `Content-Type: text/html` header
-- Have a Link response header specifying links with preconnect or preload `rel` types, such as `Link: </img/preloaded.png>; rel=preload`
+- For URIs with `.html`, `.htm`, or `.php` file extensions, or no file extension
+- On 200, 301, or 302 response return codes
+- When the response contains [link headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link) with preconnect or preload rel types, such as `Link: </img/preloaded.png>; rel=preload`
+
+<Aside type="note">
+
+Early Hints cache entries are keyed by request URI and ignore query strings.
+
+</Aside>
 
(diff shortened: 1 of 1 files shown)


products/cache/src/content/about/early-hints.md
```diff
@@ -21,20 +21,26 @@ After you sign up, you will be added to a list of beta users, and the feature wi
...
 Currently, only certain browser versions will take action to preload or preconnect on receiving Early Hints, such as Google Chrome M94 and higher. Instructions for running WebPageTest to experiment with compatible client browsers can be found in the [blog post](https://blog.cloudflare.com/early-hints/#testing-early-hints-with-web-page-test).
 
-<Aside type="note" >
-
-Early Hints responses are emitted prior to reaching the origin server or Worker. Note that if a page on your zone requires authenticated access, enabling Early Hints can serve a 103 response with previously cached Link headers to an unauthenticated visitor ahead of a 403 Forbidden response from your origin.
+Additionally, keep the following in mind:
 
-</Aside>
\ No newline at end of file
+- Early Hints responses may be emitted before reaching the origin server or Worker. If a page on your zone requires authenticated access, enabling Early Hints can serve a 103 response with previously cached Link headers to an unauthenticated visitor ahead of a 403 Forbidden response from your origin.
```
nayaabahsan: The second sentence here is hard to follow/understand. I don't have a suggestion because I don't fully understand what is being said, but I think it should be broken up.

nayaabahsan (approved):

products/cache/src/content/about/early-hints.md
dcpena: Is this a little better? I tried shifting the subject

>If pages on your website require user authentication, unauthenticated visitors may receive a 103 response with previously cached Link headers before they receive a 403 Forbidden response from the origin when Early Hints is enabled.

nayaabahsan: It is better, but I think the sentence just needs to be broken up into separate sentences. I don't know if this makes technical sense, but maybe more like:

>When Early Hints is enabled and pages on your site require authentication, unauthenticated visitors may receive a 103 response. This 103 response would be emitted with previously cached Link headers before a 403 Forbidden response from your origin.

dcpena: @nayaabahsan omg, teamwork! thanks for finding a spot to break up the sentence. I'll make that change now

PR #2664, opened 2021-11-09
Add startTime to player docs

third774:
cc @zaidf

### DIFF
--- a/products/stream/src/content/viewing-videos/using-the-stream-player.md
+++ b/products/stream/src/content/viewing-videos/using-the-stream-player.md
@@ -111,4 +111,8 @@ Player options are configured with querystring parameters in the iframe's `src` 
 
   - The video id from the video you've uploaded to Cloudflare Stream should be included here.
 
+- `startTime`
+
+  - A timestamp that specifies the time when playback begins. If a plain number is used such as `?startTime=123`, it will be interepreted as `123` seconds. More human readable timestamps can also be used, such as `?startTime=1h12m27s` for `1 hour, 12 minutes, and 27 seconds`.
+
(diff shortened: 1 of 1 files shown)


products/stream/src/content/viewing-videos/using-the-stream-player.md
```diff
@@ -111,4 +111,8 @@ Player options are configured with querystring parameters in the iframe's `src`
 
   - The video id from the video you've uploaded to Cloudflare Stream should be included here.
 
+- `startTime`
+
+  - A timestamp that the video will start at. If a plain number is used such as `?startTime=123`, it will be interepreted as `123` seconds. More human readable timestamps can also be used, such as `?startTime=1h12m27s` for `1 hour, 12 minutes, and 27 seconds`.
```
dcpena: What about "A timestamp that specifies when the video begins." ?

dcpena (approved):
Approving because my comment is a suggestion and non-blocking

products/stream/src/content/viewing-videos/using-the-stream-player.md
third774: Iterating based on that, how do you feel about "A timestamp that specifies the time from which playback will begin."?

I like that it kinda distinguishes between the video itself and playback.

dcpena: Ah, good callout! Can we iterate one more time? haha.

"A timestamp that specifies the time when playback begins."

third774: Adopted in a force push!

PR #2680, opened 2021-11-11
[Workers] DO revision

deadlypants1973:


lukeed (approved):

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -6,15 +6,15 @@ pcx-content-type: configuration
 
 <Aside type="warning" header="Beta">
 
-Durable Objects are currently in beta and are available to anyone with a Workers subscription. You can enable them for your account in [the Cloudflare dashboard](https://dash.cloudflare.com/) by navigating to “Workers” and then “Durable Objects”.
+Durable Objects are currently in beta and are available to anyone with a Workers subscription. You can enable them for your account in [the Cloudflare dashboard](https://dash.cloudflare.com/) by navigating to **Workers** > **your Worker** > **Durable Objects**.
```
koeninger: Is this actually accurate?  @mpint

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -164,7 +164,7 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
+      - By default, the system will pause outgoing network messages from the Durable Object until all previous writes have been confirmed flushed to disk. In the unlikely event that the write fails, the system will reset the Object, discard all outgoing messages, and respond to any clients with errors instead. This way, Durable Objects can be allowed to continue executing in parallel with a write being performed, without having to worry about prematurely confirming writes, because it is impossible for any external party to observe the Object's actions unless the write actually succeeds. However, this does mean that after any write, subsequent network messages may be slightly delayed. Some applications may consider it acceptable to communicate on the basis of unconfirmed writes and may prefer to permit network traffic immediately. In this case, `allowUnconfirmed` may be set to `true` to opt out of the default behavior. Refer to [this blog post for an in-depth discussion](https://blog.cloudflare.com/durable-objects-easy-fast-correct-choose-three/).
```
pedrosousa:
```suggestion
      - By default, the system will pause outgoing network messages from the Durable Object until all previous writes have been confirmed flushed to disk. In the unlikely event that the write fails, the system will reset the Object, discard all outgoing messages, and respond to any clients with errors instead. This way, Durable Objects can continue executing in parallel with a write operation, without having to worry about prematurely confirming writes, because it is impossible for any external party to observe the Object's actions unless the write actually succeeds. However, this does mean that after any write, subsequent network messages may be slightly delayed. Some applications may consider it acceptable to communicate on the basis of unconfirmed writes and may prefer to allow network traffic immediately. In this case, set `allowUnconfirmed` to `true` to opt out of the default behavior. Refer to [this blog post for an in-depth discussion](https://blog.cloudflare.com/durable-objects-easy-fast-correct-choose-three/).
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -174,13 +174,13 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
+    `put()` returns a `Promise`, but most applications can simply discard this promise without `await`ing it. The `Promise` usually completes immediately anyway, because `put()` writes to an in-memory write buffer that is flushed to disk asynchronously. However, if an application performs a very large number of `put()`s without waiting for any I/O, the write buffer could theoretically grow large enough to cause the isolate to exceed its 128MB memory limit. To avoid this scenario, such applications should `await` the `Promise`s returned by `put()`. The system will then apply backpressure onto the application, slowing it down so that the write buffer has time to flush. Note that these `await`s will disable automatic write coalescing.
```
pedrosousa:
```suggestion
    `put()` returns a `Promise`, but most applications can discard this promise without `await`ing it. The `Promise` usually completes immediately, because `put()` writes to an in-memory write buffer that is flushed to disk asynchronously. However, if an application performs a very large number of `put()`s without waiting for any I/O, the write buffer could theoretically grow large enough to cause the isolate to exceed its 128MB memory limit. To avoid this scenario, such applications should `await` the `Promise`s returned by `put()`. The system will then apply backpressure onto the application, slowing it down so that the write buffer has time to flush. Note that these `await`s will disable automatic write coalescing.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -204,7 +204,7 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
+  - Returns all keys and values associated with the current Durable Object in ascending lexicographic sorted order. The type of each returned value in the [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) will be whatever was previously written for the corresponding key. Be aware of how much data may be stored in your Durable Object before calling this version of `list` without options because it will all be loaded into the Durable Object's memory, potentially hitting its [limit](/platform/limits). If that is a concern, pass options to `list` as documented below.
```
pedrosousa:
```suggestion
  - Returns all keys and values associated with the current Durable Object in ascending lexicographic sorted order. The type of each returned value in the [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) will be whatever was previously written for the corresponding key. Be aware of how much data may be stored in your Durable Object before calling this version of `list` without options because all the data will be loaded into the Durable Object's memory, potentially hitting its [limit](/platform/limits). If that is a concern, pass options to `list` as documented below.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -250,7 +250,7 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
     <Aside type="note" header="Deprecated">
 
-    Explicit transactions are no longer necessary. Any series of write operations with no intervening `await` will automatically be submitted atomically, and the system will prevent concurrent events from executing while `await`ing a read operation (unless `allowConcurrency: true` is used). Hence, a series of reads followed by a series of writes (with no other intervening I/O) are automatically atomic and behave like a transaction.
+    Explicit transactions are no longer necessary. Any series of write operations with no intervening `await` will automatically be submitted atomically, and the system will prevent concurrent events from executing while `await`ing a read operation (unless `allowConcurrency: true` is used). Therefore, a series of reads followed by a series of writes (with no other intervening I/O) are automatically atomic and behave like a transaction.
```
pedrosousa:
```suggestion
    Explicit transactions are no longer necessary. Any series of write operations with no intervening `await` will automatically be submitted atomically, and the system will prevent concurrent events from executing while `await`ing a read operation (unless you use `allowConcurrency: true`). Therefore, a series of reads followed by a series of writes (with no other intervening I/O) are automatically atomic and behave like a transaction.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -274,55 +274,57 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
 ### `fetch()` handler method
 
-The `fetch()` method of a Durable Object namespace is called by the system when an HTTP request is sent to the Object. These requests are not sent from the public Internet, but from other Workers, using a Durable Object namespace binding (see below).
+The `fetch()` method of a Durable Object namespace is called by the system when an HTTP request is sent to the Object. These requests are not sent from the public Internet, but from other [Workers using a Durable Object namespace binding](#accessing-a-durable-object-from-a-worker).
```
pedrosousa:
```suggestion
The system calls the `fetch()` method of a Durable Object namespace when an HTTP request is sent to the Object. These requests are not sent from the public Internet, but from other [Workers using a Durable Object namespace binding](#accessing-a-durable-object-from-a-worker).
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -274,55 +274,57 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
+Namespace bindings have two jobs: generating Object IDs and connecting to Objects.
 
 ### Generating IDs randomly
 
 ```js
 let id = OBJECT_NAMESPACE.newUniqueId()
 ```
 
-Creates a new object ID randomly. This method will never return the same ID twice, and thus it is guaranteed that the object does not yet exist and has never existed at the time the method returns.
+The `newUniqueId()` method on a Durable Object namespace creates a new Object ID randomly. This method will never return the same ID twice, and thus, it is guaranteed that the object does not yet exist and has never existed at the time the method returns.
```
pedrosousa:
```suggestion
The `newUniqueId()` method on a Durable Object namespace creates a new Object ID randomly. This method will never return the same ID twice, and thus, it is guaranteed that the Object does not yet exist and has never existed at the time the method returns.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -274,55 +274,57 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
 
 When generating an ID randomly, you need to store the ID somewhere in order to be able to reach the same object again in the future. You could, for example, store the ID in Workers KV, in an external database, or in a cookie in the user's browser.
 
-Unique IDs are unguessable, therefore they can be used in URL-based access control, sometimes known as "anyone with the link can access."
+Unique IDs are unguessable, therefore they can be used in URL-based access control.
 
-To store the ID in external storage, use its `.toString()` method to convert it into a hex string, and `OBJECT_NAMESPACE.idFromString()` to convert the string back into an ID later.
+To store the ID in external storage, use its `.toString()` method to convert it into a hex string and `OBJECT_NAMESPACE.idFromString()` to convert the string back into an ID later.
```
pedrosousa:
```suggestion
To store the ID in external storage, use its `.toString()` method to convert it into a hexadecimal string and `OBJECT_NAMESPACE.idFromString()` to convert the string back into an ID later.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -274,55 +274,57 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
 ```js
 let id = OBJECT_NAMESPACE.newUniqueId({jurisdiction: "eu"})
 ```
 
-Creates a new object ID that will only run and persist data within the European Union. The jurisdiction feature is useful for building applications that are compliant with regulations such as the [GDPR](https://gdpr-info.eu/). Jurisdiction constraints can only be used with ids created by `newUniqueId()` and are not currently compatible with ids created by `idFromName(name)`.
+The `jurisdiction` option for the `newUniqueId()` method creates a new Object ID that will only run and persist data within the European Union. The jurisdiction feature is useful for building applications that are compliant with regulations such as the [GDPR](https://gdpr-info.eu/). Jurisdiction constraints can only be used with IDs created by `newUniqueId()` and are not currently compatible with IDs created by `idFromName(name)`.
+
+<Aside type="note" header="ID logging">
+
+Note that Object IDs will be logged outside of the EU even if you specify a jurisdiction.
```
pedrosousa:
```suggestion
Object IDs will be logged outside of the EU even if you specify a jurisdiction.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -274,55 +274,57 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
...
 
-<Aside header="ID logging">
-Note that object IDs will be logged outside of the EU even if you specify a jurisdiction.
 </Aside>
 
-Note that objects that are constrained to a jurisdiction may still be accessed by your Workers from anywhere in the world. The jurisdiction constraint only controls where the Durable Object itself runs and persists data. Consider using [Regional Services](https://blog.cloudflare.com/introducing-regional-services/) to control the regions from which Cloudflare responds to requests.
+Note that Objects that are constrained to a jurisdiction may still be accessed by your Workers from anywhere in the world. The jurisdiction constraint only controls where the Durable Object itself runs and persists data. Consider using [Regional Services](https://blog.cloudflare.com/introducing-regional-services/) to control the regions from which Cloudflare responds to requests.
```
pedrosousa:
```suggestion
Your Workers may still access Objects constrained to a jurisdiction from anywhere in the world. The jurisdiction constraint only controls where the Durable Object itself runs and persists data. Consider using [Regional Services](https://blog.cloudflare.com/introducing-regional-services/) to control the regions from which Cloudflare responds to requests.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -337,15 +339,15 @@ let id = OBJECT_NAMESPACE.idFromName(name)
...
+The first time you access a Durable Object based on an ID derived from a name, the system does not know anything about the Object. It is possible that a Worker on the opposite side of the world could have coincidentally decided to access the same Object at the same time. In order to guarantee that only one instance of the Object is created worldwide, the system must check whether the Object has been created anywhere else. Due to the inherent limit of the speed of light, this round-the-world check can take up to a few hundred milliseconds. After this check, the Object will be instantiated near where it was first requested.
```
pedrosousa:
```suggestion
The first time you access a Durable Object based on an ID derived from a name, the system does not know anything about the Object. It is possible that a Worker on the opposite side of the world could have coincidentally decided to access the same Object at the same time. To guarantee that only one instance of the Object is created worldwide, the system must check whether the Object has been created anywhere else. Due to the inherent limit of the speed of light, this round-the-world check can take up to a few hundred milliseconds. After this check, the Object will be instantiated near where it was first requested.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -385,21 +387,21 @@ let stub = OBJECT_NAMESPACE.get(id)
 
 </Definitions>
 
-This method constructs an object "stub", which is a local client that provides access to a remote Durable Object.
+This method constructs an Object stub, which is a local client that provides access to a remote Durable Object.
 
-If the remote object does not already exist, it will be created. Thus, there will always be something for the stub to talk to.
+If the remote Object does not already exist, it will be created. Thus, there will always be something for the stub to talk to.
```
pedrosousa:
```suggestion
If the remote Object does not already exist, it will be created. Thus, there will always be an Object accessible from the stub.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -385,21 +387,21 @@ let stub = OBJECT_NAMESPACE.get(id)
...
+Stubs implement E-order semantics. When you make multiple calls to the same stub, it is guaranteed that the calls will be delivered to the remote Object in the order in which you made them. This ordering guarantee often makes many distributed programming problems easier. However, there is a cost: due to random network disruptions or other transient issues, a stub may become disconnected from its remote Object. Once a stub is disconnected, it is permanently broken, and all in-flight calls and future calls will fail with exceptions. In order to make new requests to the Durable Object, you must call `OBJECT_NAMESPACE.get(id)` again to get a new stub, and you must keep in mind that there are no ordering guarantees between requests to the new stub versus the old one. If ordering is not a concern, you can create a new stub for every request.
```
pedrosousa:
```suggestion
Stubs implement E-order semantics. When you make multiple calls to the same stub, it is guaranteed that the calls will be delivered to the remote Object in the order in which you made them. This ordering guarantee often makes many distributed programming problems easier. However, there is a cost: due to random network disruptions or other transient issues, a stub may become disconnected from its remote Object. Once a stub is disconnected, it is permanently broken, and all in-flight calls and future calls will fail with exceptions. To make new requests to the Durable Object, you must call `OBJECT_NAMESPACE.get(id)` again to get a new stub, keeping in mind that there are no ordering guarantees between requests to the new stub versus the old one. If ordering is not a concern, you can create a new stub for every request.
```

products/workers/src/content/runtime-apis/durable-objects.md
```diff
@@ -412,6 +414,6 @@ let response = await stub.fetch(request)
 let response = await stub.fetch(url, options)
 ```
 
-The `fetch()` method of a stub has the exact same signature as the [global `fetch`](/runtime-apis/fetch). However, instead of sending an HTTP request to the internet, the request is always sent to the Durable Object to which the stub points.
+The `fetch()` method of a stub has the exact same signature as the [global `fetch`](/runtime-apis/fetch). However, instead of sending an HTTP request to the Internet, the request is always sent to the Durable Object to which the stub points.
 
 Any uncaught exceptions thrown by the Durable Object's `fetch()` handler are propagated to the caller's `fetch()` promise.
```
pedrosousa:
```suggestion
Any uncaught exceptions thrown by the Durable Object's `fetch()` handler will be propagated to the caller's `fetch()` promise.
```

pedrosousa (approved):

PR #2707, opened 2021-11-15
Attempt to fix TableWrap usage on workers pricing page

a-robinson:
To fix broken DO pricing table.

This is mostly a shot in the dark, but the previous fix attempt (#2704) still doesn't appear to have helped in prod, so it shouldn't hurt to make the whitespace around these couple `TableWrap`s more similar to the others on the same page.

PR #2708, opened 2021-11-15
Fix DO pricing table

WalshyDev:
Fixes the DO pricing table, it currently shows as broken due to a space instead of an empty line

Current:
image

After fix:
image


WalshyDev:
Superceeded by https://github.com/cloudflare/cloudflare-docs/pull/2707

PR #2709, opened 2021-11-15
Change package manager to yarn

adesege:
# Why?

The `coa` package put out a release with compromised code. This caused pipelines to fail.

# What?

This PR generates lock files for the root package, developers.cloudflare.com package and all the sub-products packages to prevent installing new versions on every install.

Going forward, new products must have their lock files committed to git.

To fix conflicts during rebase, delete the conflicting package-lock.json file and run `npm install` to regenerate it.

# How?

To test this PR, ensure that nothing breaks when any of the products is built locally.


.github/workflows/preview.yml
```diff
@@ -52,48 +51,48 @@ jobs:
...
       - name: Build updated docs
         if: ${{ env.pr_number != '0' }}
         run: |
           while read tile; do
             echo "Building $tile..."
             if [ $tile == "developers.cloudflare.com" ]; then
               pushd developers.cloudflare.com
-              npm install
-              npm run build
+              yarn install
```
jrf0110: `yarn install` sill still update lockfiles and update patch versions of packages. We should run `yarn install --pure-lockfile`

products/workers/src/content/tutorials/store-data-with-fauna/index.md
```diff
@@ -119,7 +119,7 @@ First, install the Fauna JavaScript driver.
 ---
 header: Installing the Fauna driver
 ---
-$ npm install faunadb
```
jrf0110: We shouldn't be touching any of the docs themselves. Let's exclude all `.md` changes unless they're directly related to the docs engine

jrf0110:
I feel like the title and description of this PR could use a little work

> Generate lock files to avoid downloading new versions on every install 

The real thing that we're doing here is moving to yarn.

> The coa package put out a release with compromised code. This caused pipelines to fail.

This could have been avoided by using the [npm ci](https://docs.npmjs.com/cli/v8/commands/npm-ci) command. So if our stated purpose is to fix that issue, then probably a more straight-forward approach would be to run `npm ci` instead of `npm install`

products/workers/src/content/tutorials/store-data-with-fauna/index.md
adesege: Since lock files will be generated using yarn, it's important we reflect this in the documentation as well. Otherwise, a new package-lock.json will be generated whenever `npm install` is ran locally which will not respect the locked versions in yarn.lock and that can introduce unintended behaviour.

adesege:
@jrf0110 Thanks for the feedback. While `npm ci` will resolve generating package-lock,json in ci, we are moving away from npm to yarn so we can take advantage of `resolutions` to lock the dependencies of downstream/upstream packages to specific version. See [the comments here](https://jira.cfops.it/browse/MRK-14197).

products/pages/src/content/platform/build-configuration.md
```diff
@@ -25,7 +25,7 @@ If you are not using a framework, you may input `exit 0` into the **Build comman
 | ---------------------------- | ------------------------------------ | --------------------------- |
 | Angular (Angular CLI)        | `ng build`                           | `dist`                      |
 | Brunch                       | `brunch build --production`          | `public`                    |
-| Docusaurus                   | `npm run build`                      | `build`                     |
```
jrf0110: Changes in this file should be reverted

adesege: I've removed all changes in *.md files. KIndly review again

jrf0110 (approved):

jrf0110 (approved):

jrf0110 (approved):

PR #2710, opened 2021-11-15
fix: remove space from table wrap

eidam:


a-robinson:
Looks fine, but is this doing something that #2707 didn't?

eidam:
> Looks fine, but is this doing something that #2707 didn't?

no, I am just sooo late to the fixing party (I was ping in DM to take a look at it)

a-robinson:
No worries, thanks! The page appears to be fixed in prod now :)

PR #2744, opened 2021-11-18
Update signature for bulk put and delete

vlovich:
The documentation notes there's an additional option but the function signature doesn't list it.


a-robinson (approved):

a-robinson:
Thanks!

PR #2762, opened 2021-11-21
s/req/request

threepointone:
This snippet generates an error when run because there's no object called `req`. This PR fixes the reference.

via https://github.com/cloudflare/wrangler2/issues/37

### DIFF
--- a/products/pages/src/content/platform/functions.md
+++ b/products/pages/src/content/platform/functions.md
@@ -309,7 +309,7 @@ filename: _worker.js
 ---
 export default {
   async fetch(request, env) {
-    if (req.url.startsWith('/api/')) {
+    if (request.url.startsWith('/api/')) {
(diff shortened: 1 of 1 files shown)


WalshyDev:
Code example is also incorrectly assuming `url` is the path rather than full URL. Code should instead be:
```js
export default {
  async fetch(request, env) {
    const { pathname } = new URL(request.url);
    if (pathname.startsWith('/api/')) {
      // TODO: Custom /api/* Worker logic
      return new Response('TODO: add logic')
    }

    // Otherwise, serve static asset(s).
    // Without this, Worker will error and no assets will be served.
    return env.ASSETS.fetch(request)
  }
}
```

Electroid (approved):

lukeed (approved):

PR #2782, opened 2021-11-24
fix(pages): add `api` to path in example

jbergstroem:
As I was reading this part of the documentation, the example with `api` was confusing. Perhaps `api` is meant to be magical? If so, feel free to close the PR.

Also, seems like my editor removed dangling/trailing spaces. Let me know if I should omit that from the commit.


lukeed (approved):

deadlypants1973 (approved):
You are awesome for doing this! Thank you :)

PR #2798, opened 2021-11-28
Specify and speculate around entry-point.

janat08:
https://discord.com/channels/595317990191398933/799437470004412476/914317025969590323
I got confused by the old description of entry-point.

### DIFF
--- a/products/workers/src/content/platform/sites/configuration.md
+++ b/products/workers/src/content/platform/sites/configuration.md
@@ -42,7 +42,7 @@ There are a few specific configuration settings for Workers Sites in your `wrang
     - The directory containing your static assets, path relative to your `wrangler.toml`. Example: `bucket = "./public"`.
 
   - `entry-point` <PropMeta>optional</PropMeta>
-    - The location of your Worker script, default is `workers-site`. Example: `entry-point = "./workers-site"`.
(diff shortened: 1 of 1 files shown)


jkup:
Hi @janat08! Thanks for the PR! This doesn't read very clearly to me. Can you help me understand what issue you ran into? Maybe I could work with you on updating the docs!

janat08:
This spurred the pr: https://discord.com/channels/595317990191398933/799437470004412476/914315249346289665
I did read it the longform docs that entry-point has default behavior that he described, so I don't think this pr is speculation anymore.

jkup:
@janat08 Hey again! Walk us through this please. Here is what I'm seeing:

1. I create a Workers Sites project with `wrangler generate proj --site`
2. At this point, `wrangler dev` runs correctly!
3. I rename the Worker folder with `mv workers-site foo`
4. I update `wrangler.toml`, adding `entry-point = "./foo"`
5. It's still working correctly!

How can I reproduce the issue you originally ran into?

janat08:
I actually no longer remember what the problem is, and I only care about it in so far that I hate it. You describe entry point as something other than what requires package.json which then will throw errors, and without any explanation for why the thing exists. I decided to go into detail as to how it can screw you over, between the default behavior and weird requirements. If you like you can just close the issue, this is a troll pr. You don't use this optional param and it will silently fail, and it will complain about lack of package.json but not any entry files that have js extension.

PR #2801, opened 2021-11-29
Add fetch-html-site.md

MaXFalstein:
Send a request to a remote server, read HTML from the response, and serve that HTML extended to support pathnames of links clicked on the initial HTML response page.

This may need to be updated with a Cloudflare link such as an https://examples.cloudflareworkers.com/demos as used in the original [fetch-html example](https://github.com/cloudflare/cloudflare-docs/blob/production/products/workers/src/content/examples/fetch-html.md) replacing the [BBC News website](https://bbc.co.uk/news) used in testing due to its use of non-FQDN links for articles.

### DIFF
--- a/products/workers/src/content/examples/fetch-html-site.md
+++ b/products/workers/src/content/examples/fetch-html-site.md
@@ -1,0 +1,73 @@
+---
+order: 3
+type: example
+summary: Send a request to a remote server, read HTML from the response, and serve that HTML extended to support pathnames of links clicked on the initial HTML response page.
+tags:
+  - Originless
+pcx-content-type: configuration
+---
+
+# Fetch HTML Site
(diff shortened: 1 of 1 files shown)


kristianfreeman:
Hey @MaXFalstein, I spent some time reading over this example, but I'm not sure I understand what it's doing. Can you tell me a bit more about the use-case so I can understand it better?

lukeed (commented):
This is a fork of the [`fetch-html`](https://github.com/cloudflare/cloudflare-docs/blob/production/products/workers/src/content/examples/fetch-html.md) example, adding BBC-specific hostname and some 404 conditions.

TBH I'm not sure it's unique enough to be added as an entirely new example.

PR #2815, opened 2021-11-29
Update durable-objects.md - s/`await`/`ctx.waitUntil`

threepointone:
The docs say that the different between the two code samples is an await statement, but there's no await statement in either of the samples. I think it meant to say that you can omit the call the `ctx.waitUntil`, because that _is_ the difference between the two.


lukeed (approved):
My bad, thanks!

PR #2832, opened 2021-12-02
[Workers] Fix typo: `colo-local` in how-the-cache-works.md

kino-ma:
- fix typo: `colo-local` to `co-local`

Although I'm not a native speaker of English, I could not find a reasonable meaning of the prefix "colo-" or the word "colo-local". So I thought this was a typo for "co-local".

For that reason, I correct `colo-local` to `co-local` in this PR.

### DIFF
--- a/products/workers/src/content/learning/how-the-cache-works.md
+++ b/products/workers/src/content/learning/how-the-cache-works.md
@@ -5,7 +5,7 @@ pcx-content-type: concept
 
 # How the Cache works
 
-Since Workers was built atop Cloudflare’s network, since its early days, it was designed to allow developers to interact directly with the Cloudflare cache. The cache can provide ephemeral, colo-local storage, as a convenient way to frequently access static or dynamic content.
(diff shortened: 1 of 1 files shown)


WalshyDev:
Colo is short for colocation. Colocation is basically having some servers in a datacentre (generally rented space).

Each PoP (Point of Presence) CF has (for example AMS, MAN, etc) is done through colocation. So this is basically saying colocation-local, meaning local to that PoP.

So yeah, not incorrect but maybe could be made a bit clearer.

kino-ma:
Thank you for your kind response! 
I didn’t know that abbreviation. 😢 

Then, I will make changes to make it clearer after I learned more about CF itself and caches.

(But please feel free to close this PR, I will open another then!)

deadlypants1973:
Hi @kino-ma , thanks for the PR. This doc is due for revision and we'll make sure to update it to be more clear and easily understandable :)

PR #2858, opened 2021-12-06
Update private-networks.md

abracchi-tw:
Proposed new steps to connect a network to Cloudflare through Tunnel.

### DIFF
--- a/products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
+++ b/products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
@@ -34,7 +34,7 @@ To connect a private network to Cloudflare's edge, follow the guide below. You c
     $ cloudflared tunnel create acme-network
     ```
 
-1. Finally, configure your tunnel with the IP/CIDR range of your private network.
+1. Configure your tunnel with the IP/CIDR range of your private network.
 
     ```bash
     $ cloudflared tunnel route ip add 12.55.0.0/16
@@ -60,7 +60,7 @@ To connect a private network to Cloudflare's edge, follow the guide below. You c
     $ cloudflared tunnel run acme-network
     ```
 
-## (optional) Ensure that traffic can reach your network
+## Ensure that traffic can reach your network
 
 By default, Cloudflare for Teams [excludes traffic](/connections/connect-devices/warp/exclude-traffic/split-tunnels) to a specific set of destinations from WARP encryption. If you have not changed your default Split Tunnel settings, you can skip this step. If you have altered this configuration, for example by adding/removing entries to the Exclude list or by setting Split Tunnels mode to **Include**, make sure that traffic to the IP/CIDR you are associating with your private network is indeed set to be encrypted by WARP.
 
--- a/products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
+++ b/products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
@@ -14,7 +14,7 @@ To connect a private network to Cloudflare's edge, follow the guide below. You c
 
 ## Before you start
 
-* [Install](/connections/connect-apps/install-and-setup/installation).
+* [Install](/connections/connect-apps/install-and-setup/installation) `cloudflared`.
 * [Enable HTTP filtering](/policies/filtering/http-policies) by turning on the **Proxy** switch under **Settings** > **Network** > **L7 Firewall**.
 * [Create device enrollment rules](/connections/connect-devices/warp/warp-settings#device-enrollment-permissions) to determine which devices can enroll to your Teams organization.
 * [Install the WARP client](/connections/connect-devices/warp) on the devices you want to allow into your network.
--- a/products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
+++ b/products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
@@ -6,18 +6,15 @@ title: "Private networks"
 
 # Connect private networks
 
-You can connect private networks and the services running in those networks to Cloudflare using [Cloudflare Tunnel](/glossary#cloudflare-tunnel). End users can then connect to those resources using the [WARP client](/connections/connect-devices/warp) by first authenticating into your organization's account. When users connect to an IP made available through Cloudflare Tunnel, WARP sends their connection through Cloudflare's network to the corresponding tunnel.
+You can connect private networks and the services running in those networks to Cloudflare using [Cloudflare Tunnel](/glossary#cloudflare-tunnel). End users can then connect to those resources using the [WARP client](/connections/connect-devices/warp). When users connect to an IP made available through Cloudflare Tunnel, WARP sends their connection through Cloudflare's network to the corresponding tunnel.
 
-Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those connections. Administrators define the IPs available in that environment and associate them with the Tunnel. Users in your organization can then reach the service by enrolling into your organization's Cloudflare for Teams account and using the WARP agent.
-
-Once enrolled, user endpoints will be able to connect to private [RFC 1918](https://tools.ietf.org/html/rfc1918) IP space and other ranges that you control. Applications running on those endpoints will be able to reach those private IPs as well in a private network model.
-Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those connections.
+Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those connections. Administrators define the IPs available in that environment and associate them with the tunnel. Users in your organization can then reach the service by [enrolling](/connections/connect-devices/warp/deployment) into your organization's Cloudflare for Teams account and using the WARP client.
 
 To connect a private network to Cloudflare's edge, follow the guide below. You can also check out our [tutorial](/tutorials/warp-to-tunnel).
 
 ## Before you start
 
-* [Install](/connections/connect-apps/install-and-setup/installation) `cloudflared` on your network.
+* [Install](/connections/connect-apps/install-and-setup/installation).
 * [Enable HTTP filtering](/policies/filtering/http-policies) by turning on the **Proxy** switch under **Settings** > **Network** > **L7 Firewall**.
 * [Create device enrollment rules](/connections/connect-devices/warp/warp-settings#device-enrollment-permissions) to determine which devices can enroll to your Teams organization.
 * [Install the WARP client](/connections/connect-devices/warp) on the devices you want to allow into your network.
@@ -25,21 +22,22 @@ To connect a private network to Cloudflare's edge, follow the guide below. You c
 
 ## Create a tunnel to connect your network
 
-1. Authenticate `cloudflared` with the command below, which will launch a browser window and prompt you to log in with your Cloudflare account:
+1. Authenticate `cloudflared` with the command below. The command will launch a browser window where you will be prompted to log in with your Cloudflare account and pick any zone you have added to Cloudflare.
 
     ```bash
     $ cloudflared login
     ```
-1. Next, you’ll create a tunnel with a user-friendly name to identify your network or environment. 
+
+1. Create a tunnel with a user-friendly name to identify your network or environment. 
 
     ```bash
     $ cloudflared tunnel create acme-network
     ```
 
-1. Finally, configure your tunnel with the IP/CIDR range of your private network. By doing this, you’re making the Cloudflare WARP agent aware that any requests to this IP range need to be routed to our new tunnel. 
+1. Finally, configure your tunnel with the IP/CIDR range of your private network. By doing this, you are creating a private network and making the WARP client aware that any requests to this IP range need to be routed to your new tunnel. 
 
     ```bash
-    $ cloudflared tunnel route ip add 192.168.0.1/32
+    $ cloudflared tunnel route ip add 12.55.0.0/16
     ```
 1. Confirm the routes enrolled with the following command:
 
@@ -47,7 +45,7 @@ To connect a private network to Cloudflare's edge, follow the guide below. You c
     $ cloudflared tunnel route ip show
     ```
 
-1. Next, create a configuration file for the tunnel. In this example, 
+1. Next, create a [configuration file](/connections/connect-apps/configuration/configuration-file) for the tunnel. The configuration file will be structured as follows: 
 
     ```txt
     tunnel: <Tunnel-UUID>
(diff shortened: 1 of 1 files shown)


products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+
+* [Install](/connections/connect-apps/install-and-setup/installation) `cloudflared` on your network.
+* [Enable HTTP filtering](/policies/filtering/http-policies) by turning on the **Proxy** switch under **Settings** > **Network** > **L7 Firewall**.
+* [Create device enrollment rules](/connections/connect-devices/warp/warp-settings#device-enrollment-permissions) to determine which devices can enroll to your Teams organization.
+* [Install the WARP client](/connections/connect-devices/warp) on the devices you want to allow into your network.
+* (optional) [Integrate your preferred identity provider](/identity/idp-integration) with Cloudflare for Teams. 
+
+## Create a tunnel to connect your network
+
+1. Authenticate `cloudflared` with the command below, which will launch a browser window and prompt you to log in with your Cloudflare account:
```
sudarshan-reddy: Can we merge 1 and 2 into a single step?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+    ```bash
+    $ cloudflared login
+    ```
+1. Next, you’ll create a tunnel with a user-friendly name to identify your network or environment. 
+
+    ```bash
+    $ cloudflared tunnel create acme-network
+    ```
+
+1. Finally, configure your tunnel with the IP/CIDR range of your private network. By doing this, you’re making the Cloudflare WARP agent aware that any requests to this IP range need to be routed to our new tunnel.
```
sudarshan-reddy: > IP/CIDR range of your private network

By configuring this, the user is creating a "private network". Let's make that unambiguous?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+    ```bash
+    $ cloudflared tunnel route ip add 192.168.0.1/32
+    ```
+1. Confirm the routes enrolled with the following command:
+
+    ```bash
+    $ cloudflared tunnel route ip show
+    ```
+
+1. Next, create a configuration file for the tunnel. In this example,
```
sudarshan-reddy: Actually this is all they need always. We should say it's as easy as adding only 3 lines perhaps.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+
+1. Run the Tunnel. Traffic inside of your organization, from enrolled WARP agents, will be sent to this instance when the destination is this private IP range:
+
+    ```
+    $ cloudflared tunnel run acme-network
+    ```
+
+## Allow users to connect to your network
+
+On the Teams Dashboard, navigate to **Settings** > **Network** > **Split Tunnels** and click **Manage**.
```
sudarshan-reddy: This seems to be an optional step if I'm not mistaken.

WDYT @nmldiegues ?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+
+## Configure devices to connect to your network
+
+1. On the devices you want to connect to your network, install the Cloudflare root certificate. Find the certificate under **Settings** > **Devices** > **Certificates** and follow [these instructions](/connections/connect-devices/warp/install-cloudflare-cert).
+
+1. Next, log in to the WARP client with Cloudflare for Teams on each of the devices you want to connect.
+    * For desktop clients, go to **Account** > **Login with Cloudflare for Teams** and enter your organization's [team name](/glossary#team-name).
+    * For mobile clients, open the *1.1.1.1 Faster Internet* application, tap the Menu button, go to **Accounts** and enter your organization's [team name](/glossary#team-name).
+
+Users will now be able to reach resources on your network by navigating to any IP address in the range you have specified.
```
sudarshan-reddy: > resources

Perhaps we should explain more? Resources are any http or tcp(for now) based services that are running in the same machine. 

We should also mention that they access this by <IP they previously ran cloudflared route>:<resource_port>/...

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -1,9 +1,10 @@
 ---
 order: 4
 pcx-content-type: concept
+title: "Private networks"
 ---
 
-# Private networks
+# Connect private networks
 
 You can connect private networks and the services running in those networks to Cloudflare using [Cloudflare Tunnel](/glossary#cloudflare-tunnel). End users can then connect to those resources using the [WARP client](/connections/connect-devices/warp) by first authenticating into your organization's account. When users connect to an IP made available through Cloudflare Tunnel, WARP sends their connection through Cloudflare's network to the corresponding tunnel.
```
nmldiegues: `tunnel` -> do we systematically want to refer to it as Tunnel (capitalized)?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
 Once enrolled, user endpoints will be able to connect to private [RFC 1918](https://tools.ietf.org/html/rfc1918) IP space and other ranges that you control. Applications running on those endpoints will be able to reach those private IPs as well in a private network model.
```
nmldiegues: `Once enrolled` -> "enrolled" comes a bit out of nowhere here;  above we say "authenticating into your organization's account", and here we say "enrolled", and I bet the user still doesn't know what either mean in fact, since it looks like the user just needs to login to the dashboard or something

maybe you can replace each of these by a formatted hyperlink that points to the WARP page that explains the enrolment process

nmldiegues: In fact, it's not clear to me what this paragraph is adding.  It seems to be redundant with the paragraph before.

This sentence:
> Applications running on those endpoints will be able to reach those private IPs as well in a private network model.

is particularly confusing to me

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
nmldiegues: I think we need to explain somewhere that `cloudflared login` will ask the user to pick a zone. That zone will not matter for this documentation, but he needs to have one, and he needs to choose (any).

This is completely non-sensical, but until we fix it, we have to be able to explain it. Otherwise the user is immediately blocked on step 1 :)

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+1. Next, you’ll create a tunnel with a user-friendly name to identify your network or environment. 
+
+    ```bash
+    $ cloudflared tunnel create acme-network
+    ```
+
+1. Finally, configure your tunnel with the IP/CIDR range of your private network. By doing this, you’re making the Cloudflare WARP agent aware that any requests to this IP range need to be routed to our new tunnel. 
+
+    ```bash
+    $ cloudflared tunnel route ip add 192.168.0.1/32
```
nmldiegues: Let's make this slightly more realistic (a network Space rather than a full IP) and use a CIDR that does not overlap with default excluded CIDRs (in Split Tunnels):  12.55.0.0/16

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+    $ cloudflared tunnel route ip show
+    ```
+
+1. Next, create a configuration file for the tunnel. In this example, 
+
+    ```txt
+    tunnel: <Tunnel-UUID>
+    credentials-file: /root/.cloudflared/credentials-file.json
+    warp-routing:
+    enabled: true
```
nmldiegues: This needs correct spacing. Also, use `yaml` instead of `txt`

```yaml
tunnel: <Tunnel-UUID>
credentials-file: /root/.cloudflared/credentials-file.json
warp-routing:
    enabled: true
```

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
nmldiegues: We use device and agent in different places. I think we should be consistent about it.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
nmldiegues: I agree. The default Split Tunnels may work since it is a list of exclusions by default.

My recommendation is to change the CIDR used in the example so that it does not overlap with a default exclusion. That way, we can then mark this step as optional:  this is only required to be modified if the CIDR routed to a Tunnel matches an excluded CIDR from WARP, or if the customer wants to modify the Split Tunnel configuration (e.g. instead of Exclusion List, define an Inclusion List)

Also, I think we should add hyperlink to https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/exclude-traffic/split-tunnels

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+    ```
+
+1. Run the Tunnel. Traffic inside of your organization, from enrolled WARP agents, will be sent to this instance when the destination is this private IP range:
+
+    ```
+    $ cloudflared tunnel run acme-network
+    ```
+
+## Allow users to connect to your network
```
nmldiegues: This is more about telling WARP what traffic is meant for Cloudflare or not

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+
+On the Teams Dashboard, navigate to **Settings** > **Network** > **Split Tunnels** and click **Manage**.
+* If you are using the feature in **Exclude** mode, the IP ranges you see listed are those that Cloudflare excludes from WARP encryption by default. If your network's IP/CIDR range is listed on this page, delete it.
+* If you are using the feature in **Include** mode, the IP ranges you see listed are the only one Cloudflare is encrypting through WARP. Add your network's IP/CIDR range to the list.
+
+## Configure devices to connect to your network
+
+1. On the devices you want to connect to your network, install the Cloudflare root certificate. Find the certificate under **Settings** > **Devices** > **Certificates** and follow [these instructions](/connections/connect-devices/warp/install-cloudflare-cert).
```
nmldiegues: I'd say this step can be marked as optional if the customer has Settings -> Network -> Firewall -> TLS decryption disabled

It can be helpful to get started that way, since this is a very hard step for most users (as it is done on the "eyeball side"), and maybe this can be marked as an "Advanced step" optional?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -12,4 +13,67 @@ Cloudflare Tunnel relies on a piece of software, `cloudflared`, to create those
...
+
+1. Next, log in to the WARP client with Cloudflare for Teams on each of the devices you want to connect.
```
nmldiegues: Can we add an hyperlink to the docs that explain this better?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
nmldiegues: I suggest to also add a self-check step here: go to https://help.teams.cloudflare.com/ and make sure that
 - HTTP filtering has WARP and Gateway Proxy both "enabled"
 - Team name matches the expected organization (must belong to the same Cloudflare account where the Tunnel is running from)
 - "Your network is fully protected." is written there

nmldiegues:
An overall comment about where this page is sitting:   it should be within `Connect networks` in my opinion

Being inside `Connect resources -> Configure tunnels -> Private networks` (side-by-side with Configuration file and Arguments) is not the right place a user would expect it to be in.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abracchi-tw: But what if someone has Split Tunnels in **Include** mode? They need to add the CIDR to the list, no? Maybe we can advise users to check their Split Tunnels setup anyway (and agreed, will hyperlink to that page, good catch)

nmldiegues: > But what if someone has Split Tunnels in **Include** mode? They need to add the CIDR to the list, no? Maybe we can advise users to check their Split Tunnels setup anyway (and agreed, will hyperlink to that page, good catch)

The default is Exclude mode.  We will still have this step here, we're just saying that if the account hasn't messed up with the defaults, then there should be nothing to do if the CIDR chosen is not on that list.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abracchi-tw: I'm noticing we don't really have a guide for end users to enroll their devices — in the meantime I've linked to WARP deployment docs where the process is covered (among other things)

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abracchi-tw: it depends! I'm trying to use `tunnel` for tunnel = object. The product name should always be `Cloudflare Tunnel`. does that make sense?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abracchi-tw: got it, makes sense

abracchi-tw: cc @abelinkinbio

abelinkinbio: I'm not sure I agree that we should convey this as an optional step. It will likely lead to more confusion than clarity. Technically, as the documentation stands, I can see why we'd say that, but the more complete fix would be to specify an RFC1918 address in the `cloudflared tunnel route ip add` step and then state this step as a requirement. You either need to remove this CIDR from the exclude list or change the default setting and add it if you're using an include list. 

I think we're trying to account for an unlikely edge case here that will be distracting.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
sudarshan-reddy: `Nit`: Additionally `credentials-file` should be `<credentials-file-name.json>` perhaps?

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
sudarshan-reddy: @abelinkinbio your comment actually makes me lean more towards this being an optional step. People setting up tunnel should not worry about include/exclude lists etc unless they've already had a configuration/setup for this. Let's imagine someone with a vanilla dash/gateway configuration onboarding. Would they need to do this? (Especially if the CIDR is not a default exclusion one) No. So, this step should definitely be optional and not take up their cognitive load.

abelinkinbio: @sudarshan-reddy here is my comprehension of this flow as it stands today. 

1. User creates a tunnel
2. Adds an RFC1918 route
3. Runs tunnel
4. Installs WARP
5. By default, *all* traffic to an RFC 1918 address is _not_ sent to WARP

By stating this as optional, users think they've completed the critical path of steps, but private traffic never makes it to their tunnel.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abelinkinbio: +1

Upfront, we should outline the [topology of a private network on Cloudflare](https://blog.cloudflare.com/extending-cloudflares-zero-trust-platform-to-support-udp-and-internal-dns/) as we do in this blog. There we can make the distinction that the agent is installed on end-users device. Then moving forward, we simply have to refer to the agent.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abelinkinbio: Agreed. It's a silly step for now, and one that can lead to a lot of confusion if we don't explain it. (Users believe their zone is doing something magical in the background).

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abelinkinbio: Great catch, Nuno. This makes it seems like the applications themselves are reaching out to their own IP when I think what we mean is that these are now addressable by end-users through WARP.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abelinkinbio: This is another one that will make more sense if we split out the topology. 

One side is connecting your network to Cloudflare. 
The other side is connecting your users to that network through Cloudflare.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abelinkinbio: Agree with the feedback here. We need to be succinct, but explicit about what every step is accomplishing. 

Here, we should say "if you're planning to use the Cloudflare Agent for secure web filtering with Cloudflare Gateway, you can install the Cloudflare root certificate at this time. Find the certificate under **Settings** > **Devices** > **Certificates** and follow [these instructions](/connections/connect-devices/warp/install-cloudflare-cert). If you do not plan to secure your users outbound traffic you can skip this step." 

cc: @Encore-Encore

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abelinkinbio: We should also be pointing to the dashboard as much as possible. We have onboarding guides and we should add screenshots so users know they can complete this all in one place.

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
sudarshan-reddy: > By default, all traffic to an RFC 1918 address is not sent to WARP

This has never been my experience. Traffic `may` not be sent to WARP on the chance the user chose an address that had to be excluded/included. This is why this step should be optional.

abracchi-tw: @sudarshan-reddy @abelinkinbio @nmldiegues hey everyone, can you let me know if I should leave the step as optional or if you want me to include it in the critical path? thanks! 🙏

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
```diff
@@ -1,15 +1,88 @@
...
+    $ cloudflared login
+    ```
+
+1. Create a tunnel with a user-friendly name to identify your network or environment. 
+
+    ```bash
+    $ cloudflared tunnel create acme-network
+    ```
+
+1. Finally, configure your tunnel with the IP/CIDR range of your private network.
```
nayaabahsan: This is super minor, but is there a reason this says "Finally" even though it's not the last step?

nayaabahsan (approved):

products/cloudflare-one/src/content/connections/connect-apps/configuration/private-networks.md
abracchi-tw: merging this PR, we can always go back and rework this.

PR #2862, opened 2021-12-07
fix numbers and add install link on line 187

chris-martinelli:
Fixed numbers through document.
line 187 - "Follow steps 1 through 3 above to download and install cloudflared on the client desktop" referred to installing the service, which is not needed at this step.


kodster28:
Howdy @chris-martinelli, per our [style guide](https://wiki.cfops.it/display/Prod/Tables+and+Lists), we actually prefer all numbered lists to use `1` (how it's structured currently) b/c markdown auto-renders them to the correct number. It avoids confusion when you have a numbered step in markdown that's different from the numbered step in the output.

chris-martinelli:
closing PR to open new one without number change

PR #2867, opened 2021-12-07
Angela/add documentation for logpush for authoritative and secondary dns

angelampcosta:


products/logs/src/content/reference/glossary.md
```diff
@@ -61,6 +61,10 @@ HTTP response code sent from Cloudflare to the client (end user). The Cloudflare
 
 HTTP response code sent from the origin server to Cloudflare
 
+#### EDNS Client Subnet (ECS)
+
+ECS is forwarded by recursive resolvers and contains information about the network that the DNS query is originating from. Not all resolvers choose to forward ECS, but if they do, usually a part of the IP address is omitted. The subnet length indicates the size of the client subnet in bits that is forwarded. So if the last octet of an IPv4 address is omitted (e.g. 192.0.2.x.), the subnet length will be 24.
```
deadlypants1973:
```suggestion
ECS is forwarded by recursive resolvers and contains information about the network that the DNS query is originating from. Not all resolvers choose to forward ECS, but if they do, usually a part of the IP address is omitted. The subnet length indicates the size of the client subnet in bits that is forwarded. So if the last octet of an IPv4 address is omitted (for example, 192.0.2.x.), the subnet length will be 24.
```

deadlypants1973 (approved):

products/logs/src/content/reference/log-fields/zone/dns_logs.md
```diff
@@ -0,0 +1,25 @@
...
+title: DNS logs 
+---
+
+# DNS logs
+
+The descriptions below detail the fields available for `dns_logs`.
+
+<TableWrap>
+
+| Field | Value | Type |
```
soheiokamoto: This table seems to have manual edits in descriptions (i.e., adding links), which will be lost on the next code generation.

crwaters16 (approved):

PR #2916, opened 2021-12-13
Add note under Health Check Regions

Kruthinarayana:
Add below note under Health Check Regions 

<Aside type="note">
    
**Note**:  
Difference between selecting the **All Data Center vs All regions** for the Health Check Region.
- All datacenters will send requests from over 200+ colos and therefore has an understanding about the entire state of the world. 
- All regions will kick off probes from our 13 regions and sends 39 requests, for example choosing three main colos per region can calculate quorum and ensure the correct status is communicated. 
    
</Aside>


kodster28:
Howdy @Kruthinarayana, thanks for the suggestion, but we actually already have (and link to) that information in http://developers.cloudflare.com/load-balancing/understand-basics/health-details#how-an-origin-becomes-unhealthy

PR #2920, opened 2021-12-13
Update review-malicious-scripts.md

zeinjaber:
In the Script Details UI, we call it "Malicious content".


pedrosousa:
Hi @zeinjaber 
Thank you for the update, but in fact it's the UI that is not up to date (it should mention "code" instead of "content"). The UI is going to be updated very soon.

zeinjaber:
@pedrosousa, thanks for the feedback. Please disregard this PR then.

PR #2956, opened 2021-12-15
Correct X-Forwarded-For example

ad8-bdl:
X-Forwarded-For example currently incorrectly shows the proxy addresses being prepended; they are [should be] appended. Ref, the rest of the doc page, as well as https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For


ad8-bdl:
FTR: I haven't validated Cf's behaviour with `X-Forwarded-For` - I trust it correctly _appends_ proxy addresses, leaving the client IP at the beginning of the list. Ref. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

Then I found this: https://community.cloudflare.com/t/x-forwarded-for-documentaiton-seems-wrong/221504

abracchi-tw (approved):

kodster28 (changes requested):
Actually, we need to verify this example internally

kodster28 (approved):

kodster28:
Thanks for the correction, @ad8-bdl

PR #2971, opened 2021-12-16
Add clarification on Logpush Sumo Logic integration

angelampcosta:


deadlypants1973 (approved):

tanushree-sharma:
Hey Anglea. The description you're adding should mention that the same collector can be used - 

"Note that the same collector can be used for multiple Logpush jobs, but each job must have a dedicated source address."

soheiokamoto (approved):

tanushree-sharma (changes requested):
See my comment above!

tanushree-sharma (approved):

PR #2978, opened 2021-12-16
[Teams] add device posture checks for GW

abracchi-tw:


products/cloudflare-one/src/content/policies/filtering/http-policies/index.md
```diff
@@ -115,6 +115,14 @@ The `Host` selector matches the exact entry input by a customer in the value fie
 
 </Aside>
 
+#### Device Posture
+
+With the Device Posture selector, admins can use signals from end-user devices to secure access to their internal and external resources. For example, a security admin can choose to limit all access to internal applications based on whether specific software is installed on a device, and/or if the device or software are configured in a particular way.
```
kodster28:
```suggestion
With the Device Posture selector, admins can use signals from end-user devices to secure access to their internal and external resources. For example, a security admin can choose to limit all access to internal applications based on whether specific software is installed on a device and/or if the device or software are configured in a particular way. 
```

kodster28 (approved):

PR #3023, opened 2021-12-26
[Workers] Fix DO pricing

WalshyDev:
The pricing section for Durable Objects is a little broken due to the stars. It is making text italic thus not making it clear how the pricing is actually calculated.

Before:
image

After:
image


Electroid (approved):
Thanks for fixing :)

lukeed (approved):

PR #3040, opened 2021-12-30
Update Durable Objects size limit

### LINKED ISSUE #3039: Raise max value size in Durable Objects from 32 KiB to 128 KiB
### Expected/Actual Behavior/Section that requires update
https://developers.cloudflare.com/workers/platform/limits#durable-objects - Indicates limit of 32 KiB even though release notes indicate higher limit of 128 KiB

Release notes: https://community.cloudflare.com/t/2021-12-10-workers-runtime-release-notes/334982

its-jman:
- Increase from 32 KiB to 128 KiB per release notes: https://community.cloudflare.com/t/2021-12-10-workers-runtime-release-notes/334982

#3039

### DIFF
--- a/products/workers/src/content/platform/limits.md
+++ b/products/workers/src/content/platform/limits.md
@@ -282,7 +282,7 @@ The size of chunked response bodies (`Transfer-Encoding: chunked`) is not known 
 
 - Storage keys of up to 2 KiB (2048 bytes)
 
-- Storage values of up to 32 KiB (32768 bytes)
+- Storage values of up to 128 KiB (131072 bytes)
 
 - 30s of CPU time per request, including websocket messages
 


a-robinson (commented):
Hi @jbmanning , thanks for noticing and sending a fix! There are a couple more places in the docs that also need to be updated though. Let's see if I can add them on as an additional commit here.

a-robinson (changes requested):
It's very possible that it's my fault, but github is rejecting my attempts to push an additional commit to your branch. I opened #3048 instead with the extra commit added, but if you'd prefer to have this PR merged then feel free to just copy my changes over here.

its-jman:
No worries, looks good! Good catch on those other locations, thanks!

PR #3050, opened 2022-01-03
Fixed tunnel run command with config parameter

Luuuuuis:
Exactly the same issue as I already mentioned in #3049 

The first 'run' caused the tunnel run command to fail because the config parameter is a tunnel command option, not a subcommand option.


abracchi-tw (approved):

abracchi-tw:
thanks so much for checking!

PR #3078, opened 2022-01-06
Update collect-pcaps.md

zeinjaber:
filter_v1 uses a wirefilter expression to filter the pcap.


products/magic-firewall/src/content/how-to/collect-pcaps.md
```diff
@@ -58,7 +58,8 @@ curl -X POST https://api.cloudflare.com/client/v4/accounts/${account_id}/pcaps \
         "system": "magic-transit"
 }'
 ```
-`filter_v1` can be left empty to collect all packets without any filtering. 
+
+`filter_v1` uses Wirefilter expressions to filter the PCAP. This field can be left empty to collect all packets without any filtering.
```
pedrosousa:
```suggestion
Leaving the `filter_v1` object empty will collect all packets without any filtering. 
```
**UPDATE:** Updated the above suggestion to avoid passive voice. The questions below were already answered in this comment thread. Cc: @dcpena @zeinjaber 

- Can you enter a Wirefilter expression for each field value (`source_address`, `source_port`, etc.), or will the different field values be combined to form a single Wirefilter expression?
    A: We do not support generic wirefilter expressions in `filter_v1`.

- Do you need to add the object without any fields when you don't want to perform any filtering (e.g., `"filter_v1": {}`), or can you omit it?
    A: Both options work, so we don't need to go into details.

dcpena: @arges and @zeinjaber, would you all be able to provide a little more insight to the questions Pedro posed? thanks so much!

arges: As it is written does not seem correct. We do not support generic wirefilter expressions in `filter_v1`.

arges (changes requested):

products/magic-firewall/src/content/how-to/collect-pcaps.md
tejaswi-cf: Hey @pedrosousa,

> Can you enter a Wirefilter expression for each field value (source_address, source_port, etc.), or will the different field values be combined to form a single Wirefilter expression?

Each field value is a single address/port/protocol. It cannot be set to a wirefilter expression, it has to be a single value. We don't want to expose how these values will internally be used to filter traffic.

> Do you need to add the object without any fields when you don't want to perform any filtering (e.g., "filter_v1": {}), or can you omit it?

Either way is fine. In both these cases traffic will not be filtered. The existing sentence about `filter_v1` looks correct.

pedrosousa: As discussed with @tejaswi-cf, I'll remove the "Wirefilter expression" reference (since it doesn't apply here) and just fix the passive voice in the suggestion above.

tejaswi-cf (approved):
Looks good. Thanks Pedro!

KimJ15:
This PR was opened before the developer doc site was moved to Cloudflare Pages. There are many conflicting branches. I'm going to close it now, but feel free to open a new issue or ask any questions.

PR #3084, opened 2022-01-06
Link to the documentation of the previous version of Rate Limiting rules

linzinha:
I replaced the image as well, it should change to what's currently shown on the Cloudflare Dashboard


pedrosousa:
Instead of changing the instructions to the ones already in the Support KB, we should link to those instructions, since they refer to the previous implementation of Rate Limiting rules, and they're not supported by the Ruleset Engine.

I've introduced my proposed changes in this PR.

nayaabahsan (approved):

PR #3099, opened 2022-01-08
Fix Pages functions routing example

leomp12:
I think the current example of [functions routing on Cloudflare Pages](https://developers.cloudflare.com/pages/platform/functions#functions-routing) is wrong 😬 
Example leads to believe that `/functions/api/[[path]].ts` would render index (for example), but at least with `wrangler dev pages` it doesn't, `/functions/api/[[path]].ts` actually does the trick...

### DIFF
--- a/products/pages/src/content/platform/functions.md
+++ b/products/pages/src/content/platform/functions.md
@@ -26,15 +26,15 @@ For example, assume this directory structure:
 ```
 ├── ...
 ├── functions
-|   └── api
-│       ├── [[path]].ts
-│       ├── [username]
-│       │   └── profile.ts
-│       ├── time.ts
-│       └── todos
-│           ├── [[path]].ts
-│           ├── [id].ts
-│           └── index.ts
+|   ├── api
+│   │   ├── time.ts
+│   │   └── todos
+│   │       ├── [[path]].ts
+│   │       ├── [id].ts
+│   │       └── index.ts
+│   ├── [[path]].ts
+│   └── [username]
+│       └── profile.ts
 └── ...
 ```
 
@@ -45,8 +45,8 @@ The following routes will be generated based on the file structure, mapping the 
 /api/todos => ./functions/api/todos/index.ts
 /api/todos/* => ./functions/api/todos/[id].ts
 /api/todos/*/** => ./functions/api/todos/[[path]].ts
-/*/profile => ./functions/api/[username]/profile.ts
-/** => ./functions/api/[[path]].ts
+/*/profile => ./functions/[username]/profile.ts
+/** => ./functions/[[path]].ts
 ```
 
 ### Path segments


Ekwuno:
Hey, @leomp12  so the examples are accurate. By updating the existing folder structure you have also updated the way the URLs will be displayed. So the exiting docs aren’t inaccurate your PR changes the structure of both(which is current but not really an update). 

  For example.  `/functions/todos/[[path]].ts` will return an array of paths so 

 path[0] = '1'
 path[1] = 'monday'

will give `/todos/1/monday`

[[path]] is a catch-all in any layer of the path, including the root. With the updated PR you sent you removed the API root.

Another look is if the path is `functions/[[path.js]]`  then the URL will be `/1/monday` depending on how you handle it.

products/pages/src/content/platform/functions.md
```diff
@@ -45,8 +45,8 @@ The following routes will be generated based on the file structure, mapping the
 /api/todos => ./functions/api/todos/index.ts
 /api/todos/* => ./functions/api/todos/[id].ts
 /api/todos/*/** => ./functions/api/todos/[[path]].ts
-/*/profile => ./functions/api/[username]/profile.ts
-/** => ./functions/api/[[path]].ts
```
leomp12, line 49: Hey @Ekwuno ,
Actually I didn't update the routes. On current example you're pointing
/** (eg. /index) rendered by ./functions/api/[[path]].ts, but it doesn't, am I wrong?

products/pages/src/content/platform/functions.md
```diff
@@ -45,8 +45,8 @@ The following routes will be generated based on the file structure, mapping the
 /api/todos => ./functions/api/todos/index.ts
 /api/todos/* => ./functions/api/todos/[id].ts
 /api/todos/*/** => ./functions/api/todos/[[path]].ts
-/*/profile => ./functions/api/[username]/profile.ts
```
leomp12, line 48: I guess also /*/profile route will not be rendered by ./functions/api/[username]/profile.ts , /api/*/profile would be...

leomp12 (commented):
Hey @Ekwuno ,
Are you sure about that?
I may be confused, but actually I kept both routes /*/profile and /** , editing only the function files to remove the /api path, am I wrong?

leomp12:
> For example.  `/functions/todos/[[path]].ts` will return an array of paths so 
> 
>  path[0] = '1'
>  path[1] = 'monday'
> 
> will give `/todos/1/monday`

Yup, clear to me here...

> [[path]] is a catch-all in **any layer of the path, including the root**.

Ahh, so `./functions/api/[[path]].ts` really should render /index for example ? I've tried it with wrangler pages dev and didn't work, worked only with `./functions/[[path]].ts` or `./functions/index` :thinking:

Maybe I'll test again, thank you!

PR #3100, opened 2022-01-08
add dynamic route section to Next.js Pages guide

rkusa:
This PR adds a section of how to use Next.js dynamic routes on Cloudflare Pages. This is an information I myself was missing when reading the docs, and something that also came up in Discord just recently. So I think it is something a lot of people would like to know when reading the docs.

It is probably a good idea to have a technical writer look over the PR as English is not my native language. If this lands, it probably makes sense to add a similar section to the Gatsby docs (and maybe other frameworks, but Next.js and Gatsby the only ones where I've used dynamic routes in the past).

### DIFF
--- a/products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
+++ b/products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
@@ -57,6 +57,20 @@ For the complete guide to deploying your first site to Cloudflare Pages, refer t
 After deploying your site, you will receive a unique subdomain for your project on `*.pages.dev`.
 Every time you commit new code to your Next.js site, Cloudflare Pages will automatically rebuild your project and deploy it. You will also get access to [preview deployments](/platform/preview-deployments) on new pull requests, so you can preview how changes look to your site before deploying them to production.
 
+## Dynamic Routes
+
+Next.js [Dynamic Routes](https://nextjs.org/docs/routing/dynamic-routes) require a rewrite that can be added using [Functions](/pages/platform/functions). Consider the following dynamic route: `pages/post/[pid].js`. To intercept all requests matching that route and serve `/post/[pid].html` (created when exporting your Next.js site) to them, add a function at `/functions/post/[pid].js` with the following content:
+
+```js
+export function onRequestGet({ env, request }) {
+  return env.ASSETS.fetch(
+    new Request(new URL("/post/[pid]", request.url).toString(), request)
+  );
+}
+```
+
+This function assumes that the Next.js [`trailingSlash` option](https://nextjs.org/docs/api-reference/next.config.js/trailing-slash) is disabled (default). If you have the `trailingSlash` option enabled, the rewrite must point to `"/profiles/[profile]/"` (notice the trailing slash).
+
 ## Learn more
 
 By completing this guide, you have successfully deployed your Next.js site to Cloudflare Pages. To get started with other frameworks, [refer to the list of Framework guides](/framework-guides).


kristianfreeman:
@rkusa looks great! Do you have an example repo we can test this with?

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -57,6 +57,20 @@ For the complete guide to deploying your first site to Cloudflare Pages, refer t
...
 Every time you commit new code to your Next.js site, Cloudflare Pages will automatically rebuild your project and deploy it. You will also get access to [preview deployments](/platform/preview-deployments) on new pull requests, so you can preview how changes look to your site before deploying them to production.
 
+## Dynamic Routes
+
+Next.js [Dynamic Routes](https://nextjs.org/docs/routing/dynamic-routes) require a rewrite that can be added using [Functions](/pages/platform/functions). Consider the following dynamic route: `pages/post/[pid].js`. To intercept all requests matching that route and serve `/post/[pid].html` (created when exporting your Next.js site) to them, add a function at `/functions/post/[pid].js` with the following content:
+
+```js
+export function onRequestGet({ env, request }) {
+  return env.ASSETS.fetch(
+    new Request(new URL("/post/[pid]", request.url).toString(), request)
```
lukeed, line 67: As written, this will be a hardcoded path to `/post/[pid]`.

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -57,6 +57,20 @@ For the complete guide to deploying your first site to Cloudflare Pages, refer t
...
+  );
+}
```
lukeed, line 69: No need to rewrite the URL. It's the same request but serving from a different location. 

```suggestion
export function onRequestGet({ env, request }) {
  return env.ASSETS.fetch(request);
}
```

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -57,6 +57,20 @@ For the complete guide to deploying your first site to Cloudflare Pages, refer t
...
+```
+
+This function assumes that the Next.js [`trailingSlash` option](https://nextjs.org/docs/api-reference/next.config.js/trailing-slash) is disabled (default). If you have the `trailingSlash` option enabled, the rewrite must point to `"/profiles/[profile]/"` (notice the trailing slash).
```
lukeed, line 72: I don't think this makes a difference for static asset serving (what we're doing above). Pretty sure Next.js is equipped to handle `/about` and `/about/` the same way, even if one redirects to the other.

rkusa:
> Do you have an example repo we can test this with?

@codewithkristian Yes, here is one: https://github.com/rkusa/cf-pages-next-test

I've added some examples:
- [`main`](https://github.com/rkusa/cf-pages-next-test) branch: uses the function from the code snippet of this PR; example: https://cf-pages-next-test.pages.dev/post/42
- [`no-rewrite`](https://github.com/rkusa/cf-pages-next-test/tree/no-rewrite) branch: same dynamic route, but no function to rewrite; example: https://cb0c5741.cf-pages-next-test.pages.dev/post/42 (you'll get a 404 without the rewrite)
- [`trailing-slash`](https://github.com/rkusa/cf-pages-next-test/tree/trailing-slash) branch: enabled `trailingSlash` Next.js option without any other change; example: https://c3629fae.cf-pages-next-test.pages.dev/post/42 (redirect loop)
- [`trailing-slash-fixed`](https://github.com/rkusa/cf-pages-next-test/tree/trailing-slash-fixed) branch: fixed the above by adding the trailing slash `/` to the assets request; example: https://5b584593.cf-pages-next-test.pages.dev/post/42/

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
rkusa: I don't think this is the case. What this should achieve is to fetch `/post/[pid].html` for any request matching `/post/[pid]`, so eg. `/post/1`, `/post/2` etc.
With the suggested change, if the incoming request is for `/post/1`, it would try to open `/post/1.html` (or `/post/1`), which does not exist.

Edit: In case this could be the cause of confusion, the suggested change of this PR assumes an exported Next.js site (build command of the docs is: `npm run export`) and not a Next.js server the requests can be forwarded to.

products/pages/src/content/framework-guides/deploy-a-nextjs-site.md
rkusa: From what I've gathered, `env.ASSETS.fetch` must not return a redirect or you'll end up with a redirect loop. I've added the two last examples here https://github.com/cloudflare/cloudflare-docs/pull/3100#issuecomment-1009117445 as an example for that. The second last shows that you'll end up with a redirect loop if the trailing slash is missing, and the last example adds the trailing slash, which fixes the redirect loop.

The redirect happens because with `trailingSlash: true`, the file is `/post/[pid]/index.html` and not `/post/[pid].html`. We can neither request `/post/[pid]/index.html` nor  `/post/[pid]`, because both redirect to `/post/[pid]/` (this is a Cloudflare redirect and not a Next.js one).

But you are right, with the last example, it doesn't matter if you open either of the following:
- https://5b584593.cf-pages-next-test.pages.dev/post/42/
- https://5b584593.cf-pages-next-test.pages.dev/post/42

rkusa:
```suggestion
This function assumes that the Next.js [`trailingSlash` option](https://nextjs.org/docs/api-reference/next.config.js/trailing-slash) is disabled (default). If you have the `trailingSlash` option enabled, the rewrite must point to `"/post/[pid]/"` (notice the trailing slash).
```

(I'll wait with this change, to see if there is anything else that need to change).

lukeed: The `env.ASSETS.fetch` will auto-resolve the final file in the same way it would for a static Pages deployment. The bindings applies the same file-matching logic that the rest of Pages uses, so sending it `/foo` vs `/foo/` will both resolve to the `foo/index.html` file.

rkusa: @lukeed Yes, it will resolve to `foo/index.html`, but by returning a redirect and not by rewriting it internally. Taking the examples from https://github.com/cloudflare/cloudflare-docs/pull/3100#issuecomment-1009117445:

https://c3629fae.cf-pages-next-test.pages.dev/post/42 has `trailingSlash` enabled and ends up with a redirect loop.

It's fixed by adding the trailing slash as described in the document (see commit: https://github.com/rkusa/cf-pages-next-test/commit/7e1adaf3d806b25f4bccedfa8443025c86f92044). This fixes the redirect loop: https://5b584593.cf-pages-next-test.pages.dev/post/42/

PR #3103, opened 2022-01-10
Remove unused variable in cache api example

Cyb3r-Jak3:
`const cacheUrl = new URL(request.url)` appears to be unused as the cache is generated from `cacheUrl.toString()`


lukeed:
Thank you but it's used 2 lines below :)

JordanMG:
It is "used", yes, but why is it necessary? The code converts a string to a `URL` only to convert it back into a string again.

lukeed:
Yes but if you'd like to make that change, please make the full change. The PR – as is – makes the snippet invalid & would throw a runtime error if executed.

PR #3129, opened 2022-01-12
Add another example for next() in functions

Ekwuno:


products/pages/src/content/platform/functions.md
```diff
@@ -264,6 +264,30 @@ async function errorHandler(context) {
 export const onRequest = errorHandler;
 ```
 
+Another usecase for `next` function is passing the request cycle from the current middleware function to the next function in the stack if the current function does not end the request-response cycle. Using the next()function will pass control to the next middleware function, depending on the order of execution. For example:
```
deadlypants1973:
```suggestion
Another use case for the `next` function is passing the request cycle from the current middleware function to the next function in the stack if the current function does not end the request-response cycle. Using the `next()` function will pass control to the next middleware function, depending on the order of execution. For example: 
```

deadlypants1973 (commented):
Reviewed but also would recommend DA review for this as it includes code.

products/pages/src/content/platform/functions.md
```diff
@@ -264,6 +264,30 @@ async function errorHandler(context) {
 export const onRequest = errorHandler;
 ```
 
+Another use case for the `next` function is passing the request cycle from the current middleware function to the next function in the stack if the current function does not end the request-response cycle. Using the `next()` function will pass control to the next middleware function, depending on the order of execution. For example:
```
lukeed, line 267: These two sentences say effectively the same thing. Of the two, the first has a little more info

products/pages/src/content/platform/functions.md
```diff
@@ -264,6 +264,30 @@ async function errorHandler(context) {
...
+                statusText: "Internal Server Error"
+            })
+        }
+
+    },
+ ({request,next})=>{
+     return new Response("Hello from next base middleware");
+ }
+];
+
```
lukeed:
```suggestion
// Attach multiple handlers
export const onRequest = [
  async ({ request,next }) => {
    try {
      // Call the next handler in the stack
      const response = await next()
      const responseText = await response.text()
      //~> "Hello from next base middleware"
      return new Response(responseText + " from middleware")
    } catch (thrown) {
      return new Response(`Error ${thrown}`, {
        status: 500,
        statusText: "Internal Server Error"
      })
    }
  },
  ({ request, next }) => {
    return new Response("Hello from next base middleware");
  }
];
```

products/pages/src/content/platform/functions.md
Ekwuno: I was trying to say, in the second sentence, that "If next is called after a function, the function would run before passing over to the next handler" Should I reword it to say just that?

lukeed: on second look, i think this is fine :) ignore me

lukeed (approved):

Issue #2260, opened 2021-09-09
`CF_*` environment variables supported by wrangler


dlqqq:
In the documentation, I've noticed that you can specify `CF_ZONE_ID` and `CF_ACCOUNT_ID` to dynamically change `zone_id` and `account_id` during deployment based on the environment. This is useful for having dynamic deployment behavior. However, there are also other Wrangler options. Namely `CF_NAME`, which we use to spin up a custom `wrangler.dev` subdomain for each branch in our serverless repository. For example, if I push to a branch `beta-fixes`, I get the `beta-fixes.[ORGANIZATION].wrangler.dev` subdomain allocated to that function.

 I can't find any mention of this in the docs; could somebody please point me to a source or add documentation to save the time of future users?

dlqqq:
Related: https://github.com/cloudflare/wrangler/issues/1885

jkup:
Added a note to the docs. Would love to get feedback on whether you feel this covers it or if there is more we should add! https://github.com/cloudflare/cloudflare-docs/pull/3178

PR #3178, opened 2022-01-19
Add section on Wrangler env variables

jkup:
Noticing we don't document that you can pass in TOML keys as Env variables

addresses https://github.com/cloudflare/cloudflare-docs/issues/2260 and https://github.com/cloudflare/wrangler/issues/1885

Should we add more or does this cover it?

### DIFF
--- a/products/workers/src/content/cli-wrangler/commands.md
+++ b/products/workers/src/content/cli-wrangler/commands.md
@@ -1023,7 +1023,7 @@ y
 
 ## Environment variables
 
-Wrangler also supports any `wrangler.toml` keys passed in as environment variables. This works by passing in `CF_` + any toml key uppercased. So you can pass them in like:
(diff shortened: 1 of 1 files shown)


products/workers/src/content/cli-wrangler/commands.md
```diff
@@ -1021,6 +1021,14 @@ y
 
 --------------------------------
 
+## Environment variables
+
+Wrangler also supports any `wrangler.toml` keys passed in as environment variables. This works by passing in `CF_` + any toml key uppercased. So you can pass them in like:
```
deadlypants1973:
```suggestion
Wrangler supports any `wrangler.toml` keys passed in as environment variables. This works by passing in `CF_` + any uppercased TOML key. For example:
```

deadlypants1973 (approved):

PR #3185, opened 2022-01-19
Update custom-origin.md

zeinjaber:
https://jira.cfops.it/browse/PSPEC-832


kodster28:
Howdy @zeinjaber, we already have the limitations for custom origin server documented at the top of the page. Since SNI rewrite is a sub-feature of custom origin servers, I don't think we'd need to re-add the reminder to this section.

Screen Shot 2022-01-19 at 3 20 46 PM
.

zeinjaber:
That makes sense, @kodster28

PR #3192, opened 2022-01-20
rm redundant "if you are deploying to..."

ignoramous:
[A conditional so nice, they added it twice](https://en.wikipedia.org/wiki/New_York,_New_York_(So_Good_They_Named_It_Twice))...?


kodster28 (approved):

kodster28:
Thanks for the catch, @ignoramous! 
(also appreciate the reference)

PR #3207, opened 2022-01-20
Change Workers changelog dates to ISO-8601

Electroid:
We'd like to stick to one standard of dates, and ISO-8601 seems like the best choice.

### DIFF
--- a/products/workers/src/content/platform/changelog.md
+++ b/products/workers/src/content/platform/changelog.md
@@ -1,14 +1,14 @@
 # Changelog
 
-## 1/7/2022
+## 2022-01-07
 
 - The `workers_api_getters_setters_on_prototype` configuration flag corrects the way Workers attaches property getters and setters to API objects so that they can be properly subclassed.
 
-## 12/22/2021
+## 2021-12-22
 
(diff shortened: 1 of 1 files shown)


deadlypants1973 (approved):

products/workers/src/content/platform/changelog.md
```diff
@@ -215,12 +215,12 @@ Changes in the prior release:
...
 - Fixed a rare bug where a WritableStream could be garbage collected while it still had writes queued, causing those writes to be lost.
 
-## 12/10/2020
+## 2020-12-10
 
 Changes this week:
 - Major V8 update: 8.7.220.29 -> 8.8.278.8
 
-## 9/19/2019
+## 2019-09-19
```
WalshyDev, line 223: 2019 entry in the middle of 2020 ones? Think this is in the wrong place

jkup (approved):
Thanks for doing this! LGTM if we can move that one Walshy pointed out is in the wrong spot!

products/workers/src/content/platform/changelog.md
Electroid: Ah yes! Though let's put that in another PR.

PR #3227, opened 2022-01-23
[Workers] fix typo in tutorials/build-a-slackbot

tniessen:
- Change `that the the webhook` to `that the webhook`.


kodster28 (approved):
Thanks for the fix, @tniessen!

PR #3231, opened 2022-01-24
chore: MRK-14167 Update yarn.lock

younggeeks:
### Background
We've added a new meta field `pcx-content-type` which will be used by Coveo Scrapper.    
    - In this PR we are updating `Yarn.lock` for the field to reflect in individual product sites
    
    
    #### Screenshot
   
Screenshot 2022-01-24 at 16 09 34


kevin-wynn (approved):
This looks good to me, the main thing is that docs-engine is pointing towards https://github.com/cloudflare/cloudflare-docs-engine.git#3e7f905b9db824845c4ab8bb9bda1abc9cc8011d which is the correct git hash for the pcx content type change.

I do wonder if we need all the other changes for patch version upgrades in here with it?

adesege:
Tested locally and it's all good to me

kodster28 (changes requested):
Hey! I still need to test this out, but we need to delay this until we can verify that the docs engine can still build pages that _do not_ yet have a content type.

There's a small subset of our pages that don't have that value currently and this change would prevent us from making any changes until that's addressed.

kodster28:
Okay, have verified that the value does not affect whether or not the page is rendered. Tested on the `WAF` docs since a lot of the changelog pages do not have `pcx-content-type` values.
(sorry about the caution... but I broke all our pages last year when I made the DocSearch component mandatory).

I'd still want @pedrosousa to have a look at it if he's able to, so leaving as pending until then.

pedrosousa:
Seems to be working properly, both when using the development server and when building the final Gatsby site. Tested with a few products.

A couple of final questions:

1) The following folders also have `yarn.lock` files, but they're not being updated. Is there any particular reason for this? They should probably use the same engine version.

  * `browser-isolation`
  * `dns`
  * `docs-engine`
  * `events`
  * `magic-firewall`
  * `magic-wan`
  * `pages`
  * `spectrum`
  * `workers`

2) The `yarn.lock` file in `products/api-security` is the only file in that folder. Can you remove both the file and the folder, since they're not needed?

pedrosousa:
I'm also getting the following deprecation warning when running the development server and when building:

```
warn Deprecation warning - adding inferred resolver for field Frontmatter.pcx_content_type. In Gatsby v3, only fields with an explicit directive/extension will get a resolver.
```

younggeeks:
Hey @pedrosousa,
I've updated the PR accordingly.

kodster28:
@younggeeks, I think this change has been merged in with another PR done by @adesege.

PR #3245, opened 2022-01-25
Document comments format in _headers

rspilker:
In the `_headers` documentation, I could not find how to do comments.

I've asked on Discord, and verified the answer.

I also noticed that `_redirects` lacks the same documentation, and though I suspect the syntax is the same, I did not verify that.

### DIFF
--- a/products/pages/src/content/platform/headers.md
+++ b/products/pages/src/content/platform/headers.md
@@ -82,6 +82,22 @@ filename: _headers
   x-movie-name: You are watching ":title"
 ```
 
+### Comments
+
+If you want to document the rules or headers, you can put comments on a separate lines, starting with an octothorpe (`#`). 
+
+```txt
+---
+filename: _headers
+---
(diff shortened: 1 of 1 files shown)


rspilker:
Hmm, there is a small typo: "on a separate lines". The 'a' should be removed.

lukeed:
Hey, thank you for this! Instead of adding an entire section just for this, can you add a single comment around L30? That should be enough for people to realize how to do them.

Something like this:

```txt
---
filename: _headers
---
# This is a comment
/secure/page
  X-Frame-Options: DENY
  X-Content-Type-Options: nosniff
  Referrer-Policy: no-referrer

/static/*
  Access-Control-Allow-Origin: *
  X-Robots-Tag: nosnippet

https://myproject.pages.dev/*
  X-Robots-Tag: noindex
```

rspilker:
@lukeed Good suggestion, implemented.

lukeed (approved):

Issue #3148, opened 2022-01-13
Having root and output set with Exit 0 command causing builds to fail


ghost:
### Expected Behavior
When `exit 0` is set in the **Build Command** field and no prebuild environment is chosen, only one field should be specified (output or root?).

### Actual Behavior
Specifying both the root directory and output directory with the `exit 0` command causes builds to fail.

### Section that requires update
[Provide a link to the page that needs an update and be specific about which section requires the update.](https://developers.cloudflare.com/pages/framework-guides/deploy-anything#deploying-with-cloudflare-pages)

### Additional information
A simple clarification in the docs about this issue could be very useful, especially for beginners/non-devs.

WalshyDev:
Both fields can be set fine with `exit 0` or any command. What's the error you're getting?

(also, I'd recommend getting support on Discord - https://workers.community)

ghost:
> Both fields can be set fine with `exit 0` or any command. What's the error you're getting?
> 
> (also, I'd recommend getting support on Discord - https://workers.community)

[I just came from there](https://discord.com/channels/595317990191398933/789155108529111069/931311033795104790)

WalshyDev:
Ah, I see. Greg was referring to the fact that you should set your output or root since you're attempting to point them both to the same location. Right now your output folder would be `/public/public` not `/public` hence why you would only need to set one of them :)

Clarification on that could be improved in the docs. However, it isn't related to having both directories and a command. Hopefully, this clears it up!

lukeed:
Closing as this was about the relationship between the root & output directories, not about `exit 0` or any build command.

PR #3256, opened 2022-01-26
Update instructions for exit 0

Ekwuno:
This PR addresses #3148

### DIFF
--- a/products/pages/src/content/framework-guides/deploy-anything.md
+++ b/products/pages/src/content/framework-guides/deploy-anything.md
@@ -34,7 +34,7 @@ Deploy your site to Pages by logging into the [Cloudflare dashboard](https://das
 
 </TableLayout>
 
-Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field.
+Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field and then specify the **Build output directory**, your build command will run in the folder specified and you will not need to add the same directory as a path in the **Root directory**.
 
 After configuring your site, you can begin your first deploy. Your custom build command (if provided) will run, and Pages will deploy your static site.
 
--- a/products/pages/src/content/framework-guides/deploy-anything.md
+++ b/products/pages/src/content/framework-guides/deploy-anything.md
@@ -34,7 +34,7 @@ Deploy your site to Pages by logging into the [Cloudflare dashboard](https://das
 
 </TableLayout>
 
-Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field and then specify the **Build output directory**, your build command will run in the folder specified and you will not need to add the same directory as a path in the **Root directory**.
+Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input `exit 0` into the **Build command** field and specify a **Build output directory** where your application's content lives.
 
 After configuring your site, you can begin your first deploy. Your custom build command (if provided) will run, and Pages will deploy your static site.
 
--- a/products/pages/src/content/framework-guides/deploy-anything.md
+++ b/products/pages/src/content/framework-guides/deploy-anything.md
@@ -34,7 +34,7 @@ Deploy your site to Pages by logging into the [Cloudflare dashboard](https://das
 
 </TableLayout>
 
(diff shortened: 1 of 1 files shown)


products/pages/src/content/framework-guides/deploy-anything.md
```diff
@@ -34,7 +34,7 @@ Deploy your site to Pages by logging into the [Cloudflare dashboard](https://das
 
 </TableLayout>
 
-Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field.
+Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field and then specify the **Build output directory**, your build command will run in the folder specified and you will not need to add the same directory as a path in the **Root directory**.
```
deadlypants1973:
```suggestion
Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input `exit 0` into the **Build command** field and specify a **Build output directory** where your application's content lives.

Comment (Ekwuno) [reply products/pages/src/content/framework-guides/deploy-anything.md]: We need to mention that if you set the build output directory as  `./public` for example and it's the same as the root. There is no need to have that twice;  because instead of building at `/public` Pages will look in `/public/public`

Comment (lukeed) [reply products/pages/src/content/framework-guides/deploy-anything.md]: Not quite – they're serving different purposes. 

Made up example: if you have your site living in the `public` directory, then:

```
root: / (default)
build: public
result ==> /public ✅

root: /public 
build: /
result ==> /public ✅

root: /public
build: public
result ==> /public/public ❌
```

We don't want to recommend people change the root directory unless they _actually_ need the work (the build command) to run from that location. 

The pipeline workflow is essentially this:

```sh
git clone <repo> tmpdir
cd tmpdir
if (root != "/") cd <root>
run <build command>
deploy <build output dir>
```

So if you don't have a build command, you're just resolving a destination folder somehow (eg "/public"). There _are_ two ways to get to that destination, but the _more correct_ way is to avoid `root` changes when unnecessary.

Ekwuno: > We don't want to recommend people change the root directory unless they actually need the work (the build command) to run from that location. 

@lukeed  That's what I am saying "you will not need to add the same directory as a path in the **Root directory** "... 

This PR addresses the confusion in Issue 3148, because it's not specified in docs, people might do this: 

```
root: /public
build: public
result ==> /public/public ❌
```
Which will lead to the build failing. 

In the UI there is a prompt to tell you to leave it alone if your site isn't in a subdirectory.

Ekwuno: "Your project’s root directory, where Cloudflare runs the build command. If your site is not in a subdirectory, leave this path value empty."

GregBrimble: I think @deadlypants1973 's suggestion is more accurate that @Ekwuno 's original suggestion.

> Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field and then specify the **Build output directory**, your build command will run in the folder specified and you will not need to add the same directory as a path in the **Root directory**.

The build command will not run in the build output directory. It will run in the root.

I don't think it's worth specifying that the user shouldn't put the same directory as both the build output directory and the root just because one person tripped up on that. The UI is relatively clear that most people won't need to set it, and by bringing it up here, I honestly think it just muddies the waters.

image

Ekwuno: I still think this should be documented a little differently or at least feedback in the UI to stop this mistake. I am happy to land this PR right now, and we can revisit at a later time.

products/pages/src/content/framework-guides/deploy-anything.md
```diff
@@ -34,7 +34,7 @@ Deploy your site to Pages by logging into the [Cloudflare dashboard](https://das
 
 </TableLayout>
 
-Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input  `exit 0` into the **Build command** field.
+Unlike many of our framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, input `exit 0` into the **Build command** field and specify a **Build output directory** where your application's content lives.
```
deadlypants1973:
```suggestion
Unlike many of the framework guides, the build command and build directory for your site are going to be completely custom. If you do not need a build step, leave the **Build command** field empty and specify a **Build output directory**. The build output directory is where your application's content lives.
```

deadlypants1973 (approved):

lukeed:
This has to be rebased (or probably recreated) since it was created before the migration. I suggest recreate the PR since the changes are few

PR #3275, opened 2022-01-27
fix diagram type

odyslam:
It's a sequence diagram

I am a contributoooor


kodster28:
Howdy @odyslam, thanks for the contribution, but we're going to defer to the Auth0 documentation here. They call it a flow in their docs so we'll just stick with their wording.

PR #3277, opened 2022-01-27
[Workers] fix vulnerability in signature example

tniessen:
Without proper separation/padding, the existing example potentially allows bypassing signature/expiry checks.

For example, a signature for the pathname `"/get/123456"` with expiry `1643291710779` would also be valid for the pathname `"/get/12345"` with expiry `61643291710779`, thus extending the signature's validity by two millennia, while the pathname might still be valid. Similarly, if no access to `"/get/12345"` was intended, this allows privilege escalation.


kristianfreeman (approved):

products/workers/src/content/examples/signing-requests.md
```diff
@@ -54,9 +54,11 @@ async function verifyAndFetch(request) {
 
   // Extract the query parameters we need and run the HMAC algorithm on the
   // parts of the request we are authenticating: the path and the expiration
-  // timestamp.
+  // timestamp. It is crucial to pad the input data, e.g., by adding a symbol
```
deadlypants1973:
```suggestion
  // timestamp. It is crucial to pad the input data, for example, by adding a symbol
```

products/workers/src/content/examples/signing-requests.md
```diff
@@ -54,9 +54,11 @@ async function verifyAndFetch(request) {
...
+  // in-between the two fields that can never occur on the right side. Inthis
```
deadlypants1973:
```suggestion
  // in-between the two fields that can never occur on the right side. In this
```

products/workers/src/content/examples/signing-requests.md
```diff
@@ -54,9 +54,11 @@ async function verifyAndFetch(request) {
...
+  // case, we use the @ symbol to separate the fields.
```
deadlypants1973:
```suggestion
  // case, use the @ symbol to separate the fields.
```

products/workers/src/content/examples/signing-requests.md
```diff
@@ -130,7 +132,14 @@ async function generateSignedUrl(url) {
   // parameter.
   const expirationMs = 60000
   const expiry = Date.now() + expirationMs
-  const dataToAuthenticate = url.pathname + expiry
+  // The signature will be computed for the pathname and the expiry timestamp.
+  // The two fields must be separated or padded to ensure that an attacker
+  // will not be able to use the same signature for other pathname/expiry pairs.
+  // The @ symbol is guaranteed not to appear in expiry, which is a (decimal)
+  // number, so we can safely use it as a separator here. When combining more
```
deadlypants1973:
```suggestion
  // number, so you can safely use it as a separator here. When combining more
```

deadlypants1973 (commented):
PCX style guide edits :)

deadlypants1973 (approved):

PR #3285, opened 2022-01-27
Dns proxied domains

kodster28:
partially addresses PCX-3229.


products/dns/src/content/manage-dns-records/how-to/create-dns-records.md
```diff
@@ -0,0 +1,73 @@
+---
```
kodster28, line 1: This page just moved around. All the content is the same.

marciocloudflare (approved):

PR #3293, opened 2022-01-28
[Images] Replace example domain with generic example

jack828:
I for one, LOVE trying to buy domains I see on movies/tvshows - not that I've been successful so far, but I saw this one was used in your docs and it happens to be available.

I wouldn't like to copy that URL and be served something nasty (or be redirected to a rival cdn...) - so I figure best practice would be to replace it with the example.com one.

Thanks for great documentation!


kodster28 (approved):
@jack828, I'm disappointed you didn't switch it to `waynesworld.com`, but I suppose this'll do.

Thanks for the contribution!

PR #3309, opened 2022-01-31
[Workers] KV revision p2

deadlypants1973:


lukeed (approved):

products/workers/src/content/runtime-apis/kv.md
```diff
@@ -42,33 +42,35 @@ The maximum size of a value is 25 MiB.
 
 You can also [write key-value pairs from the command line with Wrangler](/cli-wrangler/commands#kvkey) and [write data via the API](https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair).
 
-Due to the eventually consistent nature of Workers KV, concurrent writes from different edge locations can end up overwriting one another. It is a common pattern to write data via Wrangler or the API but read the data from within a Worker, avoiding this issue by issuing all writes from the same location.
+Due to the eventually consistent nature of Workers KV, concurrent writes from different edge locations can end up overwriting one another. It is a common pattern to write data via Wrangler or the API so that all writes initiate from the same location. This avoids competing, concurrent writes because of the single stream. All data is still readily available within all Workers bound to the namespace.
```
koeninger: I think the emphasis on location is misleading here.  I think the original point was to avoid concurrent writes by using a single process that happens to use Wrangler or the API.  There's nothing inherent about Wrangler or the API that avoids concurrent writes..

products/workers/src/content/runtime-apis/kv.md
```diff
@@ -42,33 +42,35 @@ The maximum size of a value is 25 MiB.
...
+
+A `key` and `value` are required for each KV pair. The entire request size must be less than 100 megabytes. As of January 2022, Cloudflare does not support bulk writes from within a Worker.
 
 #### Expiring keys
 
-Many common uses of Workers KV involve writing keys that are only meant to be valid for a certain amount of time. Rather than requiring applications to remember to delete such data at the appropriate time, Workers KV offers the ability to create keys that automatically expire, either at a particular point in time or after a certain amount of time has passed since the key was last modified.
+Many common uses of Workers KV involve writing keys that are only meant to be valid for a certain amount of time. Rather than requiring applications to remember to delete such data at the appropriate time, Workers KV offers the ability to create keys that automatically expire. Expiration can occur either at a particular point in time or after a certain amount of time has passed since the key was last modified.
```
koeninger: I'd say "Expiration can be specified to occur", as it is it's not clear whether I have control over it

products/workers/src/content/runtime-apis/kv.md
```diff
@@ -121,39 +127,41 @@ You can [read key-value pairs from the command line with Wrangler](/cli-wrangler
...
 #### Metadata
 
 You can get the metadata associated with a key-value pair alongside its value by calling the `getWithMetadata` method on a namespace you have bound in your script:
 
-```
-const valueAndMetadata = await NAMESPACE.getWithMetadata(key)
-const value = valueAndMetadata.value
-const metadata = valueAndMetadata.metadata
+```js
+const { value, metadata } = await NAMESPACE.getWithMetadata(key)
```
koeninger, line 164: In the past we explicitly avoided destructuring assignment in docs to avoid confusing people who weren't familiar with it; are we walking that decision back?

marciocloudflare (approved):

products/workers/src/content/runtime-apis/kv.md
lukeed:
```suggestion
Due to the eventually consistent nature of Workers KV, concurrent writes from different edge locations can end up overwriting one another. It is a common pattern to write data via Wrangler or the API so that all writes initiate from a single source. This avoids competing, concurrent writes because of the single stream. All data is still readily available within all Workers bound to the namespace.
```

products/workers/src/content/runtime-apis/kv.md
lukeed:
```suggestion
Many common uses of Workers KV involve writing keys that are only meant to be valid for a certain amount of time. Rather than requiring applications to remember to delete such data at the appropriate time, Workers KV offers the ability to create keys that automatically expire. You may configure expiration to occur either at a particular point in time or after a certain amount of time has passed since the key was last modified.
```

products/workers/src/content/runtime-apis/kv.md
lukeed: Yes, the code snippets are all using modern JS (async/await et al) and the assumption is that JS developers already know basic syntax. A single level of destructuring is easily understood. Also the entire docs will have a unified code style, which actually uses ES2015+ language features since we already make that assumption for using common Workers/DO aspects.

products/workers/src/content/runtime-apis/kv.md
koeninger:
```suggestion
Due to the eventually consistent nature of Workers KV, concurrent writes can end up overwriting one another. It is a common pattern to write data from a single process via Wrangler or the API. This avoids competing, concurrent writes because of the single stream. All data is still readily available within all Workers bound to the namespace.
```

koeninger (approved):

PR #3324, opened 2022-02-01
Add a package

janat08:
I've packaged this with some bugfixes. My package isn't production-ready, but it's definitely better than what's in this tutorial.


kristianfreeman:
Thanks for the submission and glad you were able to find a cool alternative using Durable Objects. We've pulled this tutorial from the docs temporarily while we improve some of the content, so we won't be accepting PRs to it at the moment. Best of luck with your NPM package!

PR #3331, opened 2022-02-02
[MFW] Added disabled rule section

dcpena:
Added content for creating disabled rules. 

Addresses [PCX-3269](https://jira.cfops.it/browse/PCX-3269).

### DIFF
--- a/products/magic-firewall/src/content/how-to/add-rules.md
+++ b/products/magic-firewall/src/content/how-to/add-rules.md
@@ -17,6 +17,14 @@ To view examples for root rulesets, review the the [Magic Firewall Terraform doc
 1. Fill out the information for your new rule.
 1. When you are done, click **Add new rule**.
 
+## Create a disabled rule
+
+When you add a new rule, the rule is **Enabled** by default. 
+
+To create a **Disabled** rule, follow the steps in Add a rule above and toggle **Enabled** to off. When a rule is in the disabled state, the rule will not perform the action until is set to **Enabled**.
+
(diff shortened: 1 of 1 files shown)


products/magic-firewall/src/content/how-to/add-rules.md
```diff
@@ -17,6 +17,14 @@ To view examples for root rulesets, review the the [Magic Firewall Terraform doc
 1. Fill out the information for your new rule.
 1. When you are done, click **Add new rule**.
 
+## Create a disabled rule
+
+When you add a new rule, the rule is **Enabled** by default. 
+
+To create a **Disabled** rule, follow the steps in Add a rule above and toggle **Enabled** to off. When a rule is in the disabled state, the rule will not perform the action until is set to **Enabled**.
```
abracchi-tw: Should we have some sort of formatting around `Add a rule` since it's a title?

abracchi-tw (approved):

products/magic-firewall/src/content/how-to/add-rules.md
dcpena: I was thinking the same. Should I make it a link to the section above? I don't know if that would be confusing though because it might look like nothing's happened because it's a section right above.

I considered bold, but I used it for label names so that felt incorrect too. Open to suggestions!

abracchi-tw: Yeah I've used links in these circumstances before! I don't think it looks *too* weird lol

dcpena: Thank you! Updated

arges (approved):

PR #3373, opened 2022-02-07
[Pages] build command + others explainer

deadlypants1973:


products/pages/src/content/get-started/index.md
```diff
@@ -67,6 +69,19 @@ Cloudflare Pages begins by working from your repository's root directory. The en
 
 ![Root directory](./images/root-directory.png)
 
+<details>
+<summary>Build command, build directory and root directory</summary>
+<div>
+
+The build command is provided by your framework. For example, the Gatsby framework uses `gatsby build` as its build command. When you are working without a framework, your build command will be `exit` 0.
```
WalshyDev: We should really move away from suggesting `exit 0` as a build command. Not specifying one is even quicker and just creates a bit less confusion (people ask why and what exit 0 is)

Also, even without a framework there can be a build command. Saying will be is a bit strong worded I think. I'm not sure.

deadlypants1973:
```suggestion
The build command is provided by your framework. For example, the Gatsby framework uses `gatsby build` as its build command. When you are working without a framework, your build command will be `exit 0`. 
```

products/pages/src/content/get-started/index.md
```diff
@@ -67,6 +69,19 @@ Cloudflare Pages begins by working from your repository's root directory. The en
 
 ![Root directory](./images/root-directory.png)
 
+<details>
+<summary>Build command, build directory and root directory</summary>
```
kristianfreeman: Nice, details tags are cool! Have we verified how this looks in the actual docs? Unsure if we styled this tag ever

products/pages/src/content/get-started/index.md
kristianfreeman: I'm good with this. I also find `exit 0` confusing (IIRC, the `0` is an "ok" status to exit with, right?) and I think it's ok to tell people that it can be blank.

WalshyDev (approved):
LGTM, ty!

kristianfreeman (approved):

PR #3377, opened 2022-02-07
Update D.O. counter example to match template

koeninger:


a-robinson (approved):

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -127,7 +125,7 @@ export class Counter {
...
+The Durable Object's storage has a built-in in-memory cache of its own – if you `get()` a value that was read or written recently, the result will be instantly returned from cache. So, instead of writing initialization code like above, you could simply `get("value")` whenever you need it, and rely on the built-in cache to make this fast. Refer to the [Counter example](#example---counter) below for an example of this approach. However, in applications with more complex state, explicitly storing state in your object like above may be easier than making storage API calls on every access. Depending on the configuration of your project, write your code in the way that is easiest for you.
```
deadlypants1973:
```suggestion
The Durable Object's storage has a built-in in-memory cache of its own – if you `get()` a value that was read or written recently, the result will be instantly returned from cache. Instead of writing initialization code like above, you could `get("value")` whenever you need it, and rely on the built-in cache to make this fast. Refer to the [Counter example](#example---counter) below for an example of this approach. However, in applications with more complex state, explicitly storing state in your object like above may be easier than making storage API calls on every access. Depending on the configuration of your project, write your code in the way that is easiest for you.
```

products/workers/src/content/learning/using-durable-objects.md
```diff
@@ -417,61 +415,59 @@ We've included complete example code for both the Worker and the Durable Object
...
+      // Just serve the current value.
+      break;
+    default:
+      return new Response("Not found", {status: 404});
     }
+
+    // We don't have to worry about a concurrent request having modified the
+    // value in storage because "input gates" will automatically protect against
+    // unwanted concurrency. So, read-modify-write is safe. For more details,
+    // see: https://blog.cloudflare.com/durable-objects-easy-fast-correct-choose-three/
```
deadlypants1973:
```suggestion
    // You do not have to worry about a concurrent request having modified the
    // value in storage because "input gates" will automatically protect against
    // unwanted concurrency. So, read-modify-write is safe. For more details,
    // refer to: https://blog.cloudflare.com/durable-objects-easy-fast-correct-choose-three/
```

deadlypants1973 (approved):
PCX review

products/workers/src/content/learning/using-durable-objects.md
kentonv: Nitpick, not actually asking that we change anything, but:

I actually think "We" is more appropriate in this case because this is describing the thinking of the authors of this code, about the code itself. It's saying "we wrote this code this way because of this".

"You" can also make sense in the sense of "when you write your own code, here's what you should think about", but that's not the meaning I intended when writing the comment. To put it another way, if this were a comment in real production code, rather than a tutorial example, "you" wouldn't make sense here.

PR #3378, opened 2022-02-07
WDC-457: Document fetch credential header handling on cross-origin redirects

little-gems:
While we are currently following the `fetch` spec, I believe it would be helpful to document this behavior.


little-gems:
Context in https://github.com/node-fetch/node-fetch/pull/1449

products/workers/src/content/runtime-apis/request.md
```diff
@@ -81,15 +81,16 @@ let request = new Request(input [, init])
 
 - `headers` <Type>Headers</Type> <PropMeta>optional</PropMeta>
 
-  - A [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers).
+  - A [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers). Headers explicitly added to `fetch(url)` will always be sent, even in the case that a redirect is followed. Workers does not implement Cross-Origin Resource Sharing (CORS), as this feature is designed for browsers to decide when to implicitly send credentials like `Cookie` and `Authorization` to a origin that allows them in its CORS policy. Since Workers does not store nor implicitly add credential headers, explicitly defined `Cookie` and `Authorization` headers will be forwarded when a redirect occurs, even in a cross-origin context.
```
ejcx: Just squabbling about words here.

"Workers does not implement Cross-Origin Resource Sharing" I think is a little unclear since this is worker to origin behavior not browser to worker behavior. I might propose the alternate:

> - A [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers). Headers explicitly added to `fetch(url)` will always be sent, even in the case that a redirect is followed. ~Workers does not implement Cross-Origin Resource Sharing (CORS), as this feature is designed for browsers to decide when to implicitly send credentials like `Cookie` and `Authorization` to a origin that allows them in its CORS policy.~ Since Workers does not store nor implicitly add credential headers, explicitly defined `Cookie` and `Authorization` headers will be forwarded when a redirect occurs, even in a cross-origin context.

kentonv: It feels slightly weird to have almost the whole paragraph describing `headers` be about one obscure edge case.

I wonder if it would make sense to use an aside? `<Aside type="warning">` I think?

products/workers/src/content/runtime-apis/request.md
```diff
@@ -81,15 +81,16 @@ let request = new Request(input [, init])
...
+
 
 - `body` <Type>string | ReadableStream | FormData | URLSearchParams</Type> <PropMeta>optional</PropMeta>
 
   - The request body, if any.
 
 - `redirect` <Type>string</Type> <PropMeta>optional</PropMeta>
 
-  - The redirect mode to use: `follow`, `error`, or `manual`. The default  for a new `Request` object is `follow`. Note, however, that the incoming `Request` property of a `FetchEvent` will have redirect mode `manual`.
+  - The redirect mode to use: `follow`, `error`, or `manual`. The `fetch` method only attempts to follow redirects if the redirect mode is set to `follow` rather than `manual`. The default  for a new `Request` object is `follow`. Note, however, that the incoming `Request` property of a `FetchEvent` will have redirect mode `manual`.
```
kentonv: The word "only" here makes it sound like this isn't the default, contradicting the next sentence. Maybe:

> The `fetch` method will automatically follow 3xx redirects if the redirect mode is set to `follow`. If set to `manual`, the 3xx redirect response will be returned to the caller as-is.

products/workers/src/content/runtime-apis/request.md
little-gems: Agreed, looks better:

image

products/workers/src/content/runtime-apis/request.md
```diff
@@ -81,15 +81,15 @@ let request = new Request(input [, init])
 
 - `headers` <Type>Headers</Type> <PropMeta>optional</PropMeta>
 
-  - A [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers).
+  - A [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers). <Aside type="warning"> Headers explicitly added to `fetch(url)` will always be sent, even in the case that a redirect is followed. Since Workers does not store nor implicitly add credential headers, explicitly defined `Cookie` and `Authorization` headers will be forwarded when a redirect occurs, even in a cross-origin context.</Aside>
```
kentonv: Bikeshedding a bit, what do you think of:

```suggestion
  - A [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers). Note that, compared to browsers, Cloudflare Workers imposes very few restrictions on what headers you are allowed to send. For example, a browser will not allow you to set the `Cookie` header, since the browser is responsible for handling cookies itself. Workers, however, has no special understanding of cookies, and treats the `Cookie` header like any other header. <Aside type="warning">In the event that the response is a redirect, and the redirect mode is set to `follow` (see below), then all headers will be forwarded to the redirect destination, even if the destination is a different hostname or domain. This includes sensitive headers like `Cookie`, `Authorization`, or any application-specific headers. If this is not the behavior you want, you should set redirect mode to `manual` and implement your own redirect policy. Note that redirect mode defaults to `manual` for requests that originated from the Worker's client, so this warning only applies to `fetch()`es made by a Worker that are not proxying the original request.</Aside>
```

little-gems: I think that refactor is a good compromise from my original words and more directly addresses the intent of this change (clarification about how explicitly adding headers to our `fetch()` implementation works compared to browsers).

You do say "very few restrictions" -> I'm assuming this is referring to the IP headers? 

image

it did look a bit busy like before so I moved it into a note component:

image

ejcx:
LGTM, don't think I'm a codeowner though

products/workers/src/content/runtime-apis/request.md
kentonv: I wasn't referring to the IP headers, no. I was referring to the list of headers that browsers don't normally allow to be set using `fetch()`, like `Cookie`, `User-Agent`, `Date`, ...

I'm OK with either of those two screenshots.

kentonv (approved):

eidam (approved):

products/workers/src/content/runtime-apis/request.md
```diff
@@ -81,16 +81,15 @@ let request = new Request(input [, init])
...
 
 - `body` <Type>string | ReadableStream | FormData | URLSearchParams</Type> <PropMeta>optional</PropMeta>
 
   - The request body, if any.
 
 - `redirect` <Type>string</Type> <PropMeta>optional</PropMeta>
 
-  - The redirect mode to use: `follow`, `error`, or `manual`. The `fetch` method only attempts to follow redirects if the redirect mode is set to `follow` rather than `manual`. The default  for a new `Request` object is `follow`. Note, however, that the incoming `Request` property of a `FetchEvent` will have redirect mode `manual`.
+  - The redirect mode to use: `follow`, `error`, or `manual`. The `fetch` method will automatically follow redirects if the redirect mode is set to `follow`. If set to `manual`, the 3xx redirect response will be returned to the caller as-is. The default for a new `Request` object is `follow`. Note, however, that the incoming `Request` property of a `FetchEvent` will have redirect mode `manual`.
```
deadlypants1973:
```suggestion
  - The redirect mode to use: `follow`, `error`, or `manual`. The `fetch` method will automatically follow redirects if the redirect mode is set to `follow`. If set to `manual`, the `3xx` redirect response will be returned to the caller as-is. The default for a new `Request` object is `follow`. Note, however, that the incoming `Request` property of a `FetchEvent` will have redirect mode `manual`.
```

PR #3405, opened 2022-02-09
add 270 second idle timeout information

chris-martinelli:
Is this the best place for this information I wonder?


products/cloudflare-one/src/content/faq/teams-troubleshooting.md
```diff
@@ -119,3 +119,7 @@ There are a few different possible root causes behind the `websocket: bad handsh
 * WebSockets are not enabled. To enable them, navigate to `dash.cloudflare.com` > **Network**.
 * Your Cloudflare account has Universal SSL enabled and the SSL/TLS encryption mode is set to *Off*. To resolve, set the SSL/TLS encryption mode to any setting other than *Off*.
 * Your requests are blocked by [Super Bot Fight Mode](https://developers.cloudflare.com/bots/get-started/pro). To resolve, make sure you set **Definitely automated** to *Allow* in the bot fight mode settings.
+
+## Connections are timing out after 270 seconds
+
+We have a 270 second idle timeout on TCP connections that go through our gateway. If no data is sent, the connection is dropped. This is not mitigated by Keep-Alive packets because we're terminating TCP in the gateway and then making a new connection to the upstream. The proxy process drops the connection if there's no new data to send in either direction in 270 seconds.
```
abelinkinbio: ## TCP Connection Timeout

We enforce a 270 second idle timeout on TCP connections that go through our gateway. If no data is sent, the connection is dropped. This cannot be mitigated by Keep-Alive packets as we terminate TCP in the gateway and then make a new connection to the upstream sever. The proxy process drops the connection if there's no new data to send in either direction after 270 seconds.

abracchi-tw (approved):

PR #3468, opened 2022-02-14
Remove a duplicate question and its answer

hunts:
The 7th question `Does DNS Firewall cache responses with status SERVFAIL?` is actually same as the 3rd one `Does the DNS Firewall cache SERVFAIL?`. 

I think we can remove one of them.


ghost (approved):

kodster28 (approved):
Oops! thanks for the find @hunts

PR #3470, opened 2022-02-14
Fix broken automatic builds link

alexrutar:
The link to the automatic builds link is broken - this updates the link to the correct version.


kodster28 (approved):
Thanks for the submission, @alexrutar

PR #3471, opened 2022-02-15
Update index.md

claudiobrandt:
As it is, the example uses `browserPerf1mGroups` and returns an error:

````
{
  "data": null,
  "errors": [
    {
      "message": "unknown field \"browserPerf1mGroups\"",
      "path": null,
      "extensions": {
        "timestamp": "2022-02-14T16:22:35.585278884Z"
      }
    }
  ]
}
```

The proposed change replaces `browserPerf1mGroups` with `firewallEventsAdaptive`, which returns the expected zone limits.
Please note that if `browserPerf1mGroups` was deprecated, the table at the top of the page should also be edited to signal deprecation with `*`


angelampcosta:
@claudiobrandt Thank you for bringing this to our attention. We're actually in the process of addressing this issue and you should see updated content soon.

PR #3590, opened 2022-02-21
add service auth error to troubleshooting page

chris-martinelli:
added error "0: Bad Request. Please create a ca for application."


abracchi-tw:
Will add this content as part of #3595, thank you @chris-martinelli

PR #3612, opened 2022-02-24
Update parameters.md

nnmalex:
Added JA3 fingerprint as a characteristic.


pedrosousa:
Hi @nnmalex 
Thanks! The same update you were proposing was handled as part of a bigger PR related to the recent rate limiting announcements — PR #3826.

PR #3614, opened 2022-02-24
Updated Linux install command

siv12345:
The pipe symbol was missing for the linux install command, which caused cloudflared to not be installed.


lukeed (approved):

ranbel:
Thanks for the feedback! This issue was addressed in #3905.

PR #3619, opened 2022-02-25
[DNS] Added note about windows authentication limits

kodster28:
Addresses PCX-3420


kodster28:
Changes will be addressed by #3624

PR #3633, opened 2022-02-26
Fix MD errors in guides for popular servers

javierruizjimenez:
Markdown links were broken


kodster28:
Actually, addressing this in #3639

PR #3638, opened 2022-02-28
Fix broken link

mattferderer:
Bulk Redirects link is a 404. I'm taking a guess at the above being the correct link that should be linked to.

### DIFF
--- a/content/pages/how-to/www-redirect.md
+++ b/content/pages/how-to/www-redirect.md
@@ -9,7 +9,7 @@ In this tutorial, you will learn how to redirect the `www` subdomain to your ape
 
 {{<Aside type= "note">}}
 
-Although it is generally recommended to use [Bulk Redirects](/rules/bulk-redirects/_index.md) when redirecting between domains, for smaller use cases, such as a single redirect from `www.example.com` to `example.com`, a single Page Rule will work just as well.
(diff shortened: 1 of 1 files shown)


content/pages/how-to/www-redirect.md
```diff
@@ -9,7 +9,7 @@ In this tutorial, you will learn how to redirect the `www` subdomain to your ape
 
 {{<Aside type= "note">}}
 
-Although it is generally recommended to use [Bulk Redirects](/pages/use-bulk-redirects.md/) when redirecting between domains, for smaller use cases, such as a single redirect from `www.example.com` to `example.com`, a single Page Rule will work just as well.
+Although it is generally recommended to use [Bulk Redirects](/rules/bulk-redirects/_index.md) when redirecting between domains, for smaller use cases, such as a single redirect from `www.example.com` to `example.com`, a single Page Rule will work just as well.
```
kodster28:
```suggestion
Although it is generally recommended to use [Bulk Redirects](/pages/how-to/use-bulk-redirects/) when redirecting between domains, for smaller use cases, such as a single redirect from `www.example.com` to `example.com`, a single Page Rule will work just as well.
```

kodster28 (approved):

kodster28:
Howdy @mattferderer, updated the link to the right target, but thanks for the call out!

PR #3642, opened 2022-02-28
[ZT] fix how-to numbering

abracchi-tw:


kodster28 (approved):

lukeed (changes requested):
The numbering should rise sequentially. It will be reverted on the next `format` script pass anyway.

The numbers that are currently shown are how the markdown engine are interpreting the markdown content... and are correct. If lists reset or have the wrong count, then it's an indentation issue (which is very common to our files, unfortunately)

PR #3644, opened 2022-02-28
Document websocket message size limit

koeninger:


a-robinson (approved):
This applies to websockets in general, not just for durable objects, so perhaps it should be mentioned on https://developers.cloudflare.com/workers/learning/using-websockets/ as well. But this is fine by itself too.

For what it's worth, based on the code it looks like the limit only applies to messages received by a worker/DO rather than messages sent by a worker or proxied through a worker.

PR #3651, opened 2022-02-28
chore: batch `format` changes

lukeed:
Ran `format` on the project and kept/fine-tuned desired changes.

However, during this pass, I noticed that `format` applies some unwanted **content changes** because Prettier and Hugo are speaking slightly different dialects of markdown. Because of this, I've (temporarily) removed `format` as a npm-script and removed it as a pre-deploy step.


jkup (approved):
Change makes sense to me. I picked a few random pages in the diff and they all look good to me!

PR #3666, opened 2022-03-02
[KV] Document maximum key length

bkadatz:
Attempting to access KV with a key size over 512 bytes results in an error such as:
```
Error: KV GET failed: 414 UTF-8 encoded length of 513 exceeds key length limit of 512.
```


koeninger:
Limits are already documented here https://developers.cloudflare.com/workers/platform/limits/#kv-limits but seems fine to add here as well, since we're already discussing max value size.

koeninger (approved):

a-robinson (approved):

PR #3681, opened 2022-03-02
Migrate search to Coveo FedSearch in docs

eidam:
Deploy preview URL: https://a5537109.cloudflare-docs-7ou.pages.dev

### DIFF
--- a/assets/search.redirect.ts
+++ b/assets/search.redirect.ts
@@ -15,7 +15,7 @@
 
       if (text.length > 0) {
         redirect += '#q=' + encodeURIComponent(text);
-        redirect += '&f:source=[Developer%20docs]';
+        // redirect += '&f:source=[Developer%20docs]';
       }
 
       return location.assign(redirect);
--- a/assets/search.redirect.ts
+++ b/assets/search.redirect.ts
@@ -1,0 +1,31 @@
+(function () {
+  let $ = document.querySelector.bind(document);
+  let element = $('#DocsSearch--input') || $('#SiteSearch--input');
+
+  addEventListener('keydown', ev => {
+    let key = ev.which;
+
+    // inside search input element
+    if (ev.target === element) {
+      if (key !== 13) return;
+
+      // ENTER ~> redirect
+      let redirect = '/search/';
+      let text = (element.value || '').trim();
+
+      if (text.length > 0) {
+        redirect += '#q=' + encodeURIComponent(text);
+        redirect += '&f:source=[Developer%20docs]';
+      }
+
+      return location.assign(redirect);
+    }
+
+    // is '/' or SHIFT+'s'
+    if (key === 191 || (ev.shiftKey && key === 83)) {
+      ev.preventDefault();
+      window.scrollTo(0, 0);
+      element.focus();
+    }
+  });
+})();
--- a/assets/main.ts
+++ b/assets/main.ts
@@ -7,6 +7,7 @@ declare global {
   interface Window {
     // algolia; @see search.ts
     docsearch?(options: any): any;
+    Coveo?: any;
   }
 }
 
--- a/assets/search-coveo.ts
+++ b/assets/search-coveo.ts
@@ -1,0 +1,76 @@
+let tag = document.currentScript;
+let $ = document.querySelector.bind(document);
+let coveo = window.Coveo
+
+let dataset = tag && tag.dataset;
+let { org, token } = dataset || {};
+
+function loadCustomSearchBox() {
+  let element = $('#DocsSearch--input') || $('#SiteSearch--input');  
+  const CustomSearchbox = (function(_super) {
+    __extends(CustomSearchbox, coveo.Component);
+    function CustomSearchbox(element, options, bindings) {
+      _super.call(this, element, CustomSearchbox.ID, bindings);
+      this.type = 'CustomSearchBox';
+      coveo.Component.bindComponentToElement(element, this);
+      this.element = element;
+      this.options = coveo.ComponentOptions.initComponentOptions(element, CustomSearchbox, options);
+      this.bindings = bindings;
+      this.element.addEventListener('keyup', (e) => this.handleKeyUp(e));
+    }
+    CustomSearchbox.prototype.handleKeyUp = function(e) {
+      if (this.options.searchAsYouType) {
+        this.executeNewQuery();
+      } else if (e.key == 'Enter') {
+        this.executeNewQuery();
+      }
+    }
+    CustomSearchbox.prototype.executeNewQuery = function() {
+      this.bindings.queryStateModel.set('q', this.element.value);
+      //this.bindings.usageAnalytics.logSearchEvent({
+      //  name: 'submitSearchbox',
+      //  type: 'CustomSearchbox'
+      //});
+      this.bindings.queryController.executeQuery();
+    }
+    CustomSearchbox.options = {
+      searchAsYouType: coveo.ComponentOptions.buildBooleanOption({ defaultValue: false })
+    }
+    CustomSearchbox.ID = "CustomSearchBox";
+    coveo.Initialization.registerAutoCreateComponent(CustomSearchbox);
+  })(coveo.Component);
+
+  coveo.SearchEndpoint.configureCloudV2Endpoint(org, token);
+  coveo.initSearchbox($('.CoveoSearchInterface'), "/search")
+
+  addEventListener('keydown', ev => {
+    if (ev.target === element) return;
+
+    let key = ev.which;
+
+    // is '/' or SHIFT+'s'
+    if (key === 191 || (ev.shiftKey && key === 83)) {
+      ev.preventDefault();
+      window.scrollTo(0, 0);
+      element.focus();
+    }
+  });
+}
+
+function loadSearchResults() {
+  // The following line shows you how you could configure an endpoint against which to perform your search.
+  coveo.SearchEndpoint.configureCloudV2Endpoint(org, token);
+
+  // Initialize the framework by targeting the root in the interface.
+  // It does not have to be the document body.
+  coveo.init(document.getElementById('searchresults'));
+}
+
+// init
+(function load() {
+  if (window.location.pathname.startsWith("/search")) {
+    loadSearchResults();
+  } else {
+    loadCustomSearchBox();
+  }
+})();
--- a/assets/search-coveo.ts
+++ b/assets/search-coveo.ts
@@ -1,76 +1,82 @@
-let tag = document.currentScript;
-let $ = document.querySelector.bind(document);
-let coveo = window.Coveo
+(function () {
+  let tag = document.currentScript;
+  let $ = document.querySelector.bind(document);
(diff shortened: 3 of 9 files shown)


lukeed:
Deploy Preview is here: https://c8b02ea7.cloudflare-docs-7ou.pages.dev/
(Pages build finished before PR was opened, hence no bot comment)

eidam:
sorry, forgot to mention this is still WIP - waiting for UAT results, but any upcoming changes would be styles only - so feel free to review the rest

lukeed (changes requested):
The integration code is fine, but I have a few UX problems with Coveo that I personally consider this to be a blocker:

1. The search results are absolutely awful. the current production site's results are useful & even offer sub-section result matches. These Coveo results can't even find entire pages.
2. The search input isn't interactive. This is a major let-down & loss compared to what we have. 
3. The JS library loaded on every page is 200+ kB & it can't possibly be doing anything since there's no interactivity or anything... Pressing <kbd>ENTER</kbd> is effectively a redirect to `/search/#q=KEYWORDS` – why is this script needed?

---

And not necessarily blockers, but:

1. Searching should pre-check the `Developer docs` results filter. Users are coming from documentation & tying to search within the documentation, so this is a safe assumption to make
2. The `Developer docs` label should be renamed to `Documentation`
3. Clicking a `Documentation` result link **should not be hardcoded** to the `developers.cloudflare.com` hostname. I recognize this may not be possible, but it's also a loss as it means reviews can go from a preview deployment (`[hash].cloudflare-docs.pages.dev`) to the production site without realizing

kodster28:
Hey @lukeed, a couple responses to your comments.

Blockers:
1. Once we get in https://jira.cfops.it/browse/WDC-508, I think that'll address some of your concern, as well as non-blocker 1 We can key off your search context to promote that info more heavily. As to the other piece... we're trying to intentionally stay a more unopinionated during the initial rollout so we can get more accurate data to feed Coveo's built-in M/L models. Once those start kicking in, they should help tailor the results. I know that isn't a super satisfying answer, but we should be able to fast-follow to make the results significantly better. The other benefit -- probably bigger for other doc areas than for workers -- is that the new search allows us to pull in content from the Learning Center, Blog, and KB.
2. (also relates to 3) There's no interactivity at the moment of searching b/c we don't have enough data to safely populate their query suggestion M/L model. It basically auto-fills with suggested queries based on past user inputs. We actually went live w/ that functionality in the KB and... uh... basically folks started hammering our search with inappropriate phrases to the point that it was auto-completing phrases like `sexy videos` and others. Again, understand that it isn't super satisfying, but it should get corrected soonish (we also have some logic to try and avoid getting NSFW queries put into our data in the first place)
3. The JS library is used for the query suggestion component. So it's not technically necessary now... but it's what pre-populates search queries.

Non-blockers:
1. Valid point. I'd prefer the slightly softer approach of weighting dev doc content a bit heavier (instead of just excluding everything else). Might just be a personal preference... but I can envision folks not realizing that a filter was pre-selected and getting frustrated. I think this'll come when we address https://jira.cfops.it/browse/WDC-508.
2. We can definitely change it. At this point, we sorta have to standardize that verbiage across the other domains, so it might require a bit of communication to do so.
3. Can't really speak to the tech side, but definitely a valid point.

lukeed:
Thanks @kodster28 

Until that ticket lands, I don't think this can be merged or even reviewed fairly. The search experience is so degraded right now that it dramatically impairs the entire devdocs experience. Adding in a few results from other Cloudflare properties in no way justifies losing the ability to search – accurately and/or at all – our ~1600 documentation pages.

Without good search, content remains hidden unless & until a user takes the time to painstakingly comb through our sidebars & keep the content organization/ladder memorized for future visits.

lukeed:
Did something change in the last hour? Previously I had 4 DevDocs results when searching for "logging workers" and none of them were the [Logging from Workers](https://developers.cloudflare.com/workers/learning/logging-workers/) page I was looking for. (similar experience with `ab test` and other terms.) Now there are 28 and the desired page is the 1st result.... much better & negates my previous comment and my Blocker-1 point!

kodster28:
Ah.... so I think the change in sitemaps may have basically excluded dev docs from our index. I changed that earlier in the day, but maybe it's something to do with caching?

As in, it's fixed in Coveo now.

kodster28:
Actually, @lukeed, do you know why those changes would've just appeared now? Just wondering b/c I looked back and the re-indexing took place ~4 hours ago... but then didn't appear in your testing.

That would be a potential concern for pushing live.

lukeed:
I noticed the poor results yesterday & verified they were still occurring when I left my review today. Nothing different on my end & it was always same search terms /shrug

Do you have the updated sitemap? Not sure what you need from me.

Also, since the dropdown will come in a later iteration, can we move the Coveo JS out of this PR? 
Using this [example production page](https://developers.cloudflare.com/workers/learning/logging-workers/), there is:

* a total of 308kB total transferred
* where 58.4kB of that is JS 
* and 20kB is CSS.

The same page with [this Coveo PR](https://c8b02ea7.cloudflare-docs-7ou.pages.dev/workers/learning/logging-workers/) bumps it up to:

* **555 kB** total transfer
* **241 kB** of JS 
* **71 kB** of CSS

I know this is my personal quest, but I find this concerning... 4x'ing our JS & 3.5X'ing our CSS for what's effectively a redirect. I know it's prep-work for future dropdown, but this is a massive (performance) price to pay for what's essentially some keyboard listeners, an HTTP call, and some DOM updates. Does Coveo offer anything more sensible?

lukeed:
The Lighthouse performance score for the live/production `Logging from Workers` page is **98**

The score for the same page in the Coveo version is **74** — this will extend to **all** Dev Doc pages

IMO this verifies my concern w/ their SDK files

Screen Shot 2022-03-03 at 12 18 39 PM

Screen Shot 2022-03-03 at 12 18 29 PM

kodster28:
I think that's a valid concern on the script and -- since we aren't rolling it out yet -- it seems like it'd make sense to keep it off. Especially b/c it's basically erasing most of the optimization y'all did!!

We can also add that (and potentially you or another dev advocate) to our next meeting with them to pass along that feedback.

kristianfreeman:
> Also, since the dropdown will come in a later iteration, can we move the Coveo JS out of this PR?

Agreed on this -- since we aren't using any live functionality on most pages, I think we should remove any JS we can at the moment, and just have it redirect to that query page when someone types in a search query.

In the future, we'll obviously want to include dropdown and proper "on-page" search functionality. At that point, we should investigate if there's optimizations we can do in order to lessen the performance implications on each page across the docs. That work could probably also benefit other CF properties (MRK, support, etc.) as well, so it's definitely worth doing.

@eidam -- can you update the PR accordingly with @lukeed's suggested change? To meet our deadlines, we still want to move on this PR ASAP; I agree with @kodster28's point that any issues we may have with functionality should probably be brought up in a future group meeting for the project.

lukeed:
I've added the necessary changes to defer the new script. Will remeasure the lighthouse impact after build completes.

From there, if it's still significant, I will remove the SDK files entirely & replace with a redirect for the time being until dropdown-version is ready. 

I don't know who Coveo is or where they came from, but since this is here to stay, I'd **much rather not** use their SDK files _at all_ and apply equivalent functionality here at 1/20th the cost.

lukeed:
Screen Shot 2022-03-03 at 12 44 50 PM

> Source: https://cbe59fa8.cloudflare-docs-7ou.pages.dev/workers/learning/logging-workers/

lukeed:
In my last commit, I removed Coveo SDK files everywhere except the `/search/` page. Instead, all other pages' search input will redirect to the `/search/` page (which was already the behavior).

And I pre-selected the `Developer Docs` filter, because that makes the most sense & feels the best to me. (Easy to remove)

Will reply with the updated Lighthouse score, but it **should be** a `98` for all pages _except_ the `/search/` page, which should be a `81` now (see previous comment)

kodster28:
This is **super** late notice, so apologies, but we should delay the pushing this live until at least Monday.

Saying that not b/c there's any issues I'm seeing with the code ... but b/c I'm actually moving up to Colorado tomorrow and just found out that my phone isn't eligible to enroll in Okta for access to GChat or gmail. Was really not anticipating that second part :(.

Talking with Jenna, she thinks we should push it back to Monday since I won't be available in case we run into any issues. I wouldn't _think_ we would... but like if something weird happened with the Dev Docs source in pipeline like it did yesterday/today or something else unanticipated, it wouldn't be the best.

Also, confirmed from her side that the Friday deadline was flexible.

lukeed:
Screen Shot 2022-03-03 at 1 19 09 PM

> Source: https://cea3ac07.cloudflare-docs-7ou.pages.dev/workers/learning/logging-workers/

It got better because we dropped the ~45kB of JS for the old search too :)

kodster28:
Hey @lukeed or @eidam, before launching this, can we actually remove `Developer docs` as a pre-selected filter?

I agree that it does make sense in the Workers and Pages docs... but there's a lot of other areas where we aren't as... let's say _thorough_ in our documentation. We still have some pretty glaring holes for SSL, Speed features, etc. for example.

It'll actually be incredibly helpful to see when and for what documents folks are leaving dev docs and going to blog, community, and KB. That way, we can identify gaps in our content and prioritize fixes.

assets/search.redirect.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (ev.target === element) {
+      if (key !== 13) return;
+
+      // ENTER ~> redirect
+      let redirect = '/search/';
+      let text = (element.value || '').trim();
+
+      if (text.length > 0) {
+        redirect += '#q=' + encodeURIComponent(text);
+        redirect += '&f:source=[Developer%20docs]';
```
lukeed:
```suggestion
        // redirect += '&f:source=[Developer%20docs]';
```

kodster28:
Oh, also... as a note, if the unfiltered view gets too annoying from a Pages/Workers perspective, we can do some fancy stuff for those docs specifically once we implement https://jira.cfops.it/browse/WDC-508

kodster28 (changes requested):
Before pushing this live, we'll want to test out the Production pipeline and verify that everything looks correct.

Other than that, approved from a PCX perspective (with post-MVP tasks noted in https://jira.cfops.it/browse/WDC-508 and https://jira.cfops.it/browse/WDC-531).

kristianfreeman (approved):
Next steps: @eidam and @kodster28 to update this with production pipeline details tomorrow (3/9) for further testing, and merge later this week.

+1 from an eng perspective on our side

eidam:
@kodster28 config switched to production one

kodster28 (commented):
Approved from a PCX perspective and Coveo analytics perspective.

However, we don't yet have the green light from the higher ups... so we can't push this live quite yet :(.

kodster28 (approved):
Got steerco approval. Ready to go when y'all are!
(will let an actual developer hit the merge button though)

lukeed (approved):

PR #3694, opened 2022-03-03
Update false-positive.md

zeinjaber:
Fixing formatting.


pedrosousa:
Thanks, @zeinjaber , but I'm fixing these and a few more issues in a separate PR -- #3679.

PR #3695, opened 2022-03-03
Update false-negative.md

zeinjaber:
Fixing formatting.


pedrosousa:
Also fixed in PR #3679.
Thanks again, @zeinjaber!

PR #3702, opened 2022-03-03
[Chore]: Css fixes for table hover and link color on dark mode

Ekwuno:
This Pr 

1. Removes the hover effect we have on tables in the doc. 
2. Fixes link color in dark mode

Links have very low contrast in a dark mood. This PR fixes that.
Screenshot 2022-03-03 at 12 35 42
Screenshot 2022-03-03 at 12 35 32


eidam (approved):

pedrosousa:
Nit: In light mode, the link style is not the same inside and outside tables:

image

PR #3765, opened 2022-03-09
Fix small typo

GiladE:

- https://developers.cloudflare.com/ssl/origin-configuration/ssl-modes/#use-when-3

- `Encryption modes` >`Full (strict)` > `Use When`

### What needs to change?

- "... unless **your** are an ..."

### How should it change?

- "... unless **you** are an ..."

### Additional information

- NA


kodster28 (approved):

kodster28:
Thanks for the fix, @GiladE!

PR #3784, opened 2022-03-10
[Pages] redirect clarification

deadlypants1973:


content/pages/how-to/use-bulk-redirects.md
```diff
@@ -11,6 +11,8 @@ In this tutorial, you will learn how to use [Bulk Redirects (beta)](/rules/bulk-
 
 The redirects defined in the `_redirects` file of your build folder can work together with your Bulk Redirects. In case of duplicates, Bulk Redirects will run in front of your Pages project, where your other redirects live.
 
+For example, if you have Bulk Redirects set up to direct `abc.com` to `xyz.com` but also have `_redirects` set up to direct `xyz.com` to `foo.com`, the request will resolve to `foo.com`.
```
GregBrimble:
```suggestion
For example, if you have Bulk Redirects set up to direct `abc.com` to `xyz.com` but also have `_redirects` set up to direct `xyz.com` to `foo.com`, a request for `abc.com` will eventually redirect to `foo.com`.
```

marciocloudflare (approved):

content/pages/how-to/use-bulk-redirects.md
GregBrimble: "Resolve" makes it sound like we're flattening it, which we we're not. The client will be served a redirect to `xyz.com`. A subsequent request to `xyz.com` will then serve another redirect to `foo.com`.

GregBrimble (changes requested):

GregBrimble (approved):

PR #3814, opened 2022-03-15
[Workers] Document `minimal_subrequests` compatibility flag

harrishancock:
This new flag is live, and will become default next month.

### DIFF
--- a/content/workers/platform/compatibility-dates.md
+++ b/content/workers/platform/compatibility-dates.md
@@ -73,9 +73,9 @@ Newest changes are listed first.
   </tbody>
 </table>
 
-With the `minimal_subrequests` flag set, `fetch()` subrequests sent to endpoints on the Worker's own zone (also called "same-zone subrequests") have a reduced set of features applied to them. In general, these features should not have been applied to same-zone subrequests in the first place, and very few user-facing behavior changes are anticipated. Specifically, scripts might observe the following behavior changes with the new flag:
+With the `minimal_subrequests` flag set, `fetch()` subrequests sent to endpoints on the Worker's own zone (also called same-zone subrequests) have a reduced set of features applied to them. In general, these features should not have been applied to same-zone subrequests in the first place, and very few user-facing behavior changes are anticipated. Specifically, Workers might observe the following behavior changes with the new flag:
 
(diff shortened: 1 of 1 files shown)


harrishancock:
@codewithkristian, are you able to see why this build failed? The logs aren't accessible for me.

kentonv (approved):

content/workers/platform/compatibility-dates.md
```diff
@@ -44,6 +44,41 @@ Most developers will not need to use `compatibility_flags`; instead, Cloudflare
...
+        <strong>Flag to disable</strong>
+      </td>
+      <td>
+        <code>no_minimal_subrequests</code>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+With the `minimal_subrequests` flag set, `fetch()` subrequests sent to endpoints on the Worker's own zone (also called "same-zone subrequests") have a reduced set of features applied to them. In general, these features should not have been applied to same-zone subrequests in the first place, and very few user-facing behavior changes are anticipated. Specifically, scripts might observe the following behavior changes with the new flag:
```
deadlypants1973:
```suggestion
With the `minimal_subrequests` flag set, `fetch()` subrequests sent to endpoints on the Worker's own zone (also called same-zone subrequests) have a reduced set of features applied to them. In general, these features should not have been applied to same-zone subrequests in the first place, and very few user-facing behavior changes are anticipated. Specifically, Workers might observe the following behavior changes with the new flag:
```

Removed script because we are phasing that word out.

content/workers/platform/compatibility-dates.md
```diff
@@ -44,6 +44,41 @@ Most developers will not need to use `compatibility_flags`; instead, Cloudflare
...
+
+* Response bodies will not be opportunistically gzipped before being transmitted to the Workers runtime. If a Worker reads the response body, it will read it in plaintext, as has always been the case, so disabling this prevents unnecessary decompression. Meanwhile, if the Worker passes the response through to the client, our HTTP proxy will opportunistically gzip the response body on that side of the Workers runtime instead. The behavior change observable by a Worker script should be that some `Content-Encoding: gzip` headers will no longer appear.
```
deadlypants1973:
```suggestion
* Response bodies will not be opportunistically gzipped before being transmitted to the Workers runtime. If a Worker reads the response body, it will read it in plaintext, as has always been the case, so disabling this prevents unnecessary decompression. Meanwhile, if the Worker passes the response through to the client, Cloudflare's HTTP proxy will opportunistically gzip the response body on that side of the Workers runtime instead. The behavior change observable by a Worker script should be that some `Content-Encoding: gzip` headers will no longer appear.
```

deadlypants1973 (approved):

PR #3834, opened 2022-03-17
Update warp-to-tunnel.md

tenaciousdlg:
Added a description on where to create the config file and what it should be called.

### DIFF
--- a/content/cloudflare-one/tutorials/warp-to-tunnel.md
+++ b/content/cloudflare-one/tutorials/warp-to-tunnel.md
@@ -87,7 +87,7 @@ cloudflared tunnel route ip show
 
 ## Configure and run the Tunnel
 
-Next, create a configuration file for the Tunnel. The following template contains the required fields but can be further modified as needed.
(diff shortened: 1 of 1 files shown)


abracchi-tw:
@tenaciousdlg thanks for this. the config file can actually be named anything — we'll make this clearer across the docs

abracchi-tw (approved):

tenaciousdlg:
Sounds good but we do call it `config.yml` in other docs.
Examples:
https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide/#4-create-a-configuration-file
https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/do-more-with-tunnels/secure-server/
https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/run-tunnel/as-a-service/linux/
and about 19 other instances.

It'd be good to include both pieces of information but from a standardization/ease of search standpoint I'd vote for sticking to a naming convention yet letting clients know it is just a naming convention.

PR #3845, opened 2022-03-17
Add note about why we cannot list specific page rules disabled on Wor…

jkup:
Customer service is still getting a lot of questions on which page rules run before or after Workers. I spoke with them and we thought maybe adding this note with a few snippets of internal documentation would clear things up.

### DIFF
--- a/content/workers/platform/workers-with-page-rules.md
+++ b/content/workers/platform/workers-with-page-rules.md
@@ -20,6 +20,12 @@ When using Page Rules with Workers, the following workflow is applied.
 4.  Page Rules run as part of normal request processing with some features now disabled.
 5.  Worker executes.
 
+{{<Aside type="note" header="Listing specific page rules">}}
+
+A complete list of features which can be configured "before" versus "after" Workers would be difficult to generate, because they are defined implicitly by the features which we disallow. In general, security-related features run "before" Workers, while most others run "after"
+
+{{</Aside>}}
+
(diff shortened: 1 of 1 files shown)


content/workers/platform/workers-with-page-rules.md
```diff
@@ -20,6 +20,12 @@ When using Page Rules with Workers, the following workflow is applied.
 4.  Page Rules run as part of normal request processing with some features now disabled.
 5.  Worker executes.
 
+{{<Aside type="note" header="Listing specific page rules">}}
```
deadlypants1973:
```suggestion
{{<Aside type="note" header="Listing specific Page Rules">}}
```

content/workers/platform/workers-with-page-rules.md
```diff
@@ -20,6 +20,12 @@ When using Page Rules with Workers, the following workflow is applied.
...
+
+A complete list of features which can be configured "before" versus "after" Workers would be difficult to generate, because they are defined implicitly by the features which we disallow. In general, security-related features run "before" Workers, while most others run "after"
```
deadlypants1973:
```suggestion
A complete list of features which can be configured before and after Workers would be difficult to generate, because they are defined implicitly by the features which we disallow. In general, security-related features run before Workers, while most others run after.
```

lukeed: Instead of quotes (or omitting them) the sentence should read / say what it's trying to say in normal English.

Also, what does `implicitly by the features which we disallow` mean? Example of this maybe?

jkup:
@deadlypants1973 @lukeed what if we just simplify it like this?

deadlypants1973:
I am going to close this with the intention of adding actual documentation on the list of disabled page rules.

PR #3854, opened 2022-03-18
Update google-cloud-platform.md

SwissOS:
Removed erroneous ` at end of line 88


kodster28 (approved):
Thanks for the catch, @danielp123

PR #3875, opened 2022-03-21
Update install-cloudflare-cert.md

tenaciousdlg:
Updated Ubuntu 20.04 install instructions as the existing documentation does not work for this distro. Tested this install on Ubuntu ISO home lab and Ubuntu in GCP.


ranbel:
Thank you for bringing this to our attention! Another user experienced the same issue. I have updated the instructions in PR #3894.

PR #3879, opened 2022-03-21
[Magic WAN] Updated example IP addresses

dcpena:
Updated example addresses to use private instead of public IP addresses. Addresses [PCX-3282](https://jira.cfops.it/browse/PCX-3282).


content/magic-wan/how-to/configure-static-routes.md
```diff
@@ -22,14 +22,14 @@ For an example edge routing configuration, refer to the example below.
...
 | --------------------------- | -------------------- | ------------- |
-| GRE_1_IAD                   | **103.21.244.0/24**  | 100           |
-| GRE_2_IAD                   | **103.21.244.0/24**  | 100           |
-| GRE_3_ATL                   | **103.21.244.0/24**  | 100           |
-| GRE_4_ATL                   | **103.21.244.0/24**  | 100           |
-| GRE_1_IAD                   | 103.21.245.0/24      | 200           |
-| GRE_2_IAD                   | 103.21.245.0/24      | 200           |
-| GRE_3_ATL                   | 103.21.245.0/24      | 100           |
-| GRE_4_ATL                   | 103.21.245.0/24      | 100           |
+| GRE_1_IAD                   | **10.10.10.102/31**  | 100           |
```
kodster28, line 25: Is there a reason why these values are bolded, but the subsequent ones are not?

kodster28 (approved):

content/magic-wan/how-to/configure-static-routes.md
dcpena: Yep! The second list item above the example dropdown explains why, which is to highlight that the priority values for the prefix entries match, and Cloudflare will use ECMP packet forwarding to route the traffic.

PR #3895, opened 2022-03-23
`response` is a property of `action_parameters`

jacobbednarz:
Moves the `response` object into `action_parameters` where the payload expects it.


pedrosousa (approved):

pedrosousa:
@jacobbednarz Thank you!
Not sure how the object ended up in the wrong place, but now it's fixed.

jacobbednarz:
🤘 wicked, thanks!

PR #3905, opened 2022-03-23
Fixed linux ​​.deb install command

fbrettnich:


ranbel (approved):
Thanks for your suggestion!

PR #3910, opened 2022-03-23
Bulk Redirects → _redirects

GregBrimble:
- [x] Currently just want to see if `_redirects` works correctly in the preview environment :)~
- [x] Seems to! Now I'll work on exporting the current Bulk Redirects into this `_redirects` format.
- [x] Now just need to ensure the dynamic redirects work correctly.
- [ ] `www` → `@`


kodster28:
Howdy @GregBrimble, tested a couple of the dynamic redirects and they _seem_ to work!

That said, not sure if it's intended, but all the redirects are sending me to the live version of the page (as opposed to the preview)? That functionality is still fine with us TBH, but it sounded like y'all were looking for something different?

GregBrimble:
Good point! I'll strip the host from the destination URLs.

kodster28 (approved):

PR #3930, opened 2022-03-25
Documentation doesn't mention that minimum is default

chaouiy:


angelampcosta:
👋@chaouiy Thank you for the feedback, but we think that our current wording is still accurate. I'm going to close it now, but feel free to open a new issue or ask any questions.

PR #3937, opened 2022-03-28
[ZT] Fix Cloudflare CA URL in GCloud setup

michielappelman:
Also combines the two files with a single call to `cat`.


ranbel (approved):
Thanks for the edits!

PR #3949, opened 2022-03-29
Update config.toml

eidam:
Add a note about read-only search token

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -18,7 +18,7 @@ rssLimit = 10
 paginate = 12 # default 10
 paginatePath = "page" # default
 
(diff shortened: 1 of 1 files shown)


config.toml
```diff
@@ -18,6 +18,7 @@ rssLimit = 10
 paginate = 12 # default 10
 paginatePath = "page" # default
 
+# Coveo is a search solution - following read-only token is used from the frontend to execute search queries
```
pedrosousa:
```suggestion
# Coveo is a search solution - the following read-only token is used from the frontend to execute search queries
```

pedrosousa (approved):

PR #3954, opened 2022-03-29
[Durable Objects] Remote exceptions can be differentiated by property

cacheburner:
Previously, we couldn't determine if an exception thrown to a Worker from a remote Durable Object/Worker was created by the remote or due to a failure in reaching the remote. We can now differentiate these by using the `.remote` property on the exception.

### DIFF
--- a/content/workers/runtime-apis/durable-objects.md
+++ b/content/workers/runtime-apis/durable-objects.md
@@ -396,7 +396,7 @@ let response = await stub.fetch(url, options);
 
 The `fetch()` method of a stub has the exact same signature as the [global `fetch`](/workers/runtime-apis/fetch/). However, instead of sending an HTTP request to the Internet, the request is always sent to the Durable Object to which the stub points.
 
-Any uncaught exceptions thrown by the Durable Object's `fetch()` handler will be propagated to the caller's `fetch()` promise.
+Any uncaught exceptions thrown by the Durable Object's `fetch()` handler will be propagated to the caller's `fetch()` promise. Furthermore, if an uncaught exception is thrown by the Durable Object's `fetch()` handler, then the exception propagated to the caller's `fetch()` promise will include a property `.remote`, which will be set to `True`. If the caller's `fetch()` failed as a result of being unable to reach the Durable Object, the exception thrown to the caller's `fetch()` will not have the `.remote` property, indicating the exception was not generated remotely.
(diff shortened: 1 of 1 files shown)


cacheburner:
@koeninger @a-robinson I wonder if we should also update Worker-to-Worker docs? Though I don't know of a good spot to indicate the changes...

a-robinson:
Looks reasonable to me. I don't have a suggestion on where to put it in worker-to-worker docs, but I'll let @koeninger weigh in on that.

deadlypants1973 (approved):
Approved for style guide requirements but need @greg-mckeon or @koeninger to look at the technical content.

content/workers/runtime-apis/durable-objects.md
```diff
@@ -258,7 +258,7 @@ The system calls the `fetch()` method of a Durable Object namespace when an HTTP
 
 The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter and returns a [`Response`](/workers/runtime-apis/response/) (or a `Promise` for a `Response`).
 
-If the method fails with an uncaught exception, the exception will be thrown into the calling Worker that made the `fetch()` request.
+If the method fails with an uncaught exception, the exception will be thrown into the calling Worker that made the `fetch()` request. Furthermore, if the Durable Object caused the exception, then the resulting exception thrown to the calling Worker will include a property `.remote`, which will be set to `True`. If the method failed as a result of being unable to reach the Durable Object, the resulting exception thrown to the calling Worker will not have the `.remote` property, indicating the exception was not generated remotely.
```
koeninger: Minor, but "if the Durable Object caused the exception" raises some ambiguity in my mind as to whether my code in the DO caused it or DO system outside of my control caused it.

The "Sending HTTP Requests" section below I think is clearer when it says "Any uncaught exceptions thrown by the Durable Object's fetch() handler will be propagated to the caller's fetch() promise."  Maybe this addition would be better made in that section?

koeninger:
I don't have a good suggestion on where to put worker-to-worker related docs, all I was able to find is https://developers.cloudflare.com/workers/learning/using-services/#workers-service-bindings

@kabirsikand separate PR in the future for worker-to-worker?

koeninger (approved):

a-robinson (approved):

PR #3996, opened 2022-04-02
update if check for cookie

awareness481:
The cookie will either be string if a value has been set or `undefined` if it's missing from the Cookie header

You can test it out with the following worker
```ts
import { parse } from 'cookie'

export async function handleRequest(request: Request): Promise<Response> {
  const cookie = parse(request.headers.get('Cookie') || '')
  const c = cookie['c']
  console.log(c) // undefined
  console.log(c != null) // false
  
  return new Response('Hello World!', {
    status: 200,
  })
}
```

### DIFF
--- a/content/workers/examples/extract-cookie-value.md
+++ b/content/workers/examples/extract-cookie-value.md
@@ -19,7 +19,7 @@ const COOKIE_NAME = '__uid';
 function handleRequest(request) {
   const cookie = parse(request.headers.get('Cookie') || '');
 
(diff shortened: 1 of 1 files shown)


lukeed:
These are the same thing. The `!=` is a loose equality check, and for `null` and `undefined` [they match](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#loose_equality_using)

PR #4021, opened 2022-04-05
[Workers] route specificity for subdomains

### LINKED ISSUE #2787: A function at `/foo/[path].js` will match requests for `/foo/index.html`, but not `/foo/`.
I updated the title and description to give a clearer explanation of what I found confusing.

### Expected Behavior
I made an [example project](https://github.com/ryanjaeb/cloudflare-pages-routing) with some wildcard routes.  It's deployed to [cloudflare-pages-routing.pages.dev](https://cloudflare-pages-routing.pages.dev).

The [functions routing](https://github.com/cloudflare/cloudflare-docs/blob/production/products/pages/src/content/platform/functions.md#functions-routing) docs explain how to use `[[path]].js` and `[path].js` for wildcard matching.  When I read them, I think of the route matching in my sample project as:

1. `/foo/*`
2. `/bar/**`

There isn't an explanation for matching empty routes (ex: `/foo/` or `/bar/`) and I was expecting those to be the equivalent of the corresponding default resource (ex: `/foo/index.html`, `/bar/index.html`).  However, the behavior differs depending on whether or not you explicitly include `index.html` in a request.

I was expecting a request to `/foo/` or `/foo/index.html` to both match the route at `functions/[path].js`, but it only works for `/foo/index.html`.

The behavior for `functions/bar/[[path]].js` works like I was expecting.  A request to `/bar/` or `/bar/index.html` will match the `functions/bar/[[path]].js` route.

### Actual Behavior

* A request to `/foo/` does **not** match the `functions/foo/[path].js` route.
* A request to `/foo/index.html` **does** match the `functions/foo/[path].js` route.

Given the routes `functions/foo/[path].js` and `functions/bar/[[path]].js` and static HTML at `public/foo/index.html` and `public/bar/index.html`, you'll get the following results from requests:

| URL                                                       | Response           |
|-----------------------------------------------------------|--------------------|
| https://cloudflare-pages-routing.pages.dev/foo/           | `/foo/index.html`  |
| https://cloudflare-pages-routing.pages.dev/foo/index.html | `/foo/[pages].js`  |
| https://cloudflare-pages-routing.pages.dev/bar/           | `/bar/[[path]].js` |
| https://cloudflare-pages-routing.pages.dev/bar/index.html | `/bar/[[path]].js` |

The request to `/foo` seems like the odd one out.

The [functions routing](https://github.com/cloudflare/cloudflare-docs/blob/production/products/pages/src/content/platform/functions.md#functions-routing) docs.

### Additional information
The local behavior of `wrangler` matches the current pages.dev behavior.

### LINKED ISSUE #2879: Explain the most specific Workers route
### Expected Behavior
Users expect setting routes to be intuitive. [In the docs](https://developers.cloudflare.com/workers/platform/routes#matching-behavior), we say wildcards in the subdomain and in the subdirectory will match all requests `*.example.com/*,` but then we [write](https://developers.cloudflare.com/workers/platform/known-issues#route-specificity) that adding a wildcard in a route will match everything after the suffix `example.com*` and run the Worker - even if there's another route added with a trailing slash `example.com/*`. We also don't explain how routing executes if there's a wildcard on a subdomain.

### Actual Behavior
For adding a wildcard on a subdomain, a user asked which route `*.example.com/a` **or** `a.example.com/*` would match `a.example.com/a`?

From testing, `a.example.com/a` will match `a.example.com/*` instead of `*.example.com/a` (so this needs to be explained).

Explain how to write the most specific Workers route and add more examples.
https://developers.cloudflare.com/workers/platform/routes#matching-behavior

deadlypants1973:
closes #2879

### DIFF
--- a/content/workers/platform/routes.md
+++ b/content/workers/platform/routes.md
@@ -21,7 +21,7 @@ Cloudflare Site routes are comprised of:
 
 - Route URL (refer to [Matching Behavior](#matching-behavior))
 - Worker script to execute on matching requests
-- Failure mode for rate-limited accounts on the Free plan (see [daily request limits](/workers/platform/limits/#request-limits))
+- Failure mode for rate-limited accounts on the Free plan (refer to [daily request limits](/workers/platform/limits/#request-limits))
 
 The Routes REST API documentation can be found [in the Workers API docs](https://api.cloudflare.com/#worker-routes-properties).
 
@@ -29,9 +29,9 @@ If your route is configured to a hostname, you will need to add a DNS record to 
 
 You may enter a placeholder AAAA record pointing to [100::](https://datatracker.ietf.org/doc/html/rfc6666), which must be proxied through Cloudflare (orange-cloud in the DNS settings). This value specifically is the reserved IPv6 discard prefix but is not the only value allowed. For example, you may also use an A record pointed to 192.0.2.1 or a CNAME pointed to any resolvable target.
 
-\* _A zone that you have registered with some registrar (not workers.dev) and setup Cloudflare to serve as [a reverse proxy](https://www.cloudflare.com/learning/cdn/glossary/reverse-proxy/)._
+* _A zone that you have registered with some registrar (not workers.dev) and setup Cloudflare to serve as [a reverse proxy](https://www.cloudflare.com/learning/cdn/glossary/reverse-proxy/)._
 
-## Routes with \*.workers.dev
+## Routes with *.workers.dev
 
 Cloudflare Workers accounts come with a `*.workers.dev` subdomain that is configurable in the Cloudflare dashboard. Your `*.workers.dev` subdomain allows you to deploy Workers scripts [without attaching a custom domain as a Cloudflare zone](https://blog.cloudflare.com/announcing-workers-dev/).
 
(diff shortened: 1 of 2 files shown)


content/workers/platform/routes.md
```diff
@@ -21,17 +21,17 @@ Cloudflare Site routes are comprised of:
...
 
 The Routes REST API documentation can be found [in the Workers API docs](https://api.cloudflare.com/#worker-routes-properties).
 
 If your route is configured to a hostname, you will need to add a DNS record to Cloudflare to ensure that the hostname can be resolved externally. If your Worker acts as your origin (that is, the request terminates in a Worker), you must add a DNS record.
 
 You may enter a placeholder AAAA record pointing to [100::](https://datatracker.ietf.org/doc/html/rfc6666), which must be proxied through Cloudflare (orange-cloud in the DNS settings). This value specifically is the reserved IPv6 discard prefix but is not the only value allowed. For example, you may also use an A record pointed to 192.0.2.1 or a CNAME pointed to any resolvable target.
 
-\* _A zone that you have registered with some registrar (not workers.dev) and setup Cloudflare to serve as [a reverse proxy](https://www.cloudflare.com/learning/cdn/glossary/reverse-proxy/)._
+* _A zone that you have registered with some registrar (not workers.dev) and setup Cloudflare to serve as [a reverse proxy](https://www.cloudflare.com/learning/cdn/glossary/reverse-proxy/)._
```
ranbel, line 32: This solitary bullet point seems out of place. Is it meant to be a footnote? Maybe incorporate it into the main text or make it an "Aside" component.

ranbel (approved):

PR #4047, opened 2022-04-07
Ddos/release 2022 04 07

jdesgats:
Note: the two `scheduled.json` files will just contain scheduled changes in a JSON format to help the changelog generator from one release to the next (in order to filter previously scheduled changes correctly when re-generating the document).


content/ddos-protection/change-log/http/2022-04-07.md
```diff
@@ -0,0 +1,34 @@
...
+      <th>Rule ID</th>
+      <th>Description</th>
+      <th>Previous Action</th>
+      <th>New Action</th>
+      <th>Notes</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>c9f18c647ae745c6b81b459d8ed59b32</td>
```
pedrosousa: Rule IDs are usually trimmed in the WAF change log, keeping the last 8 characters of the ID (that's what users see in the dashboard rules list). Would it make sense to do the same here...?

content/ddos-protection/change-log/http/2022-04-07.md
```diff
@@ -0,0 +1,34 @@
+---
+title: 2022-04-07
+type: table
```
pedrosousa, line 3: We don't really need this line with the new docs tooling (we will eventually remove these from the WAF change log).

pedrosousa (approved):
Added a few comments that can be addressed later. Approving because the new change log page mentions today's date.

content/ddos-protection/change-log/http/2022-04-07.md
jdesgats: Done, I'll update the generator to do this for the next release.

content/ddos-protection/change-log/http/2022-04-07.md
jdesgats: Okay, I just copy/pasted the WAF skeleton, I'll remove it from my template for next releases.

PR #4060, opened 2022-04-08
KV list can return an empty array even when list is not complete

koeninger:
If this seems like too much info, happy to trim it down

### DIFF
--- a/content/workers/runtime-apis/kv.md
+++ b/content/workers/runtime-apis/kv.md
@@ -271,7 +271,7 @@ const cursor = value.cursor
 const next_value = await NAMESPACE.list({"cursor": cursor})
 ```
 
-Note that checking for an empty array in `keys` is not sufficient to determine whether there are more keys to fetch, you should check `list_complete`. The reason it's possible to have an empty array in `keys` but still have more keys to fetch is because [recently expired or deleted keys](https://en.wikipedia.org/wiki/Tombstone_%28data_store%29) must be iterated through but will not be included in the returned `keys`.
+Note that checking for an empty array in `keys` is not sufficient to determine whether there are more keys to fetch, you should check `list_complete`. The reason it is possible to have an empty array in `keys` but still have more keys to fetch is because [recently expired or deleted keys](https://en.wikipedia.org/wiki/Tombstone_%28data_store%29) must be iterated through but will not be included in the returned `keys`.
 
 ## KV bindings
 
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/kv.md
```diff
@@ -271,6 +271,8 @@ const cursor = value.cursor
 const next_value = await NAMESPACE.list({"cursor": cursor})
 ```
 
+Note that checking for an empty array in `keys` is not sufficient to determine whether there are more keys to fetch, you should check `list_complete`. The reason it's possible to have an empty array in `keys` but still have more keys to fetch is because [recently deleted keys](https://en.wikipedia.org/wiki/Tombstone_%28data_store%29) must be iterated through but will not be included in the returned `keys`.
```
a-robinson: It might be worth noting that the main reason this happens is due to expiration, but given that it isn't the only reason, I'm fine leaving it as is too.

a-robinson (approved):

a-robinson:
Thanks!

content/workers/runtime-apis/kv.md
koeninger:
```suggestion
Note that checking for an empty array in `keys` is not sufficient to determine whether there are more keys to fetch, you should check `list_complete`. The reason it's possible to have an empty array in `keys` but still have more keys to fetch is because [recently expired or deleted keys](https://en.wikipedia.org/wiki/Tombstone_%28data_store%29) must be iterated through but will not be included in the returned `keys`.
```

content/workers/runtime-apis/kv.md
```diff
@@ -271,6 +271,8 @@ const cursor = value.cursor
 const next_value = await NAMESPACE.list({"cursor": cursor})
 ```
 
+Note that checking for an empty array in `keys` is not sufficient to determine whether there are more keys to fetch, you should check `list_complete`. The reason it is possible to have an empty array in `keys` but still have more keys to fetch is because [recently expired or deleted keys](https://en.wikipedia.org/wiki/Tombstone_%28data_store%29) must be iterated through but will not be included in the returned `keys`.
```
deadlypants1973:
```suggestion
Note that checking for an empty array in `keys` is not sufficient to determine whether there are more keys to fetch; check `list_complete` instead. The reason it is possible to have an empty array in `keys`, but still have more keys to fetch, is because [recently expired or deleted keys](https://en.wikipedia.org/wiki/Tombstone_%28data_store%29) must be iterated through but will not be included in the returned `keys`.
```

deadlypants1973 (approved):
Approved with comment

PR #4065, opened 2022-04-09
Update windows.md

anilabhadatta:
Fixed some errors in commands and config.
Tested running config.yml from default profile dir. {Better to run from default profile dir as user can change the hostname and service and there is no need to copy the config to systemprofile dir.


ranbel (approved):
Thanks for your contribution!

PR #4066, opened 2022-04-09
Update direct-creator-uploads.md

Dectom:
It's required for anything over 200mb it isn't an optional thing so should state that it's required not recommended!!!


dcpena:
Hi @Dectom, thanks for your contribution! You can actually use copy-from-link as another upload option for large videos, so the update would technically be inaccurate. Will close this PR as a result.

PR #4098, opened 2022-04-13
replacing -vnet with --vnet

JubbaSmail:
the documentation is showing the wrong option "-vnet", it should be "--vnet" instead
to confirm you can run the command:
`cloudflared tunnel route ip add --help`


ranbel (approved):
Thanks for contributing!

PR #4104, opened 2022-04-13
advise avoiding global variables in durable objects

koeninger:


content/workers/learning/using-durable-objects.md
```diff
@@ -120,11 +120,13 @@ export class Counter {
 
     // Handle HTTP requests from clients.
     async fetch(request) {
-        ...
+        // use this.value rather than storage
     }
 }
 ```
 
+A given instance of a Durable Object may share global memory with other instances of the same class. In the example above, using a global variable `value` instead of the instance variable `this.value` would be incorrect.  Two different instances of `Counter` will each have their own separate memory for `this.value`, but might share memory for `value`, leading to unexpected results. Because of this, it is best to avoid global variables.
```
xortive:
```suggestion
A given instance of a Durable Object may share global memory with other instances of the same class. In the example above, using a global variable `value` instead of the instance variable `this.value` would be incorrect.  Two different instances of `Counter` will each have their own separate memory for `this.value`, but might share memory for the global variable `value`, leading to unexpected results. Because of this, it is best to avoid global variables.
```

xortive (approved):

content/workers/learning/using-durable-objects.md
```diff
@@ -120,11 +120,13 @@ export class Counter {
 
     // Handle HTTP requests from clients.
     async fetch(request) {
-        ...
+        // use this.value rather than storage
     }
 }
 ```
 
+A given instance of a Durable Object may share global memory with other instances of the same class. In the example above, using a global variable `value` instead of the instance variable `this.value` would be incorrect.  Two different instances of `Counter` will each have their own separate memory for `this.value`, but might share memory for the global variable `value`, leading to unexpected results. Because of this, it is best to avoid global variables.
```
deadlypants1973:
```suggestion
A given instance of a Durable Object may share global memory with other instances of the same class. In the example above, using a global variable `value` instead of the instance variable `this.value` would be incorrect. Two different instances of `Counter` will each have their own separate memory for `this.value`, but might share memory for the global variable `value`, leading to unexpected results. Because of this, it is best to avoid global variables.
```

deadlypants1973 (approved):
Approved with suggestion about spacing

PR #4115, opened 2022-04-15
Buggy code sample: Worker fetching itself in a loop

kael:
This code sample looks odd.

The worker fetches itself, it's a loop.

### DIFF
--- a/content/workers/examples/cache-api.md
+++ b/content/workers/examples/cache-api.md
@@ -30,6 +30,7 @@ async function handleRequest(event) {
       `Response for request url: ${request.url} not present in cache. Fetching and caching request.`
     );
     // If not in cache, get it from origin
+    // Here the worker fetches its URL, it's a loop.
     response = await fetch(request);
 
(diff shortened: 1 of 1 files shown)


KianNH:
That example is made to be added on a route where an origin already exists - it's a cache middleware to put in-front of your website to leverage the Cache API.

If a Worker fetches itself (or another Worker in the same zone), the request does not go to the Worker, it goes to the origin behind that route.

kristianfreeman:
Thanks for the submission - @KianNH is correct, you can use the `event.request` variable in a Workers script to make requests to your origin, so this code is correct.

kael:
Alright.

Just so you know, I'm new to Cloudflare Cache with Workers. I tried with an URL but the cache didn't seem to work (apparently due to some specific auth headers according to some quick searches on SO).

So I came across that example and tried to use it as starter.

But it's confusing though.

Please, add some contextual information about that specific use case, and perhaps a mermaid diagram describing the flow, because the cache middleware capability is not obvious for starters.

Cheers. :+1:

PR #4127, opened 2022-04-17
[Pages] Enable dropdown on sidebar for Migrations

KianNH:
Currently, the `Migration guides` in the sidebar doesn't have dropdown functionality - whereas Frameworks, which is formatted very similarly, does.

image

image

If this is intentional, feel free to close - I just think it looks a bit out of place.

### DIFF
--- a/content/pages/migrations/_index.md
+++ b/content/pages/migrations/_index.md
@@ -1,6 +1,5 @@
 ---
 type: overview
-hideChildren: true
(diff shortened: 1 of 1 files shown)


KianNH:
I think this one is pretty much just personal preference - i.e it isn't enabled on https://developers.cloudflare.com/workers/tutorials/ because otherwise the entire sidebar would be 18+ tutorials.

PR #4171, opened 2022-04-20
Update fetching-bulk-analytics.md

tobbbe:
Since you have to use API Keys, not API Token, you have to use the correct headers: X-Auth-Email and X-Auth-Key instead of Authorization.


victor-perov:
I think there is a misunderstanding. GraphQL API supports both ways to authenticate the request:

- providing a single header with API Token
- providing a pair of email / key headers

API Tokens allow you to configure access as needed: limit it to a specific set of resources (like analytics for a given zone, but not for the entire account), or set a particular range of IP addresses it can be used from. In contrast, a pair of email/key headers provide a global level of access to the account resources (zone/account analytics, etc).

In my mind, we have this PR because you might have been using the wrong API Token or one that wasn't configured to allow you access to the account-level analytics.

PR #4180, opened 2022-04-21
Update ssl-modes.md to add port notice

tomkln:
Ports other than 443 do not support Flexible SSL mode as documented CF-internally.

### DIFF
--- a/content/ssl/origin-configuration/ssl-modes.md
+++ b/content/ssl/origin-configuration/ssl-modes.md
@@ -74,7 +74,7 @@ Choose this option when you cannot set up an SSL certificate on your origin or y
 
 ### Limitations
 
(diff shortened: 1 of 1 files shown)


content/ssl/origin-configuration/ssl-modes.md
```diff
@@ -74,6 +74,8 @@ Choose this option when you cannot set up an SSL certificate on your origin or y
 
 ### Limitations
 
+Flexible mode is only supported for HTTPS connections on port 443 (default port). Other ports using HTTPS will fall back to Full mode.
```
kodster28:
```suggestion
Flexible mode is only supported for HTTPS connections on port 443 (default port). Other ports using HTTPS will fall back to [**Full** mode](#full).
```

kodster28 (approved):

PR #4186, opened 2022-04-21
update known issues

Ekwuno:
Update known issues for Pages


WalshyDev:
Do we want to add this to known issues? This is a roadmap item rather than an issue itself

Ekwuno:
> Do we want to add this to known issues? This is a roadmap item rather than an issue itself

@nevikashah asked that it be added. Just so people know we currently don’t support it yet.

deadlypants1973 (approved):

PR #4188, opened 2022-04-21
[Content maintenance] Change all redirects to slashed

kodster28:
Once this is live, we'll turn on the transform rule in the background


ranbel (approved):

ranbel (approved):

kodster28:
There's a small chance this may break something... but it's also kinda not helpful to have someone review fully for accuracy.

I'll monitor the backend and see if there's any issues, but pushing this live.

PR #4224, opened 2022-04-26
gitlab.md use ss instead of netstat

minecraftchest1:
'netstat' is depricated with the rest of 'net-tools'. Use 'ss' instead. Also 'ss' has a builtin option to show only listening '-l'


ranbel:
Hi, the tutorial screenshots would need to be updated in order to make this change. Do you have an example of the `ss -lntup` outputs before and after `gitlab-ctl configure`?

minecraftchest1:
No I do not. I should be able to get one if a few days thought. I can also see about mocking up some output as well.

minecraftchest1:
Fixed in 2bc4430f25d085e89c416a39d19462ca5efd1b8f/

ranbel (approved):

PR #4239, opened 2022-04-27
[Stream] Added content for downloading stream live videos

dcpena:
Added content for downloading stream live videos to address [PCX-3799](https://jira.cfops.it/browse/PCX-3799).

### DIFF
--- a/content/stream/stream-live/download-stream-live-videos.md
+++ b/content/stream/stream-live/download-stream-live-videos.md
@@ -1,0 +1,16 @@
+---
+pcx-content-type: how-to
+title: Download live stream videos
+---
+
+# Download live stream videos
+
+You can enable downloads for live stream videos from the Cloudflare dashboard.
+
+1. Log in to your [Cloudflare dashboard](https://dash.cloudflare.com) and select your account.
+2. Click **Stream** > **Live Inputs**.
+3. Click a live input from the list to select it.
+4. Under **Videos created by live input**, locate your video and click to select it.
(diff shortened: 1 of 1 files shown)


bkrebsbach (approved):

kodster28 (approved):

Everlag:
Inherently, a Stream Live video that's finished is a regular on-demand video, we have some preexisting docs on downloading that here:
https://developers.cloudflare.com/stream/viewing-videos/download-videos/

Since this is more dashboard oriented, maybe it makes sense to move the current contents of this PR over to the download-videos page, and have the 'Download Stream Live Video' page be a link to the regular vod page?

content/stream/stream-live/download-stream-live-videos.md
```diff
@@ -0,0 +1,16 @@
+---
+pcx-content-type: how-to
+title: Download live stream videos
+---
+
+# Download live stream videos
```
Everlag, line 6: It may be worth noting that, as far as I know, we only allow downloading a Stream Live video that has moved into a 'ready' state, which means the broadcast has ended.

dcpena: Ah, that is a great callout! I will make sure to add that information.

dcpena:
Hey @Everlag! Checked in with Zaid about placement of the download section, and we'll keep it where it is so users can easily see Stream Live supports downloads. Thanks for the feedback :)

PR #4265, opened 2022-05-01
Update CloudflareWARP.mobileconfig

theoriginalgri:
The example `CloudflareWARP.mobileconfig` was not working at all for me when installing it locally.

After some debugging, I found it was missing the `PayloadScope=System`. Without setting this, it was installed into `/Library/Managed Preferences/$USER` and Warp was unable to discover the file. Probably this is set correctly when installed via MDM but I wanted to try it out locally before doing so.

Also I simplified the configuration by removing `com.apple.ManagedClient.preferences` and `mcx_preference_settings`. This makes the configuration human readable in the Profiles overview of macOS.

The documentation should also state it's a good idea to embed the root ca which I did in my local testing but did not add this here.

Anyway, thanks for Cloudflare tunnels! 👍


ranbel (approved):
thanks for your contribution!

PR #4269, opened 2022-05-02
Fix incorrect public beta bindings example

vlovich:


content/r2/get-started.md
```diff
@@ -152,7 +152,7 @@ async function handleRequest(request) {
       await MY_BUCKET.put(key, request.body);
       return new Response(`Put ${key} successfully!`);
    case 'GET':
-     const { value } = await MY_BUCKET.get(key);
+     const { body } = await MY_BUCKET.get(key);
```
WalshyDev: the object is nullable no? So this will error if the key doesn't exist

content/r2/get-started.md
```diff
@@ -152,7 +152,7 @@ async function handleRequest(request) {
...
 
      if (value === null) {
```
WalshyDev: needs to be updated along with line 161

WalshyDev (commented):
@Cherry did a PR here: https://github.com/cloudflare/cloudflare-docs/pull/4262

Just needs merging. For this PR other lines need updating:

content/r2/get-started.md
vlovich: If the key doesn't exist then `.get` throws an exception

vlovich:
Superseded by https://github.com/cloudflare/cloudflare-docs/pull/4262

PR #4273, opened 2022-05-02
Add missing fix content type

vlovich:
Content-type is a supported header for write operations.


deadlypants1973:
> Content-type is a supported header for write operations.

Was this included in the internal R2 S3 compat guide? We should aim to have these 2 documents as similar as possible to avoid confusion.

deadlypants1973 (approved):

PR #4274, opened 2022-05-03
[Cache] Purging cache by tag might need to specify Tag, not URL

d-lord:
Hi, not sure if the steps documented here are correct, so I've proposed a change.

If they are correct, then this behaviour is confusing and the docs could potentially use a line clarifying why.

In the dashboard, following the docs and clicking the URL button, looks like this:

Screen Shot 2022-05-03 at 10 26 21 am

But I think it should be Tag instead, which produces this form:

Screen Shot 2022-05-03 at 10 27 44 am

Tag seems like the intended behaviour in this section.


angelampcosta (approved):

angelampcosta:
👋 @d-lord Thank you for bringing this to our attention!

PR #4283, opened 2022-05-04
add install notes for PHP composer

jacobbednarz:
Stumbled across this with PHP composer today and thought we should add it to save the next person trying to find it.


kkrum (approved):

kkrum:
Thanks for the submission!

ranbel (approved):

PR #4290, opened 2022-05-04
[Rules] Add Managed Transforms

pedrosousa:
Addresses PCX-1309.


content/rules/transform/_index.md
```diff
@@ -10,23 +10,28 @@ Transform Rules allow you to adjust the URI path, query string, and HTTP headers
```
kodster28, line 11: Might want to shift this to a `layout: single` so you get the headers on the righthand side.

content/rules/transform/managed-transforms/_index.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Managed Transforms
```
kodster28, line 2: Not sure if you can totally get around this, but it's sorta confusing seeing `Managed transforms` as a folder name and then a separate how-to page for `Manage Transform Rules`. Maybe switch the latter to be `Work with...` just so there's more scannability.

kodster28 (approved):

content/rules/transform/managed-transforms/_index.md
pedrosousa: Yes, makes sense. Thanks!

PR #4296, opened 2022-05-05
Fix path regex example

petr-tichy:
The `path` regex example was misleading. It is indeed a Regex match, so to match files ending in `.jpg` etc., the regex should be `\.(jpg|png|css|js)$`

### DIFF
--- a/content/cloudflare-one/connections/connect-apps/configuration/local-management/ingress.md
+++ b/content/cloudflare-one/connections/connect-apps/configuration/local-management/ingress.md
@@ -54,7 +54,7 @@ ingress:
     service: https://localhost:8000
   # Rules can match the request's path to a regular expression:
   - hostname: static.example.com
-    path: /*.(jpg|png|css|js)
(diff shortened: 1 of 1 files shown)


CryptoGodfatherVA38 (approved):

ranbel:
Hello, thanks for your comment. Could you elaborate on why the provided syntax did not work for you?

petr-tichy:
The provided example regex `/*.(jpg|png|css|js)` matches (among others) the following paths:
`////xjpg`
`xjpg`
`foojpgbar`
`-json`
and so on.

IMHO the intention was to provide an example of using a Golang regex syntax in the `path` parameter to match static files ending with the `jpg|png|css|js` extensions. One way is to write the regex as in my PR:
`path: \.(jpg|png|css|js)$`
this matches
`some/path/file.jpg`
but not
`xjpg`
`foojpgbar`
`-json`

joliveirinha (approved):

ranbel (approved):

PR #4306, opened 2022-05-05
[R2] get started script

adaptive:
The get started example didn't work for me. I also increased the GET method to show `contentType` and `eTag`.
_PS: This is a working script at the time of this PR, as R2 is beta and runtime is always changing._

cc: @vlovich


WalshyDev:
These are the old bindings/API, make sure your compatibility date is set to today or a recent time

adaptive:
@WalshyDev the code was pushed on the Dashboard so there is no compatibility date, which date is Dashboard quick edit setting? Not today?

WalshyDev:
The quick edit doesn't use a date so Sept 2021 is no date.

PR #4317, opened 2022-05-06
[Search updates] Added product facet filter + card info

kodster28:
Added Product facet filter to lefthand filters and product name to the individual card.

This data is pulled from the [rendered product title](https://github.com/cloudflare/cloudflare-docs/blob/production/data/api-shield.yml#L5) during the page scraping process.

Screen Shot 2022-05-06 at 2 49 55 PM


kodster28:
Open question, is this confusing that some entries will not have a Product facet on them? 

I think there's a touch of confusion that it could introduce.... but that seems to be far outweighed by the benefit (ability to filter based on most of a specific product's info) + the sheer # of requests I've gotten for this functionality (like ~10 internally).

kodster28:
> Open question, is this confusing that some entries will not have a Product facet on them?
> 
> I think there's a touch of confusion that it could introduce.... but that seems to be far outweighed by the benefit (ability to filter based on most of a specific product's info) + the sheer # of requests I've gotten for this functionality (like ~10 internally).

From informal internal feedback, seems like this is not a valid issue.

lukeed:
I think this is great. My only nit is that the result cards should have a colon (`:`) between the label & the value:

Screen Shot 2022-05-17 at 9 32 50 AM


> Added colons, `font-style: italic; font-size: 85%`, and reduced to `margin-right: 5px` on `.coveo-field-caption` (labels)

```css
/* added */
.coveo-result-cell {
  font-style: italic;
  font-size: 85%;
}

/* modified */
.CoveoFieldValue .coveo-field-caption {
  margin-right: 5px;
}
```

lukeed:
Screen Shot 2022-05-17 at 1 40 41 PM

It's the wrong margin-right css change. The one i modified replaced a `10px` value

kodster28:
Couldn't exactly find the property you were talking about, but it _seems_ to look a bit better with the latest update.

Screen Shot 2022-05-18 at 4 27 38 PM

kristianfreeman (approved):

PR #4319, opened 2022-05-06
State default KV cacheTtl earlier

koeninger:


a-robinson (approved):

content/workers/runtime-apis/kv.md
```diff
@@ -153,7 +153,7 @@ The `get` options object also accepts a `cacheTtl` parameter:
...
 ```
 
-The `cacheTtl` parameter must be an integer that is greater than or equal to 60. It defines the length of time in seconds that a KV result is cached in the edge location that it is accessed from. This can be useful for reducing cold read latency on keys that are read relatively infrequently. It is especially useful if your data is write-once or write-rarely. It is not recommended if your data is updated often and you need to see updates shortly after they are written, because writes that happen from other edge locations will not be visible until the cached value expires.
+The `cacheTtl` parameter must be an integer that is greater than or equal to 60, which is the default. It defines the length of time in seconds that a KV result is cached in the edge location that it is accessed from. This can be useful for reducing cold read latency on keys that are read relatively infrequently. It is especially useful if your data is write-once or write-rarely. It is not recommended if your data is updated often and you need to see updates shortly after they are written, because writes that happen from other edge locations will not be visible until the cached value expires.
```
deadlypants1973:
```suggestion
The `cacheTtl` parameter must be an integer that is greater than or equal to `60`, which is the default. It defines the length of time in seconds that a KV result is cached in the edge location that it is accessed from. This can be useful for reducing cold read latency on keys that are read relatively infrequently. It is especially useful if your data is write-once or write-rarely. It is not recommended if your data is updated often and you need to see updates shortly after they are written, because writes that happen from other edge locations will not be visible until the cached value expires.
```
If 60 is an input value for `cacheTtl`, then needs backticks.

deadlypants1973 (approved):

PR #4328, opened 2022-05-09
[Rules] Update Managed Transforms

pedrosousa:
Addresses PCX-3913.

Add "Add True-Client-IP" Managed Transform.
Update fields in API response.
Unavailable transforms will not be displayed/listed.
Conflicting options will be disabled in the UI, and will have `"has_conflict": true` in the API.

### DIFF
--- a/content/rules/transform/managed-transforms/_index.md
+++ b/content/rules/transform/managed-transforms/_index.md
@@ -9,10 +9,10 @@ layout: single
 
 Managed Transforms allow you to perform common adjustments to HTTP request and response headers with the click of a button. The available adjustments include:
 
-* Add bot protection request headers
-* Remove or add headers related to the visitor's IP address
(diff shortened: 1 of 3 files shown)


content/rules/transform/managed-transforms/_index.md
```diff
@@ -10,9 +10,8 @@ layout: single
 Managed Transforms allow you to perform common adjustments to HTTP request and response headers with the click of a button. The available adjustments include:
 
 * Add bot protection request headers
```
angelampcosta: We could add punctuation to the sentences in this list.

angelampcosta (approved):

content/rules/transform/managed-transforms/_index.md
pedrosousa: Yes, we can :)

PR #4334, opened 2022-05-09
[Workers] Add description for the new parameter startAfter for Durable Object list method content/workers/runtime-apis/durable-objects.md

jqmmes:
- `startAfter` is new parameter added to the list method of durable objects that allows listing to start after a key instead of on a key.
- This PR adds description for this new parameter in content/workers/runtime-apis/durable-objects.md


a-robinson (approved):
LGTM, thanks @jqmmes !

PR #4351, opened 2022-05-10
[WAF] Add WAF ML

pedrosousa:
Addresses PCX-3022.


content/waf/about/waf-ml.md
```diff
@@ -0,0 +1,87 @@
...
+      <th style="width: 33%">Field</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+  </tbody>
+  <tr>
+    <td><strong>WAF Attack Score</strong></td>
+    <td><code>cf.waf.ml.score</code></td>
+    <td>A total score from 1 to 99 that combines the score of each attack vector into a single score.<br/>
```
helloandre: We should either add a note that while it is *a* total score, it is not *the* total score, specifically that there is no relationship between summing (or diffing) the `.sql` and `.xss` to get the `.score`, all three are independent of each other.

pedrosousa: Updated to "global score" instead of "total score", and added a note clarifying that it's not a sum of other available scores.

helloandre (approved):

ranbel (approved):
lgtm!

PR #4364, opened 2022-05-10
Recommend users install wrangler locally, not globally

jkup:
We are in the process of making the CLI a better experience when installed globally. For now, the official recommendation is to always install it locally to your project:

https://github.com/cloudflare/wrangler2/issues/955


threepointone (approved):

deadlypants1973 (approved):

rauno56:
The run commands should also be changed then. If the package is installed locally simply running `wrangler init` doesn't work anymore. `npx wrangler`, however does. Init command actually states it correctly, but the documentation is now inconsistent.

cc: @threepointone , @deadlypants1973

jkup:
@Rauno56 Thanks for this! Working on it now.

PR #4426, opened 2022-05-13
Remove contradictory forward slash

isaac-mcfadyen:
- In the "best practices" section under "Topics and subscriptions" the 4th bullet point says that the forward slash on topics should be omitted and isn't necessary.
- However, on the first 3 points the forward slash is included in the examples. This contradicts the 4th bullet point and should be removed.


elithrar:
LGTM

elithrar (approved):

elithrar:
cc/ @dcpena

dcpena (approved):

PR #4437, opened 2022-05-14
Replace Relative Image Links with Absolute Links

helloimalastair:
Script-based replace, requested by @WalshyDev.
Script source available [here](https://gist.github.com/helloimalastair/d93b243f910a6b2c5e76fcb47c69dce3)


lukeed (changes requested):
Thank you but this is not preferred. Everything should be in absolute paths. This is so that the link checker in CI works & preserves the authenticity of deploy previews – if content does not link to itself (or links to its previous self) you can easily end up w/ a false sense of security since you're not actually referencing/relying on current content.

The internal documentation styleguide prevents these `https://developers.cloudflare.com/...` (and `../../...`) links.

PR #4438, opened 2022-05-14
fix: ensure scroll-to-section works on page load

### LINKED ISSUE #4382: Page loads don't auto-scroll to anchor tag
### Example URL(s)

https://developers.cloudflare.com/ssl/ssl-tls/certificate-statuses#monitor-certificate-statuses

### Actual Behavior

When I load a page with an anchor tag specified, the page does not auto-scroll to that anchor.

Tested this in staging and it _does_ work if you use the unslashed page request (developers.cloudflare.com/product/page#header) but doesn't work with the slashed version of the request (developers.cloudflare.com/product/page/#header).

Tried turning off my Transform rule -- which might be contributing to the behavior -- and that didn't have an effect. So think it might be a combination of something else + the transform rule.

lukeed:
Don't just rely on browser native behavior.
- Closes #4382

### DIFF
--- a/assets/events.ts
+++ b/assets/events.ts
@@ -29,6 +29,20 @@ export function $tabbable(links: NodeListOf<Element>, bool: boolean) {
   }
 }
 
+// scroll to section header
+// but only on load if `#hash` in URL
+export function load() {
+  let hash = location.hash.substring(1);
+  let item = hash && document.getElementById(hash);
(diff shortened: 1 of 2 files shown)


kodster28 (approved):
Did testing in the following browsers and confirmed that it worked for various links:
- Safari (Mac)
- Chrome (Android, Mac)

PR #4441, opened 2022-05-14
Update deploy-a-hugo-site.md

sifigi4335:
Provide a 'jump' link to the deploy section for those with a working Hugo site.


sifigi4335 (commented):
Move the "before you start" section higher up since it makes sense to be at the top. Also include a link to jump straight to the deploy section for those with working Hugo sites hosted with the supported Git providers

PR #4443, opened 2022-05-15
[ZT] Added SNI info to docs

isaac-mcfadyen:
- Currently, Cloudflare Tunnel doesn't work with [SNI](https://www.cloudflare.com/en-ca/learning/ssl/what-is-sni/) without the `originServerName` flag (locally) or the Origin Server Name option (on the Zero Trust dash) specified.
- From what I can tell, this note isn't present anywhere in the docs.
- This clarifies the docs to say that the flag is required when using servers that require SNI (such as Caddy).

### DIFF
--- a/content/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide.md
+++ b/content/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide.md
@@ -54,6 +54,10 @@ Follow these steps to connect an application through your tunnel. If you are loo
 
 3. Under **Additional application settings**, specify any parameters you would like to add to your tunnel configuration.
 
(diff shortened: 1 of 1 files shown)


ranbel:
Hi, SNI info would be a great addition to the docs. However, because this note refers to a specific use case, I recommend that we instead add it to: https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/local-management/ingress/#originservername

Feel free to open another PR with this change.

isaac-mcfadyen:
@ranbel My only concern there would be that people using Tunnel for the first time would miss it, and it impacts a significant number of reverse-proxies that expect SNI to be used to indicate the domain name (as it would be normally when hitting a webserver).

Issue #4436, opened 2022-05-14
R2 Ranged reads incorrect parameter


Anjasomc:
### Existing documentation URL(s)

- https://developers.cloudflare.com/r2/runtime-apis/#ranged-reads

### Section that requires update

"Ranged reads" under R2GetOptions.

### What needs to change?

> Ranged reads
>
> R2GetOptions accepts a range parameter, which restricts data returned in body to be range bytes, starting from offset, inclusive.
>
>    offset
>
>    The byte to begin returning data from, inclusive.
>
>range
>
>    The number of bytes to return. If more bytes are requested than exist in the object, fewer bytes than this number may be returned.

`R2GetOptions.range` requires a `length` parameter:

```js
await bucket.get(key, {
    range: {
        offset: 123,
        range: 100
    }
};
// RangeError: Invalid range. Length (123) must be an integer, not floating point

await bucket.get(key, {
    range: {
        offset: 123,
        length: 100
    }
};
// Succeeds
```

As an aside the error message is a bit confusing - showing the value for `offset` rather than `length`, and saying that's a floating point.

### How should it change?

The `R2GetOptions.range.range` parameter in the docs should be renamed `R2GetOptions.range.length`.

### Additional information

This is a seperate issue but it may come up in your testing.
Using `offset` with a value of 0 will throw an error:

```js
 await bucket.get(key, {
    range: {
        offset: 0,
        length: 100
    }
};
// Error: We encountered an internal error. Please try again.
```

mlsub:
I encountered this issue in my testing and have you found a solution to resolve the error that occurs when offset = 0?

Anjasomc:
Currently I'm just responding with a redirect to one of our non-R2 endpoints whenever that occurs. You could pad all your objects with an extra byte at the start and then always use an offset of 1, but I think it's probably just best to wait until this is fixed.

KianNH:
The `offset: 0` bug is a known one - made a PR to amend the docs regarding `range` => `length`

haleycode:
@KianNH Thanks! Please tag me to close this issue once the PR is merged (or you can close yourself if you have the right permissions).

KianNH:
Will be closed by https://github.com/cloudflare/cloudflare-docs/pull/4670

PR #4450, opened 2022-05-16
[R2] Fix incorrect parameter name in R2Range

KianNH:
As per #4436, the parameter is `length` and not `range`

This can be seen in the `workers-types` definition for `R2Range`:

https://github.com/cloudflare/workers-types/blob/4d2664ba55a5c4cc6eaa3d4a144de3156ba6ebda/index.d.ts#L1129


KianNH:
Superseded by https://github.com/cloudflare/cloudflare-docs/pull/4670

PR #4467, opened 2022-05-16
Fix link to Durable Objects Alarms blog post

irvinebroque:
"objects" instead of "object"


koeninger (approved):

koeninger:
Thanks for catching that

PR #4469, opened 2022-05-16
Move Wrangler instructions back to global

jkup:
After talking with @threepointone and @deadlypants1973 we decided that telling people to install Wrangler globally is the best path forward.

This updates the respective docs as well as mentioning you should uninstall @cloudflare/wrangler globally before installing Wrangler2.


threepointone (approved):
Approving, but I don't see where it recommend uninstalling `@cloudflare/wrangler`?

jkup:
@threepointone I had added it to https://github.com/cloudflare/cloudflare-docs/blob/production/content/workers/wrangler/migration/migrating-from-wrangler-1.md in a different commit! But I just added it to Get Started in a note as well. I was considering adding it to the general quickstart guide but I thought maybe it's best to not mention it for people who are just now getting started? What do you think cc @deadlypants1973

deadlypants1973 (approved):
@jkup approved, but do make sure that all the `npx .....` commands have also been adjusted accordingly too

PR #4474, opened 2022-05-17
r2: clarify rclone min version

elithrar:
Most package managers do not install v1.59 yet, and so many users will end up seeing errors when using their existing rclone installations.

> The S3 specification requires an `X-Amz-Content-Sha256` header on all requests. R2, sticking closely to the spec, requires this header but most other S3-compatible providers don't. Rclone version < 1.59 only sends this header when the body is empty (such as when getting an object or creating a bucket) - not when the body has content. So that's why uploads fail in older rclone versions.

### DIFF
--- a/content/r2/examples/rclone.md
+++ b/content/r2/examples/rclone.md
@@ -16,6 +16,10 @@ It is recommended that you choose a unique provider name and then rely on all de
 This will create a `rclone` configuration file, which you can then modify with the preset configuration given below.
 {{</Aside>}}
 
+{{<Aside type="note">}}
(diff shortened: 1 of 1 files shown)


dcpena (approved):

content/r2/examples/rclone.md
```diff
@@ -16,6 +16,10 @@ It is recommended that you choose a unique provider name and then rely on all de
 This will create a `rclone` configuration file, which you can then modify with the preset configuration given below.
 {{</Aside>}}
 
+{{<Aside type="note">}}
+Ensure you are running `rclone` v1.59 or greater ([rclone downloads](https://beta.rclone.org/)). Versions prior to v1.59 may return `HTTP 401: Unauthorized` errors, as earlier versions of `rclone` do not strictly align to the S3 specification in all cases.
+{{</Aside}}
```
lukeed:
```suggestion
{{</Aside>}}
```

PR #4487, opened 2022-05-18
fix(r2 workers example): correct status code

Hydrophobefireman:
send 206,200 or 304 for successful responses

### DIFF
--- a/content/r2/examples/demo-worker.md
+++ b/content/r2/examples/demo-worker.md
@@ -18,7 +18,7 @@ function parseRange(encoded: string | null): undefined | { offset: number, lengt
     return
   }
 
-  const parts = encoded.split('-')
+  const parts = encoded.split("bytes=")[1]?.split("-") ?? []
   if (parts.length !== 2) {
     throw new Error('Not supported to skip specifying the beginning/ending byte at this time')
   }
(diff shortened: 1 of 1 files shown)


vlovich:
LGTM @deadlypants1973

content/r2/examples/demo-worker.md
```diff
@@ -18,7 +18,7 @@ function parseRange(encoded: string | null): undefined | { offset: number, lengt
     return
   }
 
-  const parts = encoded.split('-')
+  const parts = encoded.split("bytes=")[1]?.split("-") ?? []
```
lukeed, line 21: this is fine for an example i think, but a `Range` header may send a suffix-only value in this format:

```
Range: <unit>=-<suffix>
```

> [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range#syntax)

So splitting on `-` in this case & throwing on `length !== 2` (below) doesn't cover all valid conditions

lukeed (approved):
As mentioned, this is technically incomplete but probably good enough for a demo snippet. Thanks!

content/r2/examples/demo-worker.md
Hydrophobefireman: Yeah, even multiple ranges are allowed in a single header and I think at that point we could just have a comment to warn future copy-pasters?

lukeed: sure sounds good 👍

deadlypants1973 (approved):

vlovich (commented):
LGTM

PR #4488, opened 2022-05-18
[ZT] Fix dashes in VNET guide

KianNH:
These dashes are the wrong character & don't work when being copy/pasted into a terminal

image


ranbel (approved):
thanks for fixing the issue!

PR #4490, opened 2022-05-18
Add NETCAT explanation - Visible ports

fjbilbao99:
I saw several people complaining that ports are still open despite using this WAF rule. Customer's usually advise they can see the ports open using netcat. 
This WAF rule will not CLOSE the port as the ports will still open. This rule will just stop forwarding data on non standar ports. 
Adding this can reduce this kind of tickets from customers. 
Example JIRA: https://jira.cfops.it/browse/CUSTESC-18238

Thanks!
Xabier from CSUP team in EMEA

### DIFF
--- a/content/firewall/_partials/_open-ports-blocked-traffic.md
+++ b/content/firewall/_partials/_open-ports-blocked-traffic.md
@@ -1,0 +1,8 @@
+---
+_build:
+  publishResources: false
+  render: never
+  list: never
+---
+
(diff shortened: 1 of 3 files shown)


pedrosousa:
Note to self: We already have a similar note in one of the firewall rules examples:
https://developers.cloudflare.com/firewall/recipes/require-specific-http-ports/

We should decide where to keep the note.

pedrosousa:
Thanks @fjbilbao99 !
Since we also had this information in a separate page, I updated the text so that we could reuse the same paragraph in the two locations.

pedrosousa (approved):

PR #4494, opened 2022-05-18
[Workers] Clarified How Workers Works page

isaac-mcfadyen:
- In the bottom paragraph of How Workers Works, it is stated that "a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop."
- The confusing wording here is the word "concurrent". While a single Workers instance may handle multiple request sequentially (an "event loop") the processing of an instance of a Worker is never occurring at the same time on the same instance.
- This PR clarifies the wording to add an additional addendum that should help clarify.
- Changes are welcome, the main thing to avoid was users thinking that they would have to lock state and similar concepts that only apply when handling state in multiple threads.

### DIFF
--- a/content/workers/learning/how-workers-works.md
+++ b/content/workers/learning/how-workers-works.md
@@ -55,10 +55,9 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
 ## Distributed execution
 
 Isolates are resilient and continuously available for the duration of a request, but in rare instances isolates may be evicted. When a script hits official [limits](/workers/platform/limits/) or when resources are exceptionally tight on the machine the request is running on, the runtime will selectively evict isolates after their events are properly resolved.
-
-Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. While you do *not* have to program with concurrency in mind, other requests may be processed during awaiting any `async` tasks (such as fetch). If there are many requests queued, the Workers runtime may initialize another isolate to help handle the load but this is not guaranteed to happen at any specific number of incoming requests.
   
-Because of this, there is no guarantee whatsoever whether any two requests will land in the same instance; therefore it is inadvisable to set or mutate global state within the event handler.
+Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request. 
+Therefore we do *not* recommend you set or mutate global state within the event handler to avoid any concurrency issues.
   
 
 ## Related resources
--- a/content/workers/learning/how-workers-works.md
+++ b/content/workers/learning/how-workers-works.md
@@ -57,7 +57,7 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
 Isolates are resilient and continuously available for the duration of a request, but in rare instances isolates may be evicted. When a script hits official [limits](/workers/platform/limits/) or when resources are exceptionally tight on the machine the request is running on, the runtime will selectively evict isolates after their events are properly resolved.
   
 Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request. 
-Therefore we do *not* recommend you set or mutate global state within the event handler to avoid any concurrency issues.
+Therefore we do not recommend you set or mutate global state within the event handler to avoid any concurrency issues.
   
 
 ## Related resources
--- a/content/workers/learning/how-workers-works.md
+++ b/content/workers/learning/how-workers-works.md
@@ -56,7 +56,7 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
 
 Isolates are resilient and continuously available for the duration of a request, but in rare instances isolates may be evicted. When a script hits official [limits](/workers/platform/limits/) or when resources are exceptionally tight on the machine the request is running on, the runtime will selectively evict isolates after their events are properly resolved.
   
-Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request. 
+Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may (or may not) be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request. 
 Therefore we do not recommend you set or mutate global state within the event handler to avoid any concurrency issues.
   
 
--- a/content/workers/learning/how-workers-works.md
+++ b/content/workers/learning/how-workers-works.md
@@ -56,7 +56,7 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
 
 Isolates are resilient and continuously available for the duration of a request, but in rare instances isolates may be evicted. When a script hits official [limits](/workers/platform/limits/) or when resources are exceptionally tight on the machine the request is running on, the runtime will selectively evict isolates after their events are properly resolved.
 
-Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. For example, if Cloudflare receives 10 HTTP requests for the same Worker at exactly the same time, the Workers runtime will start 10 instances of that Worker. Any single Worker instance will never receive more than one request at the same time but may do so one after another.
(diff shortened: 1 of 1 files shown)


eugene1g:
It might help to clarify this with an example -
`For example, if Cloudflare receives 10 HTTP requests for the same Worker at exactly the same time, Cloudflare runtime will start 10 instances of that Worker. Any single Worker instance will never receive more than one request at the same time.`

eugene1g:
I think this documentation update might be incorrect - in my testing, one Worker _can_ be processing multiple HTTP requests at the same time. I created a proxy Worker that uses `fetch` to load images from another site. These `fetch` calls take some time, and while that is happening the same worker seems to be getting & handling new HTTP requests -

```js
// worker.mjs

const instanceId = Math.random(); // create a random identifier for this instance
let processingRequests = 0; // how many concurrent HTTP requests we are handling

export default {
  async fetch(req) {
    processingRequests++;

    let myResponse;
    const reqUrl = new URL(req.url);
    if (reqUrl.pathname === "/") {
      // As a dummy test, render a plain HTML page that will trigger several simultaneous requests to this worker
      const headers = new Headers();
      headers.set("content-type", "text/html");
      myResponse = new Response(
        `<img src="/100/100"/><img src="/200/200"/><img src="/300/300"/><img src="/400/400"/>`,
        { headers }
      );
    } else {
      // Setup a proxy fetch test images from another site
      const targetUrl = new URL(reqUrl.pathname, "https://loremflickr.com");
      const response = await fetch(targetUrl.href);
      myResponse = new Response(response.body, {
        headers: response.headers,
        status: response.status,
        statusText: response.statusText,
      });
    }

    myResponse.headers.set("aa-worker-instance-id", String(instanceId));
    myResponse.headers.set("aa-worker-processing-requests", String(processingRequests));
    processingRequests--;
    return myResponse;
  },
};

```

This Worker gets several simultaneous requests, and if there was a new instance for each request, I would expect that `Math.random()` would be different in each instance, and `processingRequests` would be `1` at all times. However, if I load the worker in one of my domains (e.g. a dummy instance at https://cfproxy1.compthrive.com), I see that all responses share the same `instanceId` and accurately keep track of concurrent requests, which tells me they all go through the same JS handler and there is only one instance of the Worker.

isaac-mcfadyen:
> As a test, I created a proxy Worker that uses `fetch` to load images from another site. These calls take some time, and while `fetch` is happening, this worker does receive & handle new HTTP requests -

Yeah, I believe that's because the Worker is essentially idle when handling `fetch` requests (since there's no execution of the actual code happening). If you were to run an Unbound Worker with a CPU intensive task (such as a loop crunching numbers) then during that loop no other requests would hit.

eugene1g:
I tried a running CPU-intensive task with a _Bundled_ worker, and all simultaneous request queued up to execute serially in the same instance of the worker. So the premise that a new instance of the Worker will be started isn't holding up, even when the JS loop is blocked.

IMO these tests establish two points - 
1. It is possible for the `fetch` function inside a Worker to be handling several HTTP events at the same time. In my test, the handler was waiting for a sub-request to finish, but that could've been waiting for DO/KV/db/REST call etc, which is a common use-case for workers.
2. If the JS thread inside the Worker is busy with a CPU-intensive task, it is possible that incoming HTTP events will queue up to wait for this loop to free up, instead of new Workers coming online to take care of them.

If these points are true, this update to the docs should be ignored because neither part of my earlier example holds up - neither `...if Cloudflare receives 10 HTTP requests for the same Worker at exactly the same time, the Workers runtime will start 10 instances of that Worker` nor `Any single Worker instance will never receive more than one request at the same time but may do so one after another.`

isaac-mcfadyen:
Hmm... I'll ask one of the engineers on the Workers team for some clarification. 🙂

isaac-mcfadyen:
Alright. So quick overview of what I was told.

1. Workers don't run requests concurrently, but *asynchronously*. What this means is that while the CPU is being used the Worker will be "locked" so that any new requests coming in will essentially wait.
2. However, any async tasks such as `fetch` will "unlock" the Worker, allowing other requests to process while waiting for a network response or other async task.
3. This happens at a relatively low numbers of requests. At a certain number of simultaneous requests, another isolate **will** be spun up. However this number isn't static (it fluctuates based on a number of factors) so they weren't able to tell me exactly at what point this happens.

Based on this, I'm going to revise this PR to remove some of the incorrect information but still make it clear that programming with concurrent requests in mind is not required with Workers.

content/workers/learning/how-workers-works.md
```diff
@@ -55,8 +55,10 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
 ## Distributed execution
 
 Isolates are resilient and continuously available for the duration of a request, but in rare instances isolates may be evicted. When a script hits official [limits](/workers/platform/limits/) or when resources are exceptionally tight on the machine the request is running on, the runtime will selectively evict isolates after their events are properly resolved.
-
-Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. There is no guarantee whatsoever whether any two requests will land in the same instance; therefore it is inadvisable to set or mutate global state within the event handler.
```
koeninger, line 59: The text you removed "There is no guarantee whatsoever..." is correct, why did you remove it?

isaac-mcfadyen: I reworded it to be less passive ("it is inadvisable" vs "we don't recommend").

Should we make it clearer about why we don't recommend this though?

koeninger: It's not just a reword, you're saying something different.  Paraphrasing, the original text said "requests MIGHT NOT hit the same instance, so don't mutate global state".  You're saying "requests MIGHT hit the same instance, so don't mutate global state".  They're both arguably good advice, but they're not the same thing.

isaac-mcfadyen: Hmm... alright. Would something like the following work better?


Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request.
 
However, requests are not guaranteed to hit the same instance every time. Therefore, we do *not* recommend you set or mutate global state within the event handler to avoid any concurrency issues.

eugene1g: (external dev here):

Both behaviors are true, and both are important to document. May I suggest something like this - 

> There is no guarantee that any two user requests will be routed to the same or a different instance of your Worker. Please be careful if you plan to use or mutate global state within the event handler.

content/workers/learning/how-workers-works.md
```diff
@@ -55,8 +55,10 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
...
+  
+Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request. 
+Therefore we do *not* recommend you set or mutate global state within the event handler to avoid any concurrency issues.
```
deadlypants1973:
```suggestion
Therefore we do not recommend you set or mutate global state within the event handler to avoid any concurrency issues.
```

isaac-mcfadyen: Would something like this comment work better? https://github.com/cloudflare/cloudflare-docs/pull/4494#discussion_r879744889

Because of this: https://github.com/cloudflare/cloudflare-docs/pull/4494#discussion_r879733700

deadlypants1973: @isaac-mcfadyen my only comment here was to remove the italics

isaac-mcfadyen: Oh haha, misread that 😆 

Yeah, sounds good to me.

content/workers/learning/how-workers-works.md
deadlypants1973: @isaac-mcfadyen if you are going with @eugene1g 's comment, does this PR need to be updated?

isaac-mcfadyen: Updating now, TY 🙏

content/workers/learning/how-workers-works.md
```diff
@@ -55,8 +55,9 @@ When a request to your `*.workers.dev` subdomain or to your Cloudflare-managed d
 ## Distributed execution
 
 Isolates are resilient and continuously available for the duration of a request, but in rare instances isolates may be evicted. When a script hits official [limits](/workers/platform/limits/) or when resources are exceptionally tight on the machine the request is running on, the runtime will selectively evict isolates after their events are properly resolved.
-
-Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. There is no guarantee whatsoever whether any two requests will land in the same instance; therefore it is inadvisable to set or mutate global state within the event handler.
+  
+Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may (or may not) be processed during awaiting any `async` tasks (such as fetch) if other requests come in while processing a request.
```
deadlypants1973:
```suggestion
Like all other JavaScript platforms, a single Workers instance may handle multiple requests including concurrent requests in a single-threaded event loop. That means that other requests may (or may not) be processed during awaiting any `async` tasks (such as `fetch`) if other requests come in while processing a request. 
```

deadlypants1973 (approved):

PR #4499, opened 2022-05-19
[SSL] Fix typo in http-strict-transport-security.md

ellierider:
- fix typo "form" -> "from"

> Prevents an attacker from downgrading a first request **form** HTTPS to HTTP


kodster28 (approved):
Thanks @ellierider

PR #4511, opened 2022-05-19
[R2] Add aws-sdk-php example

KianNH:
An example demonstrating how to connect to R2's S3 API using `aws-sdk-php-v3`.


vlovich (approved):
Not a php person but seems fine after a cursory glance. Better to land & fixup if issues come up later.

deadlypants1973 (approved):

PR #4520, opened 2022-05-20
[Workers] Add environments disabled note in both environments.md

rigor789:
I was looking for the environments drop-down on the dashboard and couldn't find it. It took me a little while to find the note in the docs, as I was primarily landing on the legacy environments docs.

There are 2 pages that include a note:
https://developers.cloudflare.com/workers/learning/using-services/
https://developers.cloudflare.com/workers/wrangler/environments/ (updated the note to match the wording of the first one)

Question: Would it make sense to also update the title of the **Environments** (https://developers.cloudflare.com/workers/platform/environments/) to mention "legacy" or wrangler 1?

### DIFF
--- a/content/workers/platform/environments.md
+++ b/content/workers/platform/environments.md
@@ -7,6 +7,8 @@ title: Environments
 
 {{<Aside type="note">}}
 
+If you are looking for the new [Service Environments](/workers/learning/using-services/) please note that we have temporarily disabled the creation of Service Environments while we are improving this feature.
+
 You can only use environments with [Wrangler](/workers/wrangler/).
 
 {{</Aside>}}
--- a/content/workers/platform/environments.md
+++ b/content/workers/platform/environments.md
@@ -7,8 +7,6 @@ title: Environments
 
 {{<Aside type="note">}}
 
-If you are looking for the new [Service Environments](/workers/learning/using-services/) note that we have temporarily disabled the creation of Service Environments while we are improving this feature.
-
 You can only use environments with [Wrangler](/workers/wrangler/).
 
 {{</Aside>}}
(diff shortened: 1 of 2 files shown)


Greatshopping (approved):
rigor789:fix/worker-environments-disabled-note

content/workers/platform/environments.md
```diff
@@ -7,6 +7,8 @@ title: Environments
 
 {{<Aside type="note">}}
 
+If you are looking for the new [Service Environments](/workers/learning/using-services/) please note that we have temporarily disabled the creation of Service Environments while we are improving this feature.
```
deadlypants1973:
```suggestion
If you are looking for the new [Service Environments](/workers/learning/using-services/) note that we have temporarily disabled the creation of Service Environments while we are improving this feature.
```

deadlypants1973 (approved):

content/workers/platform/environments.md
```diff
@@ -7,6 +7,8 @@ title: Environments
 
 {{<Aside type="note">}}
 
+If you are looking for the new [Service Environments](/workers/learning/using-services/) note that we have temporarily disabled the creation of Service Environments while we are improving this feature.
+
```
deadlypants1973:
```suggestion
```

deadlypants1973 (approved):
I removed the line about Service Environments because they are still under development. However, if you feel that there is a lot of user confusion between the two and we need to have a warning, we can put one. I feel that it is more confusing now to have a warning about a feature that is not live.

rigor789:
Works for me, my main source of confusion was that I did see the service environments on the dashboard a few weeks ago, but then they were disabled. I was starting to set up multiple environments for our project, and I believe I searched for "environments" and landed on the "legacy" docs, and it took me a little while to figure out that they are not the environments I saw on the dashboard earlier...

So perhaps not a warning, but a note/link to Service Environments somewhere on the page would be helpful for anyone in a similar boat? I'm personally set as-is, as I've connected the dots and it all makes sense now.

PR #4523, opened 2022-05-20
allowConcurrency option for Durable Object put

koeninger:
The header file (and thus the public type repo) has allowConcurrency in PutOptions, is there a reason for it not to be in docs?

### DIFF
--- a/content/workers/runtime-apis/durable-objects.md
+++ b/content/workers/runtime-apis/durable-objects.md
@@ -159,6 +159,10 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
     **Supported options:**
 
+    - {{<code>}}allowConcurrency{{<param-type>}}boolean{{</param-type>}}{{</code>}}
+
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/durable-objects.md
```diff
@@ -159,6 +159,10 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
     **Supported options:**
 
+    - {{<code>}}allowConcurrency{{<param-type>}}boolean{{</param-type>}}{{</code>}}
```
a-robinson, line 162: `allowConcurrency` doesn't do anything on writes. You can pass it to us, but it doesn't change behavior at all.

koeninger: got it - does it make more sense to document that here, or try to eliminate it from the types repo?

a-robinson: Sure, that'd be fair.

edit: removing it from the types repo, I mean.

PR #4529, opened 2022-05-23
Update DO example to allow for easily switching between DOs

bretthoerner:
I use this example as a starting point when I want to do simple testing or show someone something, and the first thing I _always_ do is make the object itself an argument.

I think hard coding `"A"` does new readers a disservice and could confuse them about DO namespaces/objects.

Most of the commentary is copy pasted from elsewhere in this document, FWIW.

### DIFF
--- a/content/workers/learning/using-durable-objects.md
+++ b/content/workers/learning/using-durable-objects.md
@@ -453,12 +453,30 @@ export default {
 };
 
 async function handleRequest(request, env) {
-  let id = env.COUNTER.idFromName("A");
+  let url = new URL(request.url);
+  let name = url.searchParams.get("name");
+  if (!name) {
+    return new Response(
+      "Select a Durable Object to contact by using" +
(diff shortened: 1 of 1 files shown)


bretthoerner:
The failure is odd, but I don't have permissions to view: https://github.com/cloudflare/cloudflare-docs/pull/4529#issuecomment-1134624832

a-robinson (approved):
Yeah this makes the example much better 👍

a-robinson:
I can't view the failure details either

koeninger:
Talking to pages team to see if there's a way for us to view these errors, this isn't the first time it's happened.

As far as the PR goes, I agree with the idea, but we should probably update the templates too - I can do that if we're all ok with this change

koeninger (approved):

koeninger:
tldr from pages team seems to be push another commit and see if it builds

PR #4539, opened 2022-05-23
[Load Balancing] Update SSL/TLS Coverage for LB dev docs

patriciasantaana:
PCX-3684


dcpena (approved):

robinp-tw:
Hey @patriciasantaana ! I came over the change, and it is not very clear, if the LB subdomain is `lb`, then why does the example add a record for `www` subdomain? Also, what is the `LB` column for the `A` record? That column is normally name.

I would like to understand this before I make related changes. Thank you!

PR #4565, opened 2022-05-25
[Cloudflare One] fix Support URL heading in warp-settings.md

colinrymer:


ranbel (approved):
Good catch, thanks for fixing it!

PR #4569, opened 2022-05-25
[ZT] Change dedicated destination ip to dns resolver IP

ranbel:


ranbel:
Hey @Encore-Encore what do you think about this update? Tbh I am not sure we even need this page, since the first few paragraphs repeat what's in https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/locations/

The example policy could be moved to the DNS Common Configurations page.

ranbel:
see #4596

PR #4571, opened 2022-05-26
replace CODEOWNERS file

lukeed:
... while keeping it as a backup just in case

In an effort to reduce notification spam (for many many people), instead of auto-requesting reviews from entire team aliases, this sets up a GitHub Action to listen to the `pull_request.opened` event and then, using the same `OWNERS` information that the `assign-issue` (6a1d52042) action & workflow uses, assigns a PCX team member to review the PR.

> **Note:** Internally, each PCX member already has their own product assignments. This `owners.ts` mapping reflects those relationships & is what dictates who's assigned where.

If/when a PR is authored by the PCX owner (eg, @deadlypants1973 creates a PR for Workers/Pages/R2), then the @cloudflare/pcx team is requested as a reviewer instead. While this is still a frequent scenario, it should still cut down on the number of total notifications the PCX team (and all former codeowners) receive daily.

Additionally, if an external PR is created for a product with no PCX owner, then @haleycode is assigned (_not review-requested_) so that a PCX member can be assigned manually.

---

With CODEOWNERS gone, the repository settings (will, as of merge) still require that requested-reviewers approve the PR before merging. This should mimic the pre-CODEOWNERS GitHub era of merge/branch protection.

---

> **Note:** The `CODEOWNERS` file needed to be removed entirely because, programmatically, there is no way to ask GitHub/Ocktokit who the owners for the PR/file-list are. This means that _even if_ we were to replace every instance of `@cloudflare/pcx` in the `CODEOWNERS` file with individuals' usernames, we:
> 1. wouldn't be able to know if an owner created a PR without duplicating the information elsewhere (eg, `owners.ts` _and_ `CODEOWNERS`); and
> 2. would still "Require approval from Code Owners before merge" and yet authors cannot approve their own PRs


.github/actions/assign-pr/index.ts
```diff
@@ -0,0 +1,139 @@
...
+      // initiate "@cloudflare/pcx" team review
+      await client.rest.pulls.requestReviewers({
+        repo: repository.name,
+        owner: repository.owner.login,
+        pull_number: prnumber,
+        team_reviewers: ['@cloudflare/PCX'],
+      });
+    } else {
+      // cannot self-review
+      users.delete(author);
```
lukeed, line 114: note to self, fix me

PR #4576, opened 2022-05-26
Fix `kv_namespaces` position in wrangler config documentation

askolesov:
`kv_namespaces` is a top level config property (not a property of `[vars]`). Spent some time figuring this out.


lukeed:
Had it been `[[kv_namespaces]]` format, it wouldn’t have fallen under the vars namespace. TOML is weird

askolesov:
Agreed, `[[kv_namespaces]]` is even better option

deadlypants1973:
> Agreed, `[[kv_namespaces]]` is even better option

Hi @askolesov , Can you rewrite this PR with this option instead? :)

deadlypants1973:
fixed in #5051

PR #4590, opened 2022-05-26
[Stream] Update creator in existing videos

dcpena:
Added example for updating the creator property in existing videos to address [PCX-3958](https://jira.cfops.it/browse/PCX-3958).

### DIFF
--- a/content/stream/uploading-videos/creator-id.md
+++ b/content/stream/uploading-videos/creator-id.md
@@ -84,6 +84,16 @@ highlight: [35]
 }
 ```
 
+## Update creator in existing videos
+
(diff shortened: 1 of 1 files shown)


content/stream/uploading-videos/creator-id.md
```diff
@@ -84,6 +84,16 @@ highlight: [35]
 }
 ```
 
+## Update creator in existing videos
+
+To update the creator property in existing videos, make a POST request to the video object endpoint with a JSON payload specifying the creator property as show in the example below.
```
kodster28, line 89: I'd maybe link out to the api docs on the POST (if that endpoint is documented).

kodster28 (approved):

PR #4592, opened 2022-05-26
update partners file

lukeed:
take 2, has no owners, should assign to haley


lukeed:
cool, it (eventually) worked 😅

PR #4601, opened 2022-05-27
Fix logic in block-ip-reputation.md

AndrewLane:
Current logic seems to block all traffic from CN, TW, US, and GB even if the threat_score is 0


pedrosousa:
Thank you for the fix, @AndrewLane!

pedrosousa (approved):

PR #4614, opened 2022-05-28
[Pages] Clarify individual header length limit

KianNH:
Similar to the redirects documentation, we should include the amount of headers you can declare and the maximum length of a given header value.

Ref:
https://developers.cloudflare.com/pages/platform/limits/#headers
https://community.cloudflare.com/t/cloudflare-pages-header-file-line-limit/387052/5?u=kiannh

### DIFF
--- a/content/pages/platform/headers.md
+++ b/content/pages/platform/headers.md
@@ -56,7 +56,7 @@ An incoming request which matches multiple rules' URL patterns will inherit all 
 
 {{</table-wrap>}}
 
-A project is limited to 100 header rules. Each line in the _headers file has a 1,000 character limit. The entire line including spacing, header name and value all count towards this limit.
+A project is limited to 100 header rules. Each line in the `_headers` file has a 1,000 character limit. The entire line, including spacing, header name, and value, counts towards this limit.
 
 If a header is applied twice in the `_headers` file, the values are joined with a comma separator. Headers defined in the `_headers` file override what Cloudflare Pages ordinarily sends, so be aware when setting security headers. Cloudflare reserves the right to attach new headers to Pages projects at any time in order to improve performance or harden the security of your deployments.
 
--- a/content/pages/platform/headers.md
+++ b/content/pages/platform/headers.md
@@ -56,6 +56,8 @@ An incoming request which matches multiple rules' URL patterns will inherit all 
 
 {{</table-wrap>}}
 
+A project is limited to 100 header rules. Each header declaration has a 1,000-character limit.
+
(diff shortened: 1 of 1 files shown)


Cyb3r-Jak3:
@KianNH Do you not think it would go better in the [limits section](https://github.com/cloudflare/cloudflare-docs/blob/production/content/pages/platform/limits.md)?

KianNH:
Probably - I was just following along with how we do it in `_redirects` where the overall limit of the given rules (i.e 100 for `_headers` and 2,100 for `_redirects`) is in [limits](https://github.com/cloudflare/cloudflare-docs/blob/production/content/pages/platform/limits.md) and the per-rule character length is in the individual file, i.e [redirects](https://github.com/cloudflare/cloudflare-docs/blob/production/content/pages/platform/redirects.md)

Could put it in both? It'd make sense that if you know how to use them but want to know about limits, you might only check the specific limits page.

Cyb3r-Jak3:
Personally, I think it would be best to put it in both. When I get an issue, I check the limit's section to see if it's a known limitation. It probably also doesn't hurt to put it in two places even if it might be [changed](https://community.cloudflare.com/t/cloudflare-pages-header-file-line-limit/387052/8)

WalshyDev:
Yep we should definitely put it in both. Appreciate the PR!

content/pages/platform/headers.md
```diff
@@ -56,6 +56,8 @@ An incoming request which matches multiple rules' URL patterns will inherit all
 
 {{</table-wrap>}}
 
+A project is limited to 100 header rules. Each header declaration has a 1,000-character limit.
```
WalshyDev: Should this be an aside?

KianNH: I just stole the wording as-is from the redirects page - if we turn it into an aside here then we should there too imo

KianNH:
Updated the limit from 1,000 to 2,000

deadlypants1973 (approved):

KianNH:
The limit bump to 2k hasn't been released yet pending testing so put it back to 1k - also clarified that the limit is for the entire line, not just the header value.

content/pages/platform/headers.md
```diff
@@ -56,7 +56,7 @@ An incoming request which matches multiple rules' URL patterns will inherit all
 
 {{</table-wrap>}}
 
-A project is limited to 100 header rules. Each header declaration has a 2,000-character limit.
+A project is limited to 100 header rules. Each line in the _headers file has a 1,000 character limit. The entire line including spacing, header name and value all count towards this limit.
```
deadlypants1973:
```suggestion
A project is limited to 100 header rules. Each line in the `_headers` file has a 1,000 character limit. The entire line, including spacing, header name, and value, counts towards this limit.
```

deadlypants1973 (approved):

PR #4615, opened 2022-05-28
[Pages] Add header length limit

Cyb3r-Jak3:
* Adds a note to pages headers limit about max header length of 1000


Cyb3r-Jak3:
Was beaten by https://github.com/cloudflare/cloudflare-docs/pull/4614

PR #4627, opened 2022-05-30
Update Wrangler command docs

petebacondarwin:
Add command docs for:

- r2 bucket ...
- pages ...

Closes https://github.com/cloudflare/wrangler2/issues/1059

PR #4644, opened 2022-06-01
Revert "Update Wrangler command docs"

deadlypants1973:
Reverts cloudflare/cloudflare-docs#4627

@petebacondarwin 

Pages `create` and `publish` are described as the same command. What is the difference between these two?


content/workers/wrangler/commands.md
```diff
@@ -835,144 +785,6 @@ After starting `wrangler tail`, you will receive a live feed of console and exce
...
-{{<definitions>}}
-
-- `--project-name` {{<type>}}string{{</type>}}
-  - The name of the project you would like to list deployments for.
-
-{{</definitions>}}
-
-#### `create`
-
-Publish a directory of static assets as a Pages deployment.
```
deadlypants1973, line 926: @petebacondarwin the description for `create` matches that of `publish`. What is the difference?

sidharthachatterjee: `create` and `publish` are aliases of each other.

petebacondarwin: Thanks for the info @sidharthachatterjee - should we only have one in the docs? And if so, which one?

petebacondarwin:
@deadlypants1973 - what is the process here? Do we merge this and then I resubmit the original PR with amendments? Or do we make the amendments in this PR so that it only reverts the bits we are not happy with?

kodster28 (approved):

content/workers/wrangler/commands.md
sidharthachatterjee: I think having `publish` first class is a good idea. 

`wrangler pages publish <dir>`

petebacondarwin: OK thanks.

PR #4654, opened 2022-06-01
[Load Balancing] Geo Steering Note about China

patriciasantaana:
PCX-4081


kodster28:
I'd suggest you actually add it to the top of country steering as a non-note since it applies to both api/dash.

kodster28 (approved):

PR #4668, opened 2022-06-02
[SSL] Moved backup cert info to new page

kodster28:
Addresses PCX-4018

### DIFF
--- a/content/ssl/ssl-tls/cipher-suites.md
+++ b/content/ssl/ssl-tls/cipher-suites.md
@@ -44,6 +44,10 @@ We no longer support [RC4 cipher suites](https://blog.cloudflare.com/end-of-the-
 | AEAD-AES256-GCM-SHA384 \[^1]        | ❌      | ❌      | ❌      | ✅      |
 | AEAD-CHACHA20-POLY1305-SHA256 \[^1] | ❌      | ❌      | ❌      | ✅      |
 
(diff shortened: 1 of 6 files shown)


content/ssl/ssl-tls/cipher-suites.md
```diff
@@ -44,6 +44,10 @@ We no longer support [RC4 cipher suites](https://blog.cloudflare.com/end-of-the-
 | AEAD-AES256-GCM-SHA384 \[^1]        | ❌      | ❌      | ❌      | ✅      |
 | AEAD-CHACHA20-POLY1305-SHA256 \[^1] | ❌      | ❌      | ❌      | ✅      |
```
pedrosousa: Nit: Hadn't noticed it yet, but I'm not sure if this was the intended rendering of the footnote... The following formatting in R2 looks slightly better, and the notes are right after the table:
https://developers.cloudflare.com/r2/platform/limits/#account-plan-limits

pedrosousa (approved):

content/ssl/ssl-tls/cipher-suites.md
kodster28: Updated!

kodster28: And good suggestion.

PR #4671, opened 2022-06-02
[R2] Make cursor of R2Objects optional

KianNH:
As per the description next to it (`Only present if truncated is true.`) and the [workers-types](https://github.com/cloudflare/workers-types/blob/92f8e38fc62e23801b9bd6cc153b6cb44b96834e/index.d.ts#L1124) definitions, `cursor` is optional and only appears if necessary (`list` operation didn't return all eligible objects).

```ts
interface R2Objects {
  objects: R2Object[];
  truncated: boolean;
  cursor?: string;
  delimitedPrefixes: string[];
}
```


MateuO1 (commented):
Yes

deadlypants1973 (approved):

PR #4673, opened 2022-06-02
[R2] Explicitly define `range` on R2GetOptions

KianNH:
My wording is probably horrible, I'm not good at being concise.

Anyhow - thought it should be explicit that `range` is apart of `R2GetOptions` by defining it similar to `onlyIf` which refers to the particular section that expands upon that property.


vlovich:
LGTM

deadlypants1973 (approved):

PR #4680, opened 2022-06-03
fix: Remove extra spacing after links

pedrosousa:
Internal links have some extra empty space when they are followed by a punctuation mark. 

This seems to be caused by the newline between `</span>` and `</a>` in `render-link.html`. Removing the spaces right before the closing `</a>` didn't solve the issue, so I put all the HTML in the same line.

image


lukeed (commented):
I can no longer approve, but ✅ Approved haha

kodster28 (approved):
Well, I suppose we can trust @lukeed ;)

PR #4692, opened 2022-06-03
[Workers] Optimize phrasing around what kind of metadata `list` returns

valwalt:
@koeninger here's a minimal change that should eliminate any confusion around what kind of "values" `list` can return.

re:
image


koeninger:
@ValentinWalter thanks
I thought what tripped you up was not realizing you had to explicitly put the values in metadata?
Should we be saying something like "If your values fit in the metadata size limit, consider storing them in metadata instead" and giving an example call?

valwalt:
@koeninger what do you think of this?

koeninger (approved):

koeninger:
Seems clearer to me, thanks

deadlypants1973 (approved):

PR #4763, opened 2022-06-09
[ZT] Gateway authenticated doh

ranbel:


ranbel:
According to @jroyal , the "Enable DoH functionality for the account" API is currently a `GET` but should be a `POST`. These instructions are written for `POST`.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,292 @@
...
+  "success": true,
+  "errors": [],
+  "messages": []
+}
+```
+
+</div>
+</details>
+
+### 2. Enable DoH functionality for the account
```
jgpaiva: I think this should be:
```suggestion
### 2. Enable DoH functionality for the service token
```
I.e. if the user creates another service token and tries to follow these instructions by skipping this one, some of the other requests below will fail.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,292 @@
...
+  },
+  "success": true,
+  "errors": [],
+  "messages": []
+}
+```
+
+</div>
+</details>
+
```
jgpaiva, line 234: Might make sense to make a note here that the above steps are related with provisioning (i.e. one-time activity), whereas the ones below would happen during normal operation.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,292 @@
...
+### 4. Generate a DoH token for the user
+
+Request a DoH token for the user, using your service token to authenticate into your [team domain](/cloudflare-one/glossary/#team-domain).
```
jgpaiva, line 243: Might make sense to mention here that in general clients would be expected to have a refresh flow for the doh token instead of generating a new one for every doh query (even for performance reasons on their side, to avoid having to make 2 requests for every doh lookup).

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,292 @@
...
+Your DNS queries will now be sent to Gateway for filtering. If the site is blocked, the query will return the IP address of the Gateway block server (`162.159.36.12`).
+
+<details>
+<summary>Example response</summary>
+<div>
+
+```bash
+  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                 Dload  Upload   Total   Spent    Left  Speed
+100   187  100   187    0     0    968      0 --:--:-- --:--:-- --:--:--  1010
```
jgpaiva: Using `curl -s` avoids this extra data

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,292 @@
...
+      "TTL": 60,
+      "data": "162.159.36.12"
+    }
+  ]
+}
+```
+</div>
+</details>
+
+You can verify that the request was associated with the correct user email by checking your [Gateway DNS logs](/cloudflare-one/analytics/logs/activity-log/). To filter these requests, build a DNS policy using the [User Email](/cloudflare-one/policies/filtering/identity-selectors/#user-email) selector.
```
jgpaiva: This needs some extra testing, but I'm fairly sure users can use any of the [identity-based policies](https://developers.cloudflare.com/cloudflare-one/policies/filtering/identity-selectors/). I'd expect users to be particularly interested in the `identity.groups.name` field, which they should be able to populate with a user creation like:
```
--data '{"name":"John Doe", "email":"jdoe@acme.com","groups":["id": "02fk6b3p3majl10", "email": "awesome@acme.com", "name": "Awesome"}]}'
```
This may need confirming with the Access team if it's intended to be possible or just a side-effect of implementation.

content/cloudflare-one/static/documentation/connections/authenticated-doh.py
```diff
@@ -0,0 +1,178 @@
...
+def check_for_command(command, install):
+    try:
+        subprocess.check_output(["command", "-v", command])
+    except:
+        print(
+            f"Couldn't find {command} command. Please install with `{install}`")
+        exit(1)
+ 
+ 
+check_for_command("wg", "brew install wireguard-tools")
```
jgpaiva: oh, this is from an old version of this script, it's not needed anymore.

content/cloudflare-one/static/documentation/connections/authenticated-doh.py
```diff
@@ -0,0 +1,178 @@
...
+check_for_command("curl", "brew install curl")
```
jgpaiva: we may want to add `jq` as a requirement:
`check_for_command("jq", "brew install jq")`
(notice these checks are OS-dependent, the 'please install with x' suggestions only work for macOS)

content/cloudflare-one/static/documentation/connections/authenticated-doh.py
```diff
@@ -0,0 +1,178 @@
...
+        pprint(response)
+    return response['token']
+ 
+ 
+def request_post(url, body):
+    return request("POST", url, body)
+ 
+ 
+def request_put(url, body):
+    return request("PUT", url, body)
```
jgpaiva, line 84: can be deleted, this isn't used anymore.

content/cloudflare-one/static/documentation/connections/authenticated-doh.py
```diff
@@ -0,0 +1,178 @@
...
+    return (client_id, client_secret)
+ 
+ 
+def request_list_service_tokens():
+    return request_get(f"{CF_API}/{account_tag}/access/service_tokens/")
+ 
+ 
+def request_enable_service_token_for_doh(service_token_id):
+    url = f"{CF_API}/{account_tag}/access/organizations/doh/{service_token_id}"
+    return request_post(url)
```
jgpaiva: missing a second param for the body, this will fail to call the function. Maybe:
```suggestion
    return request_post(url,"")
```

content/cloudflare-one/static/documentation/connections/authenticated-doh.py
jgpaiva:
```suggestion
verbose = os.environ.get('VERBOSE', False)
 
 
def check_for_command(command):
    try:
        subprocess.check_output(["command", "-v", command])
    except:
        print(
            f"Couldn't find required {command} command.")
        exit(1)
 
 
check_for_command("jq")
check_for_command("curl")
```
Condensed the two comments above into this suggestion (removed `wg`, added `jq`, removed os-dependent install instructions)

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
ranbel: @Encore-Encore

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
ranbel: Good point, I didn't know the token was one-time use. 

@jroyal How do you refresh a doh token?

jgpaiva: sorry, I probably created confusion. The doh token is not one-time use, but AFAIK it is only valid for a limited amount of time (I can't check now, but if you generate one and paste it in jwt.io, it should show the expiry date).
Clients can re-generate it by making the request in step 4 again (all previous steps don't need to be done again).

ranbel: When I checked the token on jwt.io, it showed a 24-hour duration. 
I also noticed was that I had to refresh the service token before I could use it to generate another DoH token. Otherwise I would get an "invalid service token" error even though the service token was not expired. Is that supposed to happen?

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
ranbel: When I try creating a user with the "groups" field, the response does not contain that data. How would you read back a user's group data?
```
{
    "result": {
        "id": "d74bcf54-3fc8-47dc-92c2-dd34f4697a5c",
        "created_at": "2022-06-14T23:52:19.539387787Z",
        "updated_at": "2022-06-14T23:52:19.539387787Z",
        "uid": "d74bcf54-3fc8-47dc-92c2-dd34f4697a5c",
        "name": "John Doe",
        "email": "jdoe@acme.com"
    },
    "success": true,
    "errors": [],
    "messages": []
}
```
I thought that it is up to the third-party reseller to come up with their own identity system. A group would be represented by a single "user", and all queries from the group would use the same DoH token.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
jgpaiva: > Is that supposed to happen?

no, this never happened during my testing. @jroyal is this expected?

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
jgpaiva: I think this needs input from @jroyal

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
jroyal: Hmm that shouldn't happen unless you set your service token duration super low, but the default is one year.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
jroyal: The create user struct is the following. I think you just have to move your groups block into the `custom` block. 
```golang
type UserCreateRequest struct {
	Name   string                 `json:"name"`
	Email  string                 `json:"email"`
	Custom map[string]interface{} `json:"custom,omitempty"`
}
```

jgpaiva: Fantastic, that does work. If the user is created with something like:
`--data '{"name":"John Doe", "email":"jdoe@acme.com","custom": {"groups":[{"id": "02fk6b3p3majl10", "email": "awesome@acme.com", "name": "Awesome"}]}}'`
When we can create a policy such as:
image
And that user will be matched according to its group.

Since this feature allows our customers to better manage big deployments, I think it might make sense to get it documented also.

ranbel:
@jroyal can you let me know when the "Enable DoH functionality for the service token" command has been changed to `POST`? I'm waiting for that to be fixed in the backend before merging the doc.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,310 @@
...
+
+Send DoH queries to the resolver at `https://<ACCOUNT_ID>.cloudflare-gateway.com/dns-query`, making sure to include the user's DoH token in the `CF-Authorization` header.
+
+```bash
+curl -s 'https://<ACCOUNT_ID>.cloudflare-gateway.com/dns-query?name=example.com' \
+     -H 'accept: application/dns-json' \
+     -H 'CF-Authorization: <USER_DOH_TOKEN>' | jq
+```
+
+If the site is blocked, the query will return the IP address of the Gateway block server (`162.159.36.12`).
```
jgpaiva: FYI, this only happens if the user set up a block page. If no block page is enabled on the rule, the response is `0.0.0.0`.

jroyal:
@ranbel will do. It's in review right now.

content/cloudflare-one/connections/connect-devices/agentless/dns/dns-over-https.md
```diff
@@ -0,0 +1,310 @@
...
+}
+```
+
+</div>
+</details>
+
+### 2. Enable DoH functionality for the service token
+
+```bash
+curl -X POST "https://api.cloudflare.com/client/v4/accounts/<ACCOUNT_ID>/access/organizations/doh/<ID>" \
```
jroyal: Sorry to change this last minute but we switched this to a PUT.

ranbel: Made the change! If there are no other updates needed, would you mind approving the PR?

jgpaiva: I added one more commit to this PR, to also fix the .py script to use PUT instead of POST. However, I think this change is not yet live: I tried the .py script yesterday and got an `HTTP/2 405` (method not allowed) error.

ranbel: I tested the PUT .py script and it seems to be working now.

jgpaiva: yup, I think we're good to merge now!

jgpaiva (approved):

PR #4771, opened 2022-06-10
[Pages] Add $ to sh codeblocks for plugins

KianNH:
You can't currently highlight them to copy/paste due to the missing `$`

without:

image

with:

image


WalshyDev:
Thanks!

Can you also add the missing ones in https://developers.cloudflare.com/pages/framework-guides/deploy-a-qwik-site/

WalshyDev (approved):

deadlypants1973 (approved):

PR #4773, opened 2022-06-10
[Pages] fix: pages functions headers modify code example

Cherry:
This fixes an issue with the example in the Pages Functions docs.

The example works in _local development_, but fails in production. My suggested change works in both environments, until local and production can operate the same way (cc @GregBrimble).

This tripped someone up as recently as yesterday [in Discord](https://canary.discord.com/channels/595317990191398933/910978223968518144/984591489998946354) as they started seeing errors in production despite things working fine in local dev.

### DIFF
--- a/content/pages/platform/functions/_index.md
+++ b/content/pages/platform/functions/_index.md
@@ -181,7 +181,8 @@ const errorHandler = async ({ next }) => {
 };
 
 const hello = async ({ next }) => {
-  const response = await next();
+  const rawResponse = await next();
+  const response = new Response(rawResponse.body, rawResponse); // clone response to create mutable headers
   response.headers.set('X-Hello', 'Hello from functions Middleware!');
   return response;
 };


GregBrimble:
Hey @Cherry , thanks for the PR and issue you filed in wrangler. I think we'd rather take a shot at fixing this in prod rather than documenting this work-around. If it takes us longer than we expect, we can revisit this docs PR. I'll close it out for now and escalate your wrangler issue.

Also, just a little side note, when cloning Responses, you should be careful to add a check for null-body-status:

```js
// https://fetch.spec.whatwg.org/#null-body-status
 return new Response(
  [101, 204, 205, 304].includes(response.status) ? null : response.body,
  response
);
```

That has bitten me numerous times!

Cherry:
Thanks @GregBrimble! And good call on the null body check!

Is there any ETA for when a fix may land in production? It’s been an issue for weeks (maybe months?) and regularly catches people out in the Discord, etc.

PR #4800, opened 2022-06-13
docs: use working CIDs in DNSLink docs

lidel:
Cosmetic changes:

- Replaces old CID with working one (wikipedia snapshot, pinned by many nodes)
- Fixes `/ipns/` content path to include example of a real and working DNSLink name
- Replaced path gateway with subdomain one
- Replaced "DNS Link" with "DNSLink", making it easier for people to search


fariborz4:
Ok

kodster28 (approved):

PR #4813, opened 2022-06-14
Add a beta section for Workers

Albertozhao:
Screen Shot 2022-06-15 at 7 36 05 AM

Screen Shot 2022-06-15 at 7 37 09 AM


KianNH:
Might be worth adding in PubSub?

Edit: nevermind, that's not strictly apart of Workers.

Albertozhao:
> Might be worth adding in PubSub?
> 
> Edit: nevermind, that's not strictly apart of Workers.

Added Pub/Sub since you're right, it's related! Pub/Sub Workers is one of the roadmap priorities.

PR #4821, opened 2022-06-15
Update macos.md

geelen:
These instructions didn't work for me as written, but I found https://github.com/cloudflare/cloudflared/issues/327#issuecomment-1087792549 when googling the error message. Seems like something's changed about the syntax?

However, I did follow the instructions in https://blog.cloudflare.com/ridiculously-easy-to-use-tunnels/ (linked from that GH issue) and migrated the config into the Dash. It also gave me the `sudo cloudflared service install` command to run, too.

Maybe this whole page can be updated given the new Dash stuff? Otherwise the proposed change here might work for some people...

FWIW, I was following this guide: https://developers.cloudflare.com/cloudflare-one/tutorials/ssh/ . I'm guessing a few of those steps can be skipped now?

### DIFF
--- a/content/cloudflare-one/connections/connect-apps/run-tunnel/as-a-service/macos.md
+++ b/content/cloudflare-one/connections/connect-apps/run-tunnel/as-a-service/macos.md
@@ -32,7 +32,7 @@ You can install the service to either run at login or at boot.
 Open a terminal window and run the following command:
 
 ```sh
-$ cloudflared service install
+$ cloudflared service install $(cloudflared tunnel token <TUNNEL>)
 ```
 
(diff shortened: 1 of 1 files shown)


ranbel:
@abelinkinbio 
Shouldn't `cloudflared service install` work without the token? Do you only need the token for a remotely-managed tunnel?

abelinkinbio:
> @abelinkinbio Shouldn't `cloudflared service install` work without the token? Do you only need the token for a remotely-managed tunnel?

That's correct. For locally managed tunnels, the token is not necessary. This should be resolved. I believe we can decline this PR as this will be better clarified in our refactor.

ranbel:
We are in the process of refactoring the tunnel docs, which will hopefully clarify dashboard vs. CLI workflows.

PR #4822, opened 2022-06-15
Change hyphens to underscores in several cookie names

timmc-edx:
I've observed these particular names with underscores, but not with hyphens, and I suspect this fix also needs to be made for other cookies listed on this page.


kodster28 (approved):
Thanks for the fix @timmc-edx (as well as the suggestion to check on the other ones).

Your instincts were right!

PR #4836, opened 2022-06-16
Remove a duplicate column

cicku:
For https://developers.cloudflare.com/ssl/ssl-tls/browser-compatibility


kodster28 (approved):

kodster28:
Thanks for the catch and fix @cicku

PR #4838, opened 2022-06-16
[SSL/TLS] fix section anchor in "Set up authenticated origin pulls"

outloudvi:
* Fix an anchor typo and update the corresponding description.


kodster28 (approved):

kodster28:
Thanks for the catch and fix @outloudvi

PR #4842, opened 2022-06-16
[R2] example for using cache-api with R2

taylorlee:
This creates an R2-specific cache-api example, by embedding the R2 [get-started-snippet](https://developers.cloudflare.com/r2/get-started/#5-access-your-r2-bucket-from-your-worker) into the [cache-api-example](https://developers.cloudflare.com/workers/examples/cache-api/)

### DIFF
--- a/content/r2/examples/cache-api.md
+++ b/content/r2/examples/cache-api.md
@@ -12,50 +12,50 @@ layout: example
 ---
 
 ```js
-async function handleRequest(event) {
-  const request = event.request;
-  const cacheUrl = new URL(request.url);
 
-  // Construct the cache key from the cache URL
-  const cacheKey = new Request(cacheUrl.toString(), request);
-  const cache = caches.default;
+export default {
+  async fetch(request, env, context) {
+    try {
+      const cacheUrl = new URL(request.url);
 
-  // Check whether the value is already available in the cache
-  // if not, you will need to fetch it from origin, and store it in the cache
-  // for future access
-  let response = await cache.match(cacheKey);
+      // Construct the cache key from the cache URL
+      const cacheKey = new Request(cacheUrl.toString(), request);
+      const cache = caches.default;
 
-  if (!response) {
-    console.log(
-      `Response for request url: ${request.url} not present in cache. Fetching and caching request.`
-    );
-    // If not in cache, get it from origin
-    response = await fetch(request);
+      // Check whether the value is already available in the cache
+      // if not, you will need to fetch it from origin, and store it in the cache
+      // for future access
+      let response = await cache.match(cacheKey);
 
-    // Must use Response constructor to inherit all of response's fields
-    response = new Response(response.body, response);
+      if (response) {
+        console.log(`Cache hit for: ${request.url}.`);
+        return response
+      }
 
-    // Cache API respects Cache-Control headers. Setting s-max-age to 10
-    // will limit the response to be in cache for 10 seconds max
+      console.log(
+        `Response for request url: ${request.url} not present in cache. Fetching and caching request.`
+      );
+      // If not in cache, get it from origin
+      response = await fetch(request);
 
-    // Any changes made to the response here will be reflected in the cached value
-    response.headers.append('Cache-Control', 's-maxage=10');
+      // Must use Response constructor to inherit all of response's fields
+      response = new Response(response.body, response);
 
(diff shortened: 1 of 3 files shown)


content/r2/examples/cache-api.md
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+      console.log(
+        `Response for request url: ${request.url} not present in cache. Fetching and caching request.`
+      );
+
+      // If not in cache, get it from R2
+      const objectKey = url.pathname.slice(1);
+      const object = await env.MY_BUCKET.get(objectKey);
+      if (!object || !object.body) {
```
mpint: Will an `object` ever not have an `object.body`?

content/r2/examples/cache-api.md
```diff
@@ -0,0 +1,70 @@
...
+      const cacheKey = new Request(url.toString(), request);
+      const cache = caches.default;
+
+      // Check whether the value is already available in the cache
+      // if not, you will need to fetch it from R2, and store it in the cache
+      // for future access
+      let response = await cache.match(cacheKey);
+
+      if (response) {
+        console.log(`Cache hit for: ${request.url}.`);
```
mpint, line 30: doc level nit: could drop the ending period after the url

content/r2/examples/cache-api.md
```diff
@@ -0,0 +1,70 @@
...
+
+```js
+
+export default {
+  async fetch(request, env, context) {
+    try {
+      const url = new URL(request.url);
+
+      // Construct the cache key from the cache URL
+      const cacheKey = new Request(url.toString(), request);
```
mpint, line 21: doc level nit: worth using `request.url` here to avoid introducing another concept (i.e. `toString()`).  then you could move the `url` declaration down to L38 to keep things a bit tighter

content/r2/examples/cache-api.md
```diff
@@ -0,0 +1,70 @@
...
+      // If not in cache, get it from R2
+      const objectKey = url.pathname.slice(1);
+      const object = await env.MY_BUCKET.get(objectKey);
+      if (!object || !object.body) {
+        return new Response('Object Not Found', { status: 404 });
+      }
+
+      // Set the appropriate object headers
+      const headers = new Headers();
+      object.writeHttpMetadata(headers);
```
mpint, line 47: not sure if I understand `object.writeHttpMetadata(headers)`. Is this writing empty headers or are the ones set below somehow absorbed in?

taylorlee: The `writeHttpMetadata` method iterates through the [R2HTTPMetadata](https://github.com/cloudflare/workers-types/blob/ca1c65b92b71fdb666c358576302a4a1573f8fda/index.d.ts#L1050) fields of the object and adds each of them to the passed Headers object. The headers below are added afterwards in addition.

content/r2/examples/cache-api.md
taylorlee: @deadlypants1973 do we generally prefer full sentences (including periods) in example logging? If not I can also update https://developers.cloudflare.com/workers/examples/cache-api/

content/r2/examples/cache-api.md
taylorlee: I believe this is done in all the cache-api examples for scripts that are not yet using the new URL compatibility flag: https://developers.cloudflare.com/workers/platform/compatibility-dates/#new-url-parser-implementation

I think we should keep the cache-api examples consistent for now.

content/r2/examples/cache-api.md
taylorlee: Good catch, no. That seems like an older example; I'll push a commit unify some of the existing R2 examples. Though [R2ObjectBody](https://github.com/cloudflare/workers-types/blob/ca1c65b92b71fdb666c358576302a4a1573f8fda/index.d.ts#L1105) is extremely similar to [Body](https://github.com/cloudflare/workers-types/blob/ca1c65b92b71fdb666c358576302a4a1573f8fda/index.d.ts#L107). I wonder if makes sense to have R2ObjectBody implement Body for easier interoperability. fyi @vlovich

content/r2/examples/cache-api.md
mpint: Agreed on consistency 👍

mpint (approved):

PR #4858, opened 2022-06-19
[R2] Add presigned links example to aws-sdk-js

KianNH:


KianNH:
Added `signatureVersion: 'v4'` into the S3 client initialization as otherwise the presigned links default to v2.

deadlypants1973 (approved):

PR #4861, opened 2022-06-20
[Doc site]: Add tab switcher shortcode

Ekwuno:
Since Workers introduced module syntax, we need a way to show the difference between service worker syntax and module syntax in our docs. 

This PR, adds a shortcode that can be used within our current docs setup as we would an `Aside` for example.  

You can test it at https://add-tab-switcher-shortcode-c.cloudflare-docs-7ou.pages.dev/pages/tutorials/forms/

Todo:
- [x]  Fix responsiveness


kodster28:
From a bit of testing, looks like only the final example (`TS (SW)`) allows you to use the scroll down functionality to see the entire snippet.

Ekwuno:
> From a bit of testing, looks like only the final example (`TS (SW)`) allows you to use the scroll down functionality to see the entire snippet.

Yea! Ignore this for now. It's a draft because I have to fix that bug. I have actually, but on a different branch.

kodster28:
One question.. can't quite grok it just from looking at the code.

Is the tab component generic enough to be used throughout the docs site? i.e., like for helping us group [examples like this](https://developers.cloudflare.com/load-balancing/how-to/create-load-balancer/#create-a-load-balancer) on this page into a two-tabbed `Dashboard` / `API` structure?

Ekwuno:
> examples like this

Yes, it is! You will write the Shortcode just as you would an Aside

Ekwuno:
@kodster28 you can test it now

static/style.css
```diff
@@ -4427,9 +4946,139 @@ blockquote .DocsMarkdown--header-anchor-positioner {
...
+  text-align: center;
+  color: currentColor;
+  cursor: pointer;
+  top: 0;
+  font-weight: 700;
+  font-size: var(--code-font-size);
+  font-family: var(--monospace-font-family);
+}
+
+.tab {
```
lauragift21, line 5052: Looks good! Just a few changes to fix the positioning of the codeblock. 

```suggestion
.tab {
  max-height: 270px;
  display: none;
  color: #2c3e50;
  border-bottom: 0.25rem solid #bdc3c7;
  overflow: scroll;
}
```
Screenshot 2022-06-30 at 15 13 06

Ekwuno: Good catch with the padding! Thanks, Gift.

jgentes:
Looks like there's an odd hover effect that goes away if `width: 25%` is removed from `.tab-label-wrapper`:
image

assets/events.ts
```diff
@@ -96,6 +96,36 @@ export function focus() {
...
+export function tabs() {
+  // Find all tab wrappers
+  let wrappers = document.querySelectorAll(".wrapper");
+
+  for (let i = 0; i < wrappers.length; i++) {
+    const labels = wrappers[i].querySelectorAll(".tab-label");
+    const tabs = wrappers[i].querySelectorAll(".tab");
+
+    // Set the first tab in a group to display
+    ((tabs[0] as HTMLElement) || null).style.display = "block";
```
kristianfreeman: This looks like it fails for me (Safari) about 50% of the time when I refresh the page, so that no active tab shows up.

Screen Shot 2022-06-30 at 3 16 27 PM

content/pages/tutorials/forms/index.md
```diff
@@ -195,6 +265,143 @@ The HTML page is also completely unstyled at this point, relying on the browsers
 
 {{</Aside>}}
 
+{{< tabs tabTotal="3" tabRightAlign="2">}}
```
kristianfreeman: what does tabRightAlign do?

kristianfreeman (commented):
quick bug I found. also, is there any sort of active tab CSS styling?

assets/events.ts
Ekwuno: Hmm, now the browser wars. Let me look into how to fix that for safari. I tested on Mozilla and chrome.  There is an active tab styling.

content/pages/tutorials/forms/index.md
Ekwuno: This was supposed to be some logic to help with moving a tab to the right but I don't need it to do that anymore so We can remove that.

assets/events.ts
Ekwuno: I am not able to recreate this. I launched safari for the first time ever! and It works as expected.

Screenshot 2022-06-30 at 15 28 58

Could this be some caching problem? I will look into it

content/pages/tutorials/forms/index.md
Ekwuno: You can ignore this for now. When we migrate code examples the shortcode will be removed from this page and the format will be gone.

Ekwuno: I have removed this code block.

Cherry (commented):
Super excited for this to land! 🎉 There's a few things I noticed that could improve this - many are nitpicks, sorry!

- There should probably be something that explains what `SW` and `ESM` actually mean in this context? Perhaps that's out of scope of this PR, but a page explaining the difference would be critical I feel for new users, who may not know which one they should choose, or why

- Sometimes, it appears to error out and doesn't render anything. I can replicate this with Chrome on Windows, by hitting CTRL + F5 about 10 times. Maybe 5/10 times it won't render and will look like this: 
	- When this occurs, I get the following JS error:
	
	  ```
		  main.min.808f9f11b83fe4a54994f62d560a56c7.js:1 Uncaught TypeError: Cannot read properties of null (reading 'style')
	      at A (main.min.808f9f11b83fe4a54994f62d560a56c7.js:1:1822)
	      at main.min.808f9f11b83fe4a54994f62d560a56c7.js:1:4740
	      at main.min.808f9f11b83fe4a54994f62d560a56c7.js:1:4753
	  ```
	  Which when formatted is this block of code:
	  ```js
	  function A() {
		  let a = document.querySelectorAll(".wrapper");
		  for (let b = 0; b < a.length; b++) {
			  let c = a[b].querySelectorAll(".tab-label")
			    , d = a[b].querySelectorAll(".tab");
			  (d[0] || null).style.display = "block";
			  for (let a = 0; a < c.length; a++)
				  c[a].addEventListener("click", D)
		  }
	  }
	  ```
	And appears to be here: https://github.com/cloudflare/cloudflare-docs/pull/4861/files#diff-6583f70c9cf0e754cdb0e9574aa14f2b76bd476cf76fa957cc043327578f3805R114 - is it possible this is somehow running too early?

- nitpick: When I load the page with JS disabled, none of the code is shown at all: .

- There appears to be a horizontal scrollbar on the code block always, even when its not needed: 
	Only showing this when there's scrollable horizontal content would be more ideal I think.

- nitpick: in the ESM version on the example page, there's a large `margin-bottom: 2em` which feels out of place:  There's also an unnecessary vertical scrollbar here, like the horizontal one mentioned previously.

- feature req: It would be very cool if switching the code type added a URL query param or hash fragment, so specific code blocks can easily be shared when linking to docs pages.

Ekwuno:
> Super excited for this to land! 🎉 There's a few things I noticed that could improve this - many are nitpicks, sorry!
> 
> 
> 
> - There should probably be something that explains what `SW` and `ESM` actually mean in this context? Perhaps that's out of scope of this PR, but a page explaining the difference would be critical I feel for new users, who may not know which one they should choose, or why
> 
> 
> 
> - Sometimes, it appears to error out and doesn't render anything. I can replicate this with Chrome on Windows, by hitting CTRL + F5 about 10 times. Maybe 5/10 times it won't render and will look like this: 
> 
> 	- When this occurs, I get the following JS error:
> 
> 	
> 
> 	  ```
> 
> 		  main.min.808f9f11b83fe4a54994f62d560a56c7.js:1 Uncaught TypeError: Cannot read properties of null (reading 'style')
> 
> 	      at A (main.min.808f9f11b83fe4a54994f62d560a56c7.js:1:1822)
> 
> 	      at main.min.808f9f11b83fe4a54994f62d560a56c7.js:1:4740
> 
> 	      at main.min.808f9f11b83fe4a54994f62d560a56c7.js:1:4753
> 
> 	  ```
> 
> 	  Which when formatted is this block of code:
> 
> 	  ```js
> 
> 	  function A() {
> 
> 		  let a = document.querySelectorAll(".wrapper");
> 
> 		  for (let b = 0; b < a.length; b++) {
> 
> 			  let c = a[b].querySelectorAll(".tab-label")
> 
> 			    , d = a[b].querySelectorAll(".tab");
> 
> 			  (d[0] || null).style.display = "block";
> 
> 			  for (let a = 0; a < c.length; a++)
> 
> 				  c[a].addEventListener("click", D)
> 
> 		  }
> 
> 	  }
> 
> 	  ```
> 
> 	And appears to be here: https://github.com/cloudflare/cloudflare-docs/pull/4861/files#diff-6583f70c9cf0e754cdb0e9574aa14f2b76bd476cf76fa957cc043327578f3805R114 - is it possible this is somehow running too early?
> 
> 
> 
> - nitpick: When I load the page with JS disabled, none of the code is shown at all: .
> 
> 
> 
> - There appears to be a horizontal scrollbar on the code block always, even when its not needed: 
> 
> 	Only showing this when there's scrollable horizontal content would be more ideal I think.
> 
> 
> 
> - nitpick: in the ESM version on the example page, there's a large `margin-bottom: 2em` which feels out of place:  There's also an unnecessary vertical scrollbar here, like the horizontal one mentioned previously.
> 
> 
> 
> - feature req: It would be very cool if switching the code type added a URL query param or hash fragment, so specific code blocks can easily be shared when linking to docs pages.
> 
> 

Not nitpicking at all I love all these! Thanks for taking the time to highlight them. I'll get to fixing. @codewithkristian also brought up the re rendering issue. I couldn't reproduce it but I'll look into it some more. 

I think with regards to `SW` and `ESM` do you think it'd be better used svgs? 

Eg. (JavaScript svg) Service worker || (JavaScript svg>)ES modules ? 

I think that before the tab switcher is used there will be a paragraph saying something like: "below we have two formats" or something like that.

Cherry:
> I think with regards to SW and ESM do you think it'd be better used svgs?

I think the text `SW` and `ESM` are totally great, as long as it's established somewhere beforehand what these acronyms are and mean (perhaps somewhere else on the docs even, with a "what is this?" button next to codeblocks?). I believe the docs mention ESM in one or two places today, but I don't think that's true for `SW`. 

When linking people to docs, these code blocks might be the first time they're introduced to Workers and these acronyms, especially in the case of tutorials - being given 4 code blocks for a single tutorial and not knowing which one is relevant to me, the "best", or why I should use them for example, could be quite overwhelming for a new user.

Ekwuno:
> > I think with regards to SW and ESM do you think it'd be better used svgs?
> 
> 
> 
> I think the text `SW` and `ESM` are totally great, as long as it's established somewhere beforehand what these acronyms are and mean (perhaps somewhere else on the docs even, with a "what is this?" button next to codeblocks?). I believe the docs mention ESM in one or two places today, but I don't that's true for `SW`. 
> 
> 
> 
> When linking people to docs, these code blocks might be the first time they're introduced to Workers and these acronyms, especially in the case of tutorials - being given 4 code blocks for a single tutorial and not knowing which one is relevant to me, the "best", or why I should use them for example, could be quite overwhelming for a new user.

Absolutely this was thinking as well! I'm all for text to be honest. I think it's more explicit and there will always be some intro before the tab component. 

I love the "what is this" idea and I see a "?" That when hovered on can give more information. That would be brilliant

assets/events.ts
Princeobinna: Approve

Ekwuno: @codewithkristian I have put up some changes. Can you check if this behaviour presets?

Cherry (commented):
Re-testing as requested by @Ekwuno! 🧡

I can't recreate the loading issue anymore, so that's great to see! It seems to load successfully every time now. The overflow scrolling only appearing when necessary is great too.

A few more pieces of feedback:

- There's a layout shift as the first tab is set to `display: block`. This can cause quite a frustrating UX, especially on slower connections, as the content jumps around when you're reading. Here's a video of this with a throttled connection to better illustrate the point: https://up.jross.me/pf9opbpr
	- Perhaps setting the first tab to `display: block` server-side would make the most sense here, and it also resolve my previous comment about the code not being visible when JS is disabled or doesn't load. Ideally, all code blocks would be visible even if JS didn't load or was disabled.

- nitpick: The grey line at the bottom feels a little unnecessary to me, especially in scenarios like this: https://up.jross.me/mc0zn1xl. This is emphasized by the margin at the bottom of the codeblock.

- nitpick: The style in the top right of the tabs looks a little strange, especially in light mode where the lack of the bottom border is apparent: https://up.jross.me/22tde2nv

- nitpick: On the mobile breakpoint, there appears to be somewhat duplicate hover styles, resulting in a double orange box: https://up.jross.me/4td6hhkh

- nitpick: Perhaps on mobile it make more sense to use `white-space: pre-wrap;` rather than requiring the user to scroll both vertically and horizontally to read the code? https://up.jross.me/4wk9p8p3

---

This is looking awesome though! I definitely think some kind of description for what these tabs are/mean is very much required. Some kind of way to link to a specific code block and/or save your selection so if you always wanted to only view ESM on first page load, too would be awesome.

content/pages/tutorials/forms/index.md
```diff
@@ -16,6 +16,76 @@ This tutorial will briefly touch upon the basics of HTML forms. For a more in-de
 
 {{</Aside>}}
 
+{{<tabs labels="js/sw | js/esm | ts/esm | ts/sw">}}
+{{<tab label="js/sw">}}
```
kristianfreeman: Perhaps we can add a parameter here called `default` or `initial` that we can use to render an initial code sample (and fix the "nothing renders with JS turned off" problem)

layouts/shortcodes/tab.html
```diff
@@ -0,0 +1,3 @@
+{{- $label := replace (.Get "label") "/" "-"}}
+
+<div class="tab" id="tab-{{$label}}">{{ $.Inner | markdownify }}</div>
```
kristianfreeman, line 8: Could check here for the presence of the aforementioned `default`/`initial` and add `style="display: none;"` if present. After render and _with_ JS enabled, the code in `events.ts` will manage the `style` attribute after any additional changes, so this should be a safe change.

kristianfreeman (commented):
Few things to note:

**Rounded code blocks clash with tab headers:**

Screen Shot 2022-07-11 at 1 01 55 PM
Screen Shot 2022-07-11 at 1 01 57 PM

**Extra padding underneath tab-wrapper**

Screen Shot 2022-07-11 at 1 02 07 PM

**No active state for currently selected tab**

Screen Shot 2022-07-11 at 1 02 16 PM

@Cherry's comment about this being blank without JS is a problem we should fix too.

assets/events.ts
```diff
@@ -76,32 +83,84 @@ function $copy(ev: MouseEvent) {
...
+        console.log({ wrappers, tabs, labels });
+      }
+    }
+  });
+}
+
+export function activeTab() {
+  var header = document.getElementById("tab-active");
+  var tabs = header.getElementsByClassName("tab-label");
+  console.log(tabs)
```
kristianfreeman: remove

assets/events.ts
```diff
@@ -76,32 +83,84 @@ function $copy(ev: MouseEvent) {
...
+      // const defaultTab = wrappers[i].querySelector(".tab-default");
+
+      if (tabs.length > 0) {
+        // Set the first tab in a group to display
+        (tabs[0] as HTMLElement).style.display = "block";
+        // (defaultTab as HTMLElement).style.display = "none";
+        for (let i = 0; i < labels.length; i++)
+          labels[i].addEventListener("click", $tab);
+      } else {
+        console.log({ wrappers, tabs, labels });
```
kristianfreeman: remove

kristianfreeman (commented):
nice, this is looking a lot better!

two things i noticed

1) formatting of the tab headers is a little messed up on low-width/mobile displays

Screen Shot 2022-07-21 at 12 45 31 PM

2) there appears to be a border or something at the bottom - i thought it was a scrollbar at first but i don't think it is

Screen Shot 2022-07-21 at 12 45 58 PM

kristianfreeman (commented):
1) Extra spacing at the end of the header row - should have some CSS or something to target the last tab selector and make it full-width to fill the rest of the space.

2) Order is weird - should probably be JS SW/ESM, then TS SW/ESM.

3) Can we add some more spacing between the icons and language label?

Screen Shot 2022-07-25 at 8 49 00 AM

Cherry (commented):
Would it be possible to get a `title` on the `JS/TS` SVGs too? Otherwise to a screen-reader, each pair of 2 tabs look identical:

Ekwuno:
> 1. Extra spacing at the end of the header row - should have some CSS or something to target the last tab selector and make it full-width to fill the rest of the space.
> 2. Order is weird - should probably be JS SW/ESM, then TS SW/ESM.
> 3. Can we add some more spacing between the icons and language label?
> 
> Screen Shot 2022-07-25 at 8 49 00 AM

1. Are you viewing this on safari? The extra spacing is added by the browser. It doesn't happen on chrome or firefox; I am a bit confused about how to address it. 
Screenshot 2022-07-25 at 15 26 37

2. The order is based on the shortcode used for this test.  In useage, if you called JS SW/ESM, then TS SW/ESM it will follow that order. I could add some `order` to the CSS, but I don’t want that to be hard coded order so that in usage, if it doesn’t follow that order screenreaders aren’t confused.  Something like this will solve it. Let me know your thoughts?

```
ul {
  display: flex;
  flex-direction: column;
}
ul li:first-child {
  order: 1;
}
ul li:nth-child(2) {
  order: 2;
}
ul li:nth-child(3) {
  order: 3;
}
ul li:nth-child(4) {
  order: 4;
}
```

3. More space between the icons makes the text roll over, which looks a bit weird to me. I tried to make the text eligible and the icon noticeable, so the text right now is in a good spot IMO.

Ekwuno:
> Would it be possible to get a `title` on the `JS/TS` SVGs too? Otherwise to a screen-reader, each pair of 2 tabs look identical: 

Thanks for pointing this out @Cherry 🙌🏾

kristianfreeman:
> > 1. Extra spacing at the end of the header row - should have some CSS or something to target the last tab selector and make it full-width to fill the rest of the space.
> > 2. Order is weird - should probably be JS SW/ESM, then TS SW/ESM.
> > 3. Can we add some more spacing between the icons and language label?
> > 
> > Screen Shot 2022-07-25 at 8 49 00 AM
> 
> 1. Are you viewing this on safari? The extra spacing is added by the browser. It doesn't happen on chrome or firefox; I am a bit confused about how to address it.
>    Screenshot 2022-07-25 at 15 26 37
> 2. The order is based on the shortcode used for this test.  In useage, if you called JS SW/ESM, then TS SW/ESM it will follow that order. I could add some `order` to the CSS, but I don’t want that to be hard coded order so that in usage, if it doesn’t follow that order screenreaders aren’t confused.  Something like this will solve it. Let me know your thoughts?
> 
> ```
> ul {
>   display: flex;
>   flex-direction: column;
> }
> ul li:first-child {
>   order: 1;
> }
> ul li:nth-child(2) {
>   order: 2;
> }
> ul li:nth-child(3) {
>   order: 3;
> }
> ul li:nth-child(4) {
>   order: 4;
> }
> ```
> 
>  3. More space between the icons makes the text roll over, which looks a bit weird to me. I tried to make the text eligible and the icon noticeable, so the text right now is in a good spot IMO.

#1/#3 - that's fine with me, disregard. For #2 if it's based on how we actually write the syntax that's great, I don't think we should add a hardcoded order. Can we fix the order here just so it merges in correctly? It's confusing as written right now.

+1 from me once we land that ^ change

Ekwuno:
Oh, I see what you mean!  So I added the shortcode to the HTML form tutorial so that we see how it is displayed. After all the reviews, I will delete it from that tutorial because it's not needed there. 

`{{<tabs labels="js/sw | js/esm | ts/sw | ts/esm">}}` manages the order so that's sorted.

kristianfreeman:
sounds good. can we remove it then and i'll +1

kristianfreeman (approved):

WalshyDev:
Some future feature requests (Not needed for MVP but worth noting):

1. Add selection into query string so we can share pages in TS+ESM for example
2. Save your selection into localStorage and switch to that tab when you visit the page (so you will always see TS+ESM for example)

layouts/shortcodes/tabs.html
```diff
@@ -0,0 +1,61 @@
...
+<div class="tabs-wrapper">
+  <nav role="navigation" class="tabs">
+    <ul id="tab-active">
+      {{ range $idx, $txt := $labels }} {{- $link := replace $txt "/" "-"}}
+      <li class="tab-label-wrapper " >
+        {{ if or (eq $link "js-sw") }}
+        <a class="tab-label active" href="#" data-link="tab-{{$link}}">
+          <svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 512 512" height="1em" width="1em"
+            xmlns="http://www.w3.org/2000/svg" data-link="tab-{{$link}}">
+            <title>JavaScript Service Workers </title>
```
Cherry:
```suggestion
            <title>JavaScript</title>
```

layouts/shortcodes/tabs.html
```diff
@@ -0,0 +1,61 @@
...
+              d="M32 32v448h448V32zm240 348c0 43.61-25.76 64.87-63.05 64.87-33.68 0-53.23-17.44-63.15-38.49l34.28-20.75c6.61 11.73 11.63 21.65 26.06 21.65 12 0 21.86-5.41 21.86-26.46V240h44zm99.35 63.87c-39.09 0-64.35-17.64-76.68-42L329 382c9 14.74 20.75 24.56 41.5 24.56 17.44 0 27.57-7.72 27.57-19.75 0-14.43-10.43-19.54-29.68-28l-10.52-4.52c-30.38-12.92-50.52-29.16-50.52-63.45 0-31.57 24.05-54.63 61.64-54.63 26.77 0 46 8.32 59.85 32.68L396 290c-7.22-12.93-15-18-27.06-18-12.33 0-20.15 7.82-20.15 18 0 12.63 7.82 17.74 25.86 25.56l10.52 4.51c35.79 15.34 55.94 31 55.94 66.16.01 37.9-29.76 57.64-69.76 57.64z" data-link="tab-{{$link}}">
+            </path>
+          </svg> Service Workers</a
+        >
+
+        {{ else if (eq $link "ts-sw") }}
+        <a class="tab-label" href="#" data-link="tab-{{$link}}">
+          <svg stroke="currentColor" fill="currentColor" stroke-width="0" role="img" viewBox="0 0 24 24" height="1em" width="1em"
+            xmlns="http://www.w3.org/2000/svg" data-link="tab-{{$link}}">
+            <title>TypeScript Service Workers</title>
```
Cherry:
```suggestion
            <title>TypeScript</title>
```

layouts/shortcodes/tabs.html
```diff
@@ -0,0 +1,61 @@
...
+            </path>
+            
+          </svg>Service Workers</a
+        >
+
+        {{ end }} {{ if or (eq $link "js-esm") }}
+        <a class="tab-label " href="#" data-link="tab-{{$link}}">
+          <svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 512 512" height="1em" width="1em"
+            xmlns="http://www.w3.org/2000/svg" data-link="tab-{{$link}}">
+            <title>JavaScript ES Modules</title>
```
Cherry:
```suggestion
            <title>JavaScript</title>
```

layouts/shortcodes/tabs.html
```diff
@@ -0,0 +1,61 @@
...
+            </path>
+            
+          </svg> ES Modules</a
+        >
+
+        {{ else if (eq $link "ts-esm") }}
+        <a class="tab-label" href="#" data-link="tab-{{$link}}">
+          <svg stroke="currentColor" fill="currentColor" stroke-width="0" role="img" viewBox="0 0 24 24" height="1em" width="1em"
+            xmlns="http://www.w3.org/2000/svg" data-link="tab-{{$link}}">
+          <title>TypeScript ES Modules</title>
```
Cherry:
```suggestion
          <title>TypeScript</title>
```

Cherry (commented):
> Thanks for pointing this out @Cherry 🙌🏾

Sorry for being overly picky, but now it's resulting in duplicate text for the tabs to a screen reader:

Ekwuno:
> Some future feature requests (Not needed for MVP but worth noting):
> 
> 1. Add selection into query string so we can share pages in TS+ESM for example
> 2. Save your selection into localStorage and switch to that tab when you visit the page (so you will always see TS+ESM for example)

We have a follow-up ticket for 2.

PR #4911, opened 2022-06-23
Update to signing-requests.md

AlexisBallo2:
Added catch for when signing algorithm generates a token with a "+". 

("+" will break authentication)


content/workers/examples/signing-requests.md
```diff
@@ -143,7 +143,9 @@ async function generateSignedUrl(url) {
   // `mac` is an ArrayBuffer, so you need to make a few changes to get
   // it into a ByteString, and then a Base64-encoded string.
   const base64Mac = btoa(String.fromCharCode(...new Uint8Array(mac)));
-
+  
+  //must convert "+" to "-" as urls encode "+" as " "
```
kristianfreeman:
```suggestion
  // must convert "+" to "-" as urls encode "+" as " " 
```

deadlypants1973 (approved):

PR #4923, opened 2022-06-24
Clear up file limit

WalshyDev:
This will need to be updated again once dash supports 20k but I want this updated now rather than waiting.


Ekwuno:
Sounds good!

Ekwuno (approved):

Hexstream:
In the nick of time!!

I was about to bump up against (and probably-painfully work around) that limit while migrating the last few of my 25+ subdomains to Cloudflare Pages!

Thank you!!

fariborz4:
Ok

PR #4928, opened 2022-06-25
Last reviewed prototype

GregBrimble:
Adds a "last reviewed" date to the footer, pulling from a page's "last-reviewed" front matter attribute.

Example page: https://last-reviewed-prototype.cloudflare-docs-7ou.pages.dev/pages/platform/headers/

### DIFF
--- a/content/pages/platform/headers.md
+++ b/content/pages/platform/headers.md
@@ -1,6 +1,7 @@
 ---
 pcx-content-type: concept
 title: Headers
+last-reviewed: "2022-05-01"
 ---
 
 # Headers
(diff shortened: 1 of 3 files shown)


KimJ15 (commented):
I can't speak to the code, but this update is great and lgtm!

pedrosousa:
I noticed that there's already an `updated` frontmatter attribute used in Workers tutorials (for example, here: https://github.com/cloudflare/cloudflare-docs/blob/production/content/workers/tutorials/workers-kv-from-rust/index.md ).
It may be confusing to have both an `updated` and a `last-reviewed` attribute.

Would it make sense to use the same attribute in both situations?

kodster28 (approved):
Not sure I can totally speak for the Workers team, but it seems like semantically they're trying to create a date more tied to updates than reviewed... so not sure if they'd be the same value.

Ideally, if we fix the date w/in the template, we could just use that value... but that seems outside the scope of this PR.

kodster28:
Feel free to re-open, but closing this out now as it's been a while and conflicting with recent changes.

PR #4930, opened 2022-06-26
Fix bash command example to include multiport

bobbyjoe30:


bobbyjoe30:
Looks like this was resolved with this pull request 
https://github.com/cloudflare/cloudflare-docs/pull/6651

PR #4944, opened 2022-06-27
Transform Rules troublshooting - link to evaluation explanation

simon-says:
I found that the fundamental information about Transform Rule evaluation might be missed on the summary page, especially when troubleshooting - so I've added a link to it from troubleshooting.


pedrosousa (approved):

pedrosousa:
Thanks, @simon-says!
I've updated the text a bit.

PR #4960, opened 2022-06-28
[R2] Amend snippet for truncated list responses

KianNH:
Amends the snippet to give a 'here's what you should do' as opposed to just 'here's what you shouldn't do' - whilst the existing snippet was there just to show you what condition (`listed.length < limit`) you *shouldn't* do, `length` doesn't exist on `R2Objects` (assuming `listed` was the result of `list()`, if it wasn't then it would be when overwritten in the block) and the operation wasn't awaited either.

Also prefixed the binding with `env.` to match the other examples in the R2 docs which now use Module Workers instead of Service Workers.

Whilst those might be obvious to someone experienced with JavaScript, a beginner who copy/pastes it would be confused for a long time.

### DIFF
--- a/content/r2/runtime-apis.md
+++ b/content/r2/runtime-apis.md
@@ -255,15 +255,19 @@ There are 3 variations of arguments that can be used in a range:
   let truncated = listed.truncated;
   let cursor = truncated ? listed.cursor : undefined;
 
-  // ❌ - if your limit can't fit into a single response or your bucket
-  // has less objects than the limit, it will be stuck here permanently
+  // ❌ - if your limit can't fit into a single response or your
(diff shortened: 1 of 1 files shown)


KianNH:
Reformatted it a bit so that it fits inside the codeblocks without needing to scroll left/right

KianNH:
It seems to be in `workers-types` too - guess it's from the runtime?

https://github.com/cloudflare/workers-types/blob/f661c288e64198fe7744a7118a61ada1320fd0b8/index.d.ts#L1089

Unrelated to this PR - but interesting

vlovich (approved):

deadlypants1973 (approved):

PR #4962, opened 2022-06-28
[SSL for SaaS] Added note about hostname churning

mmalden:


kodster28 (approved):

isaac-mcfadyen:
Small thing with this PR: me (as a user inexperienced with SaaS) would not know what `churn` means had I not looked it up. We might want to clarify this a bit. Maybe something like:

```
When an end customer "churns" (deletes their account or removes themself from your service)...
```

PR #4963, opened 2022-06-28
[Pages] do not expose what is invalid

mrtrkmn:


WalshyDev (approved):

mrtrkmn:
Also why throwing exception ? instead of re-asking credentials when username or password is invalid.

deadlypants1973 (approved):

PR #4983, opened 2022-06-29
Fix typo around mdm.xml

joel0:


kodster28 (approved):

kodster28:
Thanks for the catch @joel0

PR #5004, opened 2022-07-01
Update firewall_events.md

123123123-thinh:


kodster28:
This change wouldn't align with the other fields on the page, so going to close it out.

PR #5006, opened 2022-07-01
[Workers] add module syntax to overview page

deadlypants1973:


content/workers/_index.md
```diff
@@ -40,21 +40,35 @@ or install with `yarn`:
...
+Read more about [installing Wrangler](/workers/wrangler/get-started/).
 
 ---
 
 ## Playground
 
 View this Hello World example in the Workers playground:
 
-```javascript
+```js
```
pedrosousa, line 51: Do we know when a user will get one or the other? I tried it and always got the Service Worker syntax.
Also, the Playground-specific page (`/workers/learning/playground/`) uses only the Service Worker syntax.

Would it make sense to link to a page explaining the difference between the two, or would it be too advanced at this point?

pedrosousa (approved):

PR #5021, opened 2022-07-04
[Waiting Room] Remove API-only queueing methods header

angelampcosta:


content/_redirects
```diff
@@ -558,6 +558,7 @@
 /waiting-room/how-to/mobile-traffic/ /waiting-room/additional-options/mobile-traffic/ 301
 /waiting-room/how-to/test-waiting-room/ /waiting-room/additional-options/test-waiting-room/ 301
 /waiting-room/how-to/use-waiting-room-api/ /waiting-room/reference/waiting-room-api/ 301
+/waiting-room/reference/queueing-methods/#api-only-queueing-methods /waiting-room/reference/queueing-methods/ 301
```
pedrosousa: We cannot redirect single sections... You'd have to redirect the entire page, removing `#api-only-queueing-methods`. Since the source and target URLs would then become the same, I would not add any redirect, since it would create a circular reference and possibly cause issues.

pedrosousa (approved):

PR #5044, opened 2022-07-06
Update content on using local env variables with functions

lauragift21:
There's a typo in this example used for illustrating how to use the local `.dev.vars` file in Functions.

It should be defined as the ENV_NAME = "Value" and not with a Const assigned to the environment variable. 

Tested this locally and saw this was a typo.


Ekwuno (approved):
Thanks Gift!

PR #5068, opened 2022-07-07
document DO timeout

koeninger:
Even if we end up changing this, wanted to get it documented in the meantime, happy to hear arguments to the contrary

### DIFF
--- a/content/workers/learning/using-durable-objects.md
+++ b/content/workers/learning/using-durable-objects.md
@@ -555,4 +555,4 @@ To solve this you can either do less work per request, or send fewer requests, f
 
 #### Error: Durable Object storage operation exceeded timeout which caused object to be reset.
 
-To prevent indefinite locking, there is a limit on how much time storage operations can take. For operations such as `deleteAll()` that make partial progress, it is safe to retry until success. Otherwise contact support.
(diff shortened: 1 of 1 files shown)


content/workers/learning/using-durable-objects.md
```diff
@@ -552,3 +552,7 @@ A single instance of a Durable Object cannot do more work than is possible on a
 - `Error: Durable Object is overloaded. Requests queued for too long.` The oldest request has been in the queue too long.
 
 To solve this you can either do less work per request, or send fewer requests, for example, by splitting the requests among more instances of the Durable Object.
+
+#### Error: Durable Object storage operation exceeded timeout which caused object to be reset.
+
+To prevent indefinite locking, there is a limit on how much time storage operations can take. For operations such as `deleteAll()` that make partial progress, it is safe to retry until success. Otherwise contact support.
```
a-robinson: This only affects `deleteAll()`, doesn't it? All other storage operations are atomic.

Given that, I'd personally be more specific here. Something like:

```
To prevent indefinite locking, there is a limit on how much time storage operations can take. In objects containing a sufficiently large number of key-value pairs, `deleteAll()` may hit that time limit and fail. When this happens, note that each `deleteAll()` call does make progress and that it is safe to retry until it succeeds. Otherwise contact support.
```

a-robinson (approved):

a-robinson (approved):
👍

PR #5072, opened 2022-07-07
prevent "dpkg: error: requested operation requires superuser privileg…

spidy0x0:
…e" pop up on systems that are not automatically in root


ranbel (approved):

kodster28:
Howdy @spidy0x0, apologies on the delay here.

Merged in the change in #9513 (easier to update a new PR than this one).

PR #5107, opened 2022-07-11
Use a note for plan requirement

cicku:


kodster28:
Hey! Thanks for the submissions @cicku, but we're trying to standardize a pattern across the docs (and the H2 is more visible for folks using internal/external search).

PR #5115, opened 2022-07-11
Note about split tunnel rules and device posture

kkrum:
Add a callout explaining that device posture eval only works if the related app/authentication traffic is flowing inside the tunnel.


ranbel:
refer to #5133

PR #5138, opened 2022-07-12
[SSL] Correct the custom cert name

cicku:
In dashboard it is `Custom Modern` by default, let's keep the consistency here.


kodster28 (approved):

kodster28:
good catch, @cicku! Thanks!

PR #5141, opened 2022-07-13
early_hints should be in SSL settings instead

gijo-varghese:
`early_hints` should be in SSL settings instead of custom metadata. 
API doc: https://api.cloudflare.com/#custom-hostname-for-a-zone-create-custom-hostname


gijo-varghese:
Screenshot 2022-07-13 at 8 14 16 AM

PR #5154, opened 2022-07-14
[ZT] Use ZT dashboard directly

cicku:
Access from dashboard is deprecated


kodster28:
Howdy @cicku, thanks for the submission, but going to have to close this out as inaccurate.

Access policies w/in the core dash have been deprecated, but that doesn't mean that you can't still access ZT through the Core dash. The link goes to https://developers.cloudflare.com/cloudflare-one/setup/#start-from-the-cloudflare-dashboard, which provides more instructions on location.

PR #5185, opened 2022-07-18
Remove note about a fetch handler being required alongside DOs

a-robinson:
Because we're removing that requirement on script upload. It's still
often a good idea to bundle a fetch handler with a DO class, but will no
longer be required.


koeninger (approved):

a-robinson:
I've confirmed this behavior is live.

PR #5195, opened 2022-07-19
[Ruleset-Engine] fix http.request.version field type from Number -> String

tc80:
Minor Bug:
- `http.request.version` is a String and not a Number
- It is clear by the example values, but I also confirmed internally :)


pedrosousa:
Thanks, @tc80! I'll go ahead and merge your update.

pedrosousa (approved):

PR #5217, opened 2022-07-20
[WAF] Clarify known limitation for Firewall Analytics

lougayou:
[WAF] Clarify known limitation for Firewall Analytics: customers are wondering why they can't find events when filtering for specific ray id. (PSPEC-1315)


pedrosousa:
FYI: I removed the internal link, as we usually do for other tickets.

pedrosousa (approved):

PR #5270, opened 2022-07-25
Update false-negative.md

zeinjaber:
Addresses PSPEC-1319.


pedrosousa:
Removed the internal URL from the PR description, like we usually do.

pedrosousa (approved):

PR #5274, opened 2022-07-25
[Zero Trust] Fix typo in supported proxy modes

schultetwin1:
Previously the docs for the Zero Trust desktop client mentioned that our
proxy supports either SOCKS or HTTPS proxies. The HTTPS proxy is not
correct. We support HTTP only.


Noah-Kennedy (commented):
We should probably explain why we do HTTP rather than HTTPS here.

Noah-Kennedy:
HTTP vs HTTPS makes no difference to the threat model, since we would basically have to bundle a private key onto the machine to get a local HTTPS proxy. We should explain this here.

PR #5280, opened 2022-07-25
[Docs Site] Improved footnote experience

kodster28:
Followed instructions in https://discourse.gohugo.io/t/littlefoot-for-fancy-popup-footnotes/23991 to make footnotes more readable + easier to access in context.

### DIFF
--- a/content/load-balancing/understand-basics/health-details.md
+++ b/content/load-balancing/understand-basics/health-details.md
@@ -14,11 +14,7 @@ But how does your load balancer _know_ which servers can handle the traffic? We 
 
 ## Components
 
-Dynamic load balancing happens through a combination of:
-
-- [**Origin pools**](/load-balancing/understand-basics/pools/): Contain one or more origin servers.
-- [**Monitors**](/load-balancing/understand-basics/monitors/): Are attached to individual origin servers and issue health checks at regular intervals.
-- **Health checks**: Are issued by a monitor at regular interval and — depending on the monitor settings — return a **pass** or **fail** value to make sure an origin is still able to receive traffic.
+Dynamic load balancing happens through a combination of [origin pools](/load-balancing/understand-basics/pools/)[^1], [monitors](/load-balancing/understand-basics/monitors/)[^2], and health checks[^3]. 
 
 ![Dynamic load balancing involves pools, origins, monitors, and health checks](/load-balancing/static/images/load-balancer-components.png)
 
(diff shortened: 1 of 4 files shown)


kodster28:
Open question... suggestion to download and host locally in [the thread](https://discourse.gohugo.io/t/littlefoot-for-fancy-popup-footnotes/23991) suggesting it.

What exactly does that mean?

KianNH:
> Open question... suggestion to download and host locally in [the thread](https://discourse.gohugo.io/t/littlefoot-for-fancy-popup-footnotes/23991) suggesting it.
> 
> What exactly does that mean?



I think they mean that `https://unpkg.com` is sometimes a bottleneck - so the rest of your site has loaded but you're waiting on the library to download from their CDN, causing layout shift ( or whatever it does ).

It's [jsDelivr](https://www.jsdelivr.com/package/npm/littlefoot) as an alternative.

kodster28:
Thanks for the context, @KianNH! 

From looking at Lighthouse testing comparisons, it _seems_ like there isn't any bottleneck. Once this goes live though, I'll keep an eye out and be ready to test other solutions as needed.

marciocloudflare (approved):

kodster28:
Just closing the loop here, @KianNH, had to adjust the script slightly in #5338 , but it looks like it's good now

KianNH:
The joys of webperf 🫠

When are we getting Early Hints too? :^)

kodster28:
Oooh, good suggestion. Getting the Pages team to flip the switch for us (they're working on integrating them into Pages by default).

KianNH:
I’m excited for it! I released a Worker to add them with HTMLRewriter and Obinna made a Pages plug-in from it - now we just need @WalshyDev to make it native ;)

WalshyDev:
Enabled for docs and coming soon to a Pages site near you :^)

KianNH:
What's the go-to for handling files with hashes in their name?

i.e the docs site has no `link` headers right now so there'll be no 103 responses but the JS uses a hash in the filename for cache busting

Can you add those to the `_headers` file post-build command and Pages will be fine with that? Guess we'd also want the `littlefoot` package in there too

WalshyDev:
> Can you add those to the _headers file post-build command and Pages will be fine with that?

Yep. As long as they're there when we read the file (at publish step) it's all cool

KianNH:
Maybe I'll PR something that isn't R2 or Workers for once

PR #5282, opened 2022-07-25
[Docs Site] Add noscript alert to docs search

Cherry:
Currently when attempting to search and with JavaScript disabled, or when it's not present for any reason, you just get a blank white page like so: 

With this PR, the page now at least renders a warning to users like so: 

---

Ideally, a server-side search mechanism would be in place so that users can still search the docs when they have JS disabled (or it doesn't load for any reason - [there's lots](https://kryogenix.org/code/browser/everyonehasjs.html)), but this will at least serve as a good solution in the short term.

### DIFF
--- a/content/search.html
+++ b/content/search.html
@@ -3,6 +3,12 @@ title: Search results
 type: home
 layout: home
 ---
+<noscript>
+  <h2>Search</h2>
+  <p>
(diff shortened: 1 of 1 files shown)


kodster28 (approved):
LGTM

PR #5298, opened 2022-07-26
New microsoft-made package manager "winget" available

wellsilver:
Hi, I wanted to change the installation guide a bit for windows because Microsoft made a package manager called "winget" and cloudflared is on it
image
the tool is easy to use and pre-installed on all up-to-date windows computers


cicku:
Not a great idea.

This winget package is not well maintained. Winget reads data from https://github.com/microsoft/winget-pkgs/, which is not controlled by Cloudflare for releases.

PR #5334, opened 2022-07-28
[Doc-site]:Ignore activeTab when not needed

Ekwuno:
Removes console error for `getElementById("tab-active")` Thanks @eidam


kodster28 (approved):
LGTM

PR #5386, opened 2022-08-01
Add pcx content type to rendered content

kodster28:
addresses PCX-4345.

The `pcx-content-type` frontmatter needs to be changed to `pcx_content_type` so we can access that value w/in a Hugo template. This will not affect anything about how the page is rendered.

This PR adds a new value w/in the head section so we can pull it out when scraping (and use it for generating pre-filtered lists of content based on a certain type... i.e., Tutorials).

Screen Shot 2022-08-01 at 9 50 16 AM


marciocloudflare (approved):

VanessaRosaVi (commented):
>

Twitter1268 (commented):
Duplicate of #

Sammy361:
> addresses PCX-4345.
> 
> The `pcx-content-type` frontmatter needs to be changed to `pcx_content_type` so we can access that value w/in a Hugo template. This will not affect anything about how the page is rendered.
> 
> This PR adds a new value w/in the head section so we can pull it out when scraping (and use it for generating pre-filtered lists of content based on a certain type... i.e., Tutorials).
> 
> Screen Shot 2022-08-01 at 9 50 16 AM

PR #5408, opened 2022-08-02
Added note for frameworks

priyam-raj:
Added a note that states an issue with Cloudflare Pages according to which Redirects would only work if the site does not have any framework. 

The issue here is with the way some of these frameworks like Jekyll work, by generating a simple and static HTML/CSS/JS version of an entire project inside of a _site folder (in which we cannot write any extra file manually) and that's probably what is finally being deployed to Cloudflare's network. Because of this, even if the root folder of a project had a _redirects file, it would not be found in the deployed directory.

### DIFF
--- a/content/pages/platform/redirects.md
+++ b/content/pages/platform/redirects.md
@@ -36,7 +36,11 @@ filename: _redirects
 ```
 {{<Aside type= "note">}}
 
-Since the _redirects file needs to be present in the served root directory, some frameworks may not work right now as Cloudflare only deploys the build directory like the _site folder in case of Jekyll.
+In the case of some frameworks, such as Jekyll, you may need to manually copy and paste your `_redirects` file to the build output directory. To do this:
+
+1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com).
+2. Go to **Pages** > your Pages project > **Settings** > **Builds & deployments**.
+3. Go to **Build configurations** > **Edit configurations** > change the build command to `jekyll build && cp _redirects _site/_redirects` and select **Save**.
 
 {{</Aside>}}
 
--- a/content/pages/platform/redirects.md
+++ b/content/pages/platform/redirects.md
@@ -34,6 +34,11 @@ filename: _redirects
 /blog/* https://blog.my.domain/:splat
 /products/:code/:name /products?code=:code&name=:name
 ```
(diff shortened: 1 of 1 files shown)


WalshyDev:
Hey,

This isn’t an issue, more of an implementation detail. You can get this to work by copying the file into the _site dir after it’s built.
So, your build command could be like `jekyll build && cp _redirects _site/_redirects`

I’d say we probably just make that an aside in the Jekyll guide rather than adding an aside here. 

Thoughts @deadlypants1973 ?

priyam-raj:
Thanks for sending the workaround. This is really helpful for people like me who were confused, while a separate guide on it would be great I would still like to see a reference about this issue on the redirects docs.

content/pages/platform/redirects.md
```diff
@@ -34,6 +34,11 @@ filename: _redirects
 /blog/* https://blog.my.domain/:splat
 /products/:code/:name /products?code=:code&name=:name
 ```
+{{<Aside type= "note">}}
+
+Since the _redirects file needs to be present in the served root directory, some frameworks may not work right now as Cloudflare only deploys the build directory like the _site folder in case of Jekyll.
```
deadlypants1973, line 39: In the case of some frameworks, such as Jekyll, you may need to manually copy and paste your `_redirects` file to the build output directory.

deadlypants1973:
```suggestion
In the case of some frameworks, such as Jekyll, you may need to manually copy and paste your `_redirects` file to the build output directory. To do this:

1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com).
2. Go to **Pages** > your Pages project > **Settings** > **Builds & deployments**.
3. Go to **Build configurations** > **Edit configurations** > change the build command to `jekyll build && cp _redirects _site/_redirects` and select **Save**.
```

priyam-raj (commented):
Further note for frameworks

content/pages/platform/redirects.md
priyam-raj: @WalshyDev can we merge this now?

deadlypants1973 (approved):

WalshyDev (approved):
I think the aside should maybe be higher but approved

priyam-raj:
Thanks for the merge!

PR #5412, opened 2022-08-02
Remove sh from codeblock for module install command

skyfriends:
- Version published on website does't allow copying because the `sh` type set for the block.  `sh` tag doesn't provide any other apparent value.


JacobMGEvans (approved):
I agree with this, we don't want the UX to prevent using our own examples.

PR #5418, opened 2022-08-03
[Doc site]: Add copy to clipboard for code snippets

Ekwuno:
This PR adds copy to clipboard functionality for our code snippets. 
Screen Recording 2022-08-03 at 13 04 46

eg: 
https://copy-to-clipboard.cloudflare-docs-7ou.pages.dev/pages/platform/functions

### DIFF
--- a/assets/events.ts
+++ b/assets/events.ts
@@ -106,12 +106,13 @@ function $clicktoClipboard(ev: MouseEvent) {
 
 export function clipboardButton() {
   const copyButtonLabel = "Copy Code";
-  let blocks = document.getElementsByClassName("CodeBlock");
+  let blocks = document.getElementsByClassName("chroma");
   for (let i = 0; i < blocks.length; i++) {
     if (navigator.clipboard) {
       let button = document.createElement("button");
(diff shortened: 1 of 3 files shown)


rita3ko:
@Ekwuno it looks like the button is in active state by default: 
Screen Shot 2022-08-03 at 8 31 14 AM

It should be light gray when it's not pressed and become dark gray when you hover IMO. can you update?

rita3ko (approved):

WalshyDev:
I'd love to get feedback from @Cherry regarding design and accessibility

Ekwuno:
@Cherry come check this out!

Cherry (commented):
Thanks for the ping! This looks generally great, and the accessibility of the button works well! 🧡

My one comment would be in regards to a lack of any confirmation state that the code was actually copied. After I activate the button, it's not clear that anything actually happened. A common solution to this on the web is some kind of toast or notification. Here's a few examples:

- Google Translate: https://up.jross.me/kv902sg5
- NPM: https://up.jross.me/iem240pp
- Netlify docs: https://up.jross.me/k8o8m0pp

They're all slightly different, but each make it clear that something happened.

Ekwuno:
@Cherry, I addressed the call to action. Thank you so much.

pedrosousa:
When using the dark theme, the button icon is barely visible unless you hover it... I'm not sure if something changed in terms of CSS.
copy-button-hover

Cherry:
> @Cherry, I addressed the call to action. Thank you so much.

This looks great, thanks @Ekwuno! It appears to only be visible if I'm using a mouse though. If I tab to the copy button with my keyboard and then hit `enter` to activate it, there's no confirmation.

Ekwuno:
> > @Cherry, I addressed the call to action. Thank you so much.
> 
> This looks great, thanks @Ekwuno! It appears to only be visible if I'm using a mouse though. If I tab to the copy button with my keyboard and then hit `enter` to activate it, there's no confirmation.
Added!

pedrosousa:
~A lingering issue that seems to affect only the Windows platform (tested in both Chrome and Firefox) in one-line code blocks will be handled in a separate PR. Approving because this is not a blocking issue.~
Update: @Ekwuno solved the remaining issue.

pedrosousa (approved):

PR #5422, opened 2022-08-03
Update Workers learning section ordering to reflect more natural order

rita3ko:
Current order: 
Screen Shot 2022-08-03 at 9 26 50 AM

Updated order: 
Screen Shot 2022-08-03 at 9 27 16 AM

Note: as a future iteration, we should have sub-categories like "Workers fundementals", "Understanding Workers Storage", and "Understanding runtime APIs"


lauragift21 (approved):
Looks good to me!

PR #5424, opened 2022-08-03
Start using Module Syntax as default for overview

rita3ko:
Since we're not recommending most customers to use module syntax, we should use it as the primary example for the playground.


eidam (approved):
lgtm

PR #5425, opened 2022-08-03
update routing to custom domain

rita3ko:
Routing can be confused with routing within a Worker. Since we're pushing customers to custom domains, this is more intuitive. in the long term, we should have a "Triggers" section to reflect the same structure as the UI (will create a separate ticket)


eidam (approved):
lgtm

Ekwuno (approved):
🚢

PR #5438, opened 2022-08-03
[Docs site] Added a new value to search auto-focus

kodster28:
Should work for `CTRL` + `K` (in addition to the previous defaults of `SHIFT` + `S`, or `/`).

Will only work on pages with the top search bar (not the `/search/` page)

### DIFF
--- a/assets/search.redirect.ts
+++ b/assets/search.redirect.ts
@@ -22,7 +22,7 @@
     }
 
     // is '/' or SHIFT+'s' or COMMAND+'k'
-    if (key === 191 || (ev.shiftKey && key === 83) || (ev.metaKey && key === 75)) {
+    if (ev.key === '/' || (ev.shiftKey && key === 83) || (ev.key === 'k' && (ev.metaKey || ev.ctrlKey))) {
       ev.preventDefault();
       window.scrollTo(0, 0);
       element.focus();
--- a/assets/search.redirect.ts
+++ b/assets/search.redirect.ts
@@ -21,7 +21,7 @@
       return location.assign(redirect);
     }
 
-    // is '/' or SHIFT+'s' or CTRL+'k'
+    // is '/' or SHIFT+'s' or COMMAND+'k'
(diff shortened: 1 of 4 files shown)


kodster28:
FYI for @Cherry and @irvinebroque who independently brought this up to me. Didn't know we actually already had this functionality is our docs... but seems best to match the growing standard for CTRL + K.

assets/search.redirect.ts
```diff
@@ -21,8 +21,8 @@
       return location.assign(redirect);
     }
 
-    // is '/' or SHIFT+'s'
-    if (key === 191 || (ev.shiftKey && key === 83)) {
+    // is '/' or SHIFT+'s' or CTRL+'k'
+    if (key === 191 || (ev.shiftKey && key === 83) || (ev.ctrlKey && key === 75)) {
```
irvinebroque: Think what you want here is command-k, not ctrl-k (see Stripe and other docs for reference)

irvinebroque:
🎉  nice one!

assets/search.redirect.ts
kodster28: So close, lol

dcpena (approved):

Cherry (commented):
Oh cool! `metaKey` on Windows is the Windows key though unfortunately, and won't be captured by browsers. Using `Ctrl` for Windows would be ideal.

Algolia's approach [seems to be](https://github.com/algolia/docsearch/blob/2c64362067376a76575b7b5131972f71c920a406/packages/docsearch-react/src/useDocSearchKeyboardEvents.ts#L42):

```js
 (event.key === 'k' && (event.metaKey || event.ctrlKey))
```

WalshyDev:
Yeah, cmd is best for Mac but ctrl for Windows/Linux. Don't you love cross-platform 😅

kodster28:
Actually, @dcpena, could you review the new styling as well?

Sorta stole from Stripe's docs and added one of our shortcuts to be visible in the top search bar (on doc pages, not the main/tutorial pages).

Should also work in mobile, btw.

pedrosousa (approved):

PR #5453, opened 2022-08-05
Document `Upload-Creator` header for TUS uploads

mbetz08:


content/stream/uploading-videos/upload-video-file.md
```diff
@@ -85,6 +85,10 @@ Setting arbitrary metadata values in the `Upload-Metadata` header sets values th
 
 {{</definitions>}}
 
+### Additional supported headers
+
+Setting a creator value in the `Upload-Creator` header can be used to identify the creator of the video content, linking the way you identify your users or creators to videos in your Stream account. For more, read the [blog post](https://blog.cloudflare.com/stream-creator-management/)
```
irvinebroque: Let's link to the docs rather than the blog post.

@dcpena — can you 👁️ this and think about where in this page would be the best fit for this content?

mbetz08: re: linking to docs- I couldn't find an existing place in our docs that we reference the `creator` field, so this might require a larger change

mbetz08: Updated to point to the docs instead of blog post, leaving comment un-resolved so Denise can take a pass on where this might fit best

dcpena: To be honest, I would place this content in the same place you did @mbetz08. With the docs audit @irvinebroque and I have come up with, this page will eventually be easier to sift through and process.

At this point, my bigger concern is making the correction and ensuring it's documented. I can create a separate PCX ticket that I can prioritize higher, if it makes sense, to rework this page. What do y'all think?

irvinebroque (approved):
🚀 let's ship it and revisit where this lives later on

dcpena (approved):

PR #5458, opened 2022-08-06
Update limits.md to reflect purge ratelimits

leejt:
It's my understanding from talking to support and doing my own testing that purge-by-url doesn't count towards the 1200/5m limit. The docs were ambiguous whether the purge ratelimits are instead of, or in addition to, the global rate limit.

I think this is important information to spell out explicitly, since our team previously read this page to mean that the 1200/5m limit DID apply, which caused us to build unnecessary purging infrastructure, and might cause others to avoid using cloudflare.

Feel free to disregard this if the 1200/5m limit is meant to apply here, although be aware that it is not applied currently. Thanks!

### DIFF
--- a/content/api/limits.md
+++ b/content/api/limits.md
@@ -8,7 +8,8 @@ weight: 13
 
 The global rate limit for our API is 1200 requests per 5 minutes. If you exceed this, all API calls for the next 5 minutes will be blocked, receiving a HTTP 429 response.
 
-Some specific API calls have their own limits and are documented separately, such as the Cache Purge APIs:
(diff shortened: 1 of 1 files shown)


euwest:
The 1200/5 min rate limit did apply to purge-by-url until recently, but we are working on relaxing that limit for purge-by-url specifically. purge-by-tag, purge-by-host, purge-by-prefix, and purge-everything all still count toward the 1200/5 min rate limit, so we'll need to think about how to phrase this better. Thank you for the PR, @leejt

angelampcosta:
@leejt Thank you for the feedback. We'll track the changes and update the documentation accordingly.

PR #5460, opened 2022-08-07
Use the domain variable to get the zone for the record

declan-morris:
Use the domain variable to get the zone for the record
- var.domain is currently not used?

I think the above is more user friendly for first time users of the cloudflare provider as the domain var seems redundant previously however using it to get the zone id streamlines the process?

### DIFF
--- a/content/terraform/tutorial/initialize-terraform.md
+++ b/content/terraform/tutorial/initialize-terraform.md
@@ -32,16 +32,16 @@ provider "cloudflare" {
   api_token = "your-api-token"
 }
 
-variable "zone_id" {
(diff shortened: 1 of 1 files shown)


ranbel:
You are right that var.domain is not used in the "Initialize Terraform" tutorial. However, since both var.domain and var.zone_id are used in later tutorials (such as https://developers.cloudflare.com/terraform/tutorial/add-page-rules/), I think it's better to leave them in. 
https://github.com/cloudflare/cloudflare-docs/pull/6150 clarifies what the user needs to change in the example.

PR #5477, opened 2022-08-09
Remove duplicated curl from the PATCH example

coleplx:
There was a duplicated curl in the last example.

Also... it could be worth to mention that sending a PATCH to change ANY "settings" value will reset all other values. But that's up to you. ;)

For example:
Before:
```
      "settings": {
        "min_tls_version": "1.2"
      },
```
After:
```
      "settings": {
        "early_hints": "on"
      },
```


coleplx:
Documentation has been updated - Closing this.

:)

PR #5480, opened 2022-08-09
[Workers] Add Note to Increase Workers Limits

lauragift21:


rita3ko (approved):

arunesh90:
The form title (`Workers Limits Request Form`) currently seems to imply that it's only for limits related to Workers, but as the form is also put on the R2 and Pages limit pages, it may make sense to modify the form title to reflect it's also for those other products?

deadlynjdj:
it's probably for like if you have billions of requests to R2, you want to store petabytes in R2, or you have a massive site on pages
@arunesh90

PR #5484, opened 2022-08-09
[DRAFT] [Docs site] Learning paths PoC

kodster28:
Sample visible at https://learning-path-mvp.cloudflare-docs-7ou.pages.dev/learning-paths/get-started/


kodster28:
Figured you'd also be excited about this, @dcpena.

KianNH:
I totally wasn't looking at this when it was just a branch without a PR.

Nice work!

kodster28:
Known limitations:
- No "Clear filters" button

config.toml
```diff
@@ -43,3 +43,29 @@ tutorial_token = "xxa5a7d4c9-db72-424d-860d-ae68ff7f60ac"
 		lineNoStart = 1
 		lineNos = false
 		hl_Lines = ""
+
```
kodster28, line 46: I updated these values b/c I couldn't for the life of me figure out how to import the .json file from the `/assets/` folder into the `/data` folder (or vice versa).

This is possible if we config options for `module.mounts` values. Basically, if you set it, it'll duplicate the file in two places. There are some [limitations though](https://discourse.gohugo.io/t/absolute-paths-in-mounts/22359/7) namely that then you have to re-specify all the values for your other folder outputs.

From poking around I think this replicates the existing structure, but wanting re-assurance that this isn't mucking anything else up.

assets/learning-paths.ts
```diff
@@ -0,0 +1,77 @@
...
+    let selected_values = [];
+    for (const component of ui_components) {
+        let select = document.getElementById(component);
+        let selected_value = select.options[select.selectedIndex].value;
+        selected_values.push([component, selected_value]);
+}
+    return selected_values;
+}
+
+export function loadResults() {
```
kodster28: Theoretically, I could just let the Hugo template load the initial results and then just replace the HTML as needed... but figured it'd be easier to keep the template logic contained with a single function `buildHTML` so it doesn't live in two places.

assets/learning-paths.ts
```diff
@@ -0,0 +1,77 @@
+import * as learningPath from "./json/learning_paths.json"
```
kodster28: Feel free to call out anything that isn't super performant or could be done better. FE dev is definitely not my day job.

i.e., still wrapping my head around error handling as a concept

layouts/shortcodes/learning-path.html
```diff
@@ -0,0 +1,63 @@
+{{- $filename := .Page.RelPermalink -}}
+{{- $data := .Site.Data.learning_paths -}}
+
+{{- range where $data "path" $filename -}}
+
+<p class="viewOtherPaths"><a href="/learning-paths/"><strong>&lt</strong>  View other learning paths</a></p>
```
pedrosousa:
```suggestion
<p class="viewOtherPaths"><a href="/learning-paths/"><strong>&lt;</strong>  View other learning paths</a></p>
```

pedrosousa (approved):

PR #5506, opened 2022-08-12
Fix CircleCI deploy pages env variables

goldylucks:
Those two lines have the following problems:
1. invalid yml file
2. It doesn't work, the variables aren't passed down to wrangler
3. It actually works perfectly without them, so not only they aren't contributing, they are breaking the funcionality


WalshyDev (approved):

deadlypants1973 (approved):

goldylucks:
@deadlypants1973 why close this?

WalshyDev:
This was already fixed in an earlier PR, you can see files changed now shows no changes.

Thanks for the PR though! Please do open another if ou find more issues

PR #5517, opened 2022-08-13
[Docs site] Add table of contents highlighting

Cherry:
This PR adds a nice visual indicator to the table of contents in the docs as you scroll into each section. This is extremely useful on longer-form documents, to give users an idea of where abouts they are on the page, and how much they have left to get through.

Here's an example of it in action: https://up.jross.me/wgz6bab8

Do note that this isn't perfect, especially on shorter documents and may "skip" sections as you scroll if they're super close together. There might be an easy way to solve this, but I can't think of one at the moment given the DOM structure. It's a visual improvement and makes it way more digestible to read larger documents in my opinion, either way,

Here's a few examples of pages where it works well, from my docs fork:
- https://f1bbcd3d.cloudflare-docs-57q.pages.dev/pages/platform/functions/
- https://f1bbcd3d.cloudflare-docs-57q.pages.dev/pages/tutorials/build-an-api-with-workers/
- https://f1bbcd3d.cloudflare-docs-57q.pages.dev/stream/edit-manage-videos/edit-videos/video-clipping/
- https://f1bbcd3d.cloudflare-docs-57q.pages.dev/workers/learning/security-model/

Note: I also added a `tsconfig.json` in the `assets` folder and set it to a more modern target than the default `es5`. This is so I could use some more modern DOM iteration APIs without hurting TypeScript.

### DIFF
--- a/assets/contents.ts
+++ b/assets/contents.ts
@@ -2,6 +2,7 @@ type ListItem = HTMLLIElement & {
   h: string;
 };
 
+const ACTIVE_CLASS = 'DocsTableOfContents-link-active';
 export function toc() {
   let target = document.querySelector('ul.DocsTableOfContents');
   let article = target && document.querySelector('article.DocsMarkdown');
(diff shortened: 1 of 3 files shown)


kodster28 (approved):
Don't think there's any issue with the small inaccuracies on shorter pages, more than happy to accept that as a small limitation given the benefit.

Checked around on a couple pages and couldn't see any issues, so approving from a `It looks like it works for the business` perspective.

Any comments from @eidam or others on the actual TS part?

WalshyDev (approved):
LGTM from a code point of view

kodster28:
Coolio, feel free to merge away @Cherry!
(if you have the ability, if not just let me know and I can do that too)

Cherry:
I don't have merge perms, but am excited to see this land. Thanks @kodster28!

kodster28:
I'll keep an eye on stuff too just in case we need to revert.

PR #5519, opened 2022-08-14
Update index.md

fabform3:
added reference to static website forms


KianNH:
Hi there!

It looks like your reference is for GitHub Pages, not Cloudflare Pages - I'm not sure if that page is the right place for it.

PR #5524, opened 2022-08-15
[Workers] How to set multiple Set-Cookie headers

KianNH:
It's a bit confusing on how setting multiple cookies works.

1) You might think to use an array as the value for the header, like so...

```js
return new Response("Hello World!", {
  headers: {
    'Set-Cookie': ['cookie1=value_for_cookie_1; Path=/; HttpOnly;', 'cookie2=value_for_cookie_2; Path=/; HttpOnly;']
  }
});
```

Whilst the `HeadersInit` type is not a fan of this (being a `Record<string, string>`), this will be fine in a published Worker & joins the two strings together with a comma. But...

🚨 Folding multiple `Set-Cookie` headers isn't allowed in [RFC 6265](https://www.rfc-editor.org/rfc/rfc6265) 🚨

2) You might think to just add it twice into the `headers` object, like this...

```js
return new Response("Hello World!", {
  headers: {
    'Set-Cookie': 'cookie1=value_for_cookie_1; Path=/; HttpOnly;',
    'Set-Cookie': 'cookie2=value_for_cookie_2; Path=/; HttpOnly;'
  }
});
```

Nope - this isn't okay either since you can't have a duplicate key in an object.

🚨 `Duplicate key "Set-Cookie" in object literal [duplicate-object-key]` 🚨 

3) You probably wouldn't expect to use `Headers.append()` since the [Headers API](https://developer.mozilla.org/en-US/docs/Web/API/Headers) says this:

> The append() method of the [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) interface appends a new value onto an existing header inside a Headers object, or adds the header if it does not already exist.

This would just be the same as option 1, right? It'll append them together with a comma & we'll run into the same folding issue.

Nope! This is actually how you're supposed to set multiple `Set-Cookie` headers in Workers!

```js
const headers = new Headers();

headers.append("Set-Cookie", "cookie1=value_for_cookie_1; Path=/; HttpOnly;");
headers.append("Set-Cookie", "cookie2=value_for_cookie_2; Path=/; HttpOnly;");

console.log(headers.getAll("Set-Cookie"));
// Array(2) [ cookie1=value_for_cookie_1; Path=/; HttpOnly;, cookie2=value_for_cookie_2; Path=/; HttpOnly; ]
```

All looks good - and so is the response that we get back.

image

Since this is a special difference in how `Headers.append()` works due to the quirks of `Set-Cookie`, it should be documented.

Thank you for coming to my talk on why the `Set-Cookie` header is weird.


kristianfreeman (approved):

kristianfreeman:
@KianNH this looks good to me. Can you fix the merge conflict and then we can merge this? Thank you very much!

PR #5525, opened 2022-08-15
Add basic web example page, video.js example

irvinebroque:
See https://0e913b76.cloudflare-docs-7ou.pages.dev/stream/examples/video-js/ for how this will look

### DIFF
--- a/content/stream/examples/video-js.md
+++ b/content/stream/examples/video-js.md
@@ -1,14 +1,13 @@
-
 ---
+type: example
+summary: Example of Video.js playback with Cloudflare Stream
+tags:
+  - Playback
+pcx_content_type: configuration
 title: Video playback with Video.js
-weight: 1
-meta:
-  title: Example of Video.js playback with Cloudflare Stream
+weight: 1001
+layout: example
 ---
-
-# Video.js
-
-Example of Video.js playback with Cloudflare Stream
 
 [Video.js API docs](https://docs.videojs.com/)
 
--- a/content/stream/examples/video-js.md
+++ b/content/stream/examples/video-js.md
@@ -1,0 +1,39 @@
+
+---
+title: Video.js
+weight: 1
+meta:
+  title: Example of Video.js playback with Cloudflare Stream
+---
+
+### Video.js
+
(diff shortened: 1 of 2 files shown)


content/stream/examples/video-js.md
```diff
@@ -0,0 +1,39 @@
...
+      />
+    </video-js>
+
+    <script>
+      const vid = document.getElementById('vid1');
+      const player = videojs(vid);
+    </script>
+  </body>
+</html>
+```
```
irvinebroque, line 41: Not ideal to copy and paste code from Stackblitz example repo, but AFAIK we currently don't have a way to embed code snippets from another repository within our docs.

irvinebroque:
@dcpena — do you know if there is a way to default the 2nd-level nav hierarchy to be open rather than closed?

Ex: Instead of:

Screen Shot 2022-08-14 at 6 18 18 PM

Show:

Screen Shot 2022-08-14 at 6 18 14 PM

Having to expand the nav twice is a bit of extra work for the reader.

hussain2022a (commented):
[Hussain ]() @irvinebroque @davecardwell

kodster28:
@irvinebroque, to your question above to @dcpena, there's no way to default a navigation item to be open atm.

content/stream/examples/video-js.md
```diff
@@ -0,0 +1,43 @@
...
+pcx_content_type: configuration
+title: Video playback with Video.js
+weight: 1001
+layout: example
+---
+
+[Video.js API docs](https://docs.videojs.com/)
+
+{{<Aside>}}
+[**Run and edit this code in your browser on Stackblitz**](https://stackblitz.com/edit/cloudflare-stream-examples?file=examples%2Fvideo-js.html&initialPath=video-js&terminal=start-stackblitz)
```
dcpena, line 14: Let's save the asides for actual notes and warnings. I think a good way to still bring attention to this without using an aside would be to make the entire text a hyperlink. What do you think?

irvinebroque:
sounds good!

dcpena (approved):

PR #5529, opened 2022-08-15
updating legacy WAF ID which was mistakenly used in comments section

unknownhad:
updating legacy WAF ID which was mistakenly used in comments section


vs-mg (approved):

Princeobinna (commented):
Approve assets

PR #5540, opened 2022-08-16
Add DDoS change log entries for 2022-08-16

jdesgats:


kodster28:
This build is failing atm, but we've solved the issue in production so it'll be fine to merge in

kodster28 (approved):

PR #5559, opened 2022-08-17
[ZT] Add note about file path in application check

schultetwin1:
This note tells users the file path for a device posture application
check MUST point to an actual executable. This comes from multiple
customers attempting to checking if /Application/AppName.app is running
on a machine, when that is just the folder name for that application.


ranbel:
Good callout! Added to #5567

PR #5570, opened 2022-08-18
[ZT] Access auth cookie in Firefox private windows

ranbel:
pcx-4442


ranbel:
@jroyal 
- created a [new section](https://c66e0b7a.cloudflare-docs-7ou.pages.dev/cloudflare-one/identity/authorization-cookie/) for cookie-related topics. 
- moved CORS and JWT pages into this section
- added [note about Firefox private mode](https://c66e0b7a.cloudflare-docs-7ou.pages.dev/cloudflare-one/identity/authorization-cookie/#allow-cross-site-cookies-in-firefox)

content/cloudflare-one/identity/authorization-cookie/_index.md
```diff
@@ -0,0 +1,35 @@
+---
+pcx-content-type: concept
+title: Authorization cookie
+weight: 12
+layout: single
+---
+
+# Cloudflare authorization cookie
+
+When you protect a site with Cloudflare Access, Cloudflare checks every HTTP request bound for that site to ensure that the request has a valid `CF-Authorization` cookie. The cookie contains the user's identity in [JSON format](#access-jwts). If a request does not include the cookie, or if the identity data in the cookie does not match an Allow policy, the request will be blocked.
```
jroyal:
```suggestion
When you protect a site with Cloudflare Access, Cloudflare checks every HTTP request bound for that site to ensure that the request has a valid `CF-Authorization` cookie. The cookie contains the user's identity in [JSON format](#access-jwts). If a request does not include the cookie, or if the identity data in the cookie does not match a Service Auth policy, the request will be blocked.
```

jroyal (approved):
One small nit, but otherwise LGTM.

content/cloudflare-one/identity/authorization-cookie/_index.md
ranbel: Why does it have to be Service Auth specifically? How would a regular Allow policy know whether a logged-in user can access an application?

jroyal: honestly we should probably just drop the type entirely. You have a fair point because an Allow policy that requires Gateway will still check Gateway each time. 

I was coming it from the perspective that you didn't have a cookie. In which case the only way to login would be with a service auth policy.

ranbel: (for future reference)
Situations where a cookie exists but the user is blocked:
- Classic example is a policy that allows @[cloudflare.com](http://cloudflare.com/) and requires Gateway. If they successfully login while connected to Gateway we would issue the cookie. If they then turned off Gateway we would deny the request even if they had a valid cookie
- another option would be if the token itself was expired or had been revoked

PR #5574, opened 2022-08-19
Update rules-lists.md

michelbamps:
Currently only Open Proxies list is available for all customers. The other lists are part of Security Bundle (Core / Advanced). This must be made clear here to avoid customers trying to (frustratingly) use it.


pedrosousa (approved):

pedrosousa:
Thanks, @michelbamps!

PR #5583, opened 2022-08-21
Snippets

WalshyDev:
Products have a lot of limits. These limits are referred to all over the place which creates a problem... when we up (or remove) a limit, updating all the instances is often hard to do. This leads to there being different limits listed on different pages.

This has bugged me for a while as we've fixed multiple occurrences of this in the Pages docs. So, I wanted to solve it!

Introducing: Snippets

These are as you'd expect, little snippets of text that we can use all over the doc. This means we have a single source of truth rather than multiple. I have updated the Pages docs to use these and I'd encourage other teams to adopt them too!

(this even fixed a case where a limit was outdated, great way to show why this is needed)

Example pages:
- https://walshy-snippets.cloudflare-docs-7ou.pages.dev/pages/how-to/use-bulk-redirects/
- https://walshy-snippets.cloudflare-docs-7ou.pages.dev/pages/platform/limits/

### DIFF
--- a/layouts/shortcodes/snippet.html
+++ b/layouts/shortcodes/snippet.html
@@ -1,0 +1,23 @@
+{{- $id := .Get "id" -}}
+
+<!-- Pages -->
+{{ if (eq $id "cf-pages-limits-redirects-long") }}
+2,000 static redirects and 100 dynamic redirects, for a combined total of 2,100 redirects
+{{ else if (eq $id "cf-pages-limits-redirects-andor") }}
+2,000 static redirects and/or 100 dynamic redirects
+{{ else if (eq $id "cf-pages-limits-redirects") }}
+2,100
+{{ else if (eq $id "cf-pages-limits-headers") }}
+100
+{{ else if (eq $id "cf-pages-limits-headers-char-limit") }}
+1,000
+{{ else if (eq $id "cf-pages-limits-file-size") }}
+25 MiB
+{{ else if (eq $id "cf-pages-limits-file-count") }}
+20,000
+{{ else if (eq $id "cf-pages-limits-drag-and-drop-file-count") }}
+1,000
+{{ else if (eq $id "cf-pages-limits-projects") }}
+100
+
+{{ end }}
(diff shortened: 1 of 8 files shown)


layouts/shortcodes/snippet.html
```diff
@@ -0,0 +1,23 @@
+{{- $id := .Get "id" -}}
+
+<!-- Pages -->
+{{ if (eq $id "cf-pages-limits-redirects-long") }}
```
Ekwuno, line 4: I think this is brilliant @WalshyDev. Although, I am concerned about the syntax because this can also create another problem of its own in the sense that the "else if" will get longer with more products and the contributors can be prone to error 

First thought: Should we have a Shortcode usage section in our contribution docs? To help contributors get more comfortable when we add new shortcodes? (Thinking about this for the tab switcher one as well). 

Second thought:  Instead of a Snippet shortcode can we solve the issue of having limits all over the place differently? E.g. Updating https://developers.cloudflare.com/pages/platform/limits/ in a way that its information is always accurate and then reference it anywhere we talk about limits?

kodster28: Overall, love the idea of single sourcing! 

I think @Ekwuno's 2nd thought is valid... I'd have some concerns over this being formatted as an if/else if as well + the fact that then we'll have limits information in two separate places (the Limits page and this shortcode).

Theoretically, we could build out a JSON file with the various limit values in the **Data** folder and then use that JSON file to:
1. Pull specific data into the snippet shortcode.
2. Populate a new `Limits` shortcode that'd list out everything associated with a specific product.

Realize this'd be significantly more work and sorta run over what you've done here... but that's probably the ideal state from a reuse PoV. Might be something I or one of the other DAs could take on.

kodster28: Tracking in PCX-4576, btw

WalshyDev: Perfect, will close this out and let you take over :)

PR #5592, opened 2022-08-22
Delete Page Rule redirect

kyouheicf:
Page Rule redirect will cause redirect loop because Page Rule URL is case-insensitive.


kodster28:
You can ignore the gitguardian check, btw

angelampcosta:
@kyouheicf Thank you for your contribution!

angelampcosta (approved):

PR #5612, opened 2022-08-23
fix(assign-pr): Add REVIEWERS, and add them for the product when a PR opened by non-PCX owners

soheiokamoto:
This adds `REVIEWERS` list, and add them for the product when a PR opened by non-PCX owners.

If a PR is opened by one of PCX owners, then it would still add all other PCX owners (up to 10), and would not add anyone from REVIEWERS.

PR #5625, opened 2022-08-24
fix(assign-pr): Add REVIEWERS based on product even when PCX owner opens a PR

soheiokamoto:
This adds REVIEWERS based on product even when PCX owner opens a PR.
Currently, if non-PCX owner opens a PR, it adds REVIEWERS based on product, but if PCX owner opens a PR, it will only add all PCX OWNERS (up to 10), but not REVIEWERS.
10 reviewers limit is removed, as it seems like it allows more reviewers fine (may need to add it back later if it's not).

This is an enhancement on the change to add REVIEWERS:
 - https://github.com/cloudflare/cloudflare-docs/pull/5612
 - https://github.com/cloudflare/cloudflare-docs/pull/5618


soheiokamoto:
I don't know exact reason of current behavior (perhaps it was because there is 1 PCX owner per product now), but I think non-PCX owner product reviewers should still be added as reviewers when a PR is opened by PCX owner, when there's a change for their product.

pedrosousa (approved):

PR #5683, opened 2022-08-30
[Docs] Fix GitHub actions following the OWNERS update

pedrosousa:
I updated the doc OWNERS in #5661 but didn't regenerate the GitHub actions. This was now causing issues because Max, which I had removed, was still being referenced in the built version.

Updating the GitHub actions (built version) now in this PR.


marciocloudflare (approved):

pedrosousa:
Note: The action will still fail for this PR, since the code that is running is from the `production` branch.

PR #5695, opened 2022-08-30
updated custom domains limit

Albonycal:
changed 10 to 100


WalshyDev:
Thank you for the PR!
We are waiting on some backend changes to roll out still and want to add a bit more here. Will handle this update myself when we are ready :)

PR #5696, opened 2022-08-30
Add myself to Pages reviews

WalshyDev:


.github/actions/owners.ts
```diff
@@ -36,7 +36,7 @@ export const OWNERS: Record<string, string[]> = {
   "network-error-logging": ["dcpena"],
   "network-interconnect": ["dcpena"],
   "page-shield": ["pedrosousa"],
-  pages: ["deadlypants1973"],
+  pages: ["deadlypants1973", "WalshyDev"],
```
crwaters16:
```suggestion
  pages: ["deadlypants1973"],
```

crwaters16: We restrict owners to PCX reviewers (happy to chat about why separately) - you're welcome to add yourself to to `export const REVIEWERS` in the same way the logs/analytics products have in lines 63-65 to add yourself to reviews.

crwaters16 (changes requested):

.github/actions/owners.ts
WalshyDev: Added myself to REVIEWERS 👍 

Is there any difference between the two in terms of being assigned? My goal here is just to be assigned on all Pages PRs/issues so I can review them. I review them all anyway but just want to reduce the friction of me looking for them or them being sent to me.

WalshyDev: Ah ok I see this: https://github.com/cloudflare/cloudflare-docs/pull/5612

It would still be nice to have myself added to PRs opened by PCX just to reduce friction on needing to be pinged later.

crwaters16 (approved):

PR #5726, opened 2022-09-02
Fix incorrect field reference

tadhglewis:
Looks like this should be `code` not `error` as per https://developers.cloudflare.com/ruleset-engine/rules-language/fields/


pedrosousa:
You're right, @tadhglewis! Thank you for spotting the issue and for providing a PR to fix it.

pedrosousa (approved):

Chriss2999976 (commented):
#

PR #5310, opened 2022-07-27
Clarify API token permission needed

irvinebroque:
refs https://jira.cfops.it/browse/PCX-4384

Specify the exact permission that an API token must have in order to make requests to the Stream GraphQL Analytics API

@dcpena for edits on how to frame the "prerequisite" part of this in the context of an ordered list of steps.

PR #5752, opened 2022-09-06
[Stream] Clarifiy API token permission needed for GraphQL API

irvinebroque:
Supercedes https://github.com/cloudflare/cloudflare-docs/pull/5310

Specify the exact permission that an API token must have in order to make requests to the Stream GraphQL Analytics API

### DIFF
--- a/content/stream/getting-analytics/fetching-bulk-analytics.md
+++ b/content/stream/getting-analytics/fetching-bulk-analytics.md
@@ -147,12 +147,12 @@ header: GraphQL response
 }
 ```
 
-**About the response above:**
+About the response above:
 
-*   Each object inside videoPlaybackEventsAdaptiveGroups represents one video
-*   uid property represents the video uid
-*   count property shows the view count for one video during the specified date range
-*   timeViewedMinutes property shows the minutes viewed per video during the specified date range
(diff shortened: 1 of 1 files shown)


content/stream/getting-analytics/fetching-bulk-analytics.md
```diff
@@ -217,9 +145,16 @@ The response will look something like below. Things to remember:
   },
   "errors": null
 }
-
 ```
 
+**About the response above:**
```
dcpena: Can we place the property names within tick marks, so `videoPlaybackEventsAdaptiveGroups` and so on? Also, do you think property is necessary, or can we just say "`count` shows the view count for....?"

Let's add punctuation at the end of the sentences. 

Also, I know it's for emphasis, but part of our style guide is to refrain from capitalization or using bold, italics, or underline for emphasis on a word. With that said, I'd make the not lowercase.

irvinebroque: Addressed in https://github.com/cloudflare/cloudflare-docs/pull/5752/commits/09c3a81c8a39196e72ee634dfabc2cfee069eab0

dcpena (approved):

PR #5781, opened 2022-09-08
[Docs site] Fix weird mobile rendering for product name

kodster28:
Current mobile header lowercases certain names (`Dns`) and adds hyphens between spaces (`Waiting-Room`).

To test, shrink your view window and then check the `Waiting Room` docs + `DNS` docs in this version and our current version.


dcpena (approved):

pedrosousa:
Just noticed that the Pub/Sub tile shows `PubSub` because of the configured product name:
https://github.com/cloudflare/cloudflare-docs/blob/production/data/pub-sub.yml#L1

Since it works well in the SSL/TLS tile (where the displayed header is `SSL/TLS`, without any issues), can we rename the Pub/Sub product in the config file?

PR #5784, opened 2022-09-09
[ZT] Clarify CSV lists valid format

jannes:
Clarifies:
- no whitespace before line ending allowed
- trailing commas not needed (allowed because second column "description" is optional, but that's never mentioned anywhere, so the example file was very confusing)


ranbel:
#5786

PR #5788, opened 2022-09-09
[Docs Site] Added footer links that can be specific to the area of the docs

kodster28:
PCX-4714


kodster28:
Need to finish adding values for our other products, but this seems to be what you were asking for earlier, @irvinebroque with the specific links (if available).

data/cloudflare-for-saas.yml
```diff
@@ -12,6 +12,10 @@ meta:
   author: '@cloudflare'
   image: /cf-twitter-card.png
 
+resources:
+  dashboard_link: https://dash.cloudflare.com/?to=/:account/:zone/ssl-tls/custom-hostnames
+  discord: https://discord.com/channels/595317990191398933/897916119782928394
```
WalshyDev, line 17: You would need to be in the Discord to see this. I'd recommend using the invite instead (discord.gg/cloudflaredev) 

(and same with all the other Discord links)

ranbel (approved):

irvinebroque:
Nice one @kodster28!

PR #5810, opened 2022-09-12
Compatibility Dates API

GregBrimble:
This extracts out the compatibility dates into individual files (`content/workers/_partials/_platform-compatibility-dates/*.md`) with the following frontmatter:

```yml
# _build should always be set as follows. It tells Hugo not to build this as a real page.

_build:
  publishResources: false
  render: never
  list: never

# These two are mandatory:

name: "`HTMLRewriter` handling of `<esi:include>`" # the header of the feature
date: "2021-11-02" # the date to order the feature by

# These are all optional:

experimental: true # defaults false
enable_date: "2021-11-02" # the date from which that this feature defaults on for
enable_flag: "global_navigator" # the flag that can be used to manually turn this feature on
disable_flag: "no_global_navigator" # the flag that can be used to manually turn this feature off
```

The content of the file is just the description of the feature.

This is cool because it then means [we get an automatically generated API of the compatibility dates](https://compat-dates-api.cloudflare-docs-7ou.pages.dev/workers/platform/compatibility-dates/index.json). We plan to consume this in Pages and maybe Wrangler could use it too.

The original page remains unchanged: https://compat-dates-api.cloudflare-docs-7ou.pages.dev/workers/platform/compatibility-dates/


content/workers/_partials/_platform-compatibility-dates/fetch-improperly-interprets-unknown-protocols-as-http.md
```diff
@@ -0,0 +1,18 @@
...
+  list: never
+
+name: "`fetch()` improperly interprets unknown protocols as HTTP"
+date: "2022-11-10"
+enable_date: "2022-11-10"
+enable_flag: "fetch_refuses_unknown_protocols"
+disable_flag: "fetch_treats_unknown_protocols_as_http"
+---
+
+Originally, if the `fetch()` function was passed a URL specifying any protocol other than `http:` or `https:`, it would silently treat it as if it were `http:`. For example, `fetch()` would appear to accept `ftp:` URLs, but it was actually making HTTP requests instead.
```
deadlypants1973, line 14:
```suggestion
Originally, if the `fetch()` function was passed a URL specifying any protocol other than `http:` or `https:`, it would silently treat the protocol as if it were `http:`. For example, `fetch()` would appear to accept `ftp:` URLs, but it was actually making HTTP requests instead.
```

content/workers/_partials/_platform-compatibility-dates/htmlrewriter-handling-of-esi-include.md
```diff
@@ -0,0 +1,19 @@
...
+
+name: "`HTMLRewriter` handling of `<esi:include>`"
+date: "2021-11-02"
+experimental: true
+enable_flag: "html_rewriter_treats_esi_include_as_void_tag"
+---
+
+The HTML5 standard defines a fixed set of elements as void elements, meaning they do not use an end tag: `<area>`, `<base>`, `<br>`, `<col>`, `<command>`, `<embed>`, `<hr>`, `<img>`, `<input>`, `<keygen>`, `<link>`, `<meta>`, `<param>`, `<source>`, `<track>`, and `<wbr>`.
+
+HTML5 does not recognize XML self-closing tag syntax. For example, `<script src="foo.js" />` does not specify a script element with no body. A `</script>` ending tag is still required. The `/>` syntax simply is not recognized by HTML5 at all; it is treated the same as `>`. However, many developers still like to use this syntax, as a holdover from XHTML, a standard which failed to gain traction in the early 2000's.
```
deadlypants1973:
```suggestion
HTML5 does not recognize XML self-closing tag syntax. For example, `<script src="foo.js" />` does not specify a script element with no body. A `</script>` ending tag is still required. The `/>` syntax simply is not recognized by HTML5 at all and it is treated the same as `>`. However, many developers still like to use this syntax, as a holdover from XHTML, a standard which failed to gain traction in the early 2000's.
```

content/workers/_partials/_platform-compatibility-dates/minimal-subrequests.md
```diff
@@ -0,0 +1,18 @@
...
+  list: never
+
+name: "Minimal subrequests"
+date: "2022-04-05"
+enable_date: "2022-04-05"
+enable_flag: "minimal_subrequests"
+disable_flag: "no_minimal_subrequests"
+---
+
+With the `minimal_subrequests` flag set, `fetch()` subrequests sent to endpoints on the Worker's own zone (also called same-zone subrequests) have a reduced set of features applied to them. In general, these features should not have been applied to same-zone subrequests in the first place, and very few user-facing behavior changes are anticipated. Specifically, Workers might observe the following behavior changes with the new flag:
```
deadlypants1973:
```suggestion
With the `minimal_subrequests` flag set, `fetch()` subrequests sent to endpoints on the Worker's own zone (also called same-zone subrequests) have a reduced set of features applied to them. In general, these features should not have been initially applied to same-zone subrequests, and very few user-facing behavior changes are anticipated. Specifically, Workers might observe the following behavior changes with the new flag:
```

deadlypants1973 (approved):

PR #5844, opened 2022-09-14
MFW-891: Document multi-colo support for pcaps.

arges:


content/magic-firewall/how-to/collect-pcaps.md
```diff
@@ -86,7 +86,7 @@ header: Full PCAP example response
         "system": "magic-transit",
         "packet_limit": 10000,
         "byte_limit": 100000000,
-        "colo": "sfo06",
+        "colo": "ORD",
```
nadinelyab, line 89: You should also change the colo in the example request above to match the response

arges: done

dcpena (approved):

PR #5866, opened 2022-09-15
[Workers] FetchEvent refactor

KianNH:
**Draft**

One of many in the series of moving documents over to using TabSwitcher, providing TypeScript examples, applying standard code formatting to any examples & whatnot.


KianNH:
The changes to the tab switcher shortcake are comprised of 3 main things:

1) Fixed the Hugo conditionals, as a `js/sw` tab showed up like this:

image

2) Fixed the aria `role` attributes - they should be under the `path` rather than the `sag` element, as well as assigning the proper attribute for logo images.

3) Overall readability - this is subjective, naturally.

Let me know if that wants breaking into another PR, but note that this PR relies on point 1.

Edit: Will be splitting this into another PR due to other changes I've had to make.

https://github.com/cloudflare/cloudflare-docs/pull/5867

PR #5867, opened 2022-09-16
[Docs Site] Tab Switcher fixes

KianNH:
This PR fixes several bugs:

1) A JavaScript Service Worker has two labels, one being `js-sw`

image

This is due to the un-necessary `end` conditional in `tabs.html`.

```html
{{ end }} {{ if or (eq $link "js-esm") }}
```

2) `js` and `js-sw` have `tab-label active` by default, even if not active or defaulted.

image

```html
{{ if or (eq $link "js") }}
<a class="tab-label active" href="#" data-link="tab-{{$link}}" data-id="{{$unique_id}}">
        
{{ else if (eq $link "js-sw") }}
<a class="tab-label active" href="#" data-link="tab-{{$link}}" data-id="{{$unique_id}}">
```

3) Defaults other than the first tab do not render correctly.

image

This is because the current code will always set the first tab to display as well as set the default to display.

```js
// Set the first tab in a group to display
(tabs[0] as HTMLElement).style.display = "block";
```

4) The actively selected tab is not highlighted.

image

This is because there is nothing to set this as `active` - this includes both the default tab and when selecting a tab.

5) Aria labels.

The `role` aria label is meant for the `path` element so moved it to the appropriate place, as well as given it the `graphics-symbol` appropriate for logos.

6) General HTML readability changes.


assets/events.ts
```diff
@@ -173,10 +173,31 @@ export function tabs() {
...
+      const defaultTab = wrappers[i].querySelector(".tab.tab-default")
 
       if (tabs.length > 0) {
-        // Set the first tab in a group to display
-        (tabs[0] as HTMLElement).style.display = "block";
+        // if a tab has been specified as default, set that
+        // as active as opposed to defaulting to the first tab
+        if (defaultTab) {
+          // changes an id (tab-js-esm-6f3904f86f90c21d) into just the type (tab-js-esm)
+          const tabId = defaultTab.id.split("-").slice(0, 3).join("-");
```
KianNH: I'll admit - I only ever write Workers, nothing to do with the DOM. Is there nicer ways to do this?

assets/events.ts
```diff
@@ -173,10 +173,31 @@ export function tabs() {
...
+
+          let defaultTabLabel;
+          let found = false;
+          labels.forEach((label) => {
```
KianNH: As an alternative, I also had the below working if that's a better solution.

```js
const defaultTabLabel = wrappers[i].querySelector(`a[data-link=${tabId}]`);
```

assets/events.ts
```diff
@@ -173,10 +173,31 @@ export function tabs() {
...
+            if (!found) {
+              if (labelId === tabId) { defaultTabLabel = label; found = true }
+            }
+          });
+
+          (defaultTab as HTMLElement).style.display = "block";
+          (defaultTabLabel as HTMLElement).classList.add("active");
+        } else {
+          (tabs[0] as HTMLElement).style.display = "block";
+          (labels[0] as HTMLElement).classList.add("active");
```
KianNH, line 193: `(labels[0] as HTMLElement).classList.add("active");` is new and means that if there is no default, the first tab label will be shown as active.

KianNH:
cc @Ekwuno @WalshyDev

assets/events.ts
rozenmd: There are other ways to do this, but the comment explaining it does the job tbh

assets/events.ts
```diff
@@ -173,10 +173,31 @@ export function tabs() {
...
+        if (defaultTab) {
+          // changes an id (tab-js-esm-6f3904f86f90c21d) into just the type (tab-js-esm)
+          const tabId = defaultTab.id.split("-").slice(0, 3).join("-");
+
+          let defaultTabLabel;
+          let found = false;
+          labels.forEach((label) => {
+            const labelId = label.getAttribute("data-link");
+            if (!found) {
+              if (labelId === tabId) { defaultTabLabel = label; found = true }
```
rozenmd:
```suggestion
              if (labelId === tabId) { 
                defaultTabLabel = label; 
                found = true;
               }
```

rozenmd (changes requested):
One small tweak for readability

rozenmd (approved):
LGTM 👍

assets/events.ts
KianNH: I've just realised whilst testing it on the `unstable_dev` API page that it doesn't work with `tab-js`/`tab-ts` since it's assuming it'll be `tab-js-esm`/etc

KianNH:
```js
// split by - and get the length of the array
let parts = defaultTab.id.split("-").length;
// if there is 3, the id is tab-js-id
// otherwise the id is tab-js-esm-id
const end = parts = 3 ? 2 : 3
// changes an id (i.e tab-js-esm-6f3904f86f90c21d) into just the type (tab-js-esm)
const tabId = defaultTab.id.split("-").slice(0, end).join("-")
```

This _works_ but I don't like it.

KianNH:
```js
const tabId = defaultTab.id.split("-").slice(0, parts - 1).join("-")
```

Of course I can just do this. Rubber ducky always works.

assets/events.ts
```diff
@@ -173,10 +173,36 @@ export function tabs() {
...
+          let defaultTabLabel;
+          let found = false;
+          labels.forEach((label) => {
+            const labelId = label.getAttribute("data-link");
+            if (!found) {
+              if (labelId === tabId) { 
+                defaultTabLabel = label; 
+                found = true;
+               }
+            }
```
WalshyDev: no need for an inner if
```suggestion
            if (!found && labelId === tabId) {
              defaultTabLabel = label; 
              found = true;
            }
```

assets/events.ts
WalshyDev: yeah I was surprised you didn't do this. I'd prefer this

layouts/shortcodes/tabs.html
```diff
@@ -5,80 +5,101 @@
     <ul class="tab-active" block-id="{{$unique_id}}">
       {{ range $idx, $txt := $labels }} {{- $link := replace $txt "/" "-"}}
       <li class="tab-label-wrapper">
-        {{ if or (eq $link "js") }}
-        <a class="tab-label active" href="#" data-link="tab-{{$link}}" data-id="{{$unique_id}}">
+        {{ if (eq $link "js") }}
+        <a class="tab-label" href="#" data-link="tab-{{$link}}" data-id="{{$unique_id}}">
           <svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 512 512" height="1em" width="1em"
-            xmlns="http://www.w3.org/2000/svg" data-link="tab-{{$link}}" role="presentation">
```
WalshyDev, line 11: noob with a11y but why did the role move from the svg to the path?

KianNH: The current `role="presentation"` isn't valid for `svg` iirc & the proper role, `graphics-symbol`, is only accepted on the `path` since that's the actual image.

Before:
image

After:
image

WalshyDev: interesting, it was James who said to add that. Sounds good :)

KianNH: `role="img"` seems happy - just not `presentation`

image

image

Cherry: The reason I suggested this was so a screen-reader didn't read the SVG title, and then the human-written too, so you don't get "JavaScript JavaScript" read out with both the `title` in the SVG and the text outside.

Seems like `aria-hidden=true` would be better solution though - the SVG provides no semantic meaning here (does it?) so simply hiding it would be ideal: https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs

KianNH: > Assigning role="[presentation](https://www.w3.org/TR/wai-aria-1.0/roles#presentation)" should be avoided, because it doesn't hide the SVG. It just cancels the native role.

That'll be why WebStorm got upset at me then

> In order to hide an SVG, it is best to give it an attribute of aria-hidden='"true". This functions in a similar way to display:none in CSS.

Are there any good extensions/DevTools to test stuff like this? Safari has an accessibility audit but it seems to be more critical stuff.

KianNH: It doesn't look like we actually have any title/content in the `svg` for a screen reader to use - we have `JavaScript` as the content for the `a` element.

`a`
image

`svg`
image

Cherry: > It doesn't look like we actually have any title/content in the `svg` for a screen reader to use - we have `JavaScript` as the content for the `a` element.
> 
> `a` image
> 
> `svg` image

Sweet, this can probably just be removed then. There was a time there was a `<title>` element in the SVG, but seems that's no longer the case.

WalshyDev (approved):

WalshyDev:
Appreciate it @KianNH 👍 

Tested this out and it's good to

PR #5879, opened 2022-09-16
[Docs site] Refactored learning paths

kodster28:
Refactored learning paths to a) directly import files (and get away from module tinkering) and b) move paths into their own file for readability.

Will update internal wiki instructions shortly, b/c this does add a touch of overhead to the process... but it's a lot more readable with the files in separate folders.


assets/json-collector.ts
```diff
@@ -0,0 +1,3 @@
+export {default as a} from "./json/get-started.json"
```
kristianfreeman: can they have names like `get-started` instead?

kodster28: @pedrosousa had the same feedback and his suggestion (just committed) helps make that clearer

pedrosousa (approved):

PR #5881, opened 2022-09-16
Add DDoS change log entries for 2022-09-16

sbohrer:
@jdesgats I'm not sure if this PR is correct or not.  The scheduled-changes.md file looks right but the other 2022-09-16.md file has GRE filters that I don't think were touched recently.


sbohrer:
Actually after review this PR does look correct.  We disabled the two filters listed in 2022-09-16.md which is why it shows they were updated.

jdesgats (approved):

PR #5897, opened 2022-09-19
[R2] 09-19, 09-27 & 09-28 changelogs

KianNH:
https://community.cloudflare.com/t/2022-9-16-workers-runtime-release-notes/420496?u=kiannh

Will want to confirm the types (i.e if `delete()` now takes `string[]`) before merging.

### DIFF
--- a/content/r2/data-access/workers-api/workers-api-reference.md
+++ b/content/r2/data-access/workers-api/workers-api-reference.md
@@ -223,6 +223,12 @@ There are 3 variations of arguments that can be used in a range:
 
   - A map of custom, user-defined metadata that will be stored with the object.
 
+{{<Aside type="note">}}
+
+Only a single hashing algorithm can be specified at once.
+
+{{</Aside>}}
+
 - {{<code>}}md5{{<param-type>}}ArrayBuffer{{</param-type>}}|{{<param-type>}}string{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}{{</code>}}
 
   - A md5 hash to use to check the received object's integrity.
@@ -379,7 +385,7 @@ If an additional checksum was provided when using the `put()` binding, it will b
 
 {{<definitions>}}
 
-- {{<code>}}md5{{</code>}} {{<param-type>}}ArrayBuffer{{</param-type>}}
+- {{<code>}}md5{{</code>}} {{<param-type>}}ArrayBuffer{{</param-type>}} {{<prop-meta>}}optional{{</prop-meta>}}
 
   - The MD5 checksum of the object.
 
--- a/content/r2/data-access/workers-api/workers-api-reference.md
+++ b/content/r2/data-access/workers-api/workers-api-reference.md
@@ -78,7 +78,7 @@ export default {
   - Stores the given {{<code>}}value{{</code>}} and metadata under the associated {{<code>}}key{{</code>}}. Once the write succeeds, returns an `R2Object` containing metadata about the stored Object.
   - R2 writes are strongly consistent. Once the Promise resolves, all subsequent read operations will see this key value pair globally.
 
-- {{<code>}}delete(key{{<param-type>}}string | string[]{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
+- {{<code>}}delete(keys{{<param-type>}}string | string[]{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
 
   - Deletes the given {{<code>}}values{{</code>}} and metadata under the associated {{<code>}}keys{{</code>}}. Once the delete succeeds, returns {{<code>}}void{{</code>}}.
    - R2 deletes are strongly consistent. Once the Promise resolves, all subsequent read operations will no longer see the provided key value pairs globally.
--- a/content/r2/data-access/workers-api/workers-api-reference.md
+++ b/content/r2/data-access/workers-api/workers-api-reference.md
@@ -78,10 +78,10 @@ export default {
   - Stores the given {{<code>}}value{{</code>}} and metadata under the associated {{<code>}}key{{</code>}}. Once the write succeeds, returns an `R2Object` containing metadata about the stored Object.
   - R2 writes are strongly consistent. Once the Promise resolves, all subsequent read operations will see this key value pair globally.
 
-- {{<code>}}delete(key{{<param-type>}}string{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
+- {{<code>}}delete(key{{<param-type>}}string[]{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
 
-  - Deletes the given {{<code>}}value{{</code>}} and metadata under the associated {{<code>}}key{{</code>}}. Once the delete succeeds, returns {{<code>}}void{{</code>}}.
-   - R2 deletes are strongly consistent. Once the Promise resolves, all subsequent read operations will no longer see this key value pair globally.
+  - Deletes the given {{<code>}}values{{</code>}} and metadata under the associated {{<code>}}keys{{</code>}}. Once the delete succeeds, returns {{<code>}}void{{</code>}}.
+   - R2 deletes are strongly consistent. Once the Promise resolves, all subsequent read operations will no longer see the provided key value pairs globally.
 
 - {{<code>}}list(options{{<param-type>}}R2ListOptions{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}) {{<type>}}Promise\<{{<param-type>}}R2Objects{{</param-type>}}|{{<param-type>}}null{{</param-type>}}>{{</type>}}{{</code>}}
 
@@ -207,6 +207,10 @@ There are 3 variations of arguments that can be used in a range:
 
 {{<definitions>}}
 
+- {{<code>}}onlyIf{{<param-type>}}R2Conditional{{</param-type>}}{{</code>}}
+
+  - Specifies that the object should only be stored given satisfaction of certain conditions in the `R2Conditional`. Refer to [Conditional operations](#conditional-operations).
+
 - {{<code>}}httpMetadata{{<param-type>}}R2HTTPMetadata{{</param-type>}}|{{<param-type>}}Headers{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}{{</code>}}
 
   - Various HTTP headers associated with the object. Refer to [HTTP Metadata](#http-metadata).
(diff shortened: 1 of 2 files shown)


deadlypants1973 (approved):

Frederik-Baetens:
Hi Kian, just saw this PR. I made the change to allow delete to take string[] and also just tested it in a production worker to check that the deletion of multiple objects actually works :)

KianNH:
Awesome! I tested it at the time. Well, I tried `'foo', 'bar'` and it took me a minute to realise it should be `['foo', 'bar']` but we don't talk about that :^)

I think `R2Checksums` should be accurate - it's mostly a guess of what the type will look like until we get a new workers-types release

Frederik-Baetens:
Yeah, I should've been a bit clearer in the release note.

I checked the type in workers-types and it'll be this:
```
  "R2Checksums": {
    "name": "R2Checksums",
    "members": [
      {
        "name": "md5",
        "type": {
          "name": "ArrayBuffer",
          "optional": true
        }
      },
      {
        "name": "sha1",
        "type": {
          "name": "ArrayBuffer",
          "optional": true
        }
      },
      {
        "name": "sha256",
        "type": {
          "name": "ArrayBuffer",
          "optional": true
        }
      },
      {
        "name": "sha384",
        "type": {
          "name": "ArrayBuffer",
          "optional": true
        }
      },
      {
        "name": "sha512",
        "type": {
          "name": "ArrayBuffer",
          "optional": true
        }
      }
    ],
    "comment": {
      "text": "The checksums associated with the object.",
      "renamed": true
    },
    "kind": "struct"
  },
```

So I think what you wrote is correct.

I wonder if it would be possible to document the hashes in putOptions using " | " so that it's clearer that only one hash can be passed at a time. 

Perhaps something like this:

```
- {{<code>}}md5|sha1|sha256|sha384|sha512{{<param-type>}}ArrayBuffer{{</param-type>}}|{{<param-type>}}string{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}{{</code>}}

  - A hash to use to check the received object's integrity. Only one hashing algorithm can be specified at once.
```
Although this is a bit ugly.

@deadlypants1973 do you have an opinion on how we might best format this? Since these are indeed all separate fields, doing what Kian did in this PR makes a lot of sense, but I feel like it's a bit messy considering that those fields are mutually exclusive.

KianNH:
Is `md5` optional on `R2Checksums`? I believe I've always seen it present, even if a user-provided `sha???` is provided, since that's what is used for the ETag.

Good point about the checksums - I found that when playing around & trying to provide two but forgot to add it in here.

> Although this is a bit ugly.

True, but already done for the types that `value` can be when using `put()` - so even if we did it now with a vision to refactor it all in the future to be presented better, it'd be consistent.

Frederik-Baetens:
md5 should indeed be optional. It won't be present for multipart uploaded objects since it's quite difficult architecturally to compute the object hash as it's being uploaded with a multipart object.

KianNH:
That's a very good point - I suppose multipart is still `<randomhash>-<number of parts>`

KianNH:
We could use an aside?

image

Frederik-Baetens:
That looks really nice as well, but I'm still a bit concerned about just how messy still it is. I realize this is a direct result of the api being messy, but I couldn't get around that because it seemed more important to be consistent with the already existing md5 field. It's a tough choice between the messiness listing all those fields separately and the abnormality of using " | " for field names.

KianNH:
Overall, I've wanted to refactor the Runtime APIs page for awhile.

I prefer KV's layout with a lot more nuance around the 'why', with examples, as opposed to the 'raw' types and a little sidenote. https://developers.cloudflare.com/workers/runtime-apis/kv/#writing-key-value-pairs

As a rough concept, with poor wording, this is what I'd prefer:

image

It can be expanded to say 'if the hash does not match the checksum that R2 computes, you will get `<error desc here>`' and stuff like that.

I think the current docs provide value to those who know what to do with types and methods - but the way I'd rather have them is a lot more approachable and serves to teach people as opposed to guide them.

Frederik-Baetens:
That looks fantastic to me, but a slightly more structured approach might be better since then we can try to accommodate both the needs of people needing just an api reference, and the needs of people looking for more expansive documentation.

To me, the rust documentation seems like the best documentation I've seen anywhere. E.g.: https://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html It beautifully combines a more extensive general description, useful examples with still having a very structured general layout. Do you think we could work towards a structure like that?

E.g. You could keep the examples and explanation you show there, and perhaps put it under such a `md5|sha1|sha256|sha384|sha512` entry?

I realize that that still isn't as beautiful and readable for newer users looking for an introduction as what you wrote, but perhaps we should put more walktrough level explanations on separate pages? I think checksums are a bit too niche to put them on [this general walktrough page](https://developers.cloudflare.com/r2/data-access/workers-api/workers-api-usage/) but perhaps we could make entirely separate pages for guides like this? E.g. AWS does this as well, on top of explaining the various checksum parameters in their api reference: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html Not sure if that's applicable here yet since we don't have quite as much to say about object integrity as S3 does.

KianNH:
Rust & Stripe docs have always been the ones that I look to if I need inspiration or a code sample.

It'd be a pretty big change to how the other APIs for Workers are documented, but that's not an issue - if we can get a layout that works then it's easy to change the other documentation for things like KV to match.

Especially since if we get a good layout that we want to stick with, we can standardise as opposed to each runtime API having a different format.

We could look at moving the more in-depth examples to the examples page, going for the same sort of style as https://developers.cloudflare.com/workers/examples/ with focused, individual examples and explanations to achieve a single goal.

I feel like the existing list example (that shows you how to avoid falling into the length trap due to truncation) would be an example under a `list` equivalent of Rust's implementation section and only showing the proper way.

We could maintain a small block, probably 10 lines at most, under `put` to show conditionals/hashes/metadata and then leave the "tutorial" style to its own page.

KianNH:
I suppose another consideration is how we make the distinction between R2Bucket, R2Object and R2ObjectBody - the equivalent of Rust's structs which are separate on docs.rs

Are we documenting the available methods and properties on both buckets and objects in the same page? Do we make them separate?

KianNH:
Examples:

image

image

image

Frederik-Baetens:
Wow, that looks fantastic! I'd be 100% in favor of that. I hadn't even thought about making it expandable.

I had a look around and you're right that there isn't much coherence in the docs. For example the [durable object docs](https://developers.cloudflare.com/workers/runtime-apis/durable-objects/) take an approach that's a bit more similar to the api reference style that R2 has, but their actual structure is significantly different from what the R2 api reference looks like. This feels really messy and could probably use a general review so that we can come to one consistent structure for the docs.

In the meantime, adding additional information and short examples under each method and field seems like the best short term approach. Longer examples could indeed go on a separate examples page like with the general workers examples.

(disclaimer: I'm quite new on the R2 team, so I'm not a position to comment with authority on what the structure of the docs should look like, but I really like your last proposal)

KianNH:
We'll probably want to get the go-ahead from @deadlypants1973 before I spend an evening rewriting it all to find out it clashes with the PCX style guide - but beyond that, I'm a big fan of the collapsible style. 

We can provide lots of value without having to worry about an essay of a page with everything being pushed away by ~10 lines of parameters/examples under every method or property.

image

I've seen it elsewhere on the docs but I also stole the idea of using it for methods/APIs from docs.rs

> (disclaimer: I'm quite new on the R2 team, so I'm not a position to comment with authority on what the structure of the docs should look like, but I really like your last proposal)

I'm just a bored community contributor who likes writing docs - so I'll also defer to Kate haha

KianNH:
image

I'm on the fence about including types/return types in the heading of the collapsible tab since the text is quite large and it can get unruly (it'll be multiple lines for checksums or the different types for a `put` body)

Frederik-Baetens:
I prefer having the full type signature listed somewhere. If you really think it's to messy an option might be to only show the full signature with types when the documentation is expanded, but I think putting the full signature with types & return value in the header is probably acceptable too if we can make it look right.

KianNH:
These would be the two options, I suppose

image

image

I'd quite like it if the heading text could be like a code block, so monospaced with syntax highlighting - but that's a larger overall 'docs site' piece I think. Worth a look either way.

KianNH:
image

image

I spent a bit more time coming up with a good example.

Frederik-Baetens:
Really like that as well. I think this PR is probably fine as is, and we can take those great ideas into a comprehensive review of what the general style & structure of R2 or all workers api docs should look like.

(small nit, but I think it's strange to have a "•" in front of both the fields and the information about those fields. It's probably cleaner to only have the "•" in front of the field and then the comment after it without a dot. That way the dot serves as a clearer separator of fields.)

KianNH:
> (small nit, but I think it's strange to have a "•" in front of both the fields and the information about those fields. It's probably cleaner to only have the "•" in front of the field and then the comment after it without a dot. That way the dot serves as a clearer separator of fields.)

image

If there's a dot or not (or if that dot is generated by a `-` or `*`) is another thing that varies page to page haha - but yeah, I think this looks better.

> Really like that as well. I think this PR is probably fine as is, and we can take those great ideas into a comprehensive review of what the general style & structure of R2 or all workers api docs should look like.

I'll make a new issue to evaluate what layouts we can look to refactor the APIs with.

KianNH:
https://github.com/cloudflare/cloudflare-docs/issues/5917

content/r2/data-access/workers-api/workers-api-reference.md
```diff
@@ -78,10 +78,10 @@ export default {
   - Stores the given {{<code>}}value{{</code>}} and metadata under the associated {{<code>}}key{{</code>}}. Once the write succeeds, returns an `R2Object` containing metadata about the stored Object.
   - R2 writes are strongly consistent. Once the Promise resolves, all subsequent read operations will see this key value pair globally.
 
-- {{<code>}}delete(key{{<param-type>}}string{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
+- {{<code>}}delete(key{{<param-type>}}string | string[]{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
```
Frederik-Baetens: Would it make sense to change key -> keys here? I feel like the plural makes more sense if both one or multiple keys are allowed.

KianNH: Makes sense to me - will `workers-types` show `keys` when that has a new release?

cc @mrbbot :wave: don't suppose you know when `workers-types` will re-run to include the new R2 changes?

Frederik-Baetens: Yes, workers-types will show keys:

```
      {
        "name": "delete",
        "type": {
          "params": [
            {
              "name": "keys",
              "type": {
                "name": "|",
                "args": [
                  {
                    "name": "string"
                  },
                  {
                    "name": "[]",
                    "args": [
                      {
                        "name": "string"
                      }
                    ]
                  }
                ]
              }
            }
          ],
          "returns": {
            "name": "Promise",
            "args": [
              {
                "name": "void"
              }
            ]
          }
        }
      },
```

KianNH:
```suggestion
- {{<code>}}delete(keys{{<param-type>}}string | string[]{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}void {{</param-type>}}>{{</type>}}{{</code>}}
```

KianNH:
@deadlypants1973 added the 2 recent releases changelogs onto this one - ready to go imo

content/r2/platform/changelog.md
```diff
@@ -6,6 +6,21 @@ weight: 3
 
 # Changelog
 
+## 2022-09-28
+
+- Fixed a bug where CORS configuration was not being applied to S3 endpoint.
+- No-longer render the `Access-Control-Expose-Headers` response header if `ExposeHeader` is not defined.
+- Public buckets will no-longer return the `Content-Range` response header unless the response is partial.
+- Fixed CORS rendering for the S3 `HeadObject` operation.
+- Fixed a bug where no matching CORS configuration could result in a 403 response.
```
deadlypants1973:
```suggestion
- Fixed a bug where no matching CORS configuration could result in a `403` response.
```

deadlypants1973 (approved):

deadlypants1973 (approved):

Issue #5584, opened 2022-08-21
Cloudflare for Saas Workers origin


garretcharp:
### Which Cloudflare product does this pertain to?

Cloudflare for SaaS

### Existing documentation URL(s)

- https://developers.cloudflare.com/cloudflare-for-saas/workers-for-platforms/worker-as-origin/

### Section that requires update

Point #4

### What needs to change?

It currently states that if you want all traffic on your zone to go to the worker to set a route pattern as `*/*` going to the worker (this part is correct). However it says if you do not want that to set the route `*.<zonename>.com/*` to no worker. When in actuality you need to see both the `*/*` route to the worker and `*.<zonename>.com/*` to no worker otherwise all your customers domains would get error 522.

### How should it change?

Needs to give accurate info on how to setup the worker origin if you do not want everything on your zone to go to the saas handler worker. Had to debug this quite a bit in discord to get it to properly work because it was either getting error 522 or overriding all of my other workers previously and any other origins that might be setup that proxy through cloudflare.

kodster28:
Howdy @rita3ko, can you help with this?

rita3ko:
@tanushree-sharma @kabirsikand what's the right answer here?

rita3ko:
cc @dinasaur404

kabirsikand:
The updated guidance here is to use Custom Domains to route traffic on a zone to the Worker, by setting the custom domain as a fallback origin or custom origin for your custom hostnames (e.g. worker.example.com); this however will also route traffic on that domain name (worker.example.com) to the worker. In that regard it is different than both methods listed on the documentation now.

garretcharp:
Even when using custom domains you must still have a route of `*/*` otherwise you will get the following error (522):
![image](https://user-images.githubusercontent.com/32212214/186289357-667c8fb2-1365-41f3-904d-f9620395378d.png)

And if you dont want your entire zone to be taken over by the SASS worker you need to add 2 worker exclusions in your zone settings and the `*/*` route So you would need all of the following:

```
<zonename>.com/* -> no worker
*.<zonename>.com/* -> no worker
*/* -> sass worker
```

garretcharp:
@kabirsikand is there any update on this? People still get confused by this in the Cloudflare Dev Discord. Would be nice to update docs here to show how to set this up properly and to update that the worker routes page now calls it a "service" in the route selection menu.

henrhoi:
[...]

PR #5898, opened 2022-09-19
Replace tinyurl link with actual link

Cyb3r-Jak3:
There was a tinyurl link in the workers headers. I replaced it with the actual link. The tinyurl also points to the wrong location.

### DIFF
--- a/content/workers/runtime-apis/headers.md
+++ b/content/workers/runtime-apis/headers.md
@@ -77,7 +77,7 @@ The intended purpose of this header is to provide a means for recipients (for ex
 
 {{<Aside type="note">}}
 
-When configuring firewall rules, do not match on this header. Firewall rules are applied before Cloudflare adds the `CF-Worker` header. Instead, use the `cf.worker.upstream_zone`(https://tinyurl.com/2wx4senh) dynamic field, which contains the same value and exists for the same purpose.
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/headers.md
```diff
@@ -77,7 +77,7 @@ The intended purpose of this header is to provide a means for recipients (for ex
 
 {{<Aside type="note">}}
 
-When configuring firewall rules, do not match on this header. Firewall rules are applied before Cloudflare adds the `CF-Worker` header. Instead, use the `cf.worker.upstream_zone`(https://tinyurl.com/2wx4senh) dynamic field, which contains the same value and exists for the same purpose.
+When configuring firewall rules, do not match on this header. Firewall rules are applied before Cloudflare adds the `CF-Worker` header. Instead, use the `cf.worker.upstream_zone`(https://developers.cloudflare.com/ruleset-engine/rules-language/fields/#standard-fields) dynamic field, which contains the same value and exists for the same purpose.
```
kodster28:
```suggestion
When configuring firewall rules, do not match on this header. Firewall rules are applied before Cloudflare adds the `CF-Worker` header. Instead, use the [`cf.worker.upstream_zone`](/ruleset-engine/rules-language/fields/#standard-fields) dynamic field, which contains the same value and exists for the same purpose.
```

kodster28 (approved):

Dricecrushme (commented):
https://github.com/cloudflare/cloudflare-docs/issues/5584#issue-1345626157

PR #5900, opened 2022-09-19
R2: Fixing up redirects

codymgriffin:
To keep some previous links working, redirects have been added for the S3 and Runtime API references

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -558,6 +558,13 @@
 /stream/edit-manage-videos/manage-video-library/creator-id/ /stream/manage-video-library/creator-id/ 301
 /stream/edit-manage-videos/manage-video-library/searching/ /stream/manage-video-library/searching/ 301
 
+# r2
+/r2/platform/s3-compatibility/ /r2/data-access/s3-api/ 301
+/r2/platform/s3-compatibility/api/ /r2/data-access/s3-api/api/ 301
+/r2/platform/s3-compatibility/tokens/ /r2/data-access/s3-api/tokens/ 301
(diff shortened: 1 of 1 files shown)


mnewswanger-cloudflare (approved):

KianNH:
It might be best to use a dynamic redirect (under the appropriate section under the `# dynamic redirects`) heading as there's two other S3 docs that are high up on Google.

`/r2/platform/s3-compatibility/* /r2/data-access/s3-api/:splat`

codymgriffin:
@KianNH - good catch, but I had already broken out the specific links (its also nice that if we rename or change the path of docs under s3-api we'll be able to see it more obviously)

KianNH:
Hi @codymgriffin - fair enough, that makes sense.

I posted this on the Discord but I'll post it here too for visibility: 

`/r2/runtime-apis /r2/data-access/workers-api/workers-api-reference/ 301` appears to be missing the trailing slash from the source path. Google has `/r2/runtime-apis/` and that's what Pages would have generated.

```sh
➜  ~ curl https://production.cloudflare-docs-github.pages.dev/r2/runtime-apis -sSI | rg 'HTTP/'
HTTP/2 301
➜  ~ curl https://production.cloudflare-docs-github.pages.dev/r2/runtime-apis/ -sSI | rg 'HTTP/'
HTTP/2 404
```

<https://developers.cloudflare.com/r2/runtime-apis> and <https://developers.cloudflare.com/r2/runtime-apis/> both seem to still return a 404 oddly enough, would have expected `runtime-apis` to work but it'll be `runtime-apis/` that needs to work anyhow.

PR #5906, opened 2022-09-20
[R2] document deleting multiple R2 keys

Frederik-Baetens:


Frederik-Baetens:
Closing because an earlier open PR already documents this. Not 100% sure about the atomicity of the deleteObjects api anyways.

PR #5909, opened 2022-09-20
Fix linux support with disk encryption

kkrum:
Linux disk encryption checks have not been working consistently since day 1 so we are removing support for them for the time being


content/cloudflare-one/identity/devices/warp-client-checks/disk-encryption.md
```diff
@@ -69,3 +69,11 @@ All disks on the system must be encrypted for the posture check to pass.
 1. **Protection Status** must be set to **On**.
 
 All disks on the system must be encrypted for the posture check to pass.
+
+### On Linux
+
+Disk encryption is not supported on Linux at the moment
```
ranbel:
```suggestion
Disk encryption checks are not currently supported on Linux.
```

ranbel (approved):

PR #5952, opened 2022-09-22
Update migrate-legacy-tunnels.md

nrussell008:
I want to add instructions for customers to verify on their own if they still have legacy tunnels as we are receiving multiple tickets for this type of request for support to check for them.


ranbel:
merged in https://github.com/cloudflare/cloudflare-docs/pull/5967

PR #5957, opened 2022-09-23
[Docs] Fix highlight issues with indentation

pedrosousa:


KianNH:
[URL](https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/r2/data-access/workers-api/workers-api-reference/#r2listoptions)


- [x] Code block is formatted properly & indented with the list item.
- [x] MacOS - Safari
- [x] MacOS - Chrome
- [x] MacOS - Firefox
- [x] iOS - Safari

pedrosousa:
List of pages to test:

https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/r2/data-access/workers-api/workers-api-reference/#r2listoptions

https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/stream/uploading-videos/direct-creator-uploads/#step-1-generate-a-unique-one-time-upload-url

https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/cloudflare-one/tutorials/google-workspace-saas/#2-create-a-certificate-from-your-public-key

https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/pages/platform/custom-domains/#caa-records

https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/ssl/edge-certificates/custom-certificates/uploading/#update-a-certificate

https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/workers/examples/images-workers/

_Update:_ Also testing the page mentioned in #6431:
https://pedro-2022-09-23-fix-highlig.cloudflare-docs-7ou.pages.dev/workers/runtime-apis/streams/readablestream/

pedrosousa:
Hmm... we're losing the empty lines in that R2 code block. Let me investigate this a bit more (moving the PR back to draft).
Meanwhile, let's merge the other PR temporarily removing the code block indentation to make it readable.

KianNH:
> we're losing the empty lines in that R2 code block

I didn't even notice haha - but that makes sense.

I guess we need to look at replacing the blank lines with `<br>` in the highlight script too

pedrosousa:
Yes, I tested precisely that change and it worked. However, the "Copy to clipboard" button in every code block doubles every empty line -- was this already happening?

bin/highlight.ts
```diff
@@ -77,8 +77,10 @@ async function markdown(file: string): Promise<void> {
 
     let html = highlight(inner, lang);
 
+    html = html.replace(/[\r\n]/g, '').replace(/\{\{\</g, '{\\{<');
```
KianNH: > I guess we need to look at replacing the blank lines with `<br>` in the highlight script too

```suggestion
    html = html.replace(/[\r\n]/g, '<br>').replace(/\{\{\</g, '{\\{<');
```

This *seems* to work?

image

image

KianNH: > However, the "Copy to clipboard" button in every code block doubles every empty line -- was this already happening?

I haven't been able to replicate this on production or on my development server (w/ the `<br>` change) 🤔 

image

KianNH:
image

Looks good with the latest commit.

pedrosousa:
Still having issues in both Chrome and Firefox:
image

Can this be due to the different line endings in Windows (`\r\n` vs `\n`)...?

Well, I guess this issue could be there already for Windows folks.

KianNH:
I've spent an hour or so logging the output from the `\r\n` replace, comparing the outputted shortcode/HTML when you run `npm run prebuild`, LF & CRLF, this and that but I can never get the extra whitespace.

I'm currently on MacOS so I feel like I'll need to hop onto Windows to even be able to replicate it.

pedrosousa:
Just pushed a new commit adjusting copy to clipboard. It fixed the Windows issue (tested in Chrome and Firefox), which was also present in Chrome for Android (also fixed now). Can you check if it broke something for you on MacOS / iPhone?

If this can break something, I can remove this last change.

Context: Seems to be related to some [strange `innerText` behavior](http://perfectionkills.com/the-poor-misunderstood-innerText/#update), but it's strange that's not affecting every platform.

KianNH:
image

It looks fine in the documentation but copy to clipboard seems to have no newlines

KianNH:
It's weird - since the code is doing...

```js
function $clicktoClipboard(ev: MouseEvent) {
  const button = ev.target as HTMLElement;
  const pre = button.parentElement;
  if (pre) {
    const code = pre.getElementsByTagName("code")[0];
    const text = code.innerText;
    if (text) {
      try {
        //copy to clipboard
        navigator.clipboard.writeText(text);
        //change SVG
        button.innerHTML = `<svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 16 16" height="1em" width="1em" xmlns="http://www.w3.org/2000/svg"  style= "width:1rem; pointer-events: none;"  aria-label="Copied to clipboard button" focusable="true"><title>Copied Button</title><path d="M8.97 4.97a.75.75 0 0 1 1.07 1.05l-3.99 4.99a.75.75 0 0 1-1.08.02L2.324 8.384a.75.75 0 1 1 1.06-1.06l2.094 2.093L8.95 4.992a.252.252 0 0 1 .02-.022zm-.92 5.14.92.92a.75.75 0 0 0 1.079-.02l3.992-4.99a.75.75 0 1 0-1.091-1.028L9.477 9.417l-.485-.486-.943 1.179z"></path></svg>`;
        setTimeout(() => {
          button.innerHTML = `<svg stroke="currentColor" fill="currentColor" stroke-width="0" viewBox="0 0 24 24" height="1em" width="1em" xmlns="http://www.w3.org/2000/svg" style= "width:1rem; pointer-events: none;" aria-label="Copy to clipboard button" focusable="true"><title>Copy Button</title><path fill="none" d="M0 0h24v24H0z"></path><path d="M16 1H4c-1.1 0-2 .9-2 2v14h2V3h12V1zm3 4H8c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h11c1.1 0 2-.9 2-2V7c0-1.1-.9-2-2-2zm0 16H8V7h11v14z"></path></svg>`;
        }, 1500);
      } catch (err) {
        /* no support */
      }
    }
  }
}
```

Notably `code.innerText` - but if I replicate that in the console:

image

Edit: Ah, didn't notice your change to the clipboard code.

image

pedrosousa:
My `innerText` had the extra `\n`s already; yours does not. I just undid my previous commit.

KianNH:
Does `text.innerText.replace(/\r\n?|\n/g, '\n')` fix the double lines for you?

with `text` being `const text = document.querySelector('#main > article > div:nth-child(30) > ul > li:nth-child(5) > ul > li:nth-child(3) > pre > code')`

Edit: it does not (Windows, Firefox)

The issue is that there's a `<span class="CodeBlock--token-plain">` that contains just `<br>`

This doesn't do anything in the browser:

image

But does when copied:

image

If you remove that `<span class="CodeBlock--token-plain">`

image

MacOS generates them on LF:

image

but not on CRLF:

image

Windows generates them on LF:

image

but not on CRLF:

image

I guess CRLF is actually the one playing nicely.

My main confusion stems from find & replace `\r?\n` => `<br/>` results in the exact same output if you do it on a CRLF or LF example.

pedrosousa:
I was only able to test this now. As you said, no change -- I still get double new lines instead of single empty lines.
However, at least locally there's a slight improvement: I only get doubled new lines, instead of some additional _random_ new lines appearing where there was one.

I'll push the change to GitHub to see if at least this part improves in Cloudflare Pages' builds.

Edit1: Tried with your proposed standardization of line endings and without `<br>`'s but it didn't work. Trying now with the `<br>`'s still in place.
Edit2: It didn't work. Besides double empty lines, I get random empty lines inserted in the middle of the code. I'll revert my last commit.

kodster28:
seems like this could fix #6431

pedrosousa:
I'll update the branch and resolve the conflict.

I guess we can make it a proper PR and merge it even if it has a Windows-only issue — we didn't find a complete solution for it and the problem already exists (I just checked the live site). 

~Can someone double-check that we didn't break the copy behavior for Mac/Linux users, at least? It should keep the same newlines as in the code block being copied.~
**Update:** It seems that the solution to the indent issues I was proposing in this PR breaks the "Copy to clipboard" functionality on Macs (on Windows it was already half broken...). Until we find a proper solution, we should remove the indentation of code blocks with issues.

PR #5966, opened 2022-09-23
[Bots] Verified bots policy

patriciasantaana:


content/bots/reference/verified-bots-policy.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Bot Policy
+
+### Minimum Traffic
+
+A bot or proxy must have a minimum amount of traffic for Cloudflare to be able find it in the sampled data. 
+
+{{<Aside type="example" header="Example">}}
+
+The minimum traffic should have more than 1000 requests per day across multiple domains.
```
ranbel: Does the 1000 requests/day criteria apply to all bots? If so, then I would suggest labeling it as a requirement rather than an example.

content/bots/reference/verified-bots-policy.md
```diff
@@ -0,0 +1,119 @@
...
+
+{{<Aside type="example" header="Example">}}
+
+`GoogleBot/1.0` is a valid UA. 
+
+{{</Aside>}}
+
+### Domain Owner Consent
+
+Domains should only be crawled with the explicit or implicit consent of the zone's owner or terms of use. For search engines crawlers, it's obligatory to read the `robots.txt` to exclude paths to crawl from owner.
```
ranbel: I'm having some trouble understanding what this means: "it's obligatory to read the `robots.txt` to exclude paths to crawl from owner."

ranbel: Maybe rephrase as "Search engine crawlers must check `robots.txt` and exclude paths specified by the owner.

ranbel (approved):

PR #5979, opened 2022-09-26
Swap Verizon Edgecast IP addresses

wolruf:
IP addresses in Documentation should be compliant with https://datatracker.ietf.org/doc/html/rfc5737 I swapped 93.xxx addresses from Verizon Edgecast with RFC-compliant IP addresses


pedrosousa:
Hi @wolruf!
Thank you for your PR! In fact, we already started using the IP ranges reserved documentation purposes, but this page hadn't been updated yet.

pedrosousa (approved):

PR #5988, opened 2022-09-26
ChromeOS posture check OS version note

kkrum:
Clarify new behavior of Android client 6.17 and higher with our OS Version posture check works.

### DIFF
--- a/content/cloudflare-one/identity/devices/warp-client-checks/os-version.md
+++ b/content/cloudflare-one/identity/devices/warp-client-checks/os-version.md
@@ -71,3 +71,8 @@ Linux currently relies on the system Kernel version instead of a specific distro
 
 1. The valid SemVer would be the first 3 whole numbers of the output you obtain in the previous step. For instance, if the command above returned `5.14.0-25.el9.x86_64`, the valid SemVer would be `5.14.0`.
 
+
+### On ChromeOS
+
+On ChromeOS Zero Trust WARP services is served via our Android client which technically runs inside an Android VM on ChromeOS. WARP for Android version 6.16 and below repor the Android VM version as Android OS version. Starting with client 6.17 we introduced a distinct ChromeOS OS Version posture check and the Android client now only returns the actual Chrome OS version.
+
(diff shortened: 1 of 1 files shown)


content/cloudflare-one/identity/devices/warp-client-checks/os-version.md
```diff
@@ -71,3 +71,8 @@ Linux currently relies on the system Kernel version instead of a specific distro
 
 1. The valid SemVer would be the first 3 whole numbers of the output you obtain in the previous step. For instance, if the command above returned `5.14.0-25.el9.x86_64`, the valid SemVer would be `5.14.0`.
 
+
+### On ChromeOS
+
+On ChromeOS Zero Trust WARP services is served via our Android client which technically runs inside an Android VM on ChromeOS. WARP for Android version 6.16 and below repor the Android VM version as Android OS version. Starting with client 6.17 we introduced a distinct ChromeOS OS Version posture check and the Android client now only returns the actual Chrome OS version.
```
ranbel:
```suggestion
On Chromebooks, the WARP client runs as an Android application inside an Android VM. For the OS version check, WARP version 6.16 and below reports the Android VM version and not the ChromeOS version. Version 6.17 and above returns the actual ChromeOS version. 

To determine the ChromeOS version on your device, select the time and go to **Settings** > **About ChromeOS**.
```

ranbel: @kkrum How would a user determine the Android VM version?

kkrum: Go into Settings and About ChromeOS (just like you do on iOS / Android). We have instructions on Windows/Mac/Linux because those platforms make it slightly more difficult to find the right version number.

ranbel: I'm having a hard time visualizing the difference between the Android VM version and the ChromeOS version.  Doesn't Settings -> About ChromeOS get you the ChromeOS version?

kkrum: That was actually the problem, you are correct, figuring out how to get the Android VM version was tough (and not the right approach). Now we just rely on the ChromeOS version.

ranbel: I guess we can leave out Android VM instructions since it only applies to the old version of WARP.

kkrum: That sounds good, that section is only relevant if you care about the history of things which most folsk migh not

ranbel (approved):

PR #5993, opened 2022-09-26
[PoC] Dynamic learning paths

kodster28:


assets/App.vue
```diff
@@ -0,0 +1,11 @@
+export App = "<template>
```
kodster28: Need to clean up this file as it does nothing right now

assets/myapp.js
```diff
@@ -0,0 +1,84 @@
...
+    if (paths[item].path == location.pathname) {
+        current_path = paths[item];
+    }
+}
+
+let filteredElements = current_path.elements.filter (element => { return element.visible_by_default !== false})
+
+Vue.createApp({
+    methods: {
+        test() {
```
kodster28: Need to remove this function

bin/crawl.ts
```diff
@@ -4,27 +4,27 @@
  * - all images (<img>) do not have broken sources
```
kodster28, line 4: This was just from running the `format document` command.

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+
+let filteredElements = current_path.elements.filter(element => { return element.visible_by_default !== false})
+
+Vue.createApp({
+    methods: {
+        onRadioButtonChange() {
+            const selectedOptions = document.querySelectorAll('input[type=radio]:checked');
+            this.elements = current_path.elements.filter (element => {
+                let keepItem = true;
+                if (element.variables !== undefined) {
```
pedrosousa:
```suggestion
                if (element.variables) {
```

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+    let amended_path = location.pathname
+    if (!location.pathname.match(regex)) {
+        amended_path += '/'
+    }
+    if (paths[item].path === amended_path) {
+        current_path = paths[item];
+    }
+}
+
+let filteredElements = current_path.elements.filter(element => { return element.visible_by_default !== false})
```
pedrosousa:
```suggestion
let filteredElements = current_path.elements.filter(element => element.visible_by_default !== false)
```

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+// need to make sure the page exactly matches via ending slash (if not already there).
+
+let current_path;
+const regex = '\/$'
+
+for (const item in paths) {
+    let amended_path = location.pathname
+    if (!location.pathname.match(regex)) {
+        amended_path += '/'
+    }
```
pedrosousa:
```suggestion
for (const item in paths) {
    let amended_path = location.pathname;
    if (!location.pathname.endsWith('/')) {
        amended_path += '/';
    }
```
`endsWith` should have better performance.

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+    </div>`,
+    data() {return {
+        elements: filteredElements
+    }},
+    computed: {
+        onlyModules() {
+            return this.elements.filter(item => item.type === "module");
+        },
+        overallTimeEstimate() {
+            onlyTimeEstimates = this.elements.filter(item => item.estimated_time).map(item => item.estimated_time);
```
pedrosousa:
```suggestion
            let onlyTimeEstimates = this.elements.filter(item => item.estimated_time).map(item => item.estimated_time);
```

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+            let total_min = onlyTimeEstimates.reduce((previousValue, currentValue) => previousValue + currentValue,
```
pedrosousa: I'd use a single convention for variable names: either snake case (`total_min`) or camel case (`totalMin`).

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+                }
+                })
+            },
+        calculateModuleNumber(module) {
+            filteredModules = this.elements.filter(item => item.type === "module");
+            for (let element in filteredModules) {
+                if (module.title === filteredModules[element].title) {
+                    return (parseFloat(element) + 1).toString();
+                }
+            }
```
pedrosousa:
```suggestion
            let foundIdx = filteredModules.findIndex((elem, idx) => module.title === filteredModules[idx].title);
            return (foundIdx + 1).toString();
```
There are [a few quirks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in) when using `for...in` for array iteration -- I'd rather use some other control structure/function instead. [`findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) seems to work here.
Worst-case scenario: The function would return `"0"` when no index was found (`foundIdx => -1` / `(foundIdx + 1).toString() => "0"`).

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+        },
+    template: `
+    <div>
+        <p class="estimate">This learning path contains <span>[[ onlyModules.length ]] modules</span> and should take you around <span>[[ overallTimeEstimate ]]</span>.</p>
+    </div>
+    <div class="background">
+    <div v-for="(element, index) in elements" v-on:change="onRadioButtonChange">
+        <div class="learningPathModule" v-if="element.type === 'module'">
+        <div class="moduleHeader">
+            <h2 :id="element.title.toLowerCase().replaceAll(' ', '-')"><span class="DocsMarkdown--header-anchor-positioner">
```
pedrosousa: We could extract `stringVar.toLowerCase().replaceAll(' ', '-')` to a separate utility function since this code is used in a few places.
And then here we could do something like:

```
<h2 :id="slugify(element.title)">(...)
```

I probably don't know enough about Vue, but this _should_ work in an attribute?

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
+import { learning_paths as paths } from "json-collector";
```
pedrosousa, line 1: Before merging, run `prettier` on this file.

pedrosousa (approved):

assets/dynamic-learning-path-vue.js
```diff
@@ -0,0 +1,129 @@
...
+                        } 
+                    }
+                    return keepItem;
+                } else {
+                    return keepItem;
+                }
+                })
+            },
+        calculateModuleNumber(module) {
+            filteredModules = this.elements.filter(item => item.type === "module");
```
pedrosousa:
```suggestion
            let filteredModules = this.elements.filter(item => item.type === "module");
```

assets/dynamic-learning-path-vue.js
kodster28: Integrated in my local branch

assets/dynamic-learning-path-vue.js
kodster28: Added to my local branch

assets/dynamic-learning-path-vue.js
kodster28: Added to my local

assets/dynamic-learning-path-vue.js
kodster28: Added to my local

assets/dynamic-learning-path-vue.js
kodster28: Added to my local

assets/dynamic-learning-path-vue.js
kodster28: Added in my local branch. And yes, it _totally_ works like that!

assets/dynamic-learning-path-vue.js
kodster28: Addressed in changes that I just pushed

PR #5994, opened 2022-09-26
Update bfd.md

ajtaylor6:
Article incorrectly described BFD as "Bidirectional Forwarding Direction" instead of Bidirectional Forwarding Detection (https://www.rfc-editor.org/rfc/rfc5881) The direction is bidirectional after all.


dcpena:
Going to mark this pull request as closed because this fix was implemented in PR https://github.com/cloudflare/cloudflare-docs/pull/6222.

PR #6008, opened 2022-09-27
[R2] demo worker example fix

deadlypants1973:
closes #5516

### DIFF
--- a/content/r2/api/workers/demo-worker.md
+++ b/content/r2/api/workers/demo-worker.md
@@ -12,23 +12,6 @@ interface Env {
   MY_BUCKET: R2Bucket
 }
 
-function parseRange(encoded: string | null): undefined | { offset: number, end: number, length: number } {
-  if (encoded === null) {
-    return
-  }
-
-  const parts = encoded.split("bytes=")[1]?.split("-") ?? []
-  if (parts.length !== 2) {
-    throw new Error('Not supported to skip specifying the beginning/ending byte at this time')
-  }
-
-  return {
-    offset: Number(parts[0]),
-    end:    Number(parts[1]),
-    length: Number(parts[1]) + 1 - Number(parts[0]),
-  }
-}
-
 function objectNotFound(objectName: string): Response {
   return new Response(`<html><body>R2 object "<b>${objectName}</b>" not found</body></html>`, {
     status: 404,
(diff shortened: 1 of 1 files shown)


ranbel (approved):

content/r2/examples/demo-worker.md
```diff
@@ -67,9 +50,8 @@ export default {
       }
 
       if (request.method === 'GET') {
-        const range = parseRange(request.headers.get('range'))
         const object = await env.MY_BUCKET.get(objectName, {
-          range,
+          range: request.headers,
```
KianNH: This should be sound, but I can't get it to work in any variety without throwing an internal error. 

Spoken to Harshal who is going to see if he can figure out why it's throwing that.

content/r2/examples/demo-worker.md
```diff
@@ -80,10 +62,10 @@ export default {
         const headers = new Headers()
         object.writeHttpMetadata(headers)
         headers.set('etag', object.httpEtag)
-        if (range) {
-          headers.set("content-range", `bytes ${range.offset}-${range.end}/${object.size}`)
+        if (object.range) {
```
KianNH: `object.range` will exist even on a non-ranged request so we'll likely have to continue checking for the existence of a `range` header on the request.

content/r2/examples/demo-worker.md
```diff
@@ -80,10 +62,10 @@ export default {
...
+          headers.set("content-range", `bytes ${object.range.offset}-${object.range.end}/${object.size}`)
         }
-        const status = object.body ? (range ? 206 : 200) : 304
+        const status = object.body ? (object.range ? 206 : 200) : 304
```
KianNH: Same as above - but one question is on the semantics of 206.

Do we always return 206 in response to a ranged request, even if we return the full content? Will need to do some RFC/MDN reading.

vlovich: From my reading, yes. MDN indicates 206 is returned always even if you end up satisfying the entire file.

content/r2/examples/demo-worker.md
Frederik-Baetens: This should be fixed now. Let me know if you need anything else from me to unblock merging this PR.

PR #6026, opened 2022-09-28
Add note that multiple keys are allowd in ca.pub

rhodesn:
This is already detailed in the sshd_config docs: https://man.openbsd.org/sshd_config#TrustedUserCAKeys, however considering the Cloudflare docs are seemingly meant to be read standalone, it would make sense to note that multiple keys are allowed here.


ranbel:
will merge in #6045

PR #6040, opened 2022-09-28
[Queues] Update estimated bill formula on pricing.md

ChrisBoik:
The formula provided did not take into account the 'per million' cost


Electroid (approved):
Thanks for catching this!

PR #6049, opened 2022-09-28
Update Boolean

sdayman:
CSV upload will return a JSON error if 1s and 0s are used.


pedrosousa:
Hi @sdayman 

I just tested it and you're right, it doesn't work — but it should. :)
I filled a bug internally and I'll remove the 0/1 from the documentation for now as you suggest. I'll add it back once the issue is solved.

pedrosousa (approved):

PR #6056, opened 2022-09-29
Remove API only mention ?

wolruf:
Unless I'm missing something, the feature can be configured using the Cloudflare Dashboard. API is still available but it's not "required" as the only way to configure Custom Error Response. Right ?

### DIFF
--- a/content/rules/custom-error-responses/_index.md
+++ b/content/rules/custom-error-responses/_index.md
@@ -17,8 +17,6 @@ To configure a custom error response, create a custom error response rule at the
 
 During the beta, you can define custom error responses using inline templates and specify the response's content type and HTTP status code.
 
-Additionally, at this stage you can only create custom error response rules [using the API](/rules/custom-error-responses/create-api/).
-
 {{</Aside>}}
 
 ## How it works


pedrosousa:
Hi @wolruf

I just checked and the feature is still API only. Are you referring to the error pages you can configure via [Custom Pages](https://support.cloudflare.com/hc/en-us/articles/200172706-Configuring-Custom-Pages-Error-and-Challenge-)?

Custom Pages is a different feature, less powerful than custom error responses -- it only depends on the status code returned by the origin server, and you cannot set up an expression to configure an error response for a subset of your traffic.

As stated in the custom error responses documentation, if you configure both, custom error responses will override Custom Pages.

wolruf:
Hi @pedrosousa I'm referring to custom error response (beta), that, I believe, is what is seen in WAF Custom Rule configuration
Screenshot 2022-09-29 at 14 20 03

pedrosousa:
OK, I understand. That's also a different feature.

The custom response for _Block_ actions (in your screenshot) allows you to define a response to send the visitor when there's a match for a rule configured with the _Block_ action. In this situation, the origin server is not contacted — Cloudflare determined that, based on the rule configuration, a given request should be blocked. You can configure the response content and status code sent to the website visitor for this particular case.

On the other hand, custom error response rules (currently API only) allow you to customize what is sent to the visitor when both A) the origin server (or a Cloudflare product like Workers) returns an error status code, and B) there is a match for the rule expression. In this case, we'll be adjusting the final response sent to the visitor because there was an error returned elsewhere, not due to a blocked request.

wolruf:
Hi @pedrosousa where can we document that a custom response for Block actions is a great way to have a maintenance page when such a maintenance is happening on the Origin server (in lieu of using the Origin server) ?

pedrosousa:
Let's track this internally first — I need to check with PMs what's our best recommendation for this use case. 
I created PCX-5140 for this purpose.

PR #6060, opened 2022-09-29
Update www-redirect.md

rquinlivan:
Went through these instructions today and noticed that the 'Bulk Redirects' menu item is under 'Rules'. It wasn't apparent from the instructions where to find it.

### DIFF
--- a/content/pages/how-to/www-redirect.md
+++ b/content/pages/how-to/www-redirect.md
@@ -15,7 +15,7 @@ To do this:
 
 ![Adding a www record in the Cloudflare DNS dashboard](/pages/how-to/media/www_subdomain.png)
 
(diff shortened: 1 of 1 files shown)


pedrosousa:
Actually, that's a shortcut available at the zone level (for convenience) that will navigate you to the account-level Bulk Redirects page, available at **Account Home** > **Bulk Redirects**.

kodster28 (approved):

kodster28:
In this case, @pedrosousa, think it makes sense to navigate them to the shortcut since they're already w/in the context of a specific zone w/ adding the DNS record.

pedrosousa:
Looks good! 
I was just adding a heads-up that it's an account-level configuration, even though there's a shortcut at the zone level to get to the correct dashboard page faster.

rquinlivan:
Thanks :)

PR #6094, opened 2022-10-03
Analytics Engine function additions

46bit:
In `analytics-engine-api` PR #90 I'm adding some new SQL functions (`min` and `max`) and adding support for the `DISTINCT` keyword to `count`, `avg` and `sum`. This PR updates the documentation with information on those new features.

This PR also quick-fixes a page that still listed us as being in private beta.

~⚠️ Don't merge this until the accompanying private PR is merged~ **Ready to go live**

### DIFF
--- a/content/analytics/analytics-engine/sql-reference.md
+++ b/content/analytics/analytics-engine/sql-reference.md
@@ -296,7 +296,7 @@ Usage:
 SUM([DISTINCT] column_name)
 ```
 
-Sum is an aggregation function that returns the sum of column values across all rows in each group or results set. Sum also supports `DISTINCT`, and then it will only sum the unique values in the column.
+Sum is an aggregation function that returns the sum of column values across all rows in each group or results set. Sum also supports `DISTINCT`, but in this case it will only sum the unique values in the column.
 
 Example:
 ```SQL
--- a/content/analytics/analytics-engine/sql-reference.md
+++ b/content/analytics/analytics-engine/sql-reference.md
@@ -284,7 +284,7 @@ Count can also be used to count the number of distinct (unique) values in each c
 Example:
 ```SQL
 -- return the total number of rows
-COUNT(item_cost)
+COUNT()
 -- return the number of unique item costs
 COUNT(DISTINCT item_cost)
 ```
--- a/content/analytics/analytics-engine/sql-reference.md
+++ b/content/analytics/analytics-engine/sql-reference.md
@@ -323,6 +323,36 @@ AVG(item_cost)
 AVG(DISTINCT item_cost)
 ```
 
+### MIN
+
+Usage:
+```SQL
+MIN(column_name)
+```
+
+Min is an aggregation function that returns the minimum value of a column across all rows.
+
+Example:
+```SQL
+-- return the minimum item cost
+MIN(item_cost)
+```
+
+### MAX
+
+Usage:
+```SQL
+MAX(column_name)
+```
+
+Max is an aggregation function that returns the maximum value of a column across all rows.
+
+Example:
+```SQL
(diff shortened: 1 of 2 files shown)


content/analytics/analytics-engine/sql-reference.md
```diff
@@ -274,38 +274,83 @@ column 1 value  column 2 value
...
 ### SUM
 
 Usage:
 ```SQL
-SUM(column_name)
+SUM([DISTINCT] column_name)
 ```
 
-Sum is an aggregation function that returns the sum of column values across all rows in each group or results set.
+Sum is an aggregation function that returns the sum of column values across all rows in each group or results set. Sum also supports `DISTINCT`, and then it will only sum the unique values in the column.
```
angelampcosta:
```suggestion
Sum is an aggregation function that returns the sum of column values across all rows in each group or results set. Sum also supports `DISTINCT`, but in this case it will only sum the unique values in the column.
```

content/analytics/analytics-engine/sql-reference.md
```diff
@@ -274,38 +274,83 @@ column 1 value  column 2 value
...
 ### AVG
 
 Usage:
 ```SQL
-AVG(column_name)
+AVG([DISTINCT] column_name)
 ```
 
-Avg is an aggregation function that returns the mean of column values across all rows in each group or results set.
+Avg is an aggregation function that returns the mean of column values across all rows in each group or results set. Avg also supports `DISTINCT`, and then it will only average the unique values in the column.
```
angelampcosta:
```suggestion
Avg is an aggregation function that returns the mean of column values across all rows in each group or results set. Avg also supports `DISTINCT`, but in this case it will only average the unique values in the column.
```

angelampcosta (approved):

content/analytics/analytics-engine/sql-reference.md
```diff
@@ -274,38 +274,83 @@ column 1 value  column 2 value
...
 ```
 
 Count is an aggregation function that returns the number of rows in each group or results set.
 
+Count can also be used to count the number of distinct (unique) values in each column:
+
+Example:
+```SQL
+-- return the total number of rows
+COUNT(item_cost)
```
jimhawkridge: The usage suggest the options are empty brackets, or DISTINCT?
```suggestion
COUNT()
```

content/analytics/analytics-engine/sql-reference.md
```diff
@@ -274,38 +274,83 @@ column 1 value  column 2 value
...
+-- return the number of unique item costs
```
jimhawkridge: How about `COUNT(DISTINCT product_variety)` (counts the number of distinct product types). Or something.
I think counting distinct costs could be a confusing/unlikely thing to want to do.

jimhawkridge (changes requested):
I've requested one change (to the `COUNT()`).
I've also suggested a change to the example to make it more "real world". I'm stuggling to think of good examples for AVG and SUM with DISTINCT though. I'll think on that, but either way the current examples are fine really - I'm just being picky.

content/analytics/analytics-engine/sql-reference.md
46bit: I've run queries like that before but yeah it's hard to describe a usecase briefly. I'll go with your idea 👍🏻

46bit: Pushed up a commit that should help a bit

jimhawkridge (approved):

46bit:
@angelampcosta @jimhawkridge The changes this PR documents are live now. I don't have merge permissions for this PR, could either of you merge? Thanks :)

PR #6113, opened 2022-10-04
Turnstile: remove 'hostname' from error response

punkeel:
We won't include this field in non-successful responses (as it may not be known).


punkeel:
cc @migueldemoura @worenga @patriciasantaana

WalshyDev (approved):

migueldemoura (approved):

PR #6118, opened 2022-10-04
Proposal to guide users to use Managed Challenge

wolruf:
With both Managed Challenge and JS challenge, customers can hesitate still using JS challenge. As Managed Challenge is more feature-rich, my proposal is to give more guidance to use Managed Challenge and avoid JS challenge unless for specific reasons. Hence, wording isn't perfect yet.


kodster28:
Addressing this change in a separate PR

PR #6126, opened 2022-10-04
[Turnstile] Removed domain example

isaac-mcfadyen:
- In the Turnstile docs, several domains are provided as examples for how Domain Management works.
- One of the example domains was previously unregistered, has now been registered (not by Cloudflare) and now redirects to a Discord link not associated with Cloudflare.
- This PR removes this domain from the docs, as a precaution in case someone pastes the link into an address bar or otherwise visits the domain and is redirected away from what they are looking for.

On an unrelated note, I'm questioning the `:8080` part on the second example domain as the text in the paragraph above says a port shouldn't be specified?

### DIFF
--- a/content/turnstile/get-started/domain-management.md
+++ b/content/turnstile/get-started/domain-management.md
@@ -21,7 +21,6 @@ For example, using the `www.example.com` value will allow widgets on the followi
 but not on the following domains:
 * `example.com`
 * `dash.example.com`
-* `attacker-website.cf`
 * `not-example.com`
 
 When the widget is embedded on a domain not listed, it will show an error message.
--- a/content/turnstile/get-started/domain-management.md
+++ b/content/turnstile/get-started/domain-management.md
@@ -21,7 +21,7 @@ For example, using the `www.example.com` value will allow widgets on the followi
 but not on the following domains:
 * `example.com`
 * `dash.example.com`
-* `not-example.com`
+* `workerd.dev`
 
 When the widget is embedded on a domain not listed, it will show an error message.
 
--- a/content/turnstile/get-started/domain-management.md
+++ b/content/turnstile/get-started/domain-management.md
@@ -21,7 +21,7 @@ For example, using the `www.example.com` value will allow widgets on the followi
 but not on the following domains:
 * `example.com`
 * `dash.example.com`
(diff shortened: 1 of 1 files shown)


content/turnstile/get-started/domain-management.md
```diff
@@ -21,7 +21,6 @@ For example, using the `www.example.com` value will allow widgets on the followi
 but not on the following domains:
 * `example.com`
 * `dash.example.com`
-* `attacker-website.cf`
 * `not-example.com`
```
WalshyDev: Let's also remove this or point it to one we own

isaac-mcfadyen: Resolved, replaced with a Cloudflare controlled domain. 🙂

WalshyDev (approved):

migueldemoura:
> On an unrelated note, I'm questioning the :8080 part on the second example domain as the text in the paragraph above says a port shouldn't be specified?

The host with a port is mentioned as an example of an origin that would be allowed, not as an example of what to provide in the domain list for the widget.

content/turnstile/get-started/domain-management.md
```diff
@@ -21,13 +21,12 @@ For example, using the `www.example.com` value will allow widgets on the followi
 but not on the following domains:
 * `example.com`
 * `dash.example.com`
-* `attacker-website.cf`
-* `not-example.com`
+* `workerd.dev`
```
migueldemoura: Can we use `cloudflare.com` instead?

isaac-mcfadyen:
Oh OK, makes sense. I wonder if we should remove that/clarify that? When I first read it I read it as:
- no port should be specified
- but the example user specifies port

content/turnstile/get-started/domain-management.md
isaac-mcfadyen: Edited! 👍

migueldemoura (approved):

migueldemoura:
> Oh OK, makes sense. I wonder if we should remove that/clarify that? When I first read it I read it as:
> - no port should be specified
> - but the example user specifies port

Maybe we can follow-up on this in another PR, cc @patriciasantaana.

PR #6129, opened 2022-10-04
[Docs Site] clipboard: do not copy unselectable elements from codeblocks.

### LINKED ISSUE #5474: Copy btn in code snippets will copy the `$` sign at the beginning of a command line
### Example URL(s)

https://developers.cloudflare.com/workers/wrangler/get-started/
https://developers.cloudflare.com/workers/wrangler/commands/

### Actual Behavior

Copying code snippets such as 
Screenshot 2022-08-09 at 13 42 29
will paste the command *including* the `$` at the beginning of the command line. If pasted in the terminal, users will have to first remove that character in order to run the command

### Expected Behavior

It would be nice if the `$` sign is trimmed, so that users can run the copied command directly in the terminal

### Additional information

_No response_

themetanull:
The codeblocks seem to tag tokens that should not be copied with the class name`CodeBlock--token-unselectable`. We use that to cloneNode() the code element and remove children with that class name, before getting the innerText.

This safely removes the leading `$` in languages like bash, while keeping it in other languages like JS. Other languages should work as expected according to the syntax definitions.

Fixes #5474

### DIFF
--- a/assets/events.ts
+++ b/assets/events.ts
@@ -92,19 +92,28 @@ function $clicktoClipboard(ev: MouseEvent) {
   const button = ev.target as HTMLElement;
   const pre = button.parentElement;
   if (pre) {
-    const code = pre.getElementsByTagName("code")[0];
-    const text = code.innerText;
-    if (text) {
-      try {
-        //copy to clipboard
-        navigator.clipboard.writeText(text);
-        //change SVG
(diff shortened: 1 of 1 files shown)


themetanull:
Is there a way to generate the pages preview. I see that on other pull requests.

kristianfreeman:
Hey @themetanull - unfortunately we don't support deploy preview PRs on forks right now, but I def see why this is something we can and should get merged. Can you work up a quick video demo-ing it real quick? Or even some screenshots? It will probably be faster than any other way of reviewing :D

themetanull:
Added `deep: true` to the cloneNode(). 
Here is a screen recording :) 

https://user-images.githubusercontent.com/94995512/195790611-d7d32ffb-135c-427c-ae88-fb18811f3a07.mov

kristianfreeman (approved):

PR #6134, opened 2022-10-05
cloudflare-tunnels-faq.md: transport-log -> transport-loglevel

tobiaspal:
It's slightly confusing that cloudflare-tunnels-faq.md mentions the configuration option `transport-log`, but the correct option is `transport-loglevel`.


ranbel (approved):
Thanks for the fix!

PR #6135, opened 2022-10-05
Update require-gateway.md

jason-schell:
Clarified requirements for using the "require Gateway"-policy.


ranbel:
Thanks for bringing this to our attention! You definitely need Proxy enabled, however you don't need the cert or tls decryption. We're updating the docs in #https://github.com/cloudflare/cloudflare-docs/pull/6147.

PR #6153, opened 2022-10-06
[Workers] add DigestStream to WebCrypto Docs

helloimalastair:
Addresses [2021-12-10 Changelog](https://github.com/cloudflare/cloudflare-docs/commit/52f31fbd721fdff4195bb1122264048238cb1d38), `crypto.DigestStream` entry.

Added in [Methods](https://webcrypto-digeststream-docs.pages.dev/workers/runtime-apis/web-crypto/#methods) section.

Tab switcher is placed outside of `{{</definitions>}}` on purpose, as placing it within results in HTML being rendered within the code block.

May need to be corrected to match Styleguide.


content/workers/runtime-apis/web-crypto.md
```diff
@@ -52,7 +52,53 @@ The Web Crypto API differs significantly from Node’s Crypto API. If you want t
 
   - Generates a new random (version 4) UUID as defined in [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122.txt).
 
+- {{<code>}}new crypto.DigestStream(algorithm){{</code>}} {{<type>}}DigestStream{{</type>}}
+
+  - A non-standard extension to the `crypto` API that supports generating a hash digest from streaming data. The `DigestStream` itself is a [`WritableStream`](/workers/runtime-apis/streams/writablestream/) that does not retain the data written into it; instead, it generates a digest hash automatically when the flow of data has ended.
```
deadlypants1973:
```suggestion
  - A non-standard extension to the `crypto` API that supports generating a hash digest from streaming data. The `DigestStream` itself is a [`WritableStream`](/workers/runtime-apis/streams/writablestream/) that does not retain the data written into it. Instead, it generates a digest hash automatically when the flow of data has ended.
```

deadlypants1973 (commented):
PCX review

content/workers/runtime-apis/web-crypto.md
harrishancock:
```suggestion
  - A non-standard extension to the `crypto` API that supports generating a hash digest from streaming data. The `DigestStream` itself is a [`WritableStream`](/workers/runtime-apis/streams/writablestream/) that does not retain the data written into it; instead, it generates a hash digest automatically when the flow of data has ended.
```

content/workers/runtime-apis/web-crypto.md
```diff
@@ -52,7 +52,53 @@ The Web Crypto API differs significantly from Node’s Crypto API. If you want t
...
+{{<tabs labels="js/esm | ts/esm">}}
+{{<tab label="js/esm" default="true">}}
+```js
+export default {
+  async fetch(req) {
+    const res = await fetch(req);
+    const digestStream = new crypto.DigestStream("SHA-256");
+    res.clone().body.pipeTo(digestStream);
+    const digest = await digestStream.digest;
+    res.headers.set("x-content-digest", `SHA-256=${digest}`);
```
harrishancock: We need to hex-encode the digest, otherwise `digest` will be stringified as `[object ArrayBuffer]` or something like that. Also need to make the headers mutable by recreating the response. Example: https://cloudflareworkers.com/#b63ef59864953fc07d4f61f9f457fc9c:https://tutorial.cloudflareworkers.com

content/workers/runtime-apis/web-crypto.md
```diff
@@ -52,7 +52,53 @@ The Web Crypto API differs significantly from Node’s Crypto API. If you want t
...
+    return res;
+  }
+}
```
WalshyDev: Fixed the code and added comments

```suggestion
export default {
  async fetch(req) {
    // Fetch from origin
    const res = await fetch(req);
    
    // We need to read the body twice so we `tee` it (get two instances)
    const [bodyOne, bodyTwo] = res.body.tee();

    // Make a new response so we can set the headers (responses from `fetch` are immutable)
    const newRes = new Response(bodyOne, res);

    // Create a SHA-256 digest stream and pipe the body into it
    const digestStream = new crypto.DigestStream("SHA-256");
    bodyTwo.pipeTo(digestStream);
    // Get the final result
    const digest = await digestStream.digest;

    // Turn it into a hex string
    const hexString = [...new Uint8Array(digest)]
      .map(b => b.toString(16).padStart(2, '0'))
      .join('')

    // Set a header with the SHA-256 hash and return the response
    newRes.headers.set("x-content-digest", `SHA-256=${hexString}`);
    return newRes;
  }
}
```

content/workers/runtime-apis/web-crypto.md
```diff
@@ -52,7 +52,53 @@ The Web Crypto API differs significantly from Node’s Crypto API. If you want t
...
+      res.clone().body?.pipeTo(digestStream);
+      const digest = await digestStream.digest;
+      res.headers.set("x-content-digest", `SHA-256=${digest}`);
+    } else {
+      res.headers.set("x-content-digest", "SHA-256=empty");
+    }
+    return res;
+  }
+}
+export default handler;
```
WalshyDev:
```suggestion
const handler: ExportedHandler = {
  async fetch(req) {
    // Fetch from origin
    const res = await fetch(req);
    
    // We need to read the body twice so we `tee` it (get two instances)
    const [bodyOne, bodyTwo] = res.body!.tee();

    // Make a new response so we can set the headers (responses from `fetch` are immutable)
    const newRes = new Response(bodyOne, res);

    // Create a SHA-256 digest stream and pipe the body into it
    const digestStream = new crypto.DigestStream("SHA-256");
    bodyTwo.pipeTo(digestStream);
    // Get the final result
    const digest = await digestStream.digest;

    // Turn it into a hex string
    const hexString = [...new Uint8Array(digest)]
      .map(b => b.toString(16).padStart(2, '0'))
      .join('')

    // Set a header with the SHA-256 hash and return the response
    newRes.headers.set("x-content-digest", `SHA-256=${hexString}`);
    return newRes;
  }
}
export default handler;
```

WalshyDev (changes requested):

helloimalastair:
PCX Revisions were applied in the previous commit. Changes are live at [preview](https://webcrypto-digeststream-docs.pages.dev/workers/runtime-apis/web-crypto/).

content/workers/runtime-apis/web-crypto.md
```diff
@@ -52,7 +52,75 @@ The Web Crypto API differs significantly from Node’s Crypto API. If you want t
 
   - Generates a new random (version 4) UUID as defined in [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122.txt).
 
+- {{<code>}}new crypto.DigestStream(algorithm){{</code>}} {{<type>}}DigestStream{{</type>}}
```
harrishancock: Looking at the preview, it looks a little weird to me to put this under "Methods". Would putting this in a separate section called "Constructors" or "Nested Types, and dropping the `new`, make sense?

harrishancock (approved):
Left one more comment to think about, but it looks good to me.

content/workers/runtime-apis/web-crypto.md
helloimalastair: Sure, that sounds good. I looked over the docs for a bit, trying to find a similar example. The closest I found was the constructor for `HTMLRewriter`, but that didn't seem right, so I wasn't sure what to do with it.

helloimalastair:
Moved `DigestStream` into the `Constructors` block, above `Methods`. Change visible on [preview](https://webcrypto-digeststream-docs.pages.dev/workers/runtime-apis/web-crypto/#constructors).

WalshyDev (approved):

deadlypants1973 (approved):

PR #6161, opened 2022-10-06
The default method is `GET` not `POST`

tavoyne:
The default method is `GET` not `POST`. The values are serialized and written as URL params.


deadlypants1973 (approved):

deadlypants1973:
> The default method is `GET` not `POST`. The values are serialized and written as URL params.

Superb catch, thank you.

PR #6180, opened 2022-10-07
fix: update example port to number instead of string

pyk:
Using an example throw the following error:

```bash
[exchange:dev:*server] ✘ [ERROR] Processing wrangler.toml configuration:
[exchange:dev:*server]
[exchange:dev:*server]     - Expected "dev.port" to be of type number but got "3000".
```


deadlypants1973:
fixed by #6203

Issue #5348, opened 2022-07-28
[Docs Site] Add link headers for 103 Early Hints


KianNH:
### Example URL(s)

- https://developers.cloudflare.com/

### Actual Behavior

No `link` headers.

### Expected Behavior

All the `link` headers.

### Additional information

cc @kodster28 @WalshyDev 

Now that we have [Early Hints enabled on the docs site](https://github.com/cloudflare/cloudflare-docs/pull/5280#issuecomment-1198554066), we can leverage 103 Early Hints to make the site super fast 🔥

## Preloads

In my opinion, this would include:

- https://developers.cloudflare.com/home.css
- https://developers.cloudflare.com/search.redirect.min.826764246c0056e88cd0af076274b8d9.js
- https://developers.cloudflare.com/home.min.d720e27e667103421a4ba0e7d5a9c1bb.js

![image](https://user-images.githubusercontent.com/94662631/181628911-7a84e017-be04-48fb-8cbf-c960078173a6.png)

I haven't actually looked into what the JS does - or what any of the third party stuff does.

Since the JS is cache-busted with a hash, we'd need a post-build step to add them to the `_headers` file.

## Preconnects

Important third party domains hosting assets:

- www.googletagmanager.com

## Suggestions

Happy to hear any thoughts about if anything needs adding or removing - obviously with preloads, we want to keep it to the critical render path otherwise you'll defeat the benefit of stuff like lazy loading or deferring.

KianNH:
On another note, is the code block highlighting added through GTM? My fork doesn't have it and it looks ugly - but have never looked into why

kodster28:
If you're talking what I think you're talking about, you can preview code block highlighting by running `npm run prebuild` prior to `npm run dev` (and then run `npm run postbuild` afterwards to strip out all the highlighting).

kodster28:
Also, in terms of Early Hints, heard that we might actually get the *Smart Early Hints* mentioned by [the blog](https://blog.cloudflare.com/early-hints/) relatively soon, so the links might get added automatically.

KianNH:
> you can preview code block highlighting by running `npm run prebuild` prior to `npm run dev`

![image](https://user-images.githubusercontent.com/94662631/181769193-a2f150d6-c816-4291-9790-0f746d6343bf.png)

Ooooooooooooooooooooooo - thanks.

[...]

PR #6183, opened 2022-10-07
Testing adding rel preload headers

kodster28:
Addresses #5348 .

From looking at [this thread](https://stackoverflow.com/questions/46787921/rel-preload-for-stylesheet-isnt-applying-the-styles-once-downloaded)... seems acceptable to have two `rel` attributes to actually apply the spreadsheet.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -3,7 +3,7 @@
   <head>
     {{- partial "head.meta" (dict "Context" . "Product" .Section) -}}
 
-    <link rel="stylesheet" href="/style.css"/>
+    <link rel="stylesheet" rel="preload" as="style" href="/style.css" />
     {{- block "styles" . -}}{{- end -}}
 
     {{- partial "script" (dict "src" "theme.ts" "inline" true "format" "iife") -}}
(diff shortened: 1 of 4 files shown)


kodster28:
@GregBrimble, not sure if you had any thoughts on the implementation? i.e., which of the approaches works best (2 `rel` links, two stylesheet declarations, having a `onload="this.rel='stylesheet'"` property... or maybe something I'm not thinking of.

Just swapping to the `rel=preload as="style"` doesn't apply the stylesheet to the page.

kodster28:
Definitely does speed up the homepage load...

Original
Screen Shot 2022-10-07 at 11 53 51 AM

Preload headers
Screen Shot 2022-10-07 at 11 53 43 AM

Preconnect headers
Screen Shot 2022-10-10 at 9 10 32 AM

GregBrimble:
Nifty! Adding two `rel` values together in the same `<link>` element isn't currently supported for [the recently announced automatic Pages and Early Hints integration](https://blog.cloudflare.com/early-hints-on-cloudflare-pages/), but it probably could be.

I think basically we need to change [this line](https://github.com/cloudflare/wrangler2/blob/main/packages/pages-shared/asset-server/handler.ts#L328) to be:

```diff
- .on("link[rel=preconnect],link[rel=preload]", {
+ .on("link[rel~=preconnect],link[rel~=preload]", {
```

GregBrimble:
Looks like HTMLRewriter is actually a bit picky here (you could possibly call this a bug with HTMLRewriter?):

https://cloudflareworkers.com/#ef343d99a11b6bf64a7b684577c62427:https://tutorial.cloudflareworkers.com/

The attribute order changes how these selectors work. In summary:

- `rel="stylesheet preconnect"` ✅
- `rel="preconnect stylesheet"` ✅
- `rel="preconnect" rel="stylesheet"` ✅
- `rel="stylesheet" rel="preconnect" ` ❌

GregBrimble:
image

[W3C](https://validator.w3.org/nu/) doesn't like two of the same attributes, but it's fine with one attribute with two whitespace-separated values! So I'd say you should probably change to that, and we'll look at making that change to the Early Hints stuff :)

pedrosousa (approved):

PR #6185, opened 2022-10-07
workers: remove deprecated experimental-enable-local-persistence flag in commands.md

justinnoel:
The flag `--experimental-enable-local-persistence` is deprecated and no longer available.

When attempting to use this flag, wrangler says:

```
🚧 'wrangler pages <command>' is a beta command. Please report any issues to https://github.com/cloudflare/wrangler2/issues/new/choose
▲ [WARNING] --experimental-enable-local-persistence is deprecated.

  Move any existing data to .wrangler/state and use --persist, or
  use --persist-to=./wrangler-local-state to keep using the old path.
```

### DIFF
--- a/content/workers/wrangler/commands.md
+++ b/content/workers/wrangler/commands.md
@@ -154,13 +154,6 @@ None of the options for this command are required. Many of these options can be 
 - `--local` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: false){{</prop-meta>}}
 
   - Run the preview of the Worker directly on your local machine.
-
-    {{<Aside type="warning">}}
-
-This runs an ephemeral local version of your Worker, and will not be able to access data stored on Cloudflare's Edge (for instance, this includes your data stored on KV). If you'd like to persist data locally, the experimental option `--experimental-enable-local-persistence` will store data in the `wrangler-local-state` subdirectory.
-
-{{</Aside>}}
-
 - `--experimental-local` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: false){{</prop-meta>}}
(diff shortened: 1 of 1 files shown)


content/workers/wrangler/commands.md
```diff
@@ -154,13 +154,6 @@ None of the options for this command are required. Many of these options can be
 - `--local` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: false){{</prop-meta>}}
 
   - Run the preview of the Worker directly on your local machine.
-
-    {{<Aside type="warning">}}
-
-This runs an ephemeral local version of your Worker, and will not be able to access data stored on Cloudflare's Edge (for instance, this includes your data stored on KV). If you'd like to persist data locally, the experimental option `--experimental-enable-local-persistence` will store data in the `wrangler-local-state` subdirectory.
```
rozenmd, line 160: Rather than deleting, probably better to include the message from the deprecation warning:
```
This runs an ephemeral local version of your Worker, and will not be able to access data stored on Cloudflare's Edge (for instance, this includes your data stored on KV). If you'd like to persist data locally, use --persist to use a default path (`.wrangler/state`), or use --persist-to=<YOUR_PATH_HERE> to set a custom persistence path.
```

rozenmd (changes requested):

content/workers/wrangler/commands.md
justinnoel: The message you suggested adding is the one I'm suggesting deleting. There's no need to add something that's already there.

My point is that the **docs** don't need to document the effects of a command that is not only deprecated but **also** doesn't exist in the current code base anymore.

rozenmd: No, it isn't. We deprecated `--experimental-enable-local-persistence` in favour of `--persist` and `--persist-to`.

justinnoel:
Okay.

PR #6187, opened 2022-10-07
Fix custom domain route example

Skye-31:
Custom domain workers routes should not include the wildcard, and wrangler throws an error for this
`X [ERROR] Cannot use "example.com/*" as a Custom Domain; wildcard operators (*) are not allowed`


Skye-31:
Done in #6188 (walshy stole my pr)

PR #6203, opened 2022-10-10
fix(docs): port number type

DSergiu:


rozenmd (approved):
Confirming that wrangler actually throws an error if `"8080"` is used, `8080` is correct.

deadlypants1973 (approved):

PR #6206, opened 2022-10-10
Explain how to paginate with the Analytics Engine SQL API

46bit:
This commit's diff is quite self-explanatory. This is a common request and it's best to have it in the documentation.

### DIFF
--- a/content/analytics/analytics-engine/sql-api.md
+++ b/content/analytics/analytics-engine/sql-api.md
@@ -86,6 +86,46 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
 
 Additionally, the [QUANTILEWEIGHTED function](../sql-reference/#quantileweighted) is designed to be used with sample interval as the third argument.
 
+## Pagination
+
+When you have a lot of data stored, the SQL API may not be able to retrieve all the data with one `SELECT` query. Instead you will need to retrieve it using multiple queries.
+
+Traditionally the SQL `OFFSET` feature has been used for pagination, but we do not support it because it is inefficient with big datasets. Instead you can use the timestamp of each row as a cursor.
+
+An example is where you want to retrieve all of your data from a dataset: `SELECT * FROM your_dataset`. If you only have a few thousand rows then that query should execute fine, but if you have millions of rows then it is likely to fail. Instead you can break down the query into multiple sequential queries.
+
+First run your desired query, adding `ORDER BY timestamp DESC` and a sensible `LIMIT`:
+
+```
+> SELECT * FROM your_dataset ORDER BY timestamp DESC LIMIT 10000
+{
+    "data": [
+        [...],
+        {
+            "timestamp": "2022-10-10 13:56:48",
+            [...]
+        }
+    ]
+}
+```
+
+Then run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
+
+```
+> SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 1000
+{
+    "data": [
+        [...],
+        {
+            "timestamp": "2022-10-09 21:33:00",
+            [...]
+        }
+    ]
+}
+```
+
+Use the new oldest timestamp to run another query, and keep doing that until you run out of results.
+
 ## Example queries
 
 ### Select data with column aliases
--- a/content/analytics/analytics-engine/sql-api.md
+++ b/content/analytics/analytics-engine/sql-api.md
@@ -10,7 +10,7 @@ meta:
 
 The Workers Analytics Engine SQL API is an HTTP API that allows executing SQL queries against your Workers Anaytics Engine datasets.
 
-The API is hosted at `https://api.cloudflare.com/client/v4/accounts/<account_id>/analytics_engine/sql`. 
+The API is hosted at `https://api.cloudflare.com/client/v4/accounts/<account_id>/analytics_engine/sql`.
 
 ## Authentication
 
@@ -88,16 +88,21 @@ Additionally, the [QUANTILEWEIGHTED function](../sql-reference/#quantileweighted
 
 ## Pagination
 
-When you have a lot of data stored, the SQL API may not be able to retrieve all the data with one `SELECT` query. Instead you will need to retrieve it using multiple queries.
+When you have a lot of data stored, the SQL API may not be able to retrieve all the data with one `SELECT` query. Instead you will need to retrieve it using multiple queries. We suggest retrieving 1000 records at a time.
 
 Traditionally the SQL `OFFSET` feature has been used for pagination, but we do not support it because it is inefficient with big datasets. Instead you can use the timestamp of each row as a cursor.
 
 An example is where you want to retrieve all of your data from a dataset: `SELECT * FROM your_dataset`. If you only have a few thousand rows then that query should execute fine, but if you have millions of rows then it is likely to fail. Instead you can break down the query into multiple sequential queries.
 
-First run your desired query, adding `ORDER BY timestamp DESC` and a sensible `LIMIT`:
+First run your desired query, adding `ORDER BY timestamp DESC` and our recommended `LIMIT` of 1000:
 
 ```
-> SELECT * FROM your_dataset ORDER BY timestamp DESC LIMIT 10000
+SELECT * FROM your_dataset ORDER BY timestamp DESC LIMIT 1000
+```
+
+The response will include up to 1000 rows. Since the rows are ordered by descending timestamp, the oldest row has the oldest timestamp:
+
+```
 {
     "data": [
         [...],
@@ -109,10 +114,15 @@ First run your desired query, adding `ORDER BY timestamp DESC` and a sensible `L
 }
 ```
 
-Then run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
+Next run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
 
 ```
-> SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 1000
+SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 1000
+```
+
+You'll get the next 1000 older rows:
+
+```
 {
     "data": [
         [...],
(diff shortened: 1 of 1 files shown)


content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,46 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
+            [...]
+        }
+    ]
+}
+```
+
+Then run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
+
+```
+> SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 1000
```
jimhawkridge:
```suggestion
> SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 10000
```

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,46 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
 
+## Pagination
+
+When you have a lot of data stored, the SQL API may not be able to retrieve all the data with one `SELECT` query. Instead you will need to retrieve it using multiple queries.
+
+Traditionally the SQL `OFFSET` feature has been used for pagination, but we do not support it because it is inefficient with big datasets. Instead you can use the timestamp of each row as a cursor.
+
+An example is where you want to retrieve all of your data from a dataset: `SELECT * FROM your_dataset`. If you only have a few thousand rows then that query should execute fine, but if you have millions of rows then it is likely to fail. Instead you can break down the query into multiple sequential queries.
+
+First run your desired query, adding `ORDER BY timestamp DESC` and a sensible `LIMIT`:
```
jimhawkridge: Can we offer any guidance on what a sensible value might be?
What factors affect this?

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,46 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
+
+```
+> SELECT * FROM your_dataset ORDER BY timestamp DESC LIMIT 10000
```
jimhawkridge: I think it might be worth breaking each of these examples into two code blocks with appropriate syntax highlighting on each.

content/analytics/analytics-engine/sql-api.md
46bit: I think I'm going to suggest `1000` items as a sensible limit. Queries are expensive to run and if we suggest something like `100` it'll add up a lot more.

victor-perov:
Thanks for providing a pagination example!

Described concept of paginating by N rows ordered by timestamp is good enough only if you are 100% sure that you have no more than one event per second. For example, you paginate over aggregates by timestamp (`... GROUP BY timestamp...`). 

It would make more sense if the concept pointed out the importance of understanding the data before running into a pagination SQL query. It will save time for our fellow users to understand their results.

Also, the pagination is primarily used inside a script that runs data extraction in the loop. Therefore, examples with a so-called "preparation step" (like getting an initial "offset") are hard to use as-is.

I want to suggest structuring this information in blocks as follows:

1. describe the points users should be aware of, as this will help to write the needed query:
    
    * LIMIT N OFFSET is not supported;
    * describe hard limits of the API (like 1000 rows of the result set);
    * datasets might have more than one event per second, so users should be aware of inconsistent sorting if they re-run the same query with sorting by a timestamp only;
    * adding more high cardinality fields to `ORDER BY` clause is very helpful;
    * having `timestamp > datetimeA AND timestamp <= datetimeB` is vital as you won't read twice the exact second;
    * including more fields to the filter criteria and using narrow time ranges improves chances of not hitting hard limits;
    * using GROUP BY should be considered when the hard limit is reached, and the user runs out of filter options.

2. Prepare user to understand the given query by describing the nature of data and the rate of events per granule (like one event per second per index);

3. Provide a SQL query that has appropriate WHERE criteria to the rate of events;

4. Finally, show a complete snippet where the given SQL query is injected into a loop that gets a timestamp and iterates by incrementing it by hour/day/etc.

victor-perov (changes requested):
Please, see my comment above.

46bit:
If you have a strong vision I’d suggest writing an alternate PR. This is a small PR intended as a brief how-to and not a detailed article.

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,56 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
 
 Additionally, the [QUANTILEWEIGHTED function](../sql-reference/#quantileweighted) is designed to be used with sample interval as the third argument.
 
+## Pagination
+
+When you have a lot of data stored, the SQL API may not be able to retrieve all the data with one `SELECT` query. Instead you will need to retrieve it using multiple queries. We suggest retrieving 1000 records at a time.
+
+Traditionally the SQL `OFFSET` feature has been used for pagination, but we do not support it because it is inefficient with big datasets. Instead you can use the timestamp of each row as a cursor.
+
+An example is where you want to retrieve all of your data from a dataset: `SELECT * FROM your_dataset`. If you only have a few thousand rows then that query should execute fine, but if you have millions of rows then it is likely to fail. Instead you can break down the query into multiple sequential queries.
```
jimhawkridge: Few thousand seems at odds with the recommendation of 1000.

jimhawkridge (approved):
I think this is fine to cover off the basics of pagination. We can always add a more detailed page with some of Viktor's suggestions later on.
I especially agree it would be nice to have a story around how to cope with duplicate timestamps. I don't think there is actually a good answer to that at the moment. 
How granular are timestamps? This will have a big effect on how much of an issue this is.

content/analytics/analytics-engine/sql-api.md
46bit: I think I might leave this for now, as we don't have a fixed limit. 1000 is intended as a sensible suggestion but you can get back a lot more than that if you try -- it's just, at some point it will error.

I've added DS-12845 to think about how we handle large responses.

46bit:
Looks like this can't be merged with @victor-perov's outstanding requested changes

content/analytics/analytics-engine/sql-api.md
angelampcosta:
```suggestion
An example might be when you want to retrieve all of your data from a dataset: `SELECT * FROM your_dataset`. If you only have a few thousand rows then that query should execute fine, but if you have millions of rows then it is likely to fail. Instead you can break down the query into multiple sequential queries.
```

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,56 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
+
+First run your desired query, adding `ORDER BY timestamp DESC` and our recommended `LIMIT` of 1000:
```
angelampcosta:
```suggestion
First, run your desired query, adding `ORDER BY timestamp DESC` and our recommended `LIMIT` of 1000:
```

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,56 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
+        [...],
+        {
+            "timestamp": "2022-10-10 13:56:48",
+            [...]
+        }
+    ]
+}
+```
+
+Next run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
```
angelampcosta:
```suggestion
Next, run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
```

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,56 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
+
+```sql
+SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 1000
+```
+
+You'll get the next 1000 older rows:
```
angelampcosta:
```suggestion
You will get the next 1000 older rows:
```

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,56 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
 
 Additionally, the [QUANTILEWEIGHTED function](../sql-reference/#quantileweighted) is designed to be used with sample interval as the third argument.
 
+## Pagination
+
+When you have a lot of data stored, the SQL API may not be able to retrieve all the data with one `SELECT` query. Instead you will need to retrieve it using multiple queries. We suggest retrieving 1000 records at a time.
+
+Traditionally the SQL `OFFSET` feature has been used for pagination, but we do not support it because it is inefficient with big datasets. Instead you can use the timestamp of each row as a cursor.
```
angelampcosta:
```suggestion
Traditionally, the SQL `OFFSET` feature has been used for pagination, but we do not support it because it is inefficient with big datasets. Instead you can use the timestamp of each row as a cursor.
```

angelampcosta (approved):

victor-perov:
> If you have a strong vision I’d suggest writing an alternate PR. This is a small PR intended as a brief how-to and not a detailed article.

I see, thanks for clarification. In that case, is it possible to re-phrase the bit about "millions of rows"? In my understanding it doesn't match with further solution and might cause confusion.

Would be nice to get all parts of the guide aligned together.

tlozoot (approved):

tlozoot (approved):

angelampcosta:
Hi @victor-perov! Could you please review again to make sure your comments have been addressed. Thanks!

content/analytics/analytics-engine/sql-api.md
```diff
@@ -86,6 +86,56 @@ The rate at which the data is sampled is exposed via the `_sample_interval` colu
...
+}
+```
+
+Next, run a second query, using the timestamp of the final row in the `WHERE` clause of the next query:
+
+```sql
+SELECT * FROM your_dataset WHERE timestamp < toDateTime("2022-10-10 13:56:48") ORDER BY timestamp DESC LIMIT 1000
+```
+
+You will get the next 1000 older rows:
```
victor-perov: What if the next batch will get 1000 rows with the same timestamp?

I think we should highlight here an assumption that it works for a scenario when we have at most 1 request per second.

Would also be nice to notice that we SQL API does not support `LIMIT n OFFSET m`.

angelampcosta:
Hi @victor-perov! @jimhawkridge addressed your last comment. Do you think we can merge this PR?

KimJ15:
@victor-perov @jimhawkridge See Angela's comment above. We would like to merge this PR and close out the request. Thanks!

haleycode:
Hi all, this PR has been open for quite some time without resolution, so I'm going to close it out for now. We can always reopen or revisit the issue in future if needed!

PR #6260, opened 2022-10-13
Update feedback-loop.md

zeinjaber:
- Custom Rules don't include the allow action. They need to use the skip action instead.
- Also added two notes to differentiate what applies to Bot Management.


content/bots/concepts/feedback-loop.md
```diff
@@ -203,12 +203,25 @@ We appreciate any comments you wish to leave in the description field that might
...
+{{<Aside>}}
 
-* Allowing an **IP address**: Only use an IP address to allow traffic if the IP is a dedicated resource that belongs only to the traffic source you wish to allow. <br>If the traffic you want to allow shares an IP with other traffic sources, or if the IP changes frequently, consider an alternative to allowing by IP address.
+**Note**: The instructions below apply to Enterprise subscription with Bot Management only.
+
+{{</Aside>}}
+
+After submitting a false positive, you can explicitly allow the traffic if you are confident that this traffic source cannot be used for abuse in the future. To allow traffic, you can create a **Firewall Rule** with an [allow](https://developers.cloudflare.com/firewall/cf-firewall-rules/actions/) action or **Custom Rule** with a [Skip the remaining custom rules](https://developers.cloudflare.com/waf/custom-rules/skip/options/) action that matches the characteristics of your false positive report. We recommend any allow or skip rule that you create uses the most narrow possible scope, including restricting the request methods and URIs that the expected traffic has access to, in order to limit potential abuse.
```
kodster28:
```suggestion
After submitting a false positive, you can explicitly allow the traffic if you are confident that this traffic source cannot be used for abuse in the future. To allow traffic, you can create a **Firewall Rule** with an [allow](/firewall/cf-firewall-rules/actions/) action or **Custom Rule** with a [Skip the remaining custom rules](/waf/custom-rules/skip/options/) action that matches the characteristics of your false positive report. We recommend any allow or skip rule that you create uses the most narrow possible scope, including restricting the request methods and URIs that the expected traffic has access to, in order to limit potential abuse.   
```

kodster28 (approved):

PR #6266, opened 2022-10-13
[Workers] Update compat date enable dates

WalshyDev:
Enable dates were added here: https://github.com/cloudflare/workerd/pull/86/files

https://walshy-update-enable-dates.cloudflare-docs-7ou.pages.dev/workers/platform/compatibility-dates/


GregBrimble:
Can we sort the other `date`s too, so they appear in the correct order? I think there's a bit of weirdness if you set a date actually into the future, but can you make them relative anyway? I can look into supporting future dates later.

GregBrimble:
Just pushed directly on to your branch here, hope you don't mind!

GregBrimble (approved):
LGTM lol

PR #6268, opened 2022-10-13
[Bots] IP Validation and Breach of Policy

patriciasantaana:
PCX-5020

### DIFF
--- a/content/bots/reference/verified-bots-policy.md
+++ b/content/bots/reference/verified-bots-policy.md
@@ -108,10 +108,14 @@ A set of validation methods and requirements to gather set IP ranges for a verif
 
 ## Breach of Policy 
 
-If any of the requirements to validate are breached, a service will be removed from the global allowlist. See the list below.
+If any of the requirements to validate are breached, a service will be removed from the global allowlist.
+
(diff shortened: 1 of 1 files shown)


content/bots/reference/verified-bots-policy.md
```diff
@@ -77,3 +77,41 @@ If a bot skips `robots.txt`, it will be rejected.
...
+### Machine Learning 
+
+* A valid user-agent.
+* More than 1000 requests per day.
+* Dynamic IPs are solely used by bot owner. 
+* A user-agent match pattern.
+
+## Breach of Policy 
+
+If any of the requirements to validate are breached, a service will be removed from the global allowlist. See the list below.
```
deadlypants1973:
```suggestion
If any of the following requirements to validate are breached, a service will be removed from the global allowlist.
```

deadlypants1973: I'm not sure about the connection between the list and the breach of policy so adding this as a clarification. But also the list below doesn't seem like a list of requirements because requirements need to have a 'Must...' clause.

dcpena (approved):

PR #6271, opened 2022-10-13
[R2] clarify md5 being available by default for non-multipart objects

Frederik-Baetens:


content/r2/data-access/workers-api/workers-api-reference.md
```diff
@@ -388,7 +388,8 @@ Generally, these fields match the HTTP metadata passed when the object was creat
 
 ## Checksums
 
-If an additional checksum was provided when using the `put()` binding, it will be available on the returned object under the `checksums` property.
+If a checksum was provided when using the `put()` binding, it will be available on the returned object under the `checksums` property.
+The md5 checksum will be included by default for non-multipart objects.
```
deadlypants1973:
```suggestion
If a checksum was provided when using the `put()` binding, it will be available on the returned object under the `checksums` property. The MD5 checksum will be included by default for non-multipart objects.
```

deadlypants1973 (approved):
PCX approved with comment

PR #6276, opened 2022-10-14
Use constant time function to compare HMACs

gorankarlic:
Using simple string comparison to compare a HMACs is not recommended as it may allow a timing-based side channel attack.

### DIFF
--- a/content/workers/tutorials/github-sms-notifications-using-twilio/index.md
+++ b/content/workers/tutorials/github-sms-notifications-using-twilio/index.md
@@ -166,7 +166,7 @@ async function checkSignature(formData, headers) {
   let expectedSignature = await createHexSignature(formData);
   let actualSignature = headers.get('X-Hub-Signature');
 
-  return expectedSignature === actualSignature;
+  return crypto.timingSafeEqual(Buffer.from(expectedSignature, 'hex'), Buffer.from(actualSignature, 'hex'));
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gorankarlic:
Closed by mistake 😵‍💫

content/workers/tutorials/github-sms-notifications-using-twilio/index.md
```diff
@@ -166,7 +166,7 @@ async function checkSignature(formData, headers) {
   let expectedSignature = await createHexSignature(formData);
   let actualSignature = headers.get('X-Hub-Signature');
 
-  return expectedSignature === actualSignature;
+  return crypto.timingSafeEqual(Buffer.from(expectedSignature, 'hex'), Buffer.from(actualSignature, 'hex'));
```
jasnell: Given that `crypto.timingSafeEqual()` will throw an error if the inputs are not the same length, you might consider checking the lengths of each first before.

```suggestion
  const expectedBuffer = Buffer.from(expectedSignature, 'hex');
  const actualBuffer = Buffer.from(actualSignature, 'hex');
  return expectedBuffer.byteLength == actualBuffer.byteLength &&
             crypto.timingSafeEqual(expectedBuffer, actualBuffer);
```

That said, keep in mind that the two `Buffer.from(..., 'hex')` operations are not timing safe, so it might still be possible to extract minute timing information from each. Not sure if that matters enough here, however.

gorankarlic: Excellent point about checking that buffers are of same length.

I do not see how the timing in `Buffer.from` could be abused by an attacker as that timing should not depend on the (prefix) equality between expected and actual HMAC (whereas string comparison timing does exclusively/heavily depend on prefix equality).

Also the primary intent here is that (prominent) example code should always incorporate security best practices and thus serve as a learning opportunity for novices.

jasnell (approved):

deadlypants1973 (approved):

deadlypants1973:
@gorankarlic could you please
`git pull origin production`
`git push -f`
so we can get this merged? :)

PR #6291, opened 2022-10-14
[turnstile] Document response-field & response-field-name

Le0Developer:
Merge when ready to support this.

image

PR #6292, opened 2022-10-14
[turnstile] Document widget sizes & size=compact

Le0Developer:
Compact size 👀 

~~Dimensions are TBA, so I just took a guess. This is not yet in staging but soon:tm:? Staging 👀~~ In production now

~~`size` parameter/attribute is already accepted for hCaptcha compatiblity but using `compact` is currently causing a warning.~~ No longer causes a warning


worenga (commented):
don't we set it to height 65px?

kodster28 (approved):

kodster28:
Howdy @Le0Developer, looks like this is conflicting with some of the recent changes merged in for #6291.

I'd imagine all the fields should be listed, but could you confirm or push up revised changes?

Le0Developer:
@kodster28 Conflicts resolved.

PR #6294, opened 2022-10-14
[PoC] Dynamic Dashboard landing page

kodster28:
Provides recommendations based on the dash page you're leaving (which is supplied by a query in the URL string)

### DIFF
--- a/assets/dashboard-landing.js
+++ b/assets/dashboard-landing.js
@@ -16,6 +16,9 @@ Vue.createApp({
       if (params.dash_area !== null) {
         this.visibleData = data.filter((item) => item.dashPath === params.dash_area);
         if (this.visibleData.length === 0) {
+          this.visibleData = data.filter((item) => params.dash_area.includes(item.dashPath));
+        }        
+        if (this.visibleData.length === 0) {
           this.visibleData = data.filter((item) => item.id === "Default");
         }
       } else {
--- a/assets/dashboard-landing.js
+++ b/assets/dashboard-landing.js
@@ -1,23 +1,22 @@
 
 import * as recommendations from "./dashboard-landing.json";
 
-let data = recommendations["default"]["resources"]
+const data = recommendations["default"]["resources"]
 
-const params = new Proxy(new URLSearchParams(window.location.search), {
-  get: (searchParams, prop) => searchParams.get(prop),
-});
+const dashPathMap = data.map((x) => x.dashPath )
 
 Vue.createApp({
   methods: {
-    filterByParam() {
+    filterByParam() {      
       const params = new Proxy(new URLSearchParams(window.location.search), {
         get: (searchParams, prop) => searchParams.get(prop),
       });
       if (params.dash_area !== null) {
-        this.visibleData = data.filter((item) => item.dashPath === params.dash_area);
-        if (this.visibleData.length === 0) {
+        if (params.dash_area in this.dashPathMap) {
+          this.visibleData = data.filter((item) => item.dashPath === params.dash_area);
+        } else {
           this.visibleData = data.filter((item) => params.dash_area.includes(item.dashPath));
-        }        
+        }         
         if (this.visibleData.length === 0) {
           this.visibleData = data.filter((item) => item.id === "Default");
         }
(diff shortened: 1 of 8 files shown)


assets/dashboard-landing.js
```diff
@@ -0,0 +1,42 @@
...
+          this.visibleData = data.filter((item) => item.dashPath === params.dash_area);
+        } else {
+          this.visibleData = data.filter((item) => params.dash_area.includes(item.dashPath));
+        }         
+        if (this.visibleData.length === 0) {
+          this.visibleData = data.filter((item) => item.id === "Default");
+        }
+      } else {
+        this.visibleData = data.filter((item) => item.id === "Default");
+      }
```
pedrosousa:
```suggestion
        if (this.dashPathMap.includes(params.dash_area)) {
          this.visibleData = data.find((item) => item.dashPath === params.dash_area);
        } else {
          this.visibleData = data.find((item) => params.dash_area.includes(item.dashPath));
        }
      }
      if (!this.visibleData) {
        this.visibleData = data.find((item) => item.id === DEFAULT_ITEM_ID);
      }
```

assets/dashboard-landing.js
```diff
@@ -0,0 +1,42 @@
+
+import * as recommendations from "./dashboard-landing.json";
+
+const data = recommendations["default"]["resources"]
+
+const dashPathMap = data.map((x) => x.dashPath )
```
pedrosousa:
```suggestion
const dashPathMap = data.map((x) => x.dashPath)

const DEFAULT_ITEM_ID = "Default"
```

assets/dashboard-landing.js
```diff
@@ -0,0 +1,42 @@
...
+
+Vue.createApp({
+  methods: {
+    filterByParam() {      
+      const params = new Proxy(new URLSearchParams(window.location.search), {
+        get: (searchParams, prop) => searchParams.get(prop),
+      });
+      if (params.dash_area !== null) {
+        if (params.dash_area in this.dashPathMap) {
```
pedrosousa: `in` would search for a _property_ with the given name in the object, not a value (replaced with `includes()` in the suggestion below).

assets/dashboard-landing.js
```diff
@@ -0,0 +1,42 @@
...
+          this.visibleData = data.filter((item) => item.dashPath === params.dash_area);
+        } else {
+          this.visibleData = data.filter((item) => params.dash_area.includes(item.dashPath));
+        }         
+        if (this.visibleData.length === 0) {
```
pedrosousa: We can skip this condition check here and handle the "not found" case right after closing this inner `if`, since we already have the same code there.

assets/dashboard-landing.js
```diff
@@ -0,0 +1,42 @@
...
+
+Vue.createApp({
+  methods: {
+    filterByParam() {      
+      const params = new Proxy(new URLSearchParams(window.location.search), {
+        get: (searchParams, prop) => searchParams.get(prop),
+      });
+      if (params.dash_area !== null) {
+        if (params.dash_area in this.dashPathMap) {
+          this.visibleData = data.filter((item) => item.dashPath === params.dash_area);
```
pedrosousa: We're always using the first item only, so I'd replace `filter()` with `find()` and remove `[0]` from the template.

assets/dashboard-landing.js
```diff
@@ -0,0 +1,42 @@
...
+        this.visibleData = data.filter((item) => item.id === "Default");
+      }
+    }
+  },
+  beforeMount() {
+    this.filterByParam();
+  },
+  template: `
+        <ul class="recommendationList">
+            <li class="recommendationLink" v-for="element in visibleData[0].recommendations"><a :href='element.url_path' target='_blank'>[[ element.title ]]</a></li>
```
pedrosousa:
```suggestion
            <li class="recommendationLink" v-for="element in visibleData.recommendations"><a :href='element.url_path' target='_blank'>[[ element.title ]]</a></li>
```

pedrosousa (approved):

assets/dashboard-landing.js
```diff
@@ -0,0 +1,44 @@
...
+const DEFAULT_ITEM_ID = "Default"
+let longestMatch;
+
+Vue.createApp({
+  methods: {
+    filterByParam() {      
+      const params = new Proxy(new URLSearchParams(window.location.search), {
+        get: (searchParams, prop) => searchParams.get(prop),
+      });
+      if (params.dash_area !== null) {
```
pedrosousa, line 16: You're missing an `else` for this `if` (`dash_area` value not provided) -- we should probably show the default recommendations, instead of none. 

Even though there's an `else` that is visually aligned with this `if`, it applies to another `if`. You should probably run `prettier` on this file.

PR #6300, opened 2022-10-17
Clarified DDoS managed ruleset customers scope

wolruf:
I found the previous statement to lead to think that only MT and Spectrum benefit from it. I took the statement found on: https://developers.cloudflare.com/ddos-protection/managed-rulesets/network/


pedrosousa (approved):

pedrosousa:
Thanks, @wolruf!

PR #5732, opened 2022-09-02
Update Direct Creator Upload docs with examples

irvinebroque:
Updates the Direct Creator Upload docs to be simpler, shorter, and link to runnable examples.

Preview: https://3608f320.cloudflare-docs-7ou.pages.dev/stream/uploading-videos/direct-creator-uploads/

- [x] Merge https://github.com/cloudflare/templates/pull/98 first

PR #6337, opened 2022-10-19
Add HTML ID for jumplink

harlantwood:
The jumplink 

```
[generate URLs that use the tus protocol](/stream/uploading-videos/direct-creator-uploads#tus)
```

is broken.  This is one way to fix it. 😄


dcpena:
Hi Harlan, thanks so much for opening this PR. It looks like this link got fixed in PR https://github.com/cloudflare/cloudflare-docs/pull/5732.

PR #6374, opened 2022-10-21
Fix broker create line

sksar:
The command to create broker was wrong.
FIxed the line with the proper command.

When I tried the command it failed, I was stuck at this for a while, until I figured it out that the command in the documentation is wrong.


sksar:
@dcpena any updates?

dcpena (approved):

PR #6382, opened 2022-10-24
Wrong code - use fetch() instead of new Response()

alexander-josef:
Isn't the provided code wrong? Use fetch() in the return statement to deliver back the image from imagedelivery. new Response() only returns the URL as text.


kristianfreeman (approved):

kristianfreeman:
oopsies you are absolutely right

PR #6404, opened 2022-10-24
turnstile: clarify the size only affects Managed/Non-interactive

punkeel:


kodster28 (approved):

Le0Developer:
Currently you can force a managed/non-interactive to be hidden by using `size=invisible`. Is this considered a bug?

image
image

Repro: https://turnstile-demo.pages.dev/ with `0x4AAAAAAAA6ulk8fhx8AjD-` as sitekey.

punkeel:
Indeed, good catch! Will get this fixed.

ferreiro:
Hey @punkeel, we're starting to use turnstile to validate thousands of users daily :) however, this change broke our user interface (at least in development). We have the [widget on invisible mode](https://twitter.com/JGFerreiro/status/1585323892222021632), and after this change (we were passing invisible) it's showing the widget.

Can u let me know how to fix it? Shot a video: https://www.loom.com/share/0bfe4e916c3b400bb9cd9c1ded7aae88

PR #6407, opened 2022-10-25
stream docs: fix markdown code block to render docs correctly

pew:


pedrosousa (approved):

pedrosousa:
Thank you for the fix, @pew!

PR #6412, opened 2022-10-25
Fix broken SQL example

jimhawkridge:
I think this broke at some point.
Or maybe I never tested it!?


46bit:
Oh, I think I broke this with the semantics limitations :(
That should change with DS-12438, no ETA yet though

46bit (approved):

PR #6415, opened 2022-10-25
Update _index.md

jjxtra:
Denote IPBan support (https://github.com/DigitalRuby/IPBan/blob/master/Recipes/Windows/Integrations/CloudflareBlock.ps1)


pedrosousa:
Thank you for your proposal, @jjxtra. 
However, except for integration guides we do not reference third-party tools unless it's a well-known, free tool that anyone can use.

jjxtra:
@pedrosousa IPBan is definitely free, definitely anyone can use (Linux + Windows support, MIT license) and has over 1k stars on github. Would love some clarification on what about it does not meet your criteria for well-known and free that anyone can use?

https://github.com/DigitalRuby/IPBan

PR #6417, opened 2022-10-25
[ZT] Add polling frequency

ranbel:


content/cloudflare-one/identity/devices/_index.md
```diff
@@ -40,3 +39,17 @@ You can now use your device posture check in an [Access policy](/cloudflare-one/
 * The IdP used to authenticate to Cloudflare Zero Trust if posture check is part of an Access policy.
 * `<your-team-name>.cloudflareaccess.com` if posture check is part of an Access policy.
 * The application protected by the Access or Gateway policy.
+
+## Policy enforcement rate
+
+Access detects changes in device posture at the same rate as the [polling frequency](#polling-frequency) configured for the posture check. Due to how Gateway caches data, Gateway policies are subject to an additional five minute delay. For example, if your polling frequency is set to 10 minutes, it may take up to 15 minutes for Gateway to detect posture changes on a device.
```
kkrum: Can we change "Due to how Gateway aches data" to something like "Because Gateway evaluates NETWORK and HTTP policies on every request, it maintains local cache of posture results that is only updated every 5 minutes" or something like that? Trying to be more specific about what data is cached

kkrum (approved):

PR #6430, opened 2022-10-26
remove "ssh" from _tunnel-cloudflared-access.md

pableu:
It's also included in the pages for SMB and RDP.


ranbel (approved):
Good catch, thanks!

PR #6454, opened 2022-10-27
[DNS] Change proxy status in MX records example in email-records.md

MeesJ:
The A/AAAA record used to refer to a mailserver should not be proxied, as this will cause email delivery problems because Cloudflare does not proxy port 25 (https://support.cloudflare.com/hc/en-us/articles/200168876-Email-undeliverable-when-using-Cloudflare).


kodster28:
We actually just updated that article to remove the warning about proxied records. The current setup in the MX records documentation is correct.

PR #6456, opened 2022-10-27
[R2] Cleanup S3 operations

vlovich:
Add "account level" operations and reorder the ops on the page so that everything is correct (ListObjects for example was incorrectly listed as an object-level operation).

Add some color commentary to the `data-access/s3-api` and `data-access/workers-api` pages so that it's a bit of a softer landing explaining how those things work (e.g. if you link to these from some other piece, we should describe a bit as to what they are instead of only providing a table of contents).

### DIFF
--- a/content/r2/data-access/s3-api/api.md
+++ b/content/r2/data-access/s3-api/api.md
@@ -27,9 +27,27 @@ When using the S3 API, the region for an R2 bucket is `auto`. For compatibility 
 
 This also applies to the `LocationConstraint` for the `CreateBucket` API.
 
+## Account-level operations
+
+The following tables are related to operations you perform directly on the account itself. The URL
+will be `https://<ACCOUNT>.r2.cloudflarestorage.com/`. There may be search parameters but there
+is no additional path in the URL.
+
+### Implemented account-level operations
+
+| API Name                                | Feature                           |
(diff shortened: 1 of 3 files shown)


content/r2/data-access/s3-api/api.md
```diff
@@ -27,9 +27,27 @@ When using the S3 API, the region for an R2 bucket is `auto`. For compatibility
...
+The following tables are related to operations you perform directly on the account itself. The URL
+will be `https://<ACCOUNT>.r2.cloudflarestorage.com/`. There may be search parameters but there
+is no additional path in the URL.
+
+### Implemented account-level operations
+
+| API Name                                | Feature                           |
+| --------------------------------------- |---------------------------------- |
+| ✅ [ListBuckets](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html)     |                                    |
+| ✅ [HeadBucket](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html)       | ❌ Bucket Owner: <br> &emsp;  ❌ x-amz-expected-bucket-owner |
```
vlovich, line 41: These are wrong. `ListBuckets` is the only account-level operation. HeadBucket/CreateBucket/DeleteBucket should remain bucket-level operations.

PR #6485, opened 2022-10-31
Call out rate limiting for public buckets

mnewswanger-cloudflare:
Add notes around public bucket usage and rate limiting

### DIFF
--- a/content/r2/data-access/public-buckets.md
+++ b/content/r2/data-access/public-buckets.md
@@ -99,6 +99,14 @@ The domain is no longer connected to your bucket and will no longer appear in th
 Enabling managed public access for your bucket will make the content of your bucket available to the Internet through a `r2.dev` Cloudflare subdomain.
 To enable public access for your buckets:
 
+{{<Aside type="note">}}
+
+Public access through `r2.dev` subdomains are rate limited and should only be used for development purposes.
+
+To enable access management, Cache and bot management features, you must set up a custom domain when enabling public access to your bucket.
+
+{{</Aside>}}
+
 1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com) > select your account > **R2**.
 
 ![Choose R2 from the side menu](/r2/static/public-buckets-r2.png)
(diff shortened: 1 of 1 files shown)


bthwaites (approved):

content/r2/data-access/public-buckets.md
```diff
@@ -99,6 +99,14 @@ The domain is no longer connected to your bucket and will no longer appear in th
 Enabling managed public access for your bucket will make the content of your bucket available to the Internet through a `r2.dev` Cloudflare subdomain.
 To enable public access for your buckets:
 
+{{<Aside type="note">}}
+
+Public access through `r2.dev` subdomains are rate limited and should only be used for development purposes.
+
+To enable access management, Cache and bot management features, you must set up a custom domain when enabling public access to your bucket.
```
harshal317, line 106: Should this maybe be instead of enabling public access to your bucket

harshal317 (approved):

content/r2/data-access/public-buckets.md
Frederik-Baetens, line 106: How about something like:

```suggestion
For production workloads requiring public access to your bucket, you must set up a custom domain. This also enables the use of Cache, bot management and access management features.
```

That makes it clearer that custom domains are intended for production workloads and that the difference isn't just those extra features.

mnewswanger-cloudflare: The latter bit there is just moving up from lower in the doc (no changes here). I actually prefer the phrasing calling out the functionality first followed by what action users need to take.

Frederik-Baetens: Yeah, but I think we should explicitly communicate that custom domains are intended for production workloads/don't have rate limits. I already saw confusion about this in the discord.

mnewswanger-cloudflare: That copy's unchanged from earlier (just moved up so the info's above the steps to enable). I think the phrasing gets a little confusing with the phrasing of `Managed Pubic Access` vs more generic "public access" so I'd lean toward just doing nothing as it's already been reviewed.

mnewswanger-cloudflare: I'll merge this one as-is for now; if we continue to see noise and confusion related to it, we can revisit and easily update later.

PR #6510, opened 2022-11-02
[ZT] Add new client MDM IP override parameters

schultetwin1:
The next release of the WARP client will include new MDM parameters which allow users to change the IP address uses to talk with the API, DoH, and WARP endpoints.

This change adds documentations for those MDM parameters

### DIFF
--- a/content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
+++ b/content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
@@ -111,6 +111,66 @@ When the WARP client is deployed via MDM, the in-app **Send Feedback** button is
 - `https://<support.example.com>` &mdash; Use an `https://` link to open your company's internal help site.
 - `mailto:<yoursupport@example.com>` &mdash; Use a `mailto:` link to open your default mail client.
 
+### `override_api_endpoint`
+
+Overrides the IP address the WARP client will use to communicate with the client
+orchestration API. This parameter is only usable by clients with a version newer
+than 2022.10.0.0. Most IT admins should not set this setting as it will redirect
+all API traffic to a new IP. If the parameter is used, please be sure to update
+your organization's firewall to ensure this IP is allowed through.
+
+Although the IP has changed the domain name used will remain the same, meaning
+this new IP address may not terminate the TLS connection.
+
+**Value Type:** `string`
+
+**Value:**
+
+- `1.2.3.4` &mdash; Redirect all client orchestration API calls to 1.2.3.4.
+
+The string must be a valid IPv4 or IPv6 address, otherwise the WARP client will
+fail to parse the entire mdm file.
+
+### `override_doh_endpoint`
+
+Overrides the IP address the WARP client will use to resolve DNS queries using
+DNS over HTTPS. This parameter is only usable by clients with a version newer
+than 2022.10.0.0. Most IT admins should not set this setting as it will redirect
+all DNS over HTTPS traffic to a new IP. If the parameter is used, please be sure
+to update your organization's firewall to ensure this IP is allowed through.
+
+Although the IP has changed the domain name used will remain the same, meaning
+this new IP address may not terminate the TLS connection.
+
+**Value Type:** `string`
+
+**Value:**
+
+- `1.2.3.4` &mdash; Redirect all client DNS over HTTPS lookups to 1.2.3.4.
+
+The string must be a valid IPv4 or IPv6 address, otherwise the WARP client will
+fail to parse the entire mdm file.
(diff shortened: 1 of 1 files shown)


content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
```diff
@@ -111,6 +111,66 @@ When the WARP client is deployed via MDM, the in-app **Send Feedback** button is
 - `https://<support.example.com>` &mdash; Use an `https://` link to open your company's internal help site.
 - `mailto:<yoursupport@example.com>` &mdash; Use a `mailto:` link to open your default mail client.
 
+### `override_api_endpoint`
+
+Overrides the IP address the WARP client will use to communicate with the client
+orchestration API. This parameter is only usable by clients with a version newer
+than 2022.10.0.0. Most IT admins should not set this setting as it will redirect
```
ranbel, line 118: If most admins should not use these settings, what's the use case for these override parameters?

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
```diff
@@ -111,6 +111,66 @@ When the WARP client is deployed via MDM, the in-app **Send Feedback** button is
 - `https://<support.example.com>` &mdash; Use an `https://` link to open your company's internal help site.
 - `mailto:<yoursupport@example.com>` &mdash; Use a `mailto:` link to open your default mail client.
 
+### `override_api_endpoint`
+
+Overrides the IP address the WARP client will use to communicate with the client
```
ranbel, line 116: Where can a user see the default api_endpoint, doh_endpoint, and warp_endpoint?

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
```diff
@@ -111,6 +111,66 @@ When the WARP client is deployed via MDM, the in-app **Send Feedback** button is
...
+orchestration API. This parameter is only usable by clients with a version newer
+than 2022.10.0.0. Most IT admins should not set this setting as it will redirect
+all API traffic to a new IP. If the parameter is used, please be sure to update
+your organization's firewall to ensure this IP is allowed through.
+
+Although the IP has changed the domain name used will remain the same, meaning
```
ranbel, line 122: > Although the IP has changed the domain name used will remain the same, meaning this new IP address may not terminate the TLS connection.

Could you clarify what this sentence means?

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
kkrum: This is primarily for Zero Trust usage of employees inside China and will require a separate endpoint the customer will need to support.

I think we should add some language like "This should be primarily used by customers who need to route traffic through 3rd party providers (ex. companies operating inside China that have negotiated path through the firewall with 3rd parties)." or something like that. We'll have blogs and more documentation coming at a later date.

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
kkrum: End users won't be able to see this at all, this is purely an admin controlled thing.

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
kkrum: @schultetwin1 how does this impact us at all? The only communication with these endpoints should be our daemon, which all works in this mode.

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
schultetwin1: If this override is set, users will be able to see it via `warp-cli settings` which will output the override value

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/parameters.md
schultetwin1: The point of this sentence was to make it clear you could not redirect cloudflare's traffic to a destination not owned by Cloudflare. I'm concerned people may read this and think they can host their own server that captures WARP's API requests or DNS requests.

schultetwin1:
Covered by #6643 so closing this one

PR #6523, opened 2022-11-02
Update debugging-pages.md

joekrump:
Fix a typo on one of the docs pages.


deadlypants1973:
Hello, thank you for your suggestion. This is intentional. Cloudflare style guide doesn't include - before "re" prefixes.

joekrump:
Fair enough. Thanks for letting me know. I found it confusing because I though it was "read" with an extra "d" at first. I thought that the hyphen would provide greater clarity to readers.

PR #6533, opened 2022-11-03
Revert documentation of Queues waitUntil behavior

a-robinson:
This reverts commit 0c66c539a8d0b7cf3044be352994935e950ae73f.

We have not yet changed the behavior to work like this, and based on conversations this week it sounds like if we want to go ahead and do so that we'll also have to update how waitUntil works in cron triggers.


deadlypants1973 (approved):

jbw1991 (approved):
👍

PR #6548, opened 2022-11-04
[Stream] Added missing code block end to Example webhook payload

cranberry3148:


TheFruxz:
Never thought, that such a small change would entirely change the way, of how I look at code. LGTM

dcpena (approved):

PR #6551, opened 2022-11-04
Update rdp.md

deansundquist:
The Connect as a User section here just isn't friendly with Windows users. 

It would seem that we're making an assumption here that we're not on a Windows machine, where I'd suspect the majority of user's are using Windows, if they're using RDP.  Also a majority of those user's likely also have port 3389 already open to RDP to that local machine.

This gets confusing.  Adding a note is the least we can do here.  I'd suggest that we rewrite this with a port thats other than 3389.


abelinkinbio (approved):

ranbel:
> I'd suggest that we rewrite this with a port thats other than 3389.

I think this is a good idea if a different port will better suit the majority of users. Could you edit this PR with the proposed changes?

Issue #6547, opened 2022-11-04
Add detail about SSH principals to Access short-lived SSH certs


itzoey:
### Which Cloudflare product(s) does this pertain to?

Cloudflare One

### Subject Matter

SSH principals to Access short-lived SSH certs

### Content Location

Updated step 2 under https://developers.cloudflare.com/cloudflare-one/identity/users/short-lived-certificates/
Updated final step under https://developers.cloudflare.com/cloudflare-one/tutorials/ssh-cert-bastion/

### Additional information

The current docs aren't great, since they say your Access usernames have to match the *nix username of the box you sign into (not true), and the recommended workaround for if they don't (adding another user with a username that matches, and using the same uid and gid) is janky at best.

The docs should talk briefly about SSH principals and explain how to use the principal options to change who can sign in as who.

Draft of my changes is in this branch: https://github.com/cloudflare/cloudflare-docs/tree/zoey/ssh-short-lived, including the commit I cherry-picked into #6546 since it should really be its own PR.

Rendered here: https://zoey-ssh-short-lived.cloudflare-docs-7ou.pages.dev/cloudflare-one/identity/users/short-lived-certificates/

I simplified the docs into a basic description at the top, then a dropdown section for username aliasing, and another for allowing anyone (with Access perms) to sign in as a particular user (or any user).

ranbel:
This content seems like a useful addition. Would you mind opening a PR with the proposed changes?

itzoey:
done, #6589

PR #6589, opened 2022-11-08
[Access] Add detail about SSH principals to Access short-lived SSH certs

itzoey:
re: #6547

The current docs aren't great, since they say your Access usernames have to match the *nix username of the box you sign into (not true), and the recommended workaround for if they don't (adding another user with a username that matches, and using the same uid and gid) is janky at best.

I simplified the docs into a basic description at the top, then a dropdown section for username aliasing, and another for allowing anyone (with Access perms) to sign in as a particular user (or any user).

This could definitely use some copyediting, but all the crucial information should be there.

Also did some spiffing up throughout other parts of the page and updated the SSH example config in the tutorial I apparently missed in #6546

### DIFF
--- a/content/cloudflare-one/_partials/_ssh-restart-server.md
+++ b/content/cloudflare-one/_partials/_ssh-restart-server.md
@@ -5,24 +5,28 @@ _build:
   list: never
 ---
 
-Once you have modified your SSHD configuration, you still need to restart the SSH service on the remote machine.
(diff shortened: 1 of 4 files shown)


ranbel (approved):

ranbel:
The additional examples are great! If the copyedits look good to you, you can go ahead and merge the changes.

itzoey:
Read through it all, looks great to me!
Thanks for copyediting my mess :)

PR #6604, opened 2022-11-08
add time info in days.

kbitr:
add "Custom Hostname certificates are valid for 90 days and tried to be renewed 30 days before expiry."


kodster28:
Addressed in #6975

PR #6608, opened 2022-11-08
Update how-kv-works.md

koeninger:
"check in with the central datastores" led to customer confusion in discord, since it implies some kind of proactive process to update keys even if they aren't requested.


content/workers/learning/how-kv-works.md
```diff
@@ -22,7 +22,7 @@ Frequent reads from the same location return the cached value without reading fr
...
+KV achieves this performance by being eventually-consistent. Changes are usually immediately visible in the Cloudflare network edge location at which they are made but may take up to 60 seconds to be visible in other network edge locations as their cached versions of the data time out. In particular, propagation of changes takes longer to locations which have recently read a previous version of a given key (including reads that indicated the key did not exist, which are also cached locally). Workers KV is not ideal for situations where you need support for atomic operations or where values must be read and written in a single transaction.
```
garretcharp: I assume the third sentence should also be changed from propagation to visibility?


```suggestion
KV achieves this performance by being eventually-consistent. Changes are usually immediately visible in the Cloudflare network edge location at which they are made but may take up to 60 seconds to be visible in other network edge locations as their cached versions of the data time out. In particular, visibility of changes takes longer in locations which have recently read a previous version of a given key (including reads that indicated the key did not exist, which are also cached locally). Workers KV is not ideal for situations where you need support for atomic operations or where values must be read and written in a single transaction.
```

deadlypants1973 (approved):

rts-rob (approved):
LGTM

PR #6618, opened 2022-11-09
D1: first() now returns null instead of throw(D1_NORESULTS)

celso:
D1: "lastRowId" and "changes" not implemented yet

### DIFF
--- a/content/d1/client-api/index.md
+++ b/content/d1/client-api/index.md
@@ -65,19 +65,21 @@ Type conversion from Javascript inputs to D1 inputs is as follows:
 * Booleans will be turned into integers where 1 is `TRUE` and 0 is `FALSE`.
 
 ## Return object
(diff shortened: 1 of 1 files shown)


content/d1/client-api/index.md
```diff
@@ -65,23 +65,21 @@ Type conversion from Javascript inputs to D1 inputs is as follows:
...
-  lastRowId: integer, // the rowid of the last row inserted or null if it doesn't apply
-  changes: integer, // total # of rows that were inserted/updated/deleted, or `null`
   duration: number, // duration of the operation in milliseconds
   success: boolean, // true if the operation was successful, false otherwise
 }
 ```
 
 Example:
 ```js
-const { lastRowId } = await db.prepare('INSERT INTO users (name, age) VALUES (?1, ?2)').bind( "John", 42 ).run();
```
geelen, line 82: ah I really want this to be `meta.duration` but I suppose I'm running out of time to get that in...

geelen (approved):

PR #6665, opened 2022-11-11
[ZT] Revise Ansible guide

ranbel:


ranbel:
@abelinkinbio The Ansible deployment didn't work without an SSH key so I added the necessary steps to the doc. Also, I discovered that the deployment throws "Error: Failed to parse ssh private key: ssh: cannot decode encrypted private keys" if you try to use a key with a passphrase.

kodster28 (approved):

PR #6669, opened 2022-11-13
chore: use `path.Join` to avoid double slashes

jbergstroem:
In the shortcode for generating type links, `printf` is used, generating links like these: `http://localhost:1313/workers//runtime-apis/response`. Using `path.Join`, hugo handles the link generation/url massage `http://localhost:1313/workers/runtime-apis/response`.

(Example links taken from https://developers.cloudflare.com/workers/runtime-apis/cache)


kodster28 (approved):

kodster28:
Thanks for the contribution, @jbergstroem!

PR #6671, opened 2022-11-14
Clarify per-request limits

ignoramous:
This part is ambiguous and so made it clear that the limits are indeed per-request.

Also, added a missing oxford comma.


victor-perov (approved):

angelampcosta:
@ignoramous thank you for your suggestion!

PR #6672, opened 2022-11-14
Add link to cache rules

ignoramous:
Caching seems like an integral part for R2, and so, add a link to emphasize that. 

Perhaps, the R2 docs need a `Caching` section all on its own, but I am ill-placed to start one; and hence this PR, in its stead.


ignoramous:
Ah, I don't know just how to link to cache rules. But it'd be rad if someone who knew how, added it. Closing this PR.

PR #6675, opened 2022-11-14
Update github.md

pew:
A GitHub organization is required for this integration to work, if installed for a personal user no integrations are listed.


pew:
Was meaning to update a different page.

PR #6718, opened 2022-11-16
[Learning paths] Added sorting to paths + filter options

kodster28:


layouts/shortcodes/learning-path-list.html
```diff
@@ -10,10 +10,10 @@
 {{ $category_array := slice }}
 {{- range $data -}}
 {{ with .products }}
-{{- $product_array = union $product_array . -}}
+{{- $product_array = sort (union $product_array .) -}}
```
pedrosousa, line 13: Nit: We could eventually do the sort once (outside the loop) instead instead of re-sorting after adding each item, but the current implementation will have a very minimal impact in build times.

pedrosousa (approved):

PR #6725, opened 2022-11-16
queues: fix defaults

elithrar:
Fix the default days & batching config to be correct per:

```
const DefaultBatchSize int64 = 10
const DefaultMaxRetries int64 = 3
const DefaultMaxWaitTimeMS int64 = 5000
```


deadlypants1973 (approved):

Shadyboyty (commented):
Yer

PR #6740, opened 2022-11-17
CTA to Deployments for Environments page

kabirsikand:


content/workers/platform/environments.md
```diff
@@ -7,6 +7,8 @@ title: Environments
 
 {{<Aside type="note">}}
 
+We recommend using [Deployments](/workers/platforms/deployments) in place of Environments. Deployments give you a powerful audit log of changes to your application, and will soon include integrated rollbacks and automated deployment.
```
admah: @kabirsikand I think it would be good to put this paragraph at the end instead of where it is now.

admah (commented):
I made one suggestion re: ordering.

content/workers/platform/environments.md
kabirsikand: Do you think at the end of the 'Note' or end of the page? I was thinking both. We should make sure *immediately* developers know that Deployments is the feature they should use instead of environments, IMO.

admah: I meant the end of the "Note". I wouldn't think the same text needs to be repeated on the same page verbatim, but I don't see either of these as blocking. I think it's good to merge when ready.

admah (approved):

PR #6762, opened 2022-11-19
Update timeout default value

SferaDev:


content/queues/get-started.md
```diff
@@ -197,7 +197,7 @@ In your consumer Worker, you are using queues to auto batch messages using the `
 
 `max_batch_size` (defaults to 10) helps to reduce the amount of times your consumer Worker needs to be called. Instead of being called for every message, it will only be called after 10 messages have entered the queue.
 
-`max_batch_timeout` (defaults to 3 seconds) helps to reduce wait time. If the producer Worker is not sending up to 10 messages to the queue for the consumer Worker to be called, the consumer Worker will be called every 10 seconds to receive messages that are waiting in the queue.
+`max_batch_timeout` (defaults to 3 seconds) helps to reduce wait time. If the producer Worker is not sending up to 10 messages to the queue for the consumer Worker to be called, the consumer Worker will be called every 3 seconds to receive messages that are waiting in the queue.
```
elithrar, line 200: This isn't the correct change: the config above shows 10 seconds. The text is aligned with the config.

elithrar (commented):
(Not approved/this PR isn't valid, sorry)

content/queues/get-started.md
SferaDev: Hi @elithrar, when reading the docs to configure my queue I went directly to the description of the properties without checking the `toml` file you had above. I will close the PR as you're right now with the context.

I still find it confusing that the explanation of a property uses an example configuration file rather than explaining the default behaviour and how to customize it.

PR #6777, opened 2022-11-21
Update crawl rates

lindsey-em:
The marketing page indicates 7/3/daily https://www.cloudflare.com/always-online/


angelampcosta:
@lindsey-em thank you for putting this out. In this case, the marketing page is out of date, but we are working to fix this.

PR #5015, opened 2022-07-02
[Pages] Functions docs rewrite

WalshyDev:
Google doc: https://docs.google.com/document/d/1t1TPJxlLh0xsVJ_4UgtgkiIl3Jkm0mB6KST4Xj9VRpI/edit?usp=sharing

cc: @Cherry @nevikashah @Ekwuno 

closes #5374

PR #6810, opened 2022-11-23
[Pages] Fix link to middleware in Headers

Erisa:
This is needed since #5015, because middleware now has its own page


mohammadzahoor (commented):
<git>

mohammadzahoor:
<head>

WalshyDev (approved):

PR #6823, opened 2022-11-23
Creates an Action that automatically adds PRs issues to the PCX Github project

haleycode:


.github/workflows/project-add.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add open issues and PRs to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@RELEASE_VERSION
```
kodster28:
```suggestion
      - ./.github/actions/project-add
```

kodster28:
I think you're missing a few things here:
1) A corresponding entry in the .github/actions folder (https://github.com/cloudflare/cloudflare-docs/tree/production/.github/actions). The workflow calls the script stored in the action.
2) I believe for testing + other, you might want to mimic the `on` structure visible here --> https://github.com/cloudflare/cloudflare-docs/blob/production/.github/workflows/pr.yml#L4

haleycode:
This requires way more development than I expected and is beyond my skillset. Closing out due to lack of engineering resources.

haleycode:
> I think you're missing a few things here:
> 
> 1. A corresponding entry in the .github/actions folder (https://github.com/cloudflare/cloudflare-docs/tree/production/.github/actions). The workflow calls the script stored in the action.
> 2. I believe for testing + other, you might want to mimic the `on` structure visible here --> https://github.com/cloudflare/cloudflare-docs/blob/production/.github/workflows/pr.yml#L4

Just for future reference, I think we'd also need to implement these packages in the repo, which might be a non-starter: https://github.com/actions/toolkit/blob/master/README.md#packages

pedrosousa:
The assignment workflows we have are custom made, but could eventually use an action to assign a project from the GitHub Marketplace, like Haley had initially suggested.
Wouldn't the [`actions/add-to-project`](https://github.com/marketplace/actions/add-to-github-projects) action meet our needs? This wouldn't require developing a custom workflow, unless we needed custom behavior.

PR #6830, opened 2022-11-24
[R2] Update changelog

sdnts:
This had started to lag behind.

In the future, we'll create PRs updating the changelog as part of our release process, which should prevent this from happening again.

### DIFF
--- a/content/r2/platform/changelog.md
+++ b/content/r2/platform/changelog.md
@@ -6,6 +6,42 @@ weight: 3
 
 # Changelog
 
+## 2022-11-21
+
+- Fixed a bug in `ListObjects` where `startAfter` would skip over objects with keys that have numbers right after the `startAfter` prefix.
+- Add worker bindings for multipart uploads.
+
+## 2022-11-17
+
+- Unconditionally return HTTP 206 on ranged requests to match behavior of other S3 compatible implementations.
+- Fixed a CORS bug where `AllowedHeaders` in the CORS config were being treated case-sensitively.
+
+## 2022-11-08
+
+- Copying multipart objects via `CopyObject` is re-enabled.
+- `UploadPartCopy` is re-enabled.
+
+## 2022-10-28
+
+- Multipart upload part sizes are always expected to be of the same size, but this enforcement is now done when you complete an upload instead of being done very time you upload a part.
+- Fixed a performance issue where concurrent multipart part uploads would get rejected.
+
+## 2022-10-26
+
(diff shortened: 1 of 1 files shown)


Frederik-Baetens:
Can you add "Add worker bindings for multipart uploads". This was released a few days ago, and wasn't in the changelog because these changes go through workerd.

content/r2/platform/changelog.md
```diff
@@ -6,6 +6,42 @@ weight: 3
...
+- Fixed a performance issue where concurrent multipart part uploads would get rejected
+
+## 2022-10-26
+
+- Fixed ranged reads for multipart objects with part sizes unaligned to 64KiB
+
+## 2022-10-19
+
+- `HeadBucket` now sets `x-amz-bucket-region` to `auto` in the response
+
```
deadlypants1973:
```suggestion
- Fixed a bug in `ListObjects` where `startAfter` would skip over objects with keys that have numbers right after the `startAfter` prefix.
- Add Worker bindings for multipart uploads.

## 2022-11-17

- Unconditionally return HTTP `206` on ranged requests to match behavior of other S3 compatible implementations.
- Fixed a CORS bug where `AllowedHeaders` in the CORS config were being treated case-sensitively.

## 2022-11-08

- Copying multipart objects via `CopyObject` is re-enabled.
- `UploadPartCopy` is re-enabled.

## 2022-10-28

- Multipart upload part sizes are always expected to be of the same size, but this enforcement is now done when you complete an upload instead of being done very time you upload a part.
- Fixed a performance issue where concurrent multipart part uploads would get rejected.

## 2022-10-26

- Fixed ranged reads for multipart objects with part sizes unaligned to 64 KiB.

## 2022-10-19

- `HeadBucket` now sets `x-amz-bucket-region` to `auto` in the response.

```

content/r2/platform/changelog.md
```diff
@@ -6,6 +6,42 @@ weight: 3
...
+## 2022-10-06
+
+- Temporarily disabled `UploadPartCopy` while we investigate an issue
+
+## 2022-09-29
+
+- Fixed a CORS issue where `Access-Control-Allow-Headers` was not being set for preflight requests
```
deadlypants1973:
```suggestion
- Temporarily disabled `UploadPartCopy` while we investigate an issue.

## 2022-09-29

- Fixed a CORS issue where `Access-Control-Allow-Headers` was not being set for preflight requests.
```

deadlypants1973 (approved):

PR #6863, opened 2022-11-28
[Workers] update references to service environments

admah:


kabirsikand:
I think we need to remove a lot of the references to Environments here. Let's focus on Deployments instead.

admah:
@kabir I updated the `using-services` page and the `bindings` page.

kabirsikand (approved):

PR #6868, opened 2022-11-29
fix: detect invalid range requests correctly

threepointone:
When making a range request, chrome can send a request with only a beginning point (eg: `0-`, note that there's no "end" byte). In the code for the sample R2 worker, we spit the string and check the length; however, splitting this string generates 2 parts anyway ['0', '']. We need to remove empty strings from the array before testing it's length. Hence the proposed fix to filter with a `Boolean` predicate, which removes empty strings.

### DIFF
--- a/content/r2/examples/demo-worker.md
+++ b/content/r2/examples/demo-worker.md
@@ -18,7 +18,7 @@ function parseRange(encoded: string | null): undefined | { offset: number, end: 
     return
   }
 
-  const parts = encoded.split("bytes=")[1]?.split("-") ?? []
+  const parts = encoded.split("bytes=")[1]?.split("-").filter(Boolean) ?? []
   if (parts.length !== 2) {
     throw new Error('Not supported to skip specifying the beginning/ending byte at this time')
   }


threepointone:
This isn't perfect, lemme work on this some more. Thanks @KianNH for the feedback!

deadlypants1973:
> This isn't perfect, lemme work on this some more. Thanks @KianNH for the feedback!

I am going to convert this to a draft for now as you work :)

KianNH:
```ts
function parseRange(encoded: string | null): R2Range | undefined {
  if (encoded === null) {
    return undefined;
  }

  const [start, end] = encoded.split("bytes=")[1]?.split("-") ?? [];

  if (start && end) {
    // range: bytes=offset-length
    return { offset: Number(start), length: Number(end) + 1 };
  }

  if (start && !end) {
    // range: bytes=offset-
    return { offset: Number(start) };
  }

  if (!start && end) {
    // range: bytes=-suffix
    return { suffix: Number(end) };
  }

  return undefined;
}
```

I got this together - which works fine - but there's lots of type narrowing issues to deal with elsewhere like when the `content-range` response header is being set.

image

Within the `if (!start && end)` condition, which maps to `range: bytes=-suffix`, we should return a `416 Range Not Satisfiable` response if `suffix` is `0`.

```ts
if (range && "suffix" in range) {
  if (range.suffix === 0) {
    return new Response("The requested range is not satisfiable", {
      status: 416,
    });
  }
}
```

So there's a check if `range` is present, and also if it contains `suffix` to narrow the union, and then to check if it's `0` - and I hope there's a better way of doing it than the one I tested with.

If `R2GetOptions#range` worked with `Headers` and `R2Bucket#get` returned a `Response` object with all the status codes & headers ready for you, the world would be a better place :^(

kodster28:
Howdy, there hasn't been any action on this in a while + the underlying example itself has changed, so going to close this out.

If the newer example is broken, definitely raise an issue and we'll circle back around on it.

Also, general apologies that this has been sitting for so long.

PR #6876, opened 2022-11-29
[Various] Small tweaks based on search data

kodster28:


content/fundamentals/account-and-billing/account-maintenance/change-super-admin.md
```diff
@@ -0,0 +1,29 @@
+---
+pcx_content_type: how-to
+title: Change Super Admininstrator
+weight: 4
+---
+
```
kodster28, line 6: Moved from another page

data/cache.yml
```diff
@@ -6,7 +6,7 @@ product:
   url: /cache/
 
 meta:
```
kodster28, line 8: Seems to be confusion about cache vs cdn (essentially the same), so adding this so it's visible in Google and fed search results.

patriciasantaana (approved):

PR #6885, opened 2022-11-30
[WAF] Block response content type based on Accept header

pedrosousa:
Addresses PCX-5269.


anikaymaasquesa:
> Addresses PCX-5269.

What si they raison ?

kodster28:
Heyo @pedrosousa, going to close out as this is a few years old.

PR #6789, opened 2022-11-21
[ZT] WARP profiles

ranbel:

PR #6893, opened 2022-12-01
Fix required modes improperly set to "WARP with Gateway"

schultetwin1:
Local Domain Fallback works in either WARP with Gateway or WARP with DoH modes. In either case, the local DNS server can change where the DNS lookups are sent.

You can mode switch no matter which mode you are in

We do captive portal detection in both modes


content/cloudflare-one/connections/connect-devices/warp/exclude-traffic/local-domains.md
```diff
@@ -12,7 +12,7 @@ weight: 6
 
 | Operating Systems | [WARP mode required](/cloudflare-one/connections/connect-devices/warp/#warp-client-modes) | [Zero Trust plans](https://www.cloudflare.com/teams-pricing/) |
 | ----------------- | ----------------------------------------------------------------------------------------- | ------------------------------------------------------------- |
-| All systems       | WARP with Gateway                                                                         | All plans                                                     |
+| All systems       | Any mode                                                                         | All plans                                                     |
```
kkrum, line 15: This needs to be "Gateway with WARP" and "Gateway with DoH". Proxy mode and the currently hidden behind a flag device posture only mode don't support this.

kkrum (changes requested):

ranbel:
Updated the modes in #6789

PR #6898, opened 2022-12-01
Update install-cloudflare-cert.md

deansundquist:
It isn't clear in the documentation that we need to first download the .crt file (for Windows) and the .pem file (for Mac or Linux)


ranbel:
updated formatting in #6952

PR #6902, opened 2022-12-01
[Zero Trust] First round of image and alt text updates

haleycode:


patriciasantaana (approved):

haleycode:
In case this raises questions later, I screwed up and added an unnecessary file from a different branch to this PR, which I then deleted prior to merging the actual content changes.

PR #6967, opened 2022-12-07
[BYOIP] Proxy IPs

patriciasantaana:
PCX-5308


content/byoip/proxy-ips.md
```diff
@@ -0,0 +1,46 @@
...
+title: Proxy IPs
+pcx_content_type: concept
+weight: 6
+---
+
+# Proxy IPs
+
+The Address Map is a data structure which enables customers with BYOIP prefixes or account-level Static IPs to specify which IP addresses should be mapped to a zone when they are proxied on Cloudflare.
+
+The zone must also be proxied or orange-clouded. When a zone is proxied, Cloudflare authoritative DNS will respond with the address on the Address Map. The IP addresses under Records on the zone-level DNS tab will tell Cloudflare how to get to the origin.
```
kodster28: I'd cross link to managing DNS records.

kodster28: Zones aren't proxied (unless I'm mistaken). The records within a zone are proxied. Should also probably link this to https://developers.cloudflare.com/dns/manage-dns-records/reference/

I am more confident about this statement than the other one.

content/byoip/proxy-ips.md
```diff
@@ -0,0 +1,46 @@
...
+
+{{<Aside type="note">}}
+
+You can also enable, disable, and delete Address Maps. This will likely change the IP addresses used for your zones.
+
+{{</Aside>}}
+
+## Immutable Address Maps
+
+Some customers may only use BYOIP and are prohibited from using Cloudflare IP addresses. In this case, Cloudflare will create an immutable, account-wide Address Map to ensure all zones in your account receive BYOIP addresses. These Address Maps cannot be deleted.
```
kodster28: I might sync with whoever your tech lead is to clarify the wording here.

Technically, records that are proxied _respond_ with CF IP addresses, but a DNS record should never be _directly using_ a CF IP address. That'll actually give you an error if you enter it in the dashboard.

Though there may be something different with BYOIP I'm misunderstanding here, so I'm less confident here.

content/byoip/proxy-ips.md
```diff
@@ -0,0 +1,46 @@
...
+
+To specify different addresses for certain zones, [create a new Address Map](/byoip/proxy-ips/#create-address-maps).
```
kodster28:
```suggestion
To specify different addresses for certain zones, [create a new Address Map](#create-address-maps).
```

kodster28: If it's w/in the same page, you only need the anchor link.

kodster28 (changes requested):

kodster28 (commented):
Still some questions, but don't want PTO to block this going live.

kodster28 (approved):

content/byoip/proxy-ips.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Proxy IPs
+pcx_content_type: concept
+weight: 6
+---
+
+# Proxy IPs
+
+The Address Map is a data structure which enables customers with BYOIP prefixes or account-level Static IPs to specify which IP addresses should be mapped to a zone when they are proxied on Cloudflare.
```
tarnfeld:
```suggestion
The Address Map is a data structure enabling customers with BYOIP prefixes or account-level Static IPs to specify which IP addresses should be mapped to a zone's DNS records, when they are proxied through Cloudflare.
```

content/byoip/proxy-ips.md
tarnfeld: I agree, I left a suggestion at the top to this effect as well.

content/byoip/proxy-ips.md
```diff
@@ -0,0 +1,46 @@
...
+
+The zone must also be proxied or orange-clouded. When a zone is proxied, [Cloudflare authoritative DNS](/dns/manage-dns-records/how-to/create-dns-records/) will respond with the address on the Address Map. The IP addresses under Records on the zone-level DNS tab will tell Cloudflare how to get to the origin.
```
tarnfeld:
```suggestion
For an Address Map to take effect, DNS records within the mapped zone must be set to proxied or orange-clouded. When a zone is proxied, [Cloudflare authoritative DNS](/dns/manage-dns-records/how-to/create-dns-records/) will respond with the address(es) on the Address Map. Address Maps do not change how Cloudflare reaches the configured origin, the IP addresses under Records on the zone-level DNS tab continue to instruct Cloudflare how to reach the origin.
```

content/byoip/proxy-ips.md
```diff
@@ -0,0 +1,46 @@
...
+
+Creating an Address Map does not automatically change DNS configuration. DNS Responses only begin to change when a zone or account is added to a map.
```
tarnfeld:
```suggestion
Creating an Address Map does not automatically change DNS configuration. DNS Responses only begin to change when a zone or account is added to a map. Additionally, Address Maps that are not yet **Enabled** will not take effect in DNS responses.
```

tarnfeld: I am unsure if we are using the Enabled or Deployed wording in the UI. Might need to adjust this as needed.

content/byoip/proxy-ips.md
tarnfeld:
```suggestion
Some customers may only proxy zones through BYOIP addresses, and are prohibited from using Cloudflare IP addresses for proxied DNS names. In this case, Cloudflare will create an immutable, account-wide Address Map to ensure all zones in your account receive BYOIP addresses as a fallback. These Address Maps cannot be deleted. It is still possible to create more specific zone-level Address Maps with specific BYOIPs, but without one DNS will fall back to the account-wide Address Map.
```

tarnfeld (commented):
I left a few comments to clarify things.

PR #7004, opened 2022-12-12
Increase queues per account limit from 10 to 100

a-robinson:
I don't know if we have any good place to announce changes like this (e.g. a changelog), but if so we should put it there too because otherwise nobody's going to notice this. If nothing else, we could always put it in the Discord and possibly the weekly Workers release notes (even though this isn't technically a Workers-specific product).


elithrar:
See https://github.com/cloudflare/cloudflare-docs/pull/7159

PR #7022, opened 2022-12-13
[Stream] Added info about live stream transition to VOD

dcpena:
Added info about what happens when a live stream transitions to a recording to address PCX-5201.


content/stream/stream-live/watch-live-stream.md
```diff
@@ -141,3 +141,12 @@ header: Response for an inactive live stream
 ```
 
 When viewing a live stream via the live input ID, the `requireSignedURLs` and `allowedOrigins` options in the live input recording settings are used. These settings are independent of the video-level settings.
+
+## Troubleshooting
+
+When a video transitions from a live stream to the live stream's recording, users may encounter one of the following scenarios:
+
+- If a live stream transitions to the live stream's recording while a user is viewing the video, playback is temporarily interrupted. To fix the issue, reload the player and try viewing the video again.
```
irvinebroque: 1. I don't think it's quite right to say that playback is interrupted — it's that the live stream has ended.
2. We should be specific about whether we are talking about the Stream Player or using one's own player. And guidance that is specific to the Stream Player should live within the Stream Player part of the docs. The reloading guidance is specific to the Stream Player. Anyone implementing a VOD-to-live transition in their own player needs to handle this transition themselves, using https://developers.cloudflare.com/stream/stream-live/replay-recordings/

content/stream/stream-live/watch-live-stream.md
```diff
@@ -141,3 +141,12 @@ header: Response for an inactive live stream
...
+- If a viewer tries to load a video within ~60 seconds of the transition from the live stream to the live stream recording, the viewer may experience issues with playback. To fix the issue, wait at least 60 seconds for the live stream to transition to the record.
```
irvinebroque: Think this still needs to be framed less in terms of "expect there to be problems", and more in terms of "once a live stream ends, poll our API until the video UID of the recording is ready to stream, and then load this manifest" — we should be proactive, tell people what they should do to integrate Stream into their app.

Also — I would not know what "transition the record" means here — think we're lifting from how we talk about our system internally.

content/stream/stream-live/watch-live-stream.md
```diff
@@ -141,3 +141,12 @@ header: Response for an inactive live stream
 ```
 
 When viewing a live stream via the live input ID, the `requireSignedURLs` and `allowedOrigins` options in the live input recording settings are used. These settings are independent of the video-level settings.
+
+## Troubleshooting
+
+When a video transitions from a live stream to the live stream's recording, users may encounter one of the following scenarios:
```
irvinebroque: Similar to comment below and general feedback, not sure this is the right framing here. Think our goal here should be to proactively explain to people how to manage the transition from live to VOD. To me this reads as us explaining poor product behavior.

content/stream/stream-live/watch-live-stream.md
dcpena: I updated this to be more prescriptive and less about explaining an issue. Let me know what you think

content/stream/stream-live/watch-live-stream.md
dcpena: I definitely typo'd that very last sentence, but I also updated the whole thing to be more proactive in how to play back the recording.

content/stream/stream-live/watch-live-stream.md
dcpena: Updated this and also changed the header to focus more on the action the user is trying to complete.

content/stream/viewing-videos/using-the-stream-player/_index.md
```diff
@@ -157,3 +157,12 @@ The `<video>` element does not force the browser to follow the value of this opt
 ## Debug Info
 
 The Stream player Debug menu can be shown and hidden using the key combination `Shift-D` while the video is playing.
+
+## Live stream recording playback
+
+After a live stream ends, a recording is automatically generated and available within 60 seconds. To ensure successful video viewing and playback, keep the following in mind:
```
kodster28, line 163: Would it be possible to combine these two into a partial and reuse? They're pretty close to the same statements?

kodster28 (approved):

Everlag (approved):
Apart from comments from other folks, this looks good to me 👍

content/stream/viewing-videos/using-the-stream-player/_index.md
dcpena: I had the same thought, but I was having trouble coming up with a way to call out the bit about users viewing the Stream player. Suggestions?

PR #7024, opened 2022-12-13
Remove quotes from ssh config

james-w:
sshd config doesn't match usernames when quoted. I don't know how to prove this but https://man.openbsd.org/sshd_config#Match doesn't talk about quoting at all.

I copied the snippet to set up the configuration and spent ages debugging it before realising that the match lines were silently having no effect.

### DIFF
--- a/content/cloudflare-one/_partials/_ssh-usernames.md
+++ b/content/cloudflare-one/_partials/_ssh-usernames.md
@@ -26,7 +26,7 @@ By default, SSH servers authenticate the Unix username against the principals li
 To allow `jdoe@example.com` to log in as the user `johndoe`, add the following to the server's `/etc/ssh/sshd_config`:
 
 ```txt
-Match user 'johndoe'
+Match user johndoe
   AuthorizedPrincipalsCommand echo 'jdoe'
   AuthorizedPrincipalsCommandUser nobody
 ```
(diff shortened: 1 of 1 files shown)


james-w:
It would be great if you could also pass along a feature request related to this. This configuration
works with cli ssh access with `cloudflared`, but doesn't work well with the browser ssh client,
as that tries to log with the principal as the username before the user is prompted to choose
a username. It would be great if it would try (again?) after the username has been entered
so that short-lived certificates could be used when this configuration is needed.

Here's someone who found the same issue: https://community.cloudflare.com/t/short-lived-certificates-and-browser-renderings/292326

Thanks!

itzoey (approved):

ranbel:
@kennyj42 

> It would be great if you could also pass along a feature request related to this. This configuration works with cli ssh access with `cloudflared`, but doesn't work well with the browser ssh client, as that tries to log with the principal as the username before the user is prompted to choose a username. It would be great if it would try (again?) after the username has been entered so that short-lived certificates could be used when this configuration is needed.
> 
> Here's someone who found the same issue: https://community.cloudflare.com/t/short-lived-certificates-and-browser-renderings/292326
> 
> Thanks!

james-w:
Thanks!

kennyj42:
Acknowledge! I'll get a ticket in for my team to review. Thank you for flagging this James.

PR #7031, opened 2022-12-14
turnstile: add turnstile.ready

Le0Developer:
this is in prod.

maybe add a note that is async loading should be preferred unless absolutely necessary? 🤷


worenga:
Can we add this to changelog too?

patriciasantaana (approved):

PR #7043, opened 2022-12-15
update suggested NODE_VERSION for SvelteKit from 17 to 16

Conduitry:
We never intended to still be claiming to support Node 17, and we recently merged a PR (https://github.com/sveltejs/kit/pull/8174) to tighten up our `pkg.engines.node` to exclude Node 17. No one should be using Node 17 for anything now anyway because it reached EOL on June 1.

### DIFF
--- a/content/pages/framework-guides/deploy-a-svelte-site.md
+++ b/content/pages/framework-guides/deploy-a-svelte-site.md
@@ -161,7 +161,7 @@ When deploying with [`@sveltejs/adapter-cloudflare`](https://www.npmjs.com/packa
 | Production branch     | `main`                  |
 | Build command         | `npm run build`         |
 | Build directory       | `.svelte-kit/cloudflare`|
-| Environment Variables | `NODE_VERSION: 17 `|
+| Environment Variables | `NODE_VERSION: 18 `|
 
 </div>
 
--- a/content/pages/framework-guides/deploy-a-svelte-site.md
+++ b/content/pages/framework-guides/deploy-a-svelte-site.md
@@ -161,7 +161,7 @@ When deploying with [`@sveltejs/adapter-cloudflare`](https://www.npmjs.com/packa
 | Production branch     | `main`                  |
 | Build command         | `npm run build`         |
 | Build directory       | `.svelte-kit/cloudflare`|
-| Environment Variables | `NODE_VERSION: 18 `|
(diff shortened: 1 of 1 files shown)


Conduitry:
I've just heard that Node 18 is not available. If someone specifies `NODE_VERSION=16`, which version do they get? Is it at least 16.14?

content/pages/framework-guides/deploy-a-svelte-site.md
```diff
@@ -161,7 +161,7 @@ When deploying with [`@sveltejs/adapter-cloudflare`](https://www.npmjs.com/packa
 | Production branch     | `main`                  |
 | Build command         | `npm run build`         |
 | Build directory       | `.svelte-kit/cloudflare`|
-| Environment Variables | `NODE_VERSION: 17 `|
+| Environment Variables | `NODE_VERSION: 18 `|
```
Conduitry:
```suggestion
| Environment Variables | `NODE_VERSION: 16 `|
```

Conduitry:
I've updated this PR to instead switching to Node 16, as that apparently provides 16.19.0, which is new enough for SvelteKit.

WalshyDev (approved):

content/pages/framework-guides/deploy-a-svelte-site.md
```diff
@@ -161,7 +161,7 @@ When deploying with [`@sveltejs/adapter-cloudflare`](https://www.npmjs.com/packa
 | Production branch     | `main`                  |
 | Build command         | `npm run build`         |
 | Build directory       | `.svelte-kit/cloudflare`|
-| Environment Variables | `NODE_VERSION: 17 `|
+| Environment Variables | `NODE_VERSION: 16 `|
```
WalshyDev:
```suggestion
| Environment Variables | `NODE_VERSION: 16` |
```

fredguest:
@Conduitry if someone has a `.node-version` file in their Sveltekit app, Pages will prioritize that value over the `NODE_VERSION` environment variable, so anyone that had `17.x.x` in their `.node-version` file will not be helped by updating `NODE_VERSION`, they will need to change the value in `.node-version` to `16.19.0`. Not sure if that should be added to the Cloudflare docs or the Sveltekit docs but I'm sure it will throw some people off after bumping to Sveltekit 1.0.1

Issue #6048, opened 2022-09-28
Subscribe to a page


WalshyDev:
### New thing

We have quite a few pages which it'd be good for users to "subscribe" to. Examples of that are like the Workers changelog, R2 changelog and other such pages.

it would be awesome to subscribe with both an email + webhook URL

This can be done with internal tools but I'm not too familiar with them. I'd love to help out if you need it and would like to do this.
Personally, I think this would be a really nice win for users. Tracking these updates are hard right now

kodster28:
Ooh, super cool idea.

Seems like this [SO post](https://stackoverflow.com/questions/9732779/how-to-watch-only-a-directory-in-a-github-repository) provides some ideas that could be useful. 

In terms of the overall technical help, the email / webhook subscription might be a bit out of my wheelhouse.... but I have played around a bit with the github API and looking for commit histories to specific files, so theoretically could be a bit more helpful there.

kodster28:
There are definitely some limitations with the new implementation (we're really just providing links to the GH-generated RSS feed), but it should provide at least some visibility into file / folder-level changes.

You add the RSS icon to the top toolbar + the `<head>` meta by specifying either `rss: file` (just watch that file) or `rss: folder` (watch all changes in the folder) in the page frontmatter.

Examples:
- [Folder](https://github.com/cloudflare/cloudflare-docs/blob/production/content/waf/change-log/_index.md?plain=1#L7)
- [File](https://github.com/cloudflare/cloudflare-docs/blob/production/content/stream/changelog.md?plain=1#L5)

These links will be broken if you attempt to do a `rss: file` on an `_index` page, which I think would be a relative edge case.

PR #7051, opened 2022-12-16
[Docs site] Rss feed test

kodster28:
Provides an MVP solution for #6048


kodster28:
Thoughts on this, @WalshyDev and @GregBrimble?

From what I can tell about Hugo's default behavior, page-level RSS feeds aren't really supported (more meant to be when there are new posts on a blog). So really this only works for pages organized like https://rss-feed-test.cloudflare-docs-7ou.pages.dev/waf/change-log/, which add a new page for each change that'll then be picked up in the RSS feed.

Not perfect by any means and will require restructuring our current changelogs to match... but think it provides a way forward w/o a ton of custom development.

For visibility, adding an RSS icon to pages that have an RSS feed enabled.
Screenshot 2022-12-16 at 10 57 42 AM

Also adding a top-level page at https://rss-feed-test.cloudflare-docs-7ou.pages.dev/rss/ to list all the available options.

kodster28:
Doing a touch more digging, think that the structure of their [.GitInfo](https://gohugo.io/variables/git/#lastmod) variable excludes the possibility of an easy soln for page-level changes. It only stores the last commit info per a page, probably based off the use case of `.LastMod`.

kodster28:
Actually... maybe I'm overcomplicating this... it'd probably be easier just to use GitHub's built-in RSS feed per a file or folder.

https://github.com/cloudflare/cloudflare-docs/commits/production/content/waf/change-log.atom
https://github.com/cloudflare/cloudflare-docs/commits/production/content/stream/changelog.md.atom

This seems like the least amount of work + the most flexibility.

layouts/_default/rss.xml
```diff
@@ -0,0 +1,39 @@
...
+    {{ printf "<atom:link href=%q rel=\"self\" type=%q />" .Permalink .MediaType | safeHTML }}
+    {{- end -}}
+    {{ range $pages }}
+    <item>
+      <title>{{ .Title }}</title>
+      <link>{{ .Permalink }}</link>
+      <pubDate>{{ .Lastmod.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
+      {{ with .Site.Author.email }}<author>{{.}}{{ with $.Site.Author.name }} ({{.}}){{end}}</author>{{end}}
+      <guid>{{ .Permalink }}</guid>
+      <description>{{ .Summary | html }}</description>
```
pedrosousa: While this description seems to contain relevant information like CVE numbers, the format is not great: it's essentially a bunch of table cells bundled together into a sentence. 
Maybe we could require a useful summary sentence in the Markdown front matter for these changelog pages, which could include information like CVE numbers or affected rule IDs?

kodster28:
When you specify an `rss: folder` or `rss: file` property in the frontmatter of a page, that page gets:

-  an RSS icon generated in the top toolbar pointing to the [GH RSS feed](https://github.com/cloudflare/cloudflare-docs/commits/production/content/waf/change-log.atom)
- a link added to the HTML head to the same feed
- Added to the [RSS feeds page](https://rss-feed-test.cloudflare-docs-7ou.pages.dev/rss/), which is also linked in the top toolbar of home pages for visibility

There's an edge case where this logic will break if you specify `rss: file` on `_index.md` pages, but a) we can document it and b) feel like that's not an onerous burden on anyone.

Added examples for [WAF](https://rss-feed-test.cloudflare-docs-7ou.pages.dev/waf/change-log/) -- which is a folder -- and [Stream](https://rss-feed-test.cloudflare-docs-7ou.pages.dev/stream/changelog/) -- which is a file.

TBD --> Set up standards for changelog commits so they're more easily identifiable in the logs

layouts/partials/topbar.tools.html
```diff
@@ -10,12 +14,47 @@
...
+        <a class="Link Link-without-underline RSS" href="{{$rssFeed}}" target="_blank">
+          <svg viewBox="0 0 485 560" role="img" aria-labelledby="title-1234" xmlns="http://www.w3.org/2000/svg">
+            <title id="title-1234">RSS icon</title>
+            <path d="M0 64C0 46.3 14.3 32 32 32c229.8 0 416 186.2 416 416c0 17.7-14.3 32-32 32s-32-14.3-32-32C384 253.6 226.4 96 32 96C14.3 96 0 81.7 0 64zM128 416c0 35.3-28.7 64-64 64s-64-28.7-64-64s28.7-64 64-64s64 28.7 64 64zM32 160c159.1 0 288 128.9 288 288c0 17.7-14.3 32-32 32s-32-14.3-32-32c0-123.7-100.3-224-224-224c-17.7 0-32-14.3-32-32s14.3-32 32-32z"></path>
+          </svg>
+        </a>
+      </div>
+
+      <span class="Tooltip" role="tooltip" position="bottom-end">
+        View RSS Feed
```
pedrosousa: To manage expectations, how about using "View GitHub RSS feed" instead? Would that be too confusing?
My main issue with not mentioning GitHub is that all feed entry links will point to GH pull requests.

layouts/partials/rss-icon.html
```diff
@@ -0,0 +1,5 @@
+<a class="Button DocsMarkdown--resource" href="/rss">
```
pedrosousa: Are we using this file? Also, the `rss.svg` file doesn't seem to exist.

layouts/shortcodes/rss-list.html
```diff
@@ -0,0 +1,26 @@
...
+{{ range .Site.Pages }} 
+{{- $data := (index $.Site.Data .Section) -}}
+{{- $title := default .Title .Params.meta.title  -}}
+{{- $rssPath := "https://github.com/cloudflare/cloudflare-docs/commits/production/content" -}}
+{{- $contentPath :=  replaceRE `\/$` "" .RelPermalink -}}
+{{- $folderPlaceholder := "" -}}
+{{ $rssFeed := "" }}
+{{ with .Params.rss }}
+
+  {{ if eq . "folder" }}
```
pedrosousa: Not a serious issue, but we're not accounting for invalid `rss` values in frontmatter (different from `folder` and `file`). In this case, we would create a link that wouldn't point to an RSS file.

pedrosousa (approved):

layouts/shortcodes/rss-list.html
kodster28: Should be fixed in the latest commit, both for the icon + the link the `<head>`

PR #7069, opened 2022-12-19
Update bulk-redirects.md

apatwary12:
Theres a lot of factors this doesn't cover and the split '/redirect' doesn't work out of the box.

### DIFF
--- a/content/workers/examples/bulk-redirects.md
+++ b/content/workers/examples/bulk-redirects.md
@@ -22,7 +22,7 @@ const redirectMap = new Map([
 
 async function handleRequest(request) {
   const requestURL = new URL(request.url);
-  const path = requestURL.pathname.split(/[?#]/)[0];
+  const path = requestURL.pathname;
   const location = redirectMap.get(path);
   if (location) {
     return Response.redirect(location, 301);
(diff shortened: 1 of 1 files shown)


content/workers/examples/bulk-redirects.md
```diff
@@ -22,7 +22,7 @@ const redirectMap = new Map([
 
 async function handleRequest(request) {
   const requestURL = new URL(request.url);
-  const path = requestURL.pathname.split('/redirect')[1];
+  const path = requestURL.pathname.split(/[?#]/)[0];
```
WalshyDev: This is actually redundant, the query params and hash will never be in `pathname`
E.g.
```
const url = new URL('https://example.com/abcdef#abc?def=true');

console.log(url.pathname)
> /abcdef
```

this original code doesn't make much sense either though. It seems to just want the path so we'll just get the path
```suggestion
  const path = requestURL.pathname;
```

WalshyDev (changes requested):

WalshyDev (approved):

deadlypants1973 (approved):

PR #7099, opened 2022-12-23
Add DDoS change log entries for 2022-12-23

antoinecordelle:


content/ddos-protection/change-log/http/scheduled-changes.md
```diff
@@ -22,13 +22,13 @@ layout: list
   </thead>
   <tbody>
     <tr>
-      <td>2022-12-06</td>
-      <td>2022-12-19</td>
-      <td>...cd1b2ec7</td>
```
pedrosousa, line 27: Shouldn't this PR also contain a new page about this `...cd1b2ec7` rule?

antoinecordelle: We ended up not changing the corresponding rule so in this case, not new page is necessary.

jdesgats (approved):

PR #7107, opened 2022-12-26
add default-src none - more restrictive

billmcilhargey:
as defined by the CSP standards:

for the hardening is to set default-src 'none';

If no CSP is specified for a directive, it falls back to default-src.

If there is no default-src CSP, it does not fall back on any directive and as a result it allows everything.

As far as everything that falls back (that is allowed) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src

By adding none - this should be a more restrictive policy

### DIFF
--- a/content/pages/platform/headers.md
+++ b/content/pages/platform/headers.md
@@ -158,5 +158,5 @@ filename: _headers
   X-Content-Type-Options: nosniff
   Referrer-Policy: no-referrer
   Permissions-Policy: document-domain=()
-  Content-Security-Policy: script-src 'self'; frame-ancestors 'none';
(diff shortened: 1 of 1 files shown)


WalshyDev:
Hey,

Thanks for the PR however, this will have unintended side effects for anyone who just copy-pastes. For example, `style-src` will not default to `none` as it is not defined. This section as a whole isn't really great... every site differs in requirements. Wouldn't be a bad idea to just remove this @deadlypants1973 

However, for now. I will close this PR. I appreciate the desire for stricter CSP though!

PR #7108, opened 2022-12-26
add a worker-src suggestion

billmcilhargey:
Depending on how your restrictive your CSP policy is, worker-src may need to be added to support the bot detection

worker-src 'self' is required for some policy implementations to support Javascript Implementations


patriciasantaana:
Hi @computeronix ! Thanks for bringing this to our attention. As long as the `script-src` (which is always required) is fine, the `worker-src` should not matter.
I will be closing this issue now but please feel free to open another in case you would like to show a scenario in which it's required.

PR #7115, opened 2022-12-28
Add note about the `x-forwarded-for` HTTP request header

lougayou:
As per CUSTESC-24546, this header is NOT removed by the "Remove visitor IP headers" rule


lougayou:
Actually, following the comments in the Jira, we need to leave this line there.
But let's add a comment then!
Maybe something like:
- x-forwarded-for (Please note that only the x-forwarded-for header added by another CDN or proxy is removed: the initial client IP from the eyeball will be present in the header)

lougayou:
Hi @pedrosousa 
Should we close this PR? Should I raise a new one?
Or can you edit the content based on my latest comment?
Thanks!

pedrosousa:
Hi @lougayou 
It's on my personal backlog! I'll update the text, check the final version with Maurizio, and then merge.

pedrosousa (approved):

PR #7116, opened 2022-12-28
[WAF] example of how the counting expression works

satanTime:


pedrosousa (approved):

pedrosousa:
Thank you, @satanTime!

PR #7135, opened 2023-01-02
Add Queue Event Listener

omarkhairy21:


content/workers/runtime-apis/add-event-listener.md
```diff
@@ -57,3 +57,13 @@ addEventListener('scheduled', event => {
   event.waitUntil(handleScheduled(event));
 });
 ```
+
+### Queue Listener
+```js
```
deadlypants1973:
```suggestion
### Queue Listener

A Queue listener acts as [a consumer for one or more queues](https://developers.cloudflare.com/queues/javascript-apis/#consumer).

```js
```

content/workers/runtime-apis/add-event-listener.md
```diff
@@ -57,3 +57,13 @@ addEventListener('scheduled', event => {
...
+addEventListener("queue", (event) => {
+  event.waitUntil(handleMessages(event));
```
rts-rob, line 67: I would not recommend using `waitUntil` for placing messages onto the queue. If `handleMessages(event)` throws an error, the client will not know about it, as the producer worker will have already returned (presumably) a 200.

rts-rob (changes requested):
Thanks for these fixes!

We should replace `waitUntil` with a sync call as shown elsewhere in the docs so that errors are not silently lost. Otherwise g2g.

content/workers/runtime-apis/add-event-listener.md
elithrar: I would amend this to:

```js
addEventListener("queue", (event) => {
  await handleMessages(event);
});
```

content/workers/runtime-apis/add-event-listener.md
```diff
@@ -57,3 +57,13 @@ addEventListener('scheduled', event => {
   event.waitUntil(handleScheduled(event));
 });
 ```
+
+### Queue Listener
```
elithrar, line 61: Under this header, we should add more context on what a "Queue Listener" is - e.g.

> A Queue listener acts as [a consumer for one or more queues](https://developers.cloudflare.com/queues/javascript-apis/#consumer).

content/workers/runtime-apis/add-event-listener.md
deadlypants1973, line 67:
```suggestion
addEventListener("queue", (event) => {
  await handleMessages(event);
});
```

content/workers/runtime-apis/add-event-listener.md
```diff
@@ -57,3 +57,17 @@ addEventListener('scheduled', event => {
...
+
+A Queue listener acts as [a consumer for one or more queues](https://developers.cloudflare.com/queues/javascript-apis/#consumer).
+
+```js
+addEventListener("queue", (event) => {
+  await handleMessages(event);
+});
+});
```
deadlypants1973, line 68:
```suggestion
```

deadlypants1973 (approved):

content/workers/runtime-apis/add-event-listener.md
```diff
@@ -57,3 +57,16 @@ addEventListener('scheduled', event => {
...
 });
 ```
+
+### Queue Listener
+
+A Queue listener acts as [a consumer for one or more queues](https://developers.cloudflare.com/queues/javascript-apis/#consumer).
+
+```js
+addEventListener("queue", (event) => {
+  await handleMessages(event);
```
a-robinson: This is not valid code. The function defined here is not async, and so calling `await` inside it won't do what you want.

And you can't just make it async, because `addEventListener` doesn't handle async functions. The code was correct as originally written.

a-robinson (changes requested):

content/workers/runtime-apis/add-event-listener.md
deadlypants1973:
```suggestion
addEventListener("queue", (event) => {
  event.waitUntil(handleMessages(event));
```

a-robinson (approved):

PR #7138, opened 2023-01-03
Added R2 to destination section

angelampcosta:


content/logs/get-started/api-configuration.md
```diff
@@ -87,6 +87,7 @@ You can specify your cloud service provider destination via the required **desti
 As of May 2022, defining a unique destination for a Logpush job will no longer be required. As this constraint has been removed, you can now have more than one job writing to the same destination.
 {{</Aside>}}
 
+*  **Cloudflare R2**: endpoint name + bucket path + account ID + access key id + secret key; for example: `r2://<BUCKET_PATH>/account-id=<ACCOUNT_ID>&access-key-id=<R2_ACCESS_KEY_ID>&secret-access-key=<R2_SECRET_ACCESS_KEY>`
```
soheiokamoto: What's endpoint name here?

angelampcosta: I guess `r2`? I copied the example that we have here https://developers.cloudflare.com/logs/get-started/enable-destinations/r2/#manage-via-api

soheiokamoto: I think that might have been copied from s3-compatible, but with `r2://`, I don't think there is `endpoint` parameter.

angelampcosta: I see. So the `destination_conf` for R2 it's only `bucket path`, `account id`, `R2 access key id` and `R2 secret access key`, like this `r2://<BUCKET_PATH>/account-id=<ACCOUNT_ID>&access-key-id=<R2_ACCESS_KEY_ID>&secret-access-key=<R2_SECRET_ACCESS_KEY>`?

soheiokamoto: Yes.

marciocloudflare (approved):

PR #7142, opened 2023-01-03
[Docs site] If no page meta description specified, use the first 150 characters of the text

kodster28:
Known issues:
- Leaves in footnote numbers (seems like a minor concern)


kodster28:
Logic 
- If there’s a meta description specified in the page frontmatter, use it
- Otherwise, use the first `<p>` tag from the document body
- If there are errors with the `<p>` tag (an image is a known one), show the meta description from the product data file

pedrosousa (approved):

PR #7145, opened 2023-01-03
Add callout to also add the Require Rule

kennyj42:
You also need to enforce the specific Login Method via the policy. Not sure if we also want to add a picture of this. I can provide is we deem necessary.

cc @ranbel

### DIFF
--- a/content/cloudflare-one/tutorials/azuread-conditional-access.md
+++ b/content/cloudflare-one/tutorials/azuread-conditional-access.md
@@ -61,6 +61,8 @@ In this section, you will create an Access Idenetify Provider per Application or
 
 ![List of identity providers highlighting the Azure AD option](/cloudflare-one/static/zero-trust-security/azuread-access-policies/access-app-idp.png)
 
+5. Create an Access Policy Require rule with the corresponding IdP as the Login Method
+
 Repeat the two sections above for each individual application or group of applications. 
 
 ## Azure AD and Conditional Access configuration


ranbel:
Does a user need to enable API permissions for this Azure AD enterprise application?

kennyj42:
Yea, they need to do that but I think we call that out in the IdP setup section for AzureAD?

ranbel:
Great! I'm just checking for any differences between these instructions and the IdP instructions before linking out.

Regarding "Select the corresponding IdP that was just configured for this specific Conditional Access Policy." -- is it supposed to be one IdP for each **application** that's protected by Conditional Access policies? (rather than one IdP per Conditional Access policy)

kennyj42:
Yea, that's correct. you have to create one IdP instance for each application you would like to have a different conditional access criteria.

content/cloudflare-one/tutorials/azuread-conditional-access.md
```diff
@@ -61,6 +61,8 @@ In this section, you will create an Access Idenetify Provider per Application or
 
 ![List of identity providers highlighting the Azure AD option](/cloudflare-one/static/zero-trust-security/azuread-access-policies/access-app-idp.png)
 
+5. Create an Access Policy Require rule with the corresponding IdP as the Login Method
```
ranbel, line 64: Why do you need a Require rule if "Azure AD - Customer management portal" is the only accepted identity provider?

ranbel: Also what would you put as the Include rule in that policy?

ranbel:
added to #7209

PR #7154, opened 2023-01-04
[Stream] Add entry for non-video upload rejection to changelog.md

bkrebsbach:
Adds a new changelog entry for the unique error code returned when Stream detects a non-video upload.

### DIFF
--- a/content/stream/changelog.md
+++ b/content/stream/changelog.md
@@ -6,6 +6,16 @@ weight: 11
 
 # Changelog
 
+## 2023-01-04
+
+### Earlier detection (and rejection) of non-video uploads
+
+Cloudflare Stream now detects non-video content on upload using [the POST API](/stream/uploading-videos/upload-video-file/) and returns a 400 Bad Request HTTP error with code `10059`.
+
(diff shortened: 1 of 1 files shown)


dcpena (approved):

content/stream/changelog.md
```diff
@@ -6,6 +6,12 @@ weight: 11
 
 # Changelog
 
+## 2023-01-04
+
+### Earlier detection (and rejection) of non-video uploads
+
+Cloudflare Stream now detects non-video content on upload and returns a Bad Request error with code `10059`.
```
irvinebroque: 1. In the response from which specific API endpoints? (implicit, can we make it explicit)
2. Do the API docs need updating?

irvinebroque: Context that might be helpful here:

Previously, if you or one of your users attempted to upload a file that is not a video (ex: an image), the request to upload would appear successful, but then fail to be encoded later on.

With this change, Stream responds to the upload request with an error, allowing you to give users immediate feedback if they attempt to upload non-video content.

irvinebroque (approved):
🚀

PR #7217, opened 2023-01-11
[Pages] Add Pages changelog

WalshyDev:

PR #7220, opened 2023-01-11
[Docs site] Added RSS feeds where appropriate

kodster28:
Following up on #7053 and #7217, adding RSS feed links to the other changelog pages in our docs.

@GregBrimble, also think the compatibility dates page makes sense, but let me know if you think otherwise.


GregBrimble (approved):
Adding this to the compatibility dates page makes sense to me. I'll let others weigh in on their respective products.

deadlypants1973 (approved):

kodster28:
Haven't heard any objections from the folks we've reached out to internally, so going to merge in.

If there are any objections later, it's very easy to remove.

PR #7227, opened 2023-01-11
Wrong heading level in pages docs

jimhawkridge:
I spotted this while using these docs: this should be at the same level as the other binding types.


WalshyDev (approved):
Oops haha, ty!

PR #7260, opened 2023-01-13
[Docs site] Allow directory listings to pull in page descriptions to provide more context on navigation pages

kodster28:


sh471:
> 

Thank you so much

layouts/partials/head.meta.html
```diff
@@ -4,9 +4,7 @@
 
 {{- $pt := default .Context.Title .Context.Params.meta.title -}}
 {{- $title := printf "%s · %s" $pt $DATA.meta.title -}}
-{{ $raw_summary := index (split (index (split .Context.Content "</p>") 0) "<p>") 1 | plainify | htmlUnescape | chomp }}
-{{ $summary := replaceRE `External link icon\s+Open external link\s?` "" $raw_summary }}
-{{- $description:= truncate 160 (default $META.description (default $summary .Context.Params.meta.description)) -}}
+{{- $description := (partial "seo-meta-description.html" (dict "meta" $META "content" .Context.Content "params" .Context.Params "truncate" 150 "plainify" true )) -}}
```
pedrosousa: Maybe we could could declare the max size used to truncate (currently `150`) as a variable at the beginning of the template, so that folks don't have to visually parse the entire code block searching for it.

layouts/shortcodes/directory-listing.html
```diff
@@ -1,14 +1,22 @@
...
   {{- $x := .RelPermalink -}}
   {{- $self := eq $x $current -}}
   {{- $delta := sub (len (split $x "/")) $len -}}
   {{- if and (not $self) (hasPrefix $x $current) (lt $delta 2) (not .Params.hidden) -}}
   <li>
-    {{ $link := printf "[%s](%s)" .Title .RelPermalink -}}
+    {{- if $showDescriptions -}}
+      {{- $link = printf "[%s](%s): %s" .Title .RelPermalink (partial "seo-meta-description.html" (dict "content" .Content "params" .Params "truncate" $truncate )) -}}
+    {{- else -}}
+      {{ $link = printf "[%s](%s)" .Title .RelPermalink -}}
```
pedrosousa:
```suggestion
      {{- $link = printf "[%s](%s)" .Title .RelPermalink -}}
```
We're removing all spaces on the left except here. I'd also remove them unless there's a reason not to.

content/ssl/reference/_index.md
```diff
@@ -6,6 +6,6 @@ weight: 10
 
 # Reference
 
-For more on Cloudflare SSL/TLS, see these articles:
+For more on Cloudflare SSL/TLS, refer to these articles:
 
-{{<directory-listing>}}
+{{<directory-listing showDescriptions=true char_limit=300 >}}
```
pedrosousa, line 11: In this case, at least two descriptions end with a colon, which is odd. I guess we'll manually define meta descriptions when we find this kind of issue?
Regarding truncation, we should probably edit the meta description (or make sure the page has a short first paragraph) to avoid of getting an ellipsis (`...`) here.

content/ssl/edge-certificates/additional-options/_index.md
```diff
@@ -8,12 +8,4 @@ weight: 9
 
 Once you set up SSL/TLS on your application, you can adjust the following settings in **SSL/TLS** > **Edge Certificates**:
 
-*   [Total TLS](/ssl/edge-certificates/additional-options/total-tls/): {{<render file="_total-tls.md">}}
```
pedrosousa, line 11: Note: I didn't check if we could delete any of these definition-only partials.

pedrosousa (approved):

content/ssl/reference/_index.md
kodster28: Either that, or restructure the first text block.

We _theoretically_ could get fancier and avoid the `...`, but I'd prefer to keep the `...` since it's a built-in function for Hugo and it doesn't seem like there's a super easy way to do this otherwise.

Think it might make sense to define a larger starting `char_limit` for directory listings, but then still offer the input param.

content/ssl/edge-certificates/additional-options/_index.md
kodster28: You were right, they needed to be removed.

PR #7279, opened 2023-01-17
fix humidity field name

boemekeld:
double1 = temperature
double2 = humidity


jimhawkridge (approved):
Well spotted!
Thanks very much for the PR.

PR #7293, opened 2023-01-17
[R2[ Added use cases for R2 to Overview

dcpena:
Added the use cases for R2 to the Overview page. (Use cases come from the video Gift added to the Get Started page.)

Addresses PCX-5680.

### DIFF
--- a/content/r2/_index.md
+++ b/content/r2/_index.md
@@ -12,6 +12,15 @@ meta:
 # Cloudflare R2 documentation
 
 Cloudflare R2 Storage allows developers to store large amounts of unstructured data without the costly egress bandwidth fees associated with typical cloud storage services.
+
+Use R2 for: 
+
+- Static hosting
+- R2 as storage for podcast episodes
+- Storage for cloud-native applications
+- Cloud storage for web content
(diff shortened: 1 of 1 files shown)


Frederik-Baetens (approved):
Good idea! What do you think about a "- and much more" at the end? Or something similarly open ended? Not important, so you can merge as is if you want :smile:

content/r2/_index.md
```diff
@@ -12,6 +12,15 @@ meta:
 # Cloudflare R2 documentation
 
 Cloudflare R2 Storage allows developers to store large amounts of unstructured data without the costly egress bandwidth fees associated with typical cloud storage services.
+
+Use R2 for: 
+
+- Static hosting
+- R2 as storage for podcast episodes
```
bercknash, line 19: Suggest shortening to "storage for podcast episodes" since the heading is "Use R2 for", so the second R2 is redundant.  Assuming this is allowed and we're not pushing people to Streams for this... (???)

patriciasantaana (approved):

Frederik-Baetens:
LGTM?

PR #7299, opened 2023-01-18
important update on SSO setup

stefanogramm:
Change: 
{{<Aside type="note">}}

  Mail providers (e.g. `gmail.com`) will not qualify.
  SSO is enforced for all users whose **emails end in an owned given domain**.
  
{{</Aside>}}

PCX-5683


ranbel:
added to #8221

PR #7301, opened 2023-01-18
[Workers]: Add Module Workers Format to Workers Docs using Tabswitcher component

lauragift21:
This PR adds updated examples using the module workers format, also updates the learning, platform and Runtime APIs sections where module format was not present.


kristianfreeman (approved):
Spot checked these in production. Looks good to me! Great job.

PR #7306, opened 2023-01-18
The option for access applications is called configure not edit

alankemp:


ranbel (approved):
thanks for making the change!

PR #7310, opened 2023-01-18
[ZT] update Microsoft endpoint manager instructions

ranbel:


content/cloudflare-one/identity/devices/service-providers/microsoft.md
```diff
@@ -34,8 +34,10 @@ To retrieve those values:
 1. Fill in a description and how long the secret should be valid.
 1. After completing the form, immediately copy the resulting secret. This will be your Client Secret.
 1. Go to **API Permissions** and select **Add permission**.
+1. Select **Microsoft Graph**.
 1. Select **Application permissions**.
-1. Search for `DeviceManagementManagedDevices` and select the `Read` permission.
```
spatel-cf, line 38: We should leave this, and instead add a sub-bullet that states that "you may need to request admin consent to enable Cloudflare to read results from the Microsoft Graph API. See more on permissions: https://learn.microsoft.com/en-us/graph/permissions-reference#intune-device-management-permissions"

content/cloudflare-one/identity/devices/service-providers/microsoft.md
```diff
@@ -34,8 +34,10 @@ To retrieve those values:
...
+1. Add `DeviceManagementManagedDevices.Read.All`.
+1. To enable the permission, select **Grant admin consent**.
```
spatel-cf: This step may not always be necessary for all users, but certain users with certain permissions may require admin consent to be granted.

content/cloudflare-one/identity/devices/service-providers/microsoft.md
ranbel: The reference page says Admin Consent Required for `DeviceManagementManagedDevices.ReadAll`.

content/cloudflare-one/identity/devices/service-providers/microsoft.md
ranbel: I think the vast majority of users who are setting this up will have admin permissions. Even if you have admin permissions, you still have to click **Grant admin consent**.

ranbel: Would the API permissions process be similar to what we have in the Azure IdP steps? https://developers.cloudflare.com/cloudflare-one/identity/idp-integration/azuread/#2-configure-api-permissions-in-azure

spatel-cf: @ranbel yes the steps are the same, except the only permissions needed for the service-to-service posture integration is DeviceManagementManagedDevices. In which case, it would help to over-document that "Grant admin consent" might be required depending on your permissions.

patriciasantaana (approved):

PR #7313, opened 2023-01-19
Added match count to DLP profile setup

cloudflare-noelle:
As per SHIP-6699, match counts will be added to DLP profiles


ranbel (approved):

content/cloudflare-one/policies/filtering/http-policies/data-loss-prevention/_index.md
```diff
@@ -48,6 +48,12 @@ Cloudflare DLP provides [predefined profiles](/cloudflare-one/policies/filtering
 DLP scans will not start until you [create an HTTP policy](#2-create-a-dlp-policy).
 {{</Aside>}}
 
+### Advanced settings
+
+You can configure a DLP Profile with the following optional settings:
+
+- **Match count**: the number of times that any entry in the profile can be detected before an action is triggered, such as blocking or logging. For example, if you select a match count of 10, DLP will only log or block a request if the HTTP body contains 11 or more matching strings. Detections do not have to be unique.
```
ranbel: @cloudflare-noelle Does the updated description look ok?

cloudflare-noelle: Looks great, I realized I should've written "any enabled entry in the profile" in the first line though.  Can you update that and then we are good to go?

ranbel:
```suggestion
- **Match count**: the number of times that any enabled entry in the profile can be detected before an action is triggered, such as blocking or logging. For example, if you select a match count of 10, DLP will only log or block a request if the HTTP body contains 11 or more matching strings. Detections do not have to be unique.
```

PR #7332, opened 2023-01-20
Fix Svelte prerender option in svelte.config.js

BMorearty:
The docs for using SvelteKit's static adapter were based on an outdated value for the `prerender` option in svelte.config.js. Using `default: true` doesn't work any more. It resulted in a build error when I tried it.

### DIFF
--- a/content/pages/framework-guides/deploy-a-svelte-site.md
+++ b/content/pages/framework-guides/deploy-a-svelte-site.md
@@ -196,8 +196,8 @@ const config = {
 +     }
 +   ),
 +   prerender: {
-+     // This can be false if you're using a fallback (i.e. SPA mode)
-+     default: true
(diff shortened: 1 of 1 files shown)


content/pages/framework-guides/deploy-a-svelte-site.md
```diff
@@ -196,8 +196,8 @@ const config = {
 +     }
 +   ),
 +   prerender: {
-+     // This can be false if you're using a fallback (i.e. SPA mode)
-+     default: true
++     // Default option shown. This can be [] if you're using SPA mode.
```
deadlypants1973:
```suggestion
+     // Default option shown. This can be [] if you are using SPA mode.
```

deadlypants1973 (approved):

PR #7342, opened 2023-01-20
fix "Wrangler version"

tomByrer:
Needs extra -- before 'version', but some folks prefer less typing ;)


tomByrer:
seems I had Wrangler v1 installed, even though it was a 'fresh' install... confusing.

PR #7356, opened 2023-01-23
Update Deprecated Variables

pew:
Updating examples using CF_ACCOUNT_ID etc., replacing them with CLOUDFLARE_ACCOUNT_ID. Removes warnings in wrangler


KianNH:
It looks like you've updated the documentation for the old Wrangler 1 - Wrangler 1 doesn't support `CLOUDFLARE_ACCOUNT_ID`.

PR #7386, opened 2023-01-24
[Docs site] - Adding structured data for FAQ pages

kodster28:
PCX-5723 --> Adding [FAQ structured content](https://developers.google.com/search/docs/appearance/structured-data/faqpage#content-guidelines) as something we can add to specific pages.

Confirmed that the structure works as evidenced by:
- https://validator.schema.org/#url=https%3A%2F%2Ffaq-structured-data.cloudflare-docs-7ou.pages.dev%2Fturnstile%2Ffrequently-asked-questions%2F
- https://search.google.com/test/rich-results/result/r%2Ffaq?id=_f229RgbV7ZSnuK8B0aDKA


layouts/shortcodes/faq-question.html
```diff
@@ -0,0 +1,15 @@
...
+{{- $ID := $text | urlize -}}
+
+<h{{$level}} id="{{ $ID }}" itemprop="name"><span class="DocsMarkdown--header-anchor-positioner">
+    <a
+    class="DocsMarkdown--header-anchor Link Link-without-underline"
+    href="#{{ $ID }}"
+    >&#8203;​</a
+    >
+</span>
+{{ $text }}
```
pedrosousa: Maybe we should run the heading text through `markdownify` (there are a few inline code blocks in titles of FAQ/Troubleshooting pages).

layouts/shortcodes/faq-question.html
```diff
@@ -0,0 +1,15 @@
+{{- $level := (default 2 (.Get "headingLevel")) | string -}}
+{{- $text := (default "" (.Get "text")) | markdownify -}}
```
pedrosousa: I'd move `markdownify` to somewhere after the `urlize` call (line 3), so that the latter doesn't have any HTML to process.
* `markdownify` before `urlize` — A `` `something` `` heading would have ID `#codesomething%2fcode`
* `markdownify` after `urlize` — A `` `something` `` heading would have ID `#something`

We could call it on line 12 instead: `{{ $text | markdownify }}`

pedrosousa (approved):

Issue #5917, opened 2022-09-20
[R2] Workers Runtime APIs layout & formatting


KianNH:
### Which Cloudflare product does this pertain to?

R2

### Existing documentation URL(s)

- https://developers.cloudflare.com/r2/data-access/workers-api/workers-api-reference/

### Section that requires update

Layout and formatting.

### What needs to change?

The current layout, whilst satisfactory for a type reference, will struggle to scale as the APIs gain more options and require more explanations/walkthroughs on how to properly utilise them.

As an example, the addition of the new checksums supported in `R2PutOptions` means that either:

1) Every checksum is listed on it's own, with an aside that only one can be used at a time.
2) The type for the 'checksum' in `R2PutOptions` becomes `md5 | sha1 | sha256 | sha384 | sha512`

Either way, it becomes very verbose and will easily confuse people.

### How should it change?

Taking inspiration from well-regarded documentation layouts such as `docs.rs` (i.e https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html), we should come up with a consistent structure on documenting Worker's Runtime APIs that we can then extend to refactoring the pages for KV, Durable Objects and the like.

We want a good middle-ground between providing a reference, and providing a how-to, without going too in-depth that the material would be better split off into it's own how-to page.

As an example of one potential layout that was discussed by myself and @Frederik-Baetens:

<img width="643" alt="image" src="https://user-images.githubusercontent.com/94662631/191375548-9e5244fe-85be-4ca1-929a-4af793699941.png">

Using the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element, this allows us to:

1) Provide a lot of value without cluttering the rest of the page, as it's expandable.
2) Preserve the 'searchability' of the content, as CTRL+F on the majority of browsers will 'look inside' a `<details>` element.
3) Additionally include small examples, not in-depth or niche enough for their own page, under additional `<details>` elements.

The 'signature' of the function is shown to provide a type reference to those who 'know what they're doing' and just need a quick look at what the types are, with further explanations of what those types actually are further down.

The rest of the tab explains what dictates which type will be returned, and then goes on to include some bite-sized examples.

<img width="654" alt="image" src="https://user-images.githubusercontent.com/94662631/191375800-c884bf3c-dade-43a3-8aed-f3e48f1d6b21.png">

### Additional information

Original discussion: https://github.com/cloudflare/cloudflare-docs/pull/5897

[...]

PR #7396, opened 2023-01-25
[R2] Document multipart upload part size considerations in rclone example

sdnts:
plus Max number of parts allowed in a multipart upload

### DIFF
--- a/content/r2/examples/rclone.md
+++ b/content/r2/examples/rclone.md
@@ -67,17 +67,30 @@ $ rclone tree r2demo:my-bucket-name
 ## Upload and retrieve objects
 
 The [rclone copy](https://rclone.org/commands/rclone_copy/) command can be used to upload objects to an R2 bucket and vice versa - this allows you to upload files up to the 5 TB maximum object size that R2 supports.
-  
+
 ```sh
 # Upload dog.txt to the user-uploads bucket
-$ rclone copy dog.txt r2demo:user-uploads/dog.txt
+$ rclone copy dog.txt r2demo:user-uploads/
 $ rclone tree r2demo:user-uploads
 # /
 # ├── foobar.png
 # └── dog.txt
 
 # Download dog.txt from the user-uploads bucket
-$ rclone copy r2demo:user-uploads/dog.txt dog.txt
+$ rclone copy r2demo:user-uploads/dog.txt .
+```
+
+### A note about multipart upload part sizes
+
+For multipart uploads, part sizes can significantly affect the number of Class A operations that are used, which can alter how much you end up being charged.
+Every part upload counts as a separate operation, so larger part sizes will use fewer operations, but might be costly to retry if the upload fails. Also consider that a multipart upload is always going to consume at least 3 times as many operations as a single `PutObject`, because it will include at least one `CreateMultipartUpload`, `UploadPart` & `CompleteMultipartUpload` operations.
+
+Balancing part size depends heavily on your use-case, but these factors can help you minimize your bill, so they are worth thinking about.
+
+You can configure rclone's multipart upload part size using the `--s3-chunk-size` CLI argument. Note that you might also have to adjust the `--s3-upload-cutoff` argument to ensure that rclone is using multipart uploads. Both of these can be set in your configuration file as well. Generally, `--s3-upload-cutoff` will be no less than `--s3-chunk-size`.
+
+```sh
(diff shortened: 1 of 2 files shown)


content/r2/examples/rclone.md
```diff
@@ -80,6 +80,19 @@ $ rclone tree r2demo:user-uploads
...
+Every part upload counts as a separate operation, so larger part sizes will use fewer operations, but might be costly to retry if the upload fails.
+
+Balancing part size depends heavily on your use-case, but can be worth thinking about.
+You can configure rclone's multipart upload part size using the `--s3-chunk-size` CLI argument. Note that you might also have to adjust the `--s3-upload-cutoff` argument to force rclone to use multipart uploads. Both of these can be set in your configuration file as well.
+
+```sh
+# Set --s3-upload-cutoff to 0MB to force multipart uploads for all files, and --s3-chunk-size to 100MB to set the part size
+$ rclone copy long-video.mp4 r2demo:user-uploads/long-video.mp4 --s3-upload-cutoff=0M --s3-chunk-size=100M 
+```
+{{</Aside>}}
```
sdnts: This recommendation is valid pretty much anywhere we talk about (multipart) uploads, not just rclone. Is there a better place to put this in?

jahands: If there is a better place, I'd still keep it here for users who are new to Object Storage and getting started with rclone.

Nit: Recommending `--s3-upload-cutoff=0M` seems weird to me.
Generally I would set this to no less than `--s3-chunk-size` because it's using an extra API call in some cases.

Example:
`rclone copy 75MB_file.zip r2demo:user-uploads --s3-upload-cutoff=0M --s3-chunk-size=100M` will cause rclone to use a multipart upload with 3 Class A operations (CreateMultipartUpload, UploadChunk, CompleteMultipartUpload) when it could have uploaded the file with a single PUT.

Also of note: `rclone copy` copies the file to the target "directory", while `rclone copyto` copies to the exact path.

So in this example, the resulting object will actually be `r2demo:user-uploads/long-video.mp4/long-video.mp4`

sdnts: Great points, serves me right for `--dry-run`ing the command only 😬
I'll fix the command and add a better cutoff.

dcpena: Can we change this from an aside to an H2? If it's something relevant to multiple places, making it an H2 will make it easier to link to.

For the short term, let's add it here, and I will make note that this is also relevant to multipart uploads and account for that once I begin restructuring the docs and identifying the gaps. That work will start happening soon.

jahands: When testing I highly recommend passing `-vv` - it allows you to see all of the individual chunks it's uploading

KianNH: > This recommendation is valid pretty much anywhere we talk about (multipart) uploads, not just rclone. Is there a better place to put this in?

A cool pattern that the rest of the R2 examples follow is using partials - like for the [AWS API credentials](https://github.com/cloudflare/cloudflare-docs/blob/production/content/r2/_partials/_keys.md) - so you can just include them with Hugo anywhere they're necessary.

Whilst the overarching concern about multipart chunk sizes applies to any multipart upload, I suspect you can only provide guidance for `rclone` and the `aws` CLI as the rest are SDKs - which would involve you designing your own chunking from the ground up & managing multipart state yourself - so only the general 'use big chunk sizes if possible' guidance is applicable there.

dcpena: Great suggestion, @KianNH! We definitely try to utilize partials where we can in the docs, and the partials folder is already added for this project too which is nice!

KianNH: Somewhat off-topic for this but vaguely related to your mention of restructuring the docs - partials would be great for the Runtime API docs to provide more value with examples without having a giant markdown file that is hard to maintain.

It's the same path I used for https://github.com/cloudflare/cloudflare-docs/issues/5917 - with each method, like `R2Bucket.get(...)`, being a partial: you can kinda see it here, ignore the mess (also known as my WIP shortcodes) - https://github.com/KianNH/cloudflare-docs/commit/b9d96c75f5dde2c408770c2572fe7b4304db7106

sdnts: Thanks for the recommendations all, I've just addressed them, let me know if all's good!

@dcpena I decided to use an H3 instead because H2s were used as headings on the page (like for `List buckets & objects` etc.). I wanted this to show up visually "under" `Upload and retrieve objects`

jahands:
LGTM! Thanks for adding this!

dcpena (approved):

PR #7416, opened 2023-01-26
[Email routing] Correct example for Sender Rewriting Scheme in postmaster

jbd7:
Fetched correct format from https://en.wikipedia.org/wiki/Sender_Rewriting_Scheme and validated with actual CF behavior in production. 

The element rewritten by CF seems to be be `RETURN-PATH` instead of `MAIL FROM`, it may be worth adding it to the documentation.


kodster28:
Howdy @jbd7, sorry it's taken us so long to action this.

Looks like the example has been removed from the FAQ, so closing out this PR since the issue is no longer present.

PR #7417, opened 2023-01-26
[R2] Fix the aws4fetch example for presigned URLs

KianNH:


sdnts:
Nice work simplifying the example!

KianNH:
Ready for review!

Changes:

- Removed `service` and `region` from the `AwsClient` initialisation - these are not required anymore for R2.
- Removed the 'authorisation' example and substituted with a comment - the existing one was a bit confusing personally.
- Moved to using the URL constructor to make it a bit more 'obvious' what is happening where.
- Moved the `X-Amz-Expires` from Headers to URLSearchParams, as using Headers was incorrect.
- Overall linting & tidy-up.

dcpena (approved):

PR #7453, opened 2023-01-29
Makefile: bump minify to 2.12.4

SpencerIsGiddy:
Bumps minify from 2.9.24 to 2.12.4


kodster28:
This was actually addressed in #9701.
(thanks for the contribution though & sorry this was open for so long!)

PR #7463, opened 2023-01-30
Create pricing.md

ngayerie:
Adding a pricing section under Image Resizing (from SPM-1172), so the pricing is more visible that its current location lost in the "feature overview" page It will also allow adding more details or examples later on if needed


kodster28:
Heyo @ngayerie, we generally don't include pricing in dev docs.

If there's feedback about the marketing page, that'd best be provided in a WEBX ticket.

PR #7484, opened 2023-01-31
[Fundamentals] Added Transfer-Encoding and Content-Length headers to deprecation list

angelampcosta:


schien (approved):

content/fundamentals/api/reference/deprecations.md
```diff
@@ -129,3 +129,10 @@ Deprecated API:
 
 Replacement:
 Accounts API
+
+## Transfer-Encoding and Content-Length headers 
+**End of life date: March 31, 2023**
+
+Previously, `RFC2616` allowed the use of Transfer-Encoding and Content-Length in request headers at the same time. RFC7230 supersedes `RFC2616` and prohibits the use of Transfer-Encoding and Content-Length in request headers at the same time because they can cause HTTP request smuggling vulnerabilities.
+
+Starting on March 31 2021, Cloudflare will decline requests with both Transfer-Encoding and Content-Length in request headers.
```
pedrosousa:
```suggestion
Previously, RFC2616 allowed the use of `Transfer-Encoding` and `Content-Length` HTTP headers in the same request. RFC7230 supersedes RFC2616 and prohibits the use of `Transfer-Encoding` and `Content-Length` headers in the same request because they can cause HTTP request smuggling vulnerabilities.

Starting on March 31, 2023, Cloudflare will decline requests with both `Transfer-Encoding` and `Content-Length` HTTP headers.
```
We usually don't format RFCs as monospaced text, but I think we should do it for HTTP header names (our style guide only mentions "HTTP content-type values", though).

pedrosousa (approved):

PR #7485, opened 2023-01-31
[R2] document that multipart uploads need to have parts of the same size

Frederik-Baetens:


content/r2/data-access/workers-api/workers-api-reference.md
```diff
@@ -208,7 +208,7 @@ A multipart upload can be completed or aborted at any time, either through the S
 
 - {{<code>}}uploadPart(partNumber{{<param-type>}}number{{</param-type>}}, value{{<param-type>}}ReadableStream{{</param-type>}}|{{<param-type>}}ArrayBuffer{{</param-type>}}|{{<param-type>}}ArrayBufferView{{</param-type>}}|{{<param-type>}}string{{</param-type>}}|{{<param-type>}}Blob{{</param-type>}}) {{<type>}}Promise\<{{<param-type>}}R2UploadedPart{{</param-type>}}>{{</type>}}{{</code>}}
 
-  - Uploads a single part with the specified part number to this multipart upload.
+  - Uploads a single part with the specified part number to this multipart upload. Each part except the last one needs to have the same size.
```
sdnts: Would it be nice to mention that this is only enforced during `CompleteMultipartUpload` or is that too much information / unnecessary?

Frederik-Baetens: Hmm, that's something that might be worth annotating on the complete operation instead, but our current docs structure doesn't lend itself to specifying all the ways that something might throw an exception. It's worth documenting nonetheless, but I don't have a clean way of doing so right now.

For the complete, you'd want to say that it throws when the parts are non-uniform in size, when the etags don't match up, when you use a non-existent part number etc... but that's all a bit too verbose to fit into the structure we have now without expandable fields.

Frederik-Baetens: Adopting a structure like Kian suggested here might make more room for that: https://github.com/cloudflare/cloudflare-docs/issues/5917

CC @dcpena

dcpena: To Frederik's point, documenting every possible exception can be a little much, but that's a good opportunity to ensure that any error messages that get thrown are as specific as can be so that users know why something happened and what they need to do next.

I think I prefer something like, "Each part must be uniform in size with an exception for the final part which can be smaller." I'm making an assumption about the last part being smaller if the others have to be uniform in size. Feel free to correct me.

Frederik-Baetens: :+1: good suggestion. Your assumption of the last part needing to be smaller is indeed correct. I've copied your wording.

dcpena (approved):

PR #7500, opened 2023-02-01
Universal SSL date limit

AlphaK5:
Hey Team,

I'm working on this SPM-1547 to update the documentation, since it's not updated since Oct22.

Thanks!


cicku:
There isn't much need to change the wording here.

kodster28:
Since it's a past date, it implies that the migration has already happened.

PR #7501, opened 2023-02-01
Data Retention proposal section

AlphaK5:
Hello!
Working on the SPM-1518, and proposing a new section with the summary of the Data Retention to be more clear how it works. Also, added more context in the first section, about the sampling. ty!

### DIFF
--- a/content/fundamentals/data-products/about-analytics.md
+++ b/content/fundamentals/data-products/about-analytics.md
@@ -12,7 +12,11 @@ In an effort to make analytics an ubiquitous component of all Cloudflare's produ
 
 You can access root-level analytics that give you an overview of metadata related to your Cloudflare account, analytics related to specific properties and products, and the GraphQL API that gives you more control over how you visualize the analytics and log information available on the Cloudflare dashboard.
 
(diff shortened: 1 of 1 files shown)


cicku:
1. It is not in a small number of cases. We already documented the adaptive sampling behavior in a dedicated page.
2. I'm not sure why Enterprise is mentioned so much in the addition.

kodster28:
Heyo @AlphaK5, thanks for the suggestion, but I'd nudge you towards creating a PCX ticket for this work instead.

There are also a couple other issues:
- Links to other pages in our docs should always be the relative path (not absolute)
- You're duplicating content word-for-word that's in another section. This should likely be edited (per @cicku's comment above), but also separated out into a [render shortcode](https://wiki.cfdata.org/display/Prod/Markdown+components#Markdowncomponents-{{%3Crenderfile%3E}}).
- Think we might need a more elegant solution to the various data retention limits mentioned here. These vary greatly by plan and product.
- Likely makes more sense as it's own page.

angelampcosta:
@AlphaK5 I'll close the PR for now and will follow up in the ticket.

PR #7502, opened 2023-02-01
Using Cloudflare API on Windows CMD or Powershell

AlphaK5:
Hey team, working on the SPM-947.
Adding a section dedicated for the customers that have Windows CMD or powershell, and how to use the API calls to endpoints that have `--data payload`.

### DIFF
--- a/content/fundamentals/api/how-to/make-api-calls.md
+++ b/content/fundamentals/api/how-to/make-api-calls.md
@@ -15,27 +15,49 @@ $ curl "https://api.cloudflare.com/client/v4/zones/<ZONE_ID>" \
 
 Never send or store your API token secret in plaintext. Also be sure not to check it into code repositories, especially public ones.
 
+## Making API calls on Windows
 
-## Using Cloudflare API on Windows CMD or Powershell
+### Using a Command Prompt window
 
-In order to use Cloudflare API on Windows CMD or Powershell, you will need to replace the single `'` quote with `"` quote, and add another `"` next to `"` quote in the API endpoints that have `--data payload`. Otherwise, you'll get a JSON formatting error.
+To use the Cloudflare API on a Command Prompt window, you must use double quotes (`"`) as string delimiters instead of single quotes (`'`).
 
+A typical `POST` request will be similar to the following:
 
-On MAC terminal or Unix shell, the correct syntax would be:
-
-```
-GET "https://api.cloudflare.com/client/v4/zones/<ZONE_ID>/logs/control/retention/flag" -d'{"flag":true}'
-```
-
-On Windows OS, you need to replace any single-quote `'` with double-quotes `"` and replace any double-quotes `"` with additional double-quote `""`, as demonstrated in the next example:
-
-```
-GET "https://api.cloudflare.com/client/v4/zones/<ZONE_ID>/logs/control/retention/flag" -d"{""flag"":true}"
+```txt
+C:\> curl -X POST "https://api.cloudflare.com/client/v4/zones/<ZONE_ID>/logs/control/retention/flag" -H "Authorization: Bearer <API_TOKEN>" -d "{""flag"": true}"
 ```
 
+To escape a double quote character in a request body (for example, a body specified with `-d` or `--data` in a `POST` request), prepend it with another double quote (`"`) or a backslash (`\`) character.
 
-You can also try to install the [Microsoft Visual Studio](https://visualstudio.microsoft.com/) along with the [REST Client extension](https://marketplace.visualstudio.com/items?itemName=humao.rest-client) will allow for much easier interactions with the Cloudflare API, or use the [the *json* processor jq](https://developers.cloudflare.com/logs/tutorials/parsing-json-log-data/) for Windows to get a formatted output.
+To break a single command in two or more lines, use `^` as the line continuation character at the end of a line:
 
+```txt
+C:\> curl -X POST ^
+"https://api.cloudflare.com/client/v4/zones/<ZONE_ID>/logs/control/retention/flag" ^
+-H "Authorization: Bearer <API_TOKEN>" ^
+-d "{""flag"": true}"
+```
+
+### Using PowerShell
+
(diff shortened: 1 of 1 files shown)


cicku:
There is no reason to install Visual Studio and a plug-in not developed by the official Microsoft team. Instead, once you install Powershell 7 or preview from the Store, it should solve most of the issue.

Even jq is not the standard solution once you have PS7.

pedrosousa:
@cicku Can you please check the proposed changes again? I think it's useful to separate the Command Prompt window guidance from the PowerShell one.

cicku:
Here are some points:

1. `curl` is not a command always pre-installed in a Windows system so there is a chance you cannot run anything in cmd, but `Invoke-WebRequest` has been there for a long time since PS 3. If you wish to add a paragraph on this, you need to clearly document the requirements.
2. **`Invoke-WebRequest` isn't really needed for the API call, you should just use `Invoke-RestMethod`. `Invoke-WebRequest` is only useful if we need to locate the rayID, like `curl -I`.**
3. The same as `jq`, PS has `ConvertFrom-Json`. Starting from PS 6.2, `ConvertFrom-Json` supports in depth parsing, this will be useful if we have nested JSON output. But, when to specify `jq` or `ConvertFrom-Json`? See the below examples.
4. There are quirks with PS on converting timestamps, it is not mentioned as well. Do you need the raw output, or not?
5. Escaping with PS should be double checked.
6. The example call of log retention is not desired. It is an _Enterprise only_ feature, but this page is a part of the fundamentals.
7. If there are issues with a specific version of Powershell, it is recommended to upgrade to the latest stable or preview: https://apps.microsoft.com/store/detail/powershell/9MZ1SNWT0N5D / https://apps.microsoft.com/store/detail/powershell-preview/9P95ZZKTNRN4. In another word, I don't see any attempt to specify the requirement of PS.

---

An example command to retrieve the SSL certificate status in a PS way with standard expansion:

```
Invoke-RestMethod -URI https://api.cloudflare.com/client/v4/zones/<zone_id>/ssl/certificate_packs?ssl_status=all -Method 'GET' -ContentType 'application/json' -Headers @{'X-Auth-Email'='<email>';'X-Auth-Key'='<key>'} | select -ExpandProperty result

id                    : <redacted>
type                  : advanced
hosts                 : {<redacted>}
primary_certificate   : <redacted>
status                : active
certificates          : {@{id=<redacted>; hosts=System.Object[];
                        issuer=GoogleTrustServicesLLC; signature=SHA256WithRSA; status=active;
                        bundle_method=ubiquitous; zone_id=<redacted>; uploaded_on=2/1/2023
                        7:13:27 PM; modified_on=1/30/2023 8:01:53 PM; expires_on=4/30/2023 7:59:33 PM; priority=}}
created_on            : 1/30/2023 7:59:33 PM
validity_days         : 90
validation_method     : txt
certificate_authority : google

[..snip..]
```

Here is the `curl`+`jq` result in UNIX:

```
{
  "result": [
    {
      "id": "<redacted>",
      "type": "advanced",
      "hosts": [
        "<redacted>"
      ],
      "primary_certificate": "<redacted>",
      "status": "active",
      "certificates": [
        {
          "id": "<redacted>",
          "hosts": [
            "<redacted>"
          ],
          "issuer": "GoogleTrustServicesLLC",
          "signature": "SHA256WithRSA",
          "status": "active",
          "bundle_method": "ubiquitous",
          "zone_id": "<redacted>",
          "uploaded_on": "2023-02-01T19:22:57.400764Z",
          "modified_on": "2023-01-30T20:01:53.375620Z",
          "expires_on": "2023-04-30T19:59:33.000000Z",
          "priority": null
        }
      ],
      "created_on": "2023-01-30T19:59:33.001670Z",
      "validity_days": 90,
      "validation_method": "txt",
      "certificate_authority": "google"
    },
```

---

Leveraging "Convertfrom-Json" for the raw output similar to the above UNIX result (if you do not properly specify the depth, the result could be truncated since the default depth is 2, but some of the API outputs may be deeper):

```
Invoke-RestMethod -URI https://api.cloudflare.com/client/v4/zones/<zone_id>/ssl/certificate_packs?ssl_status=all -Method 'GET' -ContentType 'application/json' -Headers @{'X-Auth-Email'='<email>';'X-Auth-Key'='<key>'}  |  ConvertTo-Json -Depth 5
```

Our environment:

```
PS X:\Windows\System32> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.4.0-preview.1
PSEdition                      Core
GitCommitId                    7.4.0-preview.1
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

pedrosousa:
@cicku I believe I addressed most of you feedback.

1\. curl is included in recent Windows 10 and 11 versions (Windows 7 and 8.1 are no longer supported). Nevertheless, I added a link for folks using other versions wishing to download it and install it.
2\. / 3. There's now an example with `Invoke-RestMethod` and `ConvertFrom-Json`. I still kept the curl example for PowerShell because it's much easier to adapt one of the many curl examples we have in the docs to work in PowerShell than to use PowerShell's proper/best way of doing it. 
4\. Added a link about `ConvertFrom-Json`'s DateTime parsing (I suppose this was the behavior you wanted to highlight).
5\. Not sure what you mean by "should be double checked". Did you find any error in our current suggestions?
6\. Updated the API operation to the one your proposed for GET request examples. For PATCH examples, I used a simple operation that requires a short JSON payload. Unfortunately, I couldn't find a simple GET and POST/PATCH operation that work with tokens instead of API keys (which would be preferable).
7\. Added a note about recommended PowerShell versions.

cicku:
Approved.

marciocloudflare (approved):

pedrosousa:
Thank you for bringing this up, @AlphaK5. I've changed your initial proposal quite a bit, but we're good to go now.

PR #7512, opened 2023-02-01
Add logpush frequency

kyouheicf:
## Frequency 

The frequency at which Cloudflare sends batches of logs to your destination. Setting frequency to `high` sends your logs in larger quantities of smaller files. Setting frequency to `low` sends logs in smaller quantities of larger files.

- `high`
  - every 30 seconds or every 100,000 records, whichever comes first.
  - More than one file might be pushed per 30-second period or per 100,000 records.
- `low`
  - every 5 minutes or every 1,000,000 records, whichever comes first.
  - More than one file might be pushed per 5-minute period or per 1,000,000 records


kodster28:
These secrets aren't a concern and we've already raised this with SIRT at SIRT-9329.

May be worth just creating a new PR from production and re-making your changes.

soheiokamoto:
`frequency` was undocumented parameter, and we will replace with `max_upload_*` parameters.
Please close this PR without merging.

PR #7534, opened 2023-02-03
[Rules] Update Bulk Redirects (rules and runtime workflow)

### LINKED ISSUE #7533: Clarify Bulk Redirect priorities

Rules

https://developers.cloudflare.com/rules/url-forwarding/bulk-redirects/concepts/

The concepts would need to list additional specifications.

### What needs to change?

Some specifications are missing and cause customers to have to wait for list to be submitted to find out about errors. If specs are more granular, a large CSV will be cleaner.

### How should it change?

The concepts would need to list additional specifications:
- Source URL have to be unique, cannot have duplicate source URL
- More specific source URL wins against a less specific URL.
for example example.com/test/foo will get redirected before evaluating example.com/test

### Additional information

_No response_

pedrosousa:
Fixes #7533.

### DIFF
--- a/content/rules/url-forwarding/bulk-redirects/concepts.md
+++ b/content/rules/url-forwarding/bulk-redirects/concepts.md
@@ -55,7 +55,7 @@ URL redirects are the list items of Bulk Redirect Lists.
 
 ## Bulk Redirect Lists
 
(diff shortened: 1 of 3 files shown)


orium (approved):
LGTM

marciocloudflare (approved):

PR #7536, opened 2023-02-03
Update reference how to import and export records

hannesb0:


hannesb0:
@kodster28 we got rid of the "Advanced" button, so all other references (if there are any) should be updated, too.
This is how it looks now:
image

kodster28 (approved):

PR #7544, opened 2023-02-03
[WIP] PCX-4950 OneTrust Integration for Cookie Compliance

tsmith512:
This is a work-in-progress PR which will require testing and adjustment before it can be deployed. Reach out to `@tsmith` internally with any questions or refer to the Jira ticket for discussion.


tsmith512:
This works properly with the OneTrust `test` scripts (a8959f3), but the feature branch environment linked above seems to set the production env var because the prod script, not the test script, is included in the header --- and that will only work on the production hostname. Worth noting that this fails silently. This should be good to go, but we should review it on a call together so I can demo both states. We may also want to consider having branch builds set a development mode env, but that may have other side-effects.

kodster28:
Technically, this is also blocked by #7668, since the changes to the homepage footers will introduce merge conflicts and/or we don't want this change to get overwritten accidentally.

Currently working to get ^^^ merged in and can help with the conflicts if needed!

kodster28 (approved):

PR #7566, opened 2023-02-07
[W4P] Cleanup Dynamic Dispatch Worker

KianNH:
Since there's no types for the Dispatch Namespace binding, moved the example to be `.js` rather than `.ts` (it was already JavaScript code rather than TypeScript anyhow).

Ran the existing Worker through Prettier.


KianNH:
https://f06713c9.cloudflare-docs-github.pages.dev/cloudflare-for-platforms/workers-for-platforms/get-started/dynamic-dispatch/

WalshyDev (approved):
We should probably tab switcher this with what the type looks like too (and just like add to workers types)
But that doesn't need to be now

PR #7580, opened 2023-02-07
Modify `dev` and `build` setup by adding post-Hugo processing with Vite

penalosa:
This is a fairly big change, but it should be pretty beneficial, I hope! The _initial_ motivation for this was that I wanted to see syntax-highlighted code blocks in dev, but in a classic case of scope creep this has gone a fair bit further than that.

In brief, it adds [Vite](https://vitejs.dev/) as a post-Hugo step for build and dev. This has a few major benefits:
- Syntax highlighting in dev
- Proper bundling, which means that Mermaid isn't included on pages that don't have Mermaid diagrams
- More ergonomic setup for authoring JS components in the docs (specifically Vue components, as those were already in use)

The way this is setup is that `npm run dev` runs `hugo` and `vite` concurrently, with Vite pointing at Hugo's output folder. Vite then processes the output HTML files from Hugo. This enables syntax highlighting in development mode. Previously, syntax highlighted code blocks were added to the docs in a `postbuild` npm script—all that's changed is that Vite now runs that same script on Hugo's output HTML in development (and for the production build, which brings the development preview and the production build closer together). The syntax highlighting should be visually identical to the previous implementation. (`vite.config.ts`)

Another change is to follow the [recommendation in Hugo's docs](https://gohugo.io/content-management/diagrams/) for Mermaid diagrams, which treats them as code blocks. This means that the Hugo build process can tell if a page has a Mermaid diagram, and only inject the Mermaid script on relevant pages, which speeds up loading on the vast majority of pages that don't have Mermaid diagrams.

As part of post-processing the Hugo output with Vite, which has a stricter HTML parser than browsers, I also found a few errors in the docs. Most were minor, including `&lt` instead of `&lt;`, but some were more impactful. For instance, some Stream pages (`content/stream/viewing-videos/av1-playback.md`) had embedded figures that used React syntax (`className` etc...) rather than plain HTML, which meant that their aspect ratio wasn't being respected.

Because Vite is a modern bundler, which supports things like Vue natively, this change also meant that Vue single file components could be used, which yield higher performance that the string templates that were in use previously. I've added a shortcode:

```html
{{ <Vue name="DashboardLanding" /> }}
```

which loads a Vue SFC from `./components/DashboardLanding.vue` (or the path corresponding to the specified name), along with loading the `vue` library, and mounts it in place (which means there's no need to create an empty `<div>` with a id that has to be targeted). I've migrated the three main UI components that were in use previously to Vue SFCs (as a bonus, they're no in TypeScript!), and this should be extensible to allow adding different component frameworks in the future if needed (that's why the shortcode is called `<Vue>`, not `<Component>`—we could add a `<Qwik>` or `<Svelte>` shortcode in future to allow loading of different components).


layouts/shortcodes/worker-starter.html
```diff
@@ -17,13 +18,5 @@
...
-  <div class="WorkerStarter--command">
-    <div class="WorkerStarter--command-copy-button-wrapper">
-      <button class="Button" data-clipboard="{{ $command }}">Copy</button>
-    </div>
-
-    <div class="WorkerStarter--command-codeblock-wrapper">
-      {{- .Page.RenderString (printf "```txt\n%s\n```" $command) -}}
-    </div>
-  </div>
+  {{- .Page.RenderString (printf "```sh\n$ %s\n```" $command) -}}
```
mrbbot, line 21: Are these wrapper `div`s not needed now? What were they doing before?

vite.config.ts
```diff
@@ -0,0 +1,45 @@
...
+import glob from "glob";
+import { highlight } from "./bin/prism.config";
+import vue from "@vitejs/plugin-vue";
+const renderCodeBlock = (): PluginOption => {
+  return {
+    name: "html-transform",
+    transformIndexHtml: {
+      order: "pre",
+      handler: (html: string) => {
+        const parsedHtml = parse(html);
```
mrbbot: Could we use https://www.npmjs.com/package/html-rewriter-wasm here for the [`lol`](https://github.com/cloudflare/lol-html)s?

components/DynamicLearningPath.vue
```diff
@@ -0,0 +1,223 @@
...
+            <span v-if="element.type === 'note'">{{ element.title }}</span>
+          </h2>
+          <p v-if="element.estimated_time" class="durationEstimate">
+            ~{{ element.estimated_time }} mins
+          </p>
+        </div>
+        <div v-if="element.description" v-html="element.description"></div>
+        <details v-if="element.pages">
+          <summary v-if="element.type === 'module'">
+            Contains {{ element.pages.length }} units
```
mrbbot, line 153: I think this was a bug before, but this will show `Contains 1 units` if the `length` is 1.

components/DynamicLearningPath.vue
```diff
@@ -0,0 +1,223 @@
...
+          </summary>
+          <summary v-if="element.type === 'note'">
+            Contains {{ element.pages.length }} related resources
```
mrbbot, line 156: As above.

content/dns/zone-setups/subdomain-setup/setup.md
```diff
@@ -10,7 +10,7 @@ meta:
 
 When using a [subdomain setup](/dns/zone-setups/subdomain-setup/), the steps to create a child domain depend on the parent domain's setup and whether the child domain already exists.
 
-<div class="mermaid">
+```mermaid
```
mrbbot, line 13: Hmmm, https://0d78ae83.cloudflare-docs-7ou.pages.dev/dns/zone-setups/subdomain-setup/setup/ doesn't show a rendered diagram for me?

content/fundamentals/get-started/concepts/accounts-and-zones.md
```diff
@@ -8,7 +8,7 @@ weight: 4
 
 Within the Cloudflare ecosystem, there are three organizing concepts that control where specific settings live: user profiles, accounts, and zones.
 
-<div class="mermaid">
+```mermaid
```
mrbbot, line 11: ...and https://0d78ae83.cloudflare-docs-7ou.pages.dev/fundamentals/get-started/concepts/accounts-and-zones/. I suspect this is a problem with all of them. 😕

content/tutorials.html
```diff
@@ -42,3 +44,5 @@ <h2>Tutorials</h2>
       <div class="CoveoPager"></div>
       <div class="CoveoResultsPerPage" data-initial-choice="25"></div>
     </div>
+  </div>
+</div>
```
mrbbot, line 48:
```suggestion
</div>

```

layouts/_default/_markup/render-codeblock-mermaid.html
```diff
@@ -0,0 +1,4 @@
+<div class="mermaid">
+    {{- .Inner }}
+</div>
+{{ .Page.Scratch.Set "hasMermaid" true }}
```
mrbbot:
```suggestion
{{ .Page.Scratch.Set "hasMermaid" true }}

```

layouts/_default/_markup/render-codeblock.html
```diff
@@ -0,0 +1,12 @@
...
+    <pre data-language="{{ .Type }}" data-code="{{ .Inner | base64Encode  }}">
+    </pre>
+    <div id="vue-{{ .Page.Scratch.Get "componentId" }}">
+    </div>
+    <script type="module">
+        import { createApp } from 'vue'
+        import Component from "@/CodeCopy.vue"
+        createApp(Component).mount('#vue-{{ .Page.Scratch.Get "componentId" }}', true)
+    </script>
+</div>
```
mrbbot, line 4:
```suggestion
</div>

```

layouts/shortcodes/Vue.html
```diff
@@ -0,0 +1,10 @@
+{{- $name := .Get "name" -}}
+{{ .Page.Scratch.Add "componentId" 1 }}
+<div id="vue-{{ .Page.Scratch.Get "componentId" }}">
+    {{- .Page.RenderString .Inner -}}
+</div>
+<script type="module">
+    import { createApp } from 'vue'
+    import Component from "@/{{ $name }}.vue"
+    createApp(Component).mount('#vue-{{ .Page.Scratch.Get "componentId" }}', true)
+</script>
```
mrbbot:
```suggestion
</script>

```

layouts/shortcodes/tab.html
```diff
@@ -1,11 +1,11 @@
...
 {{ if eq $default "true"}}
-<div class="tab tab-default" id="tab-{{$label}}-UNIQUE_ID">{{ $.Inner | markdownify }}</div>
+<div class="tab tab-default" id="tab-{{$label}}-UNIQUE_ID">{{ $.Inner | .Page.RenderString }}</div>
 {{ else if eq $noCode "true" }}
 <div class="tab noCodeTab" id="tab-{{$label}}-UNIQUE_ID">{{ $.Inner | .Page.RenderString }}</div>
 {{ else }}
-<div class="tab" id="tab-{{$label}}-UNIQUE_ID">{{ $.Inner | markdownify }}</div>
-{{ end }}
+<div class="tab" id="tab-{{$label}}-UNIQUE_ID">{{ $.Inner | .Page.RenderString }}</div>
+{{ end }}
```
mrbbot, line 11:
```suggestion
{{ end }}

```

vite.config.ts
```diff
@@ -0,0 +1,45 @@
...
+        const codeBlocks = parsedHtml.querySelectorAll("pre[data-code]");
+
+        for (const block of codeBlocks) {
+          // Base64 is used in order to preserve whitespace
+          block.replaceWith(
+            highlight(
```
mrbbot: Hmmm, https://0d78ae83.cloudflare-docs-7ou.pages.dev/workers/get-started/guide/ doesn't seem to have syntax highlighting? Is there something wrong with the preview build?

mrbbot (commented):
This is so cool! 😃 There seem to be some issues with the mermaid diagrams and code highlighting in the preview build, but I'm not too sure what's gone wrong.

kristianfreeman (commented):
This is great! generally, I'm in favor of moving some of the more JS-intensive stuff into Vue as it will be easy to mix and match it with the static content. 

But it does appear that the syntax highlighting is not currently working in the previous build I've looked at.

This [D1 tutorial in preview build](https://0d78ae83.cloudflare-docs-7ou.pages.dev/d1/tutorials/build-a-comments-api/) is missing highlighting or even appears to be passing it through completely: see the last example with the `highlight` directive that is asking the syntax processor to additionally highlight a line, on top of the standard syntax highlighting.

penalosa:
Uhhh, it seems I broke this last night. I've fixed it now (but I can't push)

layouts/shortcodes/worker-starter.html
penalosa: New:

Screenshot 2023-02-08 at 15 28 11

Old. (i.e. what's currently in production)

Screenshot 2023-02-08 at 15 29 01

components/DynamicLearningPath.vue
penalosa: Is the bug the pluralisation?

mrbbot: Yep

kodster28: I would say this probably isn't worth fixing... we're planning on actually migrating these to a new format after #7612 gets merged in

mrbbot (approved):
🤩🎉✅

penalosa:
@WalshyDev @codewithkristian did you have a chance to look through this?

kodster28:
> This is great! generally, I'm in favor of moving some of the more JS-intensive stuff into Vue as it will be easy to mix and match it with the static content.
> 
> But it does appear that the syntax highlighting is not currently working in the previous build I've looked at.
> 
> This [D1 tutorial in preview build](https://0d78ae83.cloudflare-docs-7ou.pages.dev/d1/tutorials/build-a-comments-api/) is missing highlighting or even appears to be passing it through completely: see the last example with the `highlight` directive that is asking the syntax processor to additionally highlight a line, on top of the standard syntax highlighting.

Confirmed this issue is fixed in the [latest build](https://penalosa-modify-build.cloudflare-docs-7ou.pages.dev/d1/tutorials/build-a-comments-api/#test-with-an-optional-frontend).

@codewithkristian, would you be able to take another look and approve from our side?

JacobMGEvans (approved):
I really like this improvement, the use of the components to improve the end result is great but it is also a nice abstraction with how the templating is done and handled by the component.

WalshyDev:
Multi-line code copying is a bit messed up, works nice if it's a single line but a multiple (like in https://penalosa-modify-build.cloudflare-docs-7ou.pages.dev/pages/framework-guides/deploy-a-brunch-site/#create-a-github-repository) will get messed up:

Actual:
```
git init
$ git add -A
git commit -m "<YOUR_COMMIT_MESSAGE>"
$ git remote add origin <YOUR_NEW_GITHUB_REPOSITORY_URL>
git push -u origin main
```

Expected:
```
git init
git add -A
git commit -m "<YOUR_COMMIT_MESSAGE>"
git remote add origin <YOUR_NEW_GITHUB_REPOSITORY_URL>
git push -u origin main
```

Also, very strange movement while things load (happens with JS enabled or disabled but is way more noticeable if JS is disabled):
Edit: Apparently this happens even on the normal site, fun...

https://user-images.githubusercontent.com/8492901/218917607-352f0a36-5c0b-40b4-9e15-d06811d9d2e3.mp4

---

What was the reasoning to use Vue here? React is used far more internally. I don't mind Vue for these small things but ~~React > Vue~~ worth thinking about.

penalosa:
> Multi-line code copying is a bit messed up

Ah yes, that's not great. That's also the behaviour on the current site, but I can fix it in this PR.

> What was the reasoning to use Vue here? React is used far more internally. I don't mind Vue for these small things but React > Vue worth thinking about.

I used Vue because the codebase already had Vue components (they just used string templates rather than Vue SFCs). To be honest, my preference for something like this (where it's small bits of functionality on a page rather than a full JS app) would be Svelte, but I thought it'd be worth sticking to the frameworks already in use. React could definitely be adding in future though (a `<react-component name="..."/>` similar to Vue would be fairly easy to add)

kodster28:
> > Multi-line code copying is a bit messed up
> 
> Ah yes, that's not great. That's also the behaviour on the current site, but I can fix it in this PR.
> 
> > What was the reasoning to use Vue here? React is used far more internally. I don't mind Vue for these small things but React > Vue worth thinking about.
> 
> I used Vue because the codebase already had Vue components (they just used string templates rather than Vue SFCs). To be honest, my preference for something like this (where it's small bits of functionality on a page rather than a full JS app) would be Svelte, but I thought it'd be worth sticking to the frameworks already in use. React could definitely be adding in future though (a `<react-component name="..."/>` similar to Vue would be fairly easy to add)

As mentioned above, there's actually some work to quasi-deprecate the Vue stuff (#7612) and move either to Hugo base components or switch to another framework potentially.

More than happy to help with that migration (mostly the dashboard landing component) after the deprecation part (dynamic learning paths) is done.

Obviously would need some help on the other parts of the framework... but think I could assist re-writing the dashboard landing component either in vanilla JS / Svelte / React / whatever we'd rather go with.

kodster28 (approved):
From what I can see, looks good.

Spot checked a few places (learning paths pages, code highlighting, mermaid, footnotes, dash/api tabs).

WalshyDev (approved):

pedrosousa:
I see some "block rearrangement" glitch whenever I navigate to a different page (Windows 10, Chrome browser). I can also briefly see the "Skip to content" link appearing at the top:

navigation-glitch

Update: Oh, this had been reported already! I'm seeing it less often now, and only in Chrome (tested also with Firefox and Edge). However, in my case I never saw this behavior in the live site.

kodster28:
Think we're good to go on this (assuming the weird load issue doesn't show up in Production)... but could it wait until next Tuesday?

Just think it makes sense not to make any big changes right before a holiday weekend.

PR #7595, opened 2023-02-09
[ZT] WARP disk encryption check

ranbel:


spatel-cf:
@ranbel we should probably remove the line: "All disks on the system must be encrypted for the posture check to pass" since we have the ability to specify a disk.

patriciasantaana (approved):

PR #7618, opened 2023-02-12
Pages framework guides improvements

dario-piotrowicz:
- rename and update pages framework-guides for better consistency
  meaning:
   - use consistent md file names
   - always refer to the site as "site" (not also "application" or
"project")

- create new _learn-more partial for the pages framework guides
this will help reduce text duplication and also maintain consistency


dario-piotrowicz:
Regarding the _learn-more partial, I think that also the github repo creation section could be made into a partial, but since it is a bit of a time consuming process I just wanted to first check if the _learn-more approach is accepted :slightly_smiling_face: (if that's the case I can add the other partial either here or in a followup PR)

content/pages/how-to/preview-with-cloudflare-tunnel.md
```diff
@@ -22,7 +22,7 @@ cloudflared version 2021.5.9 (built 2021-05-21-1541 UTC)
...
+The easiest way to get up and running with Cloudflare Tunnel is to have an application running locally, such as a [React](/pages/framework-guides/deploy-a-react-site/) or [Svelte](/pages/framework-guides/deploy-a-svelte-site/) site. When you are developing an application with these frameworks, they will often make use of a `npm run develop` script, or something similar, which mounts the application and runs it on a `localhost` port. For example, the popular `create-react-app` tool runs your in-development React application on port 3000, making it accessible at the `http://localhost:3000` address.
```
deadlypants1973:
```suggestion
The easiest way to get up and running with Cloudflare Tunnel is to have an application running locally, such as a [React](/pages/framework-guides/deploy-a-react-site/) or [Svelte](/pages/framework-guides/deploy-a-svelte-site/) site. When you are developing an application with these frameworks, they will often make use of a `npm run develop` script, or something similar, which mounts the application and runs it on a `localhost` port. For example, the popular `create-react-app` tool runs your in-development React application on port `3000`, making it accessible at the `http://localhost:3000` address.
```

deadlypants1973 (approved):

PR #7620, opened 2023-02-13
Only Super Admin can see Payload Logging option

kyouheicf:
{{<Aside type="note">}}
Only the **Super Administrator** can see the option to configure Payload Logging. The decryption of logs is available to everyone as long as they have access to the private key. {{</Aside>}}

### DIFF
--- a/content/waf/managed-rules/payload-logging/configure.md
+++ b/content/waf/managed-rules/payload-logging/configure.md
@@ -8,6 +8,10 @@ meta:
 
 # Configure payload logging for a managed ruleset in the dashboard
 
+{{<Aside type="note">}}
+Only the **Super Administrator** can see the option to configure Payload Logging. The decryption of logs is available to everyone as long as they have access to the private key.
+{{</Aside>}}
+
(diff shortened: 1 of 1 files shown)


content/waf/managed-rules/payload-logging/configure.md
```diff
@@ -8,6 +8,10 @@ meta:
 
 # Configure payload logging for a managed ruleset in the dashboard
 
+{{<Aside type="note">}}
+Only the **Super Administrator** can see the option to configure Payload Logging. The decryption of logs is available to everyone as long as they have access to the private key.
```
helloandre: this is true-ish. We only presenent the option to decrypt in the UI to Super Admins as well. Anyone with access to the logs can decrypt with the payload with the private key, but they also need access to the logs.

pedrosousa: I updated the proposed sentence. Is it correct, @helloandre?

pedrosousa (approved):

PR #7644, opened 2023-02-14
Fix `Linux.md`

Mon-ius:
The original document was too misleading which caused unnecessary trouble. Here I provided a more safe way to prevent the unexpecting conditions. Hope you CF guy can take it into consideration, cheers :)

### DIFF
--- a/content/warp-client/get-started/linux.md
+++ b/content/warp-client/get-started/linux.md
@@ -31,8 +31,15 @@ The command line interface is the primary way to use WARP.
 
 To connect for the very first time you must call `register` first:
 
+#### Proxy Mode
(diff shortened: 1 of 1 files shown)


haleycode:
Thank you for the suggestion! We won't be updating the docs at this time. The instructions on the following page should be helpful: https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/manual-deployment/#linux

Mon-ius:
I don't mean to be unfriendly, but according to your answer, it's not related to what I reported at all. If you don't understand how your guide can misleading people, you can report it to your supervisor, young boy :)

PR #7673, opened 2023-02-15
[Queues] Codeblock cleanup

KianNH:
There was some scope creep - originally I just added `$` to some codeblocks and decided to try bring all of the code examples in-line with each other.

Some specified return types, some said `req` vs `request`, etc - generally ran them all through Prettier in line with how I've done other Workers in the docs.


KianNH:
cc @elithrar @rts-rob

elithrar (approved):

elithrar:
Thanks @KianNH - !

PR #7686, opened 2023-02-15
DRAFT: Improve Billable Duration docs in the Workers pricing page

jahands:
This is a very early WIP
We want to improve docs here for how Billable Duration is counted, and maybe improve the structure of this page a bit


kodster28:
Closing for now. May be taken up again in the future.

PR #7688, opened 2023-02-15
Link to zero trust dashboard was incorrect

thecrazytech1:
It sent you to jttps://dash.teams.cloudflare.com instead of https://dash.teams.cloudflare.com (there was a j instead of a h at the start of the link)


ranbel (approved):
thanks for the fix!

thecrazytech1:
> 

Thanks, I saw that while tryin to setup ssh with cloudflared, and decided to fix it :)

PR #7690, opened 2023-02-16
[ZT] device information only mode

ranbel:
PCX-5840


content/cloudflare-one/connections/connect-devices/warp/configure-warp/warp-modes/device-information-only.md
```diff
@@ -0,0 +1,74 @@
...
+weight: 2
+---
+
+# Enable Device Information Only mode
+
+Device Information Only mode allows you to enforce device posture rules when a user connects to your [self-hosted Access application](/cloudflare-one/applications/configure-apps/self-hosted-apps/). This mode relies on a client certificate generated from your account to establish trust between the Access application and the device.
+
+To setup Device Information Only mode:
+
+1. [Create an API token](/fundamentals/api/get-started/create-token/) with the following minimum permissions:
```
ranbel: I wasn't able to get the API token to work. Only the API key worked, though that method is not recommended. Is the token actually supported for this API?

ranbel: Will add token in later after PM confirms that it works.

kodster28 (approved):

PR #7734, opened 2023-02-21
docs(next-pages): remove section about flags no longer necessary

matteo-hertel:
## Overview

This PR updates the docs about deploying about a next js site by removing the section about the flags  `streams_enable_constructors` and `transformstream_enable_standard_constructor`, the flags not only are no longer necessary but the deployment throws an error about removing them

```
Error: Failed to publish your Function. Got error: The compatibility flag streams_enable_constructors became the default as of 2022-11-30 so does not need to be specified anymore.
The compatibility flag transformstream_enable_standard_constructor became the default as of 2022-11-30 so does not need to be specified anymore.
```

### DIFF
--- a/content/pages/framework-guides/deploy-a-nextjs-site.md
+++ b/content/pages/framework-guides/deploy-a-nextjs-site.md
@@ -139,10 +139,6 @@ The `@cloudflare/next-on-pages` CLI transforms the Edge Runtime components of yo
 
 {{</Aside>}}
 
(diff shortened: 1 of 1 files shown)


KimJ15:
@WalshyDev do you approve this update?

WalshyDev:
The guide has been updated since this PR was made and the flags are no longer referenced. Thanks for this though :)

PR #7740, opened 2023-02-22
[Home layout] Added more www material to home footer

kodster28:


static/home.css
```diff
@@ -5774,3 +5798,39 @@ blockquote .DocsMarkdown--header-anchor-positioner {
...
+}
+
+@media (min-width:900px) {
+    .legalInformation li {
+        padding-left: 0;
+    }
+
+    .legalInformation li + li:before {
+        content: "|";
+        padding-right: 1rem;
```
pedrosousa: The spacing in both sides of the pipe character (`|`) is not the same. I tried using `padding-right: 0.5rem;` here instead and it seems to have fixed it.

pedrosousa (approved):

static/home.css
kodster28:
```suggestion
        padding-right: .5rem;
```

PR #7746, opened 2023-02-22
[Docs site] Product grid update

kodster28:


content/_index.html
```diff
@@ -78,14 +78,7 @@ <h1 class="leading">Looking to improve security, <br /> performance, and reliabi
...
 <div class="ViewProducts">
-<details>
-<summary>View all products</summary>
-<div>
-
-{{<product-grid>}}
-
-</div>
-</details>
+<a href="/products/" style="font-size: 1.2em; text-decoration:underline;">View all products ></a>
```
kodster28, line 81: Changing this from a dropdown to a link since the styling doesn't really work on both pages now.

layouts/partials/head.meta.html
```diff
@@ -42,6 +42,9 @@
 {{- with .Context.Params.pcx_content_type -}}
 <meta name="pcx_content_type" content= "{{ . }}" />
 {{- end -}}
+{{- with $DATA.product.group -}}
```
kodster28, line 45: Adding a value we can scrape and then pull into our reporting. This helps the source of truth live as close to the documentation as possible.

maxvp (approved):

PR #7759, opened 2023-02-23
[Docs site] Add tabs + add data for prioritization based on product group

kodster28:
PCX-5909

### DIFF
--- a/assets/search-coveo.ts
+++ b/assets/search-coveo.ts
@@ -7,7 +7,7 @@
   const result = currentLocation.match(paramRegex);
   let productGroup
   
-  if (result[1]) {
+  if (result !== null) {
     const params = new URLSearchParams(result[1]);
     productGroup = params.get("product_group")
   }
--- a/assets/search-coveo.ts
+++ b/assets/search-coveo.ts
@@ -3,13 +3,13 @@
   let tag = document.currentScript;
   let $ = document.querySelector.bind(document);
   const currentLocation = window.location.href;
-  const paramRegex = new RegExp("\/search\/\#(.+)");
+  const paramRegex = new RegExp("/search/#(.+)");
   const result = currentLocation.match(paramRegex);
-  let productGroup
+  let productGroup = "";
   
-  if (result !== null) {
+  if (result) {
     const params = new URLSearchParams(result[1]);
-    productGroup = params.get("product_group")
+    productGroup = params.get("product_group") || "";
   }
 
   let coveo: any;
@@ -23,12 +23,10 @@
     }
   }
 
-  function setProductGroupValue() {
-    if (productGroup !== null || productGroup !== undefined) {
-      return productGroup;
-    }
+  function getProductGroupValue() {
+    return productGroup;
   }
-
+  
   function loadCustomSearchBox() {
     let element = $('#DocsSearch--input') || $('#SiteSearch--input');
     const CustomSearchbox = (function(_super) {
@@ -96,13 +94,13 @@
       let pipelineContext = coveo.$$(root).find(".CoveoPipelineContext");
       pipelineContext = coveo.get(pipelineContext);
       pipelineContext.setContextValue("referrer", setReferrerValue());
-      pipelineContext.setContextValue("product_group", setProductGroupValue());
+      pipelineContext.setContextValue("product_group", getProductGroupValue());
     })
 
     coveo.$$(root).on('changeAnalyticsCustomData', (e, args) => {
       if (args.type === 'ClickEvent' || args.type === 'CustomEvent'){
         args.metaObject.context_referrer = setReferrerValue();
-        args.metaObject.context_product_group = setProductGroupValue();
+        args.metaObject.context_product_group = getProductGroupValue();
       }});
 
     // Hacky fix to manually control search/loading icons
--- a/assets/search-coveo.ts
+++ b/assets/search-coveo.ts
@@ -2,6 +2,15 @@
 
   let tag = document.currentScript;
   let $ = document.querySelector.bind(document);
+  const currentLocation = window.location.href;
+  const paramRegex = new RegExp("\/search\/\#(.+)");
+  const result = currentLocation.match(paramRegex);
+  let productGroup
+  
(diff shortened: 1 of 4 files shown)


static/search-coveo.css
```diff
@@ -62,6 +62,33 @@
...
+  background: var(--gray-1);
+  border-bottom: 4px solid orange;
+}
+
+[theme="dark"] .CoveoTab:hover  {
+  background: var(--gray-1);
+}
+
+[theme="dark"] .coveo-tab-section {
+  background: var(--color-rgb)
```
pedrosousa:
```suggestion
  background: var(--color-rgb);
```

static/search-coveo.css
```diff
@@ -62,6 +62,33 @@
...
+}
+
+@media (max-width: 480px) {
+  [theme="dark"] .CoveoSearchInterface .coveo-tab-section {
+      background-color:var(--color-rgb)
```
pedrosousa:
```suggestion
      background-color:var(--color-rgb);
```

assets/search.redirect.ts
```diff
@@ -15,6 +16,10 @@
 
       if (text.length > 0) {
         redirect += '#q=' + encodeURIComponent(text);
+        redirect += '&t=Docs'
```
pedrosousa:
```suggestion
        redirect += '#q=' + encodeURIComponent(text) + '&t=Docs';
```

assets/search-coveo.ts
```diff
@@ -14,6 +23,12 @@
     }
   }
 
+  function setProductGroupValue() {
+    if (productGroup !== null || productGroup !== undefined) {
+      return productGroup;
+    }
+  }
+
```
pedrosousa:
```suggestion
  function getProductGroupValue() {
    return productGroup;
  }
  
```
This function actually returns (or gets) the product group value -- consider renaming the function to `getProductGroupValue`.
Also, if the product group is null/undefined, you get to the end of the function without returning anything, which means it will be returning `undefined`. In summary, there's no point in testing the value before returning it.
Note: I've added suggestions to make the default `""` instead of `undefined`.

assets/search-coveo.ts
```diff
@@ -80,11 +96,13 @@
       let pipelineContext = coveo.$$(root).find(".CoveoPipelineContext");
       pipelineContext = coveo.get(pipelineContext);
       pipelineContext.setContextValue("referrer", setReferrerValue());
+      pipelineContext.setContextValue("product_group", setProductGroupValue());
```
pedrosousa:
```suggestion
      pipelineContext.setContextValue("product_group", getProductGroupValue());
```

assets/search-coveo.ts
```diff
@@ -80,11 +96,13 @@
...
     })
 
     coveo.$$(root).on('changeAnalyticsCustomData', (e, args) => {
       if (args.type === 'ClickEvent' || args.type === 'CustomEvent'){
         args.metaObject.context_referrer = setReferrerValue();
+        args.metaObject.context_product_group = setProductGroupValue();
```
pedrosousa:
```suggestion
        args.metaObject.context_product_group = getProductGroupValue();
```

assets/search-coveo.ts
```diff
@@ -2,6 +2,15 @@
 
   let tag = document.currentScript;
   let $ = document.querySelector.bind(document);
+  const currentLocation = window.location.href;
+  const paramRegex = new RegExp("\/search\/\#(.+)");
+  const result = currentLocation.match(paramRegex);
+  let productGroup
```
pedrosousa:
```suggestion
  let productGroup = "";
```

assets/search-coveo.ts
```diff
@@ -2,6 +2,15 @@
...
+  
+  if (result !== null) {
+    const params = new URLSearchParams(result[1]);
+    productGroup = params.get("product_group")
```
pedrosousa:
```suggestion
    productGroup = params.get("product_group") || "";
```

assets/search-coveo.ts
```diff
@@ -2,6 +2,15 @@
 
   let tag = document.currentScript;
   let $ = document.querySelector.bind(document);
+  const currentLocation = window.location.href;
+  const paramRegex = new RegExp("\/search\/\#(.+)");
```
pedrosousa:
```suggestion
  const paramRegex = new RegExp("/search/#(.+)");
```
You don't need to escape these characters.

assets/search-coveo.ts
```diff
@@ -2,6 +2,15 @@
...
+  const result = currentLocation.match(paramRegex);
+  let productGroup
+  
+  if (result !== null) {
```
pedrosousa:
```suggestion
  if (result) {
```

assets/search-coveo.ts
pedrosousa: Question: Even though we're setting the context here, I'm not getting the product group checkbox selected when the search results page loads -- not sure if this was on purpose, or if the checkbox should in fact be selected.

I noticed that the following query string parameter selects the checkbox:
`f:@pcx_product_group=[<PROD_GROUP_NAME>]`
E.g.:
`https://search-group-prioritization.cloudflare-docs-7ou.pages.dev/search/#q=install&t=Docs&f:@pcx_product_group=[Developer%20platform]` when searching for "install" in the Workers tile.

kodster28: Yeah, the context is something that's set on the query pipeline, but it's not quite the same as a filter.

We absolutely have the option to add filters directly (which we'd do by updating the URL we send over in the search-redirect.ts file).

However, we use the context to help with ranking rules w/in our query pipeline. I've historically wanted to use these instead of auto-filtering b/c it still allows someone to find results outside of the tile they're in. Actually, might be worth bringing this up as a team critique topic.

pedrosousa: Yeah, I thought that the main purpose would be influencing results for now. Let's discuss the filtering option, then.

pedrosousa (approved):

PR #7762, opened 2023-02-23
[Docs site] Update priority for style.css

kodster28:
With the upgrades in speed to our baseline site functions, it looks like we actually got too fast for our current style setup (which you can see occasionally in Chrome).

That purple flash + other things are the things hidden by our styles, but it loads behind the page now. Switching to `preload` from `preconnect` seems to help.

https://user-images.githubusercontent.com/26727299/221054388-b96a4471-e28f-4a42-ada2-1562ac2f7098.mov

### DIFF
--- a/content/_headers
+++ b/content/_headers
@@ -1,2 +1,5 @@
 https://:project.pages.dev/*
  X-Robots-Tag: noindex
+
+/*
+  Link: </styles.css>; rel=preload; as=style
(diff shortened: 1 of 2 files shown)


KianNH:
Is it worth throwing in a `link` header in https://github.com/cloudflare/cloudflare-docs/blob/production/content/_headers to leverage Early Hints too?

pedrosousa (approved):

PR #7764, opened 2023-02-24
adding an advanced method for utilising cache with fetch

erfianugrah:
based off issue 7639


erfianugrah:
Will re-check build again

PR #7776, opened 2023-02-24
[Docs site] Added temp css fix for FOUC

kodster28:
Believe this was introduced in #7580...

We're having Flashes of unstyled content or [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) on our pages.

This temporary fix is based on the CSS implementation suggested by [stack overflow](https://stackoverflow.com/questions/3221561/eliminate-flash-of-unstyled-content).

Based on some [preliminary digging](https://dev.to/lyqht/what-the-fouc-is-happening-flash-of-unstyled-content-413j), it seems like it might require some edits to the _vite.config.ts_, specifically the `rewrite` function.

cc: @penalosa


kodster28:
Ugh... this appears to a) not fix the issue and b) have everything load slower.

kodster28:
`dea118c` doesn't work either.

PR #7785, opened 2023-02-27
Add a note about workersInvocationsAdaptive

### LINKED ISSUE #7756: Aggregates in GraphQL Analytics API

Analytics

- https://developers.cloudflare.com/analytics/graphql-api/tutorials/querying-workers-metrics/
- revision: https://github.com/cloudflare/cloudflare-docs/blob/178466a099956e2f5c6ae134954fd1304243cf69/content/analytics/graphql-api/tutorials/querying-workers-metrics/index.md

"API Call"

### What needs to change?

I'm not 100% sure it's wrong, but it seems wrong that this query 1) is selecting aggregation fields (`sum`, `quantiles`), but also 2) is not a `...Groups` field.

I was under the impression that in order to access aggregations, the field would have had to be named something like `workersInvocationsAdaptiveGroups`? That's based on https://developers.cloudflare.com/analytics/graphql-api/features/data-sets/ stating
> Nodes that represent aggregated data include the Groups suffix. For example, the loadBalancingRequestsAdaptiveGroups node represents aggregated data for Load Balancing requests.

and suggesting that you _can't_ do aggregates on non-Groups data sets.

### How should it change?

Just ensure the query is right. If it is, maybe explain this example since AFAICT it's the only on across all the docs that does this.

### Additional information

_No response_

victor-perov:
We have a node that does not follow the naming convention: `workersInvocationsAdaptive` (should be named with `Groups` suffix). It's a beta node, and we are looking to replace it with `workersInvocationsAdaptiveGroups` following the deprecation policy.

This commit supposed to address confusion caused by name mismatch (see #7756).

### DIFF
--- a/content/analytics/graphql-api/features/data-sets/index.md
+++ b/content/analytics/graphql-api/features/data-sets/index.md
@@ -22,7 +22,9 @@ identifies the type of data they return:
 * **Aggregated data** - Nodes that represent aggregated data include the
   `Groups` suffix. For example, the `loadBalancingRequestsAdaptiveGroups` node
   represents aggregated data for Load Balancing requests. Aggregated data is
-  returned in an array of `...Group` objects.
+  returned in an array of `...Group` objects. Please note: we have a node that
+  currently excluded from that naming convention - `workersInvocationsAdaptive`
+  (beta).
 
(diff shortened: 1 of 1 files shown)


victor-perov:
cc @angelampcosta

angelampcosta (approved):

PR #7800, opened 2023-02-28
Update default-cache-behavior.md

lindsey-em:
It's unclear what cache status will be seen when cookies are present, unless each individual status type is read. Adding this note can both reduce escalations and improve customer support accuracy.


angelampcosta:
Thank you @lindsey-em for bringing this up. We updated the docs with #7813

PR #7802, opened 2023-02-28
[PCX team] Remove preview image from .yml

kodster28:


data/api.yml
```diff
@@ -11,7 +11,7 @@ meta:
   description: Cloudflare’s API tokens allow you to make calls to our API to alter
     different settings.
   author: '@cloudflare'
-  image: /cf-twitter-card.png
+
```
pedrosousa, line 14: Nit: The images for "api" and "apps" will not work (they never did?), but let's keep handling these as in other tiles.

pedrosousa (approved):

PR #7872, opened 2023-03-03
[Durable Objects] websocket hibernation

koeninger:


content/workers/learning/using-durable-objects.md
```diff
@@ -143,7 +143,7 @@ As part of Durable Objects, Workers can act as WebSocket endpoints – including
...
 
-For more information, refer to [Using WebSockets](/workers/learning/using-websockets/). For an example of WebSockets in action within Durable Objects, review the [example chat application](https://github.com/cloudflare/workers-chat-demo).
+Durable Objects can use the web standard APIs described in [Using WebSockets](/workers/learning/using-websockets/), but the [WebSockets Hibernation API](/workers/runtime-apis/durable-objects/#websockets-hibernation-api) has significant advantages and should be preferred. For an example of WebSockets in action within Durable Objects, review the [example chat application](https://github.com/cloudflare/workers-chat-demo).
```
koeninger: update the chat demo to use hibernation, add a simple example here, or both?

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,42 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+The system calls the `webSocketMessage` method when an accepted WebSocket receives a message. The method is not called for WebSocket control frames; the system will respond to an incoming Ping automatically without interrupting hibernation. The method takes `(ws: WebSocket, message: String | ArrayBuffer)` as parameters. It does not return a result and can be `async`.
+
+#### `webSocketClose` handler method
+
+The system calls the `webSocketClose` method when a WebSocket is closed. The method takes `(ws: WebSocket, code: number, reason: string)` as parameters. It does not return a result and can be `async`.
+
+#### `webSocketError` handler method
+
+The system calls the `webSocketError` method for any non-disconnection related errors. The method takes an `Error` as the parameter. It does not return a result and can be `async`.
```
koeninger: does this take a js Error?

content/workers/learning/using-durable-objects.md
cacheburner: I think it'd be cool to have the chat demo use hibernation, but I wouldn't want it to replace the existing demo outright. Maybe have hibernation be another branch in the repo, or another repo entirely? Simple example in the docs would be a nice addition too, I feel like we don't have a lot of DO examples.

zegevlier:
It'd be nice to have a notice on the Runtime APIs -> WebSockets page and potentially on the Learning -> Using WebSockets page stating that this hibernation API exists and might be preferable. I hadn't looked at the changed files yet and had difficulty finding the change documentation on the preview site. Adding a notice there would have mitigated this issue for me.

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+- {{<code>}}state.getWebSockets(tag{{<param-type>}}string{{</param-type>}}){{</code>}} {{<type>}}Array\<WebSocket>{{</type>}}
+
+  - Gets an array of accepted WebSockets matching the given tag. Disconnected WebSockets are automatically removed from the list.
+
+- {{<code>}}state.setEventTimeout(timeoutMs{{<param-type>}}number{{</param-type>}}){{</code>}}
+
+  - Set the maximum amount of time that a single event of any type is allowed to run before the event will be canceled. This is useful for ensuring that a Durable Object cannot be held open by a hanging event leading to unexpected duration billing. This is provided as a safety measure; carefully-written code does not need it. When a WebSocket is accepted using `state.acceptWebSocket()`, the timeout applies to a single message delivery, not to the whole connection.
```
koeninger: clarify whether this applies to websocket event handlers, or any event handlers

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,42 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+Durable Objects WebSockets support includes extensions to the standard WebSocket interface, related methods on the `state` object, and handler methods that a Durable Object can implement for processing WebSocket events. These APIs allow WebSockets that are not actively handling messages to remain connected when the Durable Object is evicted from memory ("hibernation").
+
+#### WebSocket extensions
+
+- {{<code>}}WebSocket.attachment{{</code>}} {{<type>}}Any{{</type>}}
+
+  -  All WebSockets have this property. It starts out null but can be assigned to any serializable value. The property will survive hibernation. Size is limited to 2048 bytes.
+
+#### `state` methods for WebSockets
```
cacheburner: Brief update here regarding `acceptWebSocket()` and `getWebSockets()`.

### `acceptWebSocket(ws, tags)`
The second argument, `tags`, is now optional. This means `acceptWebSocket(ws)` is valid (and in fact equivalent to `acceptWebSocket(ws, [])`).

Calling `acceptWebSocket()` _without_ tags has implications for `getWebSockets()`, so that's changed a bit too.

### `getWebSockets(tag)`
The argument `tag` is now optional. Calling `getWebSockets()` will return **all websockets**, regardless of if they have a tag or not. Otherwise, behavior is unchanged.

### Implications
I want to point out that the only way to get back tagless websockets (aside from if an event is triggered, such as the `webSocketMessage` event) is to get _all_ websockets. It doesn't make sense to provide support for getting back only tagless websockets because that's what tags are for.

If your websockets are indistinguishable, and you would otherwise call `acceptWebSocket(ws, [])` or would always use the same tag for each websocket, then it makes sense to call  `acceptWebSocket(ws)` without tags since all your websockets implicitly share the empty tag.
- Not passing a tag in this case also saves us (Cloudflare) a little memory overhead, but this is outside of the isolate so no additional benefits to the end user aside from convenience.

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+#### `webSocketMessage` handler method
+
+The system calls the `webSocketMessage` method when an accepted WebSocket receives a message. The method is not called for WebSocket control frames; the system will respond to an incoming [WebSocket protocol ping](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2) automatically without interrupting hibernation. The method takes `(ws: WebSocket, message: String | ArrayBuffer)` as parameters. It does not return a result and can be `async`.
+
+#### `webSocketClose` handler method
+
+The system calls the `webSocketClose` method when a WebSocket is closed. The method takes `(ws: WebSocket, code: number, reason: string)` as parameters. It does not return a result and can be `async`.
```
cacheburner: As per https://github.com/cloudflare/workerd/pull/343#discussion_r1154712849, `webSocketClose()` will also take `wasClean: bool` as the last parameter.

content/workers/runtime-apis/durable-objects.md
```diff
@@ -306,17 +306,21 @@ Hibernation does not persist WebSocket connections across [code updates](/worker
 
 #### WebSocket extensions
 
-- {{<code>}}WebSocket.attachment{{</code>}} {{<type>}}Any{{</type>}}
+- {{<code>}}WebSocket.serializeAttachment(value{{<param-type>}}any{{</param-type>}}){{</code>}} : {{<type>}}void{{</type>}}
 
-  -  All WebSockets have this property. It starts out null but can be assigned to any serializable value. The property will survive hibernation. Size is limited to 2048 bytes.
+  - All WebSockets have this method. It keeps a copy of any serializable `value` in memory such that it will survive hibernation. If you modify `value` after calling this method, those changes will not be retained unless you call this method again. The serialized size of `value` is limited to 2048 bytes, otherwise this method will throw an error. If you need larger values to survive hibernation, use the [storage api](/workers/runtime-apis/durable-objects/#transactional-storage-api) and pass the corresponding key to this method so it can be retrieved later.
```
jasnell: It should be pointed out that these are non-standard, cloudflare-specific extensions.

jasnell: Also, it might be good to explain what a "serializable value" is.

koeninger: The heading just above this says "Durable Objects WebSockets support includes extensions to the standard WebSocket interface," is there more you'd want to see?  Change that to "Cloudflare-specific extensions" ?

koeninger: changed to re-use the existing explanation

content/workers/runtime-apis/websockets.md
```diff
@@ -7,7 +7,7 @@ title: WebSockets
 
 ## Background
 
-WebSockets allow you to communicate in real time with your Cloudflare Workers serverless functions.
+WebSockets allow you to communicate in real time with your Cloudflare Workers serverless functions. For Durable Objects, prefer using the [WebSockets Hibernation API](/workers/runtime-apis/durable-objects/#websockets-hibernation-api).
```
irvinebroque: Could we also add a note here or on this page that clarifies that the `WebSocket` object, which is otherwise documented here, has additional methods on it when used with Durable Objects? Just to cover the case where someone ends up looking here for these, when they really should be looking at `/runtime-apis/durable-objects/`

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
 
+### WebSockets Hibernation API (beta)
+
+Durable Objects WebSockets support includes Cloudflare-specific extensions to the standard WebSocket interface, related methods on the `state` object, and handler methods that a Durable Object can implement for processing WebSocket events. These APIs allow a Durable Object that is not currently running an event handler to be removed from memory while keeping its WebSockets connected ("hibernation").
+
+Hibernation does not persist WebSocket connections across [code updates](/workers/learning/using-durable-objects/#global-uniqueness). If an event occurs for a hibernated Durable Object's corresponding handler method, it will return to memory. This will call the Durable Object's constructor, so it is best to minimize work in the constructor when using WebSocket hibernation.
+
+#### WebSocket extensions
+
+- {{<code>}}WebSocket.serializeAttachment(value{{<param-type>}}any{{</param-type>}}){{</code>}} : {{<type>}}void{{</type>}}
```
jasnell:
```suggestion
- {{<code>}}webSocket.serializeAttachment(value{{<param-type>}}any{{</param-type>}}){{</code>}} : {{<type>}}void{{</type>}}
```

Using `WebSocket.serializeAttachment` implies that these are static functions on the `WebSocket` class. Rather, they are instance methods. Conventionally we would represent that using the lower-case `webSocket` or, alternatively, `WebSocket.prototype.serializeAttachment`

content/workers/runtime-apis/websockets.md
```diff
@@ -7,7 +7,7 @@ title: WebSockets
 
 ## Background
 
-WebSockets allow you to communicate in real time with your Cloudflare Workers serverless functions.
+WebSockets allow you to communicate in real time with your Cloudflare Workers serverless functions. For Durable Objects, prefer using the [WebSockets Hibernation API](/workers/runtime-apis/durable-objects/#websockets-hibernation-api) which adds Cloudflare-specific extensions to the standard `WebSocket` object.
```
irvinebroque:
```suggestion
WebSockets allow you to communicate in real time with your Cloudflare Workers serverless functions.

When using WebSockets with Durable Objects, we recommend using the [WebSockets Hibernation API](/workers/runtime-apis/durable-objects/#websockets-hibernation-api) which adds additional extensions to the standard `WebSocket` object. 
```

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+  - Keeps a copy of `value` in memory such that it will survive hibernation. The value can be any type supported by the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm), which is true of most types. If you modify `value` after calling this method, those changes will not be retained unless you call this method again. The serialized size of `value` is limited to 2048 bytes, otherwise this method will throw an error. If you need larger values to survive hibernation, use the [storage api](/workers/runtime-apis/durable-objects/#transactional-storage-api) and pass the corresponding key to this method so it can be retrieved later.
```
cacheburner, line 315: Do we want to add other limits to this page, or to the limits page?

koeninger: I'd say this page, send me a list of the current configured limits & I'll add them

content/workers/platform/pricing.md
```diff
@@ -118,9 +118,9 @@ Durable Objects are only available on the Workers Paid plan.
 
 {{</table-wrap>}}
 
-1.  Duration is billed in wall-clock time as long as the Object is active, but is shared across all requests active on an Object at once. Once your Object finishes responding to all requests, it will stop incurring duration charges. A WebSocket being connected to the Durable Object counts as the Object being active.
+1.  Duration is billed in wall-clock time as long as the Object is active, but is shared across all requests active on an Object at once. Once your Object finishes responding to all requests, it will stop incurring duration charges. Calling `.accept` on a WebSocket in an Object will incur duration charges for the entire time the WebSocket is connected. [Prefer using `state.acceptWebsocket`](/workers/runtime-apis/durable-objects/#websockets-hibernation-api), which will stop incurring duration charges once all event handlers finish running.
```
cacheburner: nit: `.accept()`
nit: `state.acceptWebSocket()` (should at least capitalize the `Socket` part)

content/workers/learning/using-durable-objects.md
cacheburner: nit: `WebSocket Hibernation API` reads better than `WebSockets Hibernation API` (though I see "WebSockets Hibernation" in a few other places so feel free to ignore).

Might also be worth mentioning that this is only for websockets that are terminated in the DO? i.e. DO as a client doesn't do hibernation

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+- {{<code>}}state.getWebSockets(tag{{<param-type>}}string{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}){{</code>}} : {{<type>}}Array\<WebSocket>{{</type>}}
+
+  - Gets an array of accepted WebSockets matching the given tag. Disconnected WebSockets are automatically removed from the list. Calling `getWebSockets()` with no `tag` argument will return all WebSockets.
+
+#### `webSocketMessage` handler method
+
+The system calls the `webSocketMessage` method when an accepted WebSocket receives a message. The method is not called for WebSocket control frames; the system will respond to an incoming [WebSocket protocol ping](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2) automatically without interrupting hibernation. The method takes `(ws: WebSocket, message: String | ArrayBuffer)` as parameters. It does not return a result and can be `async`.
```
cacheburner: nit: Instead of saying the parameters the method takes, we could show the full method signature as is done for other methods (eg. `getWebSockets()`) above. (Also applies to Close and Error handlers)

koeninger: That's the way other handlers are currently documented (e.g. https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#alarm-handler-method)  Seems like a reasonable concern, but maybe revisit outside of this PR to make it consistent?

content/workers/learning/using-durable-objects.md
koeninger: I'm going to punt on the pluralization for now since it's plural everywhere else.  Tried to clarify termination vs client

koeninger: went ahead and linked to the hibernation branch of the chat demo

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+#### `state` methods for WebSockets
+
+- {{<code>}}state.acceptWebSocket(ws{{<param-type>}}WebSocket{{</param-type>}}, tags{{<param-type>}}Array\<string>{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}){{</code>}} : {{<type>}}void{{</type>}}
+
+  - Adds a WebSocket to the set attached to this object. `ws.accept()` must NOT have been called separately. Once called, any incoming messages will be delivered by calling the Durable Object's webSocketMessage() handler, and webSocketClose() will be invoked upon disconnect. After calling this, the WebSocket is accepted, so its send() and close() methods can be used to send messages, but its addEventListener() method won't ever receive any events as they'll be delivered to the DurableObject instead. `tags` are optional string tags which can be used to look up the WebSocket with getWebSockets(). Each tag is limited to 256 characters, and each WebSocket is limited to 10 tags associated with it. This API has a maximum of 32,768 WebSockets connected per Durable Object instance, but the cpu and memory usage of a given workload may further limit the practical number of simultaneous connections.
```
Maddy-Cloudflare:
```suggestion
  - Adds a WebSocket to the set attached to this object. `ws.accept()` must NOT have been called separately. Once called, any incoming messages will be delivered by calling the Durable Object's `webSocketMessage()` handler, and `webSocketClose()` will be invoked upon disconnect. After calling this, the WebSocket is accepted, so its `send()` and `close()` methods can be used to send messages, but its `addEventListener()` method won't ever receive any events as they'll be delivered to the Durable Object instead. `tags` are optional string tags which can be used to look up the WebSocket with `getWebSockets()`. Each tag is limited to 256 characters, and each WebSocket is limited to 10 tags associated with it. This API has a maximum of 32,768 WebSockets connected per Durable Object instance, but the cpu and memory usage of a given workload may further limit the practical number of simultaneous connections. 
```

Maddy-Cloudflare (approved):

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+- {{<code>}}state.getWebSockets(tag{{<param-type>}}string{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}){{</code>}} : {{<type>}}Array\<WebSocket>{{</type>}}
+
+  - Gets an array of accepted WebSockets matching the given tag. Disconnected WebSockets are automatically removed from the list. Calling `getWebSockets()` with no `tag` argument will return all WebSockets.
+
+#### `webSocketMessage` handler method
```
Maddy-Cloudflare:
```suggestion
#### `webSocketMessage()` handler method
```

content/workers/runtime-apis/durable-objects.md
Maddy-Cloudflare:
```suggestion
The system calls the `webSocketMessage()` method when an accepted WebSocket receives a message. The method is not called for WebSocket control frames; the system will respond to an incoming [WebSocket protocol ping](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2) automatically without interrupting hibernation. The method takes `(ws: WebSocket, message: String | ArrayBuffer)` as parameters. It does not return a result and can be `async`.
```

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+The system calls the `webSocketMessage` method when an accepted WebSocket receives a message. The method is not called for WebSocket control frames; the system will respond to an incoming [WebSocket protocol ping](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2) automatically without interrupting hibernation. The method takes `(ws: WebSocket, message: String | ArrayBuffer)` as parameters. It does not return a result and can be `async`.
+
+#### `webSocketClose` handler method
```
Maddy-Cloudflare:
```suggestion
#### `webSocketClose()` handler method
```

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+The system calls the `webSocketClose` method when a WebSocket is closed. The method takes `(ws: WebSocket, code: number, reason: string, wasClean: boolean)` as parameters. `wasClean` is true if the connection closed cleanly, false otherwise. The method does not return a result and can be `async`.
```
Maddy-Cloudflare:
```suggestion
The system calls the `webSocketClose()` method when a WebSocket is closed. The method takes `(ws: WebSocket, code: number, reason: string, wasClean: boolean)` as parameters. `wasClean` is true if the connection closed cleanly, false otherwise. The method does not return a result and can be `async`.
```

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+#### `webSocketError` handler method
```
Maddy-Cloudflare:
```suggestion
#### `webSocketError()` handler method
```

content/workers/runtime-apis/durable-objects.md
```diff
@@ -298,6 +298,44 @@ The method takes a [`Request`](/workers/runtime-apis/request/) as the parameter
...
+
+The system calls the `webSocketError` method for any non-disconnection related errors. The method takes `(ws: WebSocket, error: any)` as parameters. It does not return a result and can be `async`.
```
Maddy-Cloudflare:
```suggestion
The system calls the `webSocketError()` method for any non-disconnection related errors. The method takes `(ws: WebSocket, error: any)` as parameters. It does not return a result and can be `async`.
```

Maddy-Cloudflare:
```suggestion
The system calls the `webSocketError()` method for any non-disconnection related errors. The method takes `(ws: WebSocket, error: any)` as parameters. It does not return a result and can be `async`.
```

content/workers/runtime-apis/durable-objects.md
Maddy-Cloudflare, line 329:
```suggestion
The system calls the `webSocketMessage()` method when an accepted WebSocket receives a message. The method is not called for WebSocket control frames; the system will respond to an incoming [WebSocket protocol ping](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2) automatically without interrupting hibernation. The method takes `(ws: WebSocket, message: String | ArrayBuffer)` as parameters. It does not return a result and can be `async`.
```

Maddy-Cloudflare (approved):

content/workers/runtime-apis/durable-objects.md
koeninger: wasClean isn't a function

cacheburner (approved):
Thanks for writing this!

content/workers/learning/using-websockets.md
```diff
@@ -179,4 +179,4 @@ Cloudflare Workers does not currently support WebSocket compression. This means
 
 ## Durable Objects and WebSocket state
 
-If your application needs to coordinate among multiple WebSocket connections, such as a chat room or game match, you will need to create a Durable Object so clients send messages to a single-point-of-coordination. Durable Objects are a coordinated state tool for Cloudflare Workers, which are often used in parallel with WebSockets to persist state over multiple clients and connections. Refer to [Durable Objects](/workers/learning/using-durable-objects/) to get started.
+If your application needs to coordinate among multiple WebSocket connections, such as a chat room or game match, you will need to create a Durable Object so clients send messages to a single-point-of-coordination. Durable Objects are a coordinated state tool for Cloudflare Workers, which are often used in parallel with WebSockets to persist state over multiple clients and connections. Refer to [Durable Objects](/workers/learning/using-durable-objects/) to get started, and prefer using the Durable Objects [WebSockets Hibernation API](/workers/runtime-apis/durable-objects/#websockets-hibernation-api) rather than the `.accept` method described above.
```
byule: does the anchor link #websockets-hibernation-api need -beta appended to the end? As it stands this link does not work

koeninger: good call it probably broke when beta was added to the heading

PR #7877, opened 2023-03-04
Update index.md

karthik7678:
As we noted at https://developers.cloudflare.com/fundamentals/data-products/types-of-analytics/#analytics-related-to-specific-properties - Edge Reachability is only available for Enterprise Customers


KimJ15:
Thank you for the feedback! We are consolidating and revising how plan information is shown in developer documentation. I have noted this use case and will provide feedback to the project owners so we can consistently provide plan information.

PR #7882, opened 2023-03-05
add missing shortcode closing tags

dario-piotrowicz:
some shortcodes wrongly do not have closing tags, this can produce build errors as the following:
```
   failed to extract shortcode: unclosed shortcode "resource-group"
```

fix the above by providing the missing shortcode closing tags

### DIFF
--- a/content/r2/data-access/workers-api/workers-api-reference.md
+++ b/content/r2/data-access/workers-api/workers-api-reference.md
@@ -150,6 +150,7 @@ export default {
 
   -  Retrieves the `httpMetadata` from the `R2Object` and applies their corresponding HTTP headers to the `Headers` input object. Refer to [HTTP Metadata](#http-metadata).
 
+{{</definitions>}}
 ## `R2ObjectBody` definition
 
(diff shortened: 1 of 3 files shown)


dario-piotrowicz:
Note: I am not really familiar with hugo but the above build issues prevent me from building the docs locally on my Linux machine which uses v0.111.1-39a4a3cf676533859217805c36181c7a3dfa321c+extended, I have tried on my Mac as well with v0.110.0+extended and there it does build, so I am not sure if the latter version gets stricter or if there is some configs/something else going on.

Regardless even if a build succeeds the missing close tags are an error AFAIK so I figured I would just fix them anyways.

content/r2/data-access/workers-api/workers-api-reference.md
```diff
@@ -150,6 +150,7 @@ export default {
 
   -  Retrieves the `httpMetadata` from the `R2Object` and applies their corresponding HTTP headers to the `Headers` input object. Refer to [HTTP Metadata](#http-metadata).
 
+{{</definitions>}}
```
WalshyDev, line 153:
```suggestion
{{</definitions>}}

```

WalshyDev (approved):

deadlypants1973 (approved):

PR #7917, opened 2023-03-07
[R2] Reorganizing the nav

dcpena:
Updates to the current R2 nav to align with PCX content strategy and improve user experience. Addresses PCX-5956.


content/r2/buckets/presigned-urls.md
```diff
@@ -0,0 +1,214 @@
...
+
+A binding is a how your Worker interacts with external resources such as [KV Namespaces](/workers/runtime-apis/kv/), [Durable Objects](/workers/runtime-apis/durable-objects/), or [R2 Buckets](#api). A binding is a runtime variable that the Workers runtime provides to your code. You can declare a variable name in your `wrangler.toml` file that will be bound to these resources at runtime, and interact with them through this variable. Every binding's variable name and behavior is determined by you when deploying the Worker. Refer to [Environment Variables](/workers/platform/environment-variables) for more information.
+
+A binding is defined in the `wrangler.toml` file of your Worker project's directory.
+
+{{</Aside>}}
+
+A possible use case may be restricting an application to only be able to upload to a specific URL. With presigned URLs, your central signing application might look like the following JavaScript code running on Cloudflare Workers, workerd, or another platform:
+
+```ts
```
KianNH: This looks like the old presigned URLs page rather than the current one (https://github.com/cloudflare/cloudflare-docs/blob/production/content/r2/data-access/s3-api/presigned-urls.md?plain=1#L70) which seems to be renamed in this PR to be under `content/r2/data-migration/s3-api/presigned-urls.md`

elithrar:
@dcpena — we should also be adding all of these changes to `content/_redirects` so that existing links don't break.

elithrar (changes requested):
- [x] Update `content/_redirects` so that existing links don't break.

dcpena:
> * [x]  Update `content/_redirects` so that existing links don't break.

Thanks for that handy little checkbox! Redirects are definitely part of my plan, but I was going to wait until I got the nav finalized so that I was certain on the new paths for the redirects.

content/r2/reference/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Reference
+pcx_content_type: navigation
+weight: 9
+---
+
+# Reference
```
kodster28:
```suggestion
# Reference

{{<directory-listing>}}
```

kodster28 (changes requested):
Content looks good, let me push up a change to help with the link checks.

For context, our link auditing -- for regular files + redirects file -- also completely ignores redirects. This helps with long-term maintenance + helps avoid broken links in the future (and helps us do Find/Replace fixes in the future).

kodster28 (approved):

content/r2/buckets/create-buckets.md
```diff
@@ -0,0 +1,35 @@
+---
+pcx_content_type: how-to
+title: Create new buckets
```
elithrar, line 3: This should probably be ordered first in the list

image

elithrar, line 3:
```suggestion
title: Create new buckets
weight: 1
```

content/r2/buckets/public-buckets.md
```diff
@@ -1,7 +1,6 @@
 ---
 pcx_content_type: how-to
 title: Public buckets
```
elithrar, line 3:
```suggestion
title: Public buckets
weight: 2
```

content/r2/objects/cache-api.md
```diff
@@ -1,15 +1,12 @@
...
-title: Using R2 with the Cache API
-summary: Use the Cache API to store objects from R2 in Cloudflare's cache.
-type: example
-tags:
-  - Cache API
-  - R2
-pcx_content_type: configuration
-weight: 1001
-layout: example
+title: Store objects in Cloudflare's cache
```
elithrar: Should this just be "Use the Cache API" - ?

The guide here is actually NOT as useful as using a custom domain, since the Cache API is more work for a user. I don't think this deserves a top-level guide as it's currently named. cc @jonesphillip

content/r2/reference/pricing.md
elithrar, line 1: Is this buried too deep under Reference? Is Pricing "reference" material?

content/r2/buckets/presigned-urls.md
```diff
@@ -63,77 +63,110 @@ A binding is defined in the `wrangler.toml` file of your Worker project's direct
...
-        aws: { signQuery: true },
-      }
-    );
-
-    // Caller can now use this URL to upload to that object.
-    return new Response(signed.url, { status: 200 });
-  },
-
-  // ... handle other kinds of requests
+	async fetch(request) {
```
KianNH: Bump on this - the proposed changes actually revert this page back to an incorrect example which doesn’t work.

content/r2/reference/pricing.md
dcpena: Pulled out o fReference and made it a top level nav item. Fixed redirect as well

content/r2/buckets/presigned-urls.md
dcpena: Thank you so much for this, @KianNH! Will you give this another pass? Should be fixed now

content/r2/objects/cache-api.md
dcpena: Good catch. Chatted about this with Phillip too and made the updates called out in this comment

elithrar (approved):
Looks great — thanks for making the changes! Let's ship it 🚢

Dezzj (commented):
Approved

PR #7958, opened 2023-03-10
[DNS] align dns overview pages to template

RebeccaTamachiro:
PCX-5668


layouts/shortcodes/plan.html
```diff
@@ -1,6 +1,6 @@
 {{ $map := dict "enterprise" "Enterprise-only" "pro" "Pro and above" "business"
 "Business and above" "all" "Available on all plans" "paid" "Available on paid plans"
-"add-on" "Add-on feature" }}
+"add-on" "Add-on feature" "ent-add-on" "Enterprise-only paid add-on"}}
```
RebeccaTamachiro, line 3: DNS Firewall is a bit different and after talking to Kody it seemed to make sense to add a new plan type to cover it

kodster28 (approved):

PR #7975, opened 2023-03-11
fix: remove unsupported `cloudflared tunnel` bare option

rarkins:
Running `cloudflared tunnel` alone no longer works - a url is required.

```
❯ cloudflared tunnel
You did not specify any valid additional argument to the cloudflared tunnel command.

If you are trying to run a Quick Tunnel then you need to explicitly pass the --url flag.
Eg. cloudflared tunnel --url localhost:8080/.

Please note that Quick Tunnels are meant to be ephemeral and should only be used for testing purposes.
For production usage, we recommend creating Named Tunnels. (https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide/)
```


Gasamdon23 (commented):
H

ranbel (approved):
Thanks for updating the docs!

PR #7976, opened 2023-03-12
fix: search product docs

Cherry:
This tweaks the search to only search product documentation and produce much more useful results.

For example, searching "Workers":
before: 
API references intermingled with results, which is not helpful especially for new users

after: 
only actual Workers documentation.


kodster28:
Addressing this in #7990

PR #7990, opened 2023-03-13
[Docs site] Separated api docs into separate tab

kodster28:
Replaces #7976

### DIFF
--- a/content/search.html
+++ b/content/search.html
@@ -38,7 +38,8 @@ layout: home
       <div class="CoveoBreadcrumb"></div>
       <div class="CoveoDidYouMean"></div>
       <div class="coveo-tab-section">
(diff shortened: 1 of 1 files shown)


KimJ15 (approved):

WalshyDev (approved):
Nice solution, thanks Kody!

kodster28:
A. Thanks @WalshyDev!
B. Doubt it'll take too long, but we'll need to hold this a bit until @codewithkristian can help us get the link updated on the API docs frontend.

Cherry:
This is an awesome solution - way better than my attempt! Thanks Kody 🧡

kodster28:
Thanks for the suggestion, @Cherry!

PR #8000, opened 2023-03-13
[Support] Fix codeblocks for Configuring Token Authentication

Erisa:
- Put things in codeblocks that weren't
- Convert the html madness to real markdown codeblocks
- Add language highlights

PR #8001, opened 2023-03-13
[Support] Fix up a load of codeblocks

Erisa:
After doing #8000 I realised there was a load more I could do easily, so apologies for this PR so soon after the other one that only did a single page.

This will not be all occurrences of weird or broken codeblocks, but it was all the ones I saw from a cursory glance over the pages.


jedecf (approved):
LGTM

PR #8020, opened 2023-03-14
[Workers] Update deploy button action YAML example to 2.0

kamranayub:
## Changes

- Update `cloudflare/wrangler-action` to 2.0
- Fix passing `accountId`

Tested on my own worker template.

image

## Questions

1. Does it make sense to have `pull_request` action? For example, if someone opens a PR you don't necessarily want to deploy a new worker 🤔 It's not like Pages with deploy previews, right?


kamranayub:
Thanks!

PR #8028, opened 2023-03-15
[WAF, Tenant] Zones of Tenant accounts cannot migrate to the new WAF managed rules

pedrosousa:
Addresses PCX-6204.


kodster28 (approved):
Need to confirm whether this still applies since folks can have multiple super admins now, but the added content is good.

pedrosousa:
Actually, during review we got to the conclusion that adding this clarification may be confusing in itself (not every "end customer" knows if they have an associated tenant). I'll close this PR without merging.

PR #8033, opened 2023-03-15
[R2] Added object lifecycles docs

dcpena:
Added docs for object lifecycles. Addresses PCX-5458.

### DIFF
--- a/content/r2/objects/object-lifecycles.md
+++ b/content/r2/objects/object-lifecycles.md
@@ -1,0 +1,152 @@
+---
+title: Object lifecycles
+pcx_content_type: how to
+---
+
+{{<beta heading="h1">}} Object lifecycles {{</beta>}}
+
+Object lifecycles affect how long objects uploaded to your bucket are kept.
+
+For example, you can create an object lifecycle rule to delete objects after 30 days, or you can set a rule to delete parts of multipart uploads.
+
+When you apply a rule to a bucket, the rule automatically applies to any new objects uploaded after the rule creation. For existing objects, the new rule may take a while to apply and will usually occur within 24 hours but can take longer based on the number of objects in the bucket.
+
+## Behavior
+
+- Objects will typically be removed from a bucket within 24 hours of the `x-amz-expiration` value.
+- When a new lifecycle policy is applied, the new object's `x-amz-expiration` value immediately reflects the expiration based on the new rules, but existing objects may experience a delay. Most buckets will be transitioned within 24 hours but may take longer depending on bucket size. While objects are being migrated, you may see Object Lifecycle Management apply rules from the previous configuration.
+- An object is no longer billable once it has been deleted.
+- Buckets have a default lifecycle policy to expire multipart uploads seven days after initiation.
+
+## Configure a bucket’s lifecycle policy
+
+When you create an object lifecycle rule, you can specify which prefix you would like it to apply to.
+
+{{<Aside type="note">}}
+
+Object lifecycles has a 1000 rule maximum.
+
+{{</Aside>}}
+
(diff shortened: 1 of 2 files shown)


content/r2/objects/object-lifecycles.md
```diff
@@ -0,0 +1,152 @@
+---
+title: Object lifecycles
+pcx_content_type: how to
+---
+
+{{<beta heading="h1">}} Object lifecycles {{</beta>}}
+
+Object lifecycles affect how long objects uploaded to your bucket are kept.
+
+For example, you can create an object lifecycle rule to delete objects after 30 days, or you can set a rule to delete parts of multipart uploads.
```
harshal317: `delete parts of multipart uploads` isn't quite true. Maybe `or you can set a rule to abort multipart uploads after 30 days`

content/r2/objects/object-lifecycles.md
```diff
@@ -0,0 +1,152 @@
...
+
+When you apply a rule to a bucket, the rule automatically applies to any new objects uploaded after the rule creation. For existing objects, the new rule may take a while to apply and will usually occur within 24 hours but can take longer based on the number of objects in the bucket.
```
harshal317: If we're specifying this in the behavior section is it worth having here as well?

content/r2/objects/object-lifecycles.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Behavior
+
+- Objects will typically be removed from a bucket within 24 hours of the `x-amz-expiration` value.
+- When a new lifecycle policy is applied, the new object's `x-amz-expiration` value immediately reflects the expiration based on the new rules, but existing objects may experience a delay. Most buckets will be transitioned within 24 hours but may take longer depending on bucket size. While objects are being migrated, you may see Object Lifecycle Management apply rules from the previous configuration.
```
harshal317: `the new object's` -> `newly uploaded objects`?

Also `Most buckets will be transitioned within 24 hours but may take longer depending on bucket size` -> `Most objects will be transitioned within 24 hours but may take longer depending on the number of objects in the bucket.`

kodster28 (approved):

content/r2/objects/object-lifecycles.md
ferhatelmas:
```diff
- you may see Object Lifecycle Management apply rules from the previous configuration
+ you may see old applied rules from the previous configuration
```

content/r2/objects/object-lifecycles.md
ferhatelmas, line 1: we're using policy and rule interchangeably, might be a good idea to stick to one.

harshal317: I agree. I think a rule is fine to use to describe an individual element of the policy / configuration. We should use one of policy / configuration though.

content/r2/objects/object-lifecycles.md
dcpena: ✔️

content/r2/objects/object-lifecycles.md
dcpena: Removed because it's covered in behavior

content/r2/objects/object-lifecycles.md
dcpena: ✔️

dcpena: ✔️

content/r2/objects/object-lifecycles.md
dcpena: Good catch. I swapped out the few instances of configuration for "policy" instead.

PR #8050, opened 2023-03-15
[Support] Fix and improve many more codeblocks

Erisa:
Hello it's me again with another example of the things I spend my spare time doing.

I went through what I think is every codeblock in the English support docs now, and fixed/improved any that were rendering oddly.

I also added `?` to the end of some titles, that were missing for whatever reason.


jedecf (approved):
LGTM! merging! Thank you!

PR #8056, opened 2023-03-16
Update proxytype cloudflared ingress rule

stefanogramm:
updated broken - [proxyType](#proxyType) link


stefanogramm:
closing this, will create new one.  proxyType should be proxytype ( no capital )

PR #8060, opened 2023-03-16
[Docs site] Expand product grid

kodster28:
Include products that don't have a full folder to themselves


content/cloudflare-one/policies/access/_index.md
```diff
@@ -5,6 +5,11 @@ layout: single
 weight: 2
 meta:
   title: Access policies
+product_grid:
+  show: true
+  title: Access
+  group: Cloudflare One
+
 ---
```
pedrosousa:
```suggestion
---
```

layouts/shortcodes/product-grid.html
```diff
@@ -4,19 +4,34 @@
...
 
-{{- range $index, $item := $dataArray -}}
-{{- with index (index $item "product") "group" -}}
-{{- if ne (index (index $item "product") "show") false -}}
-{{ $availableOptions = $availableOptions | append . }}
+{{- range .Site.Pages  -}}
+{{- $relpermalink := .RelPermalink -}}
+{{- with index .Params "product_grid"  -}}
+{{- $dataArray = $dataArray | append (dict "group" .group "title" .title 
+ "show" (default false .show) "url" $relpermalink "wrap" (default false .wrap)) -}}
```
pedrosousa, line 20: Nit: I have mixed feelings about requiring the `show` frontmatter entry to enable an entry in the products list.
As soon as we have a "product_grid" object in the frontmatter, we probably want to show it in the products list. It's nice to be able to disable an entry by setting `show: false`, but 99% of the times we'll have to write `show: true`.

pedrosousa (approved):

layouts/shortcodes/product-grid.html
kodster28: That's actually probably a good team critique question -- I'll add it.

My thought was I just want folks to be extra careful and intentional about what we add to the product grid, so I was preferring the opt-in method.

PR #8063, opened 2023-03-16
handleRequest

Devlious:
The way it's on the tutorial doesn't work for me.


WalshyDev:
This is the service Worker format which isn't supported in Pages. What is documented is correct. If you need support, please join our Discord - https://discord.gg/cloudflaredev

PR #8072, opened 2023-03-16
[DNS] clarify DNS scan for Ent and API use cases 

RebeccaTamachiro:
PCX-5902


content/dns/_partials/_import-scan-info.md
RebeccaTamachiro, line 1: Although its name referred "import", this partial actually contained general information about the auto scan that is invoked in DNS, and some overlap with information that was already present in partial _dns-scan-procedure. Its content has then been reviewed and consolidated into _dns-auto-scan-info, that now lives together with the _dns-scan-procedure and is used in create-dns-records.md and setup.md

marciocloudflare (approved):

PR #8091, opened 2023-03-17
[Queues] Clarify waitUntil interaction

elithrar:


content/queues/platform/javascript-apis.md
```diff
@@ -131,6 +129,20 @@ addEventListener('queue', (event) => {
 ```
 
 In service worker syntax, `event` provides the same fields and methods as `MessageBatch`, as defined below, in addition to [`waitUntil()`](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil).
+  
+#### Using ctx.waitUntil()
+  
+Using [`ctx.waitUntil()`](/workers/runtime-apis/fetch-event/#waituntil) extends the lifetime of the "queue" event handler. It accepts a Promise-based task which the Workers runtime will execute before the handler terminates but without blocking the response.
```
a-robinson: > but without blocking the response.

This isn't true as currently implemented -- promises passed to `waitUntil` *do* currently block the response in `queue()` event handlers. This differs from `fetch()` handlers but matches the behavior of `scheduled()` handlers.

As discussed before I'd like to change it to behave more like `fetch`, but that'll require a compatibility flag and until then we should document how it actually works :)

a-robinson (changes requested):

content/queues/platform/javascript-apis.md
```diff
@@ -131,6 +129,20 @@ addEventListener('queue', (event) => {
...
+  
+{{<Aside type="warn">}}
+  
+We do not generally recommend using `ctx.waitUntil()` to process message batches directly, as any failure within the context of `waitUntil` can't be caught by your own code.
```
deadlypants1973:
```suggestion
We do not generally recommend using `ctx.waitUntil()` to process message batches directly, as any failure within the context of `waitUntil` cannot be caught by your own code.
```

content/queues/platform/javascript-apis.md
deadlypants1973:
```suggestion
Using [`ctx.waitUntil()`](/workers/runtime-apis/fetch-event/#waituntil) extends the lifetime of the queue event handler. It accepts a Promise-based task which the Workers runtime will execute before the handler terminates but without blocking the response.
```

Adding this on top of Alex's comment but queue should either be in `monospaces` or left without "". We avoid "" in docs unless we are `"quoting to point to a string in code"`.

elithrar: (Fixed this — it should have said blocking. Bad copy + paste)

elithrar: Copied from https://developers.cloudflare.com/workers/runtime-apis/fetch-event/#waituntil so should fix there 😄

a-robinson (approved):

PR #8103, opened 2023-03-18
Update deploy-a-hugo-site.md

bwc69151:
Update Hugo Version.


WalshyDev:
I really don't want to just accept PRs for updating an example version all the time. Hugo does a ton of releases, the docs are basically all out of date. This is an example version not saying "use this exact version"

cc: @deadlypants1973 for ref

bwc69151:
> the docs are basically always out of date.

Then isn't it just correct to keep the example updated? You need to realize that Hugo is on development mode, so any examples you have in your tutorial will always be outdated or present a breaking change. I followed your tutorial the first time I tried to deploy a Hugo site, and it turns out even your build image version is outdated!

WalshyDev:
> Then isn't it just correct to keep the example updated? 

No. A PR twice a week to update an example version number is just not required. It takes time for all of us to review and process these PRs. If we did that for every single framework, we'd never be able to get anything else done.
It's purely an example version. It isn't a "use this" nor does it have an indication of being the latest.

PR #8125, opened 2023-03-20
[R2] Added docs for location hints in dash

dcpena:
Addresses PCX-6262.

### DIFF
--- a/content/r2/buckets/data-location.md
+++ b/content/r2/buckets/data-location.md
@@ -23,7 +23,19 @@ Location Hints are optional parameters you can provide during bucket creation to
 
 Using Location Hints can be a good choice when you expect the majority of access to data in a bucket to come from a different location than where the create bucket request originates. Keep in mind Location Hints are a best effort and not a guarantee, and they should only be used as a way to optimize performance by placing regularly updated content closer to users.
 
-Currently, you can set the Location Hint via the `LocationConstraint` parameter using the S3 API:
+### Set hints via the Cloudflare dashboard
+
+You can choose to automatically create your bucket in the closest available region based on your location or choose a specific location from the list.
+
+1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com) and select **R2**.
+2. Select **Create bucket**.
+3. Enter a name for the bucket.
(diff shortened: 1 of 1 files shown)


kodster28 (approved):

content/r2/buckets/data-location.md
```diff
@@ -23,7 +23,19 @@ Location Hints are optional parameters you can provide during bucket creation to
...
 
-Currently, you can set the Location Hint via the `LocationConstraint` parameter using the S3 API:
+### Set hints via the Cloudflare dashboard
+
+You can choose to automatically create your bucket in the closest available region based on your location or choose a specific location from the list.
+
+1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com) and select **R2**.
+2. Select **Create bucket**.
+3. Enter a name for the bucket.
+4. Under **Location**, leave *None* selected to automatically use the region closest to your bucket or select a region from the list.
```
ferhatelmas: I might say `leave *None* selected for automatic or select a region from the list.` since this is explained before steps.

content/r2/buckets/data-location.md
```diff
@@ -23,7 +23,19 @@ Location Hints are optional parameters you can provide during bucket creation to
...
+5. Select **Create bucket**.
```
ferhatelmas: Submit ?

jonesphillip: The UI submit button says "Create bucket", so probably more clear to leave Create bucket in my opinion

dcpena: I'll add on to this so that it's clearer this is the last step. We don't actually have a Submit button; it's all "Create bucket."

ferhatelmas: I mean Submit `Create bucket` button. Sounded better than select.

dcpena: Ah, I see. Our style guide has us use "Select" for buttons, but I did update the last step to be clearer and changed it to "Select Create bucket to complete the bucket creation process."

ferhatelmas (approved):

content/r2/buckets/data-location.md
ferhatelmas: Thanks, looks better 👍🏻

PR #8128, opened 2023-03-20
[Stream] Added info on download limits

dcpena:
Added info on download limits for live recordings to address PCX-6263.

### DIFF
--- a/content/stream/changelog.md
+++ b/content/stream/changelog.md
@@ -11,7 +11,7 @@ rss: file
 
 ### Limits for downloadable MP4s for live recordings
 
-Previously, generating a download for a live recording exceeding four hours would take longer than four hours to encode and result in failure.
+Previously, generating a download for a live recording exceeding four hours resulted in failure.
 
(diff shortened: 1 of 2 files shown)


kodster28 (approved):

content/stream/changelog.md
```diff
@@ -7,6 +7,14 @@ rss: file
 
 # Changelog
 
+## 2023-03-21
+
+### Limits for downloadable MP4s for live recordings
+
+Previously, generating a download for a live recording exceeding four hours would take longer than four hours to encode and result in failure.
```
mbetz08: Can we simplify to something like:
`Previously, generating a download for a live recording exceeding four hours resulted in failure.`

I think it might be a bit confusing to provide the details around how long it took to encode and reveals some internals that aren't too relevant to users.

dcpena: Done!

mbetz08 (approved):

PR #8129, opened 2023-03-20
Fix example Cron Trigger typo

calebhailey:
Looks like this example was invalid (the 23rd minute of the 59th hour)? I presume this should be `59 23 * * *`?


deadlypants1973 (approved):

deadlypants1973:
thank you!

PR #8152, opened 2023-03-21
[DNS] Add a note for mixing proxy statuses

cicku:


hannesb0:
I am fine with the text, but would rather have that paragraph under `/dns/manage-dns-records/reference/proxied-dns-records/#proxied-records` (without the "Limitations" header) and maybe even in one of these highlighted note boxes. This way users quickly spot this point.

RebeccaTamachiro:
Thanks for pushing this. 
Continued here: https://github.com/cloudflare/cloudflare-docs/pull/8259

PR #8166, opened 2023-03-22
[Workers] Update Betas table for Email Workers and D1

Erisa:
Email Workers are now Public Beta and have docs.
D1 has docs.


WalshyDev (approved):
_tries to ignore table formatting_

deadlypants1973 (approved):

Issue #7843, opened 2023-03-02
Flashes of Unstyled Content (FOUC) errors


kodster28:
### Example URL(s)

All pages

### Actual Behavior

We're seeing [flashes of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) when loading pages in dev docs.

This applies to all pages (home / non-home layouts).

https://user-images.githubusercontent.com/26727299/222509752-3f495629-a0fe-47ac-852b-555b090bbef5.mov

### Expected Behavior

For this not to happen?

### Additional information

There are two versions of style.css loading, both a hashed and unhashed version. Not sure if that's related?

You can see it clearly when you disable cache + throttle your network connection to Slow 3G.

Seems to have been introduced with #7580. 

Tested a few fixes in #7820 #7776 #7775, but none of them seemed to work.

penalosa:
This is an odd one. Having looked into it, although it's pretty difficult to reproduce (in particular, it helpfully doesn't reproduce in dev mode), it seems like elements with transitions are somehow having their default state applied on page load, resulting in a run of their transition on page load (which explains the padding transition in the content, as well as the site title colour transition). This doesn't seem to happen in Firefox, and _appears_ to be a Chrome bug/optimisation (CSS in the head _should_ be render blocking, but doesn't appear to always be for transitions).

I'd like to investigate this more to isolate whether it is actually a Chrome bug, but in the meantime the fix is fairly simple; CSS rules with transitions should be moved to inline styles. I'll open a PR tomorrow for that.

As to why this is happening more _now_, I _think_ this is because the early hints header is being manually set in a _headers file. This bug is intermittent, and much more prevalent on slower networks/when the styles aren't cached, which I think increased when the new Vite additions were merged, in which I inadvertently broke early hints. That's just a hypothesis though, I'm not super sure on that.

kodster28:
For whatever reason, this appears to be less of an issue lately.

PR #8178, opened 2023-03-23
[Docs site] Partial CSS cleanup

kodster28:
Addresses a few parts of #7843 by removing transitions entirely that didn't add a ton of value, but also removes a few style types that we don't have on the home page layouts (code blocks, sidebar navigation, tutorial listings, etc.).


static/home.css
```diff
@@ -2994,272 +2497,24 @@ span.Tooltip---root {
     overflow: hidden
 }
 
-.ArchitectureDiagram {
-    --user-code-size: 25px;
-    --accent-color-rgb: var(--orange-rgb);
-    position: relative;
+.DocsNoscript {
```
pedrosousa, line 2500: I'm not sure if we're using this style.

pedrosousa (approved):

static/home.css
kodster28: That's a weird formatting thing where I didn't touch it but it shows as added. *shrug*

PR #8179, opened 2023-03-23
Replace mentions of CF-Authorization cookie with CF_Authorization

veggiedefender:
This page describes a `CF-Authorization` cookie, but the real name of that cookie is `CF_Authorization` with an underscore rather than a dash.


ranbel (approved):
lgtm. Thanks for making the update!

PR #8192, opened 2023-03-24
Add DDoS change log entries for 2023-03-24

jdesgats:


pedrosousa (approved):

pedrosousa:
The Cloudflare Pages build failed, but it should have no impact. Even so, I'm retrying the Pages deployment.
Update: Looks good now!

PR #8227, opened 2023-03-28
[API] Handle Redirects via Functions

### LINKED ISSUE #8161: _redirects entries on /api do not work due to Function
### Example URL(s)

https://developers.cloudflare.com/api/tokens/create/permissions

### Actual Behavior

https://developers.cloudflare.com/api/tokens/create/permissions Redirects to https://developers.cloudflare.com/api/ 

### Expected Behavior

`/api/` entries in `_redirects` should be respected.
That example URL should redirect to https://developers.cloudflare.com/fundamentals/api/reference/permissions

### Additional information

All the redirects listed here: https://github.com/cloudflare/cloudflare-docs/blob/production/content/_redirects#L132-L152
Do not work since /api is a Function now and `_redirects` entries do not work alongside Functions.

These old URLs are linked in the live Terraform docs: https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/api_token#nested-schema-for-policy

helloimalastair:
Functions take precedence over `_redirects`. This PR patches redirects for `/api` to be handled via the `/api` Function, so they are functional.

Closes #8161

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -131,26 +131,7 @@
 /argo-tunnel/trycloudflare/ /cloudflare-one/connections/connect-apps/install-and-setup/ 301
 
 # api
-/api/v4docs/ /fundamentals/api/how-to/make-api-calls/ 301
-/api/tokens/ /fundamentals/api/get-started/ 301
-/api/tokens/create/ /fundamentals/api/get-started/create-token/ 301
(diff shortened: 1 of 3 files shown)


kodster28 (approved):

kodster28:
Thank you very much, @helloimalastair!

PR #8237, opened 2023-03-29
Fix incorrect information regarding Logpull

erictung1999:
Logpull has some limitations compared to Logpush, especially the log fields that made available to Logpush vs Logpull.


soheiokamoto (approved):

angelampcosta:
@erictung1999 thank you for bringing this up. You are correct Logpull and Logpush have access to different log fields, we will update the docs accordingly.

PR #8247, opened 2023-03-29
[Docs site] Add filtering to product grid

kodster28:


layouts/shortcodes/product-grid.html
```diff
@@ -103,93 +332,150 @@
...
       } else {
-        section.style.display = "";
-        visibleSections = true
+        listItems[k].style.display = 'none';
       }
     }
+  }
+  // Trigger empty state message 
+  var emptyStateMessage = document.getElementById("emptyState")
+  var productGridContent = document.getElementById("ProductGrid--content")
```
pedrosousa:
```suggestion
  let emptyStateMessage = document.getElementById("emptyState");
  let productGridContent = document.getElementById("ProductGrid--content");
```
Using `var` here would set too broad a scope -- we can use `let` to have [block scope](https://www.w3schools.com/js/js_let.asp), since it's now [broadly supported](https://caniuse.com/?search=let). Unless needed, I'd replace all variable declarations using `var` with `let`. I tested locally using only `let` and everything seemed to work.

layouts/shortcodes/product-grid.html
```diff
@@ -103,93 +332,150 @@
...
+  if (!visibleItems) {
+    emptyStateMessage.style.display = "";
+    emptyStateMessage.children[0].style.display = "";
+    emptyStateMessage.children[0].setAttribute('href', "/search/#q=" + searchTerm.toLowerCase() + "&t=Docs" )
```
pedrosousa:
```suggestion
    emptyStateMessage.children[0].setAttribute('href', "/search/#q=" + searchTerm.toLowerCase() + "&t=Docs");
```

layouts/shortcodes/product-grid.html
```diff
@@ -103,93 +332,150 @@
...
+function resetFilters() {
+  var searchBox = document.getElementById('search');
+  searchBox.value = '';
+  var checkboxes = document.getElementsByClassName("ProductCheckboxItem");
+  for (var j = 0; j < checkboxes.length; j++) {
+      if (checkboxes[j].checked) {
+        checkboxes[j].checked = false;
     }
+  }
+  filterList()
```
pedrosousa:
```suggestion
  filterList();
```

layouts/shortcodes/product-grid.html
```diff
@@ -103,93 +332,150 @@
...
+}
+   
 
+function toggleVisibility(id) {
+  const element = document.getElementById(id);
+  if (element.style.display === 'block') {
+    element.style.display = 'none';
```
pedrosousa:
```suggestion
    element.style.display = '';
```
Setting it to empty string so that the CSS width-based behavior kicks in (setting it to 'none' would forcefully hide the element also in the desktop version when closing the pop-up in the mobile version).

layouts/shortcodes/product-grid.html
pedrosousa: In addition to filtering again, I'd also close the popup in the mobile version. I would create an additional function similar to `toggleVisibility()` that would hide an element if it was currently with `display: block`, and do nothing otherwise. This way, we wouldn't be toggling the popup when clearing the filters in the desktop version.

Proposal for such a function:
```js
function hideElement(id) {
  const element = document.getElementById(id);
  if (element.style.display === 'block') {
    element.style.display = '';
  }
}
```

Line 469 would turn into the following two lines:
```js
  filterList();
  hideElement('filter-popup');
```

layouts/shortcodes/product-grid.html
```diff
@@ -103,93 +332,150 @@
...
+  var checkboxes = document.getElementsByClassName("ProductCheckboxItem");
+  var searchBox = document.getElementById('search');
+  // Get all checked checkboxes
+  var visibleItems = false;
+  var checkedValues = [];
+  for (var j = 0; j < checkboxes.length; j++) {
+    if (checkboxes[j].checked) {
+      checkedValues.push(checkboxes[j].id);
+    }
+  }
```
pedrosousa:
```suggestion
  let numCheckboxes = 0;
  for (let j = 0; j < checkboxes.length; j++) {
    if (checkboxes[j].checked) {
      checkedValues.push(checkboxes[j].id);
      numCheckboxes++;
    }
  }
  let filterButton = document.getElementById('filter-button');
  if (numCheckboxes === 0) {
    filterButton.innerText = "Filter Products";
  } else {
    let suffix = numCheckboxes > 1 ? 's' : '';
    filterButton.innerText = `Filter Products (${numCheckboxes} group${suffix})`;
  }
```
I think that in the mobile version we should hint that there are filters being applied. I suggest that we update the filter button text to `Filter Results (N groups)` when we're filtering by one or more groups.

pedrosousa (approved):

PR #8279, opened 2023-03-31
Resolve Functions Block for (#8227)

helloimalastair:
Fixes issue created by #8227 preventing the API docs from rendering.
(Sort Of) Reverts #8274

Explanation:
`wrangler pages dev` does not always appear to enforce trailing slashes. A small block was added to normalize URLs passed to the API Handler, which inadvertently began adding trailing slashes to **every** URL that didn't have them, including assets. This PR resolves this issue by automatically removing trailing slashes when required, but **only** for redirect resolution. URL used for retrieving assets is unmodified.

Apologies to @kodster28 and @ranbel for having to deal with this issue!


kodster28 (approved):

kodster28:
Absolutely no worries @helloimalastair, probably should've done a better job testing the api docs rendering myself before merging the first time.

PR #8334, opened 2023-04-04
Added a question (though I don't have the answer)

wolruf:
Can we live change a Logpush job or do we need to delete and recreate ?


cicku:
This is not a place for QA, no exact minute can be given.

angelampcosta:
We will follow up in a ticket.

PR #8342, opened 2023-04-04
bindings per environment guidance w/ example toml

Jack-R-Long:
Added a simple example for bindings per environment.  Couldn't find any docs on this and it caused a minor issue in our project.

Problem: Some bindings are not inherited by environments.  `wrangler` prints an error message about this but there is no mention of it in the docs.

Solution:  Added section to `environment.md` with guidance and example toml

### DIFF
--- a/content/workers/platform/environments.md
+++ b/content/workers/platform/environments.md
@@ -228,6 +228,30 @@ Published my-worker
   https://my-worker.<your-subdomain>.workers.dev
 ```
 
+### Bindings and Environments
+
+Bindings are defined per environment and are not shared between environments.  Below is an example of a service binding for a dev and production environment:
+
+```toml
+---
+filename: wrangler.toml
+---
+name = "my-worker-dev"
+
+[[ services ]]
+binding = "SERVICE_WORKER"
+service = "service-worker"
+environment = "dev"
+
+[env.production]
+name = "my-worker"
+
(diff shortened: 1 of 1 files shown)


Jack-R-Long:
@deadlypants1973 Ready to merge IMO.  Added the guidance and example toml.

If you would like anything added or changed please let me know! 😅

content/workers/platform/environments.md
```diff
@@ -228,6 +228,30 @@ Published my-worker
...
+```toml
+---
+filename: wrangler.toml
+---
+name = "my-worker-dev"
+
+[[ services ]]
+binding = "SERVICE_WORKER"
+service = "service-worker"
+environment = "dev"
```
KianNH: The `environment` key refers to Service Environments which aren't around anymore - so unless you have a pretty legacy Worker that used those, it's impossible to have a Worker with an environment of `dev` right now.

It's an optional property and is probably worth omitting in general for now. It'd be good to link to the other non-inheritable keys (i.e `define`) which are available at https://developers.cloudflare.com/workers/wrangler/configuration/#non-inheritable-keys

Jack-R-Long: Oh interesting, did not realize that.  Ok I removed the `environment` key and added a link to the non-inheritable keys.

Appreciate the insight!

PR #8347, opened 2023-04-04
[ZT] Add SCIM to identity-based policies

maxvp:
- Move IdP aside
- Add enforce session policy
- Add SCIM note


content/cloudflare-one/policies/filtering/identity-selectors.md
```diff
@@ -8,6 +8,17 @@ weight: 6
 
 With Cloudflare Zero Trust, you can create Secure Web Gateway policies that filter outbound traffic down to the user identity level. To do that, you can build DNS, HTTP or Network policies using a set of [identity-based selectors](#identity-based-selectors). These selectors require you to deploy the Zero Trust WARP client in [Gateway with WARP mode](/cloudflare-one/connections/connect-devices/warp/configure-warp/warp-modes/).
 
+Gateway checks identity when a user logs in or re-authenticates. To check your users' identities at regular intervals, you can [enforce a WARP client session duration](/cloudflare-one/policies/filtering/enforce-sessions/).
```
ranbel: Suggest combining Line 11 with the content in the Aside. Maybe create a new section called "When does Gateway check identity" or something like that. 

Enforce Session Duration is how an admin can force a user to re-authenticate. The two options in the aside are how the user can manually trigger the re-authentication flow.

ranbel (approved):

PR #8362, opened 2023-04-05
Fix TOML section in D1 docs

kotx:
```toml
[[ d1_databases ]]
    binding = "DB"  
```
should be this
```toml
[[d1_databases]]
    binding = "DB"  
```

Otherwise this happens
image


WalshyDev:
This syntax is valid, it looks like the "Even Better TOML" extension yells at this whereas VSCode by default, and with "Better TOML" it doesn't.

We can merge this anyway for consistency but worth noting it isn't technically wrong. Extensions just be picky.

WalshyDev (approved):

PR #8366, opened 2023-04-05
[Product grid] Allow multiple product group tags

kodster28:
PCX-6469


layouts/shortcodes/product-grid.html
```diff
@@ -443,21 +448,11 @@ <h2>Filters</h2>
...
   for (let k = 0; k < listItems.length; k++) {
-    let groupMatch = true;
-    let previewMatch = true;
-    let searchMatch = true;
-
-    if (groupIdValues.length > 0) {
-      groupMatch = (groupIdValues.indexOf(listItems[k].getAttribute('data-product-group')) >= 0);
-    } 
+    let groupTags = listItems[k].getAttribute('data-product-group').split("|").filter(x => x !== '');
+    let groupMatch = groupIdValues.length === 0 || groupTags.some(tag => groupIdValues.includes(tag));
```
kodster28, line 457: Found this online, seemed like an easier and more succinct way to get the behavior we wanted.

layouts/shortcodes/product-grid.html
```diff
@@ -6,17 +6,34 @@
 {{- $productName :=  replace .Name ".yml" "" -}}
 {{- $productData := index $.Site.Data $productName -}}
 {{- with index $productData "product" -}}
+
+{{- $additional_groups := slice -}}
+{{- with .additional_groups -}}
+{{- range . -}}
+{{- $additional_groups = $additional_groups | append . -}}
+{{- end -}}
+{{- end -}}
```
pedrosousa: (I don't know much about Hugo templating, so take this comment with a grain of salt.)
We could probably simplify the two blocks (`with` and `range`) into a single one with:
```
{{- range .additional_groups -}}
  {{- $additional_groups = $additional_groups | append . -}}
{{- end -}}
```

layouts/shortcodes/product-grid.html
```diff
@@ -6,17 +6,34 @@
...
+
 {{- $dataArray = $dataArray | append 
 (dict "group" .group "title" .title  "show" (default true .show) "url" 
-.url "wrap" (default false .wrap) "preview_tryout" (default false .preview_tryout) ) -}}
+.url "wrap" (default false .wrap) "preview_tryout" (default false .preview_tryout) 
+"additional_groups" (default ("") ($additional_groups))) -}}
```
pedrosousa: The default value for `additional_groups` value should probably be an empty slice/array (obtained with `(slice)`?) instead of an empty string. This way, you would avoid the `{{- if ne .additional_groups "" -}}` statement below, since iterating an empty slice would in fact do nothing.

layouts/shortcodes/product-grid.html
```diff
@@ -385,8 +402,9 @@ <h2>Filters</h2>
           class="ProductGrid--link"
           data-wrap-title="{{ $wrap }}"
           href="{{ .url }}"
-          data-product-group="{{ .group }}"
-          data-preview-tryout="{{ .preview_tryout }}"
+          data-product-group="{{- .group -}}{{- if ne .additional_groups "" -}}
+          {{- range .additional_groups -}}|{{.}}{{- end -}}{{- end -}}"
```
pedrosousa: Considering my previous suggestion of using an empty slice when there are no additional groups, and using the [`delimit`](https://gohugo.io/functions/delimit/) function, this could be something like the following:
```
          data-product-group="{{- delimit (union (slice .group) .additional_groups) "|" -}}"
```

layouts/shortcodes/product-grid.html
pedrosousa: Looks good! I usually check [caniuse.com](https://caniuse.com/) to make sure there's broad support for a JavaScript language element, and it seems there is for both `filter` and `some`.

data/data-localization.yml
```diff
@@ -5,6 +5,7 @@ product:
   title: Data Localization Suite
   url: /data-localization/
   group: Cloudflare essentials
+  additional_groups: [Analytics, Application performance, Application security, Cloudflare One, Developer platform, Network services, Storage]
```
pedrosousa, line 8:
```suggestion
  additional_groups: [Analytics, Application performance, Application security, Cloudflare One, Developer platform, Network services, Storage]
```
Since the group names appear repeatedly, we should consider having some sort of "keyword to group name" mapping somewhere, thinking of future localization.

data/network-analytics.yml
```diff
@@ -3,5 +3,6 @@ name: Network Analytics
 product:
   title: Network Analytics
   group: Network services
+  additional_group: Analytics
```
pedrosousa:
```suggestion
  additional_groups: [Analytics]
```

pedrosousa (approved):

layouts/shortcodes/product-grid.html
kodster28: Yep, that seems to work the exact same. Good catch.

data/data-localization.yml
kodster28: Flagged as a future improvement in PCX-6567

PR #8372, opened 2023-04-06
[DNS] Add callout regarding CNAME flattening and domain verification

Erisa:
Resolves SPM-1594


RebeccaTamachiro:
Hey Erisa! Once again, thanks for identifying this gap and for creating the PR. We really appreciate that.
Please ignore the images that my commits created here. 
About the changes to DNS content, it was really just a matter of making sure that the callout was not decontextualized within the ref page and giving it some more visibility by adding to the main CNAME flattening article.

I've also created a follow-up ticket for me (PCX-6698) thinking about ways we can make this information even more visible to customers looking to troubleshoot on their own.

RebeccaTamachiro (approved):

PR #8391, opened 2023-04-07
Clarify that no logs are stored on Cloudflare with v2

ngayerie:
Clarify that no logs are stored on Cloudflare with v2, following an email thread with JPL

### DIFF
--- a/content/logs/about/index.md
+++ b/content/logs/about/index.md
@@ -10,7 +10,7 @@ Logpush delivers logs in less than one minute with batches of no more than 100,0
 
 Prior to mid-2020, Logpush sent logs once every five minutes (referred to as Logpush v1). The change to more frequent log pushing allows Cloudflare to deliver information to you as close to real time as possible in smaller files. You may receive log files that contain fewer lines - that is expected. If you have legacy Logpush jobs configured to the old settings, use the Logpush API to upgrade your job to Logpush v2. All new jobs will use Logpush v2 by default.
 
-Please note that with the Logpush v2 no logs are stored on the Cloudflare side.
+Logpush v2 does not store logs; we attempt to send logs as quickly as they arrive. In the event of an error or loss of connection, logs may be buffered for up to a few hours before being dropped. Please note that Logpush itself does not provide a way to store or search logs. 
 
 ## Availability
 
--- a/content/logs/about/index.md
+++ b/content/logs/about/index.md
@@ -10,6 +10,8 @@ Logpush delivers logs in less than one minute with batches of no more than 100,0
 
 Prior to mid-2020, Logpush sent logs once every five minutes (referred to as Logpush v1). The change to more frequent log pushing allows Cloudflare to deliver information to you as close to real time as possible in smaller files. You may receive log files that contain fewer lines - that is expected. If you have legacy Logpush jobs configured to the old settings, use the Logpush API to upgrade your job to Logpush v2. All new jobs will use Logpush v2 by default.
 
+Please note that with the Logpush v2 no logs are stored on the Cloudflare side.
+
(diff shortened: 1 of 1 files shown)


content/logs/about/index.md
```diff
@@ -10,6 +10,8 @@ Logpush delivers logs in less than one minute with batches of no more than 100,0
 
 Prior to mid-2020, Logpush sent logs once every five minutes (referred to as Logpush v1). The change to more frequent log pushing allows Cloudflare to deliver information to you as close to real time as possible in smaller files. You may receive log files that contain fewer lines - that is expected. If you have legacy Logpush jobs configured to the old settings, use the Logpush API to upgrade your job to Logpush v2. All new jobs will use Logpush v2 by default.
 
+Please note that with the Logpush v2 no logs are stored on the Cloudflare side.
```
cascadingstyletrees: Changing the wording a bit, I suggest something like:

```For customers that have been migrated to Logpush V2, by default, full request data is not retained in Cloudflare. Logs are streamed to your cloud storage service, and Cloudflare Support only has access to partial data. ```

**Review by cascadingstyletrees — CHANGES_REQUESTED** [review]
I've suggested a few changes to the wording.

**Review by tlozoot — CHANGES_REQUESTED** [review]
I'd prefer not to add this line here in developer docs since, as always, it's a bit more complicated than that :)

It's true that *Logpush* does not store logs - we try to send logs to customers as quickly as possible and not store them. (There can be buffering for up to a few hours, but that's generally not material)

However, as defined our Privacy Policy, there is a category of data called "Customer Logs" which includes both logs products (like Logpush) as well as Analytics (e.g. what you get via the graphQL API): https://www.cloudflare.com/privacypolicy/

Using Logpush does not really impact the bigger picture of "Customer Logs"; even if Logpush is disabled, we'll still store analytics data.

Most commonly, of a customer asks us not to store logs, they actually mean the legal definition of "Customer Logs" -- they are not asking about the behavior of Logpush. And the most common reason customers ask for this is because they care _where_ logs are stored. In that case we should point them to the Data Localization Suite and in particular the Metadata Boundary: https://developers.cloudflare.com/data-localization/metadata-boundary/

BTW, none of this is meant to be a secret (which is why I'm explaining via GitHub!). Happy to incorporate into docs but I think this single line is misleading. 

cc @angelampcosta as well if she has ideas!

**Comment by tlozoot** [comment]
> I'd prefer not to add this line here in developer docs since, as always, it's a bit more complicated than that :)

Thinking about this more... I understand your intent here (which came from a comment I made!) which is to clarify that *Logpush* does not store logs in general, and that logs are not searchable after being sent via Logpush.

Maybe we can add a line like:


> Logpush v2 does not store logs; we attempt to send logs as quickly as they arrive. In the event of an error or loss of connection, logs may be buffered for up to a few hours before being dropped. Please note that Logpush itself does not provide a way to store or search logs; for that you can use _____


I'm happy to say something to that effect as long as we don't make blanket statements about "not storing logs"

File: content/logs/about/index.md (hunk 1 above)
Comment (angelampcosta) [inline content/logs/about/index.md]:
```suggestion
Logpush v2 does not store logs; we attempt to send logs as quickly as they arrive. In the event of an error or loss of connection, logs may be buffered for up to a few hours before being dropped. Please note that Logpush itself does not provide a way to store or search logs. 
```

tlozoot: If I can suggest a change to Angela's suggestion -- DLS/Metadata Boundary do not provide a way to store search logs, so I think we should cut everything after "for that you can use...."

angelampcosta:
@tlozoot can we merge the PR?

tlozoot (approved):

PR #8393, opened 2023-04-07
Update worker-variables.md fix broken function example

mountainash:
The `email` variable is passed to the `digestMessage()` function. The  `text` variable would be `undefined`.

As requested here: https://discord.com/channels/595317990191398933/917505178016579605/1093902711231086632


bjesus:
Confirming this fix is correct!

haleycode (approved):

PR #8394, opened 2023-04-07
[Docs Site] API Reference: Rename DNS Edit permission to DNS Write

dwmcc:
In the CloudFlare [permission groups API](https://developers.cloudflare.com/api/operations/permission-groups-list-permission-groups), the two zone-level DNS permissions are `DNS Read` and `DNS Write`.
However, "Edit" is listed on the docs page instead of "Write" which is not a valid permission according to the API response.

This change updates the docs page to correctly list the permission name and fixes minor markdown formatting.


kodster28 (approved):
Confirmed by looking at dash api calls. Thanks for the catch, @dwmcc

PR #8404, opened 2023-04-09
[ZT] Formatting fixes in application-check.md

colinbarr:
* Updated the `codesign` command example to include a line break. The current way it displays looks as if the `stdout` data is part of command itself.
* Fixed nested list so that it doesn't display as a codeblock - it currently shows 1, 2 and 3 in a codeblock, then `4.` as text, and 5 as `e`.


ranbel (approved):
Lgtm. Thanks for updating the docs!

colinbarr:
Thanks @ranbel! It's still not quite right, there's another line break to remove to fix the nested list- I'll open a follow up PR.

PR #8410, opened 2023-04-10
Fix algolia search duplicates

kristianfreeman:
ty @elithrar for fix :)


elithrar (approved):

elithrar:
Secret is a false positive: it's the public read-only Algolia key.

PR #8415, opened 2023-04-10
add binding cookie call out to short lived cert

kennyj42:
I think the note formatting could be better but just proposing the initial draft.

### DIFF
--- a/content/cloudflare-one/identity/users/short-lived-certificates.md
+++ b/content/cloudflare-one/identity/users/short-lived-certificates.md
@@ -12,6 +12,8 @@ Cloudflare Access can replace traditional SSH key models with short-lived certif
 
 Cloudflare Access removes the burden on the end user of generating a key, while also improving security of access to infrastructure with ephemeral certificates.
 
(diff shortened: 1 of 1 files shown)


content/cloudflare-one/identity/users/short-lived-certificates.md
```diff
@@ -12,6 +12,8 @@ Cloudflare Access can replace traditional SSH key models with short-lived certif
 
 Cloudflare Access removes the burden on the end user of generating a key, while also improving security of access to infrastructure with ephemeral certificates.
 
+Note: Short-Lived Certificate authentication is not compatible with the [Access Binding Cookie]#(cloudflare-one/identity/authorization-cookie/#binding-cookie).
```
ranbel, line 15: 1.) Why is the cookie not compatible?
2.) Where in the short-lived cert setup flow are users enabling the binding cookie, and what are they trying to accomplish?

ranbel:
We call out SSH on the binding cookie page: https://developers.cloudflare.com/cloudflare-one/identity/authorization-cookie/#when-not-to-use-binding-cookie

Issue #6726, opened 2022-11-16
Image optimizations / improve documentation website speed


jbergstroem:
### Example URL(s)

This applies site-wide.

While reviewing [pagespeed insights](https://pagespeed.web.dev/report?url=https%3A%2F%2Fdevelopers.cloudflare.com%2Fpages%2Fplatform%2Fdeploy-hooks%2F), one low hanging fruit seems to be optimizing most images. This can be done a few ways:

1. Use a image resizing service like cloudflare images
2. Optimize images (lossless or lossy)
3. Resize / use `<picture>` to serve more relevant versions of said image

There's also other small quick wins, specific to images:
1. Set proper caching headers for images / static assets ( for instance a 2MB png with 60s timeout should probably be 1d+)
2. Use `defer` 

I did a small experiment of using lossless optimization via imageoptim locally. It shaved about 30% of the file-size across all images. See below.

### Actual Behavior

Before running imageoptim on ~950 files:
```console
~/wrk/oss/cloudflare-docs ±production $ dust -n 1 -b -c content
295M ┌── content
```

### Expected Behavior

After running imageoptim on ~950 files:
```console
~/wrk/oss/cloudflare-docs ±chore/imageoptim $ dust -n 1 -b -c  content
213M ┌── content
```

### Additional information

This could also be incorporated into a github action (pngcrush or similar) to keep all changed images in check.

eih64225:
They use Hugo which also supports image optimisation. But Cloudflare images is also a plus.

kodster28:
Addressed in some areas by https://github.com/cloudflare/cloudflare-docs/pull/8423 (and will be fully addressed in future PRs).

Thanks for the suggestion, @jbergstroem!

PR #8423, opened 2023-04-11
[Docs site] PoC image optimization

### LINKED ISSUE #6199: Add `width` and `height` attributes to images to reduce CLS
Cumulative Layout Shift happens when a page has to repaint things, and then all of a sudden things shift around the page. This can be very distracting and happens a lot on the docs site due to images not having any set dimensions and therefore no aspect ratios to adhere to.

Some good references:

- https://web.dev/optimize-cls/
- https://www.smashingmagazine.com/2021/06/how-to-fix-cumulative-layout-shift-issues/
- https://jakearchibald.com/2022/img-aspect-ratio/

### Actual Behavior

Take https://developers.cloudflare.com/analytics/graphql-api/tutorials/build-your-own-analytics/ for example. This is a page with an image at the top, and the CLS is `0.128`- WAY higher than the desired `0`.

Here's a quick video demonstrating this: https://up.jross.me/r15pv1c4
Notice how the "Code" section is bumped down the page as the image loads in? That's the layout shift occurring and is very jarring especially if the image takes a while to load and I'm already reading the doc. 

This impacts every page on the docs site that has images above the fold.

### Expected Behavior

Images should have a set `width` and `height` so that the browser can reserve space for them when rendering. Setting these values won't prevent the images from being downsized when needed either, it'll simply hint the browser of the image's aspect ratio so it can reserve the space needed to render it.

### Additional information

I started work on this a bit, but my lack of experience with Hugo is making it more difficult than I'd like.

Hugo offers [Render Hooks](https://gohugo.io/templates/render-hooks/) though, which should make this very possible. I started with a `layouts/_default/markup/render-image.html` file with this:
```
{{- $src := (.Destination | safeURL) -}}
{{- $alt := .Text -}}
{{- $title := .Title -}}
{{ $resource := .Page.Resources.GetMatch .Destination }}
<!-- handle external links -->
{{- if strings.HasPrefix $src "http" -}}
	<img
		src="{{ .Destination | safeURL }}"
		alt="{{ .Text }}"
		{{ with .Title}} title="{{ . }}"{{ end }}
	/>
<!-- handle absolute URLs-->
{{- else if not (strings.HasPrefix $src "..") -}}
  {{ $imageurl := print "content/" $src }}
  {{ $imageurl }}

  {{- $img := imageConfig $imageurl -}}

  {{ with $img}}
    <img
      src="{{ $src }}"
      alt="{{ $alt }}"
      {{ with $title}} title="{{ . }}"{{ end }}
      {{ with .Width}} width="{{ . }}"{{ end }}
      {{ with .Height}} height="{{ . }}"{{ end }}
    />
  {{ end }}
<!-- handle images with relative pathing -->
{{- else -}}
  {{ $url := urls.Parse .Destination }}
  {{ $combinedpath := (path.Join .Page.File.Dir .Destination)}}
  {{$combinedpath}}

  {{ $imageurl := print "content/" $combinedpath }}
  {{ $imageurl }}

  {{- $img := imageConfig $imageurl -}}

  {{ with $img}}
    <img
      src="{{ $src }}"
      alt="{{ $alt }}"
      {{ with $title}} title="{{ . }}"{{ end }}
      {{ with .Width}} width="{{ . }}"{{ end }}
      {{ with .Height}} height="{{ . }}"{{ end }}
    />
  {{ end }}
{{- end -}}
```

It mostly works, but errors on a couple of edge-cases and needs some more thought (handling remote `http` urls, other relative paths like `../`, etc.). It should be very possible with the `Resources` and/or `imageConfig` stuff though, for someone more familiar with Hugo than me.

I also tweaked the `.DocsMarkdown img` css rule to include `height: auto`, and with both the `height` and `width` attributes set on images, it reduces the CLS on pages to zero, resulting in a much nicer reading experience for any image-heavy pages as the content loads.

kodster28:
Exploration for addressing #6726 and #6199. 

Uses render hooks to automatically compress images (will need some work to move images into the `/assets` folder) + add `height` and `width` estimates based on the intrinsic size of the images (but then also not affect the rendered size of the images).

I'm not _exactly_ sure how I feel about the estimates part of this. From looking around ([web.dev](https://web.dev/serve-images-with-correct-dimensions/#avoid-layout-shifts-by-specifying-dimensions)), seems like it's relatively kosher and better than nothing for reducing CLS. Especially doing it automatically, which could make a larger impact and is more reasonable than specifying values on each image individually.

TBD work around making this less brittle and error prone.

Original --> https://developers.cloudflare.com/assets/proxy-status-screenshot-193baccf.png (55 KB)
Updated --> https://image-optimization.cloudflare-docs-7ou.pages.dev/assets/proxy-status-screenshot_hua0b69e1ec36a9eb13505b2a0b89dbfcd_55521_900x0_resize_q70_box_3-abc6d6b4.png (16 KB) -- might be too optimized as it's a bit blurry


kodster28:
## Notes

Average compression seems to be tracking the quality inputs (~25% optimization, less in some cases & more in others). In pretty thorough spot checks, doesn't seem to affect the layout of any images that were changed.

This will be the first PR, with follow-up PRs made by me for smaller product areas and then internal tickets for some of the more beefy areas (Fundamentals / CFZT / Magic Wan).

The hardest part of the work isn't moving over assets, but rather a) verifying that current assets are still being used (quasi image audit) + b) verifying the resizing isn't causing formatting issues. 

Will discuss internally about whether a/b are truly necessary, which means this change could go along a lot faster.

Stepped approach also gives us more flexibility in terms of building out edge cases as we approach them.

## Comparison

Current state on an [image-heavy page](https://developers.cloudflare.com/analytics/graphql-api/getting-started/compose-graphql-query/) vs the [revised page](https://image-optimization.cloudflare-docs-7ou.pages.dev/analytics/graphql-api/getting-started/compose-graphql-query/).

### Current - Desktop

Screenshot 2023-04-12 at 1 50 22 PM

### Updated - Desktop

Screenshot 2023-04-12 at 1 49 55 PM

### Current - Mobile

Screenshot 2023-04-12 at 1 46 27 PM

### Updated - Mobile

Screenshot 2023-04-12 at 1 46 13 PM

Cherry:
Those are some impressive perf gains!

layouts/_default/_markup/render-image.html
```diff
@@ -0,0 +1,28 @@
+{{- $img := resources.Get .Destination -}}
+<!-- This value will only be computed if the image lives in the top-level /assets folder -->
+{{- $destination := .Destination -}}
+{{- with $img -}}
+<!-- .gifs can't be resized the same way -->
+{{- if not (in $destination ".gif") -}}
```
pedrosousa:
```suggestion
{{- with $img -}}
<!-- .gifs can't be resized the same way -->
{{- if not (in $.Destination ".gif") -}}
```
I don't think we need the `$destination` variable to get `.Destination` from the initial scope, since we can use `$`. Also, we're not using the variable again.

pedrosousa (approved):

layouts/_default/_markup/render-image.html
kodster28: Addressed in my commit

kodster28:
And extra thanks to @pedrosousa for #8470, which'll save the Q/A check from all those popup banners.

content/cloudflare-for-platforms/cloudflare-for-saas/reference/versioning.md
```diff
@@ -85,7 +85,7 @@ Before your migration, you should:
     - Via the API: `https://api.cloudflare.com/client/v4/zones/{zone_tag}/custom_hostnames?hostname_status={status}`
     - Via a CSV file (provided by your SE)
     - Via the Cloudflare dashboard:
-      ![Review SSL migration status in the dashboard](/ssl/static/ssl-migration-status.png)
+      ![Review SSL migration status in the dashboard](/cloudflare-for-platforms/static/ssl-migration-status.png)
```
marciocloudflare, line 88: So this one here doesn't go to the `/images/` root folder?

content/dns/zone-setups/partial-setup/setup.md
```diff
@@ -30,7 +30,7 @@ A partial setup is only available to customers on a Business or Enterprise plan.
 
 5.  For **Advanced Actions**, click **Convert to CNAME DNS Setup**.
 
-    ![On your domain's overview page, click Convert to CNAME DNS Setup](/dns/static/dns_cname_setup.png)
+    ![On your domain's overview page, click Convert to CNAME DNS Setup](/images/dns/dns_cname_setup.png)
```
marciocloudflare, line 33: This image is different from what we have in production, although I think you've just cropped it.

content/cloudflare-for-platforms/cloudflare-for-saas/reference/versioning.md
kodster28: Yeah, didn't get to moving those yet. Will come in a future PR.

content/dns/zone-setups/partial-setup/setup.md
kodster28: Yeah... that was intentional. Realized the other one was sorta too large to be helpful + might display the Zone ID.

marciocloudflare (approved):

PR #8430, opened 2023-04-11
Param filtering for product grid

kodster28:
PCX-6482

### DIFF
--- a/layouts/shortcodes/product-grid.html
+++ b/layouts/shortcodes/product-grid.html
@@ -397,6 +397,7 @@ span a img {
           href="{{ .url }}"
           data-product-group="{{- delimit (union (slice .group) .additional_groups) "|" -}}"
           data-preview-tryout="{{- .preview_tryout -}}"
+          style="display: none"
         >
           <span>{{ .title }}</span>
         </a>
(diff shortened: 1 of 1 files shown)


layouts/shortcodes/product-grid.html
```diff
@@ -397,6 +397,7 @@ <h2>Filters</h2>
           href="{{ .url }}"
           data-product-group="{{- delimit (union (slice .group) .additional_groups) "|" -}}"
           data-preview-tryout="{{- .preview_tryout -}}"
+          style="display: none"
```
kodster28, line 400: This is necessary to prevent flashing if you're filtering from a link. Loads a touch slower, but think it's worth avoiding the flash before the filtering.

layouts/shortcodes/product-grid.html
```diff
@@ -479,6 +505,28 @@ <h2>Filters</h2>
     emptyStateMessage.style.display = "none";
     productGridContent.style.display = "grid";
   }
+
+  // Add parameters to URL
```
kodster28: Made a decision to not add search values to the params. Think that's correct b/c the main use case is sharing a view with specific groups / preview available.

pedrosousa (approved):

PR #8440, opened 2023-04-12
[R2] Add HEAD to supported methods

KianNH:
Personally I would clarify and have it like [fetch an object's metadata](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html) as a link to the HeadObject operation but that would entail doing it for the rest so I'll leave it up to the reviewer.


KianNH:
cc @elithrar @sdnts

elithrar (approved):

PR #8481, opened 2023-04-14
Fix schema function

kristianfreeman:
This PR fixes two things in the /schema Pages function

1) Corrects an issue where the function was throwing an exception when
   trying to parse tags for an OpenAPI schema endpoint
2) When an error occurs, fall back to the raw GitHub JSON data to not
   cause any downtime on the API Docs frontend

Follow-up work will be to notify the team (Sentry or something else) in the catch portion of the block so that if an exception is occurring in this function, it can be resolved without downtime to the API Docs frontend

### DIFF
--- a/functions/schema.js
+++ b/functions/schema.js
@@ -6,35 +6,47 @@ export async function onRequestGet() {
   const cache = caches.default
   let response = await cache.match(req)
 
-  if (!response) {
-    response = await fetch(req)
-    let schema = await response.json()
+  try {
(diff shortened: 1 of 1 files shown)


kristianfreeman:
/schema endpoint is functioning again: https://kristian-schema-fn-fix.cloudflare-docs-7ou.pages.dev/schema

kodster28 (approved):
Can confirm I'm seeing the schema at `/schema` now

PR #8491, opened 2023-04-14
[R2] Transparent decompression

dcpena:
Addresses PCX-5647.


dcpena:
@KianNH I definitely used your very well written blog as I documented this, and you'll notice the examples are very familiar. 😅  would also love your feedback on this PR when you're able.

content/r2/objects/transparent-compression.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Transparent decompression
+pcx-content-type: how-to
+---
+
+# Transparent decompression
+
+Transparent decompression reduces the size of stored files, which can help with reducing bandwidth to your origin. With transparent decompression, files are automatically decompressed, but you can still access and use the files or data in their original format. When supported, Cloudflare will always request the compressed content from your web server.
```
Frederik-Baetens: > which can help with reducing bandwidth to your origin

Feel like we should leave that out since bandwidth is free with r2. Just reducing the size of stored files is the relevant part for R2.

Frederik-Baetens:
```suggestion
Transparent decompression reduces the size of stored files. With transparent decompression, compressed files are automatically decompressed, but you can still access and use the files or data in their original format.
```

Also left that last part off, since it's not relevant to r2.

content/r2/objects/transparent-compression.md
```diff
@@ -0,0 +1,82 @@
...
+    // Returns the compressed response.
+    return new Response(compressed, {
+      headers: {
+        'content-encoding': 'gzip'
+      },
+      // Specifies not to compress content. Content has already been compressed.
+      encodeBody: "manual"
+    })
+  },
+};
```
Frederik-Baetens, line 45: This example is correct, but I feel like it's not all that relevant in the context of R2. I'll work on adding some examples to a public bucket + associated curl commands that show this behavior in a way that's more specific to R2.

content/r2/objects/transparent-compression.md
```diff
@@ -0,0 +1,82 @@
...
+
+Transparent decompression reduces the size of stored files, which can help with reducing bandwidth to your origin. With transparent decompression, files are automatically decompressed, but you can still access and use the files or data in their original format. When supported, Cloudflare will always request the compressed content from your web server.
+
+## Accept-Encoding
+
+The `Accept-Encoding` request HTTP header indicates support for compressed content. However, if the header is missing, the content will still be decompressed during the request. 
+
+{{<Aside type="note">}}
+
+The `Content-Encoding` header is optional, but if you want Cloudflare to return a `gzip` compressed file, you must remove the header.
```
Frederik-Baetens: This is a bit incomplete, indeed, if an R2 object doesn't have Content-Encoding set in its metadata, we always return the object as is. However, with the (Content-Encoding: gzip) header set, we will still return the gzip compressed file, if the request from the client contains an Accept-Encoding: gzip header.

content/r2/objects/transparent-compression.md
```diff
@@ -0,0 +1,82 @@
...
+{
+  "uuid": "18b70f66-e84d-4070-8287-d114f4da1293"
+}
+```
+
+## Transfer-Encoding
+
+When the[`Transfer-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding) header is used and the directive is set to `chunked`, the data is sent in a series of chunks and the `Content-Length` is omitted by default.
+
+For version 1.1 of the HTTP protocol, the chunked transfer mechanism is always considered acceptable, even if it is not listed in the `Transfer-Encoding` request header field. The Workers platform always knows whether it has the body size in advance and will choose whether to use `Content-Length` vs. `Transfer-Encoding: chunked` header, which takes precedence over whatever was specified in the request.
```
Frederik-Baetens, line 82: I'm not sure if this applies to R2. I'll test that out.

KianNH:
Yeah, I echo some of Frederik's comments - my blog was generic to Cloudflare rather than R2 so some behavior might not be applicable.

kodster28:
Going to close this out as there hasn't been any activity in a year. Feel free to re-open if needed!

PR #8497, opened 2023-04-16
[cloudflare-docs] removed a error HTML tag sidebar.toc.html

RommHui:
Removed wrong </nav> HTML tag from sidebar.toc.html file.


kodster28 (approved):
Thanks for the catch, @RommHui

PR #8509, opened 2023-04-17
Update What is Email Address Obfuscation.md

kodster28:
Use code formatting so we aren't rendering html tags on the page


jedecf (approved):
Thanks Kody

PR #8515, opened 2023-04-17
[SUPPORT] migration of China Network content

jedecf:
PCX-6636


kodster28:
This one may actually be close enough to https://developers.cloudflare.com/china-network/concepts/icp/#displaying-your-icp-number that we could fold the content into the existing page (instead of creating a new one).

@pedrosousa, would that be something you could handle in a separate PR (maybe in the next maintenance sprint)? If so, I can handle the localized content deprecation / redirects?

content/_redirects
```diff
@@ -164,6 +164,9 @@
 /cache/best-practices/activate-polish/ /images/polish/ 301
 /cache/reference/common-cf-published-statuses/ https://support.cloudflare.com/hc/articles/4412244347917 301
 
+# china network
+/support/about-cloudflare/enterprise-documentation/understanding-and-configuring-an-icp-number/ 301
```
pedrosousa: Missing a target URL.

content/china-network/troubleshooting/Understanding and configuring an ICP number.md
pedrosousa, line 1: Not sure what happened to the `Should I choose Cloudflare or Yunjiasu?` page linked at the bottom of this page. It's not available in English in the Support tile; we only have translated versions. Should we delete these versions?

pedrosousa:
> This one may actually be close enough to https://developers.cloudflare.com/china-network/concepts/icp/#displaying-your-icp-number that we could fold the content into the existing page (instead of creating a new one).
> 
> @pedrosousa, would that be something you could handle in a separate PR (maybe in the next maintenance sprint)? If so, I can handle the localized content deprecation / redirects?

If we tackled your proposed changes in a different PR, would we keep the current PR? I could handle it directly in this PR, to avoid having an interim solution (and more redirects).

content/china-network/troubleshooting/Understanding and configuring an ICP number.md
kodster28: > > This one may actually be close enough to https://developers.cloudflare.com/china-network/concepts/icp/#displaying-your-icp-number that we could fold the content into the existing page (instead of creating a new one).
> > @pedrosousa, would that be something you could handle in a separate PR (maybe in the next maintenance sprint)? If so, I can handle the localized content deprecation / redirects?
> 
> If we tackled your proposed changes in a different PR, would we keep the current PR? I could handle it directly in this PR, to avoid having an interim solution (and more redirects).

Probably whatever is easiest for you, @pedrosousa.

kodster28: For now, probably just remove the link?

pedrosousa: Yes, that will work! I was just concerned we could have missed a page while doing the migration (from KB to Dev Docs).

kodster28: That's actually out of scope for the migration. I'll check in with the CSUP chat on this though.

kodster28: Tracking in #8534

pedrosousa (approved):

PR #8522, opened 2023-04-17
[Load Balancing] CNAME set up not supported note

patriciasantaana:
PCX-6613

### DIFF
--- a/content/load-balancing/how-to/create-load-balancer.md
+++ b/content/load-balancing/how-to/create-load-balancer.md
@@ -15,7 +15,7 @@ For more details about load balancers, refer to [Load balancers](/load-balancing
 
 {{<Aside type="note">}}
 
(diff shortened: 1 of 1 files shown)


content/load-balancing/how-to/create-load-balancer.md
```diff
@@ -13,6 +13,12 @@ meta:
 
 For more details about load balancers, refer to [Load balancers](/load-balancing/understand-basics/load-balancers/).
 
+{{<Aside type="note">}}
+
+This set up is only applicable if the target zone is on a full DNS set up. It is not supported if the target zone is on a `CNAME` set up.
```
ranbel:
```suggestion
Load balancing is only applicable if the target zone is on a full DNS setup. It is not supported if the target zone is on a `CNAME` setup.
```

ranbel (approved):

PR #8531, opened 2023-04-18
additional hostnames wildcard example missing

AlphaK5:
additional hostnames wildcard example missing


kodster28:
Actually, the wildcard on the cert is covered by the first record in the example.

PR #8540, opened 2023-04-18
[Docs site] Fixed an issue that the page can't get if the page path contains non-English characters breadcrumbs.html

RommHui:
The .RePermalink is a encoding URI. if the original filepath contains non-ASCII characters it will be encoding like "%XX%XX" and lead to can't get page correctly.

### DIFF
--- a/layouts/partials/breadcrumbs.html
+++ b/layouts/partials/breadcrumbs.html
@@ -16,7 +16,7 @@
     {{- range $index, $item := (split .RelPermalink "/") -}}
       {{- if gt (len $item ) 0 -}}
       
-        {{- $rellink = printf "%s/%s" $rellink $item -}}
+        {{- $rellink = printf "%s/%s" $rellink (urls.Parse $item).Path -}}
         {{- $page := $.Site.GetPage $rellink -}}
         
       {{- if eq ($page.RelPermalink | strings.Count "/" ) 2 -}}


kodster28:
So this does fix part of the issue, wish I'd thought of it myself!

We still aren't able to get the breadcrumbs working on the `/support/` content b/c there's some disconnect between their file names (that include spaces instead of dashes) that messes with the `.GetPage` functionality.

There's probably some way to get this in Hugo..... but probably the better approach is just gradually fixing the file names.

kodster28 (approved):

RommHui:
> So this does fix part of the issue, wish I'd thought of it myself!
> 
> We still aren't able to get the breadcrumbs working on the content b/c there's some disconnect between their file names (that include spaces instead of dashes) that messes with the functionality.`/support/``.GetPage`
> 
> There's probably some way to get this in Hugo..... but probably the better approach is just gradually fixing the file names.

I actually think the best way is to use Hugo's [.File](https://gohugo.io/variables/files/) to get the page path, but I'm worried that using it might affect something else that I don't know about.

RommHui:
> So this does fix part of the issue, wish I'd thought of it myself!
> 
> We still aren't able to get the breadcrumbs working on the content b/c there's some disconnect between their file names (that include spaces instead of dashes) that messes with the functionality.`/support/``.GetPage`
> 
> There's probably some way to get this in Hugo..... but probably the better approach is just gradually fixing the file names.

@kodster28 I actually think the best way is to use Hugo's [.File](https://gohugo.io/variables/files/) to get the page path, but I'm worried that using it might affect something else that I don't know about.

PR #8545, opened 2023-04-19
[Workers] Removed unclear wording about environment variables

isaac-mcfadyen:
- Before this change, on the Workers -> Platform -> Environment Variables page, there was a sentence `Regardless of type, bindings are always available as global variables within your Worker script`
- A user on the Discord was confused about this wording, mostly because this is only valid for the Service Workers format and the paragraph below it (about Module Workers) directly contradicts it.
- I removed the whole sentence, since it seemed unnecessary. It could also just be reworded, it just seemed out of place and a bit confusing if you're skimming the top paragraph.

### DIFF
--- a/content/workers/platform/environment-variables.md
+++ b/content/workers/platform/environment-variables.md
@@ -5,7 +5,7 @@ title: Environment variables
 
 # Environment variables
 
-In the Workers platform, environment variables, secrets, and KV namespaces are known as bindings. Regardless of type, bindings are always available as global variables within your Worker script.
+In the Workers platform, environment variables, secrets, and KV namespaces are known as bindings.
 
 ## Environment variables with module Workers
 


isaac-mcfadyen:
We were discussing this on the Discord a bit further and noticed a few other things:
1. There seems to be no section for Environment Variables in Service Workers at all, just Module Workers. Perhaps that should be added in place of this sentence?
2. There's another reference to "global variables" under the Wrangler section of that same page (`These environment variables can then be accessed within your Worker script as global variables`).
3. Every other type of binding has it's own page and the only one referenced on this page [seems to be KV](https://developers.cloudflare.com/workers/platform/environment-variables/#add-kv-namespaces-via-wrangler).

Link to a past PR that tried to clean up some of the "global variables" wording: https://github.com/cloudflare/cloudflare-docs/pull/7978

penalosa (approved):

deadlypants1973 (approved):

PR #8547, opened 2023-04-19
Small updates to the French 5xx KB

ngayerie:
Correcting typos and using better translations


jedecf (approved):
Pure translation fix :-)

PR #8573, opened 2023-04-21
workers: separate out integrations and databases

elithrar:
WIP / draft.

This PR separates out Integrations and Databases into their own top-level navigation items, so that they are more easily discovered by developers. These are both "critical path" items that we get a number of questions about.

image


elithrar:
Moving to a fresh PR.

PR #8585, opened 2023-04-23
Add note that user management requires Super Admin role

mchelen:


kodster28 (approved):

kodster28:
Thanks for the edit, @mchelen. Slightly updated your work to take advantage of a pre-existing snippet so we can re-use content.

PR #8599, opened 2023-04-24
Add algolia facet config for products

kristianfreeman:
This PR makes two changes that should improve our Algolia search results:

1) Adds back the "algolia" config for each product inside of the corresponding YAML file. This removes the index and API key values from being hardcoded into the script tags that our framework compiles.
2) Adds the `algolia:product` meta tag, which will set up a new `product` attribute on each page corresponding to the product that that page belongs to.
3) When a product is set, the search on that page will now filter directly on that product, meaning that the results will be more specific to the product you're viewing (aka search for "domain" on Pages docs, you should get results about "custom domains").

WIP while I do some testing


data/cache-reserve.yml
```diff
@@ -5,3 +5,8 @@ product:
   group: Application performance
   url: /cache/about/cache-reserve/
   grid_placeholder: true
+
+algolia:
```
kodster28: There are certain .yml files like this that don't corresponding to an actual `product` in our docs.

These are meant to show more options on the Products grid. You can identify them b/c the url references a subpage w/in a product folder.

kristianfreeman: oh sweet, makes sense. I'll remove those

kristianfreeman: This should be resolved now!

kristianfreeman:
Confirmed search still works - the filtering will come in a second phase PR

https://kristian-facets.cloudflare-docs-7ou.pages.dev/workers/

kodster28 (approved):

PR #8600, opened 2023-04-24
[Workers] Clarify when DO migrations are not needed

koeninger:
There's been recurring confusion about this in discord, try to clarify

### DIFF
--- a/content/workers/learning/using-durable-objects.md
+++ b/content/workers/learning/using-durable-objects.md
@@ -307,8 +307,14 @@ durable_objects.bindings = [
 
 ### Configuring Durable Object classes with migrations
 
-When you make changes to your list of Durable Objects classes, you must initiate a migration process. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
+You must initiate a migration process when you create a new Durable Object class, or rename, delete, or transfer an existing class. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
 
+{{<Aside type="note">}}
+
+Updating code for an existing Durable Object class does not require a migration, just [`wrangler publish`](/workers/wrangler/commands/). This is true even for changes to how code interacts with persistent storage. Because of [Global Uniqueness](/workers/learning/using-durable-objects/#global-uniqueness) you do not have to be concerned about old and new code interacting with the same storage simultaneously. However, it is your responsibility to ensure that new code is backwards compatible with existing stored data.
+
+{{</Aside>}}
+  
 The most common migration performed is a new class migration, which informs the system that a new Durable Object class is being uploaded.
 
 Migrations can also be used for transferring stored data between two Durable Object classes:
--- a/content/workers/learning/using-durable-objects.md
+++ b/content/workers/learning/using-durable-objects.md
@@ -307,7 +307,7 @@ durable_objects.bindings = [
 
 ### Configuring Durable Object classes with migrations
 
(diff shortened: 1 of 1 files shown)


content/workers/learning/using-durable-objects.md
```diff
@@ -307,8 +307,14 @@ durable_objects.bindings = [
 
 ### Configuring Durable Object classes with migrations
 
-When you make changes to your list of Durable Objects classes, you must initiate a migration process. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
+You must initiate a migration process when you create a new Durable Object class, or rename, delete, or transfer an existing class. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
```
jqmmes: This is way more detailed than previously, but I find it a bit confusing. Maybe the explanation on what a migration is should be before detailing when you actually need to use it

koeninger: Any suggestions on language you think would be clearer?

jqmmes: Maybe just simplyfing the second part is enough... What about:

You must initiate a migration process when you create a new Durable Object class, or rename, delete, or transfer an existing class. This process informs the Workers platform of the changes and provides it with instructions on how to deal with those changes.

Maddy-Cloudflare (approved):
PCX approved

jqmmes (approved):

content/workers/learning/using-durable-objects.md
```diff
@@ -307,8 +307,14 @@ durable_objects.bindings = [
 
 ### Configuring Durable Object classes with migrations
 
-When you make changes to your list of Durable Objects classes, you must initiate a migration process. A migration is informing the Workers platform of the changes and provide it with instructions on how to deal with those changes.
+You must initiate a migration process when you create a new Durable Object class, or rename, delete, or transfer an existing class. This process informs the Workers platform of the changes and provides it with instructions on how to deal with those changes.
```
deadlypants1973:
```suggestion
You must initiate a migration process when you create a new Durable Object class, or rename, delete, or transfer an existing Durable Objects class. This process informs the Workers runtime of the changes and provides it with instructions on how to deal with those changes.
```
@koeninger it was told to me that we should steer away from saying Workers platform (Developer platform = Workers as a product + other products) for architectural reason and also legal reasons (as in what does a platform need to have to be considered a platform). So changing this to 'runtime' instead, but possible change can also be 'ecosystem' if you prefer.

content/workers/learning/using-durable-objects.md
```diff
@@ -307,8 +307,14 @@ durable_objects.bindings = [
...
 
+{{<Aside type="note">}}
+
+Updating code for an existing Durable Object class does not require a migration, just [`wrangler publish`](/workers/wrangler/commands/). This is true even for changes to how code interacts with persistent storage. Because of [Global Uniqueness](/workers/learning/using-durable-objects/#global-uniqueness) you do not have to be concerned about old and new code interacting with the same storage simultaneously. However, it is your responsibility to ensure that new code is backwards compatible with existing stored data.
```
deadlypants1973:
```suggestion
Updating code for an existing Durable Object class does not require a migration. To update code for an existing Durable Object class, run [`wrangler publish`](/workers/wrangler/commands/). This is true even for changes to how code interacts with persistent storage. Because of [Global Uniqueness](/workers/learning/using-durable-objects/#global-uniqueness) you do not have to be concerned about old and new code interacting with the same storage simultaneously. However, it is your responsibility to ensure that new code is backwards compatible with existing stored data.
```

deadlypants1973 (approved):
PCX review

PR #8609, opened 2023-04-25
fix bug

Weltolk:


ranbel:
Hi, I see that `cloudflared-linux-amd64.deb` is a valid package on https://github.com/cloudflare/cloudflared/releases/. What error do you get when you run the example command?

Weltolk:
> Hi, I see that `cloudflared-linux-amd64` is a valid package on https://github.com/cloudflare/cloudflared/releases/. What error do you get when you run the example command?

sry, I made a mistake

PR #8613, opened 2023-04-25
recommend not to use Zaraz with Rocket Loader

bjesus:


content/zaraz/faq.md
```diff
@@ -38,6 +38,10 @@ If Zaraz is not loading, please verify the following:
 
 Yes. You can use the metrics in [Zaraz Monitoring](/zaraz/monitoring/) to help you find where in the workflow the problem occurred.
 
+#### Can I use Zaraz with Rocket Loader?
+
+We recommend disabling [Rocket Loader](https://developers.cloudflare.com/fundamentals/speed/rocket-loader/) when using Zaraz. While Zaraz can be used together with Rocket Loader, there's usually no need to use both. Rocket Loader can sometimes delay data from reaching Zaraz, causing issues.
```
kodster28:
```suggestion
We recommend disabling [Rocket Loader](/fundamentals/speed/rocket-loader/) when using Zaraz. While Zaraz can be used together with Rocket Loader, there's usually no need to use both. Rocket Loader can sometimes delay data from reaching Zaraz, causing issues.
```

kodster28 (approved):

PR #8660, opened 2023-04-28
need-logpush-job-update-for-new-field

kyouheicf:
In order to apply the following new security field settings to the existing Logpush jobs, you will have to edit Logpush option of jobs and make sure the new field is checked at **Select data fields**, then **Save changes**.

### DIFF
--- a/content/logs/reference/change-notices/2023-02-01-security-fields-updates.md
+++ b/content/logs/reference/change-notices/2023-02-01-security-fields-updates.md
@@ -64,7 +64,7 @@ For more information on these actions, refer to the [Actions](/ruleset-engine/ru
 
 ## HTTP Requests dataset changes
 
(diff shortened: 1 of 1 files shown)


tlozoot (changes requested):
A couple suggestions to make this more clear:
1) I think this should be lower on the page - it's confusing to have this at the top with no other context as to what this is about
2) This gives only instructions as to what to do _in the UI_ -- but many customers update their logpush jobs via API or Terraform. We should explain what needs to be done there.
3) The wording is a bit confusing. I might suggesting breaking up the description of what to do in the UI into a few sentences or a numbered list.

soheiokamoto (approved):

tlozoot (approved):
Would like @jsells1 to review this as well

angelampcosta (approved):

pedrosousa:
@jsells1 Can you please review the proposed documentation changes?

jsells1 (approved):

PR #8663, opened 2023-04-28
Update test-email-routing.md

cflcrowter:
Clarified the point about not testing from the same address and broke the paragraph up for readability.


kodster28 (approved):

kodster28:
Thanks @cflcrowter, actually ran into this issue the other day so this PR is very timely.

PR #8681, opened 2023-04-28
[SSL] Move validation level info closer to feature intro content

RebeccaTamachiro:
PCX-6660


content/ssl/reference/certificate-validation-options.md
RebeccaTamachiro, line 1: Have decided to remove this page as its content looked more like a note (moved into partials) and the Reference section of SSL docs tile is already a bit cluttered/page would not be very visible anyways. Check referred PCX ticket for traffic comparison.

marciocloudflare (approved):

PR #8686, opened 2023-05-01
Update reports.md

oyo-lab:
removed the note about customers with Leased IPs not getting reports. Since we migrated the reports to use NAv2 data, then customers with Leased IPs should also be receiving these reports.


pedrosousa (approved):

pedrosousa:
Note to self: related to PCX-6619.

PR #8728, opened 2023-05-04
[Pages] feat: Upgrade Next requirement to Node 16

alfaphi:
This commit updates the required Node.js version for Next.js to 16
Learn more: https://nextjs.org/docs/getting-started


alfaphi:
I have problem with deployment of NextJS to Cloudflare Pages
After searching my error in build log, another person also have the same problem:
https://stackoverflow.com/questions/76158141/unexpected-token-when-running-npm-run-dev-in-next-js-node-module

deadlypants1973 (approved):

PR #8771, opened 2023-05-08
[Worker] fix error code in web-crypto.md

jihuayu:
function `generateKey` parameter `length` type is number not string


deadlypants1973 (approved):
PCX review

PR #8795, opened 2023-05-09
[Workers] Runtime API: TCP Socket API Docs `connect()`

irvinebroque:
**Do not merge yet**

- [x] `/runtime-apis/connect`
- [x] `/platform/protocols` (Think this is worth adding — _"What protocols can I use with Cloudflare Workers?"._ Will cover HTTP, WebSockets, Outbound TCP, reference [Socket Workers blog post](https://blog.cloudflare.com/introducing-socket-workers/))
- [x] Address TODO comments
- [x] How do TCP connections count towards subrequest limits?
- [x] Example showing closing sockets, `allowHalfOpen` differences (https://github.com/cloudflare/cloudflare-docs/pull/8795/commits/fe93dd3a494f0c5cad1169a5d648d4dd924151f9)
- [x] Example showing error handling (https://github.com/cloudflare/cloudflare-docs/pull/8795/files#r1188052039)
- [x] Clarify that connections should not / cannot be created globally (https://github.com/cloudflare/cloudflare-docs/pull/8795/commits/a4f04859ac7f780e8e67cdf4ee4dec13adc73eba)
- [x] Add link to original blog post at top of `/platform/protocols` section, and cross-link to that from the `/runtime-apis/connect` docs page
- [x] Clarify local dev (`--experimental-local`)

### After 05/17/23

- [ ] Remove `--experimental-local` caveat

### DIFF
--- a/content/workers/platform/betas.md
+++ b/content/workers/platform/betas.md
@@ -18,6 +18,6 @@ These are the current alphas and betas relevant to the Cloudflare Workers platfo
 | Green Compute                 |               |              |  ✅          |[Blog](https://blog.cloudflare.com/earth-day-2022-green-compute-open-beta/) |
 | Pub/Sub                       |               | ✅           |              |[Docs](/pub-sub)                                                            |
 | Queues                        |               |              |  ✅          |[Docs](/queues)                                                             |
-| TCP Workers                   |               |            |     ✅         |[Docs](/workers/runtime-apis/connect)             |
+| TCP Workers                   |               |            |     ✅         |[Docs](/workers/runtime-apis/tcp-sockets)             |
 | Workers Analytics Engine      |               |             | ✅            |[Docs](/analytics/analytics-engine/)               |
 | Workers Deployments           |               |             | ✅            |[Docs](/workers/platform/deployments)               |
--- a/content/workers/platform/betas.md
+++ b/content/workers/platform/betas.md
@@ -18,6 +18,6 @@ These are the current alphas and betas relevant to the Cloudflare Workers platfo
 | Green Compute                 |               |              |  ✅          |[Blog](https://blog.cloudflare.com/earth-day-2022-green-compute-open-beta/) |
 | Pub/Sub                       |               | ✅           |              |[Docs](/pub-sub)                                                            |
 | Queues                        |               |              |  ✅          |[Docs](/queues)                                                             |
-| TCP Workers                   |               | ✅           |              |[Blog](https://blog.cloudflare.com/introducing-socket-workers/)             |
+| TCP Workers                   |               |            |     ✅         |[Blog](https://blog.cloudflare.com/introducing-socket-workers/)             |
 | Workers Analytics Engine      |               |             | ✅            |[Docs](/analytics/analytics-engine/)               |
 | Workers Deployments           |               |             | ✅            |[Docs](/workers/platform/deployments)               |
--- a/content/workers/platform/betas.md
+++ b/content/workers/platform/betas.md
@@ -18,6 +18,6 @@ These are the current alphas and betas relevant to the Cloudflare Workers platfo
 | Green Compute                 |               |              |  ✅          |[Blog](https://blog.cloudflare.com/earth-day-2022-green-compute-open-beta/) |
 | Pub/Sub                       |               | ✅           |              |[Docs](/pub-sub)                                                            |
 | Queues                        |               |              |  ✅          |[Docs](/queues)                                                             |
-| TCP Workers                   |               |            |     ✅         |[Blog](https://blog.cloudflare.com/introducing-socket-workers/)             |
+| TCP Workers                   |               |            |     ✅         |[Docs](/workers/runtime-apis/connect)             |
 | Workers Analytics Engine      |               |             | ✅            |[Docs](/analytics/analytics-engine/)               |
 | Workers Deployments           |               |             | ✅            |[Docs](/workers/platform/deployments)               |
--- a/content/workers/platform/limits.md
+++ b/content/workers/platform/limits.md
@@ -242,6 +242,7 @@ While handling a request, each Worker is allowed to have up to six connections o
 - `put()`, `match()`, and `delete()` methods of [Cache objects](/workers/runtime-apis/cache/).
 - `list()`, `get()`, `put()`, `delete()`, and `head()` methods of [R2](/r2/).
 - `send()` and `sendBatch()`, methods of [Queues](/queues/).
+- Opening a TCP socket using the [`connect()`](/workers/runtime-apis/connect/) API.
 
 Once a Worker has six connections open, it can still attempt to open additional connections. However, these attempts are put in a pending queue — the connections will not be initiated until one of the currently open connections has closed. Since earlier connections can delay later ones, if a Worker tries to make many simultaneous subrequests, its later subrequests may appear to take longer to start.
 
--- a/content/workers/platform/limits.md
+++ b/content/workers/platform/limits.md
@@ -242,7 +242,7 @@ While handling a request, each Worker is allowed to have up to six connections o
 - `put()`, `match()`, and `delete()` methods of [Cache objects](/workers/runtime-apis/cache/).
 - `list()`, `get()`, `put()`, `delete()`, and `head()` methods of [R2](/r2/).
 - `send()` and `sendBatch()`, methods of [Queues](/queues/).
-- Opening a TCP socket using the [`connect()`](/workers/runtime-apis/connect/) API.
+- Opening a TCP socket using the [`connect()`](/workers/runtime-apis/tcp-sockets/) API.
 
 Once a Worker has six connections open, it can still attempt to open additional connections. However, these attempts are put in a pending queue — the connections will not be initiated until one of the currently open connections has closed. Since earlier connections can delay later ones, if a Worker tries to make many simultaneous subrequests, its later subrequests may appear to take longer to start.
 
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -118,6 +118,31 @@ const socket = connect(address, { secureTransport: "starttls" });
 const secureSocket = socket.startTls();
 ```
 
+## Error handling
+
+- TODO - show what happens when the server closes the connection, and how to handle it gracefully.
+
+```typescript
+import { connect } from "cloudflare:sockets"
+
+const connectionUrl = "<URL>";
+
+export default {
+  async fetch(req, env) {
+
+    try {
+      const socket = connect(connectionUrl);
+
+      //...
+
+      return new Response(socket.readable, { headers: { "Content-Type": "text/plain" } });
+    } catch (error) {
+      return new Response("Socket connection failed: " + error, { status: 500 });
+    }
+  }
+};
+```
+
 Note the following:
 
 - `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -175,3 +175,4 @@ Note the following:
 
 - When developing locally with [Wrangler](/workers/wrangler/), you must pass the [`--experimental-local`](/workers/wrangler/commands/#dev) flag, instead of the `--local` flag, in order to use `connect()`.
 - TCP sockets must be created within the [`fetch()` handler](/workers/get-started/guide/#3-write-code) of a Worker. TCP sockets cannot be created in global scope and shared across requests. 
+- Each open TCP socket counts towards the maximum number of [open connections](/workers/platform/limits/#simultaneous-open-connections) that can be simultaneously open.
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -108,6 +108,11 @@ const socket = connect(address, { secureTransport: "starttls" });
 const secureSocket = socket.startTls();
 ```
 
+#### Note the following:
+
+- `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
+- Once `startTls()` is called, the initial socket is closed and no longer usable. In the example above, one would would use the newly created `secureSocket`.
+
 ## Error handling
 
 import { connect } from 'cloudflare:sockets';
@@ -164,12 +169,6 @@ socket.close();
 // After close() is called, you can no longer read from the readable side of the socket
 const reader = socket.readable.getReader(); // This fails
 ```
-
-
-Note the following:
-
-- `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
-- Once `startTls()` is called, the initial socket is closed and no longer usable. In the example above, one would would use the newly created `secureSocket`.
 
 ### Considerations
 
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -111,7 +111,8 @@ const secureSocket = socket.startTls();
 #### Note the following:
 
 - `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
-- Once `startTls()` is called, the initial socket is closed and no longer usable. In the example above, one would would use the newly created `secureSocket`.
+- Once `startTls()` is called, the initial socket is closed and can no longer be read from or written to. In the example above, anytime after `startTls()` is called, you would use the newly created `secureSocket`. Any existing readers and writers based off the original socket will no longer work — you must create new readers and writers from the newly created `secureSocket`.
+- `startTls()` should only be called once on an existing socket.
 
 ## Error handling
 
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -115,6 +115,8 @@ const secureSocket = socket.startTls();
 
 ## Error handling
 
+To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside `try..catch` blocks. This example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a 500:
+
 ```typescript
 import { connect } from 'cloudflare:sockets';
 const connectionUrl = "google.com:80";
@@ -130,13 +132,13 @@ export default {
       
       return new Response(socket.readable, { headers: { "Content-Type": "text/plain" } });
     } catch (error) {
-      return new Response("Socket connection failed: " + error, { status: 500 });
+      return new Response(`Socket connection failed: ${error}`, { status: 500 });
     }
   }
 };
 ```
 
-### Closing TCP connections
+## Closing TCP connections
 
 You can close a TCP connection by calling `close()` on the socket. This will close both the readable and writeable sides of the socket.
 
@@ -151,7 +153,7 @@ socket.close();
 const reader = socket.readable.getReader(); // This fails
 ```
 
-### Considerations
+## Considerations
 
 - When developing locally with [Wrangler](/workers/wrangler/), you must pass the [`--experimental-local`](/workers/wrangler/commands/#dev) flag, instead of the `--local` flag, in order to use `connect()`.
 - TCP sockets must be created within the [`fetch()` handler](/workers/get-started/guide/#3-write-code) of a Worker. TCP sockets cannot be created in global scope and shared across requests. 
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -68,7 +68,7 @@ export default {
 
 - `allowHalfOpen` {{<type>}}boolean{{</type>}} — Defaults to `false`
   - Defines whether the writable side of the TCP socket will automatically close on EOF. When set to `false`, the writable side of the TCP socket will automatically close on EOF. When set to `true`, the writable side of the TCP socket will remain open on EOF.
-  - This option allows for interoperability with TCP socket implementations in Node.js and Deno, which currently behave differently.
+  - This option is similar to that offered by the Node.js [`net` module](https://nodejs.org/api/net.html) and allows interoperability with code which utilises it.
 
 {{</definitions>}}
 
@@ -83,7 +83,7 @@ export default {
   - Returns the writable side of the TCP socket.
 
 - `closed()` {{<type>}}`Promise<void>`{{</type>}}
-  - TODO
+  - This promise is resolved when the socket is closed and is rejected if the socket encounters an error.
 
 - `close()` {{<type>}}`Promise<void>`{{</type>}}
   - Closes the TCP socket. If [`allowHalfOpen`](/workers/runtime-apis/connect/#socketoptions) is set to `true`, only the writeable side of the socket will be closed, and the readable side of the socket will remain readable.
@@ -102,9 +102,9 @@ export default {
 
 {{</definitions>}}
 
-## How to implement the startTLS pattern
+## How to implement Opportunistic TLS (StartTLS)
 
-Many TCP-based systems, including databases and email servers, require that clients use a pattern called StartTLS when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then "upgrades" it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
+Many TCP-based systems, including databases and email servers, require that clients use opportunistic TLS (otherwise known as [StartTLS](https://en.wikipedia.org/wiki/Opportunistic_TLS)) when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then "upgrades" it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
 
 ```typescript
 import { connect } from "cloudflare:sockets"
@@ -119,7 +119,24 @@ const secureSocket = socket.startTls();
 
 ## Error handling
 
-- TODO - show what happens when the server closes the connection, and how to handle it gracefully.
+import { connect } from 'cloudflare:sockets';
+
+const connectionUrl = "nonexistinghostname.invalid:1234";
+
+export interface Env { }
+
+export default {
+  async fetch(req: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
+    try {
+      const socket = connect(connectionUrl);
+
+      // This will throw because the `closed` promise is rejected with an exception.
+      await socket.closed;
+    } catch (error) {
+      return new Response("Socket connection failed: " + error, { status: 500 });
+    }
+  }
+};
 
 ```typescript
 import { connect } from "cloudflare:sockets"
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -115,38 +115,19 @@ const secureSocket = socket.startTls();
 
 ## Error handling
 
+```typescript
 import { connect } from 'cloudflare:sockets';
-
-const connectionUrl = "nonexistinghostname.invalid:1234";
-
+const connectionUrl = "google.com:80";
 export interface Env { }
-
 export default {
   async fetch(req: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
     try {
       const socket = connect(connectionUrl);
-
-      // This will throw because the `closed` promise is rejected with an exception.
-      await socket.closed;
-    } catch (error) {
-      return new Response("Socket connection failed: " + error, { status: 500 });
-    }
-  }
-};
-
-```typescript
-import { connect } from "cloudflare:sockets"
-
-const connectionUrl = "<URL>";
-
-export default {
-  async fetch(req, env) {
-
-    try {
-      const socket = connect(connectionUrl);
-
-      //...
-
+      const writer = socket.writable.getWriter();
+      const encoder = new TextEncoder();
+      const encoded = encoder.encode("GET / HTTP/1.0\r\n\r\n");
+      await writer.write(encoded);
+      
       return new Response(socket.readable, { headers: { "Content-Type": "text/plain" } });
     } catch (error) {
       return new Response("Socket connection failed: " + error, { status: 500 });
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -88,17 +88,8 @@ export default {
 - `close()` {{<type>}}`Promise<void>`{{</type>}}
   - Closes the TCP socket. Both the readable and writable streams are forcibly closed.
 
-- {{<code>}}startTls(options?: {{<prop-meta>}}optional{{</prop-meta>}} {{<type-link href="/workers/runtime-apis/connect/#tlsoptions">}}TlsOptions{{</type-link>}}){{</code>}} : {{<type-link href="/workers/runtime-apis/connect/#socket">}}Socket{{</type-link>}}
+- {{<code>}}startTls(){{</code>}} : {{<type-link href="/workers/runtime-apis/connect/#socket">}}Socket{{</type-link>}}
   - Upgrades an insecure socket to a secure one that uses TLS, returning a new [Socket](/workers/runtime-apis/connect#socket). Note that in order to call `startTls()`, you must set [`secureTransport`](/workers/runtime-apis/connect/#socketoptions) to `starttls` when initially calling `connect()` to create the socket.
-
-{{</definitions>}}
-
-### `TlsOptions`
-
-{{<definitions>}}
-
-- `expectedServerHostname` {{<type>}}string{{</type>}}
-  - TODO
 
 {{</definitions>}}
 
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -143,6 +143,34 @@ export default {
 };
 ```
 
+### Closing TCP connections
+
+You can close a TCP connection by calling `close()` on the socket. By default, unless the `allowHalfOpen` is explicitly set to `true`, this will close both the readable and writeable sides of the socket
+
+```typescript
+import { connect } from "cloudflare:sockets"
+
+const socket = connect("my-url.com:70");
+const reader = socket.readable.getReader();
+socket.close();
+
+// After close() is called, you can no longer read from the readable side of the socket
+const reader = socket.readable.getReader(); // This fails
+```
+
+If you need to create a socket that is compatible with existing code that assumes the pattern from Node.js, where the readable side of a socket remains open after `socket.end()` is called, you can set the `allowHalfOpen` option when creating a socket.
+
+```typescript
+import { connect } from "cloudflare:sockets"
+
+const socket = connect("my-url.com:70", { allowHalfOpen: true });
+const reader = socket.readable.getReader();
+socket.close();
+
+// After close() is called, you can continue to read from the readable side of the socket
+const reader = socket.readable.getReader(); // This works
+```
+
 Note the following:
 
 - `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -122,3 +122,7 @@ Note the following:
 
 - `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
 - Once `startTls()` is called, the initial socket is closed and no longer usable. In the example above, one would would use the newly created `secureSocket`.
+
+### Considerations
+
+- TCP sockets must be created within the [`fetch()` handler](/workers/get-started/guide/#3-write-code) of a Worker. TCP sockets cannot be created in global scope and shared across requests. 
--- a/content/workers/runtime-apis/connect.md
+++ b/content/workers/runtime-apis/connect.md
@@ -79,7 +79,7 @@ export default {
 - {{<code>}}readable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/readablestream/">}}ReadableStream{{</type-link>}}
   - Returns the readable side of the TCP socket.
 
-- {{<code>}}writeable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/writeable/">}}WriteableStream{{</type-link>}}
+- {{<code>}}writeable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/writablestream/">}}WriteableStream{{</type-link>}}
   - Returns the writable side of the TCP socket.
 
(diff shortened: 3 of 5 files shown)


content/workers/platform/betas.md
```diff
@@ -18,6 +18,6 @@ These are the current alphas and betas relevant to the Cloudflare Workers platfo
 | Green Compute                 |               |              |  ✅          |[Blog](https://blog.cloudflare.com/earth-day-2022-green-compute-open-beta/) |
 | Pub/Sub                       |               | ✅           |              |[Docs](/pub-sub)                                                            |
 | Queues                        |               |              |  ✅          |[Docs](/queues)                                                             |
-| TCP Workers                   |               | ✅           |              |[Blog](https://blog.cloudflare.com/introducing-socket-workers/)             |
+| TCP Workers                   |               |            |     ✅         |[Blog](https://blog.cloudflare.com/introducing-socket-workers/)             |
```
irvinebroque:
```suggestion
| TCP Workers                   |               |            |     ✅         |[Docs](/workers/runtime-apis/connect)             |
```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,124 @@
+---
+pcx_content_type: configuration
+title: Connect (TCP Socket API)
```
irvinebroque: This ends up being a bit long in the sidebar nav, but I think important since people otherwise don't know the `connect()` API by name yet.

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,124 @@
...
+{{<definitions>}}
+
+- {{<code>}}readable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/readablestream/">}}ReadableStream{{</type-link>}}
+  - Returns the readable side of the TCP socket.
+
+- {{<code>}}writeable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/writeable/">}}WriteableStream{{</type-link>}}
+  - Returns the writable side of the TCP socket.
+
+- `closed()` {{<type>}}`Promise<void>`{{</type>}}
+  - TODO
```
irvinebroque:
```suggestion
  - Used to check if the socket has been closed
```

@dom96 — is this true? (I actually wasn't immediately sure offhand how one might use this, imagine you have a clearer example than I do)

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,124 @@
...
+  - Upgrades an insecure socket to a secure one that uses TLS, returning a new [Socket](/workers/runtime-apis/connect#socket). Note that in order to call `startTls()`, you must set [`secureTransport`](/workers/runtime-apis/connect/#socketoptions) to `starttls` when initially calling `connect()` to create the socket.
+
+{{</definitions>}}
+
+### `TlsOptions`
+
+{{<definitions>}}
+
+- `expectedServerHostname` {{<type>}}string{{</type>}}
+  - TODO
```
irvinebroque: @dom96 — this is about [SNI](https://www.cloudflare.com/learning/ssl/what-is-sni/)? Can you drop in a brief explainer and I can edit?

https://github.com/cloudflare/workerd/blob/main/src/workerd/api/sockets.c%2B%2B#L210

content/workers/runtime-apis/connect.md
```diff
@@ -118,6 +118,31 @@ const socket = connect(address, { secureTransport: "starttls" });
 const secureSocket = socket.startTls();
 ```
 
+## Error handling
+
+- TODO - show what happens when the server closes the connection, and how to handle it gracefully.
```
irvinebroque: @dom96 — can you drop in an example?

content/workers/runtime-apis/connect.md
petebacondarwin: I would argue that for this section the title should not even have the word `connect`.
There are a number of other pages in this section where the title is more descriptive - e.g. "Durable Objects", "Service bindings", "Node compatibility".
So instead, I think this title could just be "TCP sockets" or similar. Then in the page itself introduce the actual function call that gets you a socket (i.e. `connect()`).

content/workers/runtime-apis/connect.md
petebacondarwin: Rather than "check" if the socket is closed, I would say this is a way to run some clean up "when" the socket is closed. The promise will only be fulfilled (e.g. resolved) once the socket closes. It is not something you can poll.

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+{{<definitions>}}
+
+- `expectedServerHostname` {{<type>}}string{{</type>}}
+  - TODO
+
+{{</definitions>}}
+
+## How to implement the startTLS pattern
+
+Many TCP-based systems, including databases and email servers, require that clients use a pattern called StartTLS when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then "upgrades" it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
```
petebacondarwin: I think it would be good to describe the behaviour of the old unsecured socket in this case. It was something I had to work out be trial and error. For example:
- the old socket gets closed when it is upgraded (and so resolves the `closed` promise), which I had to keep state for to avoid passing on a close event to the client of the socket.
- The new socket is an entirely new socket, so any readers and writers that were being used need to be recreated from the new socket.
- Also the developer should try to avoid calling `startTls()` multiple times on the same socket.

content/workers/runtime-apis/connect.md
dom96: Agreed with @petebacondarwin

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+      const encoder = new TextEncoder();
+      const encoded = encoder.encode(url.pathname + "\r\n");
+      await writer.write(encoded);
+
+      return new Response(socket.readable, { headers: { "Content-Type": "text/plain" } });
+    } catch (error) {
+      return new Response("Socket connection failed: " + error, { status: 500 });
+    }
+  }
+};
```
dom96: I think it would be nice if this example could be copy&pasted and work without any modifications. How about this simple example? (I tested it and it works, I also added typescript types which I think all our docs should include)

```suggestion
import { connect } from 'cloudflare:sockets';

const connectionUrl = "google.com:80";

export interface Env { }

export default {
  async fetch(req: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
    try {
      const socket = connect(connectionUrl);
      const writer = socket.writable.getWriter();
      const encoder = new TextEncoder();
      const encoded = encoder.encode("GET / HTTP/1.0\r\n\r\n");
      await writer.write(encoded);
      
      return new Response(socket.readable, { headers: { "Content-Type": "text/plain" } });
    } catch (error) {
      return new Response("Socket connection failed: " + error, { status: 500 });
    }
  }
};
```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+
+- `secureTransport` {{<type>}}"off" | "on" | "starttls"{{</type>}} — Defaults to `off`
+  - Specifies whether or not to use [TLS](https://www.cloudflare.com/learning/ssl/transport-layer-security-tls/) when creating the TCP socket.
+  - `off` — do not use TLS.
+  - `on` — use TLS.
+  - `starttls` — do not use TLS initially, but allow the socket to be upgraded to use TLS by calling [`startTls()`](/workers/runtime-apis/connect/#how-to-implement-the-starttls-pattern).
+
+- `allowHalfOpen` {{<type>}}boolean{{</type>}} — Defaults to `false`
+  - Defines whether the writable side of the TCP socket will automatically close on EOF. When set to `false`, the writable side of the TCP socket will automatically close on EOF. When set to `true`, the writable side of the TCP socket will remain open on EOF.
+  - This option allows for interoperability with TCP socket implementations in Node.js and Deno, which currently behave differently.
```
dom96: Hm, I'm not super familiar with Deno but afaik this option isn't about interoperability with it. How about this:

```suggestion
  - This option is similar to that offered by the Node.js [`net` module](https://nodejs.org/api/net.html) and allows interoperability with code which utilises it.
```

CC @jasnell

content/workers/runtime-apis/connect.md
dom96: Yeah. It also gets rejected when the socket encounters an error, so you can for example detect if the connection was unsuccessful using it. How about something like:

```suggestion
  - This promise is resolved when the socket is closed and is rejected if the socket encounters an error.
```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+  - Returns the readable side of the TCP socket.
+
+- {{<code>}}writeable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/writeable/">}}WriteableStream{{</type-link>}}
+  - Returns the writable side of the TCP socket.
+
+- `closed()` {{<type>}}`Promise<void>`{{</type>}}
+  - TODO
+
+- `close()` {{<type>}}`Promise<void>`{{</type>}}
+  - Closes the TCP socket. If [`allowHalfOpen`](/workers/runtime-apis/connect/#socketoptions) is set to `true`, only the writeable side of the socket will be closed, and the readable side of the socket will remain readable.
```
dom96: The `allowHalfOpen` option has no effect on this. 

```suggestion
  - Closes the TCP socket. Both the readable and writable streams are forcibly closed.
```

content/workers/runtime-apis/connect.md
dom96: This is not currently supported in edge, just in workerd. So I would just remove its mention and note that we will be adding TLS options in the future.

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+### `TlsOptions`
+
+{{<definitions>}}
+
+- `expectedServerHostname` {{<type>}}string{{</type>}}
+  - TODO
+
+{{</definitions>}}
+
+## How to implement the startTLS pattern
```
dom96:
```suggestion
## How to implement Opportunistic TLS (StartTLS)
```

content/workers/runtime-apis/connect.md
dom96: I know that "startTls" caused a bunch of confusion and "Opportunistic TLS" is what Wikipedia calls it so I think we should use that term primarily:

```suggestion
Many TCP-based systems, including databases and email servers, require that clients use opportunistic TLS (otherwise known as [StartTLS](https://en.wikipedia.org/wiki/Opportunistic_TLS)) when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then "upgrades" it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+
+Many TCP-based systems, including databases and email servers, require that clients use a pattern called StartTLS when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then "upgrades" it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
+
+```typescript
```
dom96: Again, it would be nice if this example could work as-is. Is there an example publicly accessibly postgres DB we could connect to?

content/workers/runtime-apis/connect.md
dom96: We already have an example of the server closing the connection in the first example ("google.com") that I posted above.

I think what would be valuable is to show how to handle connection errors, i.e. the server rejecting the connection. Here is an example that is runnable:

```suggestion
import { connect } from 'cloudflare:sockets';

const connectionUrl = "nonexistinghostname.invalid:1234";

export interface Env { }

export default {
  async fetch(req: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
    try {
      const socket = connect(connectionUrl);

      // This will throw because the `closed` promise is rejected with an exception.
      await socket.closed;
    } catch (error) {
      return new Response("Socket connection failed: " + error, { status: 500 });
    }
  }
};
```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+    } catch (error) {
+      return new Response("Socket connection failed: " + error, { status: 500 });
+    }
+  }
+};
+```
+
+### Closing TCP connections
+
+You can close a TCP connection by calling `close()` on the socket. By default, unless the `allowHalfOpen` is explicitly set to `true`, this will close both the readable and writeable sides of the socket
```
dom96:
```suggestion
You can close a TCP connection by calling `close()` on the socket. This will close both the readable and writeable sides of the socket.
```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+```typescript
+import { connect } from "cloudflare:sockets"
+
+const socket = connect("my-url.com:70", { allowHalfOpen: true });
+const reader = socket.readable.getReader();
+socket.close();
+
+// After close() is called, you can continue to read from the readable side of the socket
+const reader = socket.readable.getReader(); // This works
+```
```
dom96: The `allowHalfOpen` option only affects the implicit closure semantics.

```suggestion

```

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
...
+
+Note the following:
+
+- `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
+- Once `startTls()` is called, the initial socket is closed and no longer usable. In the example above, one would would use the newly created `secureSocket`.
```
dom96: This feels a bit out of place, we already explain this above as well, maybe we can just remove this?

dom96 (approved):
Thanks for putting this together! There are some things that need to be changed but in general looking good.

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,181 @@
+---
+pcx_content_type: configuration
+title: Connect (TCP Socket API)
+weight: 4
+---
+
+# connect()
+
+The `connect()` API provides a programming interface for creating an outbound [TCP connection](https://www.cloudflare.com/learning/ddos/glossary/tcp-ip/) from a Cloudflare Worker. It returns a TCP socket, with both a [readable](/workers/runtime-apis/streams/readablestream/) and [writable](/workers/runtime-apis/streams/writablestream/) stream of data, allowing you to read and write data on an ongoing basis, as long as the connection remains open.
```
irvinebroque:
```suggestion
The Workers runtime provides an API, `connect()`, for creating a outbound [TCP connections](https://www.cloudflare.com/learning/ddos/glossary/tcp-ip/) from Workers. It returns a TCP socket, with both a [readable](/workers/runtime-apis/streams/readablestream/) and [writable](/workers/runtime-apis/streams/writablestream/) stream of data, allowing you to read and write data on an ongoing basis, as long as the connection remains open.
```

content/workers/runtime-apis/connect.md
irvinebroque: Thinking back to the Node.js example you'd shared awhile back:

```javascript
socket.end(); // Same as writer.close() in Deno.
 
// The socket's writable side has been closed,  so now
// it is in a half-closed state. The data sent by the
// server will still be readable.
```

I gathered from this that the `allowHalfOpen` option exists to mirror this behavior? Where one can call `close()` or `end()` and still have the socket remain "half open", with the readable side still able to be read?

If not, then why does `allowHalfOpen` exist?

content/workers/runtime-apis/connect.md
irvinebroque: Sorting out details on that

content/workers/runtime-apis/connect.md
irvinebroque: Ah I see: https://github.com/cloudflare/cloudflare-docs/pull/8795/files#r1188560682

content/workers/runtime-apis/connect.md
irvinebroque: https://github.com/cloudflare/cloudflare-docs/pull/8795/commits/40a6dd10d3ba69787d78ac6f4a5868899c738b51

irvinebroque:
@deadlypants1973 — I'll add the `/platform/protocols` section shortly, but the rest of this is ready for your 👀

content/workers/runtime-apis/connect.md
```diff
@@ -0,0 +1,178 @@
...
+{{</definitions>}}
+
+### `Socket`
+
+{{<definitions>}}
+
+- {{<code>}}readable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/readablestream/">}}ReadableStream{{</type-link>}}
+  - Returns the readable side of the TCP socket.
+
+- {{<code>}}writeable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/writeable/">}}WriteableStream{{</type-link>}}
```
irvinebroque:
```suggestion
- {{<code>}}writeable(){{</code>}} : {{<type-link href="/workers/runtime-apis/streams/writablestream/">}}WriteableStream{{</type-link>}}
```

content/workers/runtime-apis/connect.md
irvinebroque: Moved further up where it was supposed to be: https://github.com/cloudflare/cloudflare-docs/pull/8795/commits/a01a57929b8544074b7d6c912acefaa9c197322f

content/workers/runtime-apis/connect.md
irvinebroque: Would be nice but don't think there's anything usable here — would get hammered with connections.

content/workers/runtime-apis/connect.md
irvinebroque: Updated

content/workers/runtime-apis/connect.md
irvinebroque: https://github.com/cloudflare/cloudflare-docs/pull/8795/commits/b4086f8172fe5ecce8ade5e0ac8733f031c9816f

content/workers/runtime-apis/connect.md
irvinebroque: https://github.com/cloudflare/cloudflare-docs/pull/8795/commits/0f42b61060219d23a5889e5253e1222a7b6e86a5

content/workers/platform/limits.md
```diff
@@ -242,6 +242,7 @@ While handling a request, each Worker is allowed to have up to six connections o
 - `put()`, `match()`, and `delete()` methods of [Cache objects](/workers/runtime-apis/cache/).
 - `list()`, `get()`, `put()`, `delete()`, and `head()` methods of [R2](/r2/).
 - `send()` and `sendBatch()`, methods of [Queues](/queues/).
+- Opening a TCP socket using the [`connect()`](/workers/runtime-apis/tcp-sockets/) API.
```
deadlypants1973, line 245: Small inconsistency, but could we also use this PR to capitalize the 'the' in line 240? So it follows the correct capitalization format introduced by this sentence.

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
+---
+pcx_content_type: configuration
+title: TCP Sockets
```
deadlypants1973:
```suggestion
title: TCP sockets
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+weight: 4
```
deadlypants1973:
```suggestion
```
Not necessary as the runtime API pages are following an alphabetical listing format? If `weight` exists on any other individual pages, it's likely having no effect.

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+---
+
+# TCP Sockets
+
+The Workers runtime provides the `connect()` API for creating a outbound [TCP connections](https://www.cloudflare.com/learning/ddos/glossary/tcp-ip/) from Workers.
+
+Many application-layer protocols are built on top of TCP, and require an underlying TCP socket API in order to work, including SSH, MQTT, SMTP, FTP, IRC, and most database wire protocols including MySQL, PostgreSQL, MongoDB and more.
```
deadlypants1973:
```suggestion
Many application-layer protocols are built on top of the Transmission Control Protocol (TCP). These application-layer protocols, including SSH, MQTT, SMTP, FTP, IRC, and most database wire protocols including MySQL, PostgreSQL, MongoDB, require an underlying TCP socket API in order to work.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
+---
+pcx_content_type: configuration
+title: TCP Sockets
+weight: 4
+---
+
+# TCP Sockets
```
deadlypants1973:
```suggestion
# TCP sockets
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+
+The Workers runtime provides the `connect()` API for creating a outbound [TCP connections](https://www.cloudflare.com/learning/ddos/glossary/tcp-ip/) from Workers.
```
deadlypants1973:
```suggestion
The Workers runtime provides the `connect()` API for creating outbound [TCP connections](https://www.cloudflare.com/learning/ddos/glossary/tcp-ip/) from Workers.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+
+Many application-layer protocols are built on top of TCP, and require an underlying TCP socket API in order to work, including SSH, MQTT, SMTP, FTP, IRC, and most database wire protocols including MySQL, PostgreSQL, MongoDB and more.
+
+## `connect()`
+
+The `connect()` function returns a TCP socket, with both a [readable](/workers/runtime-apis/streams/readablestream/) and [writable](/workers/runtime-apis/streams/writablestream/) stream of data, allowing you to read and write data on an ongoing basis, as long as the connection remains open.
+
+`connect()` is provided as a [Runtime API](/workers/runtime-apis/), and is accessed by importing the `connect` function from `cloudflare:sockets`, akin to how one imports built-in modules in Node.js. A simple example of creating a TCP socket, writing to it, and returning the readable side of the socket as a response, is shown below:
```
deadlypants1973:
```suggestion
`connect()` is provided as a [Runtime API](/workers/runtime-apis/), and is accessed by importing the `connect` function from `cloudflare:sockets`. This process is similar to how one imports built-in modules in Node.js. Refer to the following codeblock for an example of creating a TCP socket, writing to it, and returning the readable side of the socket as a response:
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+
+# TCP Sockets
+
+The Workers runtime provides the `connect()` API for creating a outbound [TCP connections](https://www.cloudflare.com/learning/ddos/glossary/tcp-ip/) from Workers.
+
+Many application-layer protocols are built on top of TCP, and require an underlying TCP socket API in order to work, including SSH, MQTT, SMTP, FTP, IRC, and most database wire protocols including MySQL, PostgreSQL, MongoDB and more.
+
+## `connect()`
+
+The `connect()` function returns a TCP socket, with both a [readable](/workers/runtime-apis/streams/readablestream/) and [writable](/workers/runtime-apis/streams/writablestream/) stream of data, allowing you to read and write data on an ongoing basis, as long as the connection remains open.
```
deadlypants1973:
```suggestion
The `connect()` function returns a TCP socket, with both a [readable](/workers/runtime-apis/streams/readablestream/) and [writable](/workers/runtime-apis/streams/writablestream/) stream of data. This allows you to read and write data on an ongoing basis, as long as the connection remains open.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+
+### `SocketAddress`
+
+{{<definitions>}}
+
+- `hostname` {{<type>}}string{{</type>}}
+  - The hostname to connect to. Ex: `cloudflare.com`.
+
+- `port` {{<type>}}number{{</type>}}
+  - The port number to connect to. Ex: `5432`.
```
deadlypants1973:
```suggestion
- `hostname` {{<type>}}string{{</type>}}
  - The hostname to connect to. Example: `cloudflare.com`.

- `port` {{<type>}}number{{</type>}}
  - The port number to connect to. Example: `5432`.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+
+### `SocketOptions`
+
+{{<definitions>}}
+
+- `secureTransport` {{<type>}}"off" | "on" | "starttls"{{</type>}} — Defaults to `off`
+  - Specifies whether or not to use [TLS](https://www.cloudflare.com/learning/ssl/transport-layer-security-tls/) when creating the TCP socket.
+  - `off` — do not use TLS.
+  - `on` — use TLS.
+  - `starttls` — do not use TLS initially, but allow the socket to be upgraded to use TLS by calling [`startTls()`](/workers/runtime-apis/tcp-sockets/#how-to-implement-the-starttls-pattern).
```
deadlypants1973:
```suggestion
  - `off` — Do not use TLS.
  - `on` — Use TLS.
  - `starttls` — Do not use TLS initially, but allow the socket to be upgraded to use TLS by calling [`startTls()`](/workers/runtime-apis/tcp-sockets/#how-to-implement-the-starttls-pattern).
```

content/workers/runtime-apis/tcp-sockets.md
deadlypants1973: Removing the passive 'is shown below' to imperative 'Refer to \<this thing>'

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+
+- `allowHalfOpen` {{<type>}}boolean{{</type>}} — Defaults to `false`
+  - Defines whether the writable side of the TCP socket will automatically close on EOF. When set to `false`, the writable side of the TCP socket will automatically close on EOF. When set to `true`, the writable side of the TCP socket will remain open on EOF.
+  - This option is similar to that offered by the Node.js [`net` module](https://nodejs.org/api/net.html) and allows interoperability with code which utilises it.
```
deadlypants1973:
```suggestion
  - This option is similar to that offered by the Node.js [`net` module](https://nodejs.org/api/net.html) and allows interoperability with code which utilizes it.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+{{<definitions>}}
+
+- `secureTransport` {{<type>}}"off" | "on" | "starttls"{{</type>}} — Defaults to `off`
+  - Specifies whether or not to use [TLS](https://www.cloudflare.com/learning/ssl/transport-layer-security-tls/) when creating the TCP socket.
+  - `off` — do not use TLS.
+  - `on` — use TLS.
+  - `starttls` — do not use TLS initially, but allow the socket to be upgraded to use TLS by calling [`startTls()`](/workers/runtime-apis/tcp-sockets/#how-to-implement-the-starttls-pattern).
+
+- `allowHalfOpen` {{<type>}}boolean{{</type>}} — Defaults to `false`
+  - Defines whether the writable side of the TCP socket will automatically close on EOF. When set to `false`, the writable side of the TCP socket will automatically close on EOF. When set to `true`, the writable side of the TCP socket will remain open on EOF.
```
deadlypants1973:
```suggestion
  - Defines whether the writable side of the TCP socket will automatically close on end-of-file (EOF). When set to `false`, the writable side of the TCP socket will automatically close on EOF. When set to `true`, the writable side of the TCP socket will remain open on EOF.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+  - Closes the TCP socket. Both the readable and writable streams are forcibly closed.
+
+- {{<code>}}startTls(){{</code>}} : {{<type-link href="/workers/runtime-apis/tcp-sockets/#socket">}}Socket{{</type-link>}}
+  - Upgrades an insecure socket to a secure one that uses TLS, returning a new [Socket](/workers/runtime-apis/tcp-sockets#socket). Note that in order to call `startTls()`, you must set [`secureTransport`](/workers/runtime-apis/tcp-sockets/#socketoptions) to `starttls` when initially calling `connect()` to create the socket.
+
+{{</definitions>}}
+
+## Opportunistic TLS (StartTLS)
+
+Many TCP-based systems, including databases and email servers, require that clients use opportunistic TLS (otherwise known as [StartTLS](https://en.wikipedia.org/wiki/Opportunistic_TLS)) when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then "upgrades" it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
```
deadlypants1973:
```suggestion
Many TCP-based systems, including databases and email servers, require that clients use opportunistic TLS (otherwise known as [StartTLS](https://en.wikipedia.org/wiki/Opportunistic_TLS)) when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then transitions it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
```

Per style guide, we should avoid saying things "" and say exactly what we mean. This is especially necessary for international readers, as the concept of "" could be difficult to translate. I understood that `upgrades` may not have the right nuance, so changed to `transitions` instead.

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+const address = {
+  hostname: "example-postgres-db.com",
+  port: 5432
+};
+const socket = connect(address, { secureTransport: "starttls" });
+const secureSocket = socket.startTls();
+```
+
+- `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
+- Once `startTls()` is called, the initial socket is closed and can no longer be read from or written to. In the example above, anytime after `startTls()` is called, you would use the newly created `secureSocket`. Any existing readers and writers based off the original socket will no longer work — you must create new readers and writers from the newly created `secureSocket`.
```
deadlypants1973:
```suggestion
- Once `startTls()` is called, the initial socket is closed and can no longer be read from or written to. In the example above, anytime after `startTls()` is called, you would use the newly created `secureSocket`. Any existing readers and writers based off the original socket will no longer work. You must create new readers and writers from the newly created `secureSocket`.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+- `startTls()` should only be called once on an existing socket.
+
+## Error handling
+
+To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside `try..catch` blocks. This example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a 500:
```
deadlypants1973:
```suggestion
To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside `try..catch` blocks. The following example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a `500` response:
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+      
+      return new Response(socket.readable, { headers: { "Content-Type": "text/plain" } });
+    } catch (error) {
+      return new Response(`Socket connection failed: ${error}`, { status: 500 });
+    }
+  }
+};
+```
+
+## Closing TCP connections
```
deadlypants1973:
```suggestion
## Close TCP connections
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+};
+const socket = connect(address, { secureTransport: "starttls" });
+const secureSocket = socket.startTls();
+```
+
+- `startTls()` can only be called if `secureTransport` is set to `starttls` when creating the initial TCP socket.
+- Once `startTls()` is called, the initial socket is closed and can no longer be read from or written to. In the example above, anytime after `startTls()` is called, you would use the newly created `secureSocket`. Any existing readers and writers based off the original socket will no longer work — you must create new readers and writers from the newly created `secureSocket`.
+- `startTls()` should only be called once on an existing socket.
+
+## Error handling
```
deadlypants1973:
```suggestion
## Handle errors
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+const reader = socket.readable.getReader();
+socket.close();
+
+// After close() is called, you can no longer read from the readable side of the socket
+const reader = socket.readable.getReader(); // This fails
+```
+
+## Considerations
+
+- When developing locally with [Wrangler](/workers/wrangler/), you must pass the [`--experimental-local`](/workers/wrangler/commands/#dev) flag, instead of the `--local` flag, in order to use `connect()`.
```
deadlypants1973:
```suggestion
- When developing locally with [Wrangler](/workers/wrangler/), you must pass the [`--experimental-local`](/workers/wrangler/commands/#dev) flag, instead of the `--local` flag, to use `connect()`.
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+- TCP sockets must be created within the [`fetch()` handler](/workers/get-started/guide/#3-write-code) of a Worker. TCP sockets cannot be created in global scope and shared across requests. 
+- Each open TCP socket counts towards the maximum number of [open connections](/workers/platform/limits/#simultaneous-open-connections) that can be simultaneously open.
+- By default, Workers cannot create outbound TCP connections on port 25 to send email to SMTP mail servers. [Cloudflare Email Workers](/email-routing/email-workers/) provides APIs to process and forward email.
```
deadlypants1973:
```suggestion
- By default, Workers cannot create outbound TCP connections on port `25` to send email to SMTP mail servers. [Cloudflare Email Workers](/email-routing/email-workers/) provides APIs to process and forward email.
```

deadlypants1973 (approved):
PCX review

content/workers/runtime-apis/tcp-sockets.md
dom96: hm, should this suggested change include the removal of the heading + first paragraph?

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+The `connect()` function returns a TCP socket, with both a [readable](/workers/runtime-apis/streams/readablestream/) and [writable](/workers/runtime-apis/streams/writablestream/) stream of data, allowing you to read and write data on an ongoing basis, as long as the connection remains open.
+
+`connect()` is provided as a [Runtime API](/workers/runtime-apis/), and is accessed by importing the `connect` function from `cloudflare:sockets`, akin to how one imports built-in modules in Node.js. A simple example of creating a TCP socket, writing to it, and returning the readable side of the socket as a response, is shown below:
+
+```typescript
+import { connect } from 'cloudflare:sockets';
+
+export default {
+  async fetch(req: Request) {
+    const gopherAddr = "gopher.floodgap.com:70";
```
dom96: Nitpicking now, but I figure we may want to be consistent and use the object form of the address everywhere.

```suggestion
    const gopherAddr = { hostname: "gopher.floodgap.com", port: 70 };
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+- Once `startTls()` is called, the initial socket is closed and can no longer be read from or written to. In the example above, anytime after `startTls()` is called, you would use the newly created `secureSocket`. Any existing readers and writers based off the original socket will no longer work — you must create new readers and writers from the newly created `secureSocket`.
+- `startTls()` should only be called once on an existing socket.
+
+## Error handling
+
+To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside `try..catch` blocks. This example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a 500:
+
+```typescript
+import { connect } from 'cloudflare:sockets';
+const connectionUrl = "google.com:80";
```
dom96:
```suggestion
const connectionUrl = { hostname: "google.com", port: 80 };
```

content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+## Closing TCP connections
+
+You can close a TCP connection by calling `close()` on the socket. This will close both the readable and writeable sides of the socket.
+
+```typescript
+import { connect } from "cloudflare:sockets"
+
+const socket = connect("my-url.com:70");
```
dom96:
```suggestion
const socket = connect({ hostname: "my-url.com", port: 70 });
```

elithrar (approved):

content/workers/runtime-apis/tcp-sockets.md
elithrar: No

elithrar:
```suggestion
Many application-layer protocols are built on top of the Transmission Control Protocol (TCP). These application-layer protocols, including SSH, MQTT, SMTP, FTP, IRC, and most database wire protocols including MySQL, PostgreSQL, MongoDB, require an underlying TCP socket API in order to work.
```

content/workers/runtime-apis/tcp-sockets.md
elithrar: `upgrades` is the right word here /  used widely. Removing quotes though.

elithrar:
```suggestion
Many TCP-based systems, including databases and email servers, require that clients use opportunistic TLS (otherwise known as [StartTLS](https://en.wikipedia.org/wiki/Opportunistic_TLS)) when connecting. In this pattern, the client first creates an insecure TCP socket, without TLS, and then upgrades it to a secure TCP socket, that uses TLS. The `connect()` API simplifies this by providing a method, `startTls()`, which returns a new `Socket` instance that uses TLS:
```

elithrar:
Committed suggestions for @irvinebroque so we can merge ahead of #8914 and before 9AM ET :)

PR #8807, opened 2023-05-09
Update identity-selectors.md

kennyj42:
the current version was really incorrect. You can't reference access groups in the Gateway policy builder.


ranbel:
The point we were trying to make is that you have to go into Access to figure out which Gateway selector to use for a particular IdP: group names, group ID, or SAML. It's not always going to be SAML.

We can improve the wording so people don't try to use Access groups in Gateway.

kennyj42:
If you're using SAML, the only option you have is SAML Attributes. The Group Name/ID fields will not work for SAML idps since we don't know what the group value is called in a custom SAML integration.

ranbel:
How about for OIDC IdPs?

kennyj42:
They're not supported in Gateway policies until we add OIDC Claims as a selector.

ranbel:
So in summary:
- User Group IDs is only available with AzureAD.
- User Group Names is only available with GitHub, Google, or Okta (OIDC). No other OIDC providers are supported.
- SAML Attributes is available for all SAML providers.
- User Group Emails --> which IdPs are supported?

kennyj42:
AzureAD is also supported via Group Name if they enable SCIM.

I believe User Group Email should be supported for all IdPs.

ranbel (approved):

PR #8820, opened 2023-05-10
[Spectrum] Document that all Spectrum ports are open

angelampcosta:


marciocloudflare (approved):

Achiel (commented):
Think we can be a tiny bit more explicit. How about

"By default, Spectrum is configured to listen on all ports, which can raise concerns for security auditors. However, it is important to note that Spectrum will only proxy connections from edge ports that are specifically configured within Cloudflare. 

When a TCP handshake is initiated to any port for a Spectrum IP, the handshake will always be completed. If there is a Spectrum application configured for the port, the connection will be proxied to origin. If no application is configured, the connection is immediately terminated and no origin connection will be opened.

Spectrum will only ever proxy traffic to an origin if there is a Spectrum application configured for that port."

angelampcosta:
@Achiel should merge it?

PR #8828, opened 2023-05-10
[Rules] Update incoming expression match in _redirect-root-to-subdomain.md

### LINKED ISSUE #8827: Redirect Rule Results in Infinite Loop

Rules

- https://developers.cloudflare.com/fundamentals/get-started/basic-tasks/manage-subdomains/#redirect-root-domain-to-a-subdomain

[Redirect root domain to a subdomain](https://developers.cloudflare.com/fundamentals/get-started/basic-tasks/manage-subdomains/#redirect-root-domain-to-a-subdomain)

### What needs to change?

The section of the documentation aims to setup a redirect rule from root domain to subdomain. However, the example expression given to match incoming requests leads to an infinite redirect loop.

The expression matches for "example.com" in the `http.request.full_uri`. The problem here is that if a user attempts to go to the subdomain such as `www.example.com`, this rule  will match that traffic and continuously redirect itself till the user receives and error about there being too many redirects.

### How should it change?

**When incoming requests match**

Using the Expression Editor:<br>
`(lower(http.host) eq "example.com")`

### Additional information

I already have a fix ready for this as I created a similar rule for one of our domains.

haelmj:
Fixes https://github.com/cloudflare/cloudflare-docs/issues/8827:

- altered step 3 to reflect change from root to subdomain
- modified Incoming expression to avoid infinite redirect loop

### DIFF
--- a/content/fundamentals/_partials/_redirect-root-to-subdomain.md
+++ b/content/fundamentals/_partials/_redirect-root-to-subdomain.md
@@ -15,14 +15,14 @@ Sometimes, you might want all traffic to your root domain (`example.com`) to act
     | -------- | -------- | ---------------- | ---------------- |
     | A        | `@`      | `192.0.2.1`      | Proxied          |
 
-3.  Create a [Single Redirect](/rules/url-forwarding/single-redirects/create-dashboard/) to forward traffic from your subdomain to your root domain.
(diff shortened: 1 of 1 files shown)


kodster28 (approved):

kodster28:
Thanks for the fix, @haelmj. Embarrassed I missed that when I was writing.

haelmj:
No worries at all...happens to the best of us😅. I appreciate the quick response🙌🏾

PR #8852, opened 2023-05-11
Update create-variables.md

Kiki-Y123:
added line 9, line 15 to focus on variables can only used on zaraz. add more detail to step 5(Create a new variable
) about the plus sign and predefined variables

add more detail about how to use variables after this sentence: Your variable is now ready to be used with tools and triggers.

### DIFF
--- a/content/zaraz/get-started/create-variables.md
+++ b/content/zaraz/get-started/create-variables.md
@@ -6,11 +6,17 @@ weight: 4
 
 # Create a variable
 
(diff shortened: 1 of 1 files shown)


haleycode:
Thank you for suggesting these changes! We'll open an internal ticket so we can address this when our team has bandwidth.

haleycode:
Closing this out for now since we're still waiting on technical validation. We can always reopen in future if needed.

PR #8869, opened 2023-05-12
d1: clarify binding naming

elithrar:
Clarifies what to name a binding.

### DIFF
--- a/content/d1/get-started.md
+++ b/content/d1/get-started.md
@@ -77,7 +77,7 @@ The value (string) you set for `<BINDING_NAME>` will be used to reference this d
 
 Your binding is available in your Worker at `env.<BINDING_NAME>` and the D1 [client API](/d1/platform/client-api/) is exposed on this binding.
 
-You will find the values for `database_name` and `database_id` in your terminal after you run the `create` command in step 3.
+Find the values for `database_name` and `database_id` in your terminal after you run the `create` command in [step 3](/d1/get-started/#3-create-your-database).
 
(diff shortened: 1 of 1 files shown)


content/d1/get-started.md
```diff
@@ -73,7 +73,11 @@ database_name = "<DATABASE_NAME>"
 database_id = "<UUID>"
 ```
 
-Set your binding name by updating the `<BINDING_NAME>` value. Your binding is available in your Worker at `env.<BINDING_NAME>`. You will find the values for `database_name` and `database_id` in your terminal after you run the `create` command in step 3.
+The value (string) you set for `<BINDING_NAME>` will be used to reference this database in your Worker. The binding must be [a valid JavaScript variable name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#variables). For example, `binding = "MY_DB"` or `binding = "productionDB"` would both be valid names for the binding.
+
+Your binding is available in your Worker at `env.<BINDING_NAME>` and the D1 [client API](/d1/platform/client-api/) is exposed on this binding.
+
+You will find the values for `database_name` and `database_id` in your terminal after you run the `create` command in step 3.
```
deadlypants1973:
```suggestion
Find the values for `database_name` and `database_id` in your terminal after you run the `create` command in [step 3](/d1/get-started/#3-create-your-database).
```

deadlypants1973 (approved):
PCX review

PR #8884, opened 2023-05-14
Add changelog entry for performance.now()

irvinebroque:


deadlypants1973 (approved):

jclee:
Ah, OK.  I had PR #8876 pending, which contains similar info, in addition to the previous week's update, but lacked the links... I can update.  I apparently do not have permission to add reviewers to it.

jclee (approved):

PR #8922, opened 2023-05-16
Adding missing parameter in example

ngayerie:
Subpath matching needs to be enabled in the "Redirect all requests for a domain to the root page of different domain" example
SPM-1228


pedrosousa:
Thanks, @ngayerie!

pedrosousa (approved):

PR #8957, opened 2023-05-17
[Load Balancing] RTT values 

patriciasantaana:


content/load-balancing/understand-basics/traffic-steering/steering-policies/dynamic-steering.md
```diff
@@ -18,4 +18,6 @@ For TCP health checks, calculated latency may not reflect the true latency to th
 
 The diagram below shows how Cloudflare would route traffic to the pool with the lowest EWMA among three regions: Eastern North America, Europe, and Australia. In this case, the ENAM pool is selected because it has the lowest RTT.
 
-![Dynamic steering routes traffic to the fastest available pool](/images/load-balancing/traffic-steering-2.png)
\ No newline at end of file
+![Dynamic steering routes traffic to the fastest available pool](/images/load-balancing/traffic-steering-2.png)
+
+RTT values are collected each time a health probe request is made and based on the response from the origin server to the monitor request. When a request is made, Cloudflare inspects the RTT data and use it to sort pools by their RTT values.
```
RebeccaTamachiro:
```suggestion
RTT values are collected each time a health probe request is made and based on the response from the origin server to the monitor request. When a request is made, Cloudflare inspects the RTT data and uses it to sort pools by their RTT values.
```

RebeccaTamachiro: Have read the page as a whole as well, but was a bit confused.
In this paragraph, we are explaining how CF would come to RTT values for each pool, is that so? I think we also talk about this in the 3rd and 4th paragraphs, so, if that is the case, maybe it would help to group them together below the diagram for example? [sth like this](https://drive.google.com/file/d/1DHaWKu0y2w9i2xk0ChmgCaN4O0DKgkPC/view?usp=share_link)

Other than that, I think it's ambiguous whether `based on the response from the origin to the monitor` refers to the `heath probe request` or to the `RTT values`.

If the latter, I would suggest something like the following
(I know it's changing a lot of the text and I don't have previous knowledge of LB, but just in case it's helpful)

> RTTs are based on responses from the origin server to the monitor request in the context of a health check. Each time there is a health probe request, Cloudflare inspects the RTT data and uses it to calculate the RTT profiles and sort the pools.

RebeccaTamachiro (approved):

PR #8958, opened 2023-05-17
beta: Socket Workers

ignoramous:
Socket Workers (private beta) are different to TCP Workers (which are now in public beta).


WalshyDev:
They're the same :) 

They're public beta, please play with them!

PR #8961, opened 2023-05-17
[Durable Objects] pricing: clarify WebSockets pricing outside of DO

ignoramous:
The docs only mention websockets pricing in the context of Durable Objects. For raw Workers themselves, each incoming message is billed as if it was a separate request.

This was mentioned by one of the cf support folks (green cloud) in the `developer-week-2023-chat` discord channel ([link](https://discord.com/channels/595317990191398933/1106627441616756788/1108486541694742560)) just today but there's no documentation on it.

---

Note: I'm not sure if outgoing websocket messages are also billed.

### DIFF
--- a/content/workers/platform/pricing.md
+++ b/content/workers/platform/pricing.md
@@ -29,7 +29,7 @@ All [Pages Functions](/pages/platform/functions/) are billed as Workers. All pri
 
 {{</table-wrap>}}
 
-1.  Requests inbound to your Worker from the Internet are charged on a unit basis for paid plans. [Subrequests](/workers/platform/limits/#subrequests) to external services are not billed on a unit basis, but network time incurred may slightly increase any duration-based billing. For WebSockets, each incoming message is billed as a separate request.
+1.  Requests inbound to your Worker from the Internet are charged on a unit basis for paid plans. [Subrequests](/workers/platform/limits/#subrequests) to external services are not billed on a unit basis, but network time incurred may slightly increase any duration-based billing. WebSockets messages are unmetered and not billed.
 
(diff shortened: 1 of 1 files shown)


rita3ko:
@ignoramous this is not accurate (based on our testing, and intent). is that what you're seeing? if so, could you sent me a note with your account details at rita at cloudflare dot com?

ignoramous:
Hi no, I am not observing otherwise, either. But it is great to know that websocket messages are free.

We were curious on how websockets would be billed if they are used standalone (without DOs). An admin (green cloud) on the group mentioned that since each websocket message is billed as a separate request for DOs, that should also be the case without DOs. I then pointed out the pricing page doesn't say so, and someone guessed that it was probably an oversight because websockets aren't meant to be used without DOs.

Hence, this PR to either clarify if websockets are billed per message (or not). I seems like they're not. So, should that be mentioned in the pricing page: `WebSocket messages are unmetered and not billed.` or something better?

---

Apologies for the confusion.

rita3ko:
It's rare for Websockets to be used without DO since they're not nearly as useful without long running state. But yeah, happy to clarify that per message pricing is not the case for regular workers.

ignoramous:
> But yeah, happy to clarify that per message pricing is not the case for regular workers.

Thanks. I've updated the commit. If the wording isn't clear or if it is more confusing, I'm okay to change it or close this PR, even.

> It's rare for Websockets to be used without DO since they're not nearly as useful without long running state

Gotcha. Outside of DO, WebSockets are now useful (to us at least) since there's possibility to make outgoing TCP connections from Workers.

koeninger:
I think this PR is confusingly worded as is, stateless worker websocket messages can still result in billing for duration, they just don't count towards request count

KimJ15:

PR #8965, opened 2023-05-18
docs(d1): :memo: add persist flag usage instruction for v3

mcnaveen:
Remove `--persist` flag from the dev local run command. Since it's a default behavior now.

Here is the response I got from the Discord server.

> Wrangler will now persist local KV, R2, D1, Cache and Durable Object data
in the .wrangler folder, by default, between reloads. This persistence
directory can be customised with the --persist-to flag. The --persist flag
has been removed, as this is now the default behaviour.

### DIFF
--- a/content/d1/get-started.md
+++ b/content/d1/get-started.md
@@ -158,7 +158,7 @@ While in your project directory, test your database locally by running:
 $ wrangler dev --local --persist
 ```
 
-> Starting from Wrangler 3 you don't need to pass the `--persist` flag.
+> Starting from Wrangler 3 you don't need to pass the `--persist` flag. By default it'll be cached in the `.wrangler` folder and used upon reloads.
 
 When you run `wrangler dev`, Wrangler will give you a URL (most likely `localhost:8787`) to review your Worker. After you visit the URL Wrangler provides, you will see this message: `Call /api/beverages to see everyone who works at Bs Beverages`.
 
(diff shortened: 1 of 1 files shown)


content/d1/get-started.md
```diff
@@ -155,7 +155,7 @@ After configuring your Worker, test your project locally.
 While in your project directory, test your database locally by running:
 
 ```sh
-$ wrangler dev --local --persist
```
elithrar, line 158: This is only the case for wrangler 3, which not everyone is using yet.

elithrar (changes requested):
See comment

content/d1/get-started.md
mcnaveen: I added a new blockquote explaining it's only from Wrangler 3.

content/d1/get-started.md
```diff
@@ -158,6 +158,8 @@ While in your project directory, test your database locally by running:
 $ wrangler dev --local --persist
 ```
 
+> Starting from Wrangler 3 you don't need to pass the `--persist` flag. By default it'll be cached in the `.wrangler` folder and used upon reloads.
```
deadlypants1973, line 161:
```suggestion
> If using Wrangler 3, you do not need to pass the `--persist` flag. By default, it will be cached in the `.wrangler` folder and used on reloads.
```

deadlypants1973 (approved):
PCX

elithrar:
This is out of date - https://developers.cloudflare.com/d1/learning/local-development/#start-a-local-development-session

PR #8967, opened 2023-05-18
[Stream] fix example date filter in fetching-bulk-analytics.md

zefer:
This example `streamMinutesViewedAdaptiveGroups` query would return zero results because the date range filters were the wrong way around. This fixes it to make the example easier to understand and use.


dcpena (approved):

dcpena:
Thanks for catch and contribution @zefer!

zefer:
@dcpena You're welcome!

Issue #8604, opened 2023-04-24
non-enterprise can create reverse dns zones


Tyler-OBrien:
### Which Cloudflare product does this pertain to?

DNS

### Existing documentation URL(s)

https://developers.cloudflare.com/dns/additional-options/reverse-zones

### Section that requires update

**Enterprise** customers who control their own IP prefix(es) can set up reverse zones with PTR records to allow reverse DNS lookups.
...
 Availability

Any customer can create PTR records, but only **Enterprise** customers can create reverse zones.
...
The following Cloudflare customers can create reverse zones (for example, 2.0.192.in-addr.arpa):

**Enterprise** customers with an IPv4 or IPv6 address space can add the IPv4 or IPv6 reverse zone for their IP space to their account, and create the required PTR records for forward resolution.

### What needs to change?

Despite the documentation saying otherwise, non-enterprise users can create & activate reverse zones. I do not know when exactly this was allowed, but it's been a thing for at least the past few months, and I have an active ipv6 reverse dns zone in my non-enterprise account. I assume this goes for IPv4 Reverse Zones as well, as can I add them to my account without issue, but I cannot verify them as I do not own any IPv4.

### How should it change?

Drop all mentions of Enterprise being required ex. "Any customer can create PTR records and reverse zones"

### Additional information

A few hobbynet operators I know of have been turned away from using Cloudflare for Reverse DNS due to this confusing documentation, I hope this can be clarified so I can recommend Cloudflare to more of them!

RebeccaTamachiro:
Thanks for flagging this. The documentation has been updated.

PR #8972, opened 2023-05-18
[DNS] Update reverse zone availability

RebeccaTamachiro:
PCX-6877
GH issue [8604](https://github.com/cloudflare/cloudflare-docs/issues/8604)

### DIFF
--- a/content/dns/additional-options/reverse-zones.md
+++ b/content/dns/additional-options/reverse-zones.md
@@ -60,7 +60,7 @@ To set up a reverse zone, you need to create a reverse DNS zone and add PTR reco
 
 ### Step 2 - Add PTR records
 
-1.  Once you finish zone onboarding, go to **DNS** > **Records**.
+1.  Go to **DNS** > **Records**.
 
 2.  For each IP within the prefix, add a PTR record using the least significant octet(s) as the subdomain.
 
--- a/content/dns/additional-options/reverse-zones.md
+++ b/content/dns/additional-options/reverse-zones.md
@@ -6,11 +6,11 @@ weight: 0
 
 # Reverse zones and PTR records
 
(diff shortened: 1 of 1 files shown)


content/dns/additional-options/reverse-zones.md
```diff
@@ -6,38 +6,34 @@ weight: 0
...
 
 Now, PTR records are primarily useful for those who own a dedicated IP space. They can help populate trace routes and security tools with human-readable domain names.
 
 As PTR records are mainly used for reverse DNS lookups, they should preferably be added to reverse zones.
 
-### Availability
+## Availability
 
-Any customer can create PTR records, but only Enterprise customers can create reverse zones.
+The following Cloudflare customers can create reverse zones.
```
jason-cf, line 23: not sure what this means

content/dns/additional-options/reverse-zones.md
```diff
@@ -62,9 +58,11 @@ To use PTR records, you need to create a reverse DNS zone and add a PTR record f
 
 4.  Skip the rest of the onboarding process.
 
-5.  Once finished with onboarding, go to **DNS** > **Records**.
+### Step 2 - Add PTR records
+
+1.  Once you finish zone onboarding, go to **DNS** > **Records**.
```
jason-cf: Does the user have to finish the zone onboarding before adding their PTR records? my guess is they can configure before adding the NS or waiting for the NS to propogate.

jason-cf (approved):

content/dns/additional-options/reverse-zones.md
RebeccaTamachiro: That makes sense. I was basically just rewording what was already there but now that I think about it, it is even somewhat inconsistent with the previous step of skipping the onboarding. Thanks for the callout 👍

PR #8974, opened 2023-05-18
[Workers] fix import in Node.js path docs

irvinebroque:


deadlypants1973:
Need @jasnell to look at technical accuracy.

irvinebroque:
@deadlypants1973 - anyone on Workers / runtime can review here, doesn't need to be just James.

All this PR is doing is updating the example to use `import` rather than `require`, as is already the case in other examples: https://developers.cloudflare.com/workers/runtime-apis/nodejs/eventemitter/

kflansburg (approved):

deadlypants1973 (approved):

deadlypants1973:
> @deadlypants1973 - anyone on Workers / runtime can review here, doesn't need to be just James.
> 
> All this PR is doing is updating the example to use `import` rather than `require`, as is already the case in other examples: https://developers.cloudflare.com/workers/runtime-apis/nodejs/eventemitter/

No pressure on James, was only referring your your review request a month ago :)

PR #8994, opened 2023-05-19
Update discord link to discord developer main invite

zloveless:
I was reading docs and noticed the "Join the Stream developer community" link linked to a channel directly instead of a proper discord invite. I found the invite via a friend, so I updated it.

The link previously only worked if you were on the discord already
image


kodster28 (approved):

kodster28:
Thanks for the catch! Fixing a few more in the linked PR.

PR #9009, opened 2023-05-19
Update Cursor terms

syeef:


syeef (commented):
use a tags instead of markdown

kodster28 (approved):

PR #9014, opened 2023-05-19
[Workers] Add outbound websockets to Protocols table

isaac-mcfadyen:
- Previously, in the [Protocols table](https://developers.cloudflare.com/workers/platform/protocols/), only MQTT over WebSockets (Pub/Sub) was listed.
- However, Workers support outgoing WebSocket connections too.
- This PR adds a link to the Workers WebSocket API to the table.


deadlypants1973 (approved):

deadlypants1973 (approved):

deadlypants1973:
I believe this is already live so closing

isaac-mcfadyen:
> I believe this is already live so closing

Doesn't look to be. See https://developers.cloudflare.com/workers/platform/protocols/ (the WebSockets section) which only mentions MQTT, which is very different from regular websockets.

PR #9025, opened 2023-05-22
result instead of return in exported_func

liweijian:
`(func (export "exported_func") (param $input i32) (result i32)` will get this error:

```
$ wat2wasm src/simple.wat -o src/simple.wasm
src/simple.wat:9:61: error: unexpected token i32, expected ).
  (func (export "exported_func") (param $input i32) (return i32)
```


kflansburg (approved):
LGTM, thanks for catching this!

deadlypants1973 (approved):
PCX

deadlypants1973:
This change is already live!

PR #9044, opened 2023-05-23
Remove --experimental-local caveat from TCP Socket API docs

irvinebroque:
Now that Wrangler v3 has been released, no longer needed.

### DIFF
--- a/content/workers/runtime-apis/tcp-sockets.md
+++ b/content/workers/runtime-apis/tcp-sockets.md
@@ -156,6 +156,6 @@ const reader = socket.readable.getReader(); // This fails
 ## Considerations
  
 - Outbound TCP sockets to [Cloudflare IP ranges](https://www.cloudflare.com/ips/) are temporarily blocked, but will be re-enabled shortly.
-- TCP sockets must be created within the [`fetch()` handler](/workers/get-started/guide/#3-write-code) of a Worker. TCP sockets cannot be created in global scope and shared across requests. 
+- TCP sockets cannot be created in global scope and shared across requests. You should always create TCP sockets within a handler (ex: [`fetch()`](/workers/get-started/guide/#3-write-code), [`scheduled()`](/workers/runtime-apis/scheduled-event/), [`queue()`](/queues/platform/javascript-apis/#consumer)) or [`alarm()`](/workers/runtime-apis/durable-objects/#alarm-handler-method).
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/tcp-sockets.md
```diff
@@ -156,7 +156,6 @@ const reader = socket.readable.getReader(); // This fails
 ## Considerations
  
 - Outbound TCP sockets to [Cloudflare IP ranges](https://www.cloudflare.com/ips/) are temporarily blocked, but will be re-enabled shortly.
-- When developing locally with [Wrangler](/workers/wrangler/), you must pass the [`--experimental-local`](/workers/wrangler/commands/#dev) flag, instead of the `--local` flag, to use `connect()`.
 - TCP sockets must be created within the [`fetch()` handler](/workers/get-started/guide/#3-write-code) of a Worker. TCP sockets cannot be created in global scope and shared across requests.
```
KianNH: Whilst this PR is open, I guess I might as well ask here - is this supposed to mean it must be created within *any* handler (i.e `scheduled` or `queue`) or is `connect()` specifically restricted to the `fetch` handler?

My assumption was this is related to the fact that `IoContext::current()` in `workerd` throws the "Some functionality ... can only be performed while handling a request" - which itself is a little confusing since as far as I can recall, it also isn't limited to a request (which implies `fetch` handler).

```cpp
IoContext& IoContext::current() {
  if (threadLocalRequest == nullptr) {
    v8::Isolate* isolate = v8::Isolate::TryGetCurrent();
    if (isolate == nullptr) {
      KJ_FAIL_REQUIRE("there is no current request on this thread");
    } else {
      kj::StringPtr message = "Some functionality, such as asynchronous I/O, timeouts, and "
                              "generating random values, can only be performed while handling a "
                              "request.";
      isolate->ThrowException(v8::Exception::Error(jsg::v8Str(isolate, message)));
      throw jsg::JsExceptionThrown();
    }
  } else {
    return *threadLocalRequest;
  }
}
```

irvinebroque: Good callout — thanks for this.

```suggestion
- TCP sockets cannot be created in global scope and shared across requests. You should always create TCP sockets within a handler (ex: [`fetch()`](/workers/get-started/guide/#3-write-code), [`scheduled()`](/workers/runtime-apis/scheduled-event/), [`queue()`](/queues/platform/javascript-apis/#consumer)) or [`alarm()`](/workers/runtime-apis/durable-objects/#alarm-handler-method).
```

dom96 (approved):

PR #9052, opened 2023-05-23
docs(pages): add steps for gatsby dynamic routes support

davwheat:
Gatsby dynamic routes don't work by default in CF Pages, as they rely on rewrites.

This adds a section to the documentation with information about how to use CF Pages Functions to perform rewrite functionality for dynamic routing.


WalshyDev:
Would proxying work here?
https://developers.cloudflare.com/pages/platform/redirects/#proxying

Could also potentially use the `gatsby-plugin-netlify` plugin to make this automatic (just disable it generating Functions) - not looked into this but that seems viable.

KimJ15:
Looks like we have a few viable options so the new documentation section isn't necessary. Closing this ticket, although we can reopen should the solutions not work and we need a new section.

davwheat:
Thanks both.

I did test the Netlify plugin and it seems to work, but I opted to [fork and re-publish](https://github.com/davwheat/gatsby-plugin-cloudflare-pages) it while removing Netlify-specific code and updating the docs for Pages.

Would it be worth putting this new plugin into the CF Pages docs?

PR #9070, opened 2023-05-24
[R2] Use the right `aws s3api` CLI subcommand in content/r2/examples/aws/aws-cli.md

miguelff:


miguelff:
This is wrong, closing. The actual correct command was `aws s3`

PR #9073, opened 2023-05-24
Update common-policies.md

maesoser:
Added an end of line regex symbol `$` to the TLD example in order to block:

```
domain.fail
```

but avoid unintended blocks like:

```
domain.fail.com
```


maxvp (approved):

maxvp:
The regex for catching unintended blocks was implemented [here](https://developers.cloudflare.com/cloudflare-one/policies/filtering/dns-policies/common-policies/#block-top-level-domains).

maxvp (changes requested):
This PR is no longer necessary.

samirbani:
في الثلاثاء، ١٨ يوليو ٢٠٢٣ ٧:٠٥ م Max Phillips ***@***.***>
كتب:

PR #9082, opened 2023-05-24
[Browser Rendering] Tabbed Browser Worker code to support JS & TS

WalshyDev:
Changes:
* Support switching both JS & TS
  * Following existing conventions, JS is default
* Code is now tabbed
* Remove redundant `type` from binding (https://github.com/cloudflare/workers-sdk/blob/4082cfcbdf08740d4a608d3d87df22e51ad0ce4a/packages/wrangler/src/worker.ts#L111C1-L113)


deadlypants1973 (approved):
PCX

PR #9130, opened 2023-05-26
[workers] add supported OS message to wrangler dev doc

admah:
Adding a message to clarify this while we continue to work on additional support.

### DIFF
--- a/content/workers/wrangler/commands.md
+++ b/content/workers/wrangler/commands.md
@@ -321,7 +321,7 @@ None of the options for this command are required. Many of these options can be 
 
 {{<Aside type="warning">}}
 
-As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS >=13 and Windows. We are working to increase support for additional operating systems.
+As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS version 13 or greater and Windows. We are working to increase support for additional operating systems.
 
 When using `wrangler dev`, you need to satisfy [`workerd`](https://github.com/cloudflare/workerd)'s `libc++1` runtime dependencies:
 
--- a/content/workers/wrangler/commands.md
+++ b/content/workers/wrangler/commands.md
@@ -321,7 +321,7 @@ None of the options for this command are required. Many of these options can be 
 
 {{<Aside type="warning">}}
 
-As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS version 13 or greater and Windows. We are working to increase support for additional operating systems.
+As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS version 13 or greater, and Windows (x86-64 architecture). We are working to increase support for additional operating systems.
 
 When using `wrangler dev`, you need to satisfy [`workerd`](https://github.com/cloudflare/workerd)'s `libc++1` runtime dependencies:
 
--- a/content/workers/wrangler/commands.md
+++ b/content/workers/wrangler/commands.md
@@ -321,6 +321,8 @@ None of the options for this command are required. Many of these options can be 
 
 {{<Aside type="warning">}}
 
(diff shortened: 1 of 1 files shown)


admah:
@mrbbot based on [this comment](https://github.com/cloudflare/workers-sdk/issues/3315#issuecomment-1564106973) is this change no longer necessary? Also, should I edit the message about needing `libc++1` runtime dependency?

content/workers/wrangler/commands.md
Cherry, line 1: It might be a good idea to mention which Windows versions are supported here. 7? 8? 10? 11? I assume this is also just x86(-64?) - what about ARM? 32 bit OS versions? wrangler 2 had a great compat story being just Node.js, so it would probably make sense to prepare folks for what OSes (and versions + architectures) are actually supported for local dev with wrangler 3.

I'd also replace "older" with "other" if it were me. I personally wouldn't consider Debian 11 "old" for example as it's the latest stable version: https://www.debian.org/releases/

content/workers/wrangler/commands.md
```diff
@@ -321,6 +321,8 @@ None of the options for this command are required. Many of these options can be
 
 {{<Aside type="warning">}}
 
+As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS >13 and Windows. We are working to increase support for older operating systems.
```
deadlypants1973: Is the mac OS >13 meant to say "macOS version 13 or greater"? Because I would change it to that if so.

admah: Yes. I will make that change. Thanks!

content/workers/wrangler/commands.md
```diff
@@ -321,6 +321,8 @@ None of the options for this command are required. Many of these options can be
 
 {{<Aside type="warning">}}
 
+As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS version 13 or greater and Windows. We are working to increase support for additional operating systems.
```
deadlypants1973:
```suggestion
As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS version 13 or greater, and Windows. We are working to increase support for additional operating systems.
```
Oxford comma vibe. Also to comment above, all Windows versions are supported or do we need to clarify this further?

admah: @deadlypants1973 I'm all about that Oxford comma! I also added the same mention of Windows support that [workerd's docs](https://github.com/cloudflare/workerd#supported-platforms) do - (x86-64 architecture).

content/workers/wrangler/commands.md
admah: Thanks, James! Updated.

deadlypants1973 (approved):

content/workers/wrangler/commands.md
```diff
@@ -321,6 +321,8 @@ None of the options for this command are required. Many of these options can be
 
 {{<Aside type="warning">}}
 
+As of Wrangler v3, `wrangler dev` is currently only supported by Debian 12 Bookworm-based distributions, macOS version 13 or greater, and Windows (x86-64 architecture). We are working to increase support for additional operating systems.
+
 When using `wrangler dev`, you need to satisfy [`workerd`](https://github.com/cloudflare/workerd)'s `libc++1` runtime dependencies:
```
isaac-mcfadyen, line 326: Should/can this be expanded on a bit maybe? If I was a new user I might have no idea what `libc++1` is and how to "satisfy" it.

KianNH: I wonder if Wrangler could just warn that it couldn't find libc++1 (rather than early exit due to false positives) - sometimes you won't even get the "no such file" error and you're just left with an obscure EPIPE error from Node.

Lots of people run into this issue: https://github.com/cloudflare/workers-sdk/issues/2583

admah: @KianNH that's a great idea. I've filed [this](https://github.com/cloudflare/workers-sdk/issues/3447) to track it.

PR #9133, opened 2023-05-28
[Constellation] Clarify wrangler commands, fix bash codeblocks

Erisa:
- Beta version of wrangler is no longer needed
- `--remote` is needed on wrangler dev
- Replace bash codeblocks with sh so they display and copy optimally

### DIFF
--- a/content/constellation/get-started/first-constellation-worker.md
+++ b/content/constellation/get-started/first-constellation-worker.md
@@ -296,6 +296,17 @@ $ npx wrangler dev --remote
 ⬣ Listening at http://0.0.0.0:8787
 ```
 
+{{<Aside type="note">}}
+
+If you're still using Wrangler v2 then run:
+
+```sh
+$ npx wrangler dev
+⬣ Listening at http://0.0.0.0:8787
+```
+
+{{</Aside>}}
+
 To classify some test images, run the following commands in your `image-classifier-worker` Worker:
 
 ```sh
--- a/content/constellation/get-started/first-constellation-worker.md
+++ b/content/constellation/get-started/first-constellation-worker.md
@@ -35,7 +35,7 @@ $ npx wrangler constellation project list
 
 Create a new [Worker](/workers/) named `image-classifier-worker`. You will install [Wrangler, the developer platform CLI, for Constellation](/constellation/platform/wrangler/#installation).
 
-```bash
+```sh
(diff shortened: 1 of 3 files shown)


content/constellation/get-started/first-constellation-worker.md
```diff
@@ -292,14 +291,14 @@ $ wget https://imagedelivery.net/WPOeHKUnTTahhk4F5twuvg/4152ee23-f9af-4b21-a636-
 
 Start a local server to test your `image-classifier-worker` Worker by running [`wrangler dev`](/workers/wrangler/commands/#dev):
 
-```bash
-$ npx wrangler dev
+```sh
+$ npx wrangler dev --remote
```
WalshyDev, line 295: We should add a `<Note>` saying if they're still on 2 to just do `npx wrangler dev`

It doesn't support --remote and it takes a while for people to migrate

Erisa: Aye that's true, I had assumed it was okay here since it's a tutorial and 3 would be installed with the command at the top.
Though, a note for clarity wouldn't be a bad idea!

WalshyDev (approved):

content/constellation/get-started/first-constellation-worker.md
deadlypants1973: If they are running through a Constellation get started guide, they should be on V3. We shouldn't add the docs maintenance work here for us because it'll be hard to update. There should be no issue with a local v3 install for a new Constellation project. Correct me if I am wrong tho.

Erisa: @WalshyDev^

deadlypants1973: Wait but also, isn't --remote deprecated because v3 is automatically remote? It would be the same for v2 and v3?

Erisa: v3 is local by default, v2 is remote by default. Using --remote is the way to get remote dev on v3

WalshyDev: My counter-point: Migration from major versions takes time, even I'm not using 3 yet. Having docs that do not work is very confusing.
If this was in a few months, I'd agree. While it's still fresh, I think this is valid.

celso (approved):

Issue #6497, opened 2022-11-01
md code shell copy unselectable chars


bgbruno:
### Example URL(s)

https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-guide/local

### Actual Behavior

for ` code shell ` - sh , bash
doesn't create element ` span .CodeBlock--token-unselectable `
https://github.com/cloudflare/cloudflare-docs/blob/019a32f673ea965b7eba3827796735a5bb22aa5f/assets/events.ts#L97

so copy block with ` button : Copy to clipboard ` means copy with text ` $ `

### Expected Behavior

copy code like here
https://developers.cloudflare.com/cloudflare-one/tutorials/cli

### Additional information

without it is difficult to reproduce the steps

kodster28:
Thanks again for reporting, @bgbruno (and sorry this took so long to get to).

Confirmed the instance you referenced is fixed.

PR #9134, opened 2023-05-28
Make some improvements to codeblocks

WalshyDev:
Made 3 changes to codeblocks:
* `bash` is now a properly supported language for commands. Which means we no longer need @Erisa to tirelessly go through the whole codebase and change them to `sh` (`bash` also just kinda makes more sense)
* Fixed `$` being copied when you copy from a command codeblock and fixed the command output being copied
* Multi-line commands are now supporteddddddd

Now when you copy from like:
```bash
$ npx command

This is the command output
```

You will only get `npx command` copied :)
Previously you'd get the entire codeblock.

Example page: https://walshy-codeblock-improvement.cloudflare-docs-7ou.pages.dev/constellation/get-started/first-constellation-worker/

### DIFF
--- a/bin/prism.config.ts
+++ b/bin/prism.config.ts
@@ -45,7 +45,7 @@ const shell = {
   },
 
   command: {
-    pattern: /[$](?:[^\r\n])+/,
+    pattern: /\$([^\n]*\\\n)*[^\n]*/g,
     inside: {
       prompt: {
         pattern: /^[$] /,
--- a/bin/prism.config.ts
+++ b/bin/prism.config.ts
@@ -32,7 +32,7 @@ import 'prismjs/components/prism-kotlin.min.js';
 import 'prismjs/components/prism-swift.min.js';
 
 // Custom `shell` grammar
-Prism.languages.sh = {
+const shell = {
   comment: {
     pattern: /(^|[^'{\\$])#.*/,
     alias: 'unselectable',
@@ -52,8 +52,11 @@ Prism.languages.sh = {
         alias: 'unselectable',
       },
     },
-  },
+  }
 };
+
+Prism.languages.sh = shell;
+Prism.languages.bash = shell;
 
 // Prism language aliases
 export const langs: Record<string, string> = {
--- a/components/CodeCopy.vue
+++ b/components/CodeCopy.vue
@@ -7,17 +7,13 @@ function copyCode(e: MouseEvent) {
     e.target as HTMLElement
   )?.parentElement?.parentElement?.querySelector("pre code");
   if (highlightedCode) {
-    // the markdown's code blocks adds a class "CodeBlock--token-unselectable", if it is not supposed to be copied.
-    // clone the code node, we do not want to modify the DOM.
-    const code = highlightedCode.cloneNode(true) as HTMLElement;
-
     // We have 2 types of code blocks, ones with commands and one with general outputs
     // For the ones with commands, we only want the commands copied, not their output
     // So let's first try selecting that
-    let lines = code.querySelectorAll("span:where(.CodeBlock--token-command):not(.CodeBlock--token-unselectable)");
+    let lines = highlightedCode.querySelectorAll("span:where(.CodeBlock--token-command):not(.CodeBlock--token-unselectable)");
     // We found no commands so let's instead copy all output
     if (lines.length === 0) {
-      lines = code.querySelectorAll(".CodeBlock--row");
+      lines = highlightedCode.querySelectorAll(".CodeBlock--row");
     }
 
     let textLines: string[] = [];
--- a/components/CodeCopy.vue
+++ b/components/CodeCopy.vue
@@ -10,13 +10,16 @@ function copyCode(e: MouseEvent) {
     // the markdown's code blocks adds a class "CodeBlock--token-unselectable", if it is not supposed to be copied.
     // clone the code node, we do not want to modify the DOM.
     const code = highlightedCode.cloneNode(true) as HTMLElement;
-    const unselectableTokens = code.getElementsByClassName(
-      "CodeBlock--token-unselectable"
-    );
-    for (let i = 0; i < unselectableTokens.length; i++) {
-      unselectableTokens[i].remove();
+
+    // We have 2 types of code blocks, ones with commands and one with general outputs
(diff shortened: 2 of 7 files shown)


Erisa:
I like the idea here. The `bash` vs `sh` distinction was a bit unique and it does seem I spend a lot of time cleaning those up.

The only thing I would be wary about is some places where `bash` is in use for multi-line codeblocks, take for example this one: https://walshy-codeblock-improvement.cloudflare-docs-7ou.pages.dev/stream/viewing-videos/securing-your-stream/#option-1-using-the-token-endpoint

It doesn't have $ so it's now unselectable but, mercifully it can be copied just fine:
image

There's also [this other one live using sh](https://developers.cloudflare.com/stream/stream-live/start-stream-live/#manage-live-inputs) which shows what would happen if the above did have `$` on the first line:
image

It displays very weird and can't be selected fully, but the copy button again works fine.

This can be sort of fixed by adding `$` to all the lines:
image

A pipe dream I would love would be for the markdown to detect lines ending with `\` and treat the following line as a command also. That would fix multi-line commands right up, but unsure on the complexity.

Erisa (approved):
I think its a positive change though :) but may want to wait for opinions from others

WalshyDev:
Yeah, there's some fun stuff here. Right now the codeblocks have 2 problems, they copy the $ and they copy command output. This fixes those but then breaks multi-line commands...

As you said, the ideal world is that we can support those. However, this looks to be more complex than I have energy for today. I will look into it more over the rest of the long weekend but I'm not sure I can get something perfect.

In my opinion, this is still better, it's 2 steps forward and 1 back. Not great but definitely better.

The key parts for supporting `\` are:
* The prism config, this is where we say that $ is unselectable and creates a command: https://github.com/cloudflare/cloudflare-docs/blob/ce99e5423411a2f19685d0448a85d1942516cad1/bin/prism.config.ts#L35-L56
  * I'm not fully sure we can really make a regex to fully support this, probably but I'm far too tired to figure it out tonight. There may also be a better way which avoids regex, not looked into it though
* The copy code: https://github.com/cloudflare/cloudflare-docs/blob/ce99e5423411a2f19685d0448a85d1942516cad1/components/CodeCopy.vue#L14-L27
  * This one is easier to support, I'm not sure fully how we'd go about it but I think it'll pretty much be detecting when a command starts, going until it ends (end of line + no `\` at the end)

pedrosousa:
My two cents:

We currently have specific internal guidelines on using either `sh` or `bash`, depending if it's a multi-line command or a single-line command (which doesn't mean we were already following these guidelines strictly). If we try to make these two "languages" have the same behavior, we will break a lot of code blocks that rely on the fact the all the lines in the example command are selectable. We would probably have to update all code blocks first before changing Prism's behavior, so that users can select complete commands (or specific lines from these commands) from the documentation without having to use the Copy button.

If we choose to add a prefix to each line belonging to a command, so that we easily know which lines to copy to the clipboard, it should probably be the continuation prompt (`>`) rather than repeating the `$` character, which could make it confusing and wouldn't match what users see in `bash`, for example.

As a final note, we could be assuming too much about what our readers do with the example code blocks we provide. They may want to copy part of an example output to use in their own (different) requests, and switching everything to the `sh` behavior would prevent this possibility.

WalshyDev:
Ok, supporting multi-line commands was easier than expected, this should now work.

> We currently have specific internal guidelines on using either sh or bash, depending if it's a multi-line command or a single-line command (which doesn't mean we were already following these guidelines strictly). If we try to make these two "languages" have the same behavior, we will break a lot of code blocks that rely on the fact the all the lines in the example command are selectable.

Was not aware of this, is this still an issue now that multi-line commands are supported @pedrosousa ?

> As a final note, we could be assuming too much about what our readers do with the example code blocks we provide. They may want to copy part of an example output to use in their own (different) requests, and switching everything to the sh behavior would prevent this possibility.

For sure, I was thinking of making the command output selectable so users can still copy this. The copy button wouldn't so it can still be nicely pasted into a terminal but also allows users to select it if they need.

kodster28:
What you just suggested @WalshyDev does seem to align with some of the internal expectations I've heard (this is anecdotally of course):

> As a final note, we could be assuming too much about what our readers do with the example code blocks we provide. They may want to copy part of an example output to use in their own (different) requests, and switching everything to the sh behavior would prevent this possibility.

I'd _think_ the added flexibility of the copy/paste probably hits any edge cases too, but I'll defer to @pedrosousa here.

kodster28:
Closes #6497

bin/prism.config.ts
```diff
@@ -45,16 +45,19 @@ Prism.languages.sh = {
   },
 
   command: {
-    pattern: /[$](?:[^\r\n])+/,
+    pattern: /\$([^\n]*\\\n)*[^\n]*/g,
```
pedrosousa, line 48: We should update the pattern so that it also works in on Windows (where line endings are `\r\n`).

pedrosousa:
I think we're slowly improving. A few issues I still detected:

1) JSON blocks seem to break break multi-line detection when switching [this example](https://developers.cloudflare.com/ruleset-engine/basic-operations/deploy-rulesets/#example) to `bash` (lines have no ending slash):
image

    Copying the code block only copies up to the opening curly bracket. The rest of the code block is also not selectable.

2) We have a few [heredoc](https://en.wikipedia.org/wiki/Here_document#Unix_shells) occurrences like the following where multi-line detection doesn't work. We can't copy or select past the first line, with the new `bash` behavior:

    image

    https://developers.cloudflare.com/terraform/tutorial/initialize-terraform/#1-define-your-first-terraform-config-file

    https://developers.cloudflare.com/1.1.1.1/encryption/dns-over-https/dns-over-https-client

    This is probably one of those cases where readers will want to copy text from the `heredoc` block.

Detecting `heredoc` may be tricky... If we used line prompt characters (`>`) that wouldn't get copied (like the `$` prompt), parsing and highlighting multi-line commands would probably be easier. It would involve doing more changes to all code blocks currently using `bash`, though, but we would help!

punkeel:
Thanks for working on this, @WalshyDev!

kodster28:
Think this has been sitting too long, unfortunately. Closing out as there are several merge conflicts and hasn't been touched in a while.

PR #9162, opened 2023-05-31
Give clearer advice for the SentinelOne application path

lfcassidy:
Updated the 'Application Path' section to reflect the fact that the agent may run in a folder with a version number.

Added advice similar to the 'Application check' page to suggest using PATH variables to keep this up-to-date.


ranbel:
This is a great idea. Thanks for the suggestion!

Doc has been updated in #9173.

PR #9165, opened 2023-05-31
[Durable Objects] Document alarm feature

Maddy-Cloudflare:
PCX-6576

### DIFF
--- a/content/workers/runtime-apis/durable-objects.md
+++ b/content/workers/runtime-apis/durable-objects.md
@@ -294,18 +294,9 @@ The system calls the `alarm()` handler method when a scheduled alarm time is rea
 
 The method takes no parameters, does not return a result, and can be `async`.
 
-To use alarms:
+#### Use `alarm()` handler method
 
-1. Add the `durable_object_alarms` compatibility flag to your `wrangler.toml`
-
-```toml
----
-filename: wrangler.toml
----
-compatibility_flags = ["durable_object_alarms"]
-```
-
-2. Implement an `alarm()` handler in your Durable Object. The `alarm()` handler will be called when the alarm executes. Call `state.storage.setAlarm()` from anywhere in your Durable Object, and pass in a time for the alarm to run at. Use `state.storage.getAlarm()` to retrieve the currently set alarm time.
+In your Durable Object, the `alarm()` handler will be called when the alarm executes. Call `state.storage.setAlarm()` from anywhere in your Durable Object, and pass in a time for the alarm to run at. Use `state.storage.getAlarm()` to retrieve the currently set alarm time.
 
 The example below will show you how to implement an `alarm()` handler that wakes the Durable Object up once every 10 seconds to batch requests to a single Durable Object. The `alarm()` handler will delay processing until there is enough work in the queue for it to be worth processing them.
 
@@ -332,7 +323,7 @@ export class Batcher {
     this.count++;
 
     // If there is no alarm currently set, set one for 10 seconds from now
-    // Any further POSTs in the next 10 seconds will be part of this kh.
+    // Any further POSTs in the next 10 seconds will be part of this batch.
     let currentAlarm = await this.storage.getAlarm();
     if (currentAlarm == null) {
       this.storage.setAlarm(Date.now() + 10 * SECONDS);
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/durable-objects.md
```diff
@@ -294,6 +294,72 @@ The system calls the `alarm()` handler method when a scheduled alarm time is rea
 
 The method takes no parameters, does not return a result, and can be `async`.
 
+To use alarms:
+
+1. Add the `durable_object_alarms` compatibility flag to your `wrangler.toml`
```
cacheburner: I don't think this is necessary anymore, we turned this compat flag on by default for all compatibility dates.

content/workers/runtime-apis/durable-objects.md
```diff
@@ -294,6 +294,72 @@ The system calls the `alarm()` handler method when a scheduled alarm time is rea
...
+    this.state.blockConcurrencyWhile(async () => {
+      let vals = await this.storage.list({ reverse: true, limit: 1 });
+      this.count = vals.size == 0 ? 0 : parseInt(vals.keys().next().value);
+    });
+  }
+  async fetch(request) {
+    this.count++;
+
+    // If there is no alarm currently set, set one for 10 seconds from now
+    // Any further POSTs in the next 10 seconds will be part of this kh.
```
cacheburner: "Any further POSTs in the next 10 seconds will be part of this **batch**."

content/workers/runtime-apis/durable-objects.md
```diff
@@ -358,7 +349,7 @@ export class Batcher {
 }
 ```
 
-The `alarm()` handler will be called once every 10 seconds. If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine, following a short delay, where it can continue processing.
+The `alarm()` handler will be called once every 10 seconds. If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine. Following a short delay, the `alarm()` handler can continue processing on the other machine.
```
cacheburner: Just to be clear, the `alarm()` handler will start from the beginning in the event an alarm is interrupted while running. The phrase "continue processing" is meant to imply that if the durable object stops running where it has been running, we'll make sure the alarm executes somewhere else.

Only pointing this out because the way it's phrased here makes it seem as though if some chunk of the `alarm()` handler code runs on one machine and the DO is terminated, then we can continue executing where we left off (but on another machine), which _is not_ the case.

content/workers/runtime-apis/durable-objects.md
```diff
@@ -294,6 +294,63 @@ The system calls the `alarm()` handler method when a scheduled alarm time is rea
 
 The method takes no parameters, does not return a result, and can be `async`.
 
+#### Use `alarm()` handler method
+
+In your Durable Object, the `alarm()` handler will be called when the alarm executes. Call `state.storage.setAlarm()` from anywhere in your Durable Object, and pass in a time for the alarm to run at. Use `state.storage.getAlarm()` to retrieve the currently set alarm time.
+
+The example below will show you how to implement an `alarm()` handler that wakes the Durable Object up once every 10 seconds to batch requests to a single Durable Object. The `alarm()` handler will delay processing until there is enough work in the queue for it to be worth processing them.
```
pedrosousa:
```suggestion
The example below implements an `alarm()` handler that wakes the Durable Object up once every 10 seconds to batch requests to a single Durable Object. The `alarm()` handler will delay processing until there is enough work in the queue.
```

content/workers/runtime-apis/durable-objects.md
pedrosousa:
```suggestion
The `alarm()` handler will be called once every 10 seconds. If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine. Following a short delay, the `alarm()` handler will run from the beginning on the other machine.
```

content/workers/runtime-apis/durable-objects.md
```diff
@@ -294,6 +294,63 @@ The system calls the `alarm()` handler method when a scheduled alarm time is rea
 
 The method takes no parameters, does not return a result, and can be `async`.
 
+#### Use `alarm()` handler method
```
pedrosousa: Maybe just `#### Example` or `#### How to use` here?

pedrosousa (approved):

content/workers/runtime-apis/durable-objects.md
Maddy-Cloudflare: If an unexpected error terminates the Durable Object, `alarm()` will start from the beginning somewhere else. 

So, it sounds like there's actually **_nothing_** that continues processing here. If there's an error, the alarm will start from the beginning, but somewhere else (where exactly, though? Is it just another machine?)

cacheburner:
Thanks Maddy!

PR #9273, opened 2023-06-06
Update use-cases.md to fix typo

mattsimonsen:


pedrosousa:
Hello @mattsimonsen 
Thank you for letting us know that we had an incorrect configuration for a sample use case in the Bulk Redirects documentation. To match the use case description and the redirects table, we were in fact missing one configuration parameter. 

The example redirects in the table matched the use case description (all incoming requests redirected to the root page of a new domain). Therefore, I kept the table as it was before, and added a new parameter to the configuration: **Preserve path suffix**.

pedrosousa (approved):

PR #8579, opened 2023-04-21
[ZT] mobile WARP client migration

ranbel:
addresses PCX-6223 and PCX-7912

PR #9286, opened 2023-06-08
Move Android WARP URL to Zero Trust Agent

wolruf:
Cloudflare employee here. Should we direct users to Zero Trust Agent instead of the 1.1.1.1 app on Android ?


ranbel:
Hi there, we are putting together a migration guide in #8579.

PR #9321, opened 2023-06-10
[turnstile] add legalese for invisible mode

Le0Developer:
This is taken 1:1 from hCaptcha ([here](https://docs.hcaptcha.com/invisible/#invisible-vs-passive) and [here](https://docs.hcaptcha.com/faq/#do-i-need-to-display-anything-on-the-page-when-using-hcaptcha-in-invisible-mode)).

reCAPTCHA does not have such a note in their docs, but they always [show their own badge](https://developers.google.com/recaptcha/docs/versions#recaptcha_v2_invisible_recaptcha_badge) instead.

**Wording probably needs to be changed to make (their) lawyers happy**

### DIFF
--- a/content/turnstile/frequently-asked-questions.md
+++ b/content/turnstile/frequently-asked-questions.md
@@ -199,3 +199,37 @@ You can safely ignore the error. It is requesting a [Private Access Token (PAT)]
 
 {{</faq-answer>}}
 {{</faq-item>}}
+
+{{<faq-item>}}
+{{<faq-question level=2 text="Do I need to display anything on the page when using Turnstile in Invisible mode?" >}}
+{{<faq-answer>}}
+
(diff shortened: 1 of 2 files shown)


patriciasantaana:
@Le0Developer  — Thank you for the feedback about how we can make improvements with the Turnstile documentation! I will be closing this issue as we will have to run this legalese through our legal team first.

patriciasantaana:
Hi @Le0Developer ! 

As noted in the hCaptcha links you provided, privacy laws and requirements vary depending on the jurisdiction(s) in which both our customers and their end users are located. Our customers are in the best position to determine which privacy laws are applicable to them and It is up to our customers to determine whether and how to use the invisible version of our Turnstile product on their website.

Please also note that while Turnstile's invisible mode does run non-interactive JavaScript challenges to gather signals about the visitor/browser environment, those challenges and signals are used with the sole purpose of providing the end user with a token as an authentication mechanism, and are not sold to any third parties, including for any advertising purposes.

PR #9364, opened 2023-06-14
[Pages] Update Pages Queue bindings in bindings.md

kewbish:
This PR updates the Queue bindings to use `context.env` instead of the nonexisting `env` variable, and correctly types the Queue as the generic `Queue<T>` in the TypeScript example.


WalshyDev (approved):
Thank you!

PR #9396, opened 2023-06-15
Update Understanding and configuring Cloudflare Page Rules (Page Rule…

jeremysimmons:
full url example isn't rendering. 
use literal code markup so that it will render in webpage.

### DIFF
--- a/content/support/Page Rules/Understanding and configuring Cloudflare Page Rules (Page Rules Tutorial).md
+++ b/content/support/Page Rules/Understanding and configuring Cloudflare Page Rules (Page Rules Tutorial).md
@@ -40,7 +40,7 @@ Cloudflare recommends ordering your rules from most specific to least
 specific.
 {{</Aside>}}
 
-A page rule matches a URL pattern based on the following format (comprised of five segments): `<scheme>://<hostname><:port>/<path>?<query\_string>`
+A page rule matches a URL pattern based on the following format (comprised of five segments): `<scheme>://<hostname><:port>/<path>?<query_string>`
 
 An example URL with these four segments looks like:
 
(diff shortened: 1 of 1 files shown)


ngayerie (approved):

content/support/Page Rules/Understanding and configuring Cloudflare Page Rules (Page Rules Tutorial).md
```diff
@@ -40,7 +40,7 @@ Cloudflare recommends ordering your rules from most specific to least
 specific.
 {{</Aside>}}
 
-A page rule matches a URL pattern based on the following format (comprised of five segments): <scheme>://<hostname><:port>/<path>?<query\_string>
+A page rule matches a URL pattern based on the following format (comprised of five segments): `<scheme>://<hostname><:port>/<path>?<query\_string>`
```
pedrosousa:
```suggestion
A page rule matches a URL pattern based on the following format (comprised of five segments): `<scheme>://<hostname><:port>/<path>?<query_string>`
```
We don't need to escape underscores now that we're using inline code formatting.

jeremysimmons:
@pedrosousa is it possible to bypass the spell-check rule?

pedrosousa:
Yes, I can still merge the PR (once it's ready) even if the spell-checking action fails.

jeremysimmons:
> Yes, I can still merge the PR (once it's ready) even if the spell-checking action fails.

Thanks @pedrosousa anything else you need from me. Thanks for being on this. It thrills me that you allow customers to contribute to the docs.

pedrosousa:
Do you mind committing my suggestion? I can't do it myself in GitHub's web interface (only by checking out the PR locally), since I was the one making the suggestion.

If that option is also not available for you (since the comment has been resolved), I can do it manually.

pedrosousa:
Thank you for the fix, @jeremysimmons!

PR #9429, opened 2023-06-18
Update set-browser-ttl.md

luishcardoso:
Hi team! We can benefit from warning the customers that setting a high Browser Cache TTL value can be dangers; because after that the cached data can only be purged locally by the website visitor on its device.


ngayerie (changes requested):
PR to be closed, following up in PCX-7353

samirbani:
في الأربعاء، ١٩ يوليو ٢٠٢٣ ١١:٢٩ ص angelampcosta ***@***.***>
كتب:

PR #9456, opened 2023-06-19
[workers] Add timing attack example

### LINKED ISSUE #7659: Examples are Vulnerable to Timing Attack

Workers

https://developers.cloudflare.com/workers/examples/auth-with-headers/
https://developers.cloudflare.com/workers/examples/basic-auth/

https://developers.cloudflare.com/workers/examples/auth-with-headers/
```js
if (psk === PRESHARED_AUTH_HEADER_VALUE) {
```

https://developers.cloudflare.com/workers/examples/basic-auth/
```js
if (BASIC_PASS !== pass) {
```

### What needs to change?

Isn't this method of string comparison not vulnerable to timing attacks? I'm new to JS and workers, so not sure what JS I could write to do the string comparisons that would be constant time.

### How should it change?

These string comparisons should be constant time. Even though these are examples, would be much better to show method that is secure because they will no doubt be copied and used often. Would also be great to build some sort of helper for constant time comparisons into the worker platform as I am trying to figure out how to do so now.

### Additional information

_No response_

kristianfreeman:
Closes #7659

### DIFF
--- a/content/workers/examples/auth-with-headers.md
+++ b/content/workers/examples/auth-with-headers.md
@@ -11,6 +11,12 @@ weight: 1001
 layout: example
 ---
 
+{{<Aside type="note">}}
+
+This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, see the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
+
+{{</Aside>}}
+
 {{<tabs labels="js | ts">}}
 {{<tab label="js" default="true">}}
 
--- a/content/workers/examples/auth-with-headers.md
+++ b/content/workers/examples/auth-with-headers.md
@@ -11,9 +11,9 @@ weight: 1001
 layout: example
 ---
 
-{{<Aside type="note">}}
+{{<Aside type="warning" header="Caution when using in production">}}
 
-This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, see the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
+This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, refer to the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
 
 {{</Aside>}}
 
--- a/content/workers/examples/basic-auth.md
+++ b/content/workers/examples/basic-auth.md
@@ -11,6 +11,12 @@ weight: 1001
 layout: example
 ---
 
+{{<Aside type="note">}}
+
+This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, see the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
+
+{{</Aside>}}
+
 {{<tabs labels="js | ts">}}
 {{<tab label="js" default="true">}}
 
--- a/content/workers/examples/basic-auth.md
+++ b/content/workers/examples/basic-auth.md
@@ -11,9 +11,9 @@ weight: 1001
 layout: example
 ---
 
-{{<Aside type="note">}}
+{{<Aside type="warning" header="Caution when using in production">}}
 
-This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, see the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
+This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, refer to the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
 
 {{</Aside>}}
 
--- a/content/workers/examples/protect-against-timing-attacks.md
+++ b/content/workers/examples/protect-against-timing-attacks.md
@@ -11,7 +11,7 @@ layout: example
 
 To avoid timing attacks in your code, you can replace equality checks with the [`crypto.timingSafeEqual`](/workers/runtime-apis/web-crypto/#timingsafeequal) function in your Workers application.
 
-To use this function, you can create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) instance, encode the values you need to compare using [`encoder.encode`](/workers/runtime-apis/encoding/#methods), and replace the standard JavaScript equality check (`===`) with `crypto.timingSafeEqual`:
+To use this function, you can create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) and encode the string values to instances of `ArrayBuffer` using [`encoder.encode`](/workers/runtime-apis/encoding/#methods). This is needed because `crypto.timingSafeEqual` compares `ArrayBuffer` instances, not strings. With the encoded values, you can then replace the standard JavaScript equality check (`===`) with `crypto.timingSafeEqual`. Note that the strings must be the same length in order to compare to `timingSafeEqual`. The below code shows how to implement string equality checks with `crypto.timingSafeEqual`:
 
 {{<tabs labels="js | ts">}}
 {{<tab label="js" default="true">}}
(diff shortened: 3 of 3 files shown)


kristianfreeman:
cc @KianNH in case you're interested in reviewing - @deadlypants1973 and I chatted this morning and she sent me your code to use as a starting place for this 🙇

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -0,0 +1,70 @@
...
+
+// The below code is vulnerable to timing attacks
+// if (string1 === string2) { ... }
+
+// You can replace it with `crypto.timingSafeEqual` by encoding the values
+// you need to compare
+const a = encoder.encode(string1);
+const b = encoder.encode(string2);
+
+let equal = crypto.timingSafeEqual(a, b)
```
KianNH, line 39:
```suggestion
let equal = crypto.timingSafeEqual(a, b)
```

This will throw if `a` and `b` are not equal lengths, so users will want to try/catch it to prevent an unhandled exception (or do the length checks beforehand).

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -0,0 +1,70 @@
...
+  - Security
+pcx_content_type: configuration
+title: Using timingSafeEqual
+weight: 1001
+layout: example
+---
+
+To avoid timing attacks in your code, you can replace equality checks with the [`crypto.timingSafeEqual`](/workers/runtime-apis/web-crypto/#timingsafeequal) function in your Workers application.
+
+To use this function, you can create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) instance, encode the values you need to compare using [`encoder.encode`](/workers/runtime-apis/encoding/#methods), and replace the standard JavaScript equality check (`===`) with `crypto.timingSafeEqual`:
```
KianNH: It might be worth telling users why we use `TextEncoder()` - which is that `timingSafeEqual` operates on ArrayBuffers rather than plain strings.

KianNH (commented):
Looks good to me!

It might be worth linking to the example on https://developers.cloudflare.com/workers/runtime-apis/web-crypto/#timingsafeequal too.

content/workers/examples/protect-against-timing-attacks.md
kristianfreeman: Thanks, fixed!

content/workers/examples/protect-against-timing-attacks.md
kristianfreeman: Thanks, added

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -0,0 +1,82 @@
...
+const encoder = new TextEncoder();
+
+const string1 = "foo"
+const string2 = "bar"
+
+if (string1.length !== string2.length) {
+  // Strings must be the same length in order to compare
+  // with crypto.timingSafeEqual
+  return false
+}
```
KianNH: We'd want to compare the lengths of the `ArrayBuffer`s outputted by the `TextEncoder` - because of funny Unicode.

```js
> "a".length
1
> "ä".length
1
> new TextEncoder().encode("a").length
1
> new TextEncoder().encode("ä").length
2
```

KianNH: More accurately, it seems to be the `byteLength` of the `ArrayBuffer` - cc @jasnell who likely knows for sure

```cpp
bool SubtleCrypto::timingSafeEqual(kj::Array<kj::byte> a, kj::Array<kj::byte> b) {
  JSG_REQUIRE(a.size() == b.size(), TypeError, "Input buffers must have the same byte length.");

  // The implementation here depends entirely on the characteristics of the CRYPTO_memcmp
  // implementation. We do not perform any additional verification that the operation is
  // actually timing safe other than checking the input types and lengths.

  return CRYPTO_memcmp(a.begin(), b.begin(), a.size()) == 0;
}
```

kristianfreeman: oh interesting! I'll wait to hear what @jasnell thinks before updating

jasnell: Yep, the comparison is on `byteLength` here. Be aware, however, that timing issues could still be introduced by the conversion of strings to utf8 using the `TextEncoder` since that conversion process is not guaranteed to be timing safe. It's not entirely possible to completely eliminate that risk but it is possible to minimize it by first comparing the string lengths before performing the conversion and the timeSafeEqual call.

content/workers/examples/auth-with-headers.md
```diff
@@ -11,6 +11,12 @@ weight: 1001
 layout: example
 ---
 
+{{<Aside type="note">}}
```
deadlypants1973:
```suggestion
{{<Aside type="warning" header="Caution when using in production">}}
```

content/workers/examples/auth-with-headers.md
```diff
@@ -11,6 +11,12 @@ weight: 1001
...
+
+This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, see the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
```
deadlypants1973:
```suggestion
This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, refer to the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
```

content/workers/examples/basic-auth.md
```diff
@@ -11,6 +11,12 @@ weight: 1001
 layout: example
 ---
 
+{{<Aside type="note">}}
+
+This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, see the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
```
deadlypants1973:
```suggestion
This code is provided as a sample, and is not suitable for production code without protecting against timing attacks. To learn how to implement production-safe code, refer to the [`timingSafeEqual` example](/workers/examples/protect-against-timing-attacks/) for more information on how to mitigate against timing attacks in your Workers code.
```

content/workers/examples/basic-auth.md
```diff
@@ -11,6 +11,12 @@ weight: 1001
 layout: example
 ---
 
+{{<Aside type="note">}}
```
deadlypants1973:
```suggestion
{{<Aside type="warning" header="Caution when using in production">}}
```

deadlypants1973 (commented):
PCX

content/workers/examples/protect-against-timing-attacks.md
kristianfreeman: Just updated again. Do you guys mint taking a look at it and making sure I'm on the right track?

jasnell (approved):

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -0,0 +1,82 @@
...
+  - Security
+pcx_content_type: configuration
+title: Using timingSafeEqual
+weight: 1001
+layout: example
+---
+
+To avoid timing attacks in your code, you can replace equality checks with the [`crypto.timingSafeEqual`](/workers/runtime-apis/web-crypto/#timingsafeequal) function in your Workers application.
+
+To use this function, you can create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) and encode the string values to instances of `ArrayBuffer` using [`encoder.encode`](/workers/runtime-apis/encoding/#methods). This is needed because `crypto.timingSafeEqual` compares `ArrayBuffer` instances, not strings. With the encoded values, you can then replace the standard JavaScript equality check (`===`) with `crypto.timingSafeEqual`. Note that the strings must be the same length in order to compare to `timingSafeEqual`. The below code shows how to implement string equality checks with `crypto.timingSafeEqual`:
```
deadlypants1973:
```suggestion
To use this function, create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) and encode the string values to instances of `ArrayBuffer` using [`encoder.encode`](/workers/runtime-apis/encoding/#methods). This is needed because `crypto.timingSafeEqual` compares `ArrayBuffer` instances, not strings. With the encoded values, replace the standard JavaScript equality check (`===`) with `crypto.timingSafeEqual`. Note that the strings must be the same length in order to compare to `timingSafeEqual`. The below code shows how to implement string equality checks with `crypto.timingSafeEqual`:
```

deadlypants1973 (approved):
PCX

PR #9460, opened 2023-06-19
Remove CNAME setup mention on child setups

cicku:
Please squash and merge due to some bad commit histories


RebeccaTamachiro:
Hi Christopher. Thanks for bringing this up.
I had seen it but have talked to Hannes about it and the reason I have not merged this yet is that, as it is, steps that refer **TXT verification record** become disconnected/ can cause confusion. We can assume users have a TXT verification record bc we previously told them to convert to partial.
Have created PCX-7471 to look into it later, but if you have the chance/want to go for it as well (or want to add more context/comment on that ticket), always welcome.

PR #9502, opened 2023-06-21
Update create-custom-hostnames.md

ngayerie:
Clarifying that min tls version is 1.0 when creating custom hostnames


kodster28:
Heyo @ngayerie, I'll need to reformat this in a bit.

We want to avoid multiple asides in a row b/c it leads folks to just skip over them.

kodster28 (approved):

PR #9507, opened 2023-06-21
[GH actions] Update assignment script to reduce noise

kodster28:


.github/actions/owners.ts
```diff
@@ -83,6 +83,7 @@ export const REVIEWERS: Record<string, string[]> = {
   pages: ["WalshyDev"],
   "pub-sub": ["CharlieBurnett", "elithrar"],
   queues: ["CharlieBurnett", "elithrar"],
+  support: ["shanecloudflare", "zeinjaber", "rprice74", "ngayerie"],
```
pedrosousa, line 86: These users will only know about PRs, not issues (just highlighting a side effect of this change).

pedrosousa (approved):

.github/actions/owners.ts
kodster28: Yeah.... I think that's intended for now. 

What isn't fun is that they're pretty active creating PRs, and the current logic then bypasses our normal assignment process.

PR #9519, opened 2023-06-21
Document Brotli compatibility flag

irvinebroque:
refs https://github.com/cloudflare/workerd/pull/708

- Add changelog entry
- Document compatibility flags under "experimental" section (Todo: move to other section once there is an enabled by default date set)
- Add a section to the `fetch()` API page with an example of making a subrequest with `Content-Encoding` specified


content/workers/runtime-apis/fetch.md
```diff
@@ -68,6 +68,32 @@ async function eventHandler(event) {
 
 ---
 
+
+## How the `Accept-Encoding` header is handled
```
irvinebroque, line 72: @deadlypants1973 — I know you're thinking about docs structure — I put this content here for now since there wasn't otherwise a clear home, but it's not necessarily the _most_ ideal or intuitive place for it, and would love to find a more platform-level (or similar, whatever the new thing ends up being) way to communicate "here are the content encoding / compression types that Workers supports".

fhanau:
Superseded by #13684 – should be ready to close.

PR #9522, opened 2023-06-22
[Logs] Update to Logpush

angelampcosta:


content/logs/reference/filters.md
```diff
@@ -216,6 +216,8 @@ The filter field has limits of approximately 30 operators and 1000 bytes. Anythi
 
 ## Logical Operators
 
+- Logical operators are optional, but if used, it must contain the `where` key.
```
kvap: This should say "Filter is optional, but if used, it must contain the `where` key." and moved to previous sections.

angelampcosta: @kvap You mean the section "## Setting filters via API or dashboard"?

kvap: Yeah, that would work.

pedrosousa (approved):

PR #9523, opened 2023-06-22
Update create-variables.md

Kiki-Y123:
added section ## Variables Types and combinations to explain different variable types and combinations and how flexible it can be as below. 

Variables types can be very flexible with different combinations as needed:
1. It can be a self-defined variable name + a predefined variable value. 
2. It can be a self-defined variable name + a self-defined variable value.
3. It can be a predefined variable with predefined name/type/value, only need to click the plus sign in the relevant field and then select a predefined variable. 
4. It can be a concatenated variable, between self defined variable values or predefined variable values.


KimJ15:
@bjesus do/can you approve this PR?

PR #9537, opened 2023-06-22
fix button alignment for algolia search

kristianfreeman:


kodster28:
Already addressed in another PR

PR #9555, opened 2023-06-23
[Support] Add Brotli compression

pedrosousa:
Also related to PCX-6689.


content/support/Speed/Optimization - File Size/What will Cloudflare compress.md
```diff
@@ -64,10 +81,14 @@ application/graphql+json
 application/geo+json
 ```
 
-Cloudflare's global network can deliver content to website visitors using gzip compression, [Brotli compression](#enable-brotli-compression), or no compression, according to the values visitors provide in the `Accept-Encoding` request header.
+Cloudflare's global network can deliver content to website visitors using GZIP compression, [Brotli compression](#enable-brotli-compression), or no compression, according to the values visitors provide in the `Accept-Encoding` request header.
 
 For responses with error status codes, Cloudflare will only compress responses if their error status code is `403` or `404`. For successful response status codes, Cloudflare will only compress responses if their status code is `200`. Responses with other status codes will not be compressed.
 
+{{<Aside type="note">}}
+Enterprise customers can use [Compression Rules](/rules/compression-rules/) to override Cloudflare's default compression behavior.
```
pedrosousa, line 89: Requires #9554 to be merged first. This link is being reported as broken (which would prevent the merge), but I'll update the branch once the PR I mentioned is live.

marciocloudflare (approved):

PR #9566, opened 2023-06-23
wae: clarify limits

elithrar:
Clarify the "25 writes per request" limit a little more:

image


tlozoot (approved):
LGTM thanks for updating this!

PR #9620, opened 2023-06-27
Fix #9604 tab targets and Pages build system tabs

### LINKED ISSUE #9604: Pages Language support and tools page has broken tabs
### Example URL(s)

https://developers.cloudflare.com/pages/platform/language-support-and-tools/

### Actual Behavior

When you first load the page, both v1 and v2 tabs are active on the tools,  frameworks, and environment tables:

image

Further, the v2 tab does not work when you click the "beta" tag - only when you click the v2 text does it activate

pages-docs-tab-bug

### Expected Behavior

- Only the v2 tab should be highlighted on page load.
- When you click any part of the v2 tab button, it should switch to the v2 tab.

### Additional information

_No response_

GregBrimble:
- Changes the tab click targets to look at the nearest ancestors to determine which tab it belongs to.
- Fixes Pages build system tabs to only provide the "tab-default" rather than both a "tab-default" and an "active" tab.


kodster28 (approved):
LGTM

PR #9622, opened 2023-06-27
Update seat-management.md

luishcardoso:
- Removed is the closest we have to a deleted or archived user record at the moment.

Suggestion: Add that information as a note under  Zero Trust  Dev doc Seat management -> Remove a user

Thanks!


ranbel:
Thanks for the suggestion! I put your note into its own section.

ranbel (approved):

PR #9655, opened 2023-06-29
workers: update beta page

elithrar:
* Make it clearer that there isn't a waitlist for public alphas and betas (this is continually glossed over)
* Rename "TCP Workers" as "TCP Sockets" for clarity, as this is a source of confusion vs. "inbound" TCP Workers


irvinebroque (approved):
thank you!

PR #9686, opened 2023-06-30
[WAF] Add `cf-terraforming` guidance to MR migration guide

pedrosousa:


content/waf/reference/migration-guides/waf-managed-rules-migration.md
```diff
@@ -376,7 +388,118 @@ For more information, refer to the following resources:
...
+        [...]
+      }
+      [...]
+    }
+    [...]
+    ```
+
+2. The previous command may return additional ruleset configurations for other Cloudflare products also based on the [Ruleset Engine](/ruleset-engine/). Since you are looking for the WAF Managed Rules configuration, keep only the Terraform resource for the `http_request_firewall_managed` phase and save it to a `.tf` configuration file. You will need the full resource name in the next step.
+
+3. Import the `cloudflare_ruleset` resource you previously identified into Terraform state using the `terraform import` command. For example:
```
cjolowicz, line 422: Should we recommend `cf-terraforming import --resource-type=cloudflare_ruleset` here? This command will print out the exact `terraform import` invocations needed. The same caveat as before applies, though: You'll have to filter out the ruleset resources for the `http_request_firewall_managed` phase (I think).

cjolowicz (approved):

content/waf/reference/migration-guides/waf-managed-rules-migration.md
pedrosousa: Since the output of `cf-terraforming import (...)`  will be full of IDs with no descriptions (the output length will vary according to the configured rulesets across different products, of course), maybe it's easier to ask the user to save the resource name (step 2) and then enter a single `terraform import (...)` command. The user will need to replace the `<ZONE_ID>` placeholder, though... What do you think?

jason-cf (approved):

content/waf/reference/migration-guides/waf-managed-rules-migration.md
cjolowicz: Thanks for clarifying. You're right -- it's just a single resource, so going via `cf-terraforming import` doesn't really help at all. Replacing the `<ZONE_ID>` placeholder seems clear enough. I think this is good as-is.

PR #9694, opened 2023-07-02
Dark mode support for Mermaid diagrams

Lekensteyn:
Use a [dark Mermaid theme](https://mermaid.js.org/config/theming.html) when dark mode is enabled. As mentioned in https://github.com/mermaid-js/mermaid/issues/2644, automatically using the device theme is not supported.

This uses a [MutationObserver](https://caniuse.com/mdn-api_mutationobserver_observe) to watch for changes to the theme. I am not that familiar with Vue.js and tried `watch` combined with `ref` in `assets/theme.ts`, but it somehow did not trigger `watch`. So I opted for the current approach. Feedback is welcome!

Example with multiple diagrams (`/ssl/troubleshooting/too-many-redirects/`):

* mermaid-light
* mermaid-dark

The original diagram that triggered me from `/workers/learning/how-the-cache-works/`. The labels on the arrows are not pretty, but at least it does not burn my eyes:

* mermaid-dark-labels


kodster28 (approved):
Thanks for the help, @Lekensteyn. LGTM.

PR #9720, opened 2023-07-05
Update get-started.md

xpat:
Functions should be implemented as server endpoints in your SvelteKit app.

### DIFF
--- a/content/pages/platform/functions/get-started.md
+++ b/content/pages/platform/functions/get-started.md
@@ -14,7 +14,7 @@ You must have a Pages project set up on your local machine or deployed on the Cl
 
 ## Create a Function
 
-To get started with generating a Pages Function, create a `/functions` directory at the root of your Pages project. 
+To get started with generating a Pages Function, create a `/functions` directory at the root of your Pages project.
 
 {{<Aside type="note" header="Advanced mode">}}
 
--- a/content/pages/platform/functions/get-started.md
+++ b/content/pages/platform/functions/get-started.md
@@ -6,7 +6,7 @@ weight: 1
 
 # Get started
 
-This guide will instruct you on creating and deploying a Pages Function.
+This guide will instruct you on creating and deploying a Pages Function. With some frameworks, such as Sveltekit, this approach will not work. Instead, Functions should be implemented as server endpoints in your SvelteKit app. For example, see https://kit.svelte.dev/docs/adapter-cloudflare#bindings
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


content/pages/platform/functions/get-started.md
```diff
@@ -6,15 +6,15 @@ weight: 1
...
+This guide will instruct you on creating and deploying a Pages Function. With some frameworks, such as Sveltekit, this approach will not work. Instead, Functions should be implemented as server endpoints in your SvelteKit app. For example, see https://kit.svelte.dev/docs/adapter-cloudflare#bindings
 
 ## Prerequisites
 
 You must have a Pages project set up on your local machine or deployed on the Cloudflare dashboard. To create a Pages project, refer to [Get started](/pages/get-started/).
 
 ## Create a Function
 
-To get started with generating a Pages Function, create a `/functions` directory at the root of your Pages project.
+To get started with generating a Pages Function, create a `/functions` directory at the root of your Pages project.
```
WalshyDev:
```suggestion
To get started with generating a Pages Function, create a `/functions` directory at the root of your Pages project.
```

dario-piotrowicz:
Hi @xpat , thank you for the PR, it's much appreciated 😄 

But I don't think it's really correct, let me explain what I mean, [Cloudflare Pages Functions](https://developers.cloudflare.com/pages/platform/functions/) are an wide-ish term that includes two different modes, the ["standard" one in which you define a `/functions`](https://developers.cloudflare.com/pages/platform/functions/routing/) directory and the [advanced one](https://developers.cloudflare.com/pages/platform/functions/advanced-mode/) in which you generate an `_worker.js`.

Both modes are referred to as Pages Functions and SvelteKit uses the advanced mode as described in [their docs](https://kit.svelte.dev/docs/adapter-cloudflare#notes).

So it is indeed using (/producing) a Pages functions application. The only thing that does not work here is the `functions/` directory but that:
 - doesn't mean that the deployment doesn't use Pages Functions because it definitely does
 - is the exact same for all meta-frameworks, SvelteKit **not** being an exception, so if anything we'd need a way to tell users that they **can** use a `/functions` directory only when they deploy a static site (but I'm not sure where the best place for that would be)
 - is already made clear later in the [SvelteKit guide](https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-site/#functions-setup) as well as in the [SvelteKit docs](https://kit.svelte.dev/docs/adapter-cloudflare#notes) themselves
 
Additionally as an example you're pointing to [this](https://kit.svelte.dev/docs/adapter-cloudflare#bindings) but as I mentioned we do already have an example below in our guide, so I again fail to see the benefit of adding an additional example here (which is also not presented as an endpoint example but an example of how to access bindings, which in my opinion makes it less clear in this context).

I totally appreciate the the Cloudflare terminology is not helping and that we are likely not conveying super clearly when and when not to use the `/functions` directory. I just don't think that the proposed changes make things any clearer, on the contrary it looks to me like they introduce more confusion/ambiguity. We could change them to mention that the `/functions` directory is not supported here but again that's already mentioned in a couple of places and applies to all meta-frameworks so I feel like we'd need to convey this is a more broad/general sense and not just in Svelte 🤔

PR #9758, opened 2023-07-07
[R2] Fix R2.list() return types

leaysgur:
Currently docs say it's return type is `Promise<R2Objects | null>`, but actually it is `Promise<R2Objects>`.

image

> https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#bucket-method-definitions

TypeScript definitions is also defined as `Promise<R2Objects>`.

image

See also https://github.com/cloudflare/workerd/blob/f5388c93bed7f753fccc9ce766d3247e3d2103bf/src/workerd/api/r2-bucket.h#L335


dcpena (approved):

dcpena:
Thanks for the contribution @leader22! Much appreciated and great catch.

PR #9772, opened 2023-07-07
[DNS] Email records shouldn't be proxied

Erisa:
Proxied email records can at worst break email delivery and at best cause rDNS issues since we automatically assign a random subdomain to reveal the real IP.

Additionally the API example underneath uses `proxied: false` which is the opposite  of the GUI example here.

This has caused confusion for customers configuring their email records.

If we want to encourage customers to proxy mail records, we could add a callout for Enterprise Spectrum.

### DIFF
--- a/content/dns/manage-dns-records/how-to/email-records.md
+++ b/content/dns/manage-dns-records/how-to/email-records.md
@@ -24,7 +24,7 @@ To route emails to your mail server, you need to [create two DNS records](/dns/m
 
      | **Type** | **Name** | **IPv4 address** | **Proxy status** |
      | -------- | -------- | ---------------- | ---------------- |
-     | A        | `mail`   | `192.0.2.1`      | Proxied          |
+     | A        | `mail`   | `192.0.2.1`      | DNS-only         |
 
      <details>
       <summary>API example</summary>


content/dns/manage-dns-records/how-to/email-records.md
```diff
@@ -155,4 +155,4 @@ To route emails to your mail server, you need to [create two DNS records](/dns/m
 
 ### Configure email security records
```
hannes-cf, line 156: @Erisa why are you suggesting that we remove this section? I think it is relevant to keep a reference to Email security records on that page.

Erisa: I made no changes to this section, the red is part of GitHub's syntax highlighting rather than part of the diff.
It would show a `-` next to the line if I had modified it (plus the very last line which the web editor touched but made no real changes to)

content/dns/manage-dns-records/how-to/email-records.md
```diff
@@ -24,7 +24,7 @@ To route emails to your mail server, you need to [create two DNS records](/dns/m
 
      | **Type** | **Name** | **IPv4 address** | **Proxy status** |
      | -------- | -------- | ---------------- | ---------------- |
-     | A        | `mail`   | `192.0.2.1`      | Proxied          |
+     | A        | `mail`   | `192.0.2.1`      | DNS-only         |
```
hannes-cf, line 27: I don't think it's necessary to change this to `DNS-only`. As you stated in your PR description, if an MX record is pointed to an orange clouded A (AAAA/CNAME) record we do automatically insert an intermediate record that is grey-clouded so that MX records don't point to Cloudflare anycast IPs. We do this for exactly the purpose so that users don't have to grey cloud the record that they use as target for their MX record.

Erisa: It causes systems relying on PTR to not function correctly, because the PTR of the IP is for `mail.example.com` though the actual MX record is a random one. This caused a lot of frustration for a customer in a Support ticket, which was what prompted the PR.
It also is different to the API example underneath, so at the least they should be unified.

hannes-cf:
> Additionally the API example underneath uses `proxied: false` which is the opposite of the GUI example here.
This is indeed something we should fix. The API call for the A record should be `proxied: true` and the API call for the MX record should omit the `proxied` parameter entirely.

> This has caused confusion for customers configuring their email records.
Can you provide some specific examples (e.g. any community threads)?

> If we want to encourage customers to proxy mail records, we could add a callout for Enterprise Spectrum.
Instead of doing this, we should probably rather add a reference for our Email Security suite (formerly Area 1).

Erisa:
> Can you provide some specific examples (e.g. any community threads)?

It was in support ticket(s). Please contact me internally and I can share details. The gist was confusion regarding PTR as mentioned in one of my review comments.

> Instead of doing this, we should probably rather add a reference for our Email Security suite (formerly Area 1).

 Agreed. I should have thougt of that.

content/dns/manage-dns-records/how-to/email-records.md
hannes-cf: Understood 👍

content/dns/manage-dns-records/how-to/email-records.md
hannes-cf: Interesting point. Okay, then let's change the table to `DNS-only` and the API example to `proxied: false`.

RebeccaTamachiro (approved):

PR #9778, opened 2023-07-07
[Docs Engine] Create shortcodes for tutorial layouts

kodster28:


layouts/shortcodes/tutorial.html
```diff
@@ -0,0 +1,30 @@
...
+        const optional = step.getAttribute("data-optional")
+        let secondSpan = step.children[0].querySelector("span:nth-child(2)");
+        let stepCounter = document.createElement("span");
+        stepCounter.className = "stepCounter";
+        stepCounter.textContent = counter;
+        if (optional === "true") {
+          let optionalFlag = document.createElement("span");
+          optionalFlag.textContent = "Optional - ";
+          secondSpan.prepend(optionalFlag);
+          stepCounter.classList.add("optionalStepCounter")
```
pedrosousa:
```suggestion
          stepCounter.classList.add("optionalStepCounter");
```

layouts/shortcodes/tutorial.html
```diff
@@ -0,0 +1,30 @@
+<div class="tutorialStepContainer">
+  {{- .Inner -}}
+</div>
+
+<script>
+  document.addEventListener("DOMContentLoaded", function() {
+  let tutorial = document.getElementsByClassName("tutorialStepContainer");
+  if (tutorial) {
+    for (const stepContainer of tutorial) {
+      let counter = 1;
```
pedrosousa: Nit: For the step counter, we could eventually use [CSS counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_counter_styles/Using_CSS_counters) instead of JavaScript code, since this seems like a simple use case.

static/style.css
```diff
@@ -5579,3 +5579,49 @@ html #ot-sdk-btn.ot-sdk-show-settings:hover {
...
+  margin-top: 4em;
+  padding-bottom: 1em;
+}
+
+.optionalStepCounter {
+  border: 2px solid gray !important;
+  border-radius: 0% !important;
+}
+
+.stepCounter {
```
pedrosousa, line 5618: Some of these CSS properties are currently being ignored, according to dev tools (the ones below with low opacity), due to the inherited property `display: inline` (which we should probably keep):

image

I believe we can remove them without any impact.

static/style.css
```diff
@@ -5579,3 +5579,49 @@ html #ot-sdk-btn.ot-sdk-show-settings:hover {
 [theme="dark"] #ot-sdk-btn.ot-sdk-show-settings:hover {
   color: var(--code-orange);
 }
+
+.tutorialStep {
```
pedrosousa, line 5583: For some reason the margins seem incorrect on mobile (all the steps are glued together).
I suggest that we add something like the following after the "normal" (non-media bound) styles:

```css
@media (max-width: 768px) {
  .tutorialStep:not(:first-of-type) {
    /* had to increase this margin, but didn't exactly understand why */
    margin-top: 7em;
  }

  .tutorialStep {
    /* using 2em seemed a bit of a waste on mobile */
    padding-left: 1em;
    /* had to increase this margin, but didn't exactly understand why */
    margin-top: 7em;
  }
}
```

I'd test this on real devices, once we have an updated preview — I volunteer an Android device.

pedrosousa (approved):

layouts/shortcodes/tutorial.html
kodster28: addressed in other commit

layouts/shortcodes/tutorial.html
kodster28: Yeah, definitely right. Also less heavy-weight / dependent on JS loading.

PR #9784, opened 2023-07-08
fix: doc bug (maybe)?

barabo:
The example etags `bce6bf66aeb76c7040fdd5f4eccb78e6` and `8165449fc15bbf43d3b674595cbcc406`, when concatenated and md5-hashed do not equal the documented value of `f77dc0eecdebcd774a2a22cb393ad2ff`.  I believe the correct example should be `158668b0d382a2d588dd141dafb5ec6f` instead.

But, I may be misunderstanding the instructions.

Context: https://community.cloudflare.com/t/documentation-bug-with-r2-multipart-upload-md5-example/530466/4

### DIFF
--- a/content/r2/objects/multipart-objects.md
+++ b/content/r2/objects/multipart-objects.md
@@ -26,4 +26,4 @@ The ETags for objects uploaded via multipart are different than those uploaded w
 
 For uploads created after June 21, 2023, R2's multipart ETags now mimic the behavior of S3.  The ETag of each individual part is the MD5 hash of the contents of the part.  The ETag of the completed multipart object is the hash of the MD5 sums of each of the constituent parts concanated together followed by a hyphen and the number of parts uploaded.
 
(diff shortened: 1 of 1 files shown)


bercknash:
It's a misunderstanding on your part.  You need to concatenate and take the md5sum of the actual sum, not the ASCII hex representation.  ie, You did the following: 
```
echo -n bce6bf66aeb76c7040fdd5f4eccb78e68165449fc15bbf43d3b674595cbcc406 | md5sum
158668b0d382a2d588dd141dafb5ec6f  - 
```

In actuality, you need to do something like: 
```
echo -n $(echo -n bce6bf66aeb76c7040fdd5f4eccb78e6 | xxd -r -p -)\
$(echo -n 8165449fc15bbf43d3b674595cbcc406 | xxd -r -p -) | md5sum

f77dc0eecdebcd774a2a22cb393ad2ff  - 
```

I'll see if I can tweak the docs to make this clearer.

dcpena:
Hi @barabo, thanks so much for your contributions! I'm going to close this PR out in favor of PR https://github.com/cloudflare/cloudflare-docs/pull/9797 which now better clarifies how multipart ETags are calculated. As always though, if you find something else, contributions are always welcome and appreciated. 😀

PR #9809, opened 2023-07-11
Clarify that the limitation for Proxy refers to HTTP traffic only

wolruf:
Cloudflare employee here.

The limitation for Proxy traffic isn't obvious that Cloudflare talks about HTTP proxy (not DNS traffic).

### DIFF
--- a/content/dns/zone-setups/zone-transfers/cloudflare-as-secondary/_index.md
+++ b/content/dns/zone-setups/zone-transfers/cloudflare-as-secondary/_index.md
@@ -22,7 +22,7 @@ B & C <--DNS lookups--> D[Resolver] <--DNS lookups--> E((User))
 ## How to
 
 - [Set up incoming zone transfers](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/setup/)
-- Proxy traffic through Cloudflare with [Secondary DNS Override](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic/)
+- Proxy HTTP traffic through Cloudflare with [Secondary DNS Override](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic/)
 
 ## Availability
 


content/dns/zone-setups/zone-transfers/cloudflare-as-secondary/_index.md
```diff
@@ -22,12 +22,12 @@ B & C <--DNS lookups--> D[Resolver] <--DNS lookups--> E((User))
 ## How to
 
 - [Set up incoming zone transfers](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/setup/)
-- Proxy traffic through Cloudflare with [Secondary DNS Override](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic/)
+- Proxy HTTP traffic through Cloudflare with [Secondary DNS Override](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic/)
```
hannes-cf, line 25: While I am not opposed to getting this updated (we probably should use `HTTPS` instead of `HTTP` if we do), I also think this is not super important in the context of "proxying traffic". Here is why:

When you use Cloudflare's nameservers, technically you don't proxy DNS traffic through Cloudflare. As we are authoritative for your domain, we basically are the last hop, the "origin" for DNS traffic / queries so to speak.

We only actually proxy DNS traffic / queries if a customer users DNS Firewall. Using that, a customer keeps their nameservers as upstream / origin and DNS queries (if not cached) are proxied through our network in a very similar way as we do for HTTPS requests.

wolruf: you're right, it's probably that the us of "proxy" without having "proxy" and "HTTPS" attached can be confusing. And that can lead to thinking  we're discussing a sub case of proxying DNS traffic. Not a big issue, just noticed a confused customer once.

RebeccaTamachiro:
Hey! Thanks for raising this.
Will be closing out this one but took note. Will get back to it in case we get other reports of confusion.

PR #9820, opened 2023-07-12
[R2] Fix R2PutOptions types

leaysgur:
It seems that `R2PutOptions.onlyIf` should accept `Headers` like `R2GetOptions.onlyIf` does.

https://github.com/cloudflare/workerd/blob/main/src/workerd/api/r2-bucket.h#L151


Frederik-Baetens (approved):
Thank you for the contribution!

dcpena (approved):

PR #9891, opened 2023-07-16
use `satisfies` for typescript worker handlers

suneettipirneni:
Previously, in order to have an default exports be typechecked in typescript you needed to first create a `const` with an explicit type attached to it and then export that `const`:

```ts
const foo: Foo = { ... }

export default { foo };
```

However, with the introduction of the [`satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator), this is no longer the case.

Instead of using an intermediate `const` we can use default exports directly while also benefitting from typechecking:

```ts
export default {
     ...
} satisfies Foo;
```

This makes the typescript code much closer to its javascript counterpart while also being safer.

This PR changes the examples and some tutorials for workers to use this syntax instead of the `const` syntax.

Similar changes for storybookjs:
- https://github.com/storybookjs/storybook/issues/17016


KianNH:
Neat - out of curiosity, is this different to doing `export default <ExportedHandler> {`?

suneettipirneni:
> Neat - out of curiosity, is this different to doing `export default <ExportedHandler> {`?

Yes, casting via the `<ExportedHandler>` syntax does provide the type inference for parameters `req`, `env` and so on, however, It doesn't check if extraneous properties within the object are allowed and/or are valid. 

For example:
```ts
export default <ExportedHandler> {
		return new Response(JSON.stringify({}, undefined, 2));
	},
	async foo(env: Env) { ... }
};
```
`foo` isn't a valid export property for `ExportedHandler` but the `<ExportedHandler>` syntax allows it anyways.

If we use `satisfies` instead we will get an error, similar to the one we'd get if we used `const handler: ExportedHandler`:
```ts
export default {
		return new Response(JSON.stringify({}, undefined, 2));
	},
 	async foo(env: Env) { ... }
} satisfies ExportedHandler;
```
```
Type '{ fetch(request: Request<unknown, IncomingRequestCfProperties<unknown>>, env: unknown, ctx: ExecutionContext): Promise<Response>; foo(env: Env): Promise<...>; }' does not satisfy the expected type 'ExportedHandler<unknown, unknown, unknown>'.
  Object literal may only specify known properties, and 'foo' does not exist in type 'ExportedHandler<unknown, unknown, unknown>'
```

In addition, `satisfies` performs type checking of the internal object rather than it's assignability like `<ExportedHandler>` does. This means that errors are much more clear when a type is incorrectly used:

Using invalid parameter type with `satisfies`:
Screenshot 2023-07-17 at 1 04 28 PM
```
Type '(request: string, env: unknown, ctx: ExecutionContext) => Promise<Response>' is not assignable to type 'ExportedHandlerFetchHandler<unknown, unknown>'.
  Types of parameters 'request' and 'request' are incompatible.
    Type 'Request<unknown, IncomingRequestCfProperties<unknown>>' is not assignable to type 'string'.ts(2322)
```

Using invalid parameter type with `<ExportedHandler>`:
Screenshot 2023-07-17 at 1 06 15 PM
```
Conversion of type '{ fetch(request: string, env: unknown, ctx: ExecutionContext): Promise<Response>; }' to type 'ExportedHandler<unknown, unknown, unknown>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property 'fetch' are incompatible.
    Type '(request: string, env: unknown, ctx: ExecutionContext) => Promise<Response>' is not comparable to type 'ExportedHandlerFetchHandler<unknown, unknown>'.
      Types of parameters 'request' and 'request' are incompatible.
        Type 'Request<unknown, IncomingRequestCfProperties<unknown>>' is not comparable to type 'string'.ts(2352)
```

deadlypants1973:
Hi @suneettipirneni , thank you for this PR. Were you able to test these examples?

suneettipirneni:
> Hi @suneettipirneni , thank you for this PR. Were you able to test these examples?

Greetings @deadlypants1973, I was able to confirm the typescript examples properly compile.

penalosa:
@suneettipirneni Thanks for this! This will make the TypeScript examples much cleaner. There are a couple restructurings of examples going through the pipeline now, but once those are merged we can get this PR in.

penalosa (approved):

penalosa:
@suneettipirneni could you resolve the conflicts here, and then this should be good to go!

suneettipirneni:
> @suneettipirneni could you resolve the conflicts here, and then this should be good to go!

Hi @penalosa, I've fixed all of the merge conflicts.

penalosa (approved):

deadlypants1973 (approved):

PR #9921, opened 2023-07-18
Replace writeMetrics with writeDataPoint

AlexDicy:
The documentation mentions a `writeMetric` method which is not actually available in the `AnalyticsEngineDataset` binding.

```ts
export interface AnalyticsEngineDataset {
  writeDataPoint(event?: AnalyticsEngineDataPoint): void;
}
```

This replaces the mention of `writeMetric` method with `writeDataPoint`


jimhawkridge (approved):
Wow! I'm surprised this has gone unnoticed for so long!
Thanks for the PR :-)

PR #9953, opened 2023-07-19
[Chore] Upgrade hugo

kodster28:


pedrosousa (approved):
Nit: We should also update the Hugo version in the [README file](https://github.com/cloudflare/cloudflare-docs/blob/production/README.md?plain=1#L11).

PR #9970, opened 2023-07-20
[Version Management] Update enable.md

AlphaK5:
PCX-6387


garrettgalow (approved):

garrettgalow:
based on internal discussion we should add a warning to this that disabling zone versioning will revert your zone configuration to what is configured in version 0 immediately. Ensuring version 0 has the correct settings first is necessary

garrettgalow (changes requested):
see previous comment

kodster28:
Addressed in another PR.

PR #9973, opened 2023-07-20
[CF One] Add best practice advise for secret variables on Terraform documentation in _terraform_input_variables.md

Lp-Francois:
- Secrets passed as variables in terraform should be labelled as sensitive in last TF versions.

TF version with the feature: https://github.com/hashicorp/terraform/blob/v0.14/CHANGELOG.md#0140-december-02-2020

### DIFF
--- a/content/cloudflare-one/_partials/_terraform_input_variables.md
+++ b/content/cloudflare-one/_partials/_terraform_input_variables.md
@@ -61,8 +61,6 @@ The following variables will be passed into your GCP and Cloudflare configuratio
     variable "cloudflare_token" {
       description = "Cloudflare API token created at https://dash.cloudflare.com/profile/api-tokens"
       type        = string
-      ## Terraform 0.14 has added the ability to mark variables as sensitive to avoid leaking secrets in logs.
-      ## uncomment if using Terraform >=0.14
-      # secure    = true
+      sensitive   = true
     }
     ```
--- a/content/cloudflare-one/_partials/_terraform_input_variables.md
+++ b/content/cloudflare-one/_partials/_terraform_input_variables.md
@@ -61,5 +61,8 @@ The following variables will be passed into your GCP and Cloudflare configuratio
     variable "cloudflare_token" {
       description = "Cloudflare API token created at https://dash.cloudflare.com/profile/api-tokens"
       type        = string
+      ## Terraform 0.14 has added the ability to mark variables as sensitive to avoid leaking secrets in logs.
+      ## uncomment if using Terraform >=0.14
+      # secure    = true
     }
     ```


kennyj42 (commented):
@jacobbednarz does this follow our best practice? Looks good to me but wanted to check with you.

jacobbednarz:
`secure` was renamed to `sensitive` in later versions of terraform (so this PR won't work as is) because it was a misnomer about what benefit it actually provided. it is by no way "secure" since will still end up in all of the log sinks for terraform, state files and anything that goes over the wire.

```
$ TF_LOG=DEBUG terraform apply -auto-approve 2>&1 | grep -i secr3t
2023-07-21T15:52:45.434+1000 [DEBUG] provider.terraform-provider-cloudflare_v4.9.0: {"result":{"id":"72fc4fb1580c5715a7b95629490c6bf4","zone_id":"0da42c8d2132a9ddaf714f9e7c920711","zone_name":"terraform.cfapi.net","name":"secr3t.terraform.cfapi.net","type":"CNAME","content":"example.com","proxiable":true,"proxied":false,"ttl":300,"locked":false,"meta":{"auto_added":false,"managed_by_apps":false,"managed_by_argo_tunnel":false,"source":"primary"},"comment":null,"tags":[],"created_on":"2023-07-21T05:49:33.546354Z","modified_on":"2023-07-21T05:49:33.546354Z"},"success":true,"errors":[],"messages":[]}
```

i'm not really concerned either way if we include this because the only gains we get are that people cannot use this directly in `output` blocks and the CLI output itself. i.e.

```hcl
variable "top_secret" {
  default = "secr3t"
}

output "secret_thing" {
  value = var.top_secret
}
```

```
$ terraform apply -auto-approve

Changes to Outputs:
  + secret_thing = "secr3t"

You can apply this plan to save these new output values to the Terraform state, without changing any real infrastructure.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

secret_thing = "secr3t"
```

vs 

```hcl
variable "top_secret" {
  default   = "secr3t"
  sensitive = true
}

output "secret_thing" {
  value = var.top_secret
  # sensitive = true
}
```

```
$ terraform apply -auto-approve
╷
│ Error: Output refers to sensitive values
│ 
│   on terraform.tf line 48:
│   48: output "secret_thing" {
│ 
│ To reduce the risk of accidentally exporting sensitive data that was intended to be only internal, Terraform requires that any root module output containing sensitive data be explicitly marked as sensitive, to confirm your intent.
│ 
│ If you do intend to export this data, annotate the output value as sensitive by adding the following argument:
│     sensitive = true
```
and 
```
$ terraform apply -auto-approve

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # cloudflare_record.mx_1 will be created
  + resource "cloudflare_record" "example" {
      + allow_overwrite = false
      + created_on      = (known after apply)
      + hostname        = (known after apply)
      + id              = (known after apply)
      + metadata        = (known after apply)
      + modified_on     = (known after apply)
      + name            = (sensitive value)    <========= masked
      + proxiable       = (known after apply)
      + ttl             = 300
      + type            = "CNAME"
      + value           = "example.com"
      + zone_id         = "0da42c8d2132a9ddaf714f9e7c920711"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
cloudflare_record.example: Creating...
cloudflare_record.example: Creation complete after 1s [id=72fc4fb1580c5715a7b95629490c6bf4]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

i would almost rather we omit this and instead have a section on storing sensitive values like credentials in a tool designed for secret management (https://blog.cloudflare.com/secrets-store/, vault, etc) as opposed to in terraform itself to avoid the false fallency this may create.

ranbel:
For now I've added the `sensitive` label to match the other variables in this guide. If you'd like to put together a tutorial on secret management tools, we can publish it in the [Terraform docs](https://developers.cloudflare.com/terraform/).

ranbel (approved):

Lp-Francois:
@jacobbednarz Thanks for the valuable comment :D

PR #9986, opened 2023-07-20
fix(pages): don't recommend an old version of wrangler for Pages

Cherry:
This is from when wrangler v1 was a thing.

Pages works fine with wrangler v2 and v3, and with v1 deprecated and barely referenced anywhere more, it's probably safe to just omit this entirely.


WalshyDev (approved):

GregBrimble (approved):

GregBrimble:
Thank you!

PR #9987, opened 2023-07-21
Update docs to use new compatibility_flags

robertaboukhalil:
Updating the docs to use the new compatibility_flags.

Otherwise, it gives errors about Node.js packages not being found.


deadlypants1973 (approved):
PCX

PR #9994, opened 2023-07-21
Fix `Using Kubernetes on GKE and AWS with Cloudflare Load Balancer.md`

ogawa0071:
Fixed a problem with the text being in code style.


kodster28:
Howdy @ogawa0071, thanks for the contribution.

Actually creating a new PR to address this at #10374

PR #10001, opened 2023-07-21
[Stream] Add info to player enhancements API

third774:
* Add context to share_link

* Touched up a couple surrounding areas.


dcpena (changes requested):
Our style guide specifies that we should use "select" vs "click" to be device agnostic. Otherwise, looks good!

third774:
Thanks @dcpena — fixed in a force push!

dcpena (approved):

PR #10002, opened 2023-07-21
Correct grammar in Pages known issues

lfcassidy:
Corrected the grammar by switching "high amount of deployments" to "high number of deployments".


aelnosu:
I think that amount is better

lfcassidy:
> I think that amount is better

"Amount" is grammatically incorrect. See:
https://grammarist.com/usage/amount-number/

deadlypants1973 (approved):
PCX

PR #8594, opened 2023-04-24
Make Pages build infra programatic

GregBrimble:
The only content changes I made were:
- changing "Gatsby.js" -> "Gatsby"
- removing the redundant word "version" in each row of the tools table
- making NPM and pip's phrasing consistent for their default versions

https://programattic-pages-build-ima.cloudflare-docs-7ou.pages.dev/pages/platform/language-support-and-tools/

cc. @jrf0110 @WalshyDev

PR #10004, opened 2023-07-21
[Pages] Create docs API for build presets and read in build-configuration.md

cmsparks:
The only changes to docs content is in build-configuration.md, but the content that users see should be identical to before. The table is pulled out into a separate shortcode `pages-build-presets-table.html` which reads from the `_build-configuration.md`. This PR allow us to keep dashboard presets in sync with the docs similar to https://github.com/cloudflare/cloudflare-docs/pull/8594.

cc: @jrf0110 @GregBrimble @WalshyDev

### DIFF
--- a/content/pages/_partials/_build-configuration.md
+++ b/content/pages/_partials/_build-configuration.md
@@ -1,0 +1,120 @@
+---
+_build:
+  publishResources: false
+  render: never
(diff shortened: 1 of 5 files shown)


cmsparks:
Closing and recreating to get preview builds

PR #10005, opened 2023-07-21
[Pages] Create docs API for build presets and read in build-configuration.md

cmsparks:
The only changes to docs content is in `build-configuration.md`, but the content that users see should be identical to before. The table is pulled out into a separate shortcode `pages-build-presets-table.html` which reads from the `_build-configuration.md`. This PR allow us to keep dashboard presets in sync with the docs similar to https://github.com/cloudflare/cloudflare-docs/pull/8594.

cc: @jrf0110 @GregBrimble @WalshyDev


cmsparks:
Fixed spell check (the remaining failing spell check is correctly spelled) and here's the relevant link: https://csparks-programmatic-build-p.cloudflare-docs-7ou.pages.dev/pages/platform/build-configuration/

WalshyDev (approved):

cmsparks:
note: realized I should have the docs serve the icons so we don't need to change icons on the dashboard

content/pages/_partials/_build-configuration.md
```diff
@@ -0,0 +1,148 @@
...
+
+build_configs:
+  create-react-app:
+    display_name: Create React App
+    build_command: npm run build
+    build_output_directory: build
+    icon: /icons/framework-icons/logo-react.svg
+  gatsby:
+    display_name: Gatsby
+    build_command: gatsby build
```
jrf0110: For any globals like this, can we change to use `npx` instead of assuming that the `gatsby` is installed globally (it is not)?

```suggestion
    build_command: npx gatsby build
```

layouts/_default/build-configuration.json.json
```diff
@@ -0,0 +1,2 @@
+{{- $file := $.Site.GetPage "content/pages/_partials/_build-configuration.md" -}}
```
jrf0110, line 1: I know this is an existing pattern, but do you know why we name these files with `.json.json`?

jrf0110 (approved):

deadlypants1973 (approved):
PCX

layouts/_default/build-configuration.json.json
cmsparks: Honestly, not sure. I'll have to ask Greg, but I'm going to keep it as is just to maintain the pattern for now.

cmsparks: Honestly not too sure, I'd have to ask @GregBrimble. I'm going to keep it as is for now

kodster28 (approved):
as long as the pages devs are happy with this ✅

PR #10026, opened 2023-07-24
Clarify Cloudflare implementation of .tee()

### LINKED ISSUE #9285: Add documentation for ReadableStream.tee()

Workers

https://developers.cloudflare.com/workers/runtime-apis/streams/readablestream/

The "Methods" section

### What needs to change?

A new section for the `tee()` method.

### How should it change?

The `ReadableStream` type has a `tee()` method (which is well-documented in other places, like the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee) docs).

The MDN docs highlight potential for unbounded memory usage, in case one consumer is much slower than the other consumer.

However, Cloudflare's implementation of `ReadableStream` has special support for this case, as documented in [this README](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/streams/README.md#what-about-tees).  This README is unfortunately hard to discover (thanks to @KianNH who pointed it out to me).  

It would be nice if the [methods section](https://developers.cloudflare.com/workers/runtime-apis/streams/readablestream/#methods) of the docs included this important info about the specific Cloudflare implementation of `tee()`

### Additional information

_No response_

irvinebroque:
refs https://github.com/cloudflare/cloudflare-docs/issues/9285

### DIFF
--- a/content/workers/runtime-apis/streams/readablestream.md
+++ b/content/workers/runtime-apis/streams/readablestream.md
@@ -34,6 +34,64 @@ A `ReadableStream` is returned by the `readable` property inside [`TransformStre
 let reader = readable.getReader({ mode: 'byob' });
 ```
 
+
+- {{<code>}}tee(){{</code>}}
+
+The `ReadableStream` API has a method `tee()` that will split the flow of data from the
+`ReadableStream` into two separate `ReadableStream` instances.
+
+In the [standard definition](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee) of the `ReadableStream` API, the `tee()` method creates two
+separate `ReadableStream` instances (called "branches") that share a single `Reader` that
+consumes the data from the original `ReadableStream` (let's call it the "trunk"). When one
+of the two branches uses the shared `Reader` to pull data from the trunk, that data is
+used to fulfill the read request from the pulling branch, and a copy of the data is pushed
+into a queue in the other branch. That copied data accumulates in memory until something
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/streams/readablestream.md
```diff
@@ -34,6 +34,64 @@ A `ReadableStream` is returned by the `readable` property inside [`TransformStre
...
+separate `ReadableStream` instances (called "branches") that share a single `Reader` that
+consumes the data from the original `ReadableStream` (let's call it the "trunk"). When one
+of the two branches uses the shared `Reader` to pull data from the trunk, that data is
+used to fulfill the read request from the pulling branch, and a copy of the data is pushed
+into a queue in the other branch. That copied data accumulates in memory until something
+starts reading from it.
+
+This spec defined behavior presents a problem for us in that it is possible for one branch
+to consume data at a far greater pace than the other, causing the slower branch to accumulate
+data in memory without any backpressure controls.
```
irvinebroque, line 53: todo: reframe to be less about us and more about what someone using this API needs to know

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee

_If only one branch is consumed, then the entire body will be enqueued in memory. Therefore, you should not use the built-in tee() to read very large streams in parallel at different speeds. Instead, search for an implementation that fully backpressures to the speed of the slower consumed branch._

jasnell (approved):

KimJ15:

PR #10056, opened 2023-07-26
[support] Billing policy clarification

kodster28:


ngayerie (approved):

ngayerie:
Just a detail @kodster28 though: shouldn't we say Coordinated Universal Time instead of Universal Time Coordinated?
https://en.wikipedia.org/wiki/Coordinated_Universal_Time

PR #10081, opened 2023-07-27
[API Shield] Schema Validation 2 updates

patriciasantaana:


content/api-shield/security/schema-validation/configure.md
```diff
@@ -223,7 +223,7 @@ header: Result
 }
 ```
 
-You can add all operations in a schema that do not already exist in Endpoint Management by combining two commands as one. The example requires the `jq` tool.
+You can add all operations in a schema that do not already exist in Endpoint Management by combining two commands as one. There is a maximum of 20 operations for this API call. The example requires the `jq` tool.
```
pedrosousa, line 226: Nit: We recommend that examples do not require `jq`, but at least it's listed here as a requirement.

content/api-shield/security/schema-validation/configure.md
```diff
@@ -369,13 +369,15 @@ header: Result
...
+You can get an overview of the schemas currently active on a zone using `GET`. 
+
+`validation_enabled=true` is an optional parameter.
 
 ```bash
 ---
 header: cURL command
 ---
-$ curl "https://api.cloudflare.com/client/v4/zones/{zone_id}/api_gateway/user_schemas?validation_enabled=true&omit_source=true" \
+$ curl "https://api.cloudflare.com/client/v4/zones/{zone_id}/api_gateway/user_schemas?omit_source=true" \
```
pedrosousa: Ideally, you would update the command according to our latest `curl` guidelines (use long parameter names, no `$` prefix), but it would become different from other commands on this page. Maybe on a separate PR?

For this command, it would look like the following:
```
curl "https://api.cloudflare.com/client/v4/zones/{zone_id}/api_gateway/user_schemas?omit_source=true" \
--header "Authorization: Bearer <API_TOKEN>" \
--header "Content-Type: application/json"
```

pedrosousa: Should we also remove `validation_enabled=true` from the aside below the result (line 410)?

pedrosousa (approved):

PR #10110, opened 2023-07-28
Update Partial Zone Wildcard certificate via Universal SSL

jedecf:
@rprice74 noted to me an edit required to Universal SSL & Partial zone wildcard certification, in that, people now require an Advanced certificate for wildcard if using a partial.

Needs confirmation before publishing.

### DIFF
--- a/content/ssl/_partials/_partial-zone-acm-dcv-wildcard.md
+++ b/content/ssl/_partials/_partial-zone-acm-dcv-wildcard.md
@@ -5,7 +5,7 @@ _build:
   list: never
 ---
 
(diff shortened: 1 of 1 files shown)


kodster28:
Just calling this out so it's clear... this'd require confirmation then from ENG or PM. PCX doesn't provide technical sign-off.

RebeccaTamachiro:
Hey! Thanks for bringing this up.
As this turns out to be a more complex case, I've created PCX-9042 to investigate further. Will be closing this PR for now and open a new one once this topic has been clarified with PM.

PR #10158, opened 2023-08-01
Web Analytics requires connect-src 'self'

mgreensmith:
Needed to connect to the  `/cdn-cgi/rum` endpoint.


kodster28:
Howdy @mgreensmith, checked with some internal folks and -- according to them -- the `self` directive isn't necessary b/c it's actually loading from a different origin entirely (`cloudflareinsights.com`).

mgreensmith:
Right, the script is fetched from your insights domain, but this PR is about destinations where the page should be allowed to send requests _to_ (CSP `connect-src`), not about where the script itself is fetched _from_ (CSP `script-src`).

When your analytics script is running on the page, it sends periodic analytics events to your [internal /cdn-cgi/ endpoint](https://developers.cloudflare.com/fundamentals/get-started/reference/cdn-cgi-endpoint/). This internal endpoint is exposed on the proxied website directly, so from the browser's perspective, it is connecting to `self`.

Please re-discuss internally; this  rule is definitely necessary to make web analytics work on a proxied domain (and it states as much on your doc that I linked).

PR #10161, opened 2023-08-01
[HOLD][R2] Updated content for bucket scoped tokens

dcpena:
Updated docs for bucket scoped tokens to address PCX-7767.

### DIFF
--- a/content/r2/api/s3/tokens.md
+++ b/content/r2/api/s3/tokens.md
@@ -36,7 +36,7 @@ Find your [account ID in the Cloudflare dashboard](/fundamentals/get-started/bas
 
 | Permission | Description |
 |------------|-------------|
-| Admin Read and Write | Allows the ability to create, list, modify, and delete buckets and their associated configurations. This includes the ability to write, read, and list objects within buckets. |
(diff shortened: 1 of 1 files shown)


content/r2/api/s3/tokens.md
```diff
@@ -29,3 +31,12 @@ You will not be able to access your **Secret Access Key** again after this step.
 The S3 endpoint is available via the `https://<ACCOUNT_ID>.r2.cloudflarestorage.com` endpoint. 
 
 Find your [account ID in the Cloudflare dashboard](/fundamentals/get-started/basic-tasks/find-account-and-zone-ids/).
+
+## Permissions
+
```
kodster28, line 36: I'm a touch confused by this table. Are all of these required.... b/c I'd think read and write covers the read only permissions?

Or is it that any of them work?

kodster28 (approved):

content/r2/api/s3/tokens.md
dcpena: Ah, good callout. They all read the same when in reality, the distinguishing factor is the Administrator option. Each option is available though so let me double-check on the Administrator option.

PR #10172, opened 2023-08-02
Update load-custom-managed-component.md

bjesus:
these features are no longer unsupported


bjesus:
@kodster28 is there anyway to skip this "simple" test? I understand the logic behind it but in this case it feels justified. and also it doesn't have anything to do with what I changed... 🫣

haleycode (approved):

haleycode:
> 

The spell checker is non-blocking, so we can go ahead and merge even if it flags something (provided it's not actually an error).

PR #10200, opened 2023-08-03
[Stream] Expiring uploads

dcpena:
Updated different pages to include information about `scheduledDeletion`. Addresses PCX-7667.

### DIFF
--- a/content/stream/stream-live/start-stream-live.md
+++ b/content/stream/stream-live/start-stream-live.md
@@ -80,6 +80,12 @@ header: Response
 
   - The `requireSignedURLs` property indicates if signed URLs are required to view the video. This setting is applied by default to all videos recorded from the input. In addition, if viewing a video via the live input ID, this field takes effect over any video-level settings.
 
+- `scheduledDeletion` {{<type>}}string{{</type>}} {{<prop-meta>}}default: `null` (any){{</prop-meta>}}
+
+  - Specifies a date and time when a video is deleted. After a video is deleted, it is no longer viewable and no longer counts towards storage for billing. Minimum value is `30`.
+
+  Note that if the value is added to a live input while a stream is live, the property will only apply to future streams. 
+
 - `allowedOrigins` {{<type>}}integer{{</type>}} {{<prop-meta>}}default: `null` (any){{</prop-meta>}}
 
   - The `allowedOrigins` property can optionally be invoked to provide a list of allowed origins. This setting is applied by default to all videos recorded from the input. In addition, if viewing a video via the live input ID, this field takes effect over any video-level settings.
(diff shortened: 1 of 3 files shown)


content/stream/stream-live/start-stream-live.md
```diff
@@ -80,6 +80,12 @@ header: Response
 
   - The `requireSignedURLs` property indicates if signed URLs are required to view the video. This setting is applied by default to all videos recorded from the input. In addition, if viewing a video via the live input ID, this field takes effect over any video-level settings.
 
+- `scheduledDeletion` {{<type>}}string{{</type>}} {{<prop-meta>}}default: `null` (any){{</prop-meta>}}
```
austin-mc: For live, the property is `deleteRecordingAfterDays` and its an integer value. Might be worth clarifying that the recording is deleted, not the input itself, and the value applies from when the stream recording is made viewable/ready to stream.

content/stream/uploading-videos/upload-video-file.md
```diff
@@ -71,6 +71,10 @@ Setting arbitrary metadata values in the `Upload-Metadata` header sets values th
 
   - If this key is present, the video playback for this video will be required to use signed urls after upload.
 
+- `scheduledDeletion`
+
+  - Specifies a date and time when a video is deleted. After a video is deleted, it is no longer viewable and no longer counts towards storage for billing.
```
austin-mc: Looks good, think it may be worth mentioning the minimum of 30 days from creation

content/stream/stream-live/start-stream-live.md
austin-mc: When the stream ends, a `scheduledDeletion` timestamp is calculated using the `deleteRecordingAfterDays` value (if present)

content/stream/stream-live/start-stream-live.md
tsmith512, line 1: "Specifies a date and time when a video is deleted."

Should we adjust to "when a video _will be_ deleted". Also ultimately (post rollout), deleted videos are actually deleted and unrecoverable. Is that clear here?

content/stream/uploading-videos/upload-video-file.md
tsmith512, line 1: Same comment as above here

tsmith512 (commented):
Looks solid, just two points to maybe take another look at, but might be okay as-is, too.

content/stream/uploading-videos/upload-video-file.md
dcpena: Done!

content/stream/stream-live/start-stream-live.md
dcpena: Ah, thanks for the callout. Fixed

content/stream/stream-live/start-stream-live.md
dcpena: Changed language to future tense. 

Legitimate question, does saying the videos are no longer viewable not convey that videos can't be recovered? Happy to add that part in if it could potentially confuse users.

patriciasantaana (approved):

PR #10207, opened 2023-08-04
[Rules] Small fixes for Bulk Redirects

pedrosousa:
Addresses PCX-7692.


content/rules/url-forwarding/bulk-redirects/faq.md
```diff
@@ -16,6 +16,18 @@ To troubleshoot runtime errors related to Bulk Redirects, refer to [Troubleshoot
 
 In this situation, Cloudflare will use the URL redirect of the first rule that triggers. This will be determined by the order of the Bulk Redirect Rules enabling each Bulk Redirect List in the `http_request_redirect` phase entry point ruleset.
 
+## I get an error when uploading a CSV list of redirects to a list. How can I solve this issue?
```
orium: Not just when uploading a CSV list.  Also, I would add the specific error to this title, since the answer is just for the "limit on the number of URL matching items on the same hostname/path" problem.

orium (approved):

patriciasantaana (approved):

PR #10232, opened 2023-08-07
Laurry gee patch 1

Laurry-gee:


pedrosousa:
Hi @Laurry-gee 
Unless otherwise noted, we only accept external contributions to the documentation itself (the content), not to other files like CODEOWNERS or their backups. I'll close this ticket.

PR #10255, opened 2023-08-07
[R2] Mark object tagging APIs as unsupported

sdnts:
Also marks ListMultipartUploads as "supported"


content/r2/api/s3/api.md
```diff
@@ -6,7 +6,7 @@ weight: 1
 
 # S3 API compatibility
 
-R2 implements the S3 API to allow users and their applications to migrate easily. When comparing to AWS S3, Cloudflare has removed some API operations' features and added others. The S3 API operations are listed below with their current implementation status. Feature implementation is currently in progress. Refer back to this page for updates.
+R2 implements the S3 API to allow users and their applications to migrate with ease. When comparing to AWS S3, Cloudflare has removed some API operations' features and added others. The S3 API operations are listed below with their current implementation status. Feature implementation is currently in progress. Refer back to this page for updates.
```
sdnts, line 9: Changed this because the inclusive language test was failing (presumably because it looks at changed files and not hunks)

dcpena (approved):

content/r2/api/s3/api.md
dcpena: Thanks for the attention to detail with this edit, Sid! for future reference, that check is non-blocking, but again, I appreciate you making the update to get it to pass. 😄

PR #10257, opened 2023-08-07
Update changelog.md with release

dom96:


content/workers/platform/changelog.md
```diff
@@ -8,6 +8,10 @@ rss: file
 
 # Changelog
 
+## 2023-08-07
+
+- No externally-visible changes this release.
+
```
irvinebroque, line 14: I think we can omit a changelog entry when there aren't developer-facing changes to call out. Thoughts?

dom96: Perhaps, though for some it may still be nice to know that something changed in case they are seeing a difference in how their worker behaves. That way they can tie it to the release.

But I don't have strong feelings. Whatever we decide we should document it in our wiki. I mainly created this because I saw other changelog entries with no externally visible changes.

irvinebroque: sorted internally

Issue #10074, opened 2023-07-27
Add XGBoost documentation on Constellation docs


GrandioseGuru:
### Which Cloudflare product does this pertain to?

Constellation

### Existing documentation URL(s)

https://developers.cloudflare.com/constellation

### What changes are you suggesting?

More than one month ago, Cloudflare announced that Constellation could use XGBoost models and they even announced that the developers documentation was updated to reflect that, but currently it still shows  ONNX as the only available runtime and it only mentions that XGBoost is in the final stages of testing.

https://blog.cloudflare.com/globally-distributed-ai-and-a-constellation-update/

I have Constellation with XGBoost enabled, but I'm not being able to use it because of this missing documentation. Please update it with at least a small example or point me in the right direction on how to use it and I could create it. Thanks in advance.

PR #10263, opened 2023-08-08
Documentation for the new Constellation API

celso:


kodster28:
Howdy @celso, does this also close #10074? Or would that be addressed in a future PR?

rita3ko (approved):
approved!

celso:
@kodster28 It does

PR #10264, opened 2023-08-08
replace manual setup steps with c3 in the the astro framework guide

pdwittig:
### What
This is a WIP to update the framework guides to use C3 instead of manually creating + configuring projects.

Changes that were made:
 - Update core path to create new project using C3.
 - Remove steps that are no longer necessary.
 - Add callout + link to new section (not core path) that handles migrating existing projects 
 
### Why
Improve the DX. C3 bootstraps a lot of the manual config steps for various frameworks, and should be the default way to create new projects for the frameworks it supports.

### Feedback ask
This is a WIP and only contains an update to Astro docs. Looking for general feedback. Once feedback is addressed, I'll update all other frameworks that C3 supports.

Framework checklist:
- [ ] Angular
- [x] Astro
- [ ] Docusaurus
- [ ] Gatsby
- [ ] Hono
- [ ] Next
- [ ] Nuxt
- [ ] Qwik
- [ ] React
- [ ] Remix
- [ ] Solid
- [ ] Svelte
- [ ] Vue

### DIFF
--- a/content/pages/framework-guides/deploy-an-astro-site.md
+++ b/content/pages/framework-guides/deploy-an-astro-site.md
@@ -11,38 +11,40 @@ Refer to the [Astro Docs](https://docs.astro.build/) to learn more about Astro o
 
 In this guide, you will create a new Astro application and deploy it using Cloudflare Pages.
 
-## Setting up a new project
 
-Create a new project directory and then initiate Astro's official setup tool by running:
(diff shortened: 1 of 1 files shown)


admah:
@pdwittig this may be a duplicate of #10357

kodster28:
Think we can close this out. If needed, feel free to open another PR making the change.

PR #10270, opened 2023-08-08
added additional note for enterprise customers

s-orodeji:
added additional information for enterprise customers who want to delete a zone


kodster28:
Addressed in #10305

PR #10272, opened 2023-08-08
Specify HC retention period

cicku:


data/plans.json
```diff
@@ -2603,9 +2603,9 @@
         "analytics": {
           "title": "Analytics",
```
victor-perov: If the idea to show how far back in time each plan can request data, then title is misleading. I would use something like "Maximum time range" or something

victor-perov (changes requested):
Could we please rename that field and put appropriate title to avoid confusion?

data/plans.json
cicku: @angelampcosta Need your opinion on how we manage the comparison here. If you are ok with the title then I will do it.

angelampcosta: Sounds good to me. We can go with Viktor's suggestion.

angelampcosta (approved):

angelampcosta:
Hi @victor-perov The PR was updated. Could you please review? Thanks!

victor-perov:
@angelampcosta @cicku 

Thanks for pinging me on this ticket.

What should we do in this case with lines 2626-2633 of the same config? Where we do use "analytics" as availability for LoadBalancing analytics.

In my understanding you're trying to update/alter meaning of the object attributed without updating all use-cases of it.

victor-perov (changes requested):
I think we should either get a linked PR / issue to get the same applied (see my comment above) for LB Analytics or change it as part of this PR too.

PR #10277, opened 2023-08-08
[LB] Update http host header override documentation for load balancer

Ltadrian:
Updating load balancing documentation to address correct behavior of host header override for Origin Rule and Load Balancing


RebeccaTamachiro (approved):
Hi Adrian. Thanks for the PR!
I've approved it. Just wanted to confirm if I can merge it already.
This is the behaviour that is already in place and we just needed to fix the docs, is that right?

Ltadrian:
Yes exactly this is behavior that was reported by customer and we have tested it to be true. So we just need to update the documentation. This can be merged, thank you!

PR #10215, opened 2023-08-04
[Workers] runtime api external links for R2, DO

deadlypants1973:
PCX-7940

PR #10281, opened 2023-08-08
[Dev plat] Update pages used purely as redirects

kodster28:
This still maintains a placeholder in the side nav... but removes the file references from our sitemap, which is helpful for both reporting + SEO purposes.


content/workers/runtime-apis/websockets/use-websockets.md
```diff
@@ -180,4 +180,4 @@ Without this compatibility flag, the Workers runtime will strip or ignore the `S
...
+If your application needs to coordinate among multiple WebSocket connections, such as a chat room or game match, you will need to create a Durable Object so clients send messages to a single-point-of-coordination. Durable Objects are a coordinated state tool for Cloudflare Workers, which are often used in parallel with WebSockets to persist state over multiple clients and connections. Refer to [Durable Objects](/durable-objects/) to get started, and prefer using the Durable Objects [WebSockets Hibernation API](/durable-objects/api/hibernatable-websockets-api/) rather than the `.accept` method described above.
```
Maddy-Cloudflare, line 183:
```suggestion
If your application needs to coordinate among multiple WebSocket connections, such as a chat room or game match, you will need to create a Durable Object so clients send messages to a single-point-of-coordination. Durable Objects are a coordinated state tool for Cloudflare Workers, which are often used in parallel with WebSockets to persist state over multiple clients and connections. Refer to [Durable Objects](/durable-objects/) to get started, and prefer using the Durable Objects [Hibernatable WebSockets API](/durable-objects/api/hibernatable-websockets-api/) rather than the `accept()` method described above.
```

Maddy-Cloudflare (approved):

kodster28:
Already addressed by #10215

kodster28:
will start a new PR

PR #10356, opened 2023-08-15
Document GraphQL API's new support for arrays of primitives

46bit:
As widely requested, GraphQL API now supports arrays of primitive types. This will be used for things like lists of tags.

This commit adjusts the existing documentation. It adds a list of the new array operators: `_has`, `_hasall` and `_hasany`. We could maybe do with usage examples for all 3, but the existing documentation just offers a very quick summary.

It also adjusts some documentation for Nested arrays, to try and disambiguate between the limited filtering of those fields vs the more featured filtering of the new arrays of primitives.

### DIFF
--- a/content/analytics/graphql-api/features/filtering/index.md
+++ b/content/analytics/graphql-api/features/filtering/index.md
@@ -82,9 +82,19 @@ filters
 
 Operator support varies, depending on the node type and node name.
 
-#### Common operators
+#### Array operators
 
-The following operators are supported for all types:
+The following operators are supported for all array types:
+
+| Operator | Comparison                                      |
+| -------- | ----------------------------------------------- |
+| `has`    | array contains a value                          |
+| `hasall` | array contains all of a list of values          |
+| `hasany` | array contains at least one of a list of values |
+
+#### Scalar operators
+
+The following operators are supported for all scalar types:
 
 | Operator | Comparison          |
 | -------- | ------------------- |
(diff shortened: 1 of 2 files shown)


content/analytics/graphql-api/features/filtering/index.md
```diff
@@ -172,6 +182,25 @@ httpRequestsAdaptiveGroups(
 WHERE datetime="2018-01-01T10:00:00Z"
   AND ((clientCountryName = "US") OR (clientCountryName = "GB"))
 ```
+
+### Filter on an array node
+
+The following GraphQL example shows how to filter an array. The SQL equivalent follows.
```
victor-perov: I think it would be very useful to demonstrate usage of `hasall` for a list of values

victor-perov (approved):

bharatnc (approved):

content/analytics/graphql-api/features/nested-structures/index.md
```diff
@@ -11,7 +11,12 @@ Two kinds of nested structures that behave in special ways are supported: **arra
 
 ## Arrays
 
-Arrays behave as a special kind of single value. There is no way to paginate through, filter, filter by, group, or group by the array.
+The GraphQL API supports two different sorts of arrays:
+
+- Some arrays contain scalar types (e.g. `[String]`) and function like ordinary fields that [can be filtered](/analytics/graphql-api/features/filtering/)
```
angelampcosta:
```suggestion
- Some arrays contain scalar types (for example, `[String]`) and function like ordinary fields that [can be filtered](/analytics/graphql-api/features/filtering/)
```

content/analytics/graphql-api/features/nested-structures/index.md
```diff
@@ -11,7 +11,12 @@ Two kinds of nested structures that behave in special ways are supported: **arra
...
+- Some arrays contain more complex types (e.g. `[Subrequest]`.) The following section describes their behaviour.
```
angelampcosta:
```suggestion
- Some arrays contain more complex types (for example, `[Subrequest]`.) The following section describes their behaviour.
```

content/analytics/graphql-api/features/filtering/index.md
```diff
@@ -172,6 +182,45 @@ httpRequestsAdaptiveGroups(
 WHERE datetime="2018-01-01T10:00:00Z"
   AND ((clientCountryName = "US") OR (clientCountryName = "GB"))
 ```
+
+### Filter an array by one value
+
+The following GraphQL examples show how to filter an array field to only return data
+that include a specific value. The SQL equivalent follows.
```
angelampcosta:
```suggestion
that includes a specific value. The SQL equivalent follows.
```

angelampcosta: Although data is a singular noun, it is used with both singular and plural verbs. Since bellow we use "includes", I think we can also use it here.

46bit:
Updated to apply suggestions.

I've also pushed a commit that works around the language policing of the spell check CI. It's been rejecting the PR because of things that are only problematic in some contexts that don't apply here.

angelampcosta:
Thanks @46bit Just FYI the spell check it's not blocking. If the suggestion doesn't apply, you can just ignore it.

angelampcosta (approved):

PR #10378, opened 2023-08-16
[ZT] Update ssh-logging prereq

ranbel:


kennyj42 (changes requested):
We actually don't need to take this step here. In step 2 of this guide, we talk about how to create the Short Lived Certificate for the user via an API command. The SSH Server connection isn't technically required.

ranbel:
> We actually don't need to take this step here. In step 2 of this guide, we talk about how to create the Short Lived Certificate for the user via an API command. The SSH Server connection isn't technically required.

So you don't need to setup a Tunnel?

kennyj42:
Nope, you only need to do that if you want to proxy it over a private IP address. If you make it available over a public IP then we'll do public key authentication to the public IP.

PR #10389, opened 2023-08-17
[Workers] npx wrangler deploy

deadlypants1973:
PCX-8105


content/workers/configuration/deployments.md
```diff
@@ -35,23 +35,23 @@ Changing triggers such as routes, custom domains, or cron triggers will not issu
 
 ## Interacting with deployments
```
kodster28, line 36: Longer-term, might be worth switching these to the [tabs component](https://github.com/cloudflare/cloudflare-docs/blob/production/content/load-balancing/how-to/create-pool.md?plain=1#L21).

kodster28 (approved):

PR #10399, opened 2023-08-17
[Stream] Added note about Chromecast limitations

dcpena:
Added note about Chromecast limitations. Addresses PCX-6925.

### DIFF
--- a/content/stream/_partials/_chromecast_limitations.md
+++ b/content/stream/_partials/_chromecast_limitations.md
@@ -7,6 +7,6 @@ _build:
 
 {{<Aside type="note">}}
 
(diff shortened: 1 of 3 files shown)


bkrebsbach (approved):

content/stream/_partials/_chromecast_limitations.md
```diff
@@ -0,0 +1,12 @@
+---
+_build:
+  publishResources: false
+  render: never
+  list: never
+---
+
+{{<Aside type="note">}}
+
+Due to Google Chromecast limitations, Chromecast does not support audio and video delivered separately. To avoid potential issues with playback, we recommend using DASH, instead of HLS, which is a Chromecast supported usecase.
```
kodster28:
```suggestion
Due to Google Chromecast limitations, Chromecast does not support audio and video delivered separately. To avoid potential issues with playback, we recommend using DASH, instead of HLS, which is a Chromecast supported use case.
```

kodster28 (approved):

PR #10437, opened 2023-08-22
Update 4xx Client Error.md

ngayerie:


ngayerie:
Hi @rprice74 @zeinjaber 
I've updated this article with the content from our macro. Is that ok?

zeinjaber:
@ngayerie LGTM in general. I'm recommending some small changes as follows. If you're okay with these suggestions, please update the PR, and I'll merge it.

To provide more context, a TCP connection must be established between Cloudflare and the website’s Origin server before any higher protocol (in this case is HTTP) started the “conversation”. In order to establish a connection, TCP uses a three-way handshake :

* SYN: Cloudflare sends three SYN packets to the origin server.
* SYN+ACK: In response, the origin server sends back an SYN+ACK packet.
* ACK: Finally, Cloudflare sends an ACK back to the origin server.

At this point, both Cloudflare and the origin server have received an acknowledgment of the connection, and communication is established. However, if the origin server does not send an SYN+ACK back to Cloudflare within 15 seconds, Cloudflare will retry one more time (another 15 seconds timeout).
Therefore, based on the timeout value set on the client's side, three different scenarios will occur, each resulting in its own specific status code:

* If the client has a shorter timeout (less than 30s), the client will give up the connection attempt, leading to Cloudflare issuing the 499 error.
* If the client has a higher timeout (more than 30s), and once the TCP connection has been successfully established, the HTTP transaction will be continued as per normal. In this scenario, Cloudflare returns a normal status code (HTTP 200).
* If the client has a higher timeout, and Cloudflare fails to establish the TCP handshake with the Origin server, an HTTP 522 status code will be returned.

zeinjaber (approved):

PR #10459, opened 2023-08-24
Clarify use of TCP Sockets in local dev of Pages projects

irvinebroque:


irvinebroque:
Don't think this is a reliable enough path to point towards, better answer is polyfill using https://github.com/Ethan-Arrowood/socket soon.

PR #10462, opened 2023-08-24
Add node:crypto to the changelog

irvinebroque:


kodster28:
If/when this change gets merged, you'll want to follow the new format (adding entries to the [YAML file](https://github.com/cloudflare/cloudflare-docs/blob/production/data/changelogs/workers.yaml)).

PR #10494, opened 2023-08-29
Update upstash.md for Redis trademark compliance

enesakar:
Redis.com asks referring their trademark notice when `Redis` is used to protect their trademark. 

Added the related note the bottom upon their request.


deadlypants1973 (approved):
Approved with @tanushreesharma-cf

PR #10501, opened 2023-08-30
Document Cloudflare Pages ignoring ports when matching _headers rules

GregBrimble:
Cloudflare Pages is going to ignore the incoming request's port when evaluating `_headers` rules to match against. It already ignores the protocol.

### DIFF
--- a/content/pages/platform/headers.md
+++ b/content/pages/platform/headers.md
@@ -22,7 +22,7 @@ Header rules are defined in multi-line blocks. The first line of a block is the 
   [name]: [value]
 ```
 
-Using absolute URLs is supported, though be aware that they must begin with `https` and specifying a port is not supported. Cloudflare Pages ignores the incoming request's port and protocol when matching against an incoming request. For example, a rule like `https://example.com/path` would match against requests to `other://example.com:1234/path`.
+Using absolute URLs is supported, though be aware that absolute URLs must begin with `https` and specifying a port is not supported. Cloudflare Pages ignores the incoming request's port and protocol when matching against an incoming request. For example, a rule like `https://example.com/path` would match against requests to `other://example.com:1234/path`.
 
 You can define as many `[name]: [value]` pairs as you require on subsequent lines. For example:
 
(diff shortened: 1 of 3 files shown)


content/pages/platform/headers.md
```diff
@@ -22,6 +22,8 @@ Header rules are defined in multi-line blocks. The first line of a block is the
   [name]: [value]
 ```
 
+Using absolute URLs is supported, though be aware that they must begin with `https` and specifying a port is not supported. Cloudflare Pages ignores the incoming request's port and protocol when matching against an incoming request. For example, a rule like `https://example.com/path` would match against requests to `other://example.com:1234/path`.
```
deadlypants1973:
```suggestion
Using absolute URLs is supported, though be aware that absolute URLs must begin with `https` and specifying a port is not supported. Cloudflare Pages ignores the incoming request's port and protocol when matching against an incoming request. For example, a rule like `https://example.com/path` would match against requests to `other://example.com:1234/path`.
```

content/pages/platform/headers.md
```diff
@@ -87,7 +89,7 @@ The matched value can be referenced within the header value as the `:splat` plac
 
 #### Placeholders
 
-A placeholder can be defined with `:placeholder_name`. A colon (`:`) indicates the start of a placeholder and the placeholder name that follows must be composed of alphanumeric characters and underscores (`:\w+`). Every named placeholder can only be referenced once. Placeholders match all characters apart from the delimiter, which when part of the host, is a period (`.`) or a forward-slash (`/`) and may only be a forward-slash (`/`) when part of the path.
+A placeholder can be defined with `:placeholder_name`. A colon (`:`) followed by a letter indicates the start of a placeholder and the placeholder name that follows must be composed of alphanumeric characters and underscores (`:[A-Za-z]\w*`). Every named placeholder can only be referenced once. Placeholders match all characters apart from the delimiter, which when part of the host, is a period (`.`) or a forward-slash (`/`) and may only be a forward-slash (`/`) when part of the path.
```
deadlypants1973: We have this on 2 pages, it should be a partial to avoid future inconsistencies.

deadlypants1973 (approved):

PR #10503, opened 2023-08-30
Clarify wording re: counting expression extending matching expression

bregenspan:
Maybe this is obvious to others, but I stumbled in understanding the wording here and wanted to clarify the difference between the default behavior for an empty counting expression vs. the behavior when a counting expression is set. The wording is technically correct already, but for me it would be helpful to hammer home the fact that the counting expression will always be equal to the rule expression when custom counting expression is empty (which is conceptually somewhat similar to extending an empty `""` counting expression with the matching expression), but that when a custom counting expression is defined, the rule expression typically should be manually copied over. This is an attempt to do so.


content/waf/rate-limiting-rules/parameters.md
```diff
@@ -67,8 +67,8 @@ Use _IP with NAT support_ to handle situations such as requests under NAT sharin
...
+  - Defines the criteria used for determining the request rate. By default, the counting expression is the same as the rule matching expression (defined in **If incoming requests match**). This default is also applied when you set this field to an empty string (`""`).
+  - If you do set a custom counting expression, it will not automatically extend the rule matching expression. Therefore, you may wish to include the matching expression in the counting expression. For example, you might want to perform rate limiting for clients sending more than five requests to `/api/` resulting in a `403` HTTP status code from the origin server. In this case, the matching expression would be `starts_with(http.request.uri.path, "/api/")` and the counting expression would be `http.response.code eq 403 and starts_with(http.request.uri.path, "/api/")`. If the counting expression did not include the matching expression (that is, if you had set the counting expression to `http.response.code eq 403`), any response with a `403` status code on any URL would increase the counter.
```
pedrosousa:
```suggestion
  - If you set a custom counting expression, it will not automatically extend the rule matching expression. Therefore, you may wish to include the matching expression in the counting expression. For example, you might want to perform rate limiting for clients sending more than five requests to `/api/` resulting in a `403` HTTP status code from the origin server. In this case, the matching expression would be `starts_with(http.request.uri.path, "/api/")` and the counting expression would be `http.response.code eq 403 and starts_with(http.request.uri.path, "/api/")`. If the counting expression did not include the matching expression (that is, if you had set the counting expression to `http.response.code eq 403`), any response with a `403` status code on any URL would increase the counter.
```

pedrosousa (approved):

pedrosousa:
Thank for your proposed change, @bregenspan!

PR #10512, opened 2023-08-30
Notifications

patriciasantaana:
PCX-7962


content/_redirects
```diff
@@ -490,6 +490,13 @@
 /magic-wan/tutorials/vyos/ /magic-wan/third-party/vyos/ 301
 /magic-wan/tutorials/warp/ /magic-wan/zero-trust/warp/ 301
 
+# notifications
```
kodster28, line 493: Might be worth doing a `/fundamentals/notifications/* /notifications/:splat 301` redirect, though that'll need to be added at the bottom of the file.

PR #10521, opened 2023-08-31
[QUEUES] Update consumer-concurrency.md

elithrar:
Fix duration units.

### DIFF
--- a/content/queues/learning/consumer-concurrency.md
+++ b/content/queues/learning/consumer-concurrency.md
@@ -107,4 +107,4 @@ Billing for consumers follows the [Workers unbound usage model](/workers/platfor
 
 ### Example
 
-A consumer Worker that takes 2 seconds ([256 GB-seconds](/workers/platform/pricing/#workers-unbound-billing-examples)) to process a batch of messages will incur the same overall costs to process 50 million (50,000,000) messages, whether it does so concurrently (faster) or individually (slower).
+A consumer Worker that takes 2 seconds ([0.256 GB-seconds](/workers/platform/pricing/#workers-unbound-billing-examples)) to process a batch of messages will incur the same overall costs to process 50 million (50,000,000) messages, whether it does so concurrently (faster) or individually (slower).


CharlieBurnett (approved):
LGTM!

content/queues/learning/consumer-concurrency.md
```diff
@@ -107,4 +107,4 @@ Billing for consumers follows the [Workers unbound usage model](/workers/platfor
 
 ### Example
 
-A consumer Worker that takes 2 seconds ([256 GB-seconds](/workers/platform/pricing/#workers-unbound-billing-examples)) to process a batch of messages will incur the same overall costs to process 50 million (50,000,000) messages, whether it does so concurrently (faster) or individually (slower).
+A consumer Worker that takes 2 seconds ([0.256 GB-seconds](/workers/platform/pricing/#workers-unbound-billing-examples)) to process a batch of messages will incur the same overall costs to process 50 million (50,000,000) messages, whether it does so concurrently (faster) or individually (slower).
```
jeansibelius, line 110: In the original Workers billing example the formula is given as follows: `(200,000 seconds) * 128 MB / 1 GB = 25,000 GB-seconds`.

As per the example, should not the value be `0.25`?

To get 0.256, I guess you're dividing with 1 GB as 1000 MB instead of 1024 MB?

Edit: link to Workers' docs for convenience. https://developers.cloudflare.com/workers/platform/pricing/#workers-unbound-billing-examples

jeansibelius (commented):
Noted a possibly incorrect value.

content/queues/learning/consumer-concurrency.md
jeansibelius: Now I started wondering, are the original Workers docs examples correct?

If yes, perhaps it would be good to clarify the units to GiB and MiB, since the current notation seems to have caused some confusion between decimal and binary.

Maddy-Cloudflare (approved):

PR #10530, opened 2023-08-31
Update faq.md

chris-martinelli:
update info around dashboard analytics


DavidJKTofan:
I recommend clarifying this more and providing context, as readers/users can be located anywhere in the world and customers can decide to set CMB to US or EU.

PR #10585, opened 2023-09-06
[workers] add wrangler changelog

admah:
Added Wrangler changelog to Cloudflare-wide changelog per guide.


kodster28:
I'll need to take more of a look at this since it's _slightly_ different from the intended use case.

admah:
@kodster28 that's fine. Can you let me know how it's different? I took these entries straight from our [changelog](https://github.com/cloudflare/workers-sdk/releases).

kodster28:
ah.... turns out most of the issue is that -- atm -- there's an assumption that there is a rendered changelog page for every entry listed in the changelog.

So adding a page at `/workers/platform/changelogs/wrangler/` solved the main issue.

kodster28:
@admah, did we want to push this live as is and then come back with whatever automated thing we figure out?

admah:
@kodster28 that's great with me. I think it looks good as-is.

kodster28 (approved):

PR #10624, opened 2023-09-08
[PubSub] Clarify PubSub get started guide to have appropriate information to successfully work

CharlieBurnett:
Added step in token creation to give proper permissions to access membership API

Added step exporting env variables so code snippet below works properly

Removed query param where a second token is generated, given the example only needs one

Changed the sh block for the token generation script to console to resolve text block treating # as a comment

### DIFF
--- a/content/pub-sub/guide.md
+++ b/content/pub-sub/guide.md
@@ -78,8 +78,9 @@ This API token requirement will be lifted prior to Pub/Sub becoming Generally Av
 4. Choose **Get Started** next to **Create Custom Token** 
 5. Name the token - e.g. "Pub/Sub Write Access"
 6. Under the **Permissions** heading, choose **Account**, select **Pub/Sub** from the first drop-down, and **Edit** as the permission.
-7. Click **Continue to Summary** at the bottom of the page, where you should see _All accounts - Pub/Sub:Edit_ as the permission
-8. Click **Create Token**, and copy the token value.
+7. Select **Add More**  below the newly created permission. Choose **User**, select **Memberships** from the first drop-down, and **Edit** as the permission.
+8. Click **Continue to Summary** at the bottom of the page, where you should see _All accounts - Pub/Sub:Edit_ as the permission
+9. Click **Create Token**, and copy the token value.
 
 In your terminal, configure a `CLOUDFLARE_API_TOKEN` environmental variable with your Pub/Sub token. When this variable is set, `wrangler` will use it to authenticate against the Cloudflare API.
 
@@ -218,10 +219,20 @@ $ which node
 $ npm i mqtt --save
 ```
 
-Generate a credential and store it in the `BROKER_TOKEN` environmental variable so the MQTT client can access it.
+Set your environment variables.
 
 ```sh
-export BROKER_TOKEN=$(curl -s -H "Authorization: Bearer ${CLOUDFLARE_API_TOKEN}" -H "Content-Type: application/json" "https://api.cloudflare.com/client/v4/accounts/${CLOUDFLARE_ACCOUNT_ID}/pubsub/namespaces/${DEFAULT_NAMESPACE}/brokers/${BROKER_NAME}/credentials?number=2&type=TOKEN&topicAcl=#" | jq '.result | to_entries | .[0].value')
+#Tip: If you want these to persist across boots, store the below variables in your `.zshrc`, `.bashrc`, or `.profile` files in your home folder
+$ export CLOUDFLARE_API_TOKEN="YourAPIToken"
+$ export CLOUDFLARE_ACCOUNT_ID="YourAccountID"
+$ export DEFAULT_NAMESPACE="TheNamespaceYouCreated"
+$ export BROKER_NAME="TheBrokerYouCreated"
+```
+
+Generate a credential and store it in the `BROKER_TOKEN` environmental variable so the MQTT client can access it.
+
+```console
+export BROKER_TOKEN=$(curl -s -H "Authorization: Bearer ${CLOUDFLARE_API_TOKEN}" -H "Content-Type: application/json" "https://api.cloudflare.com/client/v4/accounts/${CLOUDFLARE_ACCOUNT_ID}/pubsub/namespaces/${DEFAULT_NAMESPACE}/brokers/${BROKER_NAME}/credentials?type=TOKEN&topicAcl=#" | jq '.result | to_entries | .[0].value')
 ```
 
 Create a file called `index.js ` and make sure to update the `brokerEndpoint` with the address of your Pub/Sub broker.
@@ -232,6 +243,9 @@ const mqtt = require('mqtt')
 const brokerEndpoint = "mqtts://my-broker.my-namespace.cloudflarepubsub.com"
 const options = {
   port: 8883,
+  //If password alone does not successfully auth, uncomment the lines below and try again
+  //clientID: "First portion of token generated above",
+  //username: "Your username",
   password: process.env.BROKER_TOKEN,
   protocolVersion: 5, // MQTT 5
 }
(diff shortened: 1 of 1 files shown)


CharlieBurnett:
@Maddy-Cloudflare @elithrar if either of you know how to properly escape a # character inside a ```sh block, lmk. I tried a \,\`,# #, ##, and \` \`, none of which had the desired effect. 

My solution was to change the block to a ```console so the full text could be copied. This affects documentation formatting slightly to the trained eye,  but is otherwise completely workable

content/pub-sub/guide.md
```diff
@@ -78,8 +78,9 @@ This API token requirement will be lifted prior to Pub/Sub becoming Generally Av
 4. Choose **Get Started** next to **Create Custom Token** 
 5. Name the token - e.g. "Pub/Sub Write Access"
 6. Under the **Permissions** heading, choose **Account**, select **Pub/Sub** from the first drop-down, and **Edit** as the permission.
-7. Click **Continue to Summary** at the bottom of the page, where you should see _All accounts - Pub/Sub:Edit_ as the permission
-8. Click **Create Token**, and copy the token value.
+7. Select **Add More**  below the newly created permission. Choose **User**, select **Memberships** from the first drop-down, and **Edit** as the permission.
```
elithrar: Can you explain why this is necessary / why it didn't work without this?

content/pub-sub/guide.md
```diff
@@ -218,10 +219,20 @@ $ which node
 $ npm i mqtt --save
 ```
 
-Generate a credential and store it in the `BROKER_TOKEN` environmental variable so the MQTT client can access it.
+Set your environment variables.
 
 ```sh
-export BROKER_TOKEN=$(curl -s -H "Authorization: Bearer ${CLOUDFLARE_API_TOKEN}" -H "Content-Type: application/json" "https://api.cloudflare.com/client/v4/accounts/${CLOUDFLARE_ACCOUNT_ID}/pubsub/namespaces/${DEFAULT_NAMESPACE}/brokers/${BROKER_NAME}/credentials?number=2&type=TOKEN&topicAcl=#" | jq '.result | to_entries | .[0].value')
+#Tip: If you want these to persist across boots, store the below variables in your `.zshrc`, `.bashrc`, or `.profile` files in your home folder
```
elithrar:
```suggestion
```

We generally want to avoid users from storing secrets persistently in config. Since this is a private beta, it's OK for these to be temporary.

content/pub-sub/guide.md
```diff
@@ -218,10 +219,20 @@ $ which node
...
+$ export CLOUDFLARE_API_TOKEN="YourAPIToken"
+$ export CLOUDFLARE_ACCOUNT_ID="YourAccountID"
+$ export DEFAULT_NAMESPACE="TheNamespaceYouCreated"
+$ export BROKER_NAME="TheBrokerYouCreated"
```
elithrar, line 228: It's likely better to have each of these done _before_ the `wrangler ...` command that uses that name, or immediately after, so they are in context.

content/pub-sub/guide.md
```diff
@@ -232,6 +243,9 @@ const mqtt = require('mqtt')
 const brokerEndpoint = "mqtts://my-broker.my-namespace.cloudflarepubsub.com"
 const options = {
   port: 8883,
+  //If password alone does not successfully auth, uncomment the lines below and try again
+  //clientID: "First portion of token generated above",
+  //username: "Your username",
```
elithrar: I'd suggest just making this non-optional: don't make users decide during a tutorial.

elithrar (changes requested):

content/pub-sub/guide.md
CharlieBurnett: Attempting to create a namespace with a token lacking these credentials fails with a 10000 authorization error for the API

This is something that I experience when working through the docs as is and has also been experienced by devs in discord doing this same

content/pub-sub/guide.md
CharlieBurnett: That's correct. 

The only time these env variables are used is in the command following this commit, with the exception of the CLOUDFLARE_API_TOKEN. Which is first exported above, but also is included below.

 I have repeated the CLOUDFLARE_API_TOKEN in this list for clarity, but am happy to remove that line if we feel it improves the reading experience

content/pub-sub/guide.md
CharlieBurnett: The reason these lines exist is specifically because of that `mqtt.js` silent auth failure bug we talked about yesterday. 

I added them in as comments to help anyone who might have trouble logging in, but if the bug is fixed then there's no need for them. Thought it would be a nice temporary measure. 

I've got a notification alert set up to follow the bug report on Github, so if its ever resolved I can go back and update.

My preference would be removing these lines if we feel like that doesn't make sense.

content/pub-sub/guide.md
```diff
@@ -78,8 +78,9 @@ This API token requirement will be lifted prior to Pub/Sub becoming Generally Av
 4. Choose **Get Started** next to **Create Custom Token** 
 5. Name the token - e.g. "Pub/Sub Write Access"
 6. Under the **Permissions** heading, choose **Account**, select **Pub/Sub** from the first drop-down, and **Edit** as the permission.
-7. Click **Continue to Summary** at the bottom of the page, where you should see _All accounts - Pub/Sub:Edit_ as the permission
-8. Click **Create Token**, and copy the token value.
+7. Select **Add More**  below the newly created permission. Choose **User**, select **Memberships** from the first drop-down, and **Edit** as the permission.
+8. Click **Continue to Summary** at the bottom of the page, where you should see _All accounts - Pub/Sub:Edit_ as the permission
```
dcpena:
```suggestion
8. Select **Continue to Summary** at the bottom of the page, where you should see _All accounts - Pub/Sub:Edit_ as the permission.
```

content/pub-sub/guide.md
```diff
@@ -78,8 +78,9 @@ This API token requirement will be lifted prior to Pub/Sub becoming Generally Av
...
+9. Click **Create Token**, and copy the token value.
```
dcpena:
```suggestion
9. Select **Create Token** and copy the token value.
```

content/pub-sub/guide.md
dcpena:
```suggestion
7. Select **Add More** below the newly created permission. Choose **User** > **Memberships** from the first dropdown and **Edit** as the permission.
```

dcpena (commented):
Left a few suggestions to align with our style guide.

content/pub-sub/guide.md
CharlieBurnett: @Maddy-Cloudflare why do we "click" some places and "select" others? What's the standard practice for this?
Note that "click" is the way the docs were previously written. I have no preference, just want to understand which goes where

dcpena: Hi there, Charlie! Figured I'd respond while it's outside of working hours for Maddy. When the docs were originally written, the PCX team hadn't yet revisited guidance on Select vs Click. Select was the chosen option as it's device agnostic.

To avoid having that one step look completely different from the rest of the procedure, we can keep Click and just make note of revisiting the "Click" options to update them with Select at a later time.

CharlieBurnett: Thanks @dcpena makes perfect sense. I'll get a commit going changing that language for the next PR

content/pub-sub/guide.md
elithrar: I'd strongly recommend simplifying it. Reduce the choices a user has to make when starting out. "If that doesn't work, try again" is neither a) guaranteed to work and b) leads users to ask "why not just make that the default?"

CharlieBurnett: Good call. Will remove. Committed.

elithrar (approved):

dcpena (approved):

PR #10648, opened 2023-09-11
[ZT] Move macOS cert methods into tabs

maxvp:
Moves the methods for installing the Cloudflare root cert into tabs to streamline procedure.


patriciasantaana (approved):

XoL1507:
thanks

PR #10656, opened 2023-09-12
[ZT] Clarify Access Bypass policy example

ranbel:


kennyj42:
I don't love this example to be honest, given it's so counter to a Zero Trust security model. I wonder if we could include an aside that we don't recommend this?

kennyj42 (approved):

ranbel:
> I don't love this example to be honest, given it's so counter to a Zero Trust security model. I wonder if we could include an aside that we don't recommend this?

Yeah I was thinking the same. What would be a better example for the Bypass action?

kennyj42:
Often times web apps have specific endpoints that have to be public. for example, wordpress has a specific endpoint under the /admin route that has to be publicly routable. I think that would be a better example to include here.

ranbel:
What does that policy look like?

kennyj42:
It'd be something like:

create a new App

Domain = test.example.com/admin/your-specific-url

action = bypass

Include = Everyone (or country = USA or any other of the available selectors)

PR #10682, opened 2023-09-13
Update convert-partial-to-full.md

chris-martinelli:
add note about USSL being supported in a full setup


kodster28 (approved):
Just moved the content to a footnote, but looks good @chris-martinelli. Thanks!!

PR #10685, opened 2023-09-14
Update maximum length of a list name

licitdev:
The maximum name of a list name should be `20` characters instead of `50` when creating a new list in `/<CF_ACCOUNT_ID>/configurations/lists/new`.

image


pedrosousa:
Hi @licitdev 
Actually, you just found an issue in the Cloudflare dashboard.

The Cloudflare API operation for [creating lists](https://developers.cloudflare.com/api/operations/lists-create-a-list) allows you to define a list name up to 50 characters long, and the dashboard is currently limiting names to a max of 20 characters, when it should be enforcing the same 50-character limit.

This issue will be fixed soon in the dashboard. Thank you for reporting it!

PR #10705, opened 2023-09-14
[D1] update page weight and title

elithrar:
Fixes a bug where two pages had `weight: 20` - which causes one page to not be displayed. We always want the Pages doc link to be at the bottom of the list so it's easy to find/see.


rozenmd (approved):

XoL1507:
what will happen if i merge this?

PR #10709, opened 2023-09-15
[Workers] updated "add environment variables ..." in environment-variables.md

jeremy-harnois:
- Separated "In **Overview**, select your Worker > **Settings**." into two steps since they occur in separate views
- Added a missing step required to see "Under **Environment Variables**"
- Aligned selection/button text


deadlypants1973 (approved):

deadlypants1973:
For some reason the check would not run on this PR, but moved this info into another PR. Thank you for this change!

PR #10711, opened 2023-09-15
Correct the HC UA

cicku:


kodster28 (approved):

kodster28:
Thanks, confirmed via looking at internal code repos

PR #10716, opened 2023-09-15
Update Zaraz Changelog

omarmosid:


data/changelogs/zaraz.yaml
```diff
@@ -2,6 +2,12 @@
 link: "/zaraz/changelog/"
 productName: Zaraz
 entries:
+- publish_date: '2023-09-13'
+  title:
+  description: |-
+    * **Consent Management**: Add support for custom button translations.
+    * **Consent Management**: Modal stays fixed when scrolling.
```
tomkln, line 9: I think it‘s not clear if this was an issue that got fixed (no longer fixed when scrolling now) or is the result of the fix. 

Looks good apart from that :)

bjesus (approved):

PR #10627, opened 2023-09-09
Use "handler" terminology consistently

irvinebroque:
We currently are inconsistent about what we call the "handler" methods that are called when a Worker is invoked (`fetch()`, `queue()`, `scheduled()`, `email()`, etc.). Much of our docs still calls these "events", and shows Service Worker syntax. And as a result, it's hard to answer questions like:

- What is a handler, and which handlers are available within Workers? What are the kinds of inputs and triggers that can invoke a Worker?
- When I search online or in docs, should I search for "fetch handler"? "fetch event"? "fetch method"?

This PR:

1. consolidates all handlers under their own section within Runtime APIs.
2. Moves remaining Service Worker syntax content to the ["Migrate to Module Workers"](https://developers.cloudflare.com/workers/learning/migrate-to-module-workers/) page. (which should probably be renamed "ES Modules")
 
- [x] Fix redirects
- [x] Rename ["Migrate to Module Workers"](https://developers.cloudflare.com/workers/learning/migrate-to-module-workers/) to "Migrate from Service Workers to ES Modules"

PR #10722, opened 2023-09-15
[Workers] Clarify docs for Request and cf properties

irvinebroque:
refs https://github.com/cloudflare/cloudflare-docs/pull/10627 which should help with other cleanup about `fetch()` handler


content/workers/runtime-apis/request.md
```diff
@@ -178,6 +177,7 @@ All properties of an incoming `Request` object (that is, `event.request`) are re
 *   `cf` {{<type-link href="#incomingrequestcfproperties">}}IncomingRequestCfProperties{{</type-link>}} {{<prop-meta>}}read-only{{</prop-meta>}}
 
     *   An object containing properties about the incoming request provided by Cloudflare’s global network.
+    *   This property is read-only. To modify its values, pass in the new values on the [`cf` key of the `init` options argument](/workers/runtime-apis/request/#requestinit) when creating a new Request object
```
jasnell: This is close. In a scenario where I create a new `Request` from an existing `Request`, the `newRequest.cf` will not be readonly and can be modified directly.

content/workers/runtime-apis/request.md
```diff
@@ -163,7 +162,7 @@ Invalid or incorrectly-named keys in the `cf` object will be silently ignored. C
 
 ## Properties
 
-All properties of an incoming `Request` object (that is, `event.request`) are read only. To modify a request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
+All properties of an incoming `Request` object (the request you receive from the [`fetch()` handler](/workers/runtime-apis/fetch-event/)) are read only. To modify a the properties of an incoming request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
```
deadlypants1973:
```suggestion
All properties of an incoming `Request` object (the request you receive from the [`fetch()` handler](/workers/runtime-apis/fetch-event/)) are read-only. To modify a the properties of an incoming request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
```
Grammatically, I would write read-only. But the general consensus in Microsoft docs seems like the - is a necessary part of this meaning. Adding for consistency with line 180.

deadlypants1973 (approved):

content/workers/runtime-apis/request.md
kodster28: Would it be correct to say that in this section then?
cc: @jasnell 

```suggestion
    *   This property is read-only (unless created from an existing `Request`). To modify its values, pass in the new values on the [`cf` key of the `init` options argument](/workers/runtime-apis/request/#requestinit) when creating a new `Request` object.
```

content/workers/runtime-apis/request.md
```diff
@@ -161,7 +161,7 @@ Invalid or incorrectly-named keys in the `cf` object will be silently ignored. C
 
 ## Properties
 
-All properties of an incoming `Request` object (that is, `event.request`) are read only. To modify a request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
+All properties of an incoming `Request` object (the request you receive from the [`fetch()` handler](/workers/runtime-apis/fetch-event/)) are read-only. To modify a the properties of an incoming request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
```
kodster28:
```suggestion
All properties of an incoming `Request` object (the request you receive from the [`fetch()` handler](/workers/runtime-apis/fetch-event/)) are read-only. To modify the properties of an incoming request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
```

content/workers/runtime-apis/request.md
```diff
@@ -18,17 +18,17 @@ The most common way you will encounter a `Request` object is as a property of an
...
 export default {
-	async fetch(request, env, ctx) { 
+	async fetch(request, env, ctx) {
 		return new Response('Hello World!');
 	},
 };
 ```
 
-You may also want to construct a `Request` yourself when you need to modify a request object, because the incoming request parameter is immutable.
+You may also want to construct a `Request` yourself when you need to modify a request object, because the incoming `request` parameter that you receive from the [`fetch()` handler](workers/runtime-apis/fetch-event/) is immutable.
```
kodster28:
```suggestion
You may also want to construct a `Request` yourself when you need to modify a request object, because the incoming `request` parameter that you receive from the [`fetch()` handler](/workers/runtime-apis/handlers/fetch/) is immutable.
```

content/workers/runtime-apis/request.md
```diff
@@ -161,7 +161,7 @@ Invalid or incorrectly-named keys in the `cf` object will be silently ignored. C
 
 ## Properties
 
-All properties of an incoming `Request` object (that is, `event.request`) are read only. To modify a request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
+All properties of an incoming `Request` object (the request you receive from the [`fetch()` handler](/workers/runtime-apis/fetch-event/)) are read-only. To modify the properties of an incoming request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
```
kodster28:
```suggestion
All properties of an incoming `Request` object (the request you receive from the [`fetch()` handler](/workers/runtime-apis/handlers/fetch/)) are read-only. To modify the properties of an incoming request, create a new `Request` object and pass the options to modify to its [constructor](#constructor).
```

PR #10724, opened 2023-09-15
[Turnstile] 401 error on local development

patriciasantaana:
PCX-7965
https://community.cloudflare.com/t/allow-localhost-or-127-0-0-1-as-acceptable-domains-for-turnstile/423897

### DIFF
--- a/content/turnstile/frequently-asked-questions.md
+++ b/content/turnstile/frequently-asked-questions.md
@@ -220,6 +220,15 @@ If you encounter an endless challenge loop, try disabling your browser extension
 {{</faq-item>}}
 
 {{<faq-item>}}
+{{<faq-question level=2 text="What if I receive a 401 network error when developing locally?" >}}
+{{<faq-answer>}}
+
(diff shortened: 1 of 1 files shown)


content/turnstile/frequently-asked-questions.md
```diff
@@ -219,6 +219,15 @@ If you encounter an endless challenge loop, try disabling your browser extension
 {{</faq-answer>}}
 {{</faq-item>}}
 
+{{<faq-item>}}
+{{<faq-question level=2 text="What if I receive a 401 network error when developing locally?" >}}
+{{<faq-answer>}}
+
+Cloudflare returns a 401 status as part of the Private Access Token (PAT) challenge on `…/pat/…` URLs. You can use Turnstile on `127.0.0.1` or `localhost` as a domain to allow for local development.
```
punkeel, line 226: I'm not sure this is accurate, the linked post is a bit confusing.

401 happens everywhere and is PAT.

If you want to test locally you need to allow 127.0.0.1 / localhost, or you'll get errors (along the lines of "This domain is not allowed")

These two problems are not really related 😅 


We already have a Q for local mode: `Can I use Turnstile when developing locally?`

patriciasantaana: thanks for checking it out! I just took the answers from the community post and added it here. I suppose we can close this out

PR #10743, opened 2023-09-18
[Chore] Ignore a specific page in anchor link checking

kodster28:
Addresses https://github.com/cloudflare/cloudflare-docs/actions/runs/6220961176/job/16882027864#step:9:8

### DIFF
--- a/bin/.htmltest.yml
+++ b/bin/.htmltest.yml
@@ -2,6 +2,7 @@ DirectoryPath: "public"
 EnforceHTTPS: false
 IgnoreURLs:
 - "/api/*"
+- "zero-trust/index.html"
 IgnoreDirs:
 - "support/other-languages/"
 CheckLinks: true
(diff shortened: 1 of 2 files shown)


bin/.htmltest.yml
```diff
@@ -2,6 +2,7 @@ DirectoryPath: "public"
 EnforceHTTPS: false
 IgnoreURLs:
 - "/api/*"
+- "zero-trust/index.html"
```
pedrosousa:
```suggestion
- "/zero-trust/index.html"
```
Nit: Not sure if it should start with a slash, like `/api/*` does. I think we should at least be consistent (both `IgnoreURLs` and `IgnoreDirs` accept regexes).

htmltest also seems to support [tag-based ignores](https://github.com/wjdp/htmltest#see_no_evil-ignoring-content), if we wanted to ignore just a single URL. We'd have to use an HTML anchor element for this instead of Markdown, though.

pedrosousa (approved):

PR #10775, opened 2023-09-19
[Wrangler] - Automated changelog

kodster28:
Pull from GH releases api (and set the Hugo cache so we don't get rated limited).
- https://wrangler-automated-yaml.cloudflare-docs-7ou.pages.dev/changelog/?product=wrangler
- https://wrangler-automated-yaml.cloudflare-docs-7ou.pages.dev/workers/platform/changelog/wrangler/index.xml
- https://wrangler-automated-yaml.cloudflare-docs-7ou.pages.dev/workers/platform/changelog/wrangler/

cc: @admah


admah:
I can't speak to the code, but the result looks great! This is really good. 🎉

JacobMGEvans:
Super cool!

config.toml
```diff
@@ -51,3 +51,7 @@ tutorial_token = "xxa5a7d4c9-db72-424d-860d-ae68ff7f60ac"
 
 [build]
   noJSConfigInAssets = true
+
+[caches]
```
GregBrimble, line 55: TIL! But not sure if this will actually take effect in Pages since we don't currently support caching Hugo in our build system.

layouts/partials/wrangler-changelog.html
```diff
@@ -0,0 +1,16 @@
+{{- $wranglerReleases := getJSON "https://api.github.com/repos/cloudflare/workers-sdk/releases" -}}
+{{- $results := slice -}}
+
+{{- range $index, $item := $wranglerReleases -}}
+{{- if (in $item.tag_name "wrangler@3.") -}}
```
GregBrimble: Wrangler will cut v4 in the coming months. Worth just making this slightly more general with `wrangler@`?

config.toml
GregBrimble: Note for the folks working on that: https://gohugo.io/getting-started/configuration/#configure-cachedir

layouts/shortcodes/product-changelog.html
```diff
@@ -1,5 +1,10 @@
 {{- $product := .Page.Params.changelog_file_name -}}
-{{- $changelogData := index (index (index .Site.Data "changelogs") $product) "entries" -}}
+{{- $changelogData := slice -}}
+{{- if eq $product "wrangler" -}}
+{{- $changelogData = index (index (partial "wrangler-changelog.html") "wrangler") "entries" -}}
+{{- else -}}
```
GregBrimble: Kind of a shame you need this if/else branching. I'd maybe do something like:

```
<!-- layouts/shortcodes/full-changelog-data.html -->
{{- $changelogData := index .Site.Data "changelogs" -}} 
{{- $wrangler := partial "wrangler-changelog.html" -}}
{{- $changelogData = merge $changelogData $wrangler -}}
{{- return $changelogData -}}
```

And then reference that in `layouts/shortcodes/full-changelog.html`, `layouts/shortcodes/product-changelog.html` and `layouts/_defaults/changelog.rss.xml`. This'll save you needing to remember to add this if/else branch everywhere you use the changelog if you ever add more :)

config.toml
kodster28: I think the main concern is folks building locally.... that's where I'd be worried about us hitting the limit for an unauth'd endpoint.

kodster28:
Your feedback should be addressed in the latest commit, @GregBrimble. Thanks again for the advice about the partial, way better than those `if/else` statements.

layouts/partials/wrangler-changelog.html
```diff
@@ -0,0 +1,16 @@
+{{- $wranglerReleases := getJSON "https://api.github.com/repos/cloudflare/workers-sdk/releases" -}}
```
pedrosousa, line 1: Maybe we can add a comment here saying that the cache duration for JSON content is defined in `config.toml`?

pedrosousa (approved):

PR #10778, opened 2023-09-19
fix code copy btn for light mode

Mrkezii:
Fixes an issue with the `CodeCopy` component where we were not properly isolating darkmode style to darkmode only. 

Screenshot 2023-09-19 at 17 04 03


kodster28 (approved):
LGTM

PR #10785, opened 2023-09-19
[Chore] Added new inline-pill shortcode + consolidate previous heading pill shortcodes

kodster28:
Current state is demo-ing the functionality. Will remove from api param in subsequent commit.

To do:

- [x] Refactor `alpha`, `beta`, and `deprecated` shortcodes to pull from the same `heading-pill` shortcode.
- [x] Replace all instances of `alpha`, `beta`, and `deprecated`.


static/style.css
```diff
@@ -5154,6 +5154,11 @@ pre {
   align-items: center;
 }
 
+.inline-pill {
+  display:inline !important;
```
maxvp:
```suggestion
  display: inline !important;
```

pedrosousa (approved):

PR #10787, opened 2023-09-20
[Chore] Sanitize Unicode chars in file paths (list changes action)

pedrosousa:
Follows Hugo's conversion from file paths to URI paths in the action that displays a list of changed files in a PR.
Allows the list of changed files to also work for changes under Support.

(Can only test after merging.)


kodster28 (approved):
Worth testing!

PR #10788, opened 2023-09-20
[DNS] Callout caveat when reverting subdomain setup

AlphaK5:
PCX-8396


RebeccaTamachiro:
Hey! Thanks for raising this. Have left you a question on PCX-8396 and will continue to address this on a different PR: #11779

PR #10797, opened 2023-09-20
Update BYOIP developer documentation w/ Cloudflare's ASN

gigawhitlocks:
The Cloudflare documentation is missing our AS number, which causes unnecessary delays for BYOIP customers as there is a round trip discussion to the Addressing Team w/ the same question and same answer (our AS number) each time.

This change updates the docs to have our AS number present in the documentation so that this delay can be avoided. Further, the term LOA stands for Letter of Agency, not Authorization, and I have updated the documentation to reflect this.

### DIFF
--- a/content/byoip/concepts/loa.md
+++ b/content/byoip/concepts/loa.md
@@ -1,30 +1,30 @@
 ---
-title: Letter of Authorization
+title: Letter of Agency
 pcx_content_type: concept
 weight: 3
 meta:
-  title: Letter of Authorization (LOA)
+  title: Letter of Agency (LOA)
 ---
 
-# Letter of Authorization (LOA)
+# Letter of Agency (LOA)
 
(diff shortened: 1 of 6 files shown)


kodster28:
Curious on the LOA naming change. From my naive [Google search](https://www.google.com/search?hl=en&q=loa%20meaning%20network#ip=1), seems like Authorization is more common... but can absolutely defer to you on this.

gigawhitlocks:
Quoting Joe Abley from the JIRA ticket associated with this change:

>LOA stands for "Letter of Agency". Many people think it stands for "Letter of Authorization" but it really doesn't. Let's take the opportunity while we have our editing hats on to update this page in the documentation to use the correct phrase, perhaps noting that the phrase "Letter of Authorization" is often in common usage, even though it's strictly incorrect.

>Here's a [Wikipedia link about the term](https://en.wikipedia.org/wiki/Letter_of_Agency). This is mainly written in an FCC context, but (I think) that the FCC is actually the origin of the term so that doesn't seem terrible.

I'm happy to add some verbiage about how _Authorization_ is more common but _Agency_ is more correct

kodster28:
Let me push up some changes to your PR that'll address this.

kodster28 (approved):

PR #10803, opened 2023-09-20
Create how-range-requests-work-at-cloudflare.md

luishcardoso:
As per PCX-8069, we have been receiving requests from customers on which they need to understand how Range Requests work with Cloudflare. We currently have the following internal wiki for this:

https://wiki.cfdata.org/display/~gagan/How+range+requests+work+at+Cloudflare

If possible, it would be helpful for customers to have a similar public source of this flow.

### DIFF
--- a/content/cache/concepts/how-range-requests-work -at-cloudflare.md
+++ b/content/cache/concepts/how-range-requests-work -at-cloudflare.md
@@ -1,0 +1,19 @@
+---
+title: How range requests work at Cloudflare
+pcx_content_type: concept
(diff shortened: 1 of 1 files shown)


kodster28:
Howdy @luis-henriquecardoso, can you update the filename to remove the spaces and instead add the images to the `/assets/images/cache` folder?

angelampcosta:
@luis-henriquecardoso I'm closing this out as changes are planned. I'll follow up on the ticket.

PR #10808, opened 2023-09-20
[Demo] Vanilla Hugo codeblocks

kodster28:
If you run `hugo server`, you get codeblocks due to advances in Hugo's built-in [markdown render hooks](https://gohugo.io/templates/render-hooks/#render-hooks-for-code-blocks) (note that local builds will then be visible on http://localhost:1313/).

Definitely breaks our current implementation of tabs and the headings + other stuff we've added + need more padding + copy code functions.... but maybe this is a better solution than the current approach with Vite.

Note, the local dev experience is also a lot better b/c you only have to refresh a single page + builds don't hang.

cc: @penalosa, @pedrosousa

### DIFF
--- a/content/d1/examples/d1-and-remix.md
+++ b/content/d1/examples/d1-and-remix.md
@@ -24,9 +24,6 @@ The following example shows you how to define a Remix [`loader`](https://remix.r
 * Bindings are passed through on the `context.env` parameter passed to a `LoaderFunction`.
 * If you configured a [binding](/pages/platform/functions/bindings/#d1-databases) named `DB`, then you would access D1's [client API](/d1/platform/client-api/#query-statement-methods) methods via `context.env.DB`.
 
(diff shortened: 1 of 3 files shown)


content/d1/examples/d1-and-remix.md
```diff
@@ -24,9 +24,6 @@ The following example shows you how to define a Remix [`loader`](https://remix.r
 * Bindings are passed through on the `context.env` parameter passed to a `LoaderFunction`.
 * If you configured a [binding](/pages/platform/functions/bindings/#d1-databases) named `DB`, then you would access D1's [client API](/d1/platform/client-api/#query-statement-methods) methods via `context.env.DB`.
 
-{{<tabs labels="ts">}}
```
kodster28, line 27: Was causing some issues with the demo.

PR #10815, opened 2023-09-21
[CF1] Document using different usernames in browser-based SSH

Sp1l:
Using a duplicate user works as expected

### DIFF
--- a/content/cloudflare-one/_partials/_ssh-usernames.md
+++ b/content/cloudflare-one/_partials/_ssh-usernames.md
@@ -22,7 +22,13 @@ SSH certificates include one or more `principals` in their signature which indic
 By default, SSH servers authenticate the Unix username against the principals listed in the user's certificate. You can configure your SSH server to accept principals that do not match the Unix username.
 
 {{<Aside type="note">}}
-Differing usernames do not work with the browser-based terminal. If you would like to use short-lived certificates with the browser-based terminal, you will need your users' usernames to match.
+To use differing usernames with the browser-based terminal. You need to add create an additional user with the same uid/gid and home directory as the original user:
+
+```sh
+$ getent passwd jdoe
+jdoe:x:1000:1000:John Doe:/home/jdoe:/usr/bin/zsh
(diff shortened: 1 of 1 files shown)


ranbel:
Thanks for the suggestion! Unfortunately I was unable to get this configuration working in my test environment. After logging into Access as `jdoe@example.com`, the browser-based terminal automatically logs you in as the original user `jdoe`. There is no option to enter a different username (e.g. `john.doe`) through the browser when using short-lived certs. How were you able to log in as `john.doe`? 

In any case, I am going to close this PR as this is not a configuration that we officially support.

ranbel:
#13339

Sp1l:
> Thanks for the suggestion! Unfortunately I was unable to get this configuration working in my test environment. After logging into Access as `jdoe@example.com`, the browser-based terminal automatically logs you in as the original user `jdoe`. There is no option to enter a different username (e.g. `john.doe`) through the browser when using short-lived certs. How were you able to log in as `john.doe`?
> 
> In any case, I am going to close this PR as this is not a configuration that we officially support.

By adding another entry to the passwd database that has the same uid/gid as the original user, but a different username.
For all purposes and intent, that other name behaves just like the original user. Linux [`useradd(8)`](https://manpages.ubuntu.com/manpages/noble/en/man8/useradd.8.html) has the `-o` option to allow these duplicate users to be created.

This is an ancient thing, for example FreeBSD by default has a `toor` account (see https://cgit.freebsd.org/src/tree/etc/master.passwd#n2). Purpose of this is so you can set `root`'s shell to a shell that's not in the base system. If at boot you were unable to mount your `/usr/local` you'd be locked out of root (can't start shell), but `toor` would let you login.

``` 
root::0:0::0:0:Charlie &:/root:/bin/sh
toor:*:0:0::0:0:Bourne-again Superuser:/root:
```

There's nothing for Cloudflare to support here, support is in the OS 😄, documenting this would make migration easier, otherwise you'd have to go rename existing users to match whatever you get from the Identity Provider. Renaming the user would make things that rely on the username (not the id) like entries in `/etc/cron.daily` fail.

PR #10818, opened 2023-09-21
PCX-7659 - explain Android loopback address limitations

lfcassidy:
There is behaviour specific to Android and split tunnel loopback addresses that differs across versions & OS's that should be explained.


ranbel:
@dh-cf

ranbel:
This bug was fixed in CLIENT-8777. Issues with older versions of the client are generally documented in the release notes, not in the dev docs.

PR #10827, opened 2023-09-21
[workers] small updates for getting started

admah:
- update ts/js file mentions in getting started guide
- remove wrangler init mention from wrangler install and update guide

### DIFF
--- a/content/workers/get-started/guide.md
+++ b/content/workers/get-started/guide.md
@@ -55,7 +55,7 @@ For this guide, set up a basic Worker:
 
 1. Name your new Worker directory by specifying where you want to create your application.
 2. Select `"Hello World" script` as the type of application you want to create.
-3. Answer `no` to using TypeScript.
+3. Answer `yes` or `no` to using TypeScript.
 
 You will be asked if you would like to deploy the project to Cloudflare.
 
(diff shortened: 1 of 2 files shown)


content/workers/get-started/guide.md
```diff
@@ -55,7 +55,7 @@ For this guide, set up a basic Worker:
 
 1. Name your new Worker directory by specifying where you want to create your application.
 2. Select `"Hello World" script` as the type of application you want to create.
-3. Answer `no` to using TypeScript.
+3. Answer `yes` or `no` to using TypeScript.
```
deadlypants1973, line 58: In the rest of the guide, especially in the "write code" section step 3, we are specifically working with JS not TS. 

1) Could we add an aside instead of this change to say that TS is an acceptable choice? Or could you add TS code to the rest of the Get stared guide?
2) Can you update the 'worker.js' to 'index.js' in the rest of the guide where appropriate?

admah: @deadlypants1973 I added an aside noting that and the differences. I also updated the instances of `worker.js`.

deadlypants1973 (approved):

PR #10830, opened 2023-09-21
[Notifications] Incident Alerts

patriciasantaana:
PCX-7530

### DIFF
--- a/content/notifications/notification-available.md
+++ b/content/notifications/notification-available.md
@@ -147,6 +147,33 @@ During these maintenance windows, you may experience a slight increase in latenc
 </div>
 </details>
 
+<details>
+<summary>Incident Alert</summary>
+<div>
+
+**Who is it for?**
+
+Customers interested in knowing about Cloudflare incidents.
+
+The notification lets you know when Cloudflare incidents are created, updated, and resolved.
+
+**Other options / filters**
+
+Customers can filter incident alerts to specific impact levels (minor, major, critical). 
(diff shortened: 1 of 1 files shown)


content/notifications/notification-available.md
```diff
@@ -147,6 +147,33 @@ During these maintenance windows, you may experience a slight increase in latenc
...
+
+Additionally, incident alerts can be filtered to incidents affecting specific components. By default, incident alerts will trigger a notification for incident updates across all impact levels and components.
+
+**Included with**
+
+All plans.
+
+**What should you do if you receive one?**
+
+Review your [analytics](/analytics/) page to see if your domain is impacted.
```
RebeccaTamachiro, line 172: Just one thing that crossed my mind: we're currently linking to the general overview page for the Analytics tile. Should or can we be more specific? 
For example, if it is supposed to always be the account analytics page, maybe we could link to (https://incident-alerts.cloudflare-docs-7ou.pages.dev/analytics/account-and-zone-analytics/account-analytics/), just so to avoid users being uncertain/ end up going to the zone-level analytics.

RebeccaTamachiro (approved):

content/notifications/notification-available.md
patriciasantaana: This link was actually agreed upon with the engineer as there's not really a specific type of analytics we should be leading to given that it covers any kind of incident.

PR #10836, opened 2023-09-22
[SSL] Provide more guidance for AOP setup

RebeccaTamachiro:
PCX-7901


jason-cf:
@RebeccaTamachiro I would make step 1 on both pages a partial because the information should be the same, but they are currently different.

jason-cf (changes requested):
left comment in chat

RebeccaTamachiro:
> @RebeccaTamachiro I would make step 1 on both pages a partial because the information should be the same, but they are currently different.

Steps are different due to zone-level being possible to use either CF cert or custom, while per-hostname is only custom.

jason-cf (approved):

PR #10849, opened 2023-09-25
[Fundamentals] Add page to basic tasks

pedrosousa:
Addresses PCX-8172.


content/fundamentals/basic-tasks/trace-request.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Trace a request
```
angelampcosta: I think we also added beta in parenthesis, for instance https://developers.cloudflare.com/cache/how-to/cache-rules/. Not sure if there is a guideline.

angelampcosta (approved):

PR #10906, opened 2023-09-27
Update index.md for Regex101 URL

mlazzarotto:
Removed `/?flavor=rust` as choosing the flavor via URL is not supported by Regex101.
A related issue was opened here [https://github.com/firasdib/Regex101/issues/2090](https://github.com/firasdib/Regex101/issues/2090) but the issue was then closed although not fulfilled.

Although the link works, it can confuse the users into thinking they are using 'Rust' flavor, while they are not.

### DIFF
--- a/content/firewall/known-issues-and-faq/index.md
+++ b/content/firewall/known-issues-and-faq/index.md
@@ -45,7 +45,7 @@ The following image illustrates how double quotes are automatically escaped to `
 
 ### Why isn't my regular expression pattern matching working?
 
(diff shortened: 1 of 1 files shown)


pedrosousa:
Actually, the `flavor` query string parameter still works, but you must also provide `regex`, even if it's empty.
I'll update the PR.

Thank you for the heads-up that the current link wasn't working properly!

pedrosousa (approved):

PR #10923, opened 2023-09-27
[Cache/Support] Move Wildcard matching to Page Rules section

Erisa:
The document talks exclusively about Page Rules behaviour, but does not mention it's doing so. The same explainers don't apply to Cache Rules and it doesn't cover just cache, so It doesn't belong under cache section.

https://developers.cloudflare.com/cache/reference/wildcard-matching/

Please check the redirect before merging


Erisa:
The redirect works, hurray

ngayerie (approved):

Erisa:
Added the / on the end of the redirect to better align with the other redirects, it still works

zeinjaber (approved):

kodster28 (approved):

PR #10928, opened 2023-09-27
[Workers] Clarify handling of fetch Request GET payloads

jasnell:


content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
 
 Using any other type of `ReadableStream` as the body of a request will result in Chunked-Encoding being used.
 
+### Special considerations for `GET` requests with a payload body
+
+The HTTP and `fetch` specification define that `GET` requests *should not* have a payload
```
irvinebroque: link to this in spec?

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+body and that any payload that is included has no semantic meaning or value. Nor should a
+`GET` request ever contain a `content-length` header field, even if the value of that field
+is zero. The `fetch` specification even goes so far as to make it impossible to create a new
+`Request` object using the constructor if the `method` is `GET` or `POST` and the `body` is
+not `null`.
+
+For instance, the following will result in an error being thrown:
```
irvinebroque: Needs caveat of "according to spec" or something, otherwise sounds like this will throw error in Workers

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+
+```js
+const req = new Request('https://example.com', { method: 'GET', body: new Uint8Array(0) });
+```
+
+However, the reality is that there are deployed applications on the web that do -- despite
+what the standards say -- include a body payload in `GET` requests. Workers will allow for
```
irvinebroque:
```suggestion
what the standards say -- include a body payload in `GET` requests. Workers allow for
```

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+this by accepting such requests, and even forwarding those on to origins.
```
irvinebroque:
```suggestion
this by accepting such requests, and allows you to forward them to origins.
```

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+
+When a `GET` request with a payload of any length is received by workers, the `request.body`
```
irvinebroque:
```suggestion
When a `GET` request with a payload of any length is received by your Worker, the `request.body`
```

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+property will be a `ReadableStream` rather than `null` as would typically be expected per
```
irvinebroque:
```suggestion
property will be a [`ReadableStream`](/workers/runtime-apis/streams/readablestream/) rather than `null` as would typically be expected per
```

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+the `fetch` specification. This must be taken into consideration when creating a new
```
irvinebroque: Link to part of `fetch` spec that defines this?

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,51 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+      body: request.method === 'GET' || request.method === 'HEAD' ? null : request.body,
+    });
+  }
+}
+```
+
+Alternatively, calling `new Request(request)` will work to pass the `GET` request with
+the payload on through the new request. This should be done with caution, however, as existing
+HTTP infrastructure on the web can handle such payloads inconsistently. If an intermediary or
+origin server is not implemented to handle `GET` payloads correctly, bugs or security
```
irvinebroque: Can we be more specific than "not implemented to handle correctly"?

irvinebroque: Anything to link to re: vulerabilities this could introduce?

irvinebroque (approved):

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,55 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
 Using any other type of `ReadableStream` as the body of a request will result in Chunked-Encoding being used.
 
+### Special considerations for `GET` requests with a payload body
+
+The HTTP and `fetch` specification define that `GET` requests
+[*should not*](https://httpwg.org/specs/rfc9110.html#GET) have a payload body and that any
+payload that is included has no semantic meaning or value. Nor should a `GET` request ever
+contain a `content-length` header field, even if the value of that field is zero. The `fetch`
+specification even goes so far as to make it impossible to create a new `Request` object using
+the constructor if the `method` is `GET` or `POST` and the `body` is not `null`.
```
kentonv:
```suggestion
the constructor if the `method` is `GET` or `HEAD` and the `body` is not `null`.
```

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,55 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+
+For instance, per the `fetch` spec, the following must result in an error being thrown:
+
+```js
+const req = new Request('https://example.com', { method: 'GET', body: new Uint8Array(0) });
```
kentonv, line 410: When writing a code example that is wrong or doesn't work, I like to add a comment flagging it as wrong, so that people who are skimming and read the code without reading the surrounding text don't get mislead.

```suggestion
// DOESN'T WORK: GET requests cannot have a body.
const req = new Request('https://example.com', { method: 'GET', body: new Uint8Array(0) });
```

content/workers/runtime-apis/request.md
```diff
@@ -394,6 +394,55 @@ A `FixedLengthStream` is an identity `TransformStream` that permits only a fixed
...
+[per the `fetch` specification](https://fetch.spec.whatwg.org/#dom-request). This must be taken
+into consideration when creating a new `Request` from the original.
+
+Specifically, when using the following pattern, care must be taken to avoid errors:
+
+```js
+export default {
+  async fetch(request) {
+    // ...
+    const newRequest = new Request(request.url, {
```
kentonv, line 429: Hmm, I think we should push people away from this pattern altogether, since it has lots of other problems from not forwarding all properties of the request. Using it as an example here could mislead people into thinking it's OK. Is there a way we can present this without implying that?

irvinebroque: Let's add DO and DON'T examples, I think that would go a long way and serve as really valuable reference.

jasnell: @kentonv ... do you have a suggestion for a better example?

kentonv: I think I agree with @irvinebroque, we should have a "DON'T" example that does `{method: request.method, body: requset.body}`, and then a "DO" example that passes the whole request object to the new request's constructor.

I don't think we should bother suggesting a version of the code which drops the body for GET and HEAD requests, as this current code does. The body might be important, so this could be corrupting the request. If someone understands what they are doing well enough to be sure that this is actually the behavior they want, I think they can figure it out.

kodster28:
@jasnell, was this fine to merge in?

Can't exactly tell whether the comment from Kenton was addressed or not.

KimJ15:

PR #10929, opened 2023-09-27
[workers-ai] Fixes and copy on RAG tut

kristianfreeman:
Thanks @KianNH for surfacing


KianNH:
There's an incorrect `index = "vector-index"` in the `vectorize` binding (should be `index_name`) which might be worth fixing whilst we're already here.

kristianfreeman:
thanks, fixed!

KianNH:
The `[[ai]]` looks like it should be `[ai]` too (not an array).

KianNH:
The Hono examples (which people might copy/paste) are missing `import { Ai } from '@cloudflare.com/ai'`, and it would be cool for Tab Switcher & TS examples (mainly to show people how to use TS with Hono, really) but that might be a thing for a second pass PR.

irvinebroque (approved):

PR #10935, opened 2023-09-27
[Chore] Fix for uppercased filenames

kodster28:


irvinebroque (approved):

kodster28:
Actually, closing and re-opening to just address the filename, not the check. It's a simpler fix.

PR #10945, opened 2023-09-27
[Style guide] Added glossary entry info

kodster28:


haleycode (approved):

kodster28:
Thanks https://github.com/kubernetes/website/tree/main / @RinkiyaKeDad for helping us get here btw.

Totally used y'alls current shortcodes as inspiration for [our shortcodes](https://github.com/cloudflare/cloudflare-docs/blob/production/layouts/shortcodes/glossary-definition.html) (though our [implementation](https://github.com/cloudflare/cloudflare-docs/blob/production/layouts/partials/glossary-definition.html) is a bit different since the source of truth lives in YAML as opposed to rendered on a page).

PR #10992, opened 2023-09-29
[Logs] Clarify API token permissions for R2 logpush

irvinebroque:
We currently tell people their API token needs "write permissions"

When creating an API token for R2, the options actually presented in the dashboard are either Read or Edit — there is no "Write" option:

Screenshot 2023-09-28 at 5 27 40 PM

This PR changes this to be "Edit" so that someone following the docs is clear on what they need to do.

### DIFF
--- a/content/logs/get-started/enable-destinations/r2/index.md
+++ b/content/logs/get-started/enable-destinations/r2/index.md
@@ -33,7 +33,7 @@ Before getting started:
 
     - Zone scope, logs edit permissions.
 
-    - Account scope, R2 write permissions.
+    - Account scope, R2 edit permissions.
 
 ## Manage via the Cloudflare dashboard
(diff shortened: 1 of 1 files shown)


irvinebroque:
Actually, I'm not sure we should include this section at all?

https://github.com/cloudflare/cloudflare-docs/blob/a0d24b06d01036d0d1f78a915b87e4ffa335f780/content/logs/get-started/enable-destinations/r2/index.md?plain=1#L32-L36

Since when creating a Logpush job, there is no path in the dashboard to enter an API token.

Screenshot 2023-09-28 at 5 30 44 PM

angelampcosta:
Docs updated in https://github.com/cloudflare/cloudflare-docs/pull/12110

PR #10999, opened 2023-09-29
Update storage-options.md

elithrar:
Remove D1 alpha banner.


KianNH:
It looks like there's also an alpha banner on the Wrangler commands page 

https://github.com/cloudflare/cloudflare-docs/blob/production/content/workers/wrangler/commands.md?plain=1#L123

deadlypants1973 (approved):

PR #11013, opened 2023-09-29
Add redirects for constellation /get-started

irvinebroque:
PCX-8514

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -1114,6 +1114,7 @@
 # Constellation
 /constellation/platform/wrangler/ /constellation/ 301
 /constellation/platform/ /constellation/ 301
+/constellation/get-started/* /workers-ai/ 301
 
 # zaraz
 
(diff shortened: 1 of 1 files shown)


content/_redirects
```diff
@@ -1114,6 +1114,7 @@
 # Constellation
 /constellation/platform/wrangler/ /constellation/ 301
 /constellation/platform/ /constellation/ 301
+/constellation/get-started/* /workers-ai/ 301
```
kodster28: We need all the dynamic redirects at the bottom of the page. Otherwise, it breaks the static ones.

kodster28 (changes requested):

content/_redirects
irvinebroque: Ah okay, will fix

kodster28: Yeah, admittedly, the info in the pages docs isn't abundantly clear

kodster28: Actually, @irvinebroque, I just pulled down your PR and moved it myself.

kodster28 (approved):

kodster28:
also trying to reformat the redirects instructions in #11017

PR #11026, opened 2023-09-30
Fix incorrect code example in build-a-retrieval-augmented-generation-ai.md

jcytong:
Fix incorrect call for Worker AI text generation call

The field name should be `messages` instead of `inputs` according to API Schema
[LLM Text Generation API Schema](https://developers.cloudflare.com/workers-ai/models/llm/#api-schema)

```const answer = await ai.run(
      '@cf/meta/llama-2-7b-chat-int8',
      {
        messages: [
        inputs: [
          { role: 'user', content: `What is the square root of 9?` }
        ]
      }
    )
```


pdwittig (approved):

Erisa (approved):

pdwittig:
Thanks @jcytong

PR #11033, opened 2023-10-02
[Vectorize] Update client-api.md

leaysgur:
`details()` does not exists and it is may be `describe()`?

> https://github.com/cloudflare/workerd/blob/65230276cb75a62381bf29d5059533958a311ebd/src/cloudflare/internal/vectorize.d.ts#L116-L120


Maddy-Cloudflare (approved):

elithrar:
Thanks @leaysgur!

PR #11036, opened 2023-10-02
Update create-api.md

AlphaK5:
doc improvement 

PCX-8541


pedrosousa:
Hi @AlphaK5 
I updated your proposal, converting it to a "Next steps" section, both for rate limiting rules and custom rules. This way we A) don't repeat existing content, and B) unblock users that would like to perform other tasks after creating a rule but didn't know where to go.

pedrosousa (approved):

PR #11077, opened 2023-10-04
Update vary-for-images.md

justinallenmarsh:
There is an extra space in the first two curl requests which is not a valid url for the pull request.


angelampcosta (approved):

angelampcosta:
@jmarsh24 Thanks for pointing this out!

PR #11098, opened 2023-10-04
[Rules] Remove note about being unable to remove x-forwarded-for

KianNH:
Whilst you are still unable to modify the `x-forwarded-for` header, it can be removed in both a normal Transform Rule and Managed Transforms.


pedrosousa:
Quick update: Still confirming internally if this is true in all situations.

pedrosousa:
Hi @KianNH 
Thank you for the PR. However, after some internal discussions and testing, we decided to keep the paragraph as it is. At this point it is not guaranteed that the header would actually be removed in all situations.

PR #11134, opened 2023-10-05
[CF1] Consolidate cloudflared installation instructions

DevinCarr:
Favor Linux installation instructions for cloudflared that use the available package manager instead of `dpkg -i`-based installs. Additionally captures update instructions of cloudflared to simplify updates.

Backported many of these install instructions into the tutorials to reference the downloads page of cloudflared instead of using the manual `dpkg -i` method.

Update "Deprecated releases" language to match that is listed in cloudflared's main repo.

- TUN-7864


ranbel:
edited and merged in #11143

PR #11135, opened 2023-10-06
[ZT] Tunnel private networks

ranbel:
pcx-7559


sudarshan-reddy:
Can you also add @DevinCarr , @joliveirinha , @sudarshan-reddy , @jcsf and @chungthuang as reviewers please? I tried, but don't have permission to do it

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+pcx_content_type: how-to
+title: Site-to-site connectivity
+weight: 5
+meta:
+  title: Set up WARP connector
+---
+
+# Set up WARP connector
+
+Cloudflare WARP connector is a piece of software that enables site-to-site, bidirectional, and mesh networking connectivity without requiring changes to underlying network routing infrastructure. WARP connector establishes a secure Layer 3 connection between a private network and Cloudflare, allowing you to:
```
sudarshan-reddy: > piece of software

Perhaps indicate instead that it is an extension of Cloudflare Warp instead?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+
+- Connect two or more private networks to each other.
+- Connect IoT devices that cannot run external software, such as printers and IP phones.
+- Filter and log traffic server-initiated traffic, such as VoIP and SIP traffic.
+- Apply Zero Trust security policies based on the source IP of the request.
+
+```mermaid
```
sudarshan-reddy: Can this show multiple subnets behind warp connectors? The examples could even contain these bullet points (like one of them could be a SIP Phone or SIP server and maybe an IOT device.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+        router2["Device running 
+        WARP connector
+        192.168.1.97"]<-->device2["Device
+        192.168.1.100"]
+      end
+      router1<--->C((Cloudflare))<--->router2
+
+```
+
+As shown in the diagram, WARP connector acts as a router for a subnet within the private network to on-ramp and off-ramp traffic through Cloudflare. All devices on the subnet can access any services connected to Cloudflare, and all devices connected to Cloudflare can access any services on the subnet. Each subnet runs a WARP connector on a designated Linux machine (typically the default gateway router), but other devices on the network do not need to install software.
```
sudarshan-reddy, line 20: > install software

install warp-connector perhaps?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+5. Select **Save**.
+
+## 3. Turn on WARP-to-WARP
+
+To allow a WARP connector to connect to services behind another WARP connector:
+
+1. In [Zero Trust](https://one.dash.cloudflare.com), go to **Settings** > **Network**.
+2. Enable **Warp-to-Warp**.
+
+## 4. Create a WARP connector
```
sudarshan-reddy: Can we do these instructions with an example?

Maybe we can start by saying something like:

Let's think of a scenario where two subnets say 169.159.0.0/16 and 10.0.0.0/16 have no connectivity with each other.

The subsequent examples (and the flow chart above could use these specific ips. Ideally we want users to simply replace these ips with theirs in these steps and get stuff set up.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+
+Each subnet requires its own WARP connector. To create a new WARP connector:
+
+1. In [Zero Trust](https://one.dash.cloudflare.com), go to **Network** > **Tunnels**.
+2. Select **Create a tunnel**.
+3. For the connector type, select **WARP**. Select **Next**.
+4. A window will appear with a list of prerequisites. Select **Confirm** to continue.
+5. Give the tunnel any name and select **Save tunnel**.
+
+## 5. Install the WARP connector
```
sudarshan-reddy: Let us maybe (if my suggestion above is accepted) continue to use an example of two subnets here too. In this case, we should tell them they need to have run these steps twice for each of the subnets.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+
+WARP connector is only supported on Linux.  Installing on your router is the simplest setup, but if you do not have access to the router, you may choose any other machine on your network. For system requirements, refer to the [WARP downloads page](/cloudflare-one/connections/connect-devices/warp/download-warp/#linux).
+
+To install the WARP connector:
+
+1. On the **Install and run connectors** screen, select the operating system of your host machine.
+2. Copy-paste the command into a terminal window and run the command.
+3. To authenticate the WARP connector to your Zero Trust organization:
+    1. In `/var/lib/cloudflare-warp`, create an `mdm.xml` file using any text editor:
```
sudarshan-reddy: Opinionated comment, why not simply

sudo echo "<dict>
<key>organization</key>
<string><TEAM-NAME></string>
<key>auth_client_id</key>
<string><CLIENT__ID></string>
<key>auth_client-secret</key>
<string><CLIENT-SECRET></string>
<key>warp_connector_token</key>
<string><WARP-CONNECTOR-TOKEN></string>
</dict>"  > /var/lib/cloudflare-warp/mdm.xml 

?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+      <string><CLIENT__ID></string>
+      <key>auth_client-secret</key>
+      <string><CLIENT-SECRET></string>
+      <key>warp_connector_token</key>
+      <string><WARP-CONNECTOR-TOKEN></string>
+      </dict>
+      ```
+    3. To verify the registration, run the following command:
+      ```sh
+      $ warp-cli account
```
sudarshan-reddy, line 105: Maybe warp-cli status instead? It should have to say 

```
Success
Connected
```

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+      ```
+
+      If the registration did not go through, [restart the WARP service](#restart-warp).
+{{<Aside type="warning" header="Warning">}}
+If you are managing the deployment remotely over SSH, your connection may drop when you register the WARP connector. Because the connector immediately starts forwarding traffic to Cloudflare, your SSH client will see a Cloudflare IP instead of the server's public IP and terminate the connection. You can work around this issue by temporarily adding your home IP to your [Split Tunnel Exclude list](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/).
+
+{{</Aside>}}
+4. Select **Next**.
+
+## 6. Add IP routes
```
sudarshan-reddy: If we accept my suggestion above , we can modify this a bit with concrete examples of what IP routes to add.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+  $ sudo iptables -t mangle -A FORWARD -i CloudflareWARP -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu
+  $ sudo iptables -t mangle -A FORWARD -o CloudflareWARP -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu
+
+  ```
+
+## 8. Configure other devices on the network
+
+### Default gateway
+
+### Alternate gateway
```
sudarshan-reddy: I know this is TODO. Using the concrete example idea means we will show ip route commands with those IPs.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,178 @@
...
+
+## 9. Test the WARP connector
```
sudarshan-reddy: Using the concrete example idea would mean this test will use those ips.

sudarshan-reddy:
Minor nit. This PR says [ZT] Warp Connector but has changes that are not related to that work as well. Should we split it into multiple PRs so it's more focussed, and you can land those other changes irrespective of reviews/changes here?

sudarshan-reddy:
We should also have a FAQ/Gotcha page for warp-connector. Something to tell users of known quirks/gotchas approaches like ssh losing connectivity when traffic gets on-ramped to the connector as well

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: Will defer to @abelinkinbio on this. When we discussed it yesterday, it sounded like we wanted to position WARP Connector as a standalone tool.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: Yes we could. What would be a realistic example?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: I purposely used a generic term here to cover all types of software (WARP connector, WARP client, cloudflared, etc..).

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
abelinkinbio: Yeah I'm not hard pressed on it, but I think making the distinction of being a extension of warp will cause more confusion than clarity.

sudarshan-reddy: You are right. However, I think there could be a certain percentage of users for whom it might be the other way. Perhaps maybe we could just add a link or footnote to this text?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: Could there be a situation where registration is successful, but status is Disconnected?

sudarshan-reddy: Of course. In that case we should simply add a `warp-cli connect` to the instructions. My point here is that `warp-cli account` seems a bit circuitous to a reader

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,288 @@
...
+      end
+      router1<--->C((Cloudflare))<--->router2
+
+```
+
+As shown in the diagram, WARP connector acts as a router for a subnet within the private network to on-ramp and off-ramp traffic through Cloudflare. All devices on the subnet can access any services connected to Cloudflare, and all devices connected to Cloudflare can access any services on the subnet. Each subnet runs a WARP connector on a designated Linux machine (typically the default gateway router), but other devices on the network do not need to install software.
+
+This guide will cover how to connect two independent subnets, for example `10.0.0.0/24` and `192.168.1.0/24`.
+
+## 1. Create a service token
```
sudarshan-reddy, line 36: Opinonated comment: I think steps 1-7 and 9 and 10 will be guided by the UI. Can we just say GO here and follow the instructions? The UI is intuitive enough for users to do this step by step.

And we shouldn't have to do 8 in the middle of the UI setup. It can happen afterwards.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,288 @@
...
+If you do not already have a private network range, you can choose one of these [pre-defined CIDRs](https://datatracker.ietf.org/doc/html/rfc1918#section-3).
+{{</Aside>}}
+
+11. Select **Save Tunnel**.
+
+12. In your [Split Tunnel configuration](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/), ensure that your CIDR is routing through the WARP tunnel. For instructions on how to do this, refer to [Route private network IPs through WARP](/cloudflare-one/connections/connect-networks/private-net/cloudflared/#3-route-private-network-ips-through-warp).
+
+13. Repeat these steps for subnet `192.168.1.0/24`.
+
+## 7. Configure the host
```
sudarshan-reddy: Proof reading. This should be ##5

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,288 @@
...
+
+Run the following commands on each machine where you installed WARP connector.
+
+1. Enable IP forwarding:
```
sudarshan-reddy, line 201: Or "Ensure IP forwarding is enabled"?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
sudarshan-reddy: also cc: @abelinkinbio

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,288 @@
...
+
+  To set this up more permanently (persist between reboots):
+
+  ```sh
+  $ echo 'net.ipv4.ip_forward = 1' | sudo tee -a /etc/sysctl.d/99-warp-svc.conf
+  $ sudo sysctl -p /etc/sysctl.d/99-warp-svc.conf
+  ```
+
+{{<Aside type="note" header="IP forwarding on VPC">}}
+If you are setting up WARP connector on a [virtual private cloud (VPC)](/learning/cloud/what-is-a-virtual-private-cloud/), you may need to enable IP forwarding on the VM instance. Here is an example of how to do this in [Google Cloud](https://cloud.google.com/vpc/docs/using-routes#canipforward).
```
sudarshan-reddy: I think we should not send them to a Google Cloud link. Ideally this link can be something of an internal tutorial that we will put up in the future.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: I don't think there's any harm in providing extra guidance. All of our other docs, including the [dashboard tunnel guide](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/get-started/create-remote-tunnel/), walk users through steps in the UI.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: Usually we prefer to link out to external third-party documentation instead of duplicating their content. We could remove the second sentence if we want the note to be platform-agnostic?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+      </dict>
+      ```
+    3. To verify the registration:
+      ```sh
+      $ warp-cli account
+      ```
+
+      If the registration did not go through, [restart the WARP service](#restart-warp).
+{{<Aside type="warning" header="Warning">}}
+If you are managing the deployment remotely over SSH, your connection may drop when you register the WARP connector. Because the connector immediately starts forwarding traffic to Cloudflare, your SSH client will see a Cloudflare IP instead of the server's public IP and terminate the connection. You can work around this issue by temporarily adding your home IP to your [Split Tunnel Exclude list](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/).
```
DevinCarr:
```suggestion
If you are managing the deployment remotely over SSH, your connection may drop when you register the WARP connector. Because the connector immediately starts forwarding traffic to Cloudflare, the remote SSH server's traffic will be routed to Cloudflare instead of via the server's public IP and will timeout your existing connection. You can work around this issue by temporarily adding your home IP to your [Split Tunnel Exclude list](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/).
```

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+6. Select the operating system of your host machine.
+7. Copy-paste the command into a terminal window and run the command.
+8. To authenticate the WARP connector to your Zero Trust organization:
+    1. Create an `mdm.xml` file in `/var/lib/cloudflare-warp` using any text editor:
+
+      ```sh
+      $ cd /var/lib/cloudflare-warp
+      $ sudo vim mdm.xml
+      ```
+    2. Add the following text to the file. Make sure to fill in your {{<glossary-tooltip term_id="team name">}}team name{{</glossary-tooltip>}}, the Client ID and Client Secret of your [service token](#1-create-a-service-token), and the WARP connector token value (shown in the dashboard).
```
DevinCarr:
```suggestion
    2. Add the following text to the file. Make sure to fill in your {{<glossary-tooltip term_id="team name">}}team name{{</glossary-tooltip>}}, the Client ID and Client Secret of your [service token](#1-create-a-service-token), and the WARP connector token value (shown in the dashboard). As soon as you save this file the WARP client will attempt to register with the provided credentials.
```

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+
+      ```txt
+      ---
+      filename: /var/lib/cloudflare-warp/mdm.xml
+      ---
+      <dict>
+      <key>organization</key>
+      <string><TEAM-NAME></string>
+      <key>auth_client_id</key>
+      <string><CLIENT__ID></string>
```
DevinCarr:
```suggestion
      <string><CLIENT-ID></string>
```

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+      <key>auth_client-secret</key>
```
DevinCarr:
```suggestion
      <key>auth_client_secret</key>
```

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+
+11. Select **Save Tunnel**.
+
+12. In your [Split Tunnel configuration](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/), ensure that your CIDR is routing through the WARP tunnel. For instructions on how to do this, refer to [Route private network IPs through WARP](/cloudflare-one/connections/connect-networks/private-net/cloudflared/#3-route-private-network-ips-through-warp).
+
+13. Repeat these steps for subnet `192.168.1.0/24`.
+
+## 5. Configure the host
+
+Run the following commands on each machine where you installed WARP connector.
```
DevinCarr: We don't currently support HA replicas for Warp Connector like we do for cloudflared. Do we want to adjust this wording? Or is it plural because the install guide talks about installing two connectors for the two different subnets?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+{{<tab label="macos" no-code="true">}}
+
+```sh
+$ sudo route -n change default <WARP-CONNECTOR-IP> -interface en0
+```
+
+{{</tab>}}
+
+{{<tab label="windows" no-code="true">}}
+
```
DevinCarr, line 322: We need some Windows Examples. I'm adding this comment to remind us to ask the client team what their recommendation might be.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+{{</tab>}}
+{{</tabs>}}
+
+{{<Aside type="note">}}
+`100.96.0.0/12` is the CIDR for all user devices running [Cloudflare WARP](/cloudflare-one/connections/connect-devices/warp/). If you set `<DESTINATION-IP>` to `100.96.0.0/12`, this configures the server to connect to user devices through Cloudflare.
```
DevinCarr:
```suggestion
`100.96.0.0/12` is the default CIDR for all user devices running [Cloudflare WARP](/cloudflare-one/connections/connect-devices/warp/). If you set `<DESTINATION-IP>` to `100.96.0.0/12`, this configures the server to connect to user devices through Cloudflare.
```

The subnet can be adjusted in certain use-cases, so we want to refer to it as the default.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+        router2["Device running 
+        WARP connector
+        192.168.1.97"]-->device2["Device
+        192.168.1.100"]
+      end
+      router1-->C((Cloudflare))-->router2
+```
+
+{{<Aside type="note">}}
+If you are testing with curl, make sure to add the `--ipv4` to your curl commands.
```
DevinCarr:
```suggestion
If you are testing with curl using private hostnames, make sure to add the `--ipv4` to your curl commands.
```

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+{{</Aside>}}
+
+## Troubleshooting
```
DevinCarr: @abelinkinbio do we want to list the current WARP Connector limitations on this page as well? Also, which would you like to outline?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+      If the registration did not go through, [restart the WARP service](#restart-warp).
+{{<Aside type="warning" header="Warning">}}
+If you are managing the deployment remotely over SSH, your connection may drop when you register the WARP connector. Because the connector immediately starts forwarding traffic to Cloudflare, your SSH client will see a Cloudflare IP instead of the server's public IP and terminate the connection. You can work around this issue by temporarily adding your home IP to your [Split Tunnel Exclude list](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/).
+
+{{</Aside>}}
+9. Select **Next**.
+10. In **CIDR**, enter the private IPv4 address range that you wish to route through this WARP connector (for example, `10.0.0.0/24`). IPv6 is not supported at this time.
+
+{{<Aside type="note">}}
+If you do not already have a private network range, you can choose one of these [pre-defined CIDRs](https://datatracker.ietf.org/doc/html/rfc1918#section-3).
```
DevinCarr:
```suggestion
If you do not already have a private network range, you can choose a subnet from one of these [pre-defined CIDRs](https://datatracker.ietf.org/doc/html/rfc1918#section-3).
```

content/cloudflare-one/connections/connect-networks/private-net/warp-to-warp.md
```diff
@@ -1,29 +1,30 @@
...
+meta:
+  title: Create private networks with WARP-to-WARP
 ---
 
-# Create private networks with WARP
-
+# Create private networks with WARP-to-WARP
 With Cloudflare Zero Trust, you can create a private network between any two or more devices running Cloudflare WARP. This means that you can have a private network between your phone and laptop without ever needing to be connected to the same physical network. If you already have an existing Zero Trust deployment, you can also enable this feature to add device-to-device connectivity to your private network with the press of a button. This will allow you to connect to any service that relies on TCP, UDP, or ICMP-based protocols through Cloudflare’s network.
 
 Users in your organization can reach these services by enrolling into your organization's Zero Trust account. Once enrolled, each device is assigned a virtual IP address in the `100.64/10` range which will allow users or systems to address these devices directly. Administrators will then be able to build Zero Trust policies to determine who within your organization can reach those virtual IPs.
```
DevinCarr:
```suggestion
Users in your organization can reach these services by enrolling into your organization's Zero Trust account. Once enrolled, each device is assigned a virtual IP address in the `100.96.0.0/12` range which will allow users or systems to address these devices directly. Administrators will then be able to build Zero Trust policies to determine who within your organization can reach those virtual IPs.
```
@sudarshan-reddy @joliveirinha this range is 100.96/12 right? That's what we have in tunnelstore for default WARP subnet ranges.

sudarshan-reddy:
For every step, this guide should show the user how to verify that they're on the right path. Lets work on that together.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
sudarshan-reddy: and warp-account is also not foolproof btw. It could even show up details when warp-cli is not connected but just registered.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+        device["Device
+        10.0.0.2"]--default gateway-->router(["Router
+        10.0.0.1"])
+        device--alternate gateway-->warp["Device running
+        WARP connector
+        10.0.0.100"]
+
+      end
+      router--->I{Internet}
+      warp--->C((Cloudflare))
```
sudarshan-reddy: warp here still goes through the router before the Internet/Cloudflare here.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: How about we mention both warp-cli account and warp-cli connect, similar to what we have in the WARP docs?https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/manual-deployment/#linux

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+6. Select the operating system of your host machine.
+7. Copy-paste the command into a terminal window and run the command.
+8. To authenticate the WARP connector to your Zero Trust organization:
+    1. Create an `mdm.xml` file in `/var/lib/cloudflare-warp` using any text editor:
+
+      ```sh
+      $ cd /var/lib/cloudflare-warp
+      $ sudo vim mdm.xml
+      ```
+    2. Add the following text to the file. Make sure to fill in your {{<glossary-tooltip term_id="team name">}}team name{{</glossary-tooltip>}}, the Client ID and Client Secret of your [service token](#1-create-a-service-token), and the WARP connector token value (shown in the dashboard). As soon as you save this file the WARP client will attempt to register with the provided credentials.
```
abelinkinbio: I would change this to something like: "As soon as you save this file, warp-svc should automatically detect this change and register. However, to ensure this behavior run sudo systemctl restart warp-svc.service

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
abelinkinbio: It's also worth showing what a successful registration event should return after running warp-cli account

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+
+{{<Aside type="note">}}
+If you do not already have a private network range, you can choose a subnet from one of these [pre-defined CIDRs](https://datatracker.ietf.org/doc/html/rfc1918#section-3).
+{{</Aside>}}
+
+11. Select **Save Tunnel**.
+
+12. In your [Split Tunnel configuration](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/), ensure that your CIDR is routing through the WARP tunnel. For instructions on how to do this, refer to [Route private network IPs through WARP](/cloudflare-one/connections/connect-networks/private-net/cloudflared/#3-route-private-network-ips-through-warp).
+
+13. Repeat these steps for subnet `192.168.1.0/24`.
```
abelinkinbio: We could probably be more specific here. Something along the lines of repeat these steps to create an additional tunnel, connector, and IP/CIDR for the network segment 192.168/24

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+      $ warp-cli account
+      ```
+
+      If the registration did not go through, [restart the WARP service](#restart-warp).
+{{<Aside type="warning" header="Warning">}}
+If you are managing the deployment remotely over SSH, your connection may drop when you register the WARP connector. Because the connector immediately starts forwarding traffic to Cloudflare, the remote SSH server's traffic will be routed to Cloudflare instead of via the server's public IP and will timeout your existing connection. You can work around this issue by temporarily adding your home IP to your [Split Tunnel Exclude list](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/).
+
+{{</Aside>}}
+9. Select **Next**.
+10. In **CIDR**, enter the private IPv4 address range that you wish to route through this WARP connector (for example, `10.0.0.0/24`). IPv6 is not supported at this time.
```
abelinkinbio: Let's say IPv6 routes for your Tunnel to be specific.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+        10.0.0.100"]
+
+      end
+      router--->I{Internet}
+      warp--->C((Cloudflare))
+```
+
+#### Route all traffic
+
+You can configure all traffic on a device to egress through the WARP connector. All traffic will be filtered by your Gateway network policies.
```
abelinkinbio: We should be more specific here about whether this is the WARP GNCAT IP or the local IP

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### Route specific IPs
+
+You can configure only certain routes to egress through WARP connector. For example, you may only want to filter traffic destined to internal applications and devices, but allow public Internet traffic to bypass Cloudflare.
+
+{{<tabs labels="Linux | macOS | Windows ">}}
+{{<tab label="linux" no-code="true">}}
+
+```sh
+$ sudo route -n add -net <DESTINATION-IP> <WARP-CONNECTOR-IP>
```
abelinkinbio: this should be sudo **ip** route

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: What would you suggest for the wording?

abelinkinbio: Think this its mostly fine. Should just call out that this is the local IP.

content/cloudflare-one/connections/connect-networks/private-net/warp-to-warp.md
ranbel: Did we confirm if the range is actually `100.96.0.0/12`? If so we will also need to update the IPs in the "Enable WARP-to-WARP" and "Connect via WARP" sections.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: I don't have strong opinions about this, but we use the text editor approach in other parts of the docs, e.g. https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/get-started/create-local-tunnel/#4-create-a-configuration-file.
It may be easier for someone to modify their team name, service token info, etc. in a text editor.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: does the updated diagram make sense?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,316 @@
...
+
+12. In your [Split Tunnel configuration](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/split-tunnels/), ensure that your CIDR is routing through the WARP tunnel. For instructions on how to do this, refer to [Route private network IPs through WARP](/cloudflare-one/connections/connect-networks/private-net/cloudflared/#3-route-private-network-ips-through-warp).
+
+13. Repeat these steps to install an additional WARP connector on subnet `192.168.1.0/24`.
+
+## 5. Configure the host
+
+Run the following commands on the machine where you installed WARP connector. You will need to configure the host machine on each subnet.
+
+1. Ensure that IP forwarding is enabled:
```
ranbel: Could we just have the second option? How many people would do just temporary forwarding?

abelinkinbio: It's probably not necessary. Think it's thoughtful to provide both, but if we have something driving the removal I'm open to it.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: I think we are postponing these details until GA.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,272 @@
...
+  $ sudo sysctl -p /etc/sysctl.d/99-warp-svc.conf
+  ```
+
+{{<Aside type="note" header="IP forwarding on VPC">}}
+If you are setting up WARP connector on a [virtual private cloud (VPC)](/learning/cloud/what-is-a-virtual-private-cloud/), you may need to enable IP forwarding on the VM instance.
+{{</Aside>}}
+
+2. WARP's [virtual interface](/cloudflare-one/connections/connect-devices/warp/configure-warp/route-traffic/warp-architecture/#virtual-interface) has a [maximum transmission unit (MTU)](/learning/network-layer/what-is-mtu/) of 1280 bytes, whereas the standard Ethernet MTU is 1500 bytes. To avoid dropping packets that exceed 1280 bytes, clamp the [maximum segment size (MSS)](/learning/network-layer/what-is-mss/) of the host machine so that incoming payloads are less than the MTU of WARP:
+
+  ```sh
```
sudarshan-reddy: Let's note that this will not persist between restarts. So the users will have to have this script somewhere on their init if they want this change to persist between restarts. 

They have to do something like 

```bash
echo '#!/bin/bash
# Define your rules
RULES=(
  "-A FORWARD -i CloudflareWARP -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu"
  "-A FORWARD -o CloudflareWARP -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu"
)

# Apply the rules
for rule in "${RULES[@]}"; do
  iptables -t mangle $rule
done

# Save the rules
iptables-save > /etc/iptables/rules.v4
' | sudo tee /usr/local/bin/apply_iptables_rules.sh
```

and then a service

```bash
echo '[Unit]
Description=Load iptables rules at startup

[Service]
Type=oneshot
ExecStart=/sbin/iptables-restore < /etc/iptables/rules.v4

[Install]
WantedBy=multi-user.target
' | sudo tee /etc/systemd/system/iptables-persistent.service
```


perhaps this could be a drop down?

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
kcwilliamson: I wouldn't rely on the UI to navigate. If they are looking at docs, they need assistance. We should keep the addition details. Ranbel is right on this point. There's no harm here, power users can scroll through the doc if they know but folks who don't would be left out and frustrated. I wouldn't assume anything.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,367 @@
...
+    | Service Auth | Include   | Service Token | `<SERVICE-TOKEN-NAME>` |
+
+5. Select **Save**.
+
+## 3. Turn on WARP-to-WARP
+
+To allow a WARP connector to connect to services behind another WARP connector:
+
+1. In [Zero Trust](https://one.dash.cloudflare.com), go to **Settings** > **Network**.
+2. Enable **Warp-to-Warp**.
```
ranbel, line 60: When should users enable **Override local interface IP** under **Settings** > **Warp Client**?

sudarshan-reddy:
I think we should flesh out the troubleshooting a bit more. 

For instance, with `Missing Registration`, we talk about how restarting or forcing a delete will fix it. But that doesnt tell the whole story. We have to lead them to look at the logs and understand whats going on. It could even be because they didnt move the mdm.xml to the right place, formatted it poorly or forgot to device enroll the service token they generated.

abelinkinbio (approved):
Tentatively approved assuming resolution of minor comments.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
```diff
@@ -0,0 +1,367 @@
...
+
+![Default gateway routing configuration](/images/cloudflare-one/connections/connect-apps/warp-connector/default-gateway.png)
+
+### Option 2: Alternate gateway
+
+If you have access to the router but installed WARP connector on another machine, you can configure the router to forward traffic to the WARP connector. This typically involves adding a static route for the destination IPs that you want to connect to through Cloudflare. Refer to your router's documentation for specific instructions on how to add an IP route.
+
+For example, if you are on subnet `10.0.0.0/24` and want to reach applications behind subnet `192.168.1.0/24`, add a rule that routes `192.168.1.0/24` to the WARP connector IP (`10.0.0.100` in the diagram below). When a device sends a request to `192.168.1.0/24`, the router will first redirect the traffic to the WARP connector machine. WARP connector encrypts the traffic, changes its destination IP to the [WARP ingress IP](/cloudflare-one/connections/connect-devices/warp/deployment/firewall/#warp-ingress-ip), and sends it back to the router. The router will now forward this encrypted traffic to Cloudflare.
```
sudarshan-reddy, line 285: This may be obvious but let us mention that their router rules should not forward the warp-ingress ips back to the warp connector machine.

content/cloudflare-one/connections/connect-networks/private-net/_index.md
```diff
@@ -7,9 +7,17 @@ layout: single
...
 
-* Private network routes can expose both HTTP and non-HTTP resources.
-* To connect to private network resources, end users must have the [WARP client](/cloudflare-one/connections/connect-devices/warp/) installed on their device or onboard traffic via a [Magic WAN on-ramp](/magic-wan/zero-trust/cloudflare-tunnel/).
+To reach private network IPs, end users must connect their device to Cloudflare and enroll in your Zero Trust organization. The most common method is to install the [WARP client](/cloudflare-one/connections/connect-devices/warp/) on their device, or you can onboard their network traffic to Cloudflare using our WARP connector or [Magic WAN](/magic-wan/zero-trust/cloudflare-tunnel/).
```
maxvp: Is there a significant difference between `WARP` and `the WARP client`?

content/cloudflare-one/connections/connect-networks/private-net/cloudflared/private-dns.md
```diff
@@ -0,0 +1,65 @@
...
+
+Both `dig` commands will fail if the WARP client is disabled on your end user's device.
+
+## Troubleshooting
+
+Use the following troubleshooting strategies if you are running into issues while configuring your private network with Cloudflare Tunnel.
+
+- Ensure that `cloudflared` is connected to Cloudflare by visiting **Access** > **Tunnels** in Zero Trust.
+
+- Ensure that `cloudflared` is running with `quic` protocol (search for `Initial protocol quic` in its logs).
```
maxvp:
```suggestion
- Ensure that `cloudflared` is running with the `quic` protocol (search for `Initial protocol quic` in its logs).
```

content/cloudflare-one/connections/connect-networks/private-net/cloudflared/private-dns.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Ensure that the machine where `cloudflared` is running is allowed to egress via UDP to port 7844 to talk out to Cloudflare.
+
+- Ensure that end-user devices are enrolled into WARP by visiting <https://help.teams.cloudflare.com>
```
maxvp:
```suggestion
- Ensure that end-user devices are enrolled into WARP by visiting <https://help.teams.cloudflare.com>.
```

content/cloudflare-one/connections/connect-networks/private-net/cloudflared/private-dns.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Test the setup
+
+For testing, run a `dig` command for the internal DNS service:
+
+```sh
+dig AAAA www.myorg.privatecorp
+```
+
+The `dig` will work because `myorg.privatecorp` was configured above as a fallback domain. If you skip that step, you can still force `dig` to use your private DNS resolver:
```
maxvp: Nit: Can we refer to commands as "the X" or should we include "command" as well? e.g. "the `ipconfig`" vs "the `ipconfig` command"

maxvp (approved):

content/cloudflare-one/connections/connect-networks/private-net/cloudflared/private-dns.md
ranbel: I would say "the `ipconfig` command" or simply "`ipconfig`". "The `ipconfig`" sounds a bit off.

content/cloudflare-one/connections/connect-networks/private-net/warp-connector.md
ranbel: Need someone to verify these Windows commands.

PR #11161, opened 2023-10-08
Fix Header Links

helloimalastair:
For #11155.

First time doing one of these, not sure if non-employees are even allowed to do them.

Preview available [here](https://fix-header-links.cfdocs-alastair.pages.dev/).

For the glossary-related links, I don't think you **can** link to them, at least right now, since they don't expose any anchors...

### DIFF
--- a/content/rules/compression-rules/examples.md
+++ b/content/rules/compression-rules/examples.md
@@ -29,7 +29,7 @@ The following example rule will disable compression for AVIF images, based on ei
 <summary>Example API request</summary>
 <div>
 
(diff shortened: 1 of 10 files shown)


pedrosousa:
Hi @helloimalastair!

As long as you're not breaking anything, you can also fix these header links, of course. :) We usually go through a few of these (10-15 items) each week.

Regarding the glossary links, we're now using a different strategy for defining and referencing glossary entries, but we didn't update everything yet. Refer to the following page for details: https://developers.cloudflare.com/style-guide/documentation-content-strategy/component-attributes/glossary-entry/

pedrosousa (approved):

PR #11180, opened 2023-10-09
Update wildcard-dns-records.md

AlphaK5:
wildcard behavior improvement


RebeccaTamachiro:
Hey! Thanks for putting together the PR 
After considering the thread though, this seemed like a one-off doubt. In case it's actually a recurring doubt, let me know and I'll create a ticket and check with PM how to change the docs, providing more context.

PR #11197, opened 2023-10-10
Update _cloudflared-debian-install.md

edenwee:
Using single quotes will produce a literal value. In order for the variable to function correctly, double quotes must be used.


abelinkinbio (approved):

ranbel:
Thanks for flagging! This is updated in #11183.

PR #11202, opened 2023-10-10
[Pages] Fix numbered list in redirect-to-custom-domain.md

ryoheinan:
- It seems that numbered list in [this page](https://developers.cloudflare.com/pages/how-to/redirect-to-custom-domain/) (`Set up` section) is not rendered properly now. So I fixed it in this PR. 
- I formatted the entire file as well.


kodster28 (approved):
Confirmed that this is formatted correctly now. Thanks for the fix, @ryoheinan

PR #11221, opened 2023-10-10
[Pulumi] Make headers distinct to work with tutorial shortcode

kodster28:
Current behavior leads to dupe header links (and highlighted areas of side nav).

This is the lowest effort fix for now, though could revisit in the future and:
- [add a unique id](https://github.com/cloudflare/cloudflare-docs/blob/production/layouts/shortcodes/tabs.html#L2) to the end of each.... but not sure I love that for non-dupe headers.
- Rewrite the `tutorial-step` shortcode to ignore the header rendering altogether (and just look at the `.Inner`), but would that even work since it's actually already doing that? Would require updating existing tutorials to make their behavior match.

Screenshot 2023-10-10 at 2 55 23 PM


WalshyDev (approved):
fun fun

dcpena (approved):

PR #11224, opened 2023-10-11
Create isdn.yaml

ishaanvireet:


kodster28:
Howdy @ishaanvireet, we do appreciate the contribution... but think we might want to address our glossary terms in a slightly different way.

I'll flag internally and re-open if folks let me know this is relevant.

PR #11242, opened 2023-10-11
[Maintenance] Replace details HTML tags with shortcode

maxvp:


kodster28 (approved):
Did an audit check on 15 pages + Compiles is passing, so ship it!

PR #11243, opened 2023-10-11
[Workers Analytics Engine] replace powers with multiples in FAQ

typeofweb:


content/analytics/faq/wae-faqs/index.md
```diff
@@ -31,7 +31,7 @@ Thanks to sampling, the cost of writing data multiple times can be relatively lo
 
 You can use the `_sample_interval` field — again, note that this does not tell you if the results are accurate.
 
-You can tell when data is sampled at read time because sample intervals will be powers of 10, for example `20` or `700`. There is no hard and fast rule for when sampling starts at read time, but in practice reading longer periods (or more index values) will result in a higher sample interval.
+You can tell when data is sampled at read time because sample intervals will be multiples of 10, for example `20` or `700`. There is no hard and fast rule for when sampling starts at read time, but in practice reading longer periods (or more index values) will result in a higher sample interval.
```
jimhawkridge: Hi, Thanks for the PR.
You are right that this is not correct, but I think it was probably intended to say `multiples of powers of 10`.
I.e. read time sampling might sample at a rate of 10 so you will values 10, 20, 30 etc. or it might sample at a rate of 100 and you would see 100, 200, 300 etc. and so on for 1000, 10,000 etc.

If you could apply this suggestion then I will approve:
```suggestion
You can tell when data is sampled at read time because sample intervals will be multiples of powers of 10, for example `20` or `700`. There is no hard and fast rule for when sampling starts at read time, but in practice reading longer periods (or more index values) will result in a higher sample interval.
```

Thanks!

jimhawkridge (changes requested):

jimhawkridge (approved):

PR #11246, opened 2023-10-11
[Workers] Add clarification on route case sensitivity

maxwellpeterson:
- After recent changes, the host component of worker routes is no longer case sensitive
- The path component of worker routes remains case sensitive

### DIFF
--- a/content/workers/configuration/routing/routes.md
+++ b/content/workers/configuration/routing/routes.md
@@ -131,7 +131,9 @@ While they look similar to a [regex](https://en.wikipedia.org/wiki/Regular_expre
 
 - The path component of route patterns is case sensitive, for example, `example.com/Images/*` and `example.com/images/*` are two distinct routes.
 
-- The host component of route patterns is not case sensitive, for example, `example.com/*` and `Example.com/*` are equivalent routes.
+- For routes created prior to October 15th, 2023, the host component of route patterns is case sensitive, for example, `example.com/*` and `Example.com/*` are two distinct routes.
+
+- For routes created on or after October 15th, 2023, the host component of route patterns is not case sensitive, for example, `example.com/*` and `Example.com/*` are equivalent routes.
 
 A route can be specified without being associated with a Worker. This will act to negate any less specific patterns. For example, consider this pair of route patterns, one with a Workers script and one without:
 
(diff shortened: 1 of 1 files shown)


content/workers/configuration/routing/routes.md
```diff
@@ -129,7 +129,9 @@ While they look similar to a [regex](https://en.wikipedia.org/wiki/Regular_expre
 
 - Route pattern matching considers the entire request URL, including the query parameter string. Since route patterns may not contain query parameters, the only way to have a route pattern match URLs with query parameters is to terminate it with a wildcard, `*`.
 
-- Route patterns are case sensitive, for example, `example.com/Images/*` and `example.com/images/*` are two distinct routes.
+- The path component of route patterns is case sensitive, for example, `example.com/Images/*` and `example.com/images/*` are two distinct routes.
+
+- The host component of route patterns is not case sensitive, for example, `example.com/*` and `Example.com/*` are equivalent routes.
```
WalshyDev: Do we want to mention this only applies for new deploys post 15th Oct
Not that it should really matter but since this is a change in behaviour, always good to note.

maxwellpeterson: Technically this only applies to routes created after the 15th. Redeploying a worker with older routes won't change the behavior of those older routes, since we date/version routes individually.

maxwellpeterson: I added another commit that includes the API date cutoff, and agree that this is good to note.

WalshyDev (approved):
eng review

content/workers/configuration/routing/routes.md
```diff
@@ -129,7 +129,11 @@ While they look similar to a [regex](https://en.wikipedia.org/wiki/Regular_expre
 
 - Route pattern matching considers the entire request URL, including the query parameter string. Since route patterns may not contain query parameters, the only way to have a route pattern match URLs with query parameters is to terminate it with a wildcard, `*`.
 
-- Route patterns are case sensitive, for example, `example.com/Images/*` and `example.com/images/*` are two distinct routes.
+- The path component of route patterns is case sensitive, for example, `example.com/Images/*` and `example.com/images/*` are two distinct routes.
+
+- For routes created prior to October 15th, 2023, the host component of route patterns is case sensitive, for example, `example.com/*` and `Example.com/*` are two distinct routes.
```
deadlypants1973:
```suggestion
- For routes created before October 15th, 2023, the host component of route patterns is case sensitive, for example, `example.com/*` and `Example.com/*` are two distinct routes.
```
replace complex vocabulary for better readability

deadlypants1973 (approved):

deadlypants1973:
merging as per max's request

PR #11303, opened 2023-10-16
[API Shield] API Routing

patriciasantaana:
PCX-7929


content/api-shield/management-and-monitoring/api-routing.md
```diff
@@ -0,0 +1,119 @@
...
+  "errors":[],
+  "messages":[]
+}
+```
+
+### Test a Route
+
+After sending a request to your Source Endpoint, you should see the contents of the back-end service as if you called the Target Endpoint directly. 
+
+If you see unexpected results, check your Source Endpoint host, method, and path and [verify the Route](/api-shield/management-and-monitoring/api-routing/#verify-a-route) to ensure the Target Endpoint value is correct.
```
maxvp: Nit: `see` may not be inclusive. Not sure what the best alternative is... maybe `If API Shield returns unexpected results, ...`?

maxvp (approved):

content/api-shield/management-and-monitoring/api-routing.md
patriciasantaana: good catch, thanks!

PR #11339, opened 2023-10-18
[R2] Clarify presigned URLs don't work on custom domains

Erisa:
This keeps coming up in the Developer Community


content/r2/api/s3/presigned-urls.md
```diff
@@ -187,6 +187,10 @@ Presigned URLs can be generated for any S3 operation. After a presigned URL is g
 
 Choosing between presigned URLs and public buckets is dependent on your specific use case. You can also use both if your architecture should use public buckets in one situation and presigned URLs in another. It is useful to note that presigned URLs will expose your account ID and bucket name to whoever gets a copy of the URL. Public bucket URLs do not contain the account ID or bucket name. Typically, you will not share presigned URLs directly with end users or browsers, as presigned URLs are used more for internal applications.
 
+## Limitations
+
+Presigned URLs cannot be used with Custom Domains, they must be used with the `<accountid>.r2.cloudflarestorage.com` S3 API domain.
+
```
Le0Developer, line 193: Add "work-around":
```suggestion

You can use the [general purpose HMAC validation feature of the WAF instead](/ruleset-engine/rules-language/functions/#hmac-validation) (requires Pro plan or above).

```
Maybe this could be expanded into a full section with guide, but this is better than nothing.

Erisa: Thanks!

dcpena (approved):

PR #11357, opened 2023-10-18
Use ctx.req.url when fetching assets

danawoodman:
Because the fetch API expects a full URL, the example fails with:

```
Fetch API cannot load: [object Object]
```

The issues is passing the full Hono request object isn't handled by the `ASSETS.fetch` function.

Changing to `ctx.req.url` causes the assets to properly load as expected.

### DIFF
--- a/content/pages/framework-guides/deploy-a-hono-site.md
+++ b/content/pages/framework-guides/deploy-a-hono-site.md
@@ -32,7 +32,7 @@ To serve static files like CSS, image or JavaScript files, add the following to 
 
 ```javascript
 app.get("/public/*", async (ctx) => {
-  return await ctx.env.ASSETS.fetch(ctx.req.url);
+  return await ctx.env.ASSETS.fetch(ctx.req.raw);
 });
 ```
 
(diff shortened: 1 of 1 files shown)


content/pages/framework-guides/deploy-a-hono-site.md
```diff
@@ -32,7 +32,7 @@ To serve static files like CSS, image or JavaScript files, add the following to
 
 ```javascript
 app.get("/public/*", async (ctx) => {
-  return await ctx.env.ASSETS.fetch(ctx.req);
+  return await ctx.env.ASSETS.fetch(ctx.req.url);
```
WalshyDev: ctx.req.raw should work here rather, `ctx.req` will be Hono's own Request but `ctx.req.raw` will give the raw incoming Request which should work

danawoodman: ok i can update

danawoodman: Done!

WalshyDev (approved):
Tested locally and it looks good

Thanks @danawoodman !

(only code change done, no pcx needed)

WalshyDev:
@yusukebe Probably something good to add to the hono docs too in the Pages section - https://hono.dev/getting-started/cloudflare-pages

PR #11377, opened 2023-10-19
Redirect /constellation to /workers-ai

irvinebroque:
@sdayman noted this is confusing since current Constellation page doesn't mention anything about Workers AI being the right, current way to get started with AI inference on our platform.

cc @pdwittig @pedrosousa

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -1141,6 +1141,7 @@
 # Constellation
 /constellation/platform/wrangler/ /constellation/ 301
 /constellation/platform/ /constellation/ 301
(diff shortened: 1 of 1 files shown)


WalshyDev (approved):

pedrosousa:
For later: We need to understand if we still need the [Constellation Client API](https://developers.cloudflare.com/constellation/platform/client-api/) page, since it will be the only one left in the Constellation tile.

pedrosousa (approved):

PR #11381, opened 2023-10-20
Update index.md

gyterpena:
added header_Content-Type=application/json& to destination_conf without this header elastic agent will not accept this call and returns 415 wrong Content-Type header, expecting application/json


angelampcosta:
@gyterpena thank you for bringing this up.  We going to confirm with Elastic the correct configuration and update the docs accordingly.

PR #11394, opened 2023-10-20
add CLI Arguments section to C3 page

dario-piotrowicz:
Related to https://github.com/cloudflare/workers-sdk/issues/4183 as an effort to clarifying the C3 available arguments and options (without having a potentially overwhelming cli help message)

### DIFF
--- a/content/pages/get-started/c3.md
+++ b/content/pages/get-started/c3.md
@@ -80,7 +80,7 @@ $ npm create cloudflare@latest [<DIRECTORY>] [OPTIONS] [-- <NESTED ARGS..>]
     - `chatgptPlugin`: A ChatGPT plugin.
     - `openapi`: A Worker implementing an OpenAPI REST endpoint.
 
-- `--framework` {{<type>}}string{{</type>}}
+- `--framework` {{<type>}}string{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
   - The type of framework to use to create a web application (when using this option, `--type` is ignored).
 
   - The possible values for this option are:
--- a/content/pages/get-started/c3.md
+++ b/content/pages/get-started/c3.md
@@ -58,7 +58,7 @@ C3 collects any required input through a series of interactive prompts. You may 
 This is the full format of a C3 invocation alongside the possible CLI arguments:
 
 ```sh
-$ npm create cloudflare@latest [<DIRECTORY>] [OPTIONS] [-- <NESTED ARGS..>]
+$ npm create cloudflare@latest [--]  [<DIRECTORY>] [OPTIONS] [-- <NESTED ARGS...>]
 ```
 
 {{<definitions>}}
--- a/content/pages/get-started/c3.md
+++ b/content/pages/get-started/c3.md
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
 If you choose _not_ to deploy, the project will be created for you locally, and C3 will display some helpful links for further development. Go to the newly created project folder to begin development.
 
 If you choose to deploy, you will be asked to authenticate (if not logged in already), and your project will be deployed immediately. C3 will display your project's URL and some helpful links.
+
+## CLI Arguments
+
+C3 accepts CLI arguments allowing you to to specify your preference in such a was and thus skipping the relative wizard steps. Using CLI arguments you can even create C3 invocations that do not require any user interaction, making automated or repetitive uses of C3 very convenient.
+
+This is the full format of a C3 invocation alongside the possible CLI arguments:
+
+```sh
+$ npm create cloudflare@latest [<DIRECTORY>] [OPTIONS] [-- <NESTED ARGS..>]
+```
+
+{{<definitions>}}
+
+- `DIRECTORY` {{<type>}}string{{</type>}}
+  - The directory where the application should be created. The name of the application is taken from the directory name.
+- `NESTED ARGS..` {{<type>}}string[]{{</type>}}
+  - CLI arguments to pass to eventual third party CLIs C3 might invoke (in the case of full-stack applications).
+- `--type` {{<type>}}string{{</type>}}
+  - The type of application that should be created.
+
+  - The possible values for this option are:
+
+    - `webFramework`: A website or web application.
+    - `hello-world`: A basic "Hello World" Cloudflare Worker.
+    - `common`: A Cloudflare Worker which implements a common example of routing/proxying functionalities.
+    - `scheduled`: A scheduled Cloudflare Worker (triggered via Cron Triggers).
+    - `queues`: A Cloudflare Worker which is both a consumer and produced of Queues.
+    - `chatgptPlugin`: A ChatGPT plugin.
+    - `openapi`: A Worker implementing an OpenAPI REST endpoint.
+
+- `--framework` {{<type>}}string{{</type>}}
+  - The type of framework to use to create a web application (when using this option `--type` is ignored).
+
+  - The possible values for this option are:
+
+    - `angular`
+    - `astro`
+    - `docusaurus`
+    - `gatsby`
+    - `hono`
+    - `next`
+    - `nuxt`
+    - `qwik`
+    - `react`
+    - `remix`
+    - `solid`
+    - `svelte`
+    - `vue`
+
+- `--deploy` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Deploy your application after it has been created.
+
+- `--ts`{{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Use TypeScript in your application.
+
+- `--git` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Initialize a local git repository for your application.
+
+- `--open` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Open with your browser the deployed application (this option is ignored if the application is not deployed).
+
+- `-y`, `--accept-defaults` {{<type>}}boolean{{</type>}}
+
+  - Use all the default C3 options each can also be overridden by specifying it.
+
+- `--auto-update` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Automatically uses the latest version of C3.
+
+- `-v`, `--version` {{<type>}}boolean{{</type>}}
+
+  - Show version number.
+
+- `-h`, `--help` {{<type>}}boolean{{</type>}}
+
+  - Show a help message.
+
+{{<Aside type="note">}}
+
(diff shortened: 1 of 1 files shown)


XoL1507:
👍

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
 If you choose _not_ to deploy, the project will be created for you locally, and C3 will display some helpful links for further development. Go to the newly created project folder to begin development.
 
 If you choose to deploy, you will be asked to authenticate (if not logged in already), and your project will be deployed immediately. C3 will display your project's URL and some helpful links.
+
+## CLI Arguments
+
+C3 accepts CLI arguments allowing you to to specify your preference in such a was and thus skipping the relative wizard steps. Using CLI arguments you can even create C3 invocations that do not require any user interaction, making automated or repetitive uses of C3 very convenient.
```
jculvey: Suggested rewording:

```
C3 collects any required input through a series of interactive prompts. You may also specify your choices via command line arguments, which will skip these prompts. To use C3 in a non-interactive context such as CI, specify all required arguments via the command line.
```

dario-piotrowicz: updated 🙂👍

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+
+  - Show version number.
+
+- `-h`, `--help` {{<type>}}boolean{{</type>}}
+
+  - Show a help message.
+
+{{<Aside type="note">}}
+
+All the boolean options above can be specified with or without a value, for example `--open` and `--open true` have the same effect, prefixing `no-` to the option's name negates it, so for example `--no-open` and `--open false` have the same effect.
```
admah, line 136: Great callout here!

admah (approved):
This is great. Can we also put this info in the C3 README so it shows up in NPM?

jculvey (approved):

dario-piotrowicz:
> This is great. Can we also put this info in the C3 README so it shows up in NPM?

yes for sure 🙂, although in order to avoid duplication (and having to remember to change this in two different places when thing in C3 change), would just adding a link in the readme to here would suffice?

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
 If you choose to deploy, you will be asked to authenticate (if not logged in already), and your project will be deployed immediately. C3 will display your project's URL and some helpful links.
+
+## CLI Arguments
+
+C3 collects any required input through a series of interactive prompts. You may also specify your choices via command line arguments, which will skip these prompts. To use C3 in a non-interactive context such as CI, specify all required arguments via the command line.
+
+This is the full format of a C3 invocation alongside the possible CLI arguments:
+
+```sh
+$ npm create cloudflare@latest [<DIRECTORY>] [OPTIONS] [-- <NESTED ARGS..>]
```
deadlypants1973:
```suggestion
$ npm create cloudflare@latest [--]  [<DIRECTORY>] [OPTIONS] [-- <NESTED ARGS...>]
```

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+```
+
+{{<definitions>}}
+
+- `DIRECTORY` {{<type>}}string{{</type>}}
+  - The directory where the application should be created. The name of the application is taken from the directory name.
+- `NESTED ARGS..` {{<type>}}string[]{{</type>}}
+  - CLI arguments to pass to eventual third party CLIs C3 might invoke (in the case of full-stack applications).
+- `--type` {{<type>}}string{{</type>}}
```
deadlypants1973:
```suggestion
- `DIRECTORY` {{<type>}}string{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
  - The directory where the application should be created. The name of the application is taken from the directory name.
- `NESTED ARGS..` {{<type>}}string[]{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
  - CLI arguments to pass to eventual third party CLIs C3 might invoke (in the case of full-stack applications).
- `--type` {{<type>}}string{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
```

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+  - The type of application that should be created.
+
+  - The possible values for this option are:
+
+    - `webFramework`: A website or web application.
+    - `hello-world`: A basic "Hello World" Cloudflare Worker.
+    - `common`: A Cloudflare Worker which implements a common example of routing/proxying functionalities.
+    - `scheduled`: A scheduled Cloudflare Worker (triggered via Cron Triggers).
+    - `queues`: A Cloudflare Worker which is both a consumer and produced of Queues.
```
deadlypants1973:
```suggestion
    - `queues`: A Cloudflare Worker which is both a consumer and producer of [Queues]()/queues/).
```

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+  - CLI arguments to pass to eventual third party CLIs C3 might invoke (in the case of full-stack applications).
+- `--type` {{<type>}}string{{</type>}}
+  - The type of application that should be created.
+
+  - The possible values for this option are:
+
+    - `webFramework`: A website or web application.
+    - `hello-world`: A basic "Hello World" Cloudflare Worker.
+    - `common`: A Cloudflare Worker which implements a common example of routing/proxying functionalities.
+    - `scheduled`: A scheduled Cloudflare Worker (triggered via Cron Triggers).
```
deadlypants1973:
```suggestion
    - `scheduled`: A scheduled Cloudflare Worker (triggered via [Cron Triggers](/workers/configuration/cron-triggers/)).
```

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+    - `queues`: A Cloudflare Worker which is both a consumer and produced of Queues.
+    - `chatgptPlugin`: A ChatGPT plugin.
+    - `openapi`: A Worker implementing an OpenAPI REST endpoint.
+
+- `--framework` {{<type>}}string{{</type>}}
+  - The type of framework to use to create a web application (when using this option `--type` is ignored).
```
deadlypants1973:
```suggestion
  - The type of framework to use to create a web application (when using this option, `--type` is ignored).
```

content/pages/get-started/c3.md
dario-piotrowicz: should the `--` be `[--]`? (since you can not use it if you don't provide any argument?)

are the spaces between the `.`s intentional?

content/pages/get-started/c3.md
dario-piotrowicz: if we add this to these arguments then I guess we'd want to add it to all the way to line 130 right?

wouldn't that be a bit too verbose/repetitive? 🤔

content/pages/get-started/c3.md
dario-piotrowicz: I think there's an extra `)` there I'll fix it and apply the suggestion 🙂👍

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+
+- `--deploy` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Deploy your application after it has been created.
+
+- `--ts`{{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Use TypeScript in your application.
+
+- `--git` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
```
deadlypants1973:
```suggestion
- `--deploy` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}} {{<prop-meta>}}optional{{</prop-meta>}}

  - Deploy your application after it has been created.

- `--ts`{{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}} {{<prop-meta>}}optional{{</prop-meta>}}

  - Use TypeScript in your application.

- `--git` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}} {{<prop-meta>}}optional{{</prop-meta>}}
```

content/pages/get-started/c3.md
```diff
@@ -50,3 +50,91 @@ Once your project has been configured, you will be asked if you would like to de
...
+
+- `--auto-update` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}}
+
+  - Automatically uses the latest version of C3.
+
+- `-v`, `--version` {{<type>}}boolean{{</type>}}
+
+  - Show version number.
+
+- `-h`, `--help` {{<type>}}boolean{{</type>}}
```
deadlypants1973:
```suggestion
- `--open` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}} {{<prop-meta>}}optional{{</prop-meta>}}

  - Open with your browser the deployed application (this option is ignored if the application is not deployed).

- `-y`, `--accept-defaults` {{<type>}}boolean{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}

  - Use all the default C3 options each can also be overridden by specifying it.

- `--auto-update` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true){{</prop-meta>}} {{<prop-meta>}}optional{{</prop-meta>}}

  - Automatically uses the latest version of C3.

- `-v`, `--version` {{<type>}}boolean{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}

  - Show version number.

- `-h`, `--help` {{<type>}}boolean{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
```

content/pages/get-started/c3.md
deadlypants1973: I moved the `--` into the `[<DIRECTORY>]` because I think it makes more sense than saying `[--]` but I am happy to defer to you if it's better to have them separate. I don't know if running `npm create cloudflare@latest -- --framework nuxt` is something users are doing? But I think that would be the only group that `[--]` would benefit. What do you think?

dario-piotrowicz: I see.... I am not the biggest fan of having the `--` in the `[<DIRECTORY>]` as it does make it look like it's specific for the directory and that you can specify the other options even without it (which is not the case).

`[--]` does only apply if you specify the directory, I honestly am not sure how common not including the directory is, but for example one clear case which I imagine could be common is when people want to see the help message `-h` or version `-v`, in those case indeed they would not really include the directory (but would definitely need to include `--`).

So I would definitely just keep it as its own `[--]` 🤔

deadlypants1973 (approved):

PR #11406, opened 2023-10-23
updated-optional-step.md

muhammedriaz:
_Optional step shifted to the bottom_


kodster28:
Howdy @muhammedriaz, thanks for the suggestion but we need to keep as is since the flow needs to happen in that order (and is why we marked it as optional).

PR #11445, opened 2023-10-24
[Support] Update Troubleshooting Cloudflare 5XX errors.md

ngayerie:
SPM-1588


zeinjaber (changes requested):
1s isn't commonly used. Please change it to 1 second instead throughout the pragraph.

Change:
>> This is expect, it's due to the current work on implementing [Pingora, our new proxy](https://blog.cloudflare.com/how-we-built-pingora-the-proxy-that-connects-cloudflare-to-the-internet/).

To:
This is expected due to the current work...

>>As a workaround you can simply set the timeout to 1s more (121s instead of 120s for example).

Some suggestions:
As a workaround, you can adjust the timeout by adding 1 second (e.g., set it to 121 seconds instead of 120 seconds).

zeinjaber (approved):

PR #11480, opened 2023-10-26
[SSL] More descriptive about LE rate limiting

AlphaK5:
LE Rate limiting issue


RebeccaTamachiro:
Hey 👋 Thanks for raising this. We currently link to LE docs from https://developers.cloudflare.com/ssl/reference/certificate-authorities/#limitations
I understand how it could be useful to also have it under the troubleshooting page, but, until we're displaying this error on the dash/API, that could also be confusing. Will keep it tracked under PCX-9202 to update later though.

PR #11483, opened 2023-10-26
[DNS] Add note to limitations of random prefix attacks mitigation

AlphaK5:
helful information to understand why some little attacks are not mitigated, even with random prefix = true


RebeccaTamachiro:
Hey! I couldn't understand the connections here. Would you mind creating a Jira ticket with more information and context so that I can check your suggestion with the PM and better elaborate an addition to the docs? Thanks!

PR #11519, opened 2023-10-27
Pub/Sub: fix getting started guide

elithrar:
Fixes the Getting Started guide to make sure the `clientId` is set.


rts-rob (approved):
Thanks for running this down and putting up the fix. Works for me as updated.

PR #11521, opened 2023-10-27
Modify the default protocol in tunnel-run-parameters.md

razaxq:
# protocol
Default
auto -> quic


ranbel:
Hey @joliveirinha, did we change the default tunnel protocol to `quic`?

DevinCarr:
`auto` is still the default protocol for cloudflared. It defaults to QUIC under the hood, but there is no reason to adjust this as older clients will still be using `auto`.

PR #11529, opened 2023-10-27
Clarify restrictions for index names

zegevlier:
Hey!
It was not clear to someone in the Discord that Vectorize index names were not allowed to contain uppercase letters. This PR clarifies that the names should consist of lowercase or numeric characters. I also added that they must start with a letter, since starting an index name with a number also is not allowed.

The getting started guides seemed like the only appropriate place to put this.


elithrar (approved):

elithrar:
Thanks @zegevlier

zegevlier:
The spell check fails at a part that was not modified in this PR. I'm going to leave it as-is, as fixing that seems more appropriate for a separate PR.

Maddy-Cloudflare (approved):

PR #11544, opened 2023-10-29
Changelog entry for improved console output in local dev

irvinebroque:
This was called out in https://developers.cloudflare.com/workers/platform/changelog/wrangler/#3.15.0 but got a bit buried, and is worth having its own changelog entry.

Thank you again @mrbbot for making this happen!


kodster28:
Going to close this out as it's pretty old + the main benefit of a changelog seems to be highlighting recent changes (and this is 6+ months old at this point).

Feel free to re-open if I'm closing in error!

PR #11594, opened 2023-11-01
[R2] Using Sippy with bucket jurisdictions

### LINKED ISSUE #11595: R2 buckets in EU needs additional header flag to enable sippy (and other API calls I guess)

R2

- https://developers.cloudflare.com/r2/data-migration/sippy/
- https://blog.cloudflare.com/sippy-incremental-migration-s3-r2/

This API call from the example in the docs above needs an additional `cf-r2-jurisdiction: eu` header to work, otherwise gives error `10006` with `The specified bucket does not exist.`. Should be noted in the docs since it's a headache to be stuck here when all creds are actually configured properly.

### Additional information

Thanks to user @Kian on discord #r2 for pointing this out.

KianNH:
Closes https://github.com/cloudflare/cloudflare-docs/issues/11595

cc @sdnts


avocade:
Via https://github.com/cloudflare/cloudflare-docs/issues/11595

dcpena (approved):

PR #11597, opened 2023-11-01
[Chore] Migrate littlefoot out of headers

kodster28:
Don't need these resources loading on pages with the homepage layout, so adding the `preload` statements inline.

Validation:
- Littlefoot does not load on the homepage --> developer tools & `Network` tab shouldn't show `littlefoot` in any of the loaded resources
- On docs pages, `littlefoot` should load.
- On docs pages with footnotes, those should render --> https://remove-littlefoot-preheaders.cloudflare-docs-7ou.pages.dev/dns/zone-setups/partial-setup/dns-resolution/

### DIFF
--- a/static/home.css
+++ b/static/home.css
@@ -3437,6 +3437,17 @@ blockquote .DocsMarkdown--header-anchor-positioner {
     margin: 0 48px 0 16px;
 }
 
+.DocSearch-Help a {
+    text-decoration: underline;
+    color: var(--docsearch-highlight-color);
+    font-weight: 600;
+}
+
+.DocSearch-HitsFooter a {
+    color: var(--docsearch-highlight-color);
+    font-weight: 600;
+}
+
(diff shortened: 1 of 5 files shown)


static/home.css
```diff
@@ -3437,6 +3437,17 @@ blockquote .DocsMarkdown--header-anchor-positioner {
     margin: 0 48px 0 16px;
 }
 
+.DocSearch-Help a {
```
kodster28, line 3440: Move to near the other Algolia styles.

static/style.css
```diff
@@ -4444,6 +4444,17 @@ blockquote .DocsMarkdown--header-anchor-positioner {
   width: unset !important;
 }
 
+.DocSearch-Help a {
```
kodster28, line 4447: Move to near the other Algolia styles.

patriciasantaana (approved):

PR #11616, opened 2023-11-02
Redirect cleanup

kodster28:
This PR does the following:

1. Prunes static redirects:
     1. Takes our `_redirects` file from [6 months ago](https://github.com/cloudflare/cloudflare-docs/blob/1f39e97c402bbe3883f960319d936ff5f7748690/content/_redirects) to avoid removing recently added redirects that won't show traffic yet.
     2. Extracts the target.
     3. Compares the target against our Logpush job that tracks `301` error codes (at a sampling rate of 2%).
     4. Removes any redirect that doesn't have more than 2 hits in the past ~5 months (low traffic, therefore low value).
2. Prunes dynamic redirects:
     1. Looks at our current redirects file.
     2. Using a `contains` evaluation, see the amount of traffic going to different dynamic paths.
     3. If traffic is super minimal, either a) remove the dynamic redirect or b) replace with a static redirect (if all the traffic is to 1 or 2 paths).

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -204,7 +204,6 @@
 /d1/platform/wrangler-commands/ /workers/wrangler/commands/#d1 301
 
 # data loss prevention (dlp)
-/cloudflare-one/policies/data-loss-prevention/predefined-profiles/ /cloudflare-one/policies/data-loss-prevention/dlp-profiles/predefined-profiles/ 301
 /cloudflare-one/policies/data-loss-prevention/integration-profiles/ /cloudflare-one/policies/data-loss-prevention/dlp-profiles/integration-profiles/ 301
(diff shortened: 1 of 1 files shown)


ranbel (approved):

pedrosousa (approved):

Skye-31:
Would you want to include the Workers & Pages functions that could be redirect rules in this? E.g. [this](https://github.com/cloudflare/cloudflare-docs/blob/production/functions/pages/platform/build-configuration.json.ts#L3) and the other files in that directory & the workers/platform directory could just be something like `/pages/platform/build-configuration.json /pages/platform/build-configuration/index.json 200`

kodster28:
> this

I wasn't familiar with that, tbh.

My thought is that -- in the scope of this PR -- that wouldn't be necessary to include. We can evaluate migrating those into the `_redirects` file in the future (which IMO would be preferable so we can see all the behavior w/in the same file).

PR #11617, opened 2023-11-02
Update protect-against-timing-attacks.md

dfodeker:
Description:
The current documentation for the Crypto API in Cloudflare Workers suggests using crypto.timingSafeEqual(a, b) for performing a timing-safe comparison of two values. However, this method does not exist on the Crypto object. Instead, it is available through the SubtleCrypto interface.

This pull request updates the documentation to reflect the correct usage of timingSafeEqual. The proposed change includes an example  to guide developers on how to use this method correctly in the Cloudflare Workers environment.

Changes Made:

Updated the method call from crypto.timingSafeEqual(a, b) to crypto.subtle.timingSafeEqual(a, b)

### DIFF
--- a/content/workers/examples/protect-against-timing-attacks.md
+++ b/content/workers/examples/protect-against-timing-attacks.md
@@ -18,10 +18,17 @@ To use this function, create a new [`TextEncoder`](/workers/runtime-apis/encodin
 ```ts
 const encoder = new TextEncoder();
 
-const a = new TextEncoder().encode("foo")
-const b = new TextEncoder().encode("bar")
+const username = "foo";
+const password = "bar";
 
-if (string1.byteLength !== string2.byteLength) {
+if (username.length !== password.length) {
+  // Minimise the possibility of a timing attack via how long encoding takes on the strings
+}
+
+const a = encoder.encode(username)
+const b = encoder.encode(password)
+
+if (a.byteLength !== b.byteLength) {
   // Strings must be the same length in order to compare
   // with crypto.timingSafeEqual
   return false
--- a/content/workers/examples/protect-against-timing-attacks.md
+++ b/content/workers/examples/protect-against-timing-attacks.md
@@ -36,7 +36,7 @@ if (string1.byteLength !== string2.byteLength) {
 const a = encoder.encode(string1);
 const b = encoder.encode(string2);
 
-let equal = crypto.timingSafeEqual(a, b)
+let equal = crypto.subtle.timingSafeEqual(a,b)
 
 if (equal) {
   // The values are equal
@@ -52,8 +52,8 @@ if (equal) {
 ```ts
 const encoder = new TextEncoder();
 
-const string1 = new TextEncoder().encode("foo")
-const string2 = new TextEncoder().encode("bar")
+const a = new TextEncoder().encode("foo")
+const b = new TextEncoder().encode("bar")
 
 if (string1.byteLength !== string2.byteLength) {
   // Strings must be the same length in order to compare
@@ -66,12 +66,13 @@ if (string1.byteLength !== string2.byteLength) {
 
 // You can replace it with `crypto.timingSafeEqual` by encoding the values
 // you need to compare
-const a = encoder.encode(string1);
-const b = encoder.encode(string2);
+const a = encoder.encode("foo");
+const b = encoder.encode("bar");
 
-let equal = crypto.timingSafeEqual(a, b)
+//let equal = crypto.timingSafeEqual(a, b)
+let isEqual = crypto.subtle.timingSafeEqual(a,b)
 
-if (equal) {
+if (isEqual) {
   // The values are equal
 } else {
   // The values are not equal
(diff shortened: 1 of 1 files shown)


content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -66,12 +66,13 @@ if (string1.byteLength !== string2.byteLength) {
 
 // You can replace it with `crypto.timingSafeEqual` by encoding the values
 // you need to compare
-const a = encoder.encode(string1);
-const b = encoder.encode(string2);
+const a = encoder.encode("foo");
+const b = encoder.encode("bar");
 
-let equal = crypto.timingSafeEqual(a, b)
+//let equal = crypto.timingSafeEqual(a, b)
```
KianNH:
```suggestion
```

KianNH (commented):
Thanks for the PR!

There's a JS snippet just above the TS snippet that you made the other changes to (like not re-encoding `string1` and `string2`) - would you be able to make the same changes to the JS snippet?

Admittedly, there is no need for a JS and a TS snippet as the code is identical so you could remove all but one snippet (as well as the `tab` and `tabs` shortcode but it's no problem if you don't want to do that, I can do that in a follow-up PR!

content/workers/examples/protect-against-timing-attacks.md
dfodeker: I think that makes a lot of sense! 😊 I'll make the changes to the JS snippet.

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -64,19 +30,17 @@ if (string1.byteLength !== string2.byteLength) {
 // The below code is vulnerable to timing attacks
 // if (string1 === string2) { ... }
 
-// You can replace it with `crypto.timingSafeEqual` by encoding the values
+// You can replace it with `crypto.subtle.timingSafeEqual` by encoding the values
 // you need to compare
-const a = encoder.encode(string1);
-const b = encoder.encode(string2);
+const a = encoder.encode("foo");
+const b = encoder.encode("bar");
```
KianNH:
```suggestion
```

These can be removed as we already encoded the strings further up.

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -11,49 +11,15 @@ layout: example
...
 
 ```ts
 const encoder = new TextEncoder();
 
-const string1 = new TextEncoder().encode("foo")
-const string2 = new TextEncoder().encode("bar")
+const a = new TextEncoder().encode("foo")
+const b = new TextEncoder().encode("bar")
 
 if (string1.byteLength !== string2.byteLength) {
```
KianNH:
```suggestion
const username = "foo";
const password = "bar";

if (username.length !== password.length) {
  // Minimise the possibility of a timing attack via how long encoding takes on the strings
}

const a = encoder.encode(username)
const b = encoder.encode(password)

if (a.byteLength !== b.byteLength) {
```

We can use the TextEncoder instance we created earlier for this, and amend the byteLength check to use the new variable names.

This adds a string length check, to minimise timing attacks from how long encoding takes, but we keep the byteLength check due to unicode.

```js
> "a".length
1
> "ä".length
1
> new TextEncoder().encode("a").length
1
> new TextEncoder().encode("ä").length
2
```

KianNH (commented):
Thanks for making the changes! There's just a few other improvements that we can make and then it's good to go.

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -11,49 +11,15 @@ layout: example
...
+To use this function, create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) and encode the string values to instances of `ArrayBuffer` using [`encoder.encode`](/workers/runtime-apis/encoding/#methods). This is needed because `crypto.timingSafeEqual` compares `ArrayBuffer` instances, not strings. With the encoded values, replace the standard JavaScript equality check (`===`) with `crypto.timingSafeEqual`. Note that the strings must be the same length in order to compare to `timingSafeEqual`. The below code shows how to implement string equality checks with `crypto.timingSafeEqual`. Note that the example shown would apply to Typescript and Javascript:
```
KianNH:
```suggestion
To use this function, create a new [`TextEncoder`](/workers/runtime-apis/encoding/#textencoder) and encode the string values to instances of `ArrayBuffer` using [`encoder.encode`](/workers/runtime-apis/encoding/#methods). This is needed because `crypto.subtle.timingSafeEqual` compares `ArrayBuffer` instances, not strings. With the encoded values, replace the standard JavaScript equality check (`===`) with `crypto.subtle.timingSafeEqual`. Note that the strings must be the same length in order to compare to `timingSafeEqual`. The below code shows how to implement string equality checks with `crypto.subtle.timingSafeEqual`. Note that the example shown would apply to TypeScript and JavaScript:
```

dfodeker (commented):
updated comment with
// with crypto.subtle.timingSafeEqual

KianNH (commented):
Looks good to me! Thanks for the contribution.

deadlypants1973 (approved):
Thank you @KianNH for reviewing!

deadlypants1973:
And thank you so much @dfodeker for this submission!

deadlypants1973:
@dfodeker I can merge this tomorrow if there is no further work needed?

dfodeker:
Thank you @deadlypants1973 , That sounds great !  I don't have any additional changes in mind

PR #11653, opened 2023-11-06
Remove redundant return await

isaacl:
Return await does nothing here

### DIFF
--- a/content/workers/examples/debugging-logs.md
+++ b/content/workers/examples/debugging-logs.md
@@ -20,7 +20,7 @@ export default {
     const LOG_URL = "https://log-service.example.com/";
 
     async function postLog(data) {
-      return await fetch(LOG_URL, {
+      return fetch(LOG_URL, {
         method: "POST",
         body: data,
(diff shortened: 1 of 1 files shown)


isaacl:
To be clear -- based on my understanding of typescript, the 'return await' paradigm is only used for catching exceptions.  Otherwise you're simply resolving a Promise and packing into a new Promise.  But I might be wrong...

GregBrimble:
Thanks for your PR! Unfortunately though, I don't think we should merge this. We've marked the `postLog` function as asynchronous, which I think helps demonstrate how `waitUntil` works (it resolves promises), and `await fetch` also signals to readers that they're flexible in what they run in this function (e.g. they could make multiple fetches to multiple logging services), so I think we're best just keeping it as is.

isaacl:
Sure.  I understand and thanks for the quick response!

Also, after sending in this PR I noticed `return await` in many other documentation examples as well. So it'd be inconsistent to only remove here.

PR #11677, opened 2023-11-07
Changed format to earliest supported + format

eprosimon:
Changed the format for supported version listings to reduce maintenance burden. 

Reviewing the EOL statuses of the macOS versions it appears that Catalina and Big Sur are EOL and the oldest supported version is macOS Monterey 12.6.2

https://computing.cs.cmu.edu/desktop/os-lifecycle

We may need to bump the minimum supported version to the oldest version supported by apple to ensure that we are not suggesting the use of an outdated OS.


ranbel:
Should we also update the Windows and Linux sections?

eprosimon:
That would be a good Idea as this will require confirmation from product, I have raised a PCX-8920 for discussion.

ranbel (approved):

PR #11774, opened 2023-11-13
Add precision on onion service x509 termination

thibmeu:
This commit explains the expected behaviour users should see when connecting to a website with onion service enabled.


Lekensteyn (approved):
Not just host header, but also SNI, but this looks good to me!

thibmeu:
> Not just host header, but also SNI, but this looks good to me!

I've update the wording to have `host header and SNI`

ngayerie (approved):

zeinjaber (approved):

XoL1507:
dont worry i read for 3 dyas without sleep lol. i understand. but i hope we can help each other to have smoth flow of project.

PR #11816, opened 2023-11-14
Pages: Fix typo in angular framework documentation

TheConner:
Found a small typo whilst reading:

"se the" => "See the"

Edit: upon revisiting my browser tab, I think the original intended meaning was "Use" not "See". Added a commit to correct my correction


WalshyDev:
> Edit: upon revisiting my browser tab, I think the original intended meaning was "Use" not "See". Added a commit to correct my correction

I think you're right here

Thank you for catching :)

WalshyDev (approved):

deadlypants1973 (approved):

PR #11842, opened 2023-11-16
[ZT] Add details for remote tunnel debugging

eprosimon:
Add clearer guidance to customers regarding what's needed to configure logging on remote tunnels

Support uses this link as part of their macro to gather information related to tunnels. 

I was not able to find any clear guide to send to customers specifically for updating the logging which is a very common thing support asks for.

### DIFF
--- a/content/cloudflare-one/faq/cloudflare-tunnels-faq.md
+++ b/content/cloudflare-one/faq/cloudflare-tunnels-faq.md
@@ -186,141 +186,29 @@ Before contacting the Cloudflare support team:
 
 - Take note of any options you specified, either on the command line or in your configuration file, when starting your tunnel.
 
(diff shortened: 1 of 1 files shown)


ranbel (approved):

eprosimon:
The purpose of my previous additions and their verbose nature was due to the confusion that arises when we point to
https://developers.cloudflare.com//cloudflare-one/connections/connect-networks/configure-tunnels/remote-management/

As this is not a guide that relates specifically to troubleshooting but is more generic. 

I believe it would still be helpful to update the example from QUIC protocol to the log level and log file if we are not happy with having content for this specific case as it would be a sufficient example but a more useful guide for users when they get instructions from support

ranbel:
I think changing the default example would be a good solution.

eprosimon:
Thanks for making that change :)

PR #11881, opened 2023-11-20
[D1] Revisit Overview page

Maddy-Cloudflare:


rozenmd (commented):
tbh I don't know if the `<feature/>` tags are the best for features here:
image 
(due to the "Use ..." call to action), though I don't know what an alternative would be.

Maddy-Cloudflare:
> tbh I don't know if the `<feature/>` tags are the best for features here: image (due to the "Use ..." call to action), though I don't know what an alternative would be.

What's your concern exactly? I can modify the text in the box.

rozenmd:
> What's your concern exactly? I can modify the text in the box

I would change "D1 database" as the first heading to "Get started quickly" or similar, as "D1 database" isn't really a feature here. I would change the CTA to "Get started now", linking to the same get started guide.

Time Travel works great as a feature, and is a unique selling point, so that works, but the CTA feels weird, since you're not going to use the feature unless something goes wrong. I'd change it to "Learn more" or similar.

rozenmd (approved):

content/d1/_index.md
```diff
@@ -17,35 +17,55 @@ Create new serverless SQL databases to query from your Workers and Pages project
...
+{{<plan type="workers_all">}}
+
+D1 is Cloudflare’s native serverless database. D1 allows you to build applications that handle large amounts of users at no extra cost. With D1, you can restore your database to any minute within the last 30 days.
+
+Create your first D1 database by [following the Get started guide](/d1/get-started/), learn how to [import data into a database](/d1/learning/importing-data/), and how to [query your database](/d1/platform/client-api/) directly from [Workers](/workers/) or [Pages](/pages/platform/functions/bindings/#d1-databases).
 
 {{<Aside type="note" header="D1 is in public beta">}}
 
-D1 is in public beta. While we expect breaking changes and issues to be minimal, they may still occur. We generally do not recommend running large production workloads on beta products.
+D1 is in public beta. While the D1 team expect breaking changes and issues to be minimal, they may still occur. The D1 team generally do not recommend running large production workloads on beta products.
```
jason-cf:
```suggestion
D1 is in public beta. While the D1 team expects breaking changes and issues to be minimal, they may still occur. The D1 team generally does not recommend running large production workloads on beta products.
```

jason-cf (approved):

PR #11883, opened 2023-11-20
[Hugo] Create a calculator component to promote best practices

kodster28:


layouts/shortcodes/calculation.html
```diff
@@ -0,0 +1,110 @@
...
+      inputObjects.push(inputObject);
+    });
+
+    const results = document.querySelector("form.calculations");
+    const resultElements = results.querySelectorAll("input");
+
+    resultElements.forEach(function (input) {
+      let formula = input.dataset.formula;
+      let evaluatableFormula = replacePlaceholders(inputObjects, formula);
+      input.value = math.evaluate(evaluatableFormula);
```
kodster28, line 79: We don't want generic `eval()` here b/c that can [execute any input](https://www.codiga.io/blog/javascript-eval-best-practices/), so opting for [`math.evaluate`](https://mathjs.org/docs/expressions/security.html) with the values swapped out from the preceding regex.

That's also why the codeowners file has these paths locked down a bit more strictly.

.github/CODEOWNERS
```diff
@@ -11,6 +11,8 @@
 *.js @cloudflare/developer-advocacy @kristianfreeman @kodster28 @pedrosousa @haleycode
 *.ts @cloudflare/developer-advocacy @kristianfreeman @kodster28 @pedrosousa @haleycode
 /assets/json-collector.ts @cloudflare/pcx-technical-writing
+/data/calculations/ @kristianfreeman @kodster28 @pedrosousa @haleycode
```
kodster28:
```suggestion
/data/calculations/ @kristianfreeman @kodster28 @pedrosousa
```

.github/CODEOWNERS
```diff
@@ -11,6 +11,8 @@
 *.js @cloudflare/developer-advocacy @kristianfreeman @kodster28 @pedrosousa @haleycode
 *.ts @cloudflare/developer-advocacy @kristianfreeman @kodster28 @pedrosousa @haleycode
 /assets/json-collector.ts @cloudflare/pcx-technical-writing
+/data/calculations/ @kristianfreeman @kodster28 @pedrosousa
+/layouts/shortcodes/calculation.html @kristianfreeman @kodster28 @pedrosousa @haleycode
```
kodster28:
```suggestion
/layouts/shortcodes/calculation.html @kristianfreeman @kodster28 @pedrosousa
```

layouts/shortcodes/calculation.html
```diff
@@ -0,0 +1,110 @@
...
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjs/12.1.0/math.min.js"></script>
+<script>
+  const inputs = document.querySelector("form.inputs");
+  const inputElements = inputs.querySelectorAll("input");
+
+  function replacePlaceholders(inputObj, inputString) {
+    inputObj.forEach(function (inputObj) {
+      inputString = inputString.replace(
+        new RegExp(inputObj.id, "g"),
```
pedrosousa: I'd use something a bit more constrained as the regex, so that we don't inadvertently capture parts of other variable names. 
For example, given this expression:
`packet_counter + udp_packet_counter * 10`

The current replacement strategy would create an invalid expression if we replaced `packet_counter` with its value (e.g., `20`) first.
```
"packet_counter + udp_packet_counter * 10"
==> "20 + udp_20 * 10"
==> <Error ('udp_20' undefined)>
```

I would use something like the following here:
```js
new RegExp("(?<=^|[\s/+*(-])(" + inputObj.id + ")(?=[\s+/*)-]|$)", "g"),
```
(Source: https://stackoverflow.com/questions/59745015/regex-to-capture-variable-names-in-an-expression )

Had to introduce a change in the regex from that source: a lookbehind at the beginning (`?<=`), so it doesn't consume the character before the variable name when replacing. Lookbehind now seems to be [supported in recent browsers](https://caniuse.com/js-regexp-lookbehind). Source for this: https://stackoverflow.com/a/57450029

A few test cases here:
https://regex101.com/r/yddBCZ/2

layouts/shortcodes/calculation.html
```diff
@@ -0,0 +1,110 @@
...
+      const inputObject = {
+        id: input.id,
+        value: input.value,
+      };
+
+      // Push the object into the array
+      inputObjects.push(inputObject);
+    });
+
+    const results = document.querySelector("form.calculations");
```
pedrosousa: Nit: Using a general "form" element selector here will prevent us from having more than one calculation on the same page. To avoid this, we could generate and use an ID for the form.

It wouldn't be enough, though. We would also have a clash of function names (something like "calculateData() already defined") — unless Hugo already handles this somehow.

I'd say this is a concern for later!

pedrosousa (approved):

layouts/shortcodes/calculation.html
kodster28: Yeah, more than one per page will def break the implementation... and I figure we just have that as a general guideline and then re-adjust if we need to in the future.

PR #11884, opened 2023-11-20
Update purge-by-single-file.md

luishcardoso:
From a recent live chat I worked on, I understand we can benefit from adding to our docs that "Purge by URL is case sensitive". (PCX-9054)


angelampcosta:
Suggestion is addressed in https://github.com/cloudflare/cloudflare-docs/pull/11943

PR #11892, opened 2023-11-21
Update managed-networks.md

ShadowThePatriot:
This irritated me in the cli when trying to decipher the new correct command that includes no deprecated message, so here is the proposed change. "warp-cli get-alternate-network" is being deprecated, so here is the correct future command that could be displayed on the docs page.


ranbel (approved):
Thanks for updating the docs!

PR #11907, opened 2023-11-22
[Wrangler] Update "check wrangler version"

rozenmd:
We're no longer telling folks to globally install wrangler, but then immediately ask them to run a command against a globally installed wrangler 😕

### DIFF
--- a/content/workers/wrangler/install-and-update.md
+++ b/content/workers/wrangler/install-and-update.md
@@ -25,9 +25,9 @@ To install Wrangler within your Worker project, run:
 To check your Wrangler version, run:
 
 ```sh
-$ wrangler --version
+$ npx wrangler --version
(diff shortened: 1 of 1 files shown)


mrbbot (approved):

KianNH:
Is the aside below still accurate?

> Running `npx wrangler` will use the latest version of Wrangler except when there is already a Wrangler version installed in the current folder. In this case, it will use the locally installed version instead.

I thought that was removed in https://github.com/cloudflare/workers-sdk/pull/3192

Edit: I might be confusing this for running just `wrangler`

rozenmd:
Yeah that was just wrangler, `npx` runs local first then checks npm if it's not there

deadlypants1973 (approved):

PR #11908, opened 2023-11-22
[D1] Make logging in the first step

rozenmd:
Step 2 (create a D1 DB) requires being logged in anyway, may as well make logging in "Step 0"


Maddy-Cloudflare:
> Step 2 (create a D1 DB) requires being logged in anyway, may as well make logging in "Step 0"

We could just have it as step 1, we don't really have "step 0" in our docs. What do you think? :)

rozenmd:
Ah okay, I've updated the PR @Maddy-Cloudflare

Maddy-Cloudflare (approved):

PR #11930, opened 2023-11-22
[KV] Fix typo in cacheTtl parameter

brunohpaiva:
Fix `cacheTtl` parameter typo in KV docs.
Reference: https://github.com/cloudflare/workerd/blob/754bc0267643c8a00f7a29e5092cd22de1b82140/src/workerd/api/kv.h#L32


Maddy-Cloudflare (changes requested):
TTL is a concept that is shared across other Cloudflare products. We need to decide how we want to write it so that it's uniform.

elithrar:
@Maddy-Cloudflare `cacheTtl` is how KV & Workers' Cache API reference cache TTL settings.

The PR here is correct in fixing it to be `cacheTtl`. A user typing `cacheTTL` will see errors.

elithrar (approved):

Maddy-Cloudflare (approved):

PR #11933, opened 2023-11-23
[docs site] Fail fast when transforming files from Vite & disable gzip reporting

penalosa:
_Hopefully_ this provides a slight CI speedup


penalosa:
Looks like it cuts about 20 seconds ish off `make build`? Not huge, but probably worth it

kodster28 (approved):
20 seconds isn't anything to sneeze at! TYSM for helping with this.

We also did get a boost from removing the localized content recently (around ~1 min per build).

kodster28:
Feel free to merge in whenever you're ready, @penalosa.

PR #11946, opened 2023-11-23
Update create-remote-tunnel.md

wolruf:
Cloudflare employee here.

Network > Access to access Tunnels in the Dashboard


ranbel:
Screenshot 2023-12-04 at 11 37 14 AM
We created a new Networks section a few months ago, so the documentation should be correct. What do you see in your dashboard?

PR #11960, opened 2023-11-24
New text generation and speech recognition docs

celso:


content/workers-ai/models/text-generation.md
```diff
@@ -142,20 +144,122 @@ $ curl https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/ai/run/@cf/met
...
+  prompt: "tell me a joke about cloudflare"
+};
 ```
 
-### Handling streaming responses in the client
+Here's an input example of a [Mistral](https://docs.mistral.ai/llm/mistral-instruct-v0.1#chat-template) chat template prompt:
+
+```javascript
+{
+  prompt: "<s>[INST]comedian[/INST]</s>\n[INST]tell me a joke about cloudflare[/INST]"
```
pdwittig, line 169: We should consider adding a default system prompt (or a template) for users using the prompt arg, and urge users to use the messages arg if they want to use a system prompt. That would help making sure it works out of the box. I am not sure what benefit the developer gets from having to do this themselves? Not blocking, but something to consider.

pdwittig (approved):

PR #11971, opened 2023-11-26
fix incorrect bindings example in Next.js pages guide

dario-piotrowicz:


WalshyDev (approved):
we should also add js but lgtm

dario-piotrowicz:
> we should also add js but lgtm

Yes I agree, but these bindings sections usually just present the typescript version (for example see [qwik](https://developers.cloudflare.com/pages/framework-guides/deploy-a-qwik-site/#use-bindings-in-your-qwik-application) and [remix](https://developers.cloudflare.com/pages/framework-guides/deploy-a-remix-site/#create-and-add-a-binding-to-your-remix-application)).

Adding both js and ts could be quite verbose and I guess we'd need to have two different sections one for each or something like that...

I'm afraid it could increase the size of the guide and make it more overwhelming/intimidating... 🤷

We've discussed about the bindings sections with @deadlypants1973 and they should probably get overhauled at some point as now they are not too helpful/explanatory for newcomers.

So I am not really sure, I am happy to add a js section for this but my feeling is that these need to be improved anyhow at some point and for now we can at least make it correct in the most common case (typescript).

TLDR; let's see what @deadlypants1973 says, I'd be happy to add a js version of this but I am not sure how that'd look

WalshyDev:
> Adding both js and ts could be quite verbose and I guess we'd need to have two different sections one for each or something like that...

We generally just do a code switcher as there isn't much content change between the two - https://developers.cloudflare.com/pages/platform/functions/bindings/

dario-piotrowicz:
> > Adding both js and ts could be quite verbose and I guess we'd need to have two different sections one for each or something like that...
> 
> We generally just do a code switcher as there isn't much content change between the two - https://developers.cloudflare.com/pages/platform/functions/bindings/

Yes, but with in this case we need to mention the env.d.ts update for Typescript, so a code switcher wouldn't fully help, no? (unless we put all file contents in a single code snippet? 😕)

WalshyDev:
Yeah I'd just put both the env block and code block into TS switch and then just code for JS switch

Not the end of the world at all and not blocking but we should always push for JS first, not everyone uses TS (especially beginners!) -- this was very much a note not a "pls change" :)

dario-piotrowicz:
> Yeah I'd just put both the env block and code block into TS switch and then just code for JS switch

I see 🙂 

Putting both in a single block would be a bit ugly from my point of view but not the end of the world, I'm totally ok with making the change if you want, but I'd really like to see what Kate thinks as I mentioned she's has had opinions on these bindings sections 🙂 

> Not the end of the world at all and not blocking but we should always push for JS first, not everyone uses TS (especially beginners!) -- this was very much a note not a "pls change" :)

cool thanks 😁 , but either way I am totally happy with making the change here, applying the switch is going to be very straightforward anyways, it's no problem, I was just saying that I'm not sure what's best with the current structure we have 🙂

GregBrimble (approved):

content/pages/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -196,22 +196,42 @@ A [binding](/pages/platform/functions/bindings/) allows your application to inte
 
 In Next.js, add server-side code via [API Routes](https://nextjs.org/docs/pages/building-your-application/routing/api-routes), [Route Handlers](https://nextjs.org/docs/app/building-your-application/routing/router-handlers), and [getServerSideProps](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props). Then, access bindings set for your application by accessing them in your code via `process.env`.
 
-The following code block shows an example of accessing a KV namespace in Next.js.
+The following code shows an example of accessing a KV namespace in a TypeScript Next.js project.
 
+Firstly we create a new `env.d.ts` file in which we declare our binding:
```
GregBrimble:
```suggestion
First, create a new `env.d.ts` file  and declare a binding:

```

content/pages/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -196,22 +196,42 @@ A [binding](/pages/platform/functions/bindings/) allows your application to inte
...
+      // sure to include the package in your tsconfig.json's "types" field
+      MY_KV: KVNamespace;
+    }
+  }
+}
+
+export {};
+```
+
+Then such binding can be accessed directly from `process.env`:
```
GregBrimble:
```suggestion
Then, the binding can be accessed directly from `process.env`:

```

content/pages/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -196,22 +196,44 @@ A [binding](/pages/platform/functions/bindings/) allows your application to inte
...
-highlight: [4, 5]
+filename: env.d.ts
+highlight: [5,6,7]
+---
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv {
+      [key: string]: string | undefined;
+      // KVNamespace comes from @cloudflare/workers-types, make
+      // sure to include the package in your tsconfig.json's "types" field
```
dario-piotrowicz: This part [is not clear and can confuse people](https://github.com/cloudflare/next-on-pages/issues/562#issuecomment-1827460471), it needs to be improved

(by either keeping things simple and importing the workers-types in this file or by specifically explaining the tsconfig.json change needed)

dario-piotrowicz: I ended up just updating the comment for the time being, hopefully it should be clear enough

content/pages/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -196,22 +196,48 @@ A [binding](/pages/platform/functions/bindings/) allows your application to inte
 
 In Next.js, add server-side code via [API Routes](https://nextjs.org/docs/pages/building-your-application/routing/api-routes), [Route Handlers](https://nextjs.org/docs/app/building-your-application/routing/router-handlers), and [getServerSideProps](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props). Then, access bindings set for your application by accessing them in your code via `process.env`.
 
-The following code block shows an example of accessing a KV namespace in Next.js.
+The following code shows an example of accessing a KV namespace in a TypeScript Next.js project.
+
+First, create a new `env.d.ts` file  and declare a binding:
```
deadlypants1973:
```suggestion
First, create a new `env.d.ts` file and declare a binding:
```
Is there a directory they need to put this file into?

dario-piotrowicz: no there isn't, and actually the name of the file is not too important now that I think about that 😕  (but it is a common convention)

Well since this is an example I guess that should be ok? 🙂

content/pages/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -196,22 +196,48 @@ A [binding](/pages/platform/functions/bindings/) allows your application to inte
 
 In Next.js, add server-side code via [API Routes](https://nextjs.org/docs/pages/building-your-application/routing/api-routes), [Route Handlers](https://nextjs.org/docs/app/building-your-application/routing/router-handlers), and [getServerSideProps](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props). Then, access bindings set for your application by accessing them in your code via `process.env`.
 
-The following code block shows an example of accessing a KV namespace in Next.js.
+The following code shows an example of accessing a KV namespace in a TypeScript Next.js project.
+
+First, create a new `env.d.ts` file and declare a binding:
```
deadlypants1973:
```suggestion
First, create a new `env.d.ts` file and declare a [binding](/pages/platform/functions/bindings/):
```

content/pages/framework-guides/deploy-a-nextjs-site.md
```diff
@@ -196,22 +196,48 @@ A [binding](/pages/platform/functions/bindings/) allows your application to inte
...
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv {
+      [key: string]: string | undefined;
+      // The KVNamespace binding type used here comes
+      // from `@cloudflare/workers-types`, in order to
+      // use it like so make sure that you've installed
+      // the package as a dev dependency and you've added
+      // it to your `tsconfig.json` file under
+      // `compilerOptions.types`.
```
deadlypants1973:
```suggestion
      // The KV Namespace binding type used here comes
      // from `@cloudflare/workers-types`, in order to
      // use it like so, make sure that you have installed
      // the package as a dev dependency and you have added
      // it to your `tsconfig.json` file under
      // `compilerOptions.types`.
```

dario-piotrowicz:
@WalshyDev regarding adding a Javascript version, I'll do that separately across the board for all the various meta frameworks (https://github.com/cloudflare/cloudflare-docs/issues/12041) 🙂👍

PR #11979, opened 2023-11-27
Update faq.md

RyderCragie:
3rd link down on here doesn't work:

https://developers.cloudflare.com/dns/troubleshooting/faq/#why-are-dns-queries-returning-incorrect-results


kodster28:
Thanks for the catch + fix, @RyderCragie!

kodster28 (approved):

PR #11985, opened 2023-11-27
Add caveat about ETags during R2 migrations

sdnts:


kodster28:
Heyo, @sdnts might want to consider using a reusable partial if you're re-stating the exact same thing multiple times:

- Partial [definition example](https://github.com/cloudflare/cloudflare-docs/blob/production/content/r2/_partials/_keys.md)
- Partial [usage example](https://github.com/cloudflare/cloudflare-docs/blob/production/content/r2/examples/rclone.md?plain=1#L8)

sdnts:
> Heyo, @sdnts might want to consider using a reusable partial if you're re-stating the exact same thing multiple times:

Peskily enough, the text is _slightly_ different, one says "Super Slurper", and the other says "Sippy". I don't suppose we have a way to define variables in partials, do we?

kodster28:
> > Heyo, @sdnts might want to consider using a reusable partial if you're re-stating the exact same thing multiple times:
> 
> Peskily enough, the text is _slightly_ different, one says "Super Slurper", and the other says "Sippy". I don't suppose we have a way to define variables in partials, do we?

We do, actually! I'll chat you.

content/r2/data-migration/sippy.md
```diff
@@ -173,3 +173,9 @@ To create credentials with the correct permissions:
 3. Create a new user and attach the created policy to that user.
 
 You can now use both the Access Key ID and Secret Access Key when enabling Sippy.
+
+## Caveats
+
+### ETags
+
+{{<render file="_name-of-partial.md" withParameters="Sippy">}}
```
dcpena: I think `_name-of-partial.md` should be swapped out with the new partial name `_migrator-etag-caveat.md` yeah?

sdnts: Lol, yes. Done!

dcpena (approved):

PR #11988, opened 2023-11-27
Update managed-networks.md

cehrig:
Not a python expert but this example seems to deadlock if there's a single hanging TCP connection refusing to close.

Can repro with ...
```
$ openssl s_client -connect 127.0.0.1:3333
```
... and a parallel curl. 

Basically simulating two WARP clients trying to connect to the server. First client leaves connection hanging, second client can't connect and is unable to detect Managed Network.


ranbel:
@dh-cf

ranbel (approved):
LGTM. Thanks for the PR!

PR #12001, opened 2023-11-28
wildcard domains are not supported for Pages

psm384:
Just adding a cross reference from "wildcard dns records" over to the cloudflare pages known limitations (which includes that currently custome domain wildcards are not supported).

(i would suggest inserting that the current workaround is to use workers sites, but not sure that that works either)


RebeccaTamachiro (approved):

RebeccaTamachiro:
👋 @psm384 Thanks for flagging this! I've updated the DNS documentation.

PR #12058, opened 2023-12-03
Add redirect for dash logpush doc link

WalshyDev:
The docs link on the zone-level logpush Custom Fields section (https://dash.cloudflare.com/?to=/:account/:zone/analytics/logs) is going to https://developers.cloudflare.com/logs/reference/logpush-api-configuration/custom-fields (no trailing slash) which is 404ing and not being caught by the existing `/logs/reference/logpush-api-configuration/custom-fields/` redirect

This puts up a working redirect while we update the dash link

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -516,6 +516,7 @@
 /logs/logpush/s3-compatible-endpoints/ /logs/get-started/enable-destinations/s3-compatible-endpoints/ 301
 /logs/reference/logpush-api-configuration/ /logs/get-started/api-configuration/ 301
 /logs/reference/logpush-api-configuration/filters/ /logs/reference/filters/ 301
+# Non-slashed version is being used in the Cloudflare dashboard
 /logs/reference/logpush-api-configuration/custom-fields /logs/reference/custom-fields/ 301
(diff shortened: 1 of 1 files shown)


content/_redirects
```diff
@@ -516,6 +516,7 @@
 /logs/logpush/s3-compatible-endpoints/ /logs/get-started/enable-destinations/s3-compatible-endpoints/ 301
 /logs/reference/logpush-api-configuration/ /logs/get-started/api-configuration/ 301
 /logs/reference/logpush-api-configuration/filters/ /logs/reference/filters/ 301
+/logs/reference/logpush-api-configuration/custom-fields /logs/reference/custom-fields/ 301
```
pedrosousa, line 520:
```suggestion
# Non-slashed version is being used in the Cloudflare dashboard
/logs/reference/logpush-api-configuration/custom-fields /logs/reference/custom-fields/ 301
```
I'd add a comment so that it doesn't get removed inadvertently (seems like a duplicate entry).

pedrosousa (approved):

content/_redirects
WalshyDev: that's a very good idea

PR #12069, opened 2023-12-05
test

Mrkezii:


kodster28:
Howdy @Mrkezii, not entirely sure if we want to put all docs content directly accessible w/in the dash.

B/c of various concerns and potentially needing a more thorough review process.... we'd probably want this process to be more opt-in for specific pages instead of a generic opt-out.

Which'd mean more of a specifying the output specifically in the frontmatter of a page, instead of adjusting the config.toml to give them all `docs-sidebar` layout.z

(though also, realize this might just be for a PoC and not the final design. Just wanted to add our 2 cents where appropriate)

kodster28:
> Howdy @Mrkezii, not entirely sure if we want to put all docs content directly accessible w/in the dash.
> 
> B/c of various concerns and potentially needing a more thorough review process.... we'd probably want this process to be more opt-in for specific pages instead of a generic opt-out.
> 
> Which'd mean more of a specifying the output specifically in the frontmatter of a page, instead of adjusting the config.toml to give them all `docs-sidebar` layout.z
> 
> (though also, realize this might just be for a PoC and not the final design. Just wanted to add our 2 cents where appropriate)

Thought about this some more + talked with Katie... think this is more of a v2 thingy and potential req instead of v1.

So, tl;dr, ignore for now.

kodster28:
closing for now, since I think we're taking a different approach now

PR #12085, opened 2023-12-05
Update network-ports.md

AlphaK5:
DNS services


kodster28:
closing since there hasn't been any activity in a while here

PR #12086, opened 2023-12-05
[Cache] Add warning about JSD and ETags

cascadingstyletrees:
Updating Cache Article based on CUSTESC-33810

Note:
May in the future add warnings about etags in the referenced articles (JSD,E-mail obfs...)

PR #12090, opened 2023-12-05
[Bots] JSD ETags limitation

patriciasantaana:
PCX-9183


kodster28:
should eval together with #12086

kodster28 (approved):

PR #12091, opened 2023-12-05
Fix _create-tunnel.md navigation instruction

ushuz:
There's no more `Networks` sidebar in the dashboard, at least for me:

image


kodster28:
Howdy @ushuz, this is a bug with the Zero Trust dashboard and should be fixed soon.

PR #11306, opened 2023-10-16
[Workers] Update Vectorize docs to use new query format

ndisidore:
Sister to https://github.com/cloudflare/workerd/pull/1297

We're making a few changes to how Vectorize's query works in the form of a new compatibility date

PR #12098, opened 2023-12-06
[Vectorize] Add Vectorize metadata filtering docs

vy-ton:


content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+Results with metadata filtering:
+
+```json
+{"matches":[{"vectorId":"4","score":0.789848214},{"vectorId":"1","score":0.491185264}]}
+```
+
+## Limitations
+- Only newly created indexes on or after 2023-12-06 support metadata filtering.
```
vy-ton: I don't believe there is a version number to check on their index, so users have to rely on `created_on`?

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+let metadataMatches = await env.YOUR_INDEX.query(queryVector, { topK: 3, filter: { streaming_platform: "netflix" }, returnValues: true, returnMetadata: true } )
+```
+
+Results with metadata filtering:
+
+```json
+{"matches":[{"vectorId":"4","score":0.789848214},{"vectorId":"1","score":0.491185264}]}
+```
+
+## Limitations
```
vy-ton, line 168: Any other limitations to call out

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+Keys define nesting with `.` (dot)
+```json
+{ "pandas.nice": 42 } // looks for { "pandas": { "nice": 42 } }
+```
+
+## Examples
+
+### Add metadata
+
+With the following index defintion:
```
Maddy-Cloudflare:
```suggestion
With the following index definition:
```

Maddy-Cloudflare (approved):

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+Optional `filter` property on `query()` method specifies metadata filter.
+- Operators: `$eq` (equals), `$ne` (not equals)
+- `filter` must be non-empty object whose compact JSON representation must be less than 2048 bytes
+- `filter` object keys cannot be empty, contain `"`, start with `$`, or be longer than 512 characters
+
+Valid `filter` examples:
+
+Implicit `$eq` operator
+```json
+{ streaming_platform: "netflix" }
```
ndisidore: nit: probably a good idea to be consistent with quoting keys - would recommend quoting "streaming_platform" (as currently its valid js but not valid json)

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+### Query examples
+
+Use the `query()` method:
+
+```ts
+let queryVector: Array<number> = [54.8, 5.5, 3.1];
+// Best match is vector id = 5 (score closet to 1)
+let originalMatches = await env.YOUR_INDEX.query(queryVector, { topK: 3, returnValues: true, returnMetadata: true });
```
ndisidore, line 94: I don't think the results below align with these calls - here we specify `returnValues` & `returnMetadata` to be true, so we'd expect to get both a `values` property and a `metadata` property in the matching vector result set.

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+Results without metadata filtering:
+
+```json
+{"matches":[{"vectorId":"5","score":0.999909486},{"vectorId":"4","score":0.789848214},{"vectorId":"2","score":0.611976262}]}
```
ndisidore: This is the "old" return format (due to https://github.com/cloudflare/workerd/issues/1435). This will change to the new format in ~a week, so it may be worth using that here.
That new flattened format is documented here
https://github.com/cloudflare/cloudflare-docs/pull/11306/files#diff-df3b21637eae766f21dddc1d934f9d0c934a0fd56d60664798f3c874bede91dfR305

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Metadata filtering
+pcx_content_type: concept
+weight: 6
+---
+
+# Metadata Filtering
+
+In addition to providing an input vector to your query, you can also filter by [vector metadata](/vectorize/learning/insert-vectors/#metadata) associated with every vector. Query results only include vectors that match `filter` criteria.
```
ndisidore: Is it worth calling out how `filter` works with `topK`?
Filter is applied first, then `topK` is taken from the filtered result set. So you'll always get `k` results, unless after filter was applied there were less then `k` available to select from.

content/vectorize/learning/metadata-filtering.md
```diff
@@ -0,0 +1,115 @@
...
+weight: 6
+---
+
+# Metadata Filtering
+
+In addition to providing an input vector to your query, you can also filter by [vector metadata](/vectorize/learning/insert-vectors/#metadata) associated with every vector. Query results only include vectors that match `filter` criteria. 
+
+## Supported operations
+
+Optional `filter` property on `query()` method specifies metadata filter.
```
ndisidore, line 15: Worth noting the types of values the `filter` value can be: the allowed value types are `string | number | boolean | null`

content/vectorize/learning/metadata-filtering.md
vy-ton: I omitted for presentation but agree. Will have them match

content/vectorize/changelog.md
```diff
@@ -7,6 +7,18 @@ rss: file
 
 # Changelog
 
+## 2023-12-06
+
+### Metadata filtering
+
+Vectorize now supports [metadata filtering](/vectorize/learning/query-vectors/metadata-filtering) with equals (`$eq`) and not equals (`$neq`) operators. Metadata filtering limits `query()` results to only vectors that fulfill new `filter` property.
```
netgusto:
```suggestion
Vectorize now supports [metadata filtering](/vectorize/learning/query-vectors/metadata-filtering) with equals (`$eq`) and not equals (`$ne`) operators. Metadata filtering limits `query()` results to only vectors that fulfill new `filter` property.
```

content/vectorize/learning/metadata-filtering.md
sejoker: I think we should call out that we restrict "." in the property names as a result of supporting "." for property nesting.

content/vectorize/changelog.md
elithrar: I'd add a sentence/new paragraph that describes why this is useful to the non-expert too - e.g.

```
By using metadata filtering to limit the scope of a query, you can filter by specific customer IDs, tenant, product category or any other metadata you associate with your vectors.
```

content/vectorize/learning/metadata-filtering.md
sejoker: Also worth mentioning: we don't support existing indexes migration. We should say we are prioritising work to support larger indexes and we are planning to provide an API for listing vectors.

content/vectorize/learning/metadata-filtering.md
vy-ton: updated metadata examples, filed https://jira.cfdata.org/browse/PCX-9199 to fix the rest of query examples in Vectorize docs fyi @Maddy-Cloudflare

content/vectorize/learning/metadata-filtering.md
netgusto: Not specifically for metadata filtering, but stored metadata in general: metadata object keys cannot be empty, contain the dot character (.), contain the double-quote character ("), or start with the dollar character ($)

content/vectorize/learning/metadata-filtering.md
vy-ton: And nested objects?
```
{ "someKey": { "$ne": true } }
{ "pandas.nice": 42 } // looks for { "pandas": { "nice": 42 } }
```

content/vectorize/learning/metadata-filtering.md
vy-ton: added migration note, leaving roadmap/prioritization out of docs

content/vectorize/learning/metadata-filtering.md
ndisidore: Yep, and objects that have a key of the the supported operations with a value matching the same types above
The final TS type is
```
string | number | boolean | null | { '$eq'?: string | number | boolean | null; '$ne'?: string | number | boolean | null }
```

content/vectorize/learning/metadata-filtering.md
vy-ton: > Not specifically for metadata filtering, but stored metadata in general: metadata object keys cannot be empty, contain the dot character (.), contain the double-quote character ("), or start with the dollar character ($)

All noted in the `Supported Operations` section

PR #12099, opened 2023-12-06
Create Workers and APO - New page

DRayCloudflare:
Based on: https://chat.google.com/room/AAAAHL3LtZo/tGanuv5bmiI

### DIFF
--- a/content/automatic-platform-optimization/reference/Workers and APO
+++ b/content/automatic-platform-optimization/reference/Workers and APO
@@ -1,0 +1,10 @@
+---
+title: Workers and APO
+pcx_content_type: reference
+weight: 15
+---
+
+#Using a worker route and APO on the same request
+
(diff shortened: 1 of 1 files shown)


content/automatic-platform-optimization/reference/Workers and APO
```diff
@@ -0,0 +1,9 @@
+---
+title: Workers and APO
+pcx_content_type: reference
+weight: 15
+---
+
+#Using a worker route and APO on the same request
+APO is implemented by using Cloudflare First-Party Workers. These First-Party Workers are Cloudflare Workers with special permissions.
```
nofable, line 8: the concept of a first party worker is not an externally shared concept. I don't think we should expose this terminology. We can just say that APO runs before workers.

kodster28:
Howdy @DRayCloudflare, think this might make more sense broadening the current [page rule page](https://developers.cloudflare.com/automatic-platform-optimization/reference/page-rule-integration/) to instead be a page on integrations in general.

Let me know if you'd need any help with that.

kodster28:
Closing as this'll likely require more work + convo and should be addressed via a PCX ticket.

PR #12104, opened 2023-12-06
[Docs] Fix invalid aside types

pedrosousa:
Officially, we only support `note` and `warning` asides (i.e., these are the only types that have specific CSS styles).


KianNH:
We could catch this during the build by raising an error in the shortcode - if that's something PCX would be happy with then I can throw up the PR.

image

Edit: this could also apply to stuff like `inline-pill` as well.

kodster28:
> We could catch this during the build by raising an error in the shortcode - if that's something PCX would be happy with then I can throw up the PR.
> 
> image
> 
> Edit: this could also apply to stuff like `inline-pill` as well.

Appreciate the offer, @KianNH!

For now, I think we'll keep our linting a bit looser && I'll push up something soon that'll explicitly set a default and only change that default if it matches a selected list of values.

kodster28 (approved):

PR #12110, opened 2023-12-06
[Logs] Update and move API token permissions

angelampcosta:


content/logs/get-started/enable-destinations/r2/index.md
```diff
@@ -54,6 +48,11 @@ Enable Logpush to R2 via the dashboard.
 
 ## Manage via API
 
+If you want to use Logpush via API, create a Cloudflare API token with the following permissions:
+
+- Zone scope, logs edit permissions
+- Account scope, R2 edit permissions
```
deadlypants1973:
```suggestion
- Zone scope, logs edit permissions.
- Account scope, R2 edit permissions.
```

deadlypants1973 (approved):

content/logs/get-started/enable-destinations/r2/index.md
```diff
@@ -54,6 +48,11 @@ Enable Logpush to R2 via the dashboard.
 
 ## Manage via API
 
+If you want to use Logpush via API, create a Cloudflare API token with the following permissions:
```
soheiokamoto: This (before or after the change) is not described very clear to me.
Where is this token used? For creating Logpush jobs AND creating R2 token?

angelampcosta: @soheiokamoto Would it be clearer like this: In order to set up a Logpush job for sending your logs to R2 via API, you will need to generate an [API token](https://dash.cloudflare.com/profile/api-tokens). Make sure that the generated token has the following permissions before integrating it into your request:

soheiokamoto: It still doesn't sound right to me.
The API token for Logpush API usage should be just referred to https://developers.cloudflare.com/logs/get-started/permissions/#tokens

Mentioning zone scope for logs edit permission is confusing, too, as customers can be creating account scope Logpush jobs.

And, creating Logpush jobs via Logpush API itself does not require R2 permission as far as I understand. Generating R2 token may, but that would be done via UI most likely. This page does not mention about API token to generate R2 token, and I could not find the info: https://developers.cloudflare.com/r2/api/s3/tokens/

angelampcosta: Perhaps we could just remove that paragraph, since it's confusing and it's not adding any relevant information.

PR #12113, opened 2023-12-06
DNS Firewall update

AlphaK5:
prefix attack mitigation trigger
PCX-9194


kodster28:
Dupe of #11483, which was previously closed and requested clarification in a ticket.

PR #12127, opened 2023-12-06
Update change-nameservers.md

kodster28:
This section doesn't seem to make sense in the onboarding flow b/c custom nameservers are a Biz+ feature & APO / generic WP plugin is available to everyone.

cc: @william-woodhead?


nofable:
yes I agree that this section doesn't make sense here. we can remove it

RebeccaTamachiro (approved):

PR #12136, opened 2023-12-07
Configure Token Auth move /support/

jedecf:
Move: https://developers.cloudflare.com/support/firewall/learn-more/configuring-token-authentication/#implement-token-creation with redirect


kodster28:
Maybe a broader question for @pedrosousa, is this content still relevant / useful?

pedrosousa:
Yes, it's still useful. I believe there are some repeated examples we can refactor, but the token generation examples are still relevant.

pedrosousa (changes requested):
Requesting changes to prevent a merge before we update the Workers HMAC token auth example (to be addressed very soon).

Update: Waiting on @penalosa's contribution to update the Workers example, which should happen in Jan 2024.

pedrosousa:
The Workers example was updated in #12466. We can now go ahead with this PR.

pedrosousa (approved):

PR #12162, opened 2023-12-08
Fix the URL in aws bedrock cURL example.

cyrilis:
Fix the URL in aws bedrock cURL example.


kodster28:
Looks like this has been addressed in another change.

Sorry it took so long to respond to this PR, @cyrilis.

PR #12170, opened 2023-12-08
[CAT] Challenges/Turnstile default behavior

### LINKED ISSUE #12102: Update turnstile FAQ and other pages to clarify how to use turnstile on sites that are proxied by Cloudflare

Turnstile

* https://developers.cloudflare.com/turnstile/migration/
* https://developers.cloudflare.com/turnstile/frequently-asked-questions/

I am using Cloudflare DNS/nameservers to proxy a few domains for DDoS protection. I recently set up an email server (listmonk.app) on a subdomain and enabled the cloudflare proxy to protect my IP from DDoS attacks. 

My concern is that when people click the "unsubscribe" links they will be presented with a captcha puzzle and will close the tab and click "spam" instead. You can lower your security level to “Essentially off”, but that's not a good fix because it does it for the entire domain instead of just the subdomain that listmonk is installed on. I know that [Cloudflare uses hcaptcha by default](https://blog.cloudflare.com/moving-from-recaptcha-to-hcaptcha/). I learned about turnstile and it seems to be the solution for this. But there is no simple checkbox in https://dash.cloudflare.com/ to switch a website or subdomain to turnstile instead of hcaptcha. 

So I landed on the "migration from hcaptcha" page https://developers.cloudflare.com/turnstile/migration/migrating-from-hcaptcha/ since that seems to be the only option. Besides "inserting the Turnstile script snippet in your HTML’s `<head>` element", I have no clue what the rest of the instructions are telling me to do. Note that I have done nothing to set up hcaptcha. As far as I know, hcaptcha is built in and automatically enabled when you proxy a domain/subdomain with cloudflare.

After clicking on the recaptcha tab https://developers.cloudflare.com/turnstile/migration/migrating-from-recaptcha/ my understanding of these "migration" directions changed, and it now seems that they do not apply to my situation at all, and are rather for people who have explicitly set up these captchas on their website?

Can you please clarify the docs for people in my situation? One specific request would be to mention whether or not the default hcaptcha gets automatically disabled if someone in my case enables turnstile on their website. Or would that just run turnstile on top of hcaptcha? 

There really should be an option in the https://dash.cloudflare.com/ to "use turnstile instead of hcaptcha", but I don't see any place to send feedback for that. 

### Additional information

_No response_

patriciasantaana:
PCX-9203

### DIFF
--- a/content/turnstile/_index.md
+++ b/content/turnstile/_index.md
@@ -51,6 +51,10 @@ Assess the number of challenges issued, evaluate the challenge solve rate, and v
 Cloudflare bot solutions identify and mitigate automated traffic to protect your domain from bad bots.
 {{</related>}}
 
+{{<related header="WAF" href="/waf/" product="waf">}}
+Get automatic protection from vulnerabilities and the flexibility to create custom rules.
+{{</related>}}
+
 ---
 
 ## Availability
--- a/content/turnstile/_index.md
+++ b/content/turnstile/_index.md
@@ -13,7 +13,7 @@ Turnstile is Cloudflare’s smart CAPTCHA alternative. It can be embedded into a
 
 ![Turnstile Overview](/images/turnstile/turnstile-overview.png)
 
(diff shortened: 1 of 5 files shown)


kodster28:
closes #12102

ranbel (approved):

Issue #12179, opened 2023-12-10
Cloudflare should say which versions of the AWS Go SDK R2 is supposed to work with (or just fork the AWS Go SDK)


1f604:
### Which Cloudflare product does this pertain to?

R2

### Existing documentation URL(s)

https://developers.cloudflare.com/r2/examples/aws/aws-sdk-go/

### What changes are you suggesting?

I wish to add the following text at the top of the document:

The code presented in this document was written for the following version of the AWS Go SDK (put this in your `go.mod` file):
```
require (
	github.com/aws/aws-sdk-go-v2 v1.18.0
	github.com/aws/aws-sdk-go-v2/config v1.18.18
	github.com/aws/aws-sdk-go-v2/credentials v1.13.17
	github.com/aws/aws-sdk-go-v2/service/s3 v1.29.0
)
```

### Additional information

I tried to run the example script here: https://developers.cloudflare.com/r2/examples/aws/aws-sdk-go/

And I got this error:

    operation error S3: ListBuckets, https response error StatusCode: [501](https://community.cloudflare.com/t/community-tip-fixing-error-501-not-implemented/44644/), RequestID: , HostID: , api error NotImplemented: ListBuckets search parameter x-id not implemented

This looks to me like AWS updated their Go SDK to query for a new parameter x-id (no idea what that is) and R2 hasn’t implemented it yet.

However the fact that this script was on the official R2 docs suggests to me that it must have worked at some point in time.

So I downgraded my AWS SDK version to this:

require (
	github.com/aws/aws-sdk-go-v2 v1.18.0
	github.com/aws/aws-sdk-go-v2/config v1.18.18
	github.com/aws/aws-sdk-go-v2/credentials v1.13.17
	github.com/aws/aws-sdk-go-v2/service/s3 v1.29.0
)

And lo and behold, it works now.

So my suggestion is to state on the Docs which versions of the SDK R2 is supposed to work with, otherwise AWS is going to keep changing their SDK so that it doesn’t work with R2.

Another way is for Cloudflare to fork the SDK so as to prevent AWS from making changes to it that make it stop working with R2.

See: https://community.cloudflare.com/t/cloudflare-should-say-which-versions-of-the-sdk-r2-is-supposed-to-work-with-or-just-fork-the-aws-sdk/590192

1f604:
I made a PR for this here: https://github.com/cloudflare/cloudflare-docs/pull/12180

KianNH:
[...]

PR #12180, opened 2023-12-10
Add AWS Go SDK version to aws-sdk-go.md

1f604:
See https://github.com/cloudflare/cloudflare-docs/issues/12179


dcpena:
Closing this PR because this was taken care of in a different update. Also, we try not to list specific versions because then it's a little more difficult in the future to validate every new version.

PR #12181, opened 2023-12-10
Remove --cask option

dirtyhenry:
The cask flag was deprecated a while ago and running `brew install --cask ...` will return

```shell
brew install --cask cloudflare/cloudflare/cf-terraforming
Warning: Cask 'cf-terraforming' is unavailable: '/opt/homebrew/Library/Taps/cloudflare/homebrew-cloudflare/Casks/cf-terraforming.rb' does not exist.
```

### DIFF
--- a/content/terraform/advanced-topics/import-cloudflare-resources.md
+++ b/content/terraform/advanced-topics/import-cloudflare-resources.md
@@ -26,7 +26,7 @@ If you use Homebrew on macOS, open a terminal and run the following commands:
 
 ```sh
 $ brew tap cloudflare/cloudflare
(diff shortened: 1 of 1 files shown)


kodster28:
@jacobbednarz, mind providing technical sign off here?

jacobbednarz:
> cask flag was deprecated a while ago

to clarify, this isn't the issue. the `--cask` flag is still the way to install a cask via homebrew. 

the thing that changed here was how we packaged cf-terraforming; it was switched from a cask to a formula to not require notarisation and signing.

jacobbednarz (approved):

kodster28 (approved):

PR #12214, opened 2023-12-12
[Chore] Update actions to work together

kodster28:


kodster28:
This level of config would probably require a full GH app

PR #12225, opened 2023-12-13
[Gateway/Intel] Content and security category sources

ticustodio14:
Based on text provided by Product team on PCX-8795

Updated the Categorization description


maxvp:
Tracking with #12987.

PR #12229, opened 2023-12-13
Update index.md

AlphaK5:
User permissions that are necessary to have.

### DIFF
--- a/content/logs/about/index.md
+++ b/content/logs/about/index.md
@@ -22,6 +22,13 @@ Users without an Enterprise plan can still access [Workers Trace Events Logpush]
 
 {{</Aside>}}
 
+## User Permissions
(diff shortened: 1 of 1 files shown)


angelampcosta:
Hi @AlphaK5 I think the information you want to add it's already available [here](https://developers.cloudflare.com/logs/get-started/permissions/).

AlphaK5:
> Hi @AlphaK5 I think the information you want to add it's already available [here](https://developers.cloudflare.com/logs/get-started/permissions/).

You are right @angelampcosta. Wondering if that section should be more visible, but yes, we have already the information. Maybe a quick link for that section?

angelampcosta:
Sure! I can add a link.

PR #12259, opened 2023-12-15
Audit workers AI getting started guide

rita3ko:
Found two things that weren't quite as in the guide: 
1. Output of response was a bit different due to default system prompt
2. Since you have to use wrangler dev --remote, you're already logged in, so that step was superflous (and results in an error)


kodster28:
closing in favor of #14086

PR #12264, opened 2023-12-16
Fix openai-sdk-streaming.md example

### LINKED ISSUE #12060: Cloudflare Workers OpenAI stream example isn't streaming

Workers

https://developers.cloudflare.com/workers/examples/openai-sdk-streaming/

This example should pipe the stream to the response. It is currently awaiting the entire stream before returning the response which defeats the whole purpose of using a stream.

Your worker streams docs has the right idea: https://developers.cloudflare.com/workers/runtime-apis/streams/#streams

### Additional information

_No response_

aantipov:
The provided example is wrong because the Response is returned only after the streaming from OpenAI is finished. Hence the whole OpenAI response is returned in a single chunk.

Wrapping writing to the stream into self-invoked async function fixes the problem because the Response is then returned right away without waiting for Stream to finish

### DIFF
--- a/content/workers/examples/openai-sdk-streaming.md
+++ b/content/workers/examples/openai-sdk-streaming.md
@@ -31,12 +31,14 @@ export default {
 		const textEncoder = new TextEncoder();
 
 		// loop over the data as it is streamed from OpenAI and write it using our writeable
-		for await (const part of stream) {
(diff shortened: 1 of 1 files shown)


kodster28:
Howdy @aantipov, sorry we took so long on addressing this.

Closing out this PR in favor of #14336

PR #12287, opened 2023-12-18
Update certificate-authorities.md

AlphaK5:
sharing the solution for the LE rate limiting

### DIFF
--- a/content/ssl/reference/certificate-authorities.md
+++ b/content/ssl/reference/certificate-authorities.md
@@ -41,7 +41,8 @@ For publicly trusted certificates, Cloudflare partners with different certificat
 #### Limitations
 
 * Hostname on certificate can contain up to 10 levels of subdomains.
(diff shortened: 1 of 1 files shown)


nfyshhsyny113:
هنگام بازدید از www.okx.com/ با خطا مواجه شدم.

کد خطا: 1020
شناسه اشعه: 83c60f6a8e3f5d5b
کشور: IR
مرکز داده: fra14
IP: 2a01:5ec0:9021:889c:8c64:100b:663d:368f
مهر زمانی: 2023-12-28 01:34:19 UTC

RebeccaTamachiro:
Thanks for your suggestion! Since this is a reference page, adding that solution would be out of context - not every customer will necessarily find a problem with this limitation. As discussed before, PCX-9202 will be adding further information on this to a troubleshooting page once the related error messages are also surfaced to users.

PR #12296, opened 2023-12-18
[ZT] Access custom block page is Enterprise-only

ranbel:


content/cloudflare-one/applications/custom-pages.md
```diff
@@ -28,6 +28,10 @@ The login page is now updated for all of your Access applications.
 
 ## Block page
 
+{{<Aside type="note">}}
+Only available on Enterprise plans.
```
deadlypants1973, line 32: I'd add the feature or product by name explicitly in full sentence here.

X is only....

deadlypants1973 (approved):

content/cloudflare-one/applications/custom-pages.md
ranbel: In the ZT docs, we're using "Only available..." when it appears directly under a feature heading.

PR #12308, opened 2023-12-19
Update known-issues.md

DRayCloudflare:
After doing some testing to help customer with ticket: #3068577, figured it would be good to add this bit of information in to our documentation.

I had thought that maybe it should even be added into the Custom domains page directly, but was worried about making that page too bloated, so decided to just put it here for now.


deadlypants1973 (changes requested):
Can you move this to the relevant section in the Pages Custom Domain docs?

We should be documenting known issues surround a feature on the feature page rather than on a separate known issues page. Better discoverability for users.

DRayCloudflare:
Ok, good point.  have created a brand new request for updating it on the custom domains page: https://github.com/cloudflare/cloudflare-docs/pull/12337

deadlypants1973:
resolved in #13232

PR #12350, opened 2023-12-22
Response Header Modification: Add `set-cookie` caveat

bregenspan:
I'm not sure about exact wording here (suggestions/rewrites very welcome), but I think it could be very useful to call out the fact that the set-cookie header is a special case (in terms of IETF standards), and that this results in caveats for its use with the current version of the Response Header Modification feature.

I think the Response Header Modification functionality may have been written with a more general header case in mind ([RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.3:~:text=Multiple%20message%2Dheader%20fields,fields%20with%20the%20same)), whereas `set-cookie` is a special case in which multiple headers of the same name can be sent AND can't be collapsed together into a single header. The exceptional behavior of `set-cookie` leads to design constraints for any feature like Response Header Modification, and necessarily involves tradeoffs. I think it would be very valuable to list this particular tradeoff in this section of the documentation, because it's (IMO) non-obvious.

### DIFF
--- a/content/rules/transform/response-header-modification/_index.md
+++ b/content/rules/transform/response-header-modification/_index.md
@@ -40,5 +40,5 @@ To modify HTTP headers in the **request**, refer to [HTTP request header modific
 
 * Modifying `cache-control`, `CDN-Cache-Control`, or `Cloudflare-CDN-Cache-Control` headers will not change the way Cloudflare caches an object. Instead, you should create a [Cache Rule](/cache/how-to/cache-rules/).
 
-* Setting the `set-cookie` header will result in overwriting any preexisting `set-cookie` headers, including those added by other Cloudflare products. This includes the header responsible for setting a cookie used by Bot Management.
+* Setting the `set-cookie` header will result in overwriting all preexisting `set-cookie` headers, including those added by other Cloudflare products (e.g. Bot Management).
   * Instead, use "Add" to add a new `set-cookie` header. (Note that only a string literal value is supported at this time). 
--- a/content/rules/transform/response-header-modification/_index.md
+++ b/content/rules/transform/response-header-modification/_index.md
@@ -40,5 +40,6 @@ To modify HTTP headers in the **request**, refer to [HTTP request header modific
 
 * Modifying `cache-control`, `CDN-Cache-Control`, or `Cloudflare-CDN-Cache-Control` headers will not change the way Cloudflare caches an object. Instead, you should create a [Cache Rule](/cache/how-to/cache-rules/).
 
-* Setting the `set-cookie` header will result in overwriting all preexisting `set-cookie` headers, including those added by other Cloudflare products (e.g. Bot Management).
-  * Instead, use "Add" to add a new `set-cookie` header. (Note that only a string literal value is supported at this time). 
+* To add a `set-cookie` header to the response, make sure you use the _Add_ operation instead of _Set static_/_Set dynamic_. Using one of the _Set_ operations will remove any `set-cookie` headers already in the response, including those added by other Cloudflare products such as Bot Management.
+
+* Currently you can only use the _Add_ operation with a literal string value.
--- a/content/rules/transform/response-header-modification/_index.md
+++ b/content/rules/transform/response-header-modification/_index.md
@@ -40,4 +40,5 @@ To modify HTTP headers in the **request**, refer to [HTTP request header modific
 
 * Modifying `cache-control`, `CDN-Cache-Control`, or `Cloudflare-CDN-Cache-Control` headers will not change the way Cloudflare caches an object. Instead, you should create a [Cache Rule](/cache/how-to/cache-rules/).
 
-* Modifying the `set-cookie` header will result in overwriting any preexisting `set-cookie` headers, including those added by other Cloudflare products. This includes the header responsible for setting a cookie used by Bot Management.
+* Setting the `set-cookie` header will result in overwriting any preexisting `set-cookie` headers, including those added by other Cloudflare products. This includes the header responsible for setting a cookie used by Bot Management.
+  * Instead, use "Add" to add a new `set-cookie` header. (Note that only a string literal value is supported at this time). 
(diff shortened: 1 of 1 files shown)


content/rules/transform/response-header-modification/_index.md
```diff
@@ -39,3 +39,5 @@ To modify HTTP headers in the **request**, refer to [HTTP request header modific
 * Any response header modifications will also apply to Cloudflare error pages and [custom error pages](/support/more-dashboard-apps/cloudflare-custom-pages/configuring-custom-pages-error-and-challenge/).
 
 * Modifying `cache-control`, `CDN-Cache-Control`, or `Cloudflare-CDN-Cache-Control` headers will not change the way Cloudflare caches an object. Instead, you should create a [Cache Rule](/cache/how-to/cache-rules/).
+
+* Modifying the `set-cookie` header will result in overwriting any preexisting `set-cookie` headers, including those added by other Cloudflare products. This includes the header responsible for setting a cookie used by Bot Management.
```
bregenspan: I'm sort of conflating two things here:

 * The fact that in this case, the rule is not exactly "modifying a header", it's removing any and all preexisting headers of the same name and then adding a new header in their place.
 * The fact that Response Header Modification appears to run after the stage where Bot Management attaches its own `set-cookie` header.

pedrosousa:
Hi @bregenspan 
Thank you for your PR and all the context you provided.

We already support the specific case of `set-cookie` headers through the "Add" operation, instead of using the "Set static" or "Set dynamic" operations. The "Add" operation is described in Developer Docs, namely in the [Cloudflare dashboard instructions](https://developers.cloudflare.com/rules/transform/response-header-modification/create-dashboard/#:~:text=Add%20%E2%80%94%20Adds%20an%20HTTP%20response%20header%20with%20a%20static%20string%20value.%20This%20operation%20will%20not%20remove%20any%20existing%20response%20headers%20with%20the%20same%20name.):
> _Add_ — Adds an HTTP response header with a static string value. This operation will not remove any existing response headers with the same name.

Therefore, to set cookies using the modify response headers functionality and keep other cookies being set in the response you must use the "Add" operation.

bregenspan:
> Hi @bregenspan Thank you for your PR and all the context you provided.
> 
> We already support the specific case of `set-cookie` headers through the "Add" operation, instead of using the "Set static" or "Set dynamic" operations. The "Add" operation is described in Developer Docs, namely in the [Cloudflare dashboard instructions](https://developers.cloudflare.com/rules/transform/response-header-modification/create-dashboard/#:~:text=Add%20%E2%80%94%20Adds%20an%20HTTP%20response%20header%20with%20a%20static%20string%20value.%20This%20operation%20will%20not%20remove%20any%20existing%20response%20headers%20with%20the%20same%20name.):
> 
> > _Add_ — Adds an HTTP response header with a static string value. This operation will not remove any existing response headers with the same name.
> 
> Therefore, to set cookies using the modify response headers functionality and keep other cookies being set in the response you must use the "Add" operation.

@pedrosousa I think `set-cookie` is a common enough use-case that it could still be useful to call the special case out specifically in the doc here. As you noted, it's not currently possible to set a dynamic header value via Add ([related feature request](https://community.cloudflare.com/t/transform-rules-allow-ability-to-set-multiple-dynamic-headers-of-the-same-name/594221)). That's why people might be tempted to still try Set Dynamic, which can actually work OK *if* one is not relying on another set-cookie header returned by the origin or a Cloudflare product.

(Edit: I just updated the wording to clarify "modifying" => "setting" and to add a note about the Add feature, in case this is an improvement. Thank you for taking a look!)

content/rules/transform/response-header-modification/_index.md
```diff
@@ -39,3 +39,6 @@ To modify HTTP headers in the **request**, refer to [HTTP request header modific
 * Any response header modifications will also apply to Cloudflare error pages and [custom error pages](/support/more-dashboard-apps/cloudflare-custom-pages/configuring-custom-pages-error-and-challenge/).
 
 * Modifying `cache-control`, `CDN-Cache-Control`, or `Cloudflare-CDN-Cache-Control` headers will not change the way Cloudflare caches an object. Instead, you should create a [Cache Rule](/cache/how-to/cache-rules/).
+
+* Setting the `set-cookie` header will result in overwriting all preexisting `set-cookie` headers, including those added by other Cloudflare products (e.g. Bot Management).
+  * Instead, use "Add" to add a new `set-cookie` header. (Note that only a string literal value is supported at this time).
```
pedrosousa:
```suggestion
* To add a `set-cookie` header to the response, make sure you use the _Add_ operation instead of _Set static_/_Set dynamic_. Using one of the _Set_ operations will remove any `set-cookie` headers already in the response, including those added by other Cloudflare products such as Bot Management.

* Currently you can only use the _Add_ operation with a literal string value.
```

content/rules/transform/response-header-modification/_index.md
```diff
@@ -40,5 +40,6 @@ To modify HTTP headers in the **request**, refer to [HTTP request header modific
 
 * Modifying `cache-control`, `CDN-Cache-Control`, or `Cloudflare-CDN-Cache-Control` headers will not change the way Cloudflare caches an object. Instead, you should create a [Cache Rule](/cache/how-to/cache-rules/).
 
-* Setting the `set-cookie` header will result in overwriting all preexisting `set-cookie` headers, including those added by other Cloudflare products (e.g. Bot Management).
-  * Instead, use "Add" to add a new `set-cookie` header. (Note that only a string literal value is supported at this time). 
+* To add a `set-cookie` header to the response, make sure you use the _Add_ operation instead of _Set static_/_Set dynamic_. Using one of the _Set_ operations will remove any `set-cookie` headers already in the response, including those added by other Cloudflare products such as Bot Management.
```
bregenspan, line 43: Thank you, this is much clearer!

pedrosousa (approved):

PR #12352, opened 2023-12-22
[Workers] non-reference wrangler commands need 'npx' in front of them

deadlypants1973:
PCX-9321

As per guidance to install locally, all commands other than reference commands (like /workers/wrangler/commands/) need `npx`in front of them or they will not function.

### DIFF
--- a/content/workers/configuration/cron-triggers/index.md
+++ b/content/workers/configuration/cron-triggers/index.md
@@ -129,7 +129,7 @@ Changes such as adding a new Cron Trigger, updating an old Cron Trigger, or dele
 Test  Cron Triggers using `Wrangler` by passing in the `--test-scheduled` flag to [`wrangler dev`](/workers/wrangler/commands/#dev). This will expose a `/__scheduled` route which can be used to test using a HTTP request. To simulate different cron patterns, a `cron` query parameter can be passed in.
 
 ```sh
-$ wrangler dev --test-scheduled
+$ npx wrangler dev --test-scheduled
 
 $ curl "http://localhost:8787/__scheduled?cron=*+*+*+*+*"
 ```
--- a/content/workers/configuration/secrets.md
+++ b/content/workers/configuration/secrets.md
@@ -37,7 +37,7 @@ To add a secret to a Worker, run the [`wrangler secret put` command](/workers/wr
 ---
 filename: wrangler secret put
 ---
-$ wrangler secret put <KEY>
+$ npx wrangler secret put <KEY>
 ```
 
 #### Via the dashboard
@@ -63,7 +63,7 @@ To delete a secret from your Worker project, run the [`wrangler secret delete` c
 ---
 filename: wrangler secret delete
 ---
-$ wrangler secret delete <KEY>
+$ npx wrangler secret delete <KEY>
 ```
 
 ### Via the dashboard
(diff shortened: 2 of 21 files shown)


irvinebroque:
> As per guidance to install locally, all commands other than reference commands (like /workers/wrangler/commands/) need npx in front of them or they will not function.

Support this change, but just to spell this out (for myself as much as anyone else) to make sure we have a shared understanding of rationale:

- **Without [`npx`](https://docs.npmjs.com/cli/v7/commands/npx):** The globally installed version of wrangler will be used, if present.
- **With [`npx`](https://docs.npmjs.com/cli/v7/commands/npx):** The locally installed version of wrangler will be used, if present (ex: my project has wrangler 3.1.0 in its `package.json`). If wrangler is not installed locally, the latest version of wrangler will be installed (to the npm cache dir) and used.

So if you're to follow our docs today by the book, it's true that without `npx` in front, `wrangler` can't be found, since it's not installed globally and running without `npx` won't pick up a locally installed version.

Our docs currently do not instruct people to install wrangler globally:

https://developers.cloudflare.com/workers/wrangler/install-and-update/

But I wonder if it's worth elevating this note in some way as a "Always use `npx` — here's why:" kind of section that explains ^ the above.

> Running npx wrangler will use the latest version of Wrangler except when there is already a Wrangler version installed in the current folder. In this case, it will use the locally installed version instead.

irvinebroque:
Also wonder if we should be louder about suggesting that people always use `npx` on the reference commands page itself? (https://developers.cloudflare.com/workers/wrangler/commands/)

pedrosousa (approved):

deadlypants1973:
> > As per guidance to install locally, all commands other than reference commands (like /workers/wrangler/commands/) need npx in front of them or they will not function.
> 
> Support this change, but just to spell this out (for myself as much as anyone else) to make sure we have a shared understanding of rationale:
> 
> * **Without [`npx`](https://docs.npmjs.com/cli/v7/commands/npx):** The globally installed version of wrangler will be used, if present.
> * **With [`npx`](https://docs.npmjs.com/cli/v7/commands/npx):** The locally installed version of wrangler will be used, if present (ex: my project has wrangler 3.1.0 in its `package.json`). If wrangler is not installed locally, the latest version of wrangler will be installed (to the npm cache dir) and used.
> 
> So if you're to follow our docs today by the book, it's true that without `npx` in front, `wrangler` can't be found, since it's not installed globally and running without `npx` won't pick up a locally installed version.
> 
> Our docs currently do not instruct people to install wrangler globally:
> 
> https://developers.cloudflare.com/workers/wrangler/install-and-update/
> 
> But I wonder if it's worth elevating this note in some way as a "Always use `npx` — here's why:" kind of section that explains ^ the above.
> 
> > Running npx wrangler will use the latest version of Wrangler except when there is already a Wrangler version installed in the current folder. In this case, it will use the locally installed version instead.

Aimed to address this point in https://developers.cloudflare.com/workers/wrangler/commands/#background without getting into the details of the "we once supported global install in 2021, now we do not in 2023" background story. And also made it clear that this is a reference page and how to run each command with the desired package manager in background, with clear instructions to install locally in install and update page.

deadlypants1973:
> Also wonder if we should be louder about suggesting that people always use `npx` on the reference commands page itself? (https://developers.cloudflare.com/workers/wrangler/commands/)

I'll add a linkout to the background on package managers in install and update page to make it more discoverable and clear that with local install you need to run commands in the way described in the Commands page.

PR #12379, opened 2023-12-30
docs: fix puppeteer binding type

nora-soderlund:
Following the present docs, a type error will occur due to a mismatch in the Fetcher vs BrowserWorker types.

image


nora-soderlund:
Closing.
https://github.com/cloudflare/workerd/issues/1298

PR #12382, opened 2024-01-01
Fixed incorrect domain

jamie-sandbox:
`cloudflared-dns.com` is not a registered domain. Suspected typo.


ranbel (approved):
Thanks for the fix!

PR #12389, opened 2024-01-02
[Hyperdrive] Volatile functions are not cached by hyperdrive

ReppCodes:
Note in our public docs that volatile functions explicitly disable caching behavior in Hyperdrive.

### DIFF
--- a/content/hyperdrive/learning/query-caching.md
+++ b/content/hyperdrive/learning/query-caching.md
@@ -24,11 +24,13 @@ ORDER BY CURRENT_DATE() DESC
 LIMIT 50
 ```
 
-Mutating queries, including `INSERT`, `UPSERT`, or `CREATE TABLE` are not cached:
+Mutating queries (including `INSERT`, `UPSERT`, or `CREATE TABLE`) and queries that use [functions designated as `volatile` by PostgreSQL](https://www.postgresql.org/docs/current/xfunc-volatility.html) are not cached:
 
 ```sql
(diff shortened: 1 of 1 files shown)


elithrar (approved):

content/hyperdrive/learning/query-caching.md
```diff
@@ -24,11 +24,13 @@ ORDER BY CURRENT_DATE() DESC
 LIMIT 50
 ```
 
-Mutating queries, including `INSERT`, `UPSERT`, or `CREATE TABLE` are not cached:
+Mutating queries (including `INSERT`, `UPSERT`, or `CREATE TABLE`), and queries that use functions designated as `volatile` by PostgreSQL, are not cached:
```
a-robinson: Might be worth including a link to the postgres docs here: https://www.postgresql.org/docs/current/xfunc-volatility.html

Also neither comma really belongs in this sentence anymore. So I'd go with:

```suggestion
Mutating queries (including `INSERT`, `UPSERT`, or `CREATE TABLE`) and queries that use [functions designated as `volatile` by PostgreSQL](https://www.postgresql.org/docs/current/xfunc-sql.html) are not cached:
```

It'd be preferable if postgres listed the functions in their docs, but at least linking to what a `volatile` function is is better than nothing.

a-robinson (approved):
Thanks!

Maddy-Cloudflare (approved):

PR #12407, opened 2024-01-04
[Durable Objects] Create Reference section and move content

Maddy-Cloudflare:


content/durable-objects/reference/troubleshooting.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Troubleshooting
 pcx_content_type: concept
-weight: 16
+weight:
```
pedrosousa: You can remove the line if you're not providing a `weight` value.
Alternatively, you can keep the line and set its value to something like `90` to make sure it's the last page in the section.

pedrosousa (approved):

PR #12416, opened 2024-01-05
[Durable Objects] Move content from Learning to Reference

Maddy-Cloudflare:


pedrosousa (approved):

cacheburner:
I'm curious what prompted this change. IMO if I'm a beginner and I open the DO landing page, I'm less likely to open "Reference" than I am to open "Learning". @joshthoward any thoughts?

PR #12424, opened 2024-01-07
Update known-issues.md

DRayCloudflare:
EW-3421
https://community.cloudflare.com/t/cannot-convert-formdata-to-urlsearchparams-in-workers-script/184810
Checking with user manuelmeurer to confirm if it's OK to use his code as a work around.

### DIFF
--- a/content/workers/platform/known-issues.md
+++ b/content/workers/platform/known-issues.md
@@ -77,3 +77,21 @@ Do not use:
 ```js
 await fetch('http://192.0.2.1')
 ```
+
+## FormData() returns empty string
+
+The use of [`FormData()`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) to turn form data into a set of key/value pairs will instead return an empty string.  This can be worked around by manually iterating over the FormData.
+
+An example of how you can do this:
+
+```
+var formBodyEntries = []
+for (var pair of formData.entries()) {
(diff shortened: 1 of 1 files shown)


Cherry:
I don't think this is an issue with a modern [compatibility date](https://developers.cloudflare.com/workers/configuration/compatibility-dates/). `url_standard` should fix this, which has been default since `2022-10-31`.

KianNH:
Confirmed to be fixed by the user via updating the compatibility date: https://community.cloudflare.com/t/cannot-convert-formdata-to-urlsearchparams-in-workers-script/184810/8?u=kiannh

content/workers/platform/known-issues.md
```diff
@@ -77,3 +77,21 @@ Do not use:
 ```js
 await fetch('http://192.0.2.1')
 ```
+
+## FormData() returns empty string
```
deadlypants1973, line 81: Is there a place in the runtime API Workers docs or anywhere else where this could be better suited? We're trying to use Known issues less and less as a page and have known issues listed under the feature page they impact.

DRayCloudflare: Turns out this issue was already covered in: https://developers.cloudflare.com/workers/configuration/compatibility-dates/#formdata-parsing-supports-file

Because it's covered there, I will close this pull request.

The only consideration is if we modify that page to include the word: FormData() for better searchability on that other page.  Is it possible to include meta tags so that if it's googled that page may also come up in the results?

PR #12431, opened 2024-01-08
[Rules] Add examples about mobile redirects

pedrosousa:
Addresses PCX-9353.

### DIFF
--- a/content/rules/url-forwarding/single-redirects/examples.md
+++ b/content/rules/url-forwarding/single-redirects/examples.md
@@ -184,3 +184,58 @@ Request URL                             | Target URL                        | St
 `example.com/en-us/calendar?view=month` | `example.com/calendar?view=month` | `301`
 `example.com/meet-our-team`             | (unchanged)                       | n/a
 `example.com/robots.txt`                | (unchanged)                       | n/a
+
+## Perform mobile redirects
+
+The following examples will redirect visitors using mobile devices — based on the request user agent string — to a different hostname.
+
+### Redirect mobile users dropping the original URI path
+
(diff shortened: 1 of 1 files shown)


content/rules/url-forwarding/single-redirects/examples.md
```diff
@@ -184,3 +184,58 @@ Request URL                             | Target URL                        | St
...
+### Redirect mobile users dropping the original URI path
+
+This example static redirect will redirect requests for the current zone (`example.com`) from mobile users to `m.example.com` without preserving the URI path in the original HTTP request.
+
+{{<example>}}
+
+**When incoming requests match**
+
+* Enter the following expression in the [Expression Editor](/ruleset-engine/rules-language/expressions/edit-expressions/#expression-editor):<br>
+`not http.host in {"m.example.com"} and (http.user_agent contains "mobi" or http.user_agent contains "Mobi")`
```
jason-cf, line 201:
```suggestion
`not http.host in {"m.example.com"} and (http.user_agent matches "[Mm]obi")`
```

jason-cf (approved):

content/rules/url-forwarding/single-redirects/examples.md
jason-cf: optional suggestion: if you decide to use this change instead the rule from contains to matches, you should do it in the other location as well

pedrosousa: We discussed this possible solution internally, but `matches` is an BIZ/ENT operator and we would like to make our examples work in lower plans also (since it will eventually replace existing functionality currently available to all plans).

PR #12434, opened 2024-01-08
Use proper XML indentation for multi-org configuration

aw-cf:
Small change to use the correct XML indentation levels for the inline configuration example


ranbel (approved):
lgtm

PR #12439, opened 2024-01-08
[R2] Document object key length limit

sdnts:


content/r2/reference/limits.md
```diff
@@ -17,6 +17,7 @@ Managed public bucket access through an `r2.dev` subdomain is not intended for p
 | ------------------------------- | ------------------------------------- |
 | Bucket                          | 1000 buckets per account              |
 | Data storage per bucket         | Unlimited                             |
+| Object key length               | 1,024 characters                           |
```
sdnts: Should be "bytes"

dcpena (approved):

PR #12459, opened 2024-01-09
Clarified that O2O for WPEngine does not need Enterprise

sergiozygmunt:
Clarified that O2O for WPEngine does not need Enterprise.

Closes pcx-9432

### DIFF
--- a/content/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/provider-guides/wpengine.md
+++ b/content/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/provider-guides/wpengine.md
@@ -3,7 +3,7 @@ pcx_content_type: configuration
 title: WP Engine
 meta:
     title: WP Engine | Provider guides
-    description: Learn how to configure your Enterprise zone with WP Engine.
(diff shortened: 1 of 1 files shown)


jason-cf:
Waiting on internal approval for this change.

jason-cf (changes requested):
sent change requests via chat

jason-cf (approved):

hulet:
https://developers.cloudflare.com/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/provider-guides/wpengine/ still says "If you use WP Engine **and also have a Cloudflare Enterprise plan**, you can use your own Cloudflare Zone to proxy web traffic to your Zone first, then WP Engine’s (the SaaS Provider) Zone second."

If O2O for WPEngine does not need Enterprise should this be updated to remove the Cloudflare Enterprise stipulation?

I see that `_provider-guide-intro.md` is also used on the Shopify and SFCC Provider guides which both say "You can only enable O2O on the Cloudflare Enterprise plan." Maybe those providers also no longer require Cloudflare Enterprise though?

PR #12471, opened 2024-01-10
[workers] Auto generate playground links for examples

penalosa:
Adds additional code to the Vite parser to generate a link to the workers playground for all workers examples.


layouts/_default/example.html
```diff
@@ -18,14 +18,7 @@
...
         <h1>{{ .Title }}</h1>
-        {{ if .Params.Playground }}
-        <div class="DocsMarkdown--content-column">
-          <a href="{{- .Params.Playground -}}" class="DocsMarkdown--link" target="_blank" rel="noopener">
-            <span class="DocsMarkdown--link-content">Try in Workers Playground</span>
-            {{- partialCached "external.icon" . -}}
-          </a>
-        </div>
-        {{ end }}
+        <workers-playground-link></workers-playground-link>
```
kodster28, line 22: My thought -- at least based on the output in one of the [Pages functions example](https://penalosa-auto-generate-playg.cloudflare-docs-7ou.pages.dev/pages/functions/examples/ab-testing/) -- is that maybe we gate this behind a True/False `.Params.Playground` frontmatter value.

So it'd still be an opt-in, but if you do opt in, then you get the playground link for free.

kodster28: Otherwise, looks great to me!

kodster28, line 22:
```suggestion
        {{ if .Params.Playground }}
        <workers-playground-link></workers-playground-link>
        {{ end }}
```

pedrosousa: **+1** — This Boolean parameter would be useful.

With the current PR, the following page has a playground link which is not very useful:
https://penalosa-auto-generate-playg.cloudflare-docs-7ou.pages.dev/workers/examples/websockets/

The first JavaScript code block in the page is a one-liner. To be able to handle this case, we could either add a way to disable link generation (Boolean parameter described above), or a way to tag the desired JS code block that would be used for the playground. The Boolean parameter sounds fine for now.

penalosa: This is now opt-in with `playground: true`

kodster28 (approved):

PR #12478, opened 2024-01-11
Update faq.md - swap whois.net with IANA URL

wolruf:
www.whois.net doesn't resolve anymore. I suggest the IANA WHOIS which isn't a commercial entity.

### DIFF
--- a/content/dns/troubleshooting/faq.md
+++ b/content/dns/troubleshooting/faq.md
@@ -28,7 +28,7 @@ ___
 
 ## Where do I change my nameservers to point to Cloudflare?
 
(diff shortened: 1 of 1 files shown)


cicku:
lgtm

pedrosousa:
I checked the results for `cloudflare.com` and did not get the results I was expecting (the usual nameservers for the domain, for example). The service seems to be returning the results for the TLD itself (`.com`).

In other documentation pages we're using `https://lookup.icann.org/`, which we could also use here. It seems to return the expected information about the provided domain.

cicku:
The problem with ICANN WHOIS is that it does not support a lot of ccTLD which frankly renders itself useless.

cicku:
Genuinely speaking we need RDAP to replace WHOIS, but it has not been implemented at every gTLD, we need to wait for 2 years before it happens. For ccTLD it will be unknown because ICANN does not manage them directly.

PR #12498, opened 2024-01-12
[Docs] Update `list` layout

pedrosousa:
**Context**

Each section's landing page uses the `list` layout by default, which is too wide for most situations, unlike the `single` layout used for individual pages in a section. To disable this wide area for content we must manually set `layout: single` in the Markdown frontmatter.

**Proposed changes**

This PR changes the default `list` layout to look like the `single` layout — non-wide and displaying a page mini-TOC on the right.

There's also a new `wide` layout for when we really need to have more horizontal space for body text. 
All pages currently set to use the `list` layout are being updated to use the new `wide` layout.
All `_index.md` pages set to use the `single` layout are also being updated to remove this specific layout (the `list` layout now has the same behavior).


kodster28 (approved):
LGTM, agree that it probably makes sense to make the `wide` view non-standard.

PR #12513, opened 2024-01-14
[Frontend] Don't repeat Shift+D if the key is held down

Erisa:
Eliminates potentially seizure-inducing behaviour if Shift+D is held down, where the theme will repeat between dark and light at the key interval rate (which is very fast)

Property used to check: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat

### DIFF
--- a/assets/theme.ts
+++ b/assets/theme.ts
@@ -41,7 +41,7 @@
 
     // Shift+D for toggle
     addEventListener("keydown", (ev) => {
-      if (ev.target !== document.body) return;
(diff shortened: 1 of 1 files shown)


WalshyDev (approved):

WalshyDev:
Alternative is to use key up which IMO is more appropriate than key down here but non-blocking.

Erisa:
That's very true, I don't really mind either way as long as the problem is solved someway.

kodster28 (approved):

PR #12521, opened 2024-01-15
Add DDoS change log entries for 2024-01-15

antoinecordelle:


pedrosousa:
Ideally, we would also update the "Scheduled changes" last update date for feed subscribers:
https://github.com/cloudflare/cloudflare-docs/blob/production/data/changelogs/ddos-http.yaml#L7

pedrosousa (approved):

PR #12523, opened 2024-01-15
[C4SaaS] Add remark about Managed Challenges

pedrosousa:
Addresses PCX-7592.


migueldemoura (changes requested):
Let's discuss this internally before merging.

PR #12535, opened 2024-01-16
[Vectorize] Create Reference section and move content

Maddy-Cloudflare:


content/_redirects
```diff
@@ -1206,6 +1206,17 @@
 /kv/learning/kv-bindings/ /kv/reference/kv-bindings/ 301
 /kv/learning/kv-namespaces/ /kv/reference/kv-namespaces/ 301
 
+# vectorize
+
```
ranbel, line 1210: Do we need a redirect for /vectorize/changelog?

ranbel (approved):

PR #12577, opened 2024-01-18
Update text-embeddings.md to match actual API schema

mhart:
The example response was using old documentation. This updates it to reflect the current API response.


kodster28:
This appears to be out of date with the current re-org, so closing out.

PR #12581, opened 2024-01-18
[Cache] Add info about purge and Transform rules.

angelampcosta:
Closes PCX-9354.

### DIFF
--- a/content/cache/how-to/purge-cache/purge-by-single-file.md
+++ b/content/cache/how-to/purge-cache/purge-by-single-file.md
@@ -47,6 +47,6 @@ For information on how to use single-file purge to purge assets cached by a Work
 
 {{<Aside type="warning" header="Warning">}}
 
-If you have in place a [Transform Rule](/rules/transform/) that is removing part of a URL path, when performing single file purge, you must use the non-transform (end user) URL so that purge can take effect.
+If you have in place a [Transform Rule](/rules/transform/) that is modifying part of a URL path, when performing single file purge, you must use the non-transform (end user) URL so that purge can take effect.
 
 {{</Aside>}}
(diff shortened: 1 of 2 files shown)


content/cache/how-to/purge-cache/purge-by-single-file.md
```diff
@@ -43,4 +43,10 @@ Always use UTF-8 encoded URLs for single-file cache purges. Wildcards are not su
 
 For information on how to use single-file purge to purge assets cached by a Workers fetch, refer to [​​Using Workers to purge](/workers/reference/how-the-cache-works/#single-file-purge--assets-cached-by-a-worker).
 
+{{</Aside>}}
+
+{{<Aside type="warning" header="Warning">}}
+
+If you have in place a [Transform Rule](/rules/transform/) that is removing part of a URL path, when performing single file purge, you must use the non-transform (end user) URL so that purge can take effect.
```
zaidoon1: I would change `that is removing part of a URL path` to `that is modifying part of a url` since the issue is not only for removing but it could be changing/adding things. Same with the purbe by prefix messaging.

content/cache/how-to/purge-cache/purge-by-single-file.md
```diff
@@ -43,4 +43,10 @@ Always use UTF-8 encoded URLs for single-file cache purges. Wildcards are not su
 
 For information on how to use single-file purge to purge assets cached by a Workers fetch, refer to [​​Using Workers to purge](/workers/reference/how-the-cache-works/#single-file-purge--assets-cached-by-a-worker).
 
+{{</Aside>}}
+
+{{<Aside type="warning" header="Warning">}}
+
+If you have in place a [Transform Rule](/rules/transform/) that is modifying part of a URL path, when performing single file purge, you must use the non-transform (end user) URL so that purge can take effect.
```
patriciasantaana: Maybe something like this to make it clearer? Same suggestion for the other warning aside too

```suggestion
If you have a [Transform Rule](/rules/transform/) in place that is modifying part of a URL path, you must use the non-transform (end user) URL when performing single file purge so that purge can take effect.
```

patriciasantaana (approved):

PR #12582, opened 2024-01-18
[d1]: Create query-databases.md

elithrar:
This PR:

* Moves the Client API docs under "How To", as they were buried under Reference in the restructure. This is one of the most important docs for D1 and it became harder to find. It is now simply called "Query databases".
* Continues to link "Client API" to this moved doc (for the above reasons) rather than just re-directing the old link. Developers really need to find this doc, and so it's OK if it's in two places.


rozenmd (approved):

elithrar:
New: 

image

elithrar:
(The link checker breaks because it can't follow re-directs. I'll have to fix all of these manually)

Maddy-Cloudflare (approved):

Maddy-Cloudflare (changes requested):
All the other redirects will have to be fixed.

Maddy-Cloudflare (approved):

PR #12598, opened 2024-01-18
[ZT] Add WARP feature availability template

ranbel:


thibault-ml (approved):

dh-cf:
Overall looks good. Any reason why we would not add the same minimum version info to the 'Allow user to enable local network extension' feature as well?

Suggest we wait a couple of days before merging. The client versions in the switch organizations section have not, in fact, been released yet (will be in a few days). While unlikely, the version numbers may change.

Also, when the new client versions are released to GA, we could use this same merge to remove the 'beta' marker next to the 'Allow users to enable local network exclusion' feature and the 'Switch between Zero Trust organizations' feature.

ranbel:
'Allow user to enable local network extension' wasn't included in this PR because I didn't know what versions were supported. Do you have that information handy?

dh-cf:
It would be the same versions as the 'switching multi-orgs' feature.

ranbel:
'Allow user to enable local network extension' would only apply to a subset of WARP modes, right? E.g. only Gateway with WARP and Secure Web Gateway without DNS filtering?

dh-cf:
That is correct (good catch) - it is only applicable to a client in those
two modes (Gateway with WARP and Secure Web Gateway w/o DNS filtering).

Also, the documentation is correct, but the text in this comment '....local
network extension' should be '....local network exclusion'. Just wanted to
be sure that didn't become part of the documentation.

content/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/switch-organizations.md
```diff
@@ -6,6 +6,23 @@ weight: 3
...
+
+| [WARP modes](/cloudflare-one/connections/connect-devices/warp/configure-warp/warp-modes/) | [Zero Trust plans](https://www.cloudflare.com/teams-pricing/) |
+| -- | -- |
+| All modes | All plans  |
+
+| System   | Availability | Minimum WARP version |
+| ---------| -------------| ---------------------|
+| Windows  | ✅           | 2024.1.7.1           |
+| macOS    | ✅           | 2024.1.7.1           |
+| Linux    | ✅           | 2024.1.7.1           |
```
dh-cf:
```suggestion
| Linux    | X           |            |
```

dh-cf:
- Update macOS, Windows GA release versions.
- Linux not supported yet
- For 'allow users to enable local network exclusion' - add a note that this feature is not available if the split tunnel include mode is used.

patriciasantaana (approved):

PR #12606, opened 2024-01-19
[Vale] Lint "sh" codeblocks

KianNH:
A draft PR whilst PCX can discuss if scenarios where the command and output are separated by prose, like [this](https://developers.cloudflare.com/workers/tutorials/connect-to-turso-using-workers/#use-wrangler-to-create-a-workers-project), are proper usage of `sh` code blocks.

If they are, then the lint is going to have a lot of false positives since the output block has no command in it. 

In any case, I thought I'd throw up the draft. The script is [Tengo](https://github.com/d5/tengo) which is a Go-like scripting language which Vale embeds, so it isn't quite Go and therefore some things can't be cleaned up (like instantiating all the `false` variables on one line).

### DIFF
--- a/.github/styles/cloudflare/ShellCodeBlocks.yml
+++ b/.github/styles/cloudflare/ShellCodeBlocks.yml
@@ -1,0 +1,47 @@
+---
+# Warning: cloudflare.ShellCodeBlocks
+#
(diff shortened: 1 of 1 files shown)


KianNH:
Will make a new PR that is a branch on this repo rather than my fork

PR #12613, opened 2024-01-20
Sthorpe/reference arch refresh

securitypedant:
Preview of changes to reference architecture section


securitypedant:
No longer need draft PR

PR #12623, opened 2024-01-22
[Chore] Opt more workers into playground links

kodster28:
Only opted in examples that worked as is or worked with a touch of logging added

(IMO, examples that render in the playground should display something on the screen or console so it's clear that they produce the correct output)

### DIFF
--- a/assets/events.ts
+++ b/assets/events.ts
@@ -289,12 +289,13 @@ export function zarazTrackDocEvents() {
   addEventListener("DOMContentLoaded", () => {
     if (links.length > 0) {
       for (const link of links as any) {  // Type cast to any for iteration
-        if (link.hostname !== "developers.cloudflare.com") {
-          if (link.href.includes("workers.cloudflare.com/playground#")) {
+        const linkURL = new URL(link)
+        if (linkURL.hostname !== "developers.cloudflare.com") {
+          if (linkURL.hostname === "workers.cloudflare.com" && linkURL.pathname.startsWith("/playground#")) {
             link.addEventListener("click", () => {
               $zarazLinkEvent('playground link click', link);
             });
-          } else if (link.hostname.includes("cloudflare.com")) {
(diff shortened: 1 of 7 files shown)


assets/events.ts
```diff
@@ -290,7 +290,7 @@ export function zarazTrackDocEvents() {
     if (links.length > 0) {
       for (const link of links as any) {  // Type cast to any for iteration
         if (link.hostname !== "developers.cloudflare.com") {
-          if (link.href.includes("workers.cloudflare.com/playground")) {
+          if (link.href.includes("workers.cloudflare.com/playground#")) {
```
kodster28: This ensures that the `playground link click` only fires on events linking to pre-populated code samples, so it's a more direct indicator of intent.

content/workers/examples/alter-headers.md
```diff
@@ -14,6 +14,9 @@ layout: example
 {{<tab label="js" default="true">}}
 
 ```js
+---
+playground: true
+---
 export default {
   async fetch(request) {
     const response = await fetch(request);
```
penalosa:
```suggestion
    const response = await fetch("https://example.com", request);
```
Since the playground has no origin, `fetch(request)` will fail

content/workers/examples/country-code-redirect.md
```diff
@@ -30,6 +33,8 @@ export default {
 
     if (country != null && country in countryMap) {
       const url = countryMap[country];
+      // Remove this logging statement from your final output.
+      console.log(`Based on ${country}-based request, your user would go to ${url}.` )
       return Response.redirect(url);
     } else {
       return fetch(request);
```
penalosa:
```suggestion
      return fetch("https://example.com", request);
```

penalosa (approved):

haleycode (approved):

PR #12625, opened 2024-01-22
[Fundamentals] Correct notifications API Token permission

dom-cf:
Update notifications account permission to correctly reflect read/write access.


kodster28 (approved):

kodster28:
TY, @dom-cf

PR #12688, opened 2024-01-25
[Docs] Update regex to include query strings

ericalli:
The regex `http.request.uri.path` under the **Modify existing URLS to be compatible with transformations in Images** section did not include the query string needed to rewrite the `width` and `height` params accordingly.

Changed to `http.request.uri`.


pedrosousa:
Hi @ericalli 
Thank you for contributing, but I already merged this change in #12685 (it fixed the exact same thing!).

PR #12704, opened 2024-01-26
Update turso.md to correct env vars

hartraft:
Turso integration uses the TURSO_URL and TURSO_AUTH_TOKEN environment variables by default. 

First time users may not notice to update the environment variables in the Cloudflare Turso integration dialog to the examples (LIBSQL_DB_URL & LIBSQL_DB_AUTH_TOKEN) causing unnecessary friction.


kodster28 (approved):
Screenshot 2024-04-26 at 12 03 12 PM

confirmed

PR #12710, opened 2024-01-28
Update index.md

Duncan-Cloudflare:
SPM-2206

### DIFF
--- a/content/analytics/graphql-api/features/filtering/index.md
+++ b/content/analytics/graphql-api/features/filtering/index.md
@@ -110,6 +110,10 @@ The `like` operator is available for string comparisons and supports the `%` cha
 
 ## Examples
 
+{{<Aside type="note">}}
+Filtering times are based on `edgeStartTimestmps` which means requests that end after the filter may be includes in queries (as long as that start within the given time).
+{{</Aside>}}
+
 ### General example
 
 ```graphql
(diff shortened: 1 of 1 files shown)


content/analytics/graphql-api/features/filtering/index.md
```diff
@@ -110,6 +110,10 @@ The `like` operator is available for string comparisons and supports the `%` cha
 
 ## Examples
 
+{{<Aside type="note">}}
+Filtering times are based on `edgeStartTimestmps` which means requests that end after the filter may be includes in queries (as long as that start within the given time).
```
angelampcosta, line 114:
```suggestion
Filtering times are based on `edgeStartTimestmps`, which means requests that end after the filter may be included in queries (as long as they start within the given time).
```

jimhawkridge: `edgeStartTimestamp` is an internal field name. And is only applicable to some datasets.
I think if we say "Filtering times are based on event start timestamps which means..." then it will apply more generally to other datasets as well as the one we discussed in the Google Chat.

jimhawkridge (changes requested):
I suggested a wording change to make this more generally apply to most datasets.

content/analytics/graphql-api/features/filtering/index.md
angelampcosta:
```suggestion
Filtering times are based on event start timestamps, which means requests that end after the filter may be included in queries (as long as they start within the given time).
```

jimhawkridge (approved):

PR #12712, opened 2024-01-28
Remove TCP sockets from beta list

zegevlier:
Hey!

Someone pointed out on [Discord](https://discord.com/channels/595317990191398933/1148642604238524558/1201123582705668207) that TCP sockets are still listed as private beta on the docs site. This is [no longer true](https://blog.cloudflare.com/workers-tcp-socket-api-connect-databases). As I was unable to find any mention of them being beta outside of this page, I removed them from this page altogether.


zegevlier:
Erisa just pointed out that they had opened #12709 before this, so I'll close this one in favour of that.

PR #12730, opened 2024-01-29
[chore] Metadata revisions

kodster28:
Make the values embedded in meta tags surfaceable for search facets
(easier than manipulating them further down the road)


kodster28:
Think this should be solved once we merge in #12733.

Here's my local output once I updated the Hugo config, deleted `public`, and then re-built.
Screenshot 2024-01-29 at 4 33 41 PM

pedrosousa (approved):

PR #12732, opened 2024-01-29
[Workers AI] Adds a Jupyter Notebook tutorial

craigsdennis:
Adds a Jupyter Notebook tutorial that embeds an exported version.

The notebook is exported using [nbconvert](https://github.com/jupyter/nbconvert) to Markdown. 

I chose to use base64 data so we didn't have to host the output of the cells as images. When the notebooks change the images exported are bound to the cell.


craigsdennis:
Hmm didn't realize we wouldn't be able to see the code...wonder if I should "import" the file somehow but keep the top half of the tutorial visible.

irvinebroque (approved):

rita3ko (approved):

PR #12747, opened 2024-01-30
[D1] Remove How to, rename heading, remove use of gerund

Maddy-Cloudflare:


irvinebroque (approved):

content/_redirects
```diff
@@ -205,27 +205,26 @@
 /constellation/ /workers-ai/ 301
 
 # D1
-/d1/client-api/ /d1/how-to/query-databases/ 301
```
pedrosousa, line 208: Unless we're sure that no one is using these old URLs (via analytics), we should keep the redirects and update their target URL:

```diff
- /d1/client-api/ /d1/how-to/query-databases/ 301
+ /d1/client-api/ /d1/manage-databases/query-databases/ 301
```

content/_redirects
```diff
@@ -205,27 +205,26 @@
...
 /d1/platform/migrations/ /d1/reference/migrations/ 301
-/d1/reference/client-api/ /d1/how-to/query-databases/ 301
 /d1/reference/environments/ /d1/configuration/environments/ 301
 /d1/reference/metrics-analytics/ /d1/observability/metrics-analytics/ 301
+/d1/how-to/ /d1/manage-databases/ 301
+/d1/how-to/query-databases/ /d1/manage-databases/query-databases/ 301
+/d1/how-to/using-indexes/ /d1/manage-databases/using-indexes/ 301
+/d1/how-to/querying-json/ /d1/manage-databases/querying-json/ 301
+/d1/how-to/importing-data/ /d1/manage-databases/importing-data/ 301
+/d1/how-to/generated-columns/ /d1/manage-databases/generated-columns/ 301
```
pedrosousa, line 235: I believe you're missing a redirect:
```
/d1/how-to/using-d1-from-pages/ /d1/manage-databases/using-d1-from-pages/ 301
```

content/_redirects
```diff
@@ -205,27 +205,26 @@
...
 /d1/platform/environments/ /d1/configuration/environments/ 301
 /d1/platform/metrics-analytics/ /d1/observability/metrics-analytics/ 301
 /d1/platform/migrations/ /d1/reference/migrations/ 301
-/d1/reference/client-api/ /d1/how-to/query-databases/ 301
 /d1/reference/environments/ /d1/configuration/environments/ 301
 /d1/reference/metrics-analytics/ /d1/observability/metrics-analytics/ 301
+/d1/how-to/ /d1/manage-databases/ 301
+/d1/how-to/query-databases/ /d1/manage-databases/query-databases/ 301
+/d1/how-to/using-indexes/ /d1/manage-databases/using-indexes/ 301
+/d1/how-to/querying-json/ /d1/manage-databases/querying-json/ 301
```
pedrosousa:
```suggestion
/d1/how-to/querying-json/ /d1/manage-databases/query-json/ 301
```
Nit: You could use this opportunity to remove some of the gerund forms from page URLs since you're moving them. Added this suggestion as an example (you would also need to rename the `.md` files).

elithrar (approved):
Approved if @vy-ton also approves.

content/_redirects
Maddy-Cloudflare: Thanks @pedrosousa. I would have done this on a separate PR, but I might as well do it here then.

content/d1/manage-databases/_index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Manage databases
```
vy-ton: Developers will be confused with both `Manage` and `configuration`.

I suggest any of the following `Build`, `Developer 101`, `Basics`

vy-ton: - [ ] Under this heading, can we drop `Query JSON` and `Generated columns` into Reference.
- [ ] The order here should start with `Import data`, `Query databases`, `Use indexes`
- [ ] Let's move `REST API` from Reference under this heading

vy-ton (changes requested):

vy-ton (approved):

patriciasantaana (approved):

PR #12749, opened 2024-01-30
Update trace-request.md

cf-scott:
Added Data Localization Suite as a Limitation as per discussion in chat room.


cf-scott:
I messed up. It's *Data* localization suite. I'm going to try to cancel this request.

PR #12741, opened 2024-01-30
[DLS] Update trace-request.md

luishcardoso:
SPM-2215

PR #12750, opened 2024-01-30
Update trace-request.md

cf-scott:
Added Data Localization Suite as a limitation, as per discussion in trace chat room.


kodster28:
Seems to be the same change as #12741

angelampcosta (approved):

PR #12781, opened 2024-01-31
[IAM]Update scope.md

XabierCloudflare:
It is not clear that users cannot mix between Account Scope and Domain scope roles

### DIFF
--- a/content/fundamentals/setup/manage-members/scope.md
+++ b/content/fundamentals/setup/manage-members/scope.md
@@ -15,11 +15,8 @@ When you assign domain specific roles to account members, you can scope these ro
 You choose the scope of a role when you [add a member](/fundamentals/setup/manage-members/manage/) to your account.
 
 {{<Aside type="note">}}
-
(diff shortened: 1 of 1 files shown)


content/fundamentals/setup/manage-members/scope.md
```diff
@@ -14,6 +14,12 @@ When you assign domain specific roles to account members, you can scope these ro
 
 You choose the scope of a role when you [add a member](/fundamentals/setup/manage-members/manage/) to your account.
 
+{{<Aside type="note">}}
+
+You need to choose to assign either Account Scope roles or Domain Scope role. You will not be able to combine both scope roles for the same member. 
+
+{{</Aside>}}
+
```
patriciasantaana:
```suggestion
{{<Aside type="note">}}
You need to choose to assign either an Account Scope role or Domain Scope role. You cannot combine both scope roles for the same member. 
{{</Aside>}}
```

patriciasantaana (approved):

PR #12818, opened 2024-02-02
[Docs Site] Lint YAML files with Vale

KianNH:
Just a draft to see how good Vale is at linting YAML - it picked up a few issues here and there with the changelogs and glossaries when testing it locally.

cc @kodster28


KianNH:
https://github.com/KianNH/cloudflare-docs/pull/3#discussion_r1476223375

kodster28 (approved):
Tested a bit and LGTM!

PR #12848, opened 2024-02-05
[WAF] Update create-dashboard.md

luishcardoso:
SPM-2230

Document the maximum field size of 2048 bytes for Response body in WAF Custom rules.


pedrosousa:
Thank you for your contribution, @luis-henriquecardoso. Thanks to it, I noticed that we should be reusing this specific piece of content and we're not. I'll update the PR so we reuse the section that is already up to date.

pedrosousa (approved):

PR #12864, opened 2024-02-06
[Pages] env.ASSETS.fetch(): clarify request object

### LINKED ISSUE #12852: Clarify types for env.ASSETS.fetch()

Pages

https://developers.cloudflare.com/pages/functions/api-reference/#envassetsfetch

It took me a while to figure out that this function accepts a URL. Would be nice to clarify that in the API documentation.

### Additional information

_No response_

deadlypants1973:
closes #12852

### DIFF
--- a/content/pages/functions/api-reference.md
+++ b/content/pages/functions/api-reference.md
@@ -37,7 +37,11 @@ The following methods can be used to configure your Pages Function.
 
 ### `env.ASSETS.fetch()`
 
-The `env.ASSETS.fetch()` function allows you to fetch a static asset from your Pages project. Requests passed to the `env.ASSETS.fetch()` function must be to the pretty path, not directly to the asset. For example, if you had the path `/users/index.html`, you will request `/users/` instead of `/users/index.html`. This method call will run the header and redirect rules, modifying the response that is returned.
+The `env.ASSETS.fetch()` function allows you to fetch a static asset from your Pages project.
+
+You can pass a [Request](/workers/runtime-apis/request/), string, or URL to `env.ASSETS.fetch()` function. The URL must be to the pretty path, not directly to the asset.
+
+For example, if you had the path `/users/index.html`, you will request `/users/` instead of `/users/index.html`. This method call will run the header and redirect rules, modifying the response that is returned.
 
 ## Types
 
--- a/content/pages/functions/api-reference.md
+++ b/content/pages/functions/api-reference.md
@@ -39,9 +39,7 @@ The following methods can be used to configure your Pages Function.
 
 The `env.ASSETS.fetch()` function allows you to fetch a static asset from your Pages project.
 
-You can pass a [Request](/workers/runtime-apis/request/), string, or URL to `env.ASSETS.fetch()` function. The URL must be to the pretty path, not directly to the asset.
-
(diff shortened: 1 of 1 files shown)


content/pages/functions/api-reference.md
```diff
@@ -37,7 +37,11 @@ The following methods can be used to configure your Pages Function.
 
 ### `env.ASSETS.fetch()`
 
-The `env.ASSETS.fetch()` function allows you to fetch a static asset from your Pages project. Requests passed to the `env.ASSETS.fetch()` function must be to the pretty path, not directly to the asset. For example, if you had the path `/users/index.html`, you will request `/users/` instead of `/users/index.html`. This method call will run the header and redirect rules, modifying the response that is returned.
+The `env.ASSETS.fetch()` function allows you to fetch a static asset from your Pages project.
+
+You can pass a [Request](/workers/runtime-apis/request/), string, or URL to `env.ASSETS.fetch()` function. The URL must be to the pretty path, not directly to the asset.
```
WalshyDev: I wasn't sure how to phrase the string part and I'm still not.

It's a string of a URL so like `'https://google.com'` I don't think that's clear though? How can we make it clear. I wanted to write `URL string` but I don't think that's clear enough

WalshyDev (approved):

content/pages/functions/api-reference.md
KianNH: MDN says something along the lines of "a string containing the URL" - not sure if I'd say *the URL* and rather *a URL* though.

Most resources leave out the fact that you can use a URL object since that's just it being stringified like anything else. 

One source of confusion will be `workers-types`:

`Fetcher` accepts `Request | string | URL` whereas `next()` only allows `Request | string` - so they might copy what works for the global `fetch()` and `env.ASSETS.fetch()` and suddenly get TS shouting at them (despite it working at runtime).

pedrosousa: Nit: Is "pretty path" something the reader would know what it is and how to get it? I couldn't find other references in the Pages docs. Keeping the example could be useful here.

pedrosousa (approved):

content/pages/functions/api-reference.md
WalshyDev: The example is still below but yeah it definitely needs the example
May want to move the "The URL must be to the pretty path, not directly to the asset." to be in the same para as the example so it's clear that's what it's for

deadlypants1973:
```suggestion
You can pass a [Request object](/workers/runtime-apis/request/), URL string, or URL object to `env.ASSETS.fetch()` function. The URL must be to the pretty path, not directly to the asset. For example, if you had the path `/users/index.html`, you will request `/users/` instead of `/users/index.html`. This method call will run the header and redirect rules, modifying the response that is returned.
```

content/pages/functions/api-reference.md
```diff
@@ -37,7 +37,11 @@ The following methods can be used to configure your Pages Function.
...
+
+For example, if you had the path `/users/index.html`, you will request `/users/` instead of `/users/index.html`. This method call will run the header and redirect rules, modifying the response that is returned.
```
deadlypants1973:
```suggestion
```

PR #12870, opened 2024-02-06
Zero downtime failover clarification

chrisflare:
Replacing text from 3 different sections to align with one another.


kodster28:
For future maintenance, might be easier to update the partial and then reference the partial in the other places.

```
{{<render file="_dns-zero-downtime-failover.md" productFolder="fundamentals">}}
```

kodster28 (approved):
LGTM

PR #12888, opened 2024-02-07
Replace "deno" with "demo" in two places

ckhicks:
I assume this was a typo, since the libraries in question were "demo" repos and have nothing to do with Deno, correct? 👍


WalshyDev:
This was actually correct as Deno :)

We adapted the Deno Postgres and MySQL drivers. 

Thank you for the PR though!

ckhicks:
Haha, very good, keep up the good work!

PR #12936, opened 2024-02-12
[Notifications] Add missing redirect for PagerDuty

Erisa:
Once upon a time, this URL was valid: https://developers.cloudflare.com/notifications/create-pagerduty/
Sometime later, it was moved to https://developers.cloudflare.com/notifications/get-started/configure-pagerduty/

The URL https://support.cloudflare.com/hc/en-us/articles/360047358211 leads to the old path, and it is linked from the Cloudflare PagerDuty integration, so that link currently 404s:

image

The quickest way to fix this and potentially other instances of the problem is to redirect the old PagerDuty URL to the new one.

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -646,6 +646,7 @@
 # notifications
 /fundamentals/notifications/ /notifications/ 301
 /notifications/create-notifications/ /notifications/get-started/ 301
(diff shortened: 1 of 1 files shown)


patriciasantaana (approved):

patriciasantaana:
Thanks for catching this, @Erisa!

PR #12941, opened 2024-02-12
Add DDoS change log entries for 2024-02-12

jdesgats:


data/changelogs/ddos-http.yaml
```diff
@@ -8,6 +8,9 @@ entries:
   scheduled: true
   individual_page: true
   link: '/ddos-protection/change-log/http/scheduled-changes/'
```
antoinecordelle, line 10: We need to bump the date of this entry as well for the RSS feed.

antoinecordelle (approved):

PR #13011, opened 2024-02-15
Clarify Turnstile support for desktop mode

migueldemoura:
Turnstile and challenge pages now support `desktop mode` on mobile browsers.

### DIFF
--- a/content/waf/reference/cloudflare-challenges.md
+++ b/content/waf/reference/cloudflare-challenges.md
@@ -82,6 +82,10 @@ This behavior commonly occurs because an extension modifies your browser's defau
 
 {{</Aside>}}
 
+### Mobile device emulation
+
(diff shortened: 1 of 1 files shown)


pedrosousa:
> Turnstile and challenge pages now support `desktop mode` on mobile browsers.

Does it also support "mobile mode on desktop browsers"? Or was this part of the sentence being removed already outdated?

migueldemoura:
> > Turnstile and challenge pages now support `desktop mode` on mobile browsers.
> 
> Does it also support "mobile mode on desktop browsers"? Or was this part of the sentence being removed already outdated?

Discussing internaly, but tl;dr is we should probably keep that part but phrase it differently, i.e.:
`Challenges are not supported when device emulation is enabled on a browser, for example, using developer tools.`

pedrosousa (approved):

PR #13019, opened 2024-02-15
Fix typo in description of WorkerWallTimeUs

irvinebroque:
refs internal PR EW-8158


content/logs/reference/log-fields/zone/http_requests.md
```diff
@@ -102,7 +102,7 @@ The descriptions below detail the fields available for `http_requests`.
...
+| WorkerWallTimeUs | The elapsed time in milliseconds between the start of a Worker invocation, and when the Workers Runtime determines that no more JavaScript needs to run. Specifically, this measures the wall-clock time that the JavaScript context remained open. For example, when returning a response with a large body, the Workers runtime can, in some cases, determine that no more JavaScript needs to run, and closes the JS context before all the bytes have passed through and been sent. Alternatively, if you use the `waitUntil()` API to perform work without blocking the return of a response, this work may continue executing after the response has been returned, and will be included in `WorkerWallTimeUs`. | int |
```
grounded042: I believe the first sentence also needs to be updated.

```suggestion
| WorkerWallTimeUs | The elapsed time in microseconds between the start of a Worker invocation, and when the Workers Runtime determines that no more JavaScript needs to run. Specifically, this measures the wall-clock time that the JavaScript context remained open. For example, when returning a response with a large body, the Workers runtime can, in some cases, determine that no more JavaScript needs to run, and closes the JS context before all the bytes have passed through and been sent. Alternatively, if you use the `waitUntil()` API to perform work without blocking the return of a response, this work may continue executing after the response has been returned, and will be included in `WorkerWallTimeUs`. | int |
```

pedrosousa: ~Will these changes be merged back to the internal repo?~ Just found the internal PR! Thanks

kvap (approved):

kodster28 (approved):

PR #13070, opened 2024-02-20
add note regarding `getPlatformProxy`'s persist option compared to `--persist-to`

dario-piotrowicz:
add a note in the `getPlatformProxy` docs clarifying that it does not add a `v3` sub directory as `wrangler` does

___

as suggested here: https://github.com/cloudflare/next-on-pages/issues/603#issuecomment-1873476287

### DIFF
--- a/content/workers/_partials/_ai-local-usage-charges.md
+++ b/content/workers/_partials/_ai-local-usage-charges.md
@@ -7,6 +7,6 @@ _build:
 
 {{<Aside type="note" header="Workers AI local development usage charges">}}
 
-Using Workers AI always accesses your Cloudflare account in order to run AI models, and will incur usage charges even in local development.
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
Fine as a docs change but I wonder if we should do something to avoid this complexity in the tooling?

dario-piotrowicz:
> Fine as a docs change but I wonder if we should do something to avoid this complexity in the tooling?

yeah I agree.... the problem is that we cannot remove the `v3` in wrangler as it would be a breaking change... and I guess adding it to `getPlatformProxy` would also be a breaking change 😕 

on hindsight I should have added the `v3` directory right of the bat in `getPlatformProxy` 😓, I guess we can re-align them on the next major release? 😓

deadlypants1973 (approved):

PR #13084, opened 2024-02-21
[DNS] New section for Foundation DNS

RebeccaTamachiro:
PCX-7049


content/dns/foundation-dns/_index.md
```diff
@@ -0,0 +1,29 @@
...
+    * Reduced exposure to incidents or software regression
+    * More consistent nameserver assignment
+* DNSSEC keys unique to your account
+* GraphQL DNS analytics
+
+## Availability
+
+Foundation DNS is only available to Enterprise plans.
+
+Depending on your account settings or specific options you made when onboarding a zone, advanced nameservers may not be enabled. Refer to [Setup TBD](/dns/foundation-dns/setup/).
```
RebeccaTamachiro: Noting here so I don't forget. PM suggested replacing this paragraph with:
```
- By default advanced nameservers are not enabled, but can be enabled on existing zones with one click
- You can configure your account so all new zones added to the account will use advanced nameservers by default
```
Find a middle ground where it's less generic than the original but removing "with one click" and not sounding contradictory with the two different "defaults"

kodster28:
Think we can close this out as the underlying PCX ticket is closed

PR #13097, opened 2024-02-22
[R2] Update public-buckets.md with Zone Hold advisory

andre-cloudflare:
Enterprise zones would have Zone Hold enabled by default. This would prevent the activation of the Custom Domain.

### DIFF
--- a/content/r2/buckets/public-buckets.md
+++ b/content/r2/buckets/public-buckets.md
@@ -61,7 +61,7 @@ To view the added DNS record, select *...* next to the connected domain and sele
 
 {{<Aside type="note">}}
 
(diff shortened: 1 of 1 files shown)


crwaters16 (approved):

content/r2/buckets/public-buckets.md
```diff
@@ -59,6 +59,12 @@ Your domain is now connected. The status takes a few minutes to change from **In
 
 To view the added DNS record, select *...* next to the connected domain and select **Manage DNS**.
 
+{{<Aside type="note">}}
+
+If the zone is on the Enterprise plan, make sure that you [release the zone hold](/fundamentals/setup/account/account-security/zone-holds/#release-zone-holds) before adding the custom domain. A zone hold would prevent the custom subdomain from activating.
```
crwaters16:
```suggestion
If the zone is on an Enterprise plan, make sure that you [release the zone hold](/fundamentals/setup/account/account-security/zone-holds/#release-zone-holds) before adding the custom domain. A zone hold would prevent the custom subdomain from activating.
```

PR #13118, opened 2024-02-23
Update text-to-image.md

Lparksi:
The api must have a "num_steps" input, otherwise it will throw an error.


Lparksi:
According to the original document, the curl operation will receive the following response.
```json
{
  "errors": [
    {
      "message": "Error: unexpected type \"int32\" with value \"undefined\".",
      "code": 1000
    }
  ],
  "success": false,
  "result": {},
  "messages": []
}
```

kodster28:
@craigsdennis, looks like this feedback is still relevant (even after the re-org)?

kodster28:
Howdy @Lparksi, apologies on the delay here... the original doc was trying to work around a bug in the config for that model. Since then, that's been fixed (i.e., `num_steps` is an optional input).

PR #13121, opened 2024-02-23
Update deploy-a-remix-site.md

NieXi:
fix create project command


dario-piotrowicz:
Hi @NieXi, thank you for the PR 🙂 

But anyhow I don't think that the change is correct as the `--` that you are removing is actually necessary to pass the `framework` argument to `create-create` see for example: https://stackoverflow.com/questions/43046885/what-does-do-when-running-an-npm-command

or am I missing something here?

kodster28:
Closing due to lack of a response.

If this is in error, please open another PR / issue in our repo.

PR #13142, opened 2024-02-25
add compatibility_date to wfp metadata example

JoeRoddy:
Currently compatibility_date is missing as a configuration option in the Workers for Platforms documentation.  Adds it below the json example as an optional attribute.

This caused me issues when trying to use `ReadableStream()`.  Even though my `compatibility_date` was set for my WFP codebase, I was not setting it for the user workers (and it is not a required metadata property), so that feature was not working, and I spent a long time trying to figure out how I could even set that attribute.

Uses the same description for `compatibility_date` found here:
https://developers.cloudflare.com/workers/wrangler/configuration/#inheritable-keys

### DIFF
--- a/content/cloudflare-for-platforms/workers-for-platforms/reference/metadata.md
+++ b/content/cloudflare-for-platforms/workers-for-platforms/reference/metadata.md
@@ -80,6 +80,9 @@ At a minimum, the `main_module` key is required to publish a user Worker.
 
   - The path to the module entry point of the user Worker that will be executed. For example, `main.js`.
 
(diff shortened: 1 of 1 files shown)


JoeRoddy:
It would also be great if user workers inherited the `compatibility_date` from the Workers for Platforms project's `wrangler.toml` by default, if it's not already explicitly set when uploading the user worker.

It's a bit unintuitive (imo) that if you don't have to set it for user workers like you do for all other workers, that it wouldn't use the same configuration that you applied for your WFP worker.

KimJ15:

JoeRoddy:
> 👋Hi there, we’re going to close this PR due to inactivity

@KimJ15 Inactivity?  No one responded to it, hah

PR #13206, opened 2024-02-28
Missing arrow in documentation code bash snippet for the placeholder ACCOUNT_ID

abhishekwebcode:
The page assumed all token placeholder in code are represented using < > arrows, in the API key placeholder for the first code snippet the ending arrow is missing, I have added that for the placeholder ACCOUNT_ID
This is needed for consistency in documentation


kodster28 (approved):
Thanks for catching (and fixing) @abhishekwebcode

PR #13233, opened 2024-03-01
[Durable Objects] rename examples

elithrar:
This PR renames the DO pricing examples to add more context to the use-case and "scale" of each example.

### DIFF
--- a/content/durable-objects/platform/pricing.md
+++ b/content/durable-objects/platform/pricing.md
@@ -81,7 +81,7 @@ In this scenario, the estimated monthly cost would be calculated as:
 
 <sup>2</sup> The example uses 1 second because each Durable Object is active for 1 second per minute. This can also be thought of as 432 million requests that each take 10 ms to execute (4,320,000 seconds).
 
-### Example 4: Large-scale real-time app
+### Example: Large-scale real-time app
 
 This example represents a horizontally scaled Durable Objects based application using WebSockets to communicate user-specific state to a single client connected to each Durable Object.
 
--- a/content/durable-objects/platform/pricing.md
+++ b/content/durable-objects/platform/pricing.md
@@ -17,7 +17,7 @@ These examples exclude the costs for the Workers calling the Durable Objects. Wh
 * Inactive objects receiving no requests do not incur any duration charges.
 * The [WebSocket Hibernation API](/durable-objects/reference/websockets/#websocket-hibernation) can dramatically reduce duration-related charges for Durable Objects communicating with clients over the WebSocket protocol, especially if messages are only transmitted occassionally at sparse intervals.
 
-### Example 1
+### Example: Starter app
 
 This example represents a simple Durable Object used as a co-ordination service invoked via HTTP.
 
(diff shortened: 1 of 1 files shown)


content/durable-objects/platform/pricing.md
```diff
@@ -102,6 +81,27 @@ In this scenario, the estimated monthly cost would be calculated as:
 
 <sup>2</sup> The example uses 1 second because each Durable Object is active for 1 second per minute. This can also be thought of as 432 million requests that each take 10 ms to execute (4,320,000 seconds).
 
+### Example 4: Large-scale real-time app
```
a-robinson: Isn't this strictly smaller-scale than the "Medium-scale real-time app" directly above it, with the main difference being that this example doesn't use websocket hibernation? I haven't looked at the history but I suspect this example was meant to contrast with the example above that doesn't use hibernation, and that that is what is meaningful about it, not that it's "large-scale" vs "medium-scale"

a-robinson:
```suggestion
### Example: Large-scale real-time app
```

cacheburner: Not sure I agree with the phrasing of the examples, specifically looking at the scale here.

**Example 3** (Medium scale) 
- 100 DOs with 100 WS
- each client sends a message per minute
- 432 million requests a month

**Example 4** (Large scale) 
- 100 DOs with 1 WS each
- each client sends 1 message per second
- 259 million requests a month

cacheburner: Right, the goal there was to show a comparable example where duration billing pretty much disappears. That said, I'm not sure we're doing a great job at conveying how this makes scaling your real-time app horizontally with many DOs feasible. 

Might be a good idea to add an example where there's +10k Durable Objects each with maybe 100 WebSockets (using hibernation), but the clients mostly just listen to broadcasts from the DO they connect to (low incoming request count -- we don't cover this use case in the pricing AFAICT). Then your request bill would be low, and duration would likely be pretty low too, even with 1M connections

elithrar: ^ Let me scale this one out.

content/durable-objects/platform/pricing.md
```diff
@@ -102,6 +81,27 @@ In this scenario, the estimated monthly cost would be calculated as:
...
+* Clients sent one message every second of the month so that the Durable Objects were active for the entire month.
+
+In this scenario, the estimated monthly cost would be calculated as:
+
+**Requests**:
+- 100 requests to establish the WebSockets.
+- 1 message per second \* 100 connections \* 60 seconds \* 60 minutes \* 24 hours \* 30 days = 259,200,000 requests
+- (259.2 million requests - included 1 million requests) x $0.15 / 1,000,000 = $38.73
+
+**Compute Duration**:
```
Maddy-Cloudflare, line 98:
```suggestion
**Compute duration**:
```

Maddy-Cloudflare (approved):

content/durable-objects/platform/pricing.md
```diff
@@ -102,6 +81,27 @@ In this scenario, the estimated monthly cost would be calculated as:
...
+This example represents a horizontally scaled Durable Objects based application using WebSockets to communicate user-specific state to a single client connected to each Durable Object.
+
+* 100 Durable Objects each have a single WebSocket connection established to each of them.
+* Clients sent one message every second of the month so that the Durable Objects were active for the entire month.
+
+In this scenario, the estimated monthly cost would be calculated as:
+
+**Requests**:
+- 100 requests to establish the WebSockets.
+- 1 message per second \* 100 connections \* 60 seconds \* 60 minutes \* 24 hours \* 30 days = 259,200,000 requests
```
a-robinson, line 95: It still seems like this is lower scale than the "medium-scale" example above? 259M requests here vs 432M there.

elithrar:
Outdated.

PR #13235, opened 2024-03-01
durable-objects: clarify HTTP requests, RPC method calls, etc - pricing

elithrar:
Add the upcoming RPC method calls to the pricing so it is clear how much they cost.


elithrar:
Superseded.

PR #13246, opened 2024-03-01
[Workers AI] Removes weird typing on run function in code examples

craigsdennis:
I am not sure where these came from. I'm assuming a bad copy paste during import?

### DIFF
--- a/layouts/partials/models/code-examples/automatic-speech-recognition.md
+++ b/layouts/partials/models/code-examples/automatic-speech-recognition.md
@@ -23,7 +23,7 @@ export default {
       audio: [...new Uint8Array(blob)],
     };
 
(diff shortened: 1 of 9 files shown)


markdembo:
@craigsdennis this was me. This is the official typescript typing for our models. This helps we constructing and validating our inputs and soon also the response you'll get back from the models (currently typed as `Promise<any>` so not that useful yet. I propose leave as is

UPDATE: I stand to be corrected - not needed at all.

craigsdennis:
So we should squash and merge right?

craigsdennis:
@markdembo is there more?

markdembo (approved):

KimJ15 (approved):

PR #13264, opened 2024-03-04
[D1] Changelog for D1 billing

vy-ton:


data/changelogs/d1.yaml
```diff
@@ -2,6 +2,17 @@
 link: "/d1/platform/changelog/"
 productName: D1
 entries:
+  - publish_date: "2024-03-05"
+    title: Billing for D1 usage
+    description: |-
+      As of 2024-03-05, D1 usage can incur charges for an account's future billing cycle.
```
rozenmd: Can we clarify that we start counting from March 5?

Something like:

> As of 2024-03-05, D1 usage will start to be counted and may incur charges for an account's future billing cycle.

Maddy-Cloudflare (approved):

PR #13277, opened 2024-03-05
Update Hugo to v0.123.5

KianNH:
v0.123.0 was a big update for Hugo: https://github.com/gohugoio/hugo/releases/tag/v0.123.0

This comes with lots of improvements to the build process. Notably, rebuilds will now only build the affected files - this takes rebuilds from ~4s down to ~70ms.

image

Probably worth a review of the documented breaking changes at https://github.com/gohugoio/hugo/issues/11455 to see if any would affect us.


WalshyDev:
> https://github.com/gohugoio/hugo/issues/11556

Don't think we have any symlinks but would be good to check

> https://github.com/gohugoio/hugo/issues/11669

We don't currently have a trailing slash so this is a change for us. I don't see us using it in templates so we should be ok (and should probably update config) but let's sanity check the html

> https://github.com/gohugoio/hugo/issues/11999

We have a few html files in content (e.g. 404, plans). We should again check these

KianNH:
Renamed the top-level Support folders due to an issue with the `Account Management & Billing` one - subfolders & files within them still need to be changed for consistency but that can be handled in another PR.

KianNH:
`content/ddos-protection/change-log/http/2022-09-19 - Emergency-emergency.md` also now has a different path.

image

KianNH:
URL for spot checks: https://1ac94b4e.cloudflare-docs-github.pages.dev/

kodster28:
> > [gohugoio/hugo#11556](https://github.com/gohugoio/hugo/issues/11556)
> 
> Don't think we have any symlinks but would be good to check
> 
> > [gohugoio/hugo#11669](https://github.com/gohugoio/hugo/issues/11669)
> 
> We don't currently have a trailing slash so this is a change for us. I don't see us using it in templates so we should be ok (and should probably update config) but let's sanity check the html
> 
> > [gohugoio/hugo#11999](https://github.com/gohugoio/hugo/issues/11999)
> 
> We have a few html files in content (e.g. 404, plans). We should again check these

- [x] We're not using any symlinks
- [x] HTML files all have frontmatter, so none of their behavior should be changing
- [x] Loose check for trailing slashes

We should probably also update our Pages build variables to mirror this Hugo version change.

I reviewed the other Hugo breaking changes and seems like most of them related to Taxonomies and -- as such -- don't apply to us.

KianNH:
We don't use `.Site.BaseURL` anywhere so adding the trailing slash to the config shouldn't change anything (as far as we're concerned).

kodster28 (approved):

PR #13294, opened 2024-03-06
Remove un-necessary step - No need to making script executable

abhishekwebcode:
We don't need executable permission if we are running the script via bash (by specifying the script as a file argument to bash command) as the documentation suggested in 4th step

So i removed the unnecessary step, running script directly is not preferred as on some systems the user might not have sudo access to make the script executable

Even if the script needed to be directly executed, a shebang is needed which the documented did not provide

Reference: https://askubuntu.com/questions/25681/can-scripts-run-even-when-they-are-not-set-as-executable

This works for sh too

### DIFF
--- a/content/fundamentals/setup/manage-domains/add-multiple-sites-automation.md
+++ b/content/fundamentals/setup/manage-domains/add-multiple-sites-automation.md
@@ -63,12 +63,7 @@ ___
   done
 ```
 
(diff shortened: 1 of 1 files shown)


abhishekwebcode:
@RebeccaTamachiro Can you look into this, its a very small commit sitting idle since a week

PR #13296, opened 2024-03-06
[Fundamentals] Highlight the fact that origin IPs will be returned for pending zones when moving zone between accounts

dom-cf:
The fact that pending zones cannot be proxied is documented on the zone status page, however a number of users run into this specifically when moving domains between accounts. Highlighting this on the move domains page can help ensure customers are aware and prepared for this as it can result in downtime.


RebeccaTamachiro (approved):

RebeccaTamachiro:
Makes sense! Thanks for raising this, @dom-cf

PR #13315, opened 2024-03-07
Adds lightbox support for images

markdembo:
Solution based on:
https://hugocodex.org/add-ons/lightbox/
I thought for a moment, it's hugo official, sadly it's just community.

Automatically converts links to images to lightbox, if they contain an image link. Images without link will not be affected

More comments on the individual files.

### DIFF
--- a/content/kv/reference/how-kv-works.md
+++ b/content/kv/reference/how-kv-works.md
@@ -16,7 +16,7 @@ While reads are periodically revalidated in the background, requests which are n
 
 When you write to KV, your data is written to central data stores. Your data is not sent automatically to every location’s cache, but regional tiers are notified within seconds to do a purge of that key.
 
-![Your data is written to central data stores when you write to KV.](/images/kv/kv-write.svg)
+[![Your data is written to central data stores when you write to KV.](/images/kv/kv-write.svg)](/images/kv/kv-write.svg)
 
 Initial reads from a location do not have a cached value. Data must be read from the nearest regional tier, followed by a central tier, degrading finally to the central store for a truly cold global read. While the first access is slow globally, subsequent requests are faster, especially if requests are concentrated in a single region.
 
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -67,6 +67,6 @@
       {{- end -}}
     </div>
   </div>
-  <script type="text/javascript" src="/lightbox.js"></script>
+  <script type="text/javascript" src="/lightbox.js" defer></script>
   <link rel="stylesheet" href="/lightbox.css">
 </footer>
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -67,4 +67,6 @@
       {{- end -}}
     </div>
   </div>
+  <script type="text/javascript" src="/lightbox.js"></script>
+  <link rel="stylesheet" href="/lightbox.css">
 </footer>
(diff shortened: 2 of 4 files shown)


content/kv/reference/how-kv-works.md
```diff
@@ -6,27 +6,27 @@ weight: 7
...
+KV supports exceptionally high read volumes with low latency, making it possible to build highly dynamic APIs.
 
 While reads are periodically revalidated in the background, requests which are not in cache and need to hit the centralized back end can experience high latencies.
 
 ## Write data to KV and read data from KV
 
 When you write to KV, your data is written to central data stores. Your data is not sent automatically to every location’s cache, but regional tiers are notified within seconds to do a purge of that key.
 
-![Your data is written to central data stores when you write to KV.](/images/kv/kv-write.svg)
+[![Your data is written to central data stores when you write to KV.](/images/kv/kv-write.svg)](/images/kv/kv-write.svg)
```
markdembo, line 19: Example usage. Image needs to be a link to image.

layouts/partials/footer.html
```diff
@@ -67,4 +67,6 @@
       {{- end -}}
     </div>
   </div>
+  <script type="text/javascript" src="/lightbox.js"></script>
+  <link rel="stylesheet" href="/lightbox.css">
```
markdembo: Adds this script to all pages, if we want to be more conservative, this should be changed to a per-page level first. I'd then remove this includes here from the PR and then each now docs writer could include these lines in the individual PR

static/lightbox.css
```diff
@@ -0,0 +1,156 @@
+#lightbox {
```
markdembo, line 1: Nothing fancy here. Copy & Paste from original.

static/lightbox.js
```diff
@@ -0,0 +1,163 @@
+function is_youtubelink(url) {
+  var p =
+    /^(?:https?:\/\/)?(?:www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((\w|-){11})(?:\S+)?$/;
+  return url.match(p) ? RegExp.$1 : false;
+}
+function is_imagelink(url) {
+  var p = /([a-z\-_0-9\/\:\.]*\.(jpg|jpeg|png|gif|svg|avif|webp))/i;
```
markdembo, line 2: Modified original to included `svg`, `avif` and `webp`

layouts/partials/footer.html
```diff
@@ -67,4 +67,6 @@
       {{- end -}}
     </div>
   </div>
+  <script type="text/javascript" src="/lightbox.js"></script>
```
kodster28:
```suggestion
  <script type="text/javascript" src="/lightbox.js" defer></script>
```

static/lightbox.js
```diff
@@ -0,0 +1,163 @@
+function is_youtubelink(url) {
```
kodster28: I don't _think_ this'll work with our current YT videos b/c they're styled as iframes + use the YT `nocookie` url.

Maybe worth updating this code to account for the nocookie URL + this [shortcode template](https://github.com/cloudflare/cloudflare-docs/blob/production/layouts/shortcodes/youtube.html) to make them render as links (or adjusting the lightbox.js)?

Example visible at https://developers.cloudflare.com/workers-ai/tutorials/how-to-choose-the-right-text-generation-model/

markdembo:
Removed videos from scope

kodster28 (approved):
Question for the future.... is there a reason we _wouldn't_ want this to be the default for all images?

PR #13317, opened 2024-03-07
[QUEUES] Queues use JSON content type by default

jbwcloudflare:
We recently reverted these changes in https://github.com/cloudflare/cloudflare-docs/pull/13269 as they went live too early.

This PR re-adds the changes, with corrected dates about when the feature goes live.

CC: @elithrar

### DIFF
--- a/content/queues/reference/how-queues-works.md
+++ b/content/queues/reference/how-queues-works.md
@@ -58,6 +58,49 @@ A queue can have multiple producer Workers. For example, you may have multiple p
 
 Additionally, multiple queues can be bound to a single Worker. That single Worker can decide which queue to write to (or write to multiple) based on any logic you define in your code.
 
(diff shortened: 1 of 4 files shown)


elithrar (approved):
Thank you @jbwcloudflare

jbwcloudflare:
@elithrar should I hold off on merging until `2024-03-18`?

Some of the new docs in this PR imply that the default is already live, but it isn't really the "default" until `2024-03-18`. (The compat flag is live and usable now -- it just doesn't get flipped on by default until that date)

elithrar:
@jbwcloudflare I will merge and then document the flag more clearly.

elithrar (approved):

Maddy-Cloudflare (approved):

PR #13326, opened 2024-03-08
Update settings.md

DRayCloudflare:
Expanded upon how to write expressions and added an example section that can hopefully be expanded in the future as various customer requirements come up.

### DIFF
--- a/content/rules/url-forwarding/single-redirects/_index.md
+++ b/content/rules/url-forwarding/single-redirects/_index.md
@@ -8,7 +8,7 @@ meta:
 
 # Single Redirects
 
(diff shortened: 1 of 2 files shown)


pedrosousa:
Hi @DRayCloudflare 

I updated your proposed text.

I removed your example at the bottom of the page because we have a separate page for [example rules](https://developers.cloudflare.com/rules/url-forwarding/single-redirects/examples/). The use case of your example seems to be [covered already](https://developers.cloudflare.com/rules/url-forwarding/single-redirects/examples/#redirect-all-requests-to-a-different-hostname).

pedrosousa (approved):

PR #13337, opened 2024-03-08
Cloudflare for Saas - List more Limitations

Techsurfer1981:
Page rules and redirect rules (on the client zone) also won't work on custom hostnames.


RebeccaTamachiro:
Hey! Thanks for raising this. Was checking with PM and seems we should actually refer to https://developers.cloudflare.com/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/product-compatibility/ now.
Page Rules was already there. 
I've created PCX-10519 to follow up on further updates as needed and will remove 'Limitations' from that page.

PR #13344, opened 2024-03-10
WIP: Runtime APIs - modules section

irvinebroque:
Scaffolds out a new /runtime-apis/modules section to explain our ESM support and CommonJS support.

refs https://github.com/cloudflare/workerd/issues/854#issuecomment-1987362815

- [ ] Needs to cross-link
- [ ] Needs to explain other module types
- [ ] Needs some code examples to be clear about what will/won't work
- [ ] Would benefit from clearer way to visualize role of Wrangler in process for those not thinking directly about what bundler is doing vs. what runtime is providing.
- [ ] Should probably discourage CommonJS more strongly.

### DIFF
--- a/content/workers/runtime-apis/modules.md
+++ b/content/workers/runtime-apis/modules.md
@@ -1,0 +1,30 @@
+---
+title: Modules
+pcx_content_type: configuration
+meta:
+  title: Modules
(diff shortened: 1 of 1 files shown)


KimJ15:
👋Hi there, we’re going to close this PR due to inactivity and web framework migration

PR #13361, opened 2024-03-11
[Workers AI] Adds a temporary meet us page for NVidia GTC

craigsdennis:
It's been requested to add a temporary page that we can use to link to via QR codes.


craigsdennis:
Going to not do this in docs as it is marketing related. Insert gif of homer ducking into bushes.

PR #13364, opened 2024-03-11
[Workers AI] Model page refresh from config API

craigsdennis:
Going to start doing these weekly as we update models in the config API. 

This one is smaller, but want to get the LlamaGuard update in place.


deadlypants1973 (approved):

craigsdennis:
Got pulled in a newer one.

PR #13375, opened 2024-03-12
[Docs Site] Cleanup Plan shortcode

KianNH:
Changes `workers_all` and `workers_paid` to kebab-case which is in line with `add-on` and `ent-add-on` which predated them.

Added some indentation and split up the dict onto multiple lines to make it easier to read and add new additions to.

cc @kodster28


kodster28 (approved):
LGTM, thanks for cleaning up!

PR #13409, opened 2024-03-13
Update _caa-records-added-by-cf.md

mcorreiacf:
Added information about Cloudflare not creating CAA records if the customer creates their own


mcorreiacf:
New information just came along.
Please disregard this edit.

PR #13431, opened 2024-03-14
Add `Deprecation` page for Roughtime

thibmeu:
Make a clear list of deprecated server, their public key, and refers to the tutorial to migrate.

### DIFF
--- a/content/time-services/roughtime/deprecation.md
+++ b/content/time-services/roughtime/deprecation.md
@@ -1,0 +1,15 @@
+---
+title: Server Deprecation
+pcx_content_type: reference
+weight: 4
+---
+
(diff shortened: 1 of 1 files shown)


thibmeu:
update link to match the generated link, instead of `./<file>.md`, which does not seem to be a valid syntax for the documentation

lukevalenta (approved):
Thanks for adding this!

kodster28 (approved):

kodster28:
> update link to match the generated link, instead of `./<file>.md`, which does not seem to be a valid syntax for the documentation

Yeah, keeping our generated links as the standard really helps with linting for broken ones + any migration we might do in the future.

PR #13443, opened 2024-03-14
[Pages] Remove bindings dev mode aside from Svelte guide

### LINKED ISSUE #13441: Remove bindings dev mode aside from Svelte guide

https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-site/#functions-setup

Remove the aside that says bindings are not supported in dev mode when using Svelte. The `4.2.0` version of the `sveltejs/adapter-cloudflare` has bindings support out of the box.

### Additional information

_No response_

jculvey:
This aside states that bindings is dev mode `is being worked on and will be released soon.`. This feature dropped in `sveltejs/adapter-cloudflare@4.2.0` and is included by default with new Svelte projects created with C3.

### DIFF
--- a/content/pages/framework-guides/deploy-a-svelte-site.md
+++ b/content/pages/framework-guides/deploy-a-svelte-site.md
@@ -103,7 +103,7 @@ In addition to the Cloudflare adapter, review other adapters you can use in your
 
 - [`@sveltejs/adapter-auto`](https://www.npmjs.com/package/@sveltejs/adapter-auto)
 
-  SvelteKit's default adapter  automatically chooses the adapter for your current environment. If you use this adapter, [no configuration is needed](https://kit.svelte.dev/docs/adapter-auto). However, the default adapter introduces a few disadvantages for local development because it has no way of knowing what platform the application is going to be deployed to.
+  SvelteKit's default adapter automatically chooses the adapter for your current environment. If you use this adapter, [no configuration is needed](https://kit.svelte.dev/docs/adapter-auto). However, the default adapter introduces a few disadvantages for local development because it has no way of knowing what platform the application is going to be deployed to.
 
 To solve this issue, provide a `CF_PAGES` variable to SvelteKit so that the adapter can detect the Pages platform. For example, when locally building the application: `CF_PAGES=1 vite build`.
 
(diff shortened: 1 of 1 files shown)


content/pages/framework-guides/deploy-a-svelte-site.md
```diff
@@ -103,7 +103,7 @@ In addition to the Cloudflare adapter, review other adapters you can use in your
 
 - [`@sveltejs/adapter-auto`](https://www.npmjs.com/package/@sveltejs/adapter-auto)
 
-  SvelteKit's default adapter  automatically chooses the adapter for your current environment. If you use this adapter, [no configuration is needed](https://kit.svelte.dev/docs/adapter-auto). However, the default adapter introduces a few disadvantages for local development because it has no way of knowing what platform the application is going to be deployed to.
+  SvelteKit's default adapter automatically chooses the adapter for your current environment. If you use this adapter, [no configuration is needed](https://kit.svelte.dev/docs/adapter-auto). However, the default adapter introduces a few disadvantages for local development because it has no way of knowing what platform the application is going to be deployed to.
```
dario-piotrowicz, line 106: I feel like this whole section at this point can be removed, no?

What are the few disadvantages as this point? not running the worker code and not running it in workerd? that's basically the same for all frameworks anyways right?

dario-piotrowicz (approved):

content/pages/framework-guides/deploy-a-svelte-site.md
WalshyDev: imo this is fine since we say this is an issue and we point to an instruction below

is this no longer true? It seems like just the auto and no CF_PAGES doesn't get detected as Pages. Does this mean it misses out on specific Worker features? If so, we should keep. If not, is it the same? If so, we can remove.

dario-piotrowicz: oh yeah that's a good point! I did not check the text below 🙈 

mh.... yeah I think this whole section needs a bit of a rewrite as we for example tell the user to run `CF_PAGES=1 vite build` without telling them what to do with the provided output (they can preview it with `wrangler (pages) dev` or deploy it)

Not something that should block this PR 🙂👍

WalshyDev (approved):

deadlypants1973 (approved):

PR #13456, opened 2024-03-14
Add rollback notice; format as callout

microcosem:
During our initial buildout of a Cloudflare deployment, our team noticed that we could not see the "Rollback" option in the drop-down menu when the deployment only had one version (one pushed/merged commit on the branch). Thought it might be worthy to add here, and also to format it and the other note in a callout.


kodster28:
Howdy @microcosem, going to hold off on this change for now as most of that information is covered in the paragraphs above / below.

If we hear of other issues in the future around this though, 1000% will make sure to clarify this section.

PR #13458, opened 2024-03-15
Added that Aegis is not supported with health checks

DRayCloudflare:


DRayCloudflare:
Turns out Health checks can work with Aegis with Simulated Zone.

PR #13481, opened 2024-03-16
fact check health-checks-regions.md

12932:
I've checked on the Cloudflare Dashboard, and *All Regions* is for both Business and Enterprise customers:

image


kodster28 (approved):

kodster28:
Thanks for the correction, @12932!

PR #13492, opened 2024-03-18
Update _aop-configure-origin.md

jessuppi:
Improved the description of Apache and Nginx certificate paths when using Authenticated Origin Pulls, so that users are aware they must download, rename, and then upload the .PEM file provided by Cloudflare.

### DIFF
--- a/content/ssl/_partials/_aop-configure-origin.md
+++ b/content/ssl/_partials/_aop-configure-origin.md
@@ -12,27 +12,27 @@ Check the examples below for Apache and NGINX or refer to your origin web server
 
 {{<details header="Apache example">}}
 
-For this example, you would have saved the certificate to `/path/to/origin-pull-ca.pem`.
-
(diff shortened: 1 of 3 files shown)


icodeforlove:
Commit message, and pull request name should be more descriptive.

RebeccaTamachiro:
👋 @jessuppi Thanks for raising this!
Since _aop-configure-origin.md is a partial used in different docs, I had to edit it a bit so that the more specific instructions would only appear for the zone-level setup.

jessuppi:
Great stuff @RebeccaTamachiro 

I figured it might need some tweaking due to partial files. Anyway, thanks for your efforts!

Ref: https://developers.cloudflare.com/ssl/origin-configuration/authenticated-origin-pull/set-up/zone-level/

RebeccaTamachiro (approved):

PR #13496, opened 2024-03-18
Update limitations.md

XabierCloudflare:
Customer complains about changes in CA for Universal Cert. Please see ZD ticket: #3189555

It is not clear that we do not notify, and we can change this without any notification to the zone owner.


RebeccaTamachiro (approved):

RebeccaTamachiro:
Thanks for the PR, @XabierCloudflare!

PR #13499, opened 2024-03-18
Update _index.md

jochenhonig:
Clarifying validity period of unused override codes.


kennyj42:
@kkrum want to confirm this one? Seems valid to me but I want to hold off on approving until you say yes

ranbel:
Thanks for bringing this documentation gap to our attention! We are clarifying the timeout behavior in #13885.

PR #13507, opened 2024-03-18
[Pages] Add section for disabling zone caching for custom domains

karimkaylani:
Hi there!

I recently migrated my project to Cloudflare Pages and was struggling with stale data appearing on my custom domain and couldn't figure out why. However, I discovered that it was likely due to external caching from the custom domain itself from looking at community posts and discord messages. I noticed that this was a somewhat common issue ([1](https://www.answeroverflow.com/m/1158677998329938021)) ([2](https://stackoverflow.com/questions/76874341/cloudflare-page-with-custom-domain-is-cached-after-deploy)) ([3](https://community.cloudflare.com/t/how-can-i-disable-the-cache-from-a-cloudflare-pages/605730)) and figured it might be worth including somewhere in the documentation.

I apologize if this isn't the proper way to bring this suggestion up, I'm relatively new to the open-source community. Thanks!


kodster28:
Howdy @karimkaylani, thanks for the submission, but closing this out in favor of the updated advice in #14319.

PR #13515, opened 2024-03-19
[Durable Objects] Update use-kv-from-durable-objects.md

mrbbot:
Saw this in Discord. Looks like this example uses slightly incorrect types.


Maddy-Cloudflare (approved):

elithrar (approved):
Thanks Brendan!

PR #13521, opened 2024-03-19
[Images]Update _index.md

luishcardoso:
We should convey to customers Image Resizing is not compatible with BYOIP. PCX-10199


kodster28:
This is more clearly conveyed in #13501, it doesn't make sense alongside a list of positive features about image resizing.

PR #13528, opened 2024-03-19
[WAF] Update OWASP page

pedrosousa:
Addresses PCX-10223.


content/waf/managed-rules/reference/owasp-core-ruleset.md
```diff
@@ -10,17 +10,72 @@ The Cloudflare OWASP Core Ruleset is Cloudflare's implementation of the [OWASP M
...
+
+Each OWASP rule that matches the current request has an associated score. The request threat score is the sum of the individual scores of all OWASP rules that matched the request.
+
+### Score threshold
+
+The score threshold (or anomaly threshold) defines the minimum cumulative score — obtained from matching OWASP rules — for the WAF to apply the configured OWASP ruleset action.
+
+Each threshold (_Low_, _Medium_, and _High_) has an associated value (_60_, _40_, and _25_, respectively). Configuring a _Low_ threshold means that more rules will have to match the current request for the WAF to apply the configured ruleset action.
+
+### Paranoia level
```
RebeccaTamachiro, line 15: Wonder if would be even clearer if this was the first concept, kind of reflecting the workflow below, where setting the PL at ruleset level is the first step. So, going from the more generic concept to the more specific, order would be: OWASP PLs, threshold, and then understanding what happen with each request.

RebeccaTamachiro (approved):

content/waf/managed-rules/reference/owasp-core-ruleset.md
pedrosousa: Makes sense! I'll reorder the concepts.

PR #13535, opened 2024-03-20
Magic-WAN-remove-command-line-tunnel-monitor-config-Palo-Alto

alpdot:
SPM-2285. Removing tunnel monitor from the example configuration, as one of the prerequisites is to disable tunnel monitor.


kodster28:
Closing as we couldn't get technical support answering whether this was the correct approach.

PR #13554, opened 2024-03-20
Add gradual deployments documentation the `Cf-Worker-Version-Key` header.

jp4a50:


content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
 ```
 You should see 10 responses. Responses will reflect the content returned by the versions in your deployment. Responses will vary depending on the percentages configured in step #6. 
 
+## Version keys
```
irvinebroque: I think content should live here but we should add link to here from https://developers.cloudflare.com/workers/runtime-apis/headers/ so that if someone sees this, they look at headers docs in runtime APIs and can find it

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
...
+
+You may want certain requests to always be handled by the same version of your Worker. You can do this by setting the `Cf-Worker-Version-Key` header on a request. For example:
```
irvinebroque:
```suggestion
You may want requests associated with a particular identifier to be handled by a consistent version of your Worker, such that when there are two versions of your Worker deployed, the version used does not change back and forth on a per-request basis for a given user, session, or any unique ID.

You can do this by setting the `Cf-Worker-Version-Key` header on a request. For example:
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
...
+
+```sh
+curl -s https://$SCRIPT_NAME.$SUBDOMAIN.workers.dev -H 'Cf-Worker-Version-Key: foo'
+```
+
+For a given deployment, all requests with a version key set to `foo` will be handled by the same version of your Worker. The specific Worker version that the version key `foo` corresponds to is determined by the percentages you configured for each Worker version in your deployment.
```
irvinebroque:
```suggestion
For a given [deployment](/workers/configuration/deployments/), all requests with a version key set to `foo` will be handled by the same version of your Worker. The specific version of your Worker that the version key `foo` corresponds to is determined by the percentages you have configured for each Worker version in your deployment.

You can set the `Cf-Worker-Version-Key` header both when making an external request from the Internet to your Worker, as well as when making a subrequest from one Worker to another Worker using a [service binding](/workers/runtime-apis/bindings/service-bindings/).
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
irvinebroque:
```suggestion
## Version keys and session affinity
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
...
+
```
irvinebroque, line 129:
```suggestion
### Extracting version keys from URLs

If the unique identifier that you want to provide to `Cf-Worker-Version-Key` is in the URL of the incoming request, you can configure a [Ruleset Engine](/ruleset-engine/) rule on your zone to extract this key from the path or querystring, and use its value to set `Cf-Worker-Verison-Key`. This allows you to ensure version affinity, while avoiding having to modify the external client that makes the request.
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
...
+You can also:
```
irvinebroque:
```suggestion
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
...
+- use [Ruleset Engine](/ruleset-engine/) to define an [HTTP request header modification rule](/rules/transform/request-header-modification/) that sets the `Cf-Worker-Version-Key` header on the incoming request without having to modify the external client making the request.
```
irvinebroque:
```suggestion
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,20 @@ done
...
+- set the `Cf-Worker-Version-Key` header when making a request from one Worker to another Worker using a [service binding](/workers/runtime-apis/bindings/service-bindings/).
```
irvinebroque:
```suggestion
```

moved above

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,23 @@ done
 ```
 You should see 10 responses. Responses will reflect the content returned by the versions in your deployment. Responses will vary depending on the percentages configured in step #6. 
 
+## Version keys and session affinity
+
+You may want requests associated with a particular identifier to be handled by a consistent version of your Worker, such that when there are two versions of your Worker deployed, the version used does not change back and forth on a per-request basis for a given user, session, or any unique ID.
+
+You can do this by setting the `Cf-Worker-Version-Key` header on a request. For example:
```
irvinebroque:
```suggestion
You can do this by setting the `Cf-Worker-Version-Key` header on the incoming request to your Worker. For example:
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,23 @@ done
...
+
+```sh
+curl -s https://$SCRIPT_NAME.$SUBDOMAIN.workers.dev -H 'Cf-Worker-Version-Key: foo'
+```
+
+For a given [deployment](/workers/configuration/deployments/), all requests with a version key set to `foo` will be handled by the same version of your Worker. The specific version of your Worker that the version key `foo` corresponds to is determined by the percentages you have configured for each Worker version in your deployment.
+
+You can set the `Cf-Worker-Version-Key` header both when making an external request from the Internet to your Worker, as well as when making a subrequest from one Worker to another Worker using a [service binding](/workers/runtime-apis/bindings/service-bindings/).
```
irvinebroque: Fun challenge to think about for JS RPC

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,23 @@ done
...
+### Extracting version keys from URLs
+
+If the unique identifier that you want to provide to `Cf-Worker-Version-Key` is in the URL of the incoming request, you can configure a [Ruleset Engine](/ruleset-engine/) rule on your zone to extract this key from the path or querystring, and use its value to set `Cf-Worker-Verison-Key`. This allows you to ensure version affinity, while avoiding having to modify the external client that makes the request.
```
irvinebroque: Maybe example rule here that does basic "grab ID from URL and pop into header"?

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: Yep. I've been thinking the same thing about version ID override functionality too.

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,23 @@ done
...
+You can do this by setting the `Cf-Worker-Version-Key` header on the incoming request to your Worker. For example:
+
+```sh
+curl -s https://$SCRIPT_NAME.$SUBDOMAIN.workers.dev -H 'Cf-Worker-Version-Key: foo'
+```
+
+For a given [deployment](/workers/configuration/deployments/), all requests with a version key set to `foo` will be handled by the same version of your Worker. The specific version of your Worker that the version key `foo` corresponds to is determined by the percentages you have configured for each Worker version in your deployment.
+
+You can set the `Cf-Worker-Version-Key` header both when making an external request from the Internet to your Worker, as well as when making a subrequest from one Worker to another Worker using a [service binding](/workers/runtime-apis/bindings/service-bindings/).
+### Extracting version keys from URLs
```
jp4a50: I think it's probably most likely that users will want to derive a version key from cookies and/or other headers than the URL.

Maybe we can rename the header to "Creating version keys using Ruleset Engine"?

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: Sure.

I think the most likely use case will be setting the header equal to a cookie's value so probably would make most sense to show that.

I'll copy the format used in [request header modification examples](https://developers.cloudflare.com/rules/transform/request-header-modification/examples/#add-an-http-request-header-with-the-current-bot-score).

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,23 @@ done
 ```
 You should see 10 responses. Responses will reflect the content returned by the versions in your deployment. Responses will vary depending on the percentages configured in step #6. 
 
+## Version keys and session affinity
```
jp4a50, line 94: FWIW I didn't mention the word "affinity" in case we didn't want to introduce an extra term in addition to the concept of "version key". I'm OK with committing to using the term "affinity", but maybe we should just call this whole section "version affinity" (which can be achieved by setting a "version key") to make it a little more concise?

content/workers/configuration/versions-and-deployments/gradual-deployments.md
```diff
@@ -91,6 +91,23 @@ done
...
+
+You may want requests associated with a particular identifier to be handled by a consistent version of your Worker, such that when there are two versions of your Worker deployed, the version used does not change back and forth on a per-request basis for a given user, session, or any unique ID.
```
tanushree-sharma:
```suggestion
By default, the percentages configured when using gradual deployments operate on a per-request basis — a request has a X% probability of invoking one of two versions of the Worker in the [deployment](/workers/configuration/versions-and-deployments/#deployments). 

You may want requests associated with a particular identifier (such as user, session, or any unique ID) to be handled by a consistent version of your Worker such that when there are two versions of your Worker deployed, the version used does not change back and forth on a per-request basis. 
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
tanushree-sharma: "Setting the Cf-Worker-Version-Key header using Ruleset Engine" ?

content/workers/configuration/versions-and-deployments/gradual-deployments.md
tanushree-sharma: Agreed on calling it "Version affinity"

irvinebroque:
```suggestion
## Version affinity
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
irvinebroque:
```suggestion
### Setting `Cf-Worker-Version-Key` using Ruleset Engine
```

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: Uh oh. It looks like this won't work right now.

I tested the ruleset engine header transformation stuff before I landed on a name for the header. It turns out ruleset engine won't let you set any headers beginning with `Cf-` :(

I'll have to remove the rulset engine section for now...

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: Yeah we probably should add something there.

We don't appear to actually have any prior art in terms of headers that can be set by users that are interpreted by the runtime, so I'll probably have to create a new section.

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: Made the change locally.

jp4a50:
We're postponing documentation for the version key header.

I've opened a dedicated PR for gradual rollouts docs here: https://github.com/cloudflare/cloudflare-docs/pull/13564

irvinebroque:
Reopening so that we don't forget to do this with rename

content/workers/configuration/versions-and-deployments/gradual-deployments.md
irvinebroque: we should say explicitly how this will work with durable objects, and then link out to the durable objects docs on gradual deployments

jp4a50:
I'll pick it back up now.

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: I added an example that extracts the header from the URL as you originally suggested since rules engine limitations make it more complicated than it should be to extract it from a cookie (I'm pushing them to make it easier).

content/workers/configuration/versions-and-deployments/gradual-deployments.md
jp4a50: Having thought about it, since the section directly linked is runtime-specific (and it's arguably an implementation detail that this header is handled by the runtime) and the page it links to specifically discusses headers added by Cloudflare, I don't think there's an obviously place to slot this. Unless you feel strongly, I'll stick with just the docs here for now.

PR #13563, opened 2024-03-21
[Pages] Reorder Nuxt framework guide

jculvey:
Reorders the Nuxt framework guide to mention bindings before deployment so that the guide more closely resembles the user's workflow. 

There's also a few sections that were removed because they were redundant with other material in the guide or because these are now handled by C3 and don't require setup work by the user.

### DIFF
--- a/content/pages/framework-guides/deploy-a-nuxt-site.md
+++ b/content/pages/framework-guides/deploy-a-nuxt-site.md
@@ -11,33 +11,106 @@ meta:
 
 In this guide, you will create a new Nuxt application and deploy it using Cloudflare Pages.
 
-## Create a new project using the `create-cloudflare` CLI (C3)
+## Create a new project with `create-cloudflare`
 
-The [`create-cloudflare` CLI (C3)](/pages/get-started/c3/) will configure your Nuxt site for Cloudflare Pages. Run the following command in your terminal to create a new Nuxt site:
+The easiest way to create a new Nuxt project and deploy to Cloudflare Pages is to use the [`create-cloudflare`](https://www.npmjs.com/package/create-cloudflare) CLI (also known as C3). To get started, open a terminal and run:
+
+{{<tabs labels="npm | pnpm | bun">}}
+{{<tab label="npm" default="true">}}
 
 ```sh
 $ npm create cloudflare@latest my-nuxt-app -- --framework=nuxt
 ```
 
-C3 will ask you a series of setup questions and create a new project with [`nuxi` (the official Nuxt CLI)](https://github.com/nuxt/cli). C3 will also install the necessary adapters along with the [Wrangler CLI](/workers/wrangler/install-and-update/#check-your-wrangler-version).
+{{</tab>}}
+{{<tab label="pnpm">}}
 
-After creating your project, C3 will generate a new `my-nuxt-app` directory using the default Nuxt template, updated to be fully compatible with Cloudflare Pages.
+```sh
+$ pnpm create cloudflare@latest my-nuxt-app --framework=nuxt
+```
+
+{{</tab>}}
+{{<tab label="bun">}}
+
+```sh
+$ bun create cloudflare@latest my-nuxt-app --framework=nuxt
+```
+
+{{</tab>}}
+{{</tabs>}}
+
+C3 will walk you through the setup process and create a new project using `nuxi`, the official Nuxt cli. It will also install the necessary adapters along with the [Wrangler CLI](/workers/wrangler/install-and-update/#check-your-wrangler-version).
+
+{{<Aside type="note" header="Deployment">}}
+
+The final step of the C3 workflow will offer to deploy your application to Cloudflare. For more information on deployment options, see the [Deployment](#deployment) section below.
+
+{{</Aside>}}
+
+## Bindings
+
+{{<render file="/_framework-guides/_bindings_definition.md">}}
+
+In Nuxt, add server-side code via [Server Routes and Middleware](https://nuxt.com/docs/guide/directory-structure/server#server-directory). The `defineEventHandler()` method is used to define your API endpoints in which you can access Cloudflare's context via the provided `context` field. The `context` field allows you to access any bindings set for your application.
+
+The following code block shows an example of accessing a KV namespace in Nuxt.
+
+```typescript
+---
+filename: server/api/hello.ts
+highlight: [2]
+---
+export default defineEventHandler(async ({ context }) => {
+  const { MY_KV } = context.cloudflare.env;
+  const greeting = (await MY_KV.get("greeting")) ?? "hello";
+
+  return {
+    greeting,
+  };
+});
+```
+
+### Setup bindings in development
+
+Projects created via C3 come with `nitro-cloudflare-dev`, a `nitro` module that simplifies the process of working with bindings during development:
+
+```typescript
+---
+filename: nuxt.config.ts
+highlights: [2]
+---
+export default defineNuxtConfig({
+  modules: ["nitro-cloudflare-dev"],
+});
+```
+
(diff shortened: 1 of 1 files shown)


content/pages/framework-guides/deploy-a-nuxt-site.md
```diff
@@ -11,76 +11,66 @@ meta:
...
+  return {
+    greeting,
+  };
+});
+```
 
-### Set up bindings for local development
+### Setup bindings in development
 
 Projects created via C3 come with `nitro-cloudflare-dev`, a `nitro` module that simplifies the process of working with bindings during development:
```
IgorMinar, line 75: When I read this bit, I was wondering why I needed to know about these implementation details of this feature, at least it sounds like they are implementation details.

How about you keep the paragraph above but instead of this following code snippet, you add a snippet with an example wrangler.toml with a D1 binding? Wouldn't that be more relevant for the user?

IgorMinar: There is a separate use case that is not addressed in this guide - what do I do if I already have a nuxt application and want to deploy it to Cloudflare. Could this snippet and some additional instructions be moved to the end of this guide into a new section called something like "Moving and existing application to Cloudflare". What do you think?

IgorMinar (commented):
I think is much better. I added two more comments with suggestions.

content/pages/framework-guides/deploy-a-nuxt-site.md
jculvey: > When I read this bit, I was wondering why I needed to know about these implementation details of this feature, at least it sounds like they are implementation details.
> 
> How about you keep the paragraph above but instead of this following code snippet, you add a snippet with an example wrangler.toml with a D1 binding? Wouldn't that be more relevant for the user?

Good point, if you're using C3 the nitro module should be a black box you don't really need to know about, and having a usage example of wrangler would probably make more sense. I'll switch it out.

> There is a separate use case that is not addressed in this guide - what do I do if I already have a nuxt application and want to deploy it to Cloudflare. Could this snippet and some additional instructions be moved to the end of this guide into a new section called something like "Moving and existing application to Cloudflare". What do you think?

This is also a good point that Dario and I chatted about. I'd love to hear your thoughts on this. 

Initially I did include a short `Migration Guide` at the bottom but deleted it. I do think it makes sense to have a migration guide for each framework, but if we're going to do so I think we should put some time into ensuring its completeness. I worry that partial instructions are worse than no instructions, since users might follow them, get stuck on something we haven't covered, then get the impression that their framework just doesn't work on Cloudflare and cross it off the list.

In short, a good migration guide would need instructions on how to manually perform many of the steps C3 automates. For Nuxt, it would need to include: 
 - Installing the `nitro-module-cloudflare`
 - Modifying the nuxt config to use the Cloudflare preset
 - Creating an ambient type declaration for H3EventContext` that includes the Cloudflare `Env` definition.
 - Creating a placeholder `wrangler.toml` (which unfortunately wouldn't be full of examples like the one c3 gives you)
 - Making a build task to regenerate types (or informing the user on how to use `wrangler types`)
 - Installing wrangler types and configuring the latest entrypoint in tsconfig
 - Adding scripts for `dev` and `deploy`

A few of these would be somewhat common across frameworks, but there would be some framework-specific instructions in each guide.

IgorMinar: Would it be possible to "replay" the c3 automation on an existing project?

What I mean, would it be possible for me to cd into an existing nuxt project and somehow invoke c3 in a mode where it doesn't call create-nuxt, and instead it just runs all the cloudflare specific stuff?

I know that projects might vary quite a bit, but I think that we could actually have a pretty high success rate in many/most scenarios.

What do you think?

IgorMinar: One headache will be version skew. Our automated migrations might not work well for older nuxt projects. I'm not sure what to do about that.

jculvey: > Would it be possible to "replay" the c3 automation on an existing project?
> 
> What I mean, would it be possible for me to cd into an existing nuxt project and somehow invoke c3 in a mode where it doesn't call create-nuxt, and instead it just runs all the cloudflare specific stuff?
> 
> I know that projects might vary quite a bit, but I think that we could actually have a pretty high success rate in many/most scenarios.
> 
> What do you think?

I like the idea of it. C3 has the benefit of assuming we're starting from a clean slate, so I'm sure there's a few things we'll have to thing through to pull it off.

The big issues I see are:
1. What you mentioned, version skew. We could probably devise a way of storing a minimum supported version per framework, and just bail if that version is not met. I think trying to support older versions though would be a nightmare.
2. Dealing with existing file conflicts. We're able to leverage that assumption about a clean slate and write files today in places where we would need to strictly codemod. This shouldn't be too bad, but we'd have to be more defensive than we currently are about files we think might exist in a project.

content/pages/framework-guides/deploy-a-nuxt-site.md
```diff
@@ -96,85 +86,65 @@ export default defineNuxtConfig({
...
 ```
 
-### Access bindings in your Nuxt application
+{{<render file="/_framework-guides/_git-integration.md">}}
 
-In Nuxt, add server-side code via [Server Routes and Middleware](https://nuxt.com/docs/guide/directory-structure/server#server-directory). The `defineEventHandler()` method is used to define your API endpoints in which you can access Cloudflare's context via the provided `context` field. The `context` field allows you to access any bindings set for your application.
+### Create a GitHub repository
 
-The following code block shows an example of accessing a KV namespace in Nuxt.
+{{<render file="/_framework-guides/_create-gh-repo.md">}}
```
kodster28, line 119:
```suggestion
{{<render file="/_framework-guides/_create-gh-repo.md">}}
<br/>
```

kodster28: will resolve the weird text treatment that sometimes affects rendered files

content/pages/framework-guides/deploy-a-nuxt-site.md
```diff
@@ -96,85 +86,65 @@ export default defineNuxtConfig({
...
-    // ...
-  };
-});
-```
+1. Log in to the [Cloudflare dashboard](https://dash.cloudflare.com/) and select your account.
+2. Go to **Workers & Pages** > **Create application** > **Pages** > **Connect to Git** and create a new Pages project.
 
-{{</tab>}}
-{{<tab label="ts">}}
+You will be asked to authorize access to your GitHub account if you have not already done so. Cloudflare needs this so that it can monitor and deploy your projects from the source. You may narrow access to specific repositories if you prefer; however, you will have to manually update this list [within your GitHub settings](https://github.com/settings/installations) when you want to add more repositories to Cloudflare Pages.
```
kodster28, line 138: should indent so it's w/in the list.
(applies below as well)

kodster28 (approved):

content/pages/framework-guides/deploy-a-nuxt-site.md
```diff
@@ -96,85 +86,65 @@ export default defineNuxtConfig({
...
+This should be done any time you add new bindings to `wrangler.toml`.
+
 {{<Aside type="note">}}
 
 `wrangler.toml` is currently **only** used for local development. Bindings specified in it are not available remotely.
 
 {{</Aside>}}
 
-### Set up bindings for a deployed application
+### Setup bindings in deployed applications
```
kodster28, line 103:
```suggestion
### Deployed applications
```

content/pages/framework-guides/deploy-a-nuxt-site.md
```diff
@@ -11,76 +11,66 @@ meta:
...
 
-{{<render file="/_framework-guides/_bindings_definition.md">}}
+  return {
+    greeting,
+  };
+});
+```
 
-### Set up bindings for local development
+### Setup bindings in development
```
kodster28, line 73:
```suggestion
### Development
```

KimJ15:

PR #13567, opened 2024-03-21
[Workers] Update wrangler commands to include new Hyperdrive private-host option

mtlemilio:
Based on changes in https://github.com/cloudflare/workers-sdk/pull/4932


mtlemilio:
Closing for now, as we're reverting the wrangler change, pending more work internally.

PR #13581, opened 2024-03-22
[Workers] Remove Cache API per-request storage limit

fhanau:
This has been removed in https://github.com/cloudflare/workerd/pull/1863, which has been deployed to production since.
Also updates the table formatting to list the unbound and standard models together – currently the Standard column can't be seen without scrolling to the right which is confusing (https://developers.cloudflare.com/workers/platform/limits/#cache-api-limits), putting them together also matches the approach used for the table above.

### DIFF
--- a/content/workers/platform/limits.md
+++ b/content/workers/platform/limits.md
@@ -94,7 +94,7 @@ On the Unbound billing model, scheduled Workers ([Cron Triggers](/workers/config
 
 | Feature                       | Workers Free  | [Bundled](/workers/platform/pricing/#example-pricing-bundled-usage-model) | [Unbound](/workers/platform/pricing/#example-pricing-unbound-usage-model) and [Standard](/workers/platform/pricing/#example-pricing-standard-usage-model) |
 | ----------------------------- | ------------- | ------- | ------- |
-| [Max object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  |
+| [Maximum object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  |
 | [Calls/request](#cache-api-limits)   | 50     | 50      | 1,000   |
 
 {{</table-wrap>}}
--- a/content/workers/platform/limits.md
+++ b/content/workers/platform/limits.md
@@ -92,17 +92,14 @@ On the Unbound billing model, scheduled Workers ([Cron Triggers](/workers/config
 
 {{<table-wrap>}}
 
-| Feature                       | Workers Free  | [Bundled](/workers/platform/pricing/#example-pricing-bundled-usage-model) | [Unbound](/workers/platform/pricing/#example-pricing-unbound-usage-model) | [Standard](/workers/platform/pricing/#example-pricing-standard-usage-model)  |
-| ----------------------------- | ------------- | ------- | ------- | ------- |
-| [Max object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  | 512 MB  |
-| [Calls/request](#cache-api-limits)   | 50     | 50      | 1,000   | 1,000   |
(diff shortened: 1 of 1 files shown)


content/workers/platform/limits.md
```diff
@@ -92,18 +92,15 @@ On the Unbound billing model, scheduled Workers ([Cron Triggers](/workers/config
...
+| Feature                       | Workers Free  | [Bundled](/workers/platform/pricing/#example-pricing-bundled-usage-model) | [Unbound](/workers/platform/pricing/#example-pricing-unbound-usage-model) and [Standard](/workers/platform/pricing/#example-pricing-standard-usage-model) |
+| ----------------------------- | ------------- | ------- | ------- |
+| [Max object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  |
+| [Calls/request](#cache-api-limits)   | 50     | 50      | 1,000   |
 
 {{</table-wrap>}}
 
 - 50 total `put()`, `match()`, or `delete()` calls per-request, using the same quota as `fetch()`.
 
-- 5 GB total `put()` per request.
```
fhanau, line 105: @harrishancock Should we note that there used to be a per-request limit here?

irvinebroque:
Think these limits are still relevant to show on this page, since there is still a default limit, and the only change here is the mechanics of where and how the limit is defined (internal implementation detail) — is that accurate?

fhanau:
> Think these limits are still relevant to show on this page, since there is still a default limit, and the only change here is the mechanics of where and how the limit is defined (internal implementation detail) — is that accurate?

https://github.com/cloudflare/workerd/pull/1863 does introduce a material change – the per-Workers request limit has been removed entirely since we determined that the other limits are sufficient and there is no need to retain it as a configurable limit. The remaining Cache API limits continue to apply so this just removes one row in the table.

irvinebroque (approved):
@fhanau — I follow now, makes sense

content/workers/platform/limits.md
irvinebroque: Can something here for cache link over to https://developers.cloudflare.com/cache/concepts/default-cache-behavior/#customization-options-and-limits

content/workers/platform/limits.md
```diff
@@ -92,18 +92,15 @@ On the Unbound billing model, scheduled Workers ([Cron Triggers](/workers/config
...
 {{<table-wrap>}}
 
-| Feature                       | Workers Free  | [Bundled](/workers/platform/pricing/#example-pricing-bundled-usage-model) | [Unbound](/workers/platform/pricing/#example-pricing-unbound-usage-model) | [Standard](/workers/platform/pricing/#example-pricing-standard-usage-model)  |
-| ----------------------------- | ------------- | ------- | ------- | ------- |
-| [Max object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  | 512 MB  |
-| [Calls/request](#cache-api-limits)   | 50     | 50      | 1,000   | 1,000   |
-| [Storage/request](#cache-api-limits) | 5 GB   | 5 GB    | 5 GB    | 5 GB    |
+| Feature                       | Workers Free  | [Bundled](/workers/platform/pricing/#example-pricing-bundled-usage-model) | [Unbound](/workers/platform/pricing/#example-pricing-unbound-usage-model) and [Standard](/workers/platform/pricing/#example-pricing-standard-usage-model) |
+| ----------------------------- | ------------- | ------- | ------- |
+| [Max object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  |
```
deadlypants1973:
```suggestion
| [Maximum object size](#cache-api-limits) | 512 MB | 512 MB  | 512 MB  |
```

deadlypants1973 (approved):

PR #13586, opened 2024-03-22
[Pages] Clarify redirects documentation

edeca:
The documentation for redirects has a section for [Advanced redirects](https://developers.cloudflare.com/pages/configuration/redirects/#advanced-redirects). 

The example for [splats](https://developers.cloudflare.com/pages/configuration/redirects/#splats) shows a redirect from `/blog/*` to `/blog/:splat`.

This doesn't seem to make sense, as the `*` greedily matches all characters and the `:splat` adds them back to the partial URL. This results in a redirect of `/blog/foo` to `/blog/foo` - the same place.

I think it would be clearer if the table under advanced redirects showed an example which redirects to another directory, e.g. `/blog/foo` to `/posts/foo`. This is a common pattern where splats seem to be useful.


kodster28 (approved):
Thanks for the catch and fix, @edeca

PR #13599, opened 2024-03-25
NC-4351: Add affinity hints

worenga:


punkeel (changes requested):
mind holding a bit before merging? 🙏

PR #13604, opened 2024-03-25
[DLS and Analytics] Adds info about CMB error message

angelampcosta:


content/analytics/_partials/_customer_metadata_boundary_error.md
```diff
@@ -0,0 +1,12 @@
+---
+_build:
+  publishResources: false
+  render: never
+  list: never
+---
+
+If you encounter a message on the dashboard indicating that your data is unavailable due to your account's Metadata Boundary configuration, this is because you are trying to access data that is not stored in your region (that is, you are in the US and trying to access data that is only stored in the EU, or vice versa). If you receive this error message while being in the region where your data is stored, there are two potential reasons why you might see this message:
```
pedrosousa:
```suggestion
If you encounter a message on the dashboard indicating that your data is unavailable due to your account's Metadata Boundary configuration, this is because you are trying to access data that is not stored in your region (that is, you are in the US and trying to access data that is only stored in the EU, or vice versa). If you receive this error message while being in the region where your data is stored, there are two potential reasons why you might get this message:
```
Suggestion to avoid "see".

pedrosousa (approved):

PR #13607, opened 2024-03-25
Update unicode-interoperability.md to indicate macOS file names are case insensitive

kyldvs:
Noticed a minor issue while reading the docs, it lists macOS as a case-sensitive file system but by default it's not. You can see this in a terminal:

image


crwaters16 (approved):

crwaters16:
Addressed in https://github.com/cloudflare/cloudflare-docs/pull/14217

PR #13632, opened 2024-03-27
[Durable Objects] Update transactional-storage-api.md

elithrar:
Clarify `deleteAll()` vs `deleteAlarm()`.

cc @a-robinson @vy-ton

### DIFF
--- a/content/durable-objects/api/transactional-storage-api.md
+++ b/content/durable-objects/api/transactional-storage-api.md
@@ -93,7 +93,7 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
 - {{<code>}}deleteAll{{</code>}}(options{{<param-type>}}Object{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}) : {{<type>}}Promise{{</type>}}
 
-  - Deletes all keys and associated values, effectively deallocating all storage used by the Durable Object. In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted. `deleteAll()` does not proactively delete [Alarms](/durable-objects/api/alarms/). Call [`deleteAlarm()`](/durable-objects/api/alarms/#deletealarm) to delete an alarm.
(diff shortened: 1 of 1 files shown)


content/durable-objects/api/transactional-storage-api.md
```diff
@@ -93,7 +93,7 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
 - {{<code>}}deleteAll{{</code>}}(options{{<param-type>}}Object{{</param-type>}}{{<prop-meta>}}optional{{</prop-meta>}}) : {{<type>}}Promise{{</type>}}
 
-  - Deletes all keys and associated values, effectively deallocating all storage used by the Durable Object. In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted.
+  - Deletes all keys and associated values, effectively deallocating all storage used by the Durable Object. In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted. `deleteAll()` does not proactively delete [Alarms](/durable-objects/api/alarms/). Call [`deleteAlarm()`](/durable-objects/api/alarms/#deletealarm) to delete an alarm.
```
a-robinson: Clarifying in the documentation is great, but did we consider just fixing the behavior to delete the DO's alarm after deleting all its data instead? It doesn't strike me as all that difficult to do.

And if that's what people expect, then better to change it to that before documenting the opposite (even if we'd need a compat flag either way).

Maddy-Cloudflare:
```suggestion
  - Deletes all keys and associated values, effectively deallocating all storage used by the Durable Object. In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted. `deleteAll()` does not proactively delete [Alarms](/durable-objects/api/alarms/). Use [`deleteAlarm()`](/durable-objects/api/alarms/#deletealarm) to delete an alarm.
```

Maddy-Cloudflare (approved):

content/durable-objects/api/transactional-storage-api.md
elithrar: I don't know if we want to do that — there are certainly cases where you may want to delete the storage (an ongoing, persistent cost) vs. the alarm (which is more dynamic).

This question came up and this is a clarification, since it wasn't answered explicitly.

a-robinson: Alright, works for me if you / @vy-ton / @joshthoward are on board. I have no context on how often this has come up as confusing in either direction.

a-robinson (approved):

PR #13684, opened 2024-03-28
[Workers] Document Brotli compatibility flag, Fetch API compression and passthrough support

fhanau:
This flag will have a default compatibility date once https://github.com/cloudflare/workerd/pull/708 has been deployed. Adds documentation for the flag and describes the role of the Accept-Encoding header in the Fetch API, as well as how compressed data passthrough can be achieved.

### DIFF
--- a/content/workers/runtime-apis/fetch.md
+++ b/content/workers/runtime-apis/fetch.md
@@ -79,6 +79,42 @@ async function eventHandler(event) {
 
 {{</definitions>}}
 
+---
+
+## How the `Accept-Encoding` header is handled
+
+When making a subrequest with the `fetch()` API, you can specify which forms of compression to prefer that the server will respond with (if the server supports it) by including the [`Accept-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header.
+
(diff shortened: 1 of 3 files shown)


fhanau:
@irvinebroque this is a rewritten version of https://github.com/cloudflare/cloudflare-docs/pull/9519. I took an in-depth look at this yesterday and today and did a bunch of testing, this should be accurate and better describe how the compat flag and compression work in the production environment as opposed to the preview environment. There's one aspect I need to confirm internally before this is ready for review. Looking forward to discuss any questions/concerns you have once that's done!

kflansburg (approved):

content/workers/runtime-apis/fetch.md
```diff
@@ -79,6 +79,42 @@ async function eventHandler(event) {
...
+
+In addition to a change in the content encoding, recompression is also needed when a response uses an encoding not supported by the client. As an example, when a worker requests either brotli or gzip as the encoding but the client only supports gzip, recompression will still be needed if the server returns brotli-encoded data to the server [and be applied automatically]. Note that this behavior may also vary based on [compression rules](/rules/compression-rules/), which can be used to configure what compression should be applied for different types of data on the server side.
```
irvinebroque:
```suggestion
In addition to a change in the content encoding, recompression is also needed when a response uses an encoding not supported by the client. As an example, when a Worker requests either brotli or gzip as the encoding but the client only supports gzip, recompression will still be needed if the server returns brotli-encoded data to the server (and will be applied automatically). Note that this behavior may also vary based on the [compression rules](/rules/compression-rules/), which can be used to configure what compression should be applied for different types of data on the server side.
```

irvinebroque (approved):
These docs are _great_. Thank you!!

PR #13776, opened 2024-04-03
chore: Do not restrict search on product section

aseure:
**🔖 Summary**

Following some recent improvements we made for the documentation's search, this commit is a suggestion to remove an odd feature: currently, when the search feature is triggered from a specific product section, the search results which belong to that section are surfaced first, even though they are less relevant.

From my experience, and according to others, this is rather confusing, and this change is about preventing that behavior, effectively surfacing the most relevant results, no matter which page the user is currently browsing.

**🚧 Changes**

- Remove the Algolia [`optionalFilters`](https://www.algolia.com/doc/guides/managing-results/rules/merchandising-and-promoting/how-to/how-to-promote-with-optional-filters/) which were responsible for promoting lesser relevant search results from the currently browsed page.

**✅ Testing plan**

- Ran the documentation locally using `npm run dev` and looked at the before/after XHR requests going to Algolia to make sure the `optionalFilters` are not present anymore (also see the before/after screenshots showing what the search results look like).

**Before**
before

**After**
after

### DIFF
--- a/assets/search.ts
+++ b/assets/search.ts
@@ -8,11 +8,8 @@
 
   let {
     apikey: apiKey,
-    product,
     index: indexName,
   } = algoliaConfig
-
-  const facetFilters = product ? [`product:${product}`] : []
 
   function loaded() {
     window.docsearch({
@@ -32,9 +29,6 @@
       },
       getMissingResultsUrl({ query }) {
         return `/search/?q=${query}`;
-      },
-      searchParameters: {
-        optionalFilters: facetFilters
       },
       transformItems: items => {
         return items.filter(item => {


thibmeu:
I like the idea of making search more open. At the same time, I'd like to retain some context about the page the search is initiated from. Before `optionalFilter` was set, it happened quite often I got irrelevant result while searching for keywords shared across products, such as `pricing`, `api`, or `worker`.
These two search patterns seem legit but in conflict imo.

I don't know how Algolia exposes these options, but it might be interesting to have a checkbox `search in "product"` which a cookie retaining the default search preference set by the user.

rozenmd:
This suggestion breaks search for Workers.

Your preview: 
image

Prod: 
image

aseure:
@rozenmd As @thibmeu noted, this doesn't "break" the search, it's a different approach which however conflicts with the existing approach. This change is just a suggestion since a few people including myself found it confusing to have contextual search results, which vary depending on the page you navigate. I personally do not have any strong opinion on this though, I don't mind dropping this one. I'll let the PCX team decide here.

kodster28:
Howdy @aseure, this is a very opinionated choice informed by feedback from our community / internal folks (as you might've noticed from the comments :) ).

For now, let's keep this as is... but then I'll add something to our docs backlog to discuss / brainstorm about whether there's a more open way to address this.

PR #13800, opened 2024-04-03
workers: python examples

elithrar:
This PR adds examples for common Python Workers use-cases.

- [x] Returning headers
- [x] Parsing a URL + query params
- [x] Logging
- [x] Parsing incoming request JSON bodies
- [x] Returning JSON back to a client

Should this live within the Python docs, or in the general examples?

### DIFF
--- a/content/workers/languages/python/examples.md
+++ b/content/workers/languages/python/examples.md
@@ -23,11 +23,13 @@ async def on_fetch(request, env):
 
 ```python
 from js import Response, Headers
-from urllib.parse import urlparse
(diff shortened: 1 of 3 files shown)


irvinebroque:
Thank you!!! Python docs for now, for discoverability while in beta, and then think we need to extend broader docs to be able to flip between languages.

elithrar:
@irvinebroque - right, I'm hopeful we can do tabbed code examples (JS, TS, Python) for many things going forward!

This should give us some basics for common tasks. Ready to review.

irvinebroque (approved):

PR #13765, opened 2024-04-02
[Durable Objects] Initial DO JS RPC docs

vy-ton:
TODO: 

- [x] Use RPC in counter example
- [x] Rename `Configuration` section, add `Send requests to a Durable Object`
- [x] Update pricing to add RPC methods: https://github.com/cloudflare/cloudflare-docs/pull/13235/

Must merge after https://github.com/cloudflare/cloudflare-docs/pull/13252 which has Workers RPC links

DEFER to next PR:

- [ ] Change all occurrences of `state` to `ctx`
- [ ] Update `Get started` tutorial to RPC (maybe let JS RPC bake for a bit?)

PR #13252, opened 2024-03-03
Revamp Service Bindings docs for RPC

irvinebroque:
#### Consolidates Service binding documentation to be [within the Runtime APIs section of the docs](https://bib-jsrpc-api-docs.cloudflare-docs-7ou.pages.dev/workers/runtime-apis/bindings/service-bindings/).

Screenshot 2024-03-03 at 1 30 53 PM

Currently docs for configuring a Service binding, and docs for how to write code around Service bindings are in separate sections of the docs, which makes getting started hard, requires jumping back and forth between pages. Relevant content from [the configuration section](https://github.com/cloudflare/cloudflare-docs/blob/production/content/workers/configuration/bindings/about-service-bindings.md) has been moved here, and will be removed.

#### Explains what Service bindings are and what to use them for.

Screenshot 2024-03-03 at 1 31 18 PM

### Provides separate code examples RPC and HTTP modes of working with Service bindings.

refs https://github.com/cloudflare/workerd/pull/1658, https://github.com/cloudflare/workerd/pull/1692, https://github.com/cloudflare/workerd/pull/1729, https://github.com/cloudflare/workerd/pull/1756, https://github.com/cloudflare/workerd/pull/1757

PR #13801, opened 2024-04-03
Edits to JSRPC API docs

kentonv:
This is a series of edits on top of @irvinebroque's #13252. (These could have been review comments but it was easier to just edit.)

I've set the target branch to that branch, so merging this PR will append it to #13252, which should then all be merged together I think.

[EDIT: accidentally referenced #13765 initially, #13252 is what I meant)


content/workers/runtime-apis/rpc/lifecycle.md
```diff
@@ -142,22 +137,33 @@ class Foo extends RpcTarget {
...
 await counter.increment();
 ```
 
-But consider the case where the Worker service that you are calling may be far away across the network, as in the case of [Smart Placement](/workers/runtime-apis/bindings/service-bindings/#smart-placement) or [Durable Objects](/durable-objects). On the surface, you might guess that the code above makes two round trips, once when calling `getCounter()`, and again when calling `.increment()`.
+But consider the case where the Worker service that you are calling may be far away across the network, as in the case of [Smart Placement](/workers/runtime-apis/bindings/service-bindings/#smart-placement) or [Durable Objects](/durable-objects). The code above makes two round trips, once when calling `getCounter()`, and again when calling `.increment()`. We'd like to avoid this.
+
+With most RPC systems, the only way to avoid the problem would be to combine the two calls into a single "batch" call, perhaps called `getCounterAndIncrement()`. However, this makes the interface worse. You wouldn't design a local interface this way.
+
+Workers RPC allows a different approach: You can simply elide the first `await`:
```
irvinebroque:
```suggestion
Workers RPC allows a different approach: You can simply omit the first `await`:
```

More commonly used word?

irvinebroque (approved):

irvinebroque:
👍 🙌 this is great, sorry for making mess of it but think ultimately landing in good spot

content/workers/runtime-apis/rpc/lifecycle.md
kentonv: Yeah I guess.

I guess "elide" made sense to me since it's commonly used when talking about skipping unnecessary steps for performance, e.g. in compiler optimizations.

irvinebroque:
Let's just merge into my branch, before there are conflicts (@GregBrimble has a bunch of suggested edits on my branch, which then I can work through once this is merged, and then final review

PR #13802, opened 2024-04-03
[CASB] Box and Dropbox DLP support

maxvp:
PCX-10357 and 10358


kodster28 (approved):
In the future, may be worth encapsulating w/in a flexible partial (one with input variables).

PR #13859, opened 2024-04-04
[ZT] Best practices for updating WARP

ranbel:


content/cloudflare-one/connections/connect-devices/warp/download-warp/update-warp.md
```diff
@@ -0,0 +1,49 @@
...
+
+This guide covers best practices for updating the WARP client.
+
+## When to update WARP
+
+Cloudflare recommends keeping your WARP deployment as up-to-date as possible. If you run into issues that require troubleshooting or support tickets, one of the first requested actions by our support team will be to update your clients to the latest version.
+
+We also recognize that there is a cost associated for a business to go through an update cycle, potentially related to change management, QA version testing, and other critical activities for software updates. For customers with limited software update cycles, we recommend updating the client on a quarterly basis if possible. This will address ongoing bug fixes, performance improvements, and new feature development at a reasonable cadence. If quarterly updates are not possible, we recommend a minimum of twice a year update cycles.
+
+### Release schedule
```
kkrum, line 17: Let's add an additional note, maybe to this section that goes something like "Cloudflare commits to compatibility with our edge service for at least 1 year after a client is released. After 1 year, WARP Clients could potentially be unable to connect. Cloudflare will make every effort to communicate in advance through emails and in dash notifications if such a breaking change will occur".

kkrum (approved):
Looks great, only one minor addition I think we should make

ranbel:
@dh-cf Can you help review this PR?

PR #13867, opened 2024-04-05
Fix more redirects to RPC docs

irvinebroque:


irvinebroque:
prettier being silly, sorry for spam

PR #13878, opened 2024-04-05
remove --remote argument from wrangler d1 command

rezlerski:
remote not defined in the wrangler help for d1 execute, local argument defined


rozenmd (changes requested):
`wrangler d1 execute` has both --remote and --local, you might be on an old version of wrangler.

rezlerski:
> `wrangler d1 execute` has both --remote and --local, you might be on an old version of wrangler.

true, sorry

PR #13883, opened 2024-04-05
Unescaped unintentionally escaped HTML

IHaxU:
Corrected a small mistake that made it display incorrectly as you see in the following image due to HTML getting escaped

image

Instead of how it was intended to be

image


kodster28 (approved):

kodster28:
Thanks for the quick find and fix, @IHaxU!

PR #13889, opened 2024-04-05
[D1] Fixing issue where the wrong JSON function was used in an example

daskalou:
In the **Get the length of an array** section, the `json_array()` function is incorrectly used multple times when it should be the `json_array_length()` function.


kodster28 (approved):
Validated in https://github.com/cloudflare/cloudflare-docs/blob/d958843211dc8209e390fcf3fdb4c77050964f76/content/d1/build-with-d1/query-json.md?plain=1#L41

PR #13895, opened 2024-04-06
[Wrangler] mention which bindings are not supported locally

Cherry:
People are often confused about which bindings are supported locally, or not. This updates this table to be explicit about which bindings are not supported.

It also adds Rate Limiting as being supported locally.

A more concerning thing here is that if your worker is using one of the bindings that doesn't support local development, so therefore requires you to run with `--remote`, if that worker also uses Queues, there's essentially no way to test it without deploying straight to production. Any worker with Queues that you try to run with `--remote` just breaks entirely.

### DIFF
--- a/content/workers/testing/local-development.md
+++ b/content/workers/testing/local-development.md
@@ -36,18 +36,19 @@ Resources such as KV, Durable Objects, D1, and R2 will be stored and persisted l
 
 | Product                                   | Local Dev Supported | Remote Dev Supported |
 | ----------------------------------------- | ------------------- | -------------------- |
-| R2                                        | ✅                  | ✅                   |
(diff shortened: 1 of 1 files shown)


admah (approved):
This is great. Thanks, James!

kodster28 (approved):
approved from a content perspective as well

kodster28:
Fine to merge, @Cherry?

Cherry:
Go for it, thanks! This is accurate, and will just need maintaining as new bindings are added, or support is improved in future.

conngcoong1607:
warning: Queues are not yet supported in wrangler dev remote mode.
is there any way to make queue work with --remote , pls ?

PR #13945, opened 2024-04-09
[Workers AI] - Cleans up SDK docs

craigsdennis:
As we move towards a binding only world, this page will eventually change.

For now it had inconsistencies that our amazing community members pointed out.


kodster28 (approved):
approved from a content perspective (can't speak to the technical accuracy)

PR #13946, opened 2024-04-09
Workers: add known issue of types/node conflict with workers-types

admah:
Gives [this issue](https://github.com/cloudflare/workerd/issues/1298) greater visibility until we can fully address it.


WalshyDev (approved):

MattIPv4:
While this is probably the easiest recommendation in lieu of actually fixing this in worker-types, I think it should be called out that:

- This recommendation only works for folks using Node.js 20.x (not 18.x/21.x etc.)
- This recommendation prevents users from using new Node.js features in 20.9+
- This recommendation only works for package managers that support overrides (npm, not pnpm/yarm etc.)

admah:
@MattIPv4 thanks for calling those things out. This is on our roadmap to fix soon so the hope is that this workaround is a temporary fix.

josealvarez97:
@MattIPv4 @admah What is the best way to move forward for users using yarn?

Cherry:
yarn supports resolutions that should work in a similar way to npm overrides:

https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/

PR #13955, opened 2024-04-10
do: Update alarms.md

elithrar:
Add a simple example into the Alarms page + refactor the intro paragraph for clarity.

### DIFF
--- a/content/durable-objects/api/alarms.md
+++ b/content/durable-objects/api/alarms.md
@@ -8,11 +8,13 @@ weight: 1
 
 ## Background
 
-Durable Objects alarms allow you to schedule the Durable Object to be woken up at a time in the future. When the alarm's scheduled time comes, the `alarm()` handler method will be called.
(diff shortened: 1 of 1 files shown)


content/durable-objects/api/alarms.md
```diff
@@ -74,9 +76,46 @@ Alarms can be used to build distributed primitives, like queues or batching of w
  
 {{</definitions>}}
 
----
+## Example
```
joshthoward, line 79: Maybe link out to the existing [example](https://developers.cloudflare.com/durable-objects/examples/alarms-api/)?

joshthoward (approved):
I would prefer for us to link out to the existing complete example, but I'm fine as is if that's what you prefer.

content/durable-objects/api/alarms.md
elithrar: I wanted a much simpler example here - we link to the more complex one both in the intro section and at the tail of this doc, but it's far denser.

PR #13971, opened 2024-04-11
create cloudflared partial for reference

chrisflare:


content/reference-architecture/_partials/_description-of-cloudflared.md
```diff
@@ -0,0 +1,51 @@
+`cloudflared` is a lightweight daemon installed in an organizations' infrastructure that creates a tunnel via an outbound connection to Cloudflare's global network. The connector can be installed in a variety of ways:
```
kodster28, line 1:
```suggestion
---
_build:
  publishResources: false
  render: never
  list: never
---

`cloudflared` is a lightweight daemon installed in an organizations' infrastructure that creates a tunnel via an outbound connection to Cloudflare's global network. The connector can be installed in a variety of ways:
```

kodster28: This should fix your CI error :)

kodster28:
Closing since you mentioned you'd probably be using a diff approach

PR #13996, opened 2024-04-12
[Changelog] RuleID shortcode

kodster28:
There's currently a gap with WAF / other rules changes where you can't see the full rule Id in our docs / changelog entries.

This PR adds a new shortcode to assist with that, while still preserving the current table formatting.


layouts/shortcodes/rule-id.html
```diff
@@ -0,0 +1,8 @@
+{{- $ruleId := .Inner -}}
+{{- $shortenedRuleId := print "..." (substr $ruleId -8) }}
```
pedrosousa:
```suggestion
{{- $shortenedRuleId := print "..." (substr (trim $ruleId " ") -8) }}
```
Just in case the inner content includes spaces.

pedrosousa (approved):

PR #13979, opened 2024-04-11
[Workers] Adds Worker Examples For Python

megaconfidence:
This PR adds several Worker examples for Python and a new Python tab logo

PR #14007, opened 2024-04-14
Tweak ExportedHandler types to improve DX

### LINKED ISSUE #14006: Stop using ExportedHandler for TypeScript examples

There are [tonnes of places](https://github.com/search?q=repo%3Acloudflare%2Fcloudflare-docs%20ExportedHandler&type=code) where `ExportedHandler` is used, but here's  just one example:

- https://developers.cloudflare.com/workers/examples/basic-auth/

`ExportedHandler` results in significantly worsened DX when writing Workers in TypeScript. Instead of errors pointing to a place where the error actually occurs, they highlight the entire function instead, and echo a much more complex error stack back to the user. This makes it significantly harder to figure out where an issue might be occurring within a large worker.

Let's take the worker on the page linked above as an example. I've made one tiny edit in the worker that should be quick to see and resolve. Can you see it?

If I scroll up, you'll see the entire `fetch` declaration is highlighted as an error: 

And the following error:
```
Type '(request: Request<unknown, IncomingRequestCfProperties<unknown>>, env: { PASSWORD: string; }) => Promise<false | Response>' is not assignable to type 'ExportedHandlerFetchHandler<{ PASSWORD: string; }, unknown>'.
  Type 'Promise<false | Response>' is not assignable to type 'Response | Promise<Response>'.
    Type 'Promise<false | Response>' is not assignable to type 'Promise<Response>'.
      Type 'false | Response' is not assignable to type 'Response'.
        Type 'boolean' is not assignable to type 'Response'.
```

So all I know, if I'm proficient enough in TypeScript to parse that error (it's pretty complex), is that _somewhere_ in my function, I'm returning a `boolean` instead of a `Response`.

---

Now let's look at that, with the more traditional syntax of specifying function args:

All of a sudden, I'm now given the _exact_ line the error occurs on, and a much simpler error to parse:
`Type 'boolean' is not assignable to type 'Response'.`

It's a little more verbose syntax, yes, but the DX that users get from it so significantly improved

I think the following PR (at least where it removes the `req, env, ctx` type annotations) should be reverted:

- https://github.com/cloudflare/cloudflare-docs/pull/9891

And then further usage of `ExportedHandler` removed too.

Cherry:
Closes #14006

`ExportedHandler` without an explicit return type results in significantly worsened DX for users, especially when it comes to error handling. Errors are attributed the function as a whole, with IDEs marking the entire function as an issue, instead of the specific line that caused it.

<details>
<summary>Full Example</summary>

Let's take a worker from one of the examples in the docs: https://developers.cloudflare.com/workers/examples/basic-auth/

I've made one tiny edit in the worker that breaks it, but should be quick to see and resolve.

If I scroll up, you'll see the entire `fetch` declaration is highlighted as an error: 

And the following error:
```
Type '(request: Request<unknown, IncomingRequestCfProperties<unknown>>, env: { PASSWORD: string; }) => Promise<false | Response>' is not assignable to type 'ExportedHandlerFetchHandler<{ PASSWORD: string; }, unknown>'.
  Type 'Promise<false | Response>' is not assignable to type 'Response | Promise<Response>'.
    Type 'Promise<false | Response>' is not assignable to type 'Promise<Response>'.
      Type 'false | Response' is not assignable to type 'Response'.
        Type 'boolean' is not assignable to type 'Response'.
```

So all I know, if I'm proficient enough in TypeScript to parse that error (it's pretty complex), is that _somewhere_ in my function, I'm returning a `boolean` instead of a `Response`.

---

Now let's look at that, with the more traditional syntax of specifying function args:

All of a sudden, I'm now given the _exact_ line the error occurs on, and a much simpler error to parse:
`Type 'boolean' is not assignable to type 'Response'.`

It's a little more verbose syntax, yes, but the DX that users get from it so significantly improved

</details>

<details>
<summary>Contrived Example</summary>
Take the following Worker for example:

```ts
export interface Env {}

export default {
		return true;
	},
```

This errors with:
```
Type '(request: Request<unknown, IncomingRequestCfProperties<unknown>>, env: Env, ctx: ExecutionContext) => Promise<boolean>' is not assignable to type 'ExportedHandlerFetchHandler<Env, unknown>'.
  Type 'Promise<boolean>' is not assignable to type 'Response | Promise<Response>'.
    Type 'Promise<boolean>' is not assignable to type 'Promise<Response>'.
      Type 'boolean' is not assignable to type 'Response'.
```

Which again is a pretty complex error that requires reading through, and it ends up highlighting the `fetch` function in vscode as the error:

vs if you use the original setup:
```ts
export interface Env {}

export default {
	async fetch(request: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
		return true;
	},
};
```

The error you get here is so much clearer:
```
Type 'boolean' is not assignable to type 'Response'.
```

And it highlights the actual place in the code where the issue is:

</details>

By explicitly defining a return type on handlers, users are able to get better errors. So overall, this results in slightly more verbose function annotations, but _huge_ DX wins for shorter errors, and accurately highlighted errors to line/column numbers, which vastly outweighs the slightly additional verbosity.

See rest of thread for discussion.


content/workers/examples/cron-trigger.md
```diff
@@ -25,18 +25,18 @@ export default {
 
 ```ts
 export default {
-  async scheduled(event, env, ctx) {
+  async scheduled(controller: ScheduledController, env: unknown, ctx: ExecutionContext) {
```
WalshyDev: we should probably make an empty env and type with that rather than use unknown

content/workers/examples/cache-api.md
```diff
@@ -57,7 +57,7 @@ export default {
 
 ```ts
 export default {
-  async fetch(request, env, ctx) {
+  async fetch(request: Request, env: unknown, ctx: ExecutionContext): Promise<Response> {
```
WalshyDev: same here

content/workers/examples/basic-auth.md
```diff
@@ -177,7 +177,7 @@ function timingSafeEqual(a: string, b: string) {
 }
 
 export default {
-  async fetch(request, env) {
+  async fetch(request: Request, env: { PASSWORD: string }): Promise<Response> {
```
WalshyDev: also would prefer an env interface here, easier to expand for others later too if they copy this.

WalshyDev:
Gave my 2c on Discord already I'm happy with this change but interested to hear from @dario-piotrowicz given the initial PR that launched this.

content/workers/examples/cron-trigger.md
Cherry: Done, thanks!

content/workers/examples/basic-auth.md
Cherry: Done, thanks!

content/workers/examples/cache-api.md
Cherry: Done, thanks!

Cherry:
Thanks @WalshyDev! I've added that, and where appropriate added a better `Env` declaration to these examples such as in some of the R2 ones where it was previously just typed as `any`.

For anyone else reading, the discussion at https://github.com/cloudflare/workers-sdk/pull/5607#issuecomment-2056791696 is a good one to look through. I strongly think the docs should be tailored towards helping new users as much as possible.

dario-piotrowicz:
> Gave my 2c on Discord already I'm happy with this change but interested to hear from @dario-piotrowicz given the initial PR that launched this.

Thanks for pinging me @WalshyDev 🙂

Yeah after having read closely the conversation and out chat and considered the tradeoffs I agree that simpler/more verbose/less robust direct types are much simpler and clear for newcomers.

So I'm also on board with @Cherry's approach 🙂 

I wish it could also have the nice auto completion that `satisfied ExportedHandler` provides but I guess you can't have everything 😅 (well, we could include both but... 😅)

WalshyDev (approved):
Thank you James!
Code lgtm, change lgtm.

kodster28 (approved):
Trusting @WalshyDev / @dario-piotrowicz on the technical accuracy here, all the content pieces lgtm + wanted to unblock so this doesn't go stale.

kodster28:
Updated to fix the small merge conflict introduced by #13979.

At this point, we're just waiting on a review from @penalosa or someone else on the DevProd team.

RamIdeas:
I am in favour of improving DX in such ways but we need to consider future development of the platform and bringing users along with us e.g. https://github.com/cloudflare/workers-sdk/issues/5613 `ScheduledEvent` used to be the first param for the `scheduled()` handler but now there is a `ScheduledController` which is a subset of `ScheduledEvent`.

By not using `satisifes ExportedHandler`, this user didn't know of the change.

Cherry:
> I am in favour of improving DX in such ways but we need to consider future development of the platform and bringing users along with us e.g. [cloudflare/workers-sdk#5613](https://github.com/cloudflare/workers-sdk/issues/5613) `ScheduledEvent` used to be the first param for the `scheduled()` handler but now there is a `ScheduledController` which is a subset of `ScheduledEvent`.
> 
> By not using `satisifes ExportedHandler`, this user didn't know of the change.

I agree that this is the one downside of annotating the function args specifically, but there are _significant_ downsides to using `satisifes ExportedHandler` as previously discussed in this thread.

Those absolutely trump the rare cases where types like this do change in my opinion, which is _extremely_ rare.

Recommending `satisifes ExportedHandler` results in worsened DX in the vast vast majority of cases, _especially_ for new users, and that's what docs and C3 should be optimising for, in my opinion. There's some further discussion in this PR which I would recommend reading for more context: https://github.com/cloudflare/workers-sdk/pull/5607#issuecomment-2059476305

GregBrimble:
I'm kinda liking Dario's (half-joking) suggestion of both... It gives you `env` typing across the whole thing, and also gives decent typescript errors specific to that return value.

```typescript
export default {
	async fetch(request, env): Promise<Response> {
		env // typed
		return true // errors here
	},
} satisfies ExportedHandler<{ MY_BINDING: string }>;
```

We'll possibly/probably be encouraging `WorkerEntrypoint` in the future which has similar problems/solutions, and in fact, needs even the parameters specified:

```typescript
import { WorkerEntrypoint } from "cloudflare:workers";

export default class extends WorkerEntrypoint<{ MY_BINDING: string }> {
	async fetch(request: Request): Promise<Response> {
		this.env // typed
		return true // errors here
	}
}
```

Cherry:
I'm not against both if that's the consensus here. It's a little verbose, but I think it is going to be the best compromise to address everyone's concerns.

Happy to update the PR if that's what everyone would like to see.

penalosa:
I'm been out, so I'm just seeing this now, but it sounds like there's been a lot of useful discussion!

Ultimately, I think this is a tradeoff between DX and correctness, and in my mind correctness should be prioritised here. While `ExportedHandler` doesn't always give the most helpful error messages, it _does_ ensure that your handler actually conforms to the expected interface, as well as providing much better discoverability through typing of additional handlers, and the `env` and `ctx` arguments (which you'd otherwise have to specific manually). For instance, a user unfamiliar with the Cloudflare platform could very easily write something like this:

```js
export default {
	async fetch(
		request,
		env,
		ctx
	) {
		return "Hello World!"
	}
}
```

This would pass typechecking, but would fail at _runtime_ with a cryptic `Error 1101`. If they'd been using `ExportedHandler`, that would fail typechecking with:

```
Type '(request: Request<unknown, IncomingRequestCfProperties<unknown>>, env: Env, ctx: ExecutionContext) => Promise<string>' is not assignable to type 'ExportedHandlerFetchHandler<Env, unknown>'.
  Type 'Promise<string>' is not assignable to type 'Response | Promise<Response>'.
    Type 'Promise<string>' is not assignable to type 'Promise<Response>'.
      Type 'string' is not assignable to type 'Response'.
```

which is moderately cryptic, but not excessively so. 

With a solution like the one proposed in this PR, I could see someone changing the `Promise<Response>` type that `fetch` is annotated with to a `Promise<string>` and expecting that to work.

That's obviously a contrived example, but what `ExportedHandler` gives you is a guarantee that all handlers, arguments, and return types are correctly typed, rather than relying on the user _knowing_ the shape of the API and individually typing each argument and return type correctly. I think having a correctly, reliably typed handler shape provides a better DX than one which relies on the user already knowing the API shape.

However, `satisfies` is a crude tool for the assertion we're trying to make here. We're trying to say that the `fetch` handler has a certain type, and that mismatches from that type should be treated as an error _within_ the body of the handler. Instead, TypeScript is inferring a _different_ type for the fetch handler, which then doesn't match the `satisfies` constraint. I think the best solution here to improve error DX is to use that extra information we have to give the TypeScript compiler some help, and explicitly annotate the return type of the `fetch` handler.

Cherry:
> which is moderately cryptic, but not excessively so.

I would argue that this is much more than just moderately cryptic, especially if the function body is tens or hundreds of lines long, and especially if folks are new to TypeScript.

However, I do ultimately agree with your conclusion:
> I think the best solution here to improve error DX is to use that extra information we have to give the TypeScript compiler some help, and explicitly annotate the return type of the fetch handler.

So handlers in the docs would end up looking something like:
```ts
export default {
		return true 
	},
```

If everyone is in agreement, I'm happy to update this PR to conform to that.

Cherry:
I've updated this now to match the agreed syntax. Let me know if anyone wants any further changes!

kodster28:
Checked with @penalosa, @WalshyDev, and @GregBrimble and this is good to go.

Thanks for the submission, @Cherry!

PR #14021, opened 2024-04-15
turnstile: fix refresh timed-out heading

Le0Developer:
The heading is currently a h1 instead of a h2.
This causes the heading to not appear in the navbar and is just plain wrong.

https://developers.cloudflare.com/turnstile/get-started/client-side-rendering/#refresh-a-timed-out-widget


pedrosousa:
Thanks, @Le0Developer, but this was already fixed in #14039.

PR #14029, opened 2024-04-15
Release-Apr-15-2024: Releasing 9 detections

vs-mg:
Release-Apr-15-2024: Releasing 9 detections

### DIFF
--- a/data/changelogs/waf.yaml
+++ b/data/changelogs/waf.yaml
@@ -2,11 +2,14 @@
 link: "/waf/change-log/"
 productName: WAF
 entries:
-- publish_date: '2024-04-02'
-  scheduled_date: '2024-04-08'
+- publish_date: '2024-04-15'
+  scheduled_date: '2024-04-22'
   individual_page: true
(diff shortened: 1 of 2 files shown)


data/changelogs/waf.yaml
```diff
@@ -2,11 +2,14 @@
 link: "/waf/change-log/"
 productName: WAF
 entries:
-- publish_date: '2024-04-02'
-  scheduled_date: '2024-04-08'
+- publish_date: '2024-04-15'
```
kodster28: You'd actually want to keep these values what they were before. Otherwise, it'll seem like there are new scheduled changes that are coming out.

https://release-apr-15-2024.cloudflare-docs-7ou.pages.dev/waf/change-log/index.xml

Screenshot 2024-04-15 at 12 29 40 PM

vs-mg: Absolutely. Reverting them now.

pedrosousa (approved):

PR #14047, opened 2024-04-16
[Workers] Remove page which is now redirected

pedrosousa:
The [page being deleted](https://github.com/cloudflare/cloudflare-docs/blob/a78ef5684a01e8617ac5f039b7a27ea49b68eaba/content/workers/configuration/bindings/_index.md) has an [external link](https://github.com/cloudflare/cloudflare-docs/blob/a78ef5684a01e8617ac5f039b7a27ea49b68eaba/content/workers/configuration/bindings.md) on the same path, which makes in unreachable.

Due to the double function associated with the same path (actual page with content and external link), sometimes we get errors in the link checker, depending on which file Hugo parses and builds first (the external link or the unreachable page).

Also had to fix 70+ links, sometimes removing the anchor part of the target URL because the target section no longer exists.

This issue had already been identified in https://github.com/cloudflare/cloudflare-docs/pull/13252/files#r1555961030.


marciocloudflare (approved):

content/workers/wrangler/api.md
```diff
@@ -345,19 +345,19 @@ The bindings supported by `getPlatformProxy` are:
 
  * [Environment variables](/workers/configuration/environment-variables/)
 
- * [Service bindings](/workers/configuration/bindings/#service-bindings)
```
pedrosousa, line 348: Used the same links as in the [Bindings (env)](https://developers.cloudflare.com/workers/runtime-apis/bindings/) page.

PR #14057, opened 2024-04-16
Fix curl example by closing ' quote in text-embeddings.md

juanmirocks:


kodster28 (approved):
Thanks for flagging (and fixing), @juanmirocks

PR #14076, opened 2024-04-17
[Vectorize] Rename Configuration into Best Practices

Maddy-Cloudflare:


content/_redirects
pedrosousa, line 1: You're missing redirects for all child pages of `/vectorize/configuration/*` which were moved:
* `/vectorize/configuration/create-indexes/`
* `/vectorize/configuration/insert-vectors/`
* `/vectorize/configuration/query-vectors/`

You could create a dynamic redirect addressing all these three pages at once. Such a redirect entry needs to appear after all static redirects (i.e., all the ones without a wildcard in the source URL):

```
/vectorize/configuration/* /vectorize/best-practices/:splat 301
```

You can also add three static redirects near the one you added already, one for each relocated page.

pedrosousa (approved):

vy-ton (approved):

PR #14087, opened 2024-04-17
[Rules] Update edit-expressions.md

luishcardoso:
We'll benefit from publicly documenting how the Expression Builder takes care of any necessary escaping in the final expression. SPM-2353


pedrosousa:
This was addressed in a separate PR: #14412.

PR #14116, opened 2024-04-18
[Docs] Use `rule-id` shortcode for most rule/ruleset IDs

pedrosousa:
Also includes some minor formatting changes.

Updates WAF changelog pages from 2024.

### DIFF
--- a/layouts/shortcodes/rule-id.html
+++ b/layouts/shortcodes/rule-id.html
@@ -2,8 +2,8 @@
 {{- $shortenedRuleId := print "..." (substr (trim $ruleId " ") -8) }}
 
 <code class="ruleID">
-<span class="ruleID" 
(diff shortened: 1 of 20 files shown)


layouts/shortcodes/rule-id.html
```diff
@@ -2,8 +2,8 @@
 {{- $shortenedRuleId := print "..." (substr (trim $ruleId " ") -8) }}
 
 <code class="ruleID">
-<span class="ruleID" 
-        title="Click to copy full rule ID"
+<span class="ruleID"
+        title="Click to copy the full ID"
```
pedrosousa, line 6: Avoiding the word "rule" in the tooltip so that we can use the shortcode with both rule IDs and ruleset IDs.

kodster28 (approved):

PR #14154, opened 2024-04-19
[Workers AI] Adds dynamic model names to the responses section

craigsdennis:
These shouldn't be hardcoded anymore. We have the power.

### DIFF
--- a/layouts/partials/models/responses/text-generation.md
+++ b/layouts/partials/models/responses/text-generation.md
@@ -37,7 +37,7 @@ return new Response(stream, {
 
 Using the REST API:
 
-```sh
+```bash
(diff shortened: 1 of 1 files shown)


layouts/partials/models/responses/text-generation.md
```diff
@@ -38,7 +38,7 @@ return new Response(stream, {
 Using the REST API:
 
 ```sh
```
kodster28:
```suggestion
```bash
```

layouts/partials/models/responses/text-generation.md
```diff
@@ -48,7 +48,7 @@ Streaming responses use [server-sent events](https://developer.mozilla.org/en-US
...
```
kodster28:
```suggestion
```bash
```

kodster28 (approved):
Should we move the curl examples over to `bash` to allow specific text copy/pasting?
(similar to #14040)

PR #14156, opened 2024-04-20
Fixes small typo in button type attribute

1-bit-wonder:


content/pages/functions/plugins/static-forms.md
```diff
@@ -39,7 +39,7 @@ filename: public/sales-enquiry.html
   <form data-static-form-name="sales">
     <label>Email address <input type="email" name="email" /></label>
     <label>Message <textarea name="message"></textarea></label>
-    <button type="Submit">
+    <button type="submit">
```
kodster28:
```suggestion
    <button type="submit">Submit</button>
```

kodster28 (approved):
Believe there's also the closing `</button>` tag + missing button text here, at least by glancing over at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

PR #14165, opened 2024-04-22
Create tutorial: Create a sitemap for a headless CMS with Workers

johnsicili:


kristianfreeman:
Hey @johnsicili - quick feedback on some high-level stuff here.

I think this tutorial could be named "Create a sitemap for a Sanity-powered site with Workers" or something similar. There's a lot of verbiage in the tutorial that is "If you aren't using Sanity, do something different" - in my opinion, better to be focused on one tool in the tutorial. You can remove all that verbiage throughout, and just add one disclaimer at the beginning: "This tutorial uses Sanity.io, a headless CMS tool. If you are using something different, your configuration may be different"

This is all IMO, would love to hear what you think!

johnsicili:
Hey @kristianfreeman, great point, and thanks for the suggestions. I updated the article (including the folder) to be Sanity specific. I also added the command to install the Sanity client.

Let me know if you have any other suggestions/requests.

Thanks for the opportunity and love your work btw.

kodster28 (approved):

kodster28:
Heyo @johnsicili, merging this in... will do a follow up in a day or two adding in some author attribution.

TYSM for participating!

PR #14189, opened 2024-04-22
Release-Apr-22-2024: Releasing some detections.

vs-mg:
Release-Apr-22-2024: Releasing some detections.


kodster28 (approved):
Would it make sense to use the `{{<rule-id>}}` component also for today's changes?

PR #14198, opened 2024-04-23
Fix binding assignment on Qwik guide

aidhand:
The original produces a type error:

`Conversion of type 'QwikCityPlatform' to type '{ MY_KV: KVNamespace<string>; }' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

Using create-cloudflare-cli and specifying Qwik as the framework, bindings are by default set on platform.env as specified in entry.cloudflare-pages.tsx

```
declare global {
  interface QwikCityPlatform extends PlatformCloudflarePages {
    env: Env
  }
}
```


dario-piotrowicz (approved):
Thanks a bunch for the fix @aidhand! ❤️ 

Yeah it looks like we made a silly mistake here! 😅

PR #14245, opened 2024-04-24
[Workers] Add new error code

Warfields:
Adds the new 1103 status for when a worker violates Cloudflare's terms of use


Warfields:
Converted to draft, wording is possibly changing.

Warfields:
@irvinebroque and @deadlypants1973 can we take a look at this?

WalshyDev (approved):

PR #14249, opened 2024-04-25
[Feature] Add filtering to list-examples component

kodster28:
## Purpose

This adds filtering capabilities to the `/workers/examples/` page by adding the option for dynamically generated filters and filtering to the `list-examples` component.

The initial use case is specifically for Workers (highlighting various tags + languages + Workers playground links), but think this pattern can hold up for future efforts around `Workers AI`, `Rules`, CF1 examples, and more... maybe even pulling in more examples from other sections or better highlighting tutorials. Also has the benefit of getting me more experience around this + showing the immense value of metadata on specific examples.

## Visual
https://github.com/cloudflare/cloudflare-docs/assets/26727299/bedd5452-13ec-4048-a436-f22b2a689c95

## Tasks
Remaining to dos:
- [x] Write component docs
- [x] Add tracking
- [x] Review `Originless` as a category (this just sounds weird to me rn, but maybe it's still accurate)


layouts/shortcodes/list-examples.html
```diff
@@ -1,25 +1,139 @@
...
+            }
+          }
+        }
+      }
+    })
+    filterList();
+  })
+
+  // define persistent functions
+  function zarazTrackSelect(eventName, selectedOption) {
```
kodster28, line 80: This will error during preview deployments + local dev, but tested and confirmed it will work in production

layouts/shortcodes/list-examples.html
```diff
@@ -1,25 +1,139 @@
...
-      <div class="DocsCodeExamplesOverview--example-description">
-        <p>{{ .Params.summary }}</p>
-      </div>
-    </li>
-    {{- end -}}
+  {{- range $index, $page := $pages.ByWeight -}}
+  {{- $x := $page.RelPermalink -}}
+  {{- $isExample := (strings.HasPrefix $x $target) -}}
+  {{- if and (ne $x $target) $isExample -}}
+  <li class="DocsCodeExamplesOverview--example" {{ if ne (index $filters 0) "" }} {{ range $index2, $filter :=$filters
```
pedrosousa:
```suggestion
  <li class="DocsCodeExamplesOverview--example" {{ if ne (index $filters 0) "" }} {{ range $index2, $filter := $filters
```

layouts/shortcodes/list-examples.html
```diff
@@ -1,25 +1,139 @@
...
+        }
+      }
+      if (keepValue) {
+        example.style.display = "";
+      } else {
+        example.style.display = "none";
+      }
+    }
+    let newUrl = new URL(window.location);
+    newUrl.search = ''
```
pedrosousa:
```suggestion
    newUrl.search = '';
```

pedrosousa: There are a few other missing ";" below this point. Since we'd added them until here, I'd keep adding them in the rest of the file.

layouts/shortcodes/list-examples.html
```diff
@@ -1,25 +1,139 @@
...
-        <p>{{ .Params.summary }}</p>
-      </div>
-    </li>
-    {{- end -}}
+  {{- range $index, $page := $pages.ByWeight -}}
+  {{- $x := $page.RelPermalink -}}
+  {{- $isExample := (strings.HasPrefix $x $target) -}}
+  {{- if and (ne $x $target) $isExample -}}
+  <li class="DocsCodeExamplesOverview--example" {{ if ne (index $filters 0) "" }} {{ range $index2, $filter :=$filters
+    }} {{ with (index $page.Params $filter) }} data-{{$filter}}="{{ delimit (.) " |" }}" {{ end }} {{ end }} {{ end }}>
```
pedrosousa:
```suggestion
    }} {{ with (index $page.Params $filter) }} data-{{$filter}}="{{ delimit (.) "|" }}" {{ end }} {{ end }} {{ end }}>
```
This is causing filtering issues, since `"JavaScript"` != `"JavaScript "` (with a space at the end) when doing the filtering.

pedrosousa (approved):

irvinebroque (approved):

PR #14261, opened 2024-04-25
[Durable Objects] Add instructions for properties of Durable Object errors 

joshthoward:


content/durable-objects/best-practices/error-handling.md
```diff
@@ -25,7 +23,17 @@ A Durable Object stub is a client Object used to send requests to a remote Durab
...
+
+{{<Aside type="note" header="Durable Object stubs">}}
+
+If any exception is thrown from a Durable Object stub, then the stub will need to be recreated prior to retrying requests. This is to provide E-order semantics as discussed in [create Durable Object stubs and send requests](/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/#create-durable-object-stubs-and-send-requests).
+
+JavaScript Errors with the property `.retryable` set to True are guaranteed to be safely retryable. Given that requests to `ErrorThrowingObject` are idempotent, or can be applied multiple times without changing the response, the request can also be retried when `.remote` is set to True. When the `.remote` property is set to True, `.retryable` will always be false because the platform cannot verify that application logic is safely retryable.
+
+JavaScript Errors with the property `.overload` set to True should not be retried. If a Durable Object is overloaded, then retrying will worsen the overload and increase the overall error rate. When the `.remote` property is set to True, `.overload` will always be false because the platform will not suggest retrying overload errors.
```
jclee: FWIW, the property name as currently implement is `.overloaded` rather than `.overload`.

content/durable-objects/best-practices/error-handling.md
```diff
@@ -25,7 +23,17 @@ A Durable Object stub is a client Object used to send requests to a remote Durab
...
+Any uncaught exceptions thrown by a Durable Object or its infrastructure will be propagated to the callsite of the client. Catching these exceptions allows you to retry creating the Durable Object stub and sending requests.
+
+{{<Aside type="note" header="Durable Object stubs">}}
+
+If any exception is thrown from a Durable Object stub, then the stub will need to be recreated prior to retrying requests. This is to provide E-order semantics as discussed in [create Durable Object stubs and send requests](/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/#create-durable-object-stubs-and-send-requests).
+
+JavaScript Errors with the property `.retryable` set to True are guaranteed to be safely retryable. Given that requests to `ErrorThrowingObject` are idempotent, or can be applied multiple times without changing the response, the request can also be retried when `.remote` is set to True. When the `.remote` property is set to True, `.retryable` will always be false because the platform cannot verify that application logic is safely retryable.
```
jclee: As currently worded, I think this could be interpreted to mean that non-idempotent requests are also safely retryable when `.retryable` is true, which I don't think is currently the case.  The distinction between `.retryable` and `.remote` is that a `.remote` error came from an uncaught exception in the remote worker, while a `.retryable` error came from a transient error in infrastructure, like a disconnection.  I think in the latter case, the client can't know for sure if a non-idempotent request is safely retryable.

Similarly, I think it is possible for an exception to have both `.retryable` and `.remote` set, if a remote worker itself failed to handle a `.retryable` error from one of its own remote operations.

content/durable-objects/best-practices/error-handling.md
```diff
@@ -25,7 +23,17 @@ A Durable Object stub is a client Object used to send requests to a remote Durab
...
+
+If any exception is thrown from a Durable Object stub, then the stub will need to be recreated prior to retrying requests. This is to provide E-order semantics as discussed in [create Durable Object stubs and send requests](/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/#create-durable-object-stubs-and-send-requests).
+
+JavaScript Errors with the property `.retryable` set to True are suggested to be retried if requests to the Durable Object are idempotent, or can be applied multiple times without changing the response. If requests are not idempotent, then you will need to decide what is best for your application. When the `.remote` property is set to True, `.retryable` will always be false because the platform cannot verify that application logic is safely retryable.
+
+JavaScript Errors with the property `.overloaded` set to True should not be retried. If a Durable Object is overloaded, then retrying will worsen the overload and increase the overall error rate. When the `.remote` property is set to True, `.overloaded` will always be false because the platform will not suggest retrying overload errors.
```
Frederik-Baetens, line 34: > When the `.remote` property is set to True, `.overloaded` will always be false because the platform will not suggest retrying overload errors.

This sentence doesn't make much sense to me. Do we actually want to convey that `.remote` being set will mean that the DO is never overloaded? I.e. overloaded errors always have `.remote` false? That seems weird, since overloaded errors seem like remote errors by definition?

Or do we mean to convey here that when a DO is overloaded, the `.retryable` field will always be false, since we don't recommend retrying overloaded errors? (Have not looked at @jclee 's changes around error handling, so I don't know if that's actually the case, but it would make more sense to me.

jclee: I'm not quite sure myself if the mentions of `.remote` here are helpful... As I mentioned in an earlier comment,  transitively thrown errors can result in cases where `.remote` is true at the same time as either `.retryable` or `.overloaded`, so it's not actually a valid invariant.  I think it might be better to just omit the `.remote` sentences?

You're correct that one valid invariant is that the runtime will not set `.retryable` to true on an `.overloaded` exception or vice versa.  But I think the existing descriptions already imply this invariant, so I'm not sure explicitly mentioning it adds much.

jclee:
Looks like Josh's branch is in a personal repo... I couldn't find a way in the github UI to update the PR to use a different branch, so I opened a new PR here: https://github.com/cloudflare/cloudflare-docs/pull/15305

KimJ15:

PR #14263, opened 2024-04-26
Fix multiple typos under `content/workers` directory

n0vad3v:


kodster28 (approved):
Thanks for the catches + fixes, @n0vad3v

PR #13685, opened 2024-03-28
adds `--show-interactive-dev-session` flag to `dev` options

loganpowell:
I was very happy to find [this](https://github.com/cloudflare/workers-sdk/pull/5099) (I'm also using Turbo repo and the interactive prompt reduces the signal:noise), so I figured others might also be grateful for the option.

PR #13188, opened 2024-02-27
[Workers] Adds new flag for wrangler commands

KaiSpencer:
A new flag for `wrangler dev` and `wrangler pages dev` is to be added in https://github.com/cloudflare/workers-sdk/pull/5099

This change covers the documentation for this new flag.

NB: the optional meta is ordered differently for each of these, I have stayed consistent with what is currently in place for the other flags in the respective sections.

PR #14277, opened 2024-04-26
[Chore] Remake Wrangler dev --show-interactive-dev-session

kodster28:
Remakes #13685 (currently blocked by needing to merge in production)


kodster28:
whoops, looks like we're already tracking in #13188

PR #14278, opened 2024-04-26
Remove city count from Images docs

abstergail:
outdated city count, switched to a link instead


kodster28 (approved):

kodster28:
Merging b/c I need to see this below 100 for my soul :)

PR #14291, opened 2024-04-28
remove useless '\' in '\ ' which cause regex101 error, and add kaios

liudongmiao:
- `\ ` has no special meaning, and regex101 mark it as error.
- add `kaios` and converted into lowercase.

This PR makes the user-agent mobile regex same.


liudongmiao:
Please check whether cloudflare use different regex actually. (The difference is only for `kaios`).

kodster28 (approved):

PR #14307, opened 2024-04-29
Azure Active Directory now named Microsoft Entra ID

acrogenesis:
Azure Active Directory is now named Microsoft Entra ID.
https://learn.microsoft.com/en-us/entra/fundamentals/new-name


maxvp:
This issue is tracked internally and requires file renaming + setting up redirects. Closing this PR as it doesn't cover every issue.

PR #14334, opened 2024-05-01
[Workers] Adds New Python Workers Examples

megaconfidence:


content/workers/examples/cache-tags.md
```diff
@@ -12,7 +12,7 @@ weight: 1001
 layout: example
```
kodster28, line 13: Can you add `Python` to the `languages` array up top of the file? It'll help add the metadata necessary for filtering.

kodster28 (approved):
Looks good from a content perspective (aside from my previous comment).

@hoodmane / @garrettgu10, did y'all have any thoughts or suggestions on the code samples?

irvinebroque (approved):

content/workers/examples/cache-post-request.md
```diff
@@ -111,5 +111,41 @@ export default {
...
+        cache_key = Request.new(cache_url.toString(), method='GET', headers=headers)
+
+        # Find the cache key in the cache
+        cache = caches.default
+        response = await cache.match(cache_key)
+
+        # Otherwise, fetch response to POST request from origin
+        if response is None:
+            response = await fetch(request)
+            ctx.waitUntil(create_proxy(cache.put(cache_key, response.clone())))
```
hoodmane, line 143: What happens if you don't put `create_proxy` here? Seems like the result of `cache.put` should be a `Promise` which is what `ctx.waitUntil` is expecting? Ah I suppose the problem is that the result of `cache.put` gets wrapped into a `PyodideFuture` so you indeed need to extend it's lifetime. Yet another wart.

content/workers/examples/cache-post-request.md
```diff
@@ -111,5 +111,41 @@ export default {
...
+        # Hash the request body to use it as a part of the cache key
+        body = await request.clone().text()
+        body_hash = hashlib.sha256(body.encode('UTF-8')).hexdigest()
+
+        # Store the URL in cache by prepending the body's hash
+        cache_url = URL.new(request.url)
+        cache_url.pathname = "/posts" + cache_url.pathname + body_hash
+
+        # Convert to a GET to be able to cache
+        headers = Headers.new(dict(request.headers).items())
```
hoodmane, line 133: Shouldn't you be able to say:
```py
Headers.new(request.headers)
```
Or even just pass `request.headers` directly?

content/workers/examples/cache-tags.md
```diff
@@ -103,4 +103,31 @@ export default {
...
+{{<tab label="py">}}
+
+```py
+import json
+from js import Response, URL, JSON, fetch
+
+async def on_fetch(request):
+    request_url = URL.new(request.url)
+    params = request_url.searchParams
+    tags = params.get("tags").split(",") if params and params.has("tags") else []
```
hoodmane, line 115: You can also do:
```suggestion
    tags = params["tags"].split(",") if "tags" in params
```

content/workers/examples/cache-tags.md
```diff
@@ -103,4 +103,31 @@ export default {
...
+    url = params.get("uri") if params and params.has("uri") else ""
```
hoodmane: You can also do a bit shorter here:
```suggestion
    tags = params["tags"].split(",") if "tags" in params
    url = params["uri"] or ""
```

content/workers/examples/cache-tags.md
```diff
@@ -103,4 +103,31 @@ export default {
...
+
+    if url is None:
+        error = json.dumps({"error": "URL cannot be empty"})
+        return Response.json(JSON.parse(error), status=400)
```
hoodmane: Yeah would definitely be better to have something simpler here. It is also possible to use `to_js` with `dict_converter=Object.fromEntries` but it's not really shorter or clearer.

content/workers/examples/cache-tags.md
```diff
@@ -103,4 +103,31 @@ export default {
...
+
+    options = {"cf": {"cacheTags": tags}}
+
+    result = await fetch(url, options)
+    cache_status = result.headers.get("cf-cache-status")
+    last_modified = result.headers.get("last-modified")
```
hoodmane: It is possible to use `[]` to look these up too:
```suggestion
    cache_status = result.headers["cf-cache-status"]
    last_modified = result.headers["last-modified"]
```

content/workers/examples/cache-post-request.md
hoodmane: Also, this leaks the `Future` which probably doesn't own much data but it's a little unfortunate. We'll definitely want some way to make it convenient to use `waitUntil` without `create_proxy` and without leaks.

content/workers/examples/cache-tags.md
megaconfidence: Sure thing!

content/workers/examples/cache-post-request.md
megaconfidence: Yup, it goes out of scope without `create_proxy`

content/workers/examples/cache-post-request.md
megaconfidence: Nope, it turns out that Headers expect a list of tuples

content/workers/examples/cache-tags.md
megaconfidence: Done

content/workers/examples/cache-post-request.md
hoodmane: It'll be really good to have this documentation when we're working on quality of life improvements since we can look at all of the annoying stuff in here and make better alternatives.

content/workers/examples/cache-post-request.md
hoodmane: Are you sure? I just tried it out and it seems to work fine for me.

PR #14363, opened 2024-05-02
[Workers] Don't show empty page

kodster28:


kodster28:
Going to address via redirects instead

PR #14387, opened 2024-05-03
[Analytics] Add custom event for copying a code block

kodster28:


assets/events.ts
```diff
@@ -326,12 +327,22 @@ export function zarazTrackDocEvents() {
...
   }
+  if (copyCodeBlockButtons.length > 0) {
+    for (const copyButton of copyCodeBlockButtons as any) {
+      copyButton.addEventListener("click", () => {
+        const codeBlockElement = copyButton.parentElement.parentElement.firstElementChild;
+        zaraz.track('copy button link click', {
+          title: codeBlockElement.getAttribute('title') ? codeBlockElement.getAttribute('title') : 'title not set',
+          language: codeBlockElement.getAttribute('language') ? codeBlockElement.getAttribute('language') : 'language not set',});
+      });
+  }
```
pedrosousa:
```suggestion
    }
```

assets/events.ts
```diff
@@ -326,12 +327,22 @@ export function zarazTrackDocEvents() {
...
       });
   }
   }
+  if (copyCodeBlockButtons.length > 0) {
+    for (const copyButton of copyCodeBlockButtons as any) {
+      copyButton.addEventListener("click", () => {
+        const codeBlockElement = copyButton.parentElement.parentElement.firstElementChild;
+        zaraz.track('copy button link click', {
+          title: codeBlockElement.getAttribute('title') ? codeBlockElement.getAttribute('title') : 'title not set',
+          language: codeBlockElement.getAttribute('language') ? codeBlockElement.getAttribute('language') : 'language not set',});
```
pedrosousa:
```suggestion
          title: codeBlockElement.getAttribute('title') ?? 'title not set',
          language: codeBlockElement.getAttribute('language') ?? 'language not set',});
```
No need to repeat the statement; we can use the [null-ish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing), supported since 2020. However, if `title` or `language` is eventually set to an empty string, it will be used as such (it won't use the alternative default string).

assets/events.ts
```diff
@@ -286,6 +286,7 @@ export function zarazTrackDocEvents() {
   const dropdowns = document.getElementsByTagName("details")
   const glossaryTooltips = document.getElementsByClassName("glossary-tooltip")
   const playgroundLinks = document.getElementsByClassName("playground-link")
+  const copyCodeBlockButtons = document.getElementsByClassName("copyCode-button")
```
pedrosousa, line 289: Maybe we should also track these clicks in the homepage, which has code blocks... we would have to update `zarazTrackHomepageLinks()` also.

pedrosousa (approved):

PR #14394, opened 2024-05-04
Clarify ResponseInit options

kflansburg:
In https://github.com/cloudflare/workers-rs/issues/567 it was noted that our docs do not show all of the possible options for `ResponseInit`. This updates our docs to reflect what is in `worker-types`.

### DIFF
--- a/content/workers/runtime-apis/response.md
+++ b/content/workers/runtime-apis/response.md
@@ -39,16 +39,19 @@ let response = new Response(body, init);
 
 Valid options for the `options` object include: {{<definitions>}}
 
-- `status` {{<type>}}int{{</type>}}
-
-  - The status code for the response, such as `200`.
-
-- `statusText` {{<type>}}string{{</type>}}
-
-  - The status message associated with the status code, such as, `OK`.
-
+<!-- Does `cf` do anything in a Response? -->
+- `cf` {{<type>}}any | null{{</type>}}
+  - An object that contains Cloudflare-specific information. This object is not part of the Fetch API standard and is only available in Cloudflare Workers.
+- `encodeBody` {{<type>}}string{{</type>}}
+  - Workers have to compress data according to the `content-encoding` header when transmitting, to serve data that is already compressed, this property has to be set to `"manual"`, otherwise the default is `"automatic"`.
 - `headers` {{<type-link href="/runtime-apis/request#parameters">}}Headers{{</type-link>}} | {{<type-link href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String">}}ByteString{{</type-link>}}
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/response.md
```diff
@@ -60,8 +63,6 @@ Valid options for the `options` object include: {{<definitions>}}
   - A getter to get the body contents.
 - `response.bodyUsed` {{<type>}}boolean{{</type>}}
   - A boolean indicating if the body was used in the response.
-- `response.encodeBody` {{<type>}}string{{</type>}}
```
kflansburg, line 63: `workers-types` indicates that this is not actually a property of `Request`

content/workers/runtime-apis/response.md
```diff
@@ -39,16 +39,19 @@ let response = new Response(body, init);
 
 Valid options for the `options` object include: {{<definitions>}}
 
+<!-->Does `cf` do anything in a Response?<-->
```
kflansburg: Do we have more information on the possible values of `cf`?

himikof: Unfortunately, there indeed seems to be no documentation on what options are supported in `ResponseInit.cf` and what do they actually do.
Also, I'm not sure about `any | null` typing: the [C++ implementation in workerd](https://github.com/cloudflare/workerd/blob/bb8c8cd75e6a85dc88a1905553eb1ce4e1e58576/src/workerd/api/http.h#L1007-L1008) clearly expects `object | null`. Might be an issue of insufficient typing [here](https://github.com/cloudflare/workerd/blob/bb8c8cd75e6a85dc88a1905553eb1ce4e1e58576/src/workerd/api/http.h#L1015-L1018), leading to a problem in `workers-types`.

content/workers/runtime-apis/response.md
```diff
@@ -39,16 +39,19 @@ let response = new Response(body, init);
 
 Valid options for the `options` object include: {{<definitions>}}
 
+<!-- Does `cf` do anything in a Response? -->
```
irvinebroque: Not sure quite what you mean?

kflansburg: I don't know what configuring `cf` in the `Response` object does. What features are supported? We have this well documented for `Request`. Am I missing something?

irvinebroque: I'm not aware of any case where setting the value of the `cf` property on a response changes behavior

content/workers/runtime-apis/response.md
jasnell: I would see if we can handle documenting `cf` more adequately in a separate PR.

content/workers/runtime-apis/response.md
jasnell: That is correct. Setting the `cf` property on the `Response` is largely informational. Someone could take it, make modifications to it, and pass that on to a subsequent subrequest but setting the value when constructing the `Response` has no functional impact other than to set the value of the `response.cf` property.

jasnell (approved):

PR #14422, opened 2024-05-06
[D1] feat: add missing d1 `meta` object properties

hrueger:
Hi,
this adds docs about the missing properties to the `meta` object returned by `stmt.all()` and `db.batch()` d1 methods.
I'm not sure what `last_row_id` is exactly, can anyone help with that?

I think the information added here has quite some value - I did make a [pretty weird pr](https://github.com/prisma/prisma-engines/pull/4832) to another project because I did not know how to get the number of changed rows excluding indices (since there were no docs on that). Turns out - [it's pretty simple](https://github.com/prisma/prisma/pull/24083/commits/366b71dc3b891e9df2cd94b6a5d2278102373451#diff-27a4d9f7fa5e200bd36aa2ec1041e786e3e0c7ae705df17bbb0c474b60326f9c) 😅

Thanks for looking into this!

### DIFF
--- a/content/d1/build-with-d1/d1-client-api.md
+++ b/content/d1/build-with-d1/d1-client-api.md
@@ -79,7 +79,11 @@ The methods `stmt.all()` and `db.batch()` return a typed `D1Result` object that 
   meta: {
     duration: number, // duration of the operation in milliseconds
     rows_read: number, // the number of rows read (scanned) by this query
-    rows_written: number // the number of rows written by this query
+    rows_written: number, // the number of rows (including extra indices) written by this query
+    changes: number, // the number of data rows (excluding extra indices) changed by this query
+    changed_db: boolean, // whether the db was changed by this query
+    size_after: number, // the db size after executing this query
+    last_row_id: number // ToDo
   }
 }
 ```


content/d1/build-with-d1/d1-client-api.md
```diff
@@ -79,7 +79,11 @@ The methods `stmt.all()` and `db.batch()` return a typed `D1Result` object that
   meta: {
     duration: number, // duration of the operation in milliseconds
     rows_read: number, // the number of rows read (scanned) by this query
-    rows_written: number // the number of rows written by this query
+    rows_written: number, // the number of rows (including extra indices) written by this query
+    changes: number, // the number of data rows (excluding extra indices) changed by this query
+    changed_db: boolean, // whether the db was changed by this query
+    size_after: number, // the db size after executing this query
+    last_row_id: number // ToDo
```
vy-ton, line 86: @joshthoward user facing description for this field?

joshthoward: `meta` has type `Record<string, unknown>`. We shouldn't strongly type this in the docs.

hrueger: Hm. Is that expected to change frequently or what is the reason for it to be `Record<string, unknown>`? Big projects like `prisma` [use it](https://github.com/prisma/prisma/blob/271d834eb4a6d18b1caec8e70ba2107f1440d5dc/packages/adapter-d1/src/d1.ts#L83)...

joshthoward: @hrueger we wouldn't change existing fields from the response without gating the change behind a compatibility date. I want to keep our documentation and Typescript types in sync.

hrueger: That all makes perfect sense. Can I make a PR to update the types first, and after that this PR can be considered?

elithrar: Updating the types is still a breaking change.

hrueger: OK. What do you suggest?

KimJ15:

vy-ton:
@Oxyjun Can you ensure D1 results `meta` object is fully document in D1 docs?

PR #14450, opened 2024-05-07
[R2] Make wording of egress more clear in pricing.md

LiamSwayne:
Currently the R2 docs don't highlight that public R2 buckets with custom domains also do not have egress fees when listing the options that don't incur egress fees. I think it should be mentioned that custom domains can also be used with no egress fees in addition to `r2.dev` domains because `r2.dev` domains aren't recommend for production while custom domains connected to public R2 buckets are.


jonesphillip:
I think the current wording is more clear. If a user wants to connect other services to R2 (metered or otherwise), they would do so via a custom domain.

PR #14487, opened 2024-05-08
[Fundamentals] Abuse report obligations

patriciasantaana:


pedrosousa (approved):

pedrosousa:
Nit: I noticed that we have a section on [reporting abuse](https://developers.cloudflare.com/fundamentals/reference/report-abuse/) under Reference. 
Maybe these two types of content (reporting abuse and handling requests related to abuse reports) should live closer to one another?

PR #14490, opened 2024-05-08
Add TURN credential creation

third774:


third774:
Let's hold off on merging this, API might still change.

kodster28:
> Let's hold off on merging this, API might still change.

Converted to a draft just to give more indication around that

renandincer (approved):

third774:
Okay, I think we're good to go on this!

third774:
@kodster28 any thoughts/feedback before we merge?

PR #14493, opened 2024-05-09
docs: Updates postmark tutorial

### LINKED ISSUE #14467: [Workers] Postmark example issues

- https://developers.cloudflare.com/workers/tutorials/send-emails-with-postmark/

* Formatting
  * There is a newline at the start of the `fetch` handlers, it'd be good if we could run the examples through Prettier.

* Code issues
  * Whilst it is functional, `return Response.json(await response.json())` is redundant. It is parsing a JSON string into a JS object, and then stringifying that JS object back into a JSON string. It would be sufficient to just `return response`.
  
* Possible additions
  * It'd be cool to have an example of taking user input (query params, JSON body, etc) and including that in the emails, so users can have a Worker as an "email sending" API.

### Additional information

_No response_

megaconfidence:
Code change to the returned response block. Fixes #14467

### DIFF
--- a/content/workers/tutorials/send-emails-with-postmark/index.md
+++ b/content/workers/tutorials/send-emails-with-postmark/index.md
@@ -114,7 +114,7 @@ export default {
 			}),
 		});
 
-		return Response.json(await response.json());
+		return response;
 	},
 };
 ```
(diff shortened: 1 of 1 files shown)


kodster28 (approved):

kodster28:
Going to merge this in since it's addressing the clearest parts of the feedback provided.

PR #14495, opened 2024-05-09
[DNS] Add examples mapping to PTR records fields

### LINKED ISSUE #14257: Example of adding a DNS record of type PTR

https://developers.cloudflare.com/dns/additional-options/reverse-zones/

Clarification required for Step 2 - Add PTR records. This is what is says:
1. Go to DNS > Records.
2. For each IP within the prefix, add a PTR record using the least significant octet(s) as the subdomain.
IPv4 example:
Suppose you have the following configuration:
    Reverse zone: 100.51.198.in-addr.arpa
    IP address: 198.51.100.123
The subdomain for the PTR record would be 123, making the full domain for forward lookup 123.100.51.198.in-addr.arpa.

PROBLEM =  the step 2 above does NOT correspond to what appears as input fields when adding DNS record for Type PTR -- here below are the input fields that appear when adding a PTR record, please clarify what they mean, because they do not correspond to any information at the above URL.
Field: Type = PTR
Field: Name (required) = what should I input here ???
Field: Domain name (required) = what should I input here ???
Field: TTL = what value to select?

### Additional information

Screen Shot 2024-04-25 at 1 50 16 PM

RebeccaTamachiro:
PCX-10651.

closes #14257

### DIFF
--- a/content/dns/additional-options/reverse-zones.md
+++ b/content/dns/additional-options/reverse-zones.md
@@ -79,6 +79,12 @@ To set up a reverse zone, you need to create a reverse DNS zone and add PTR reco
 
 The subdomain for the PTR record would be `123`, making the full domain for forward lookup `123.100.51.198.in-addr.arpa`.
 
+{{<example>}}
+| Type | Name | Domain name | TTL |
+| --- | --- | --- | --- |
+| `PTR` | `123` | `example.com` | `Auto` |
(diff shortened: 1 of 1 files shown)


ghost:
Looks great, thanks!

pedrosousa (approved):

PR #14506, opened 2024-05-09
Clarify how Image transformations follow redirects

MattIPv4:
👋 This was somewhat unexpected behaviour which support confirmed to be true -- if the URL passed to /cdn-cgi/image returns a redirect it will be followed, even if it leaves the zone while `Resize images from any origin` is disabled. Figured this was worth adding an explicit note for, so it is documented behaviour.


kodster28 (approved):

kodster28:
Thanks for the contribution, @MattIPv4!

PR #12509, opened 2024-01-13
Added useful info when using WARP on a tightly-firewalled corporate network

jamie-sandbox:
Added useful info learned from hours of debugging and reverse-engineering when setting up WARP on a tightly-firewalled corporate network.

PR #14507, opened 2024-05-10
Add the unlisted endpoints to WARP with firewall

kyouheicf:
PCX-10848


ranbel:
@kkrum, here's another PR related to #12509

ranbel:
addressed in #14682

PR #14508, opened 2024-05-10
Add warning about polluting global scope with bindings derivatives

GregBrimble:


content/workers/runtime-apis/bindings/_index.md
```diff
@@ -40,4 +40,10 @@ export default {
...
\ No newline at end of file
+You can think of a binding as a permission and an API in one piece. With bindings, you never have to add secret keys or tokens to your Worker in order to access resources on your Cloudflare account — the permission is embedded within the API itself. The underlying secret is never exposed to your Worker's code, and therefore can't be accidentally leaked.
+
+## Making changes to bindings
+
+When creating a new Worker [version or deployment](/workers/configuration/versions-and-deployments/) and you only include changes to bindings (i.e. you don't change the code of the Worker), be aware that the any currently running Worker isolates may stay "hot" for a potentially long time if they continue to receive traffic. This is an optimization of our platform, allowing Workers to run with a dynamic set of bindings without needing to start a new isolate with every change to bindings. Broadly, this means you can expect better performance when rolling out non-code changes to your Worker.
```
irvinebroque:
```suggestion
When you deploy a change to your Worker, and only change its bindings (i.e. you don't change the Worker's code), Cloudflare may reuse existing isolates that are already running your Worker. This improves performance — you can change an environment variable or other binding without unnecessarily reloading your code.
```

content/workers/runtime-apis/bindings/_index.md
```diff
@@ -40,4 +40,10 @@ export default {
...
+
+However, this does mean that you must be careful when "polluting" global scope with derivatives of your bindings. For example, if you create an external client instance which uses a secret API key on `env`, you must ensure that you don't place this client instance in a global scope. If you do, the client instance might continue to exist despite making changes to the secret which is likely undesirable. Instead of polluting global scope, you should create a new client instance for each request, or, if you have more advanced needs, you may want to explore the [AsyncLocalStorage API](/workers/runtime-apis/nodejs/asynclocalstorage/) which provides another mechanism for exposing values down to child execution handlers.
```
irvinebroque:
```suggestion
This means you should not assign the values of bindings to variables in global scope. For example, you should not do this, because the client assigned to global scope can continue using the old value of `SECRET_KEY` even after you update it in your Worker:

let client;

export default {
  async fetch(request) {
    if (!client) { client = new ApiClient(env.SECRET_KEY) }
    // ...
  }
}

Instead, you should either create a new client instance for each request:

export default {
  async fetch(request) {
    let client = new ApiClient(env.SECRET_KEY)
    // ...
  }
}

Or use the [AsyncLocalStorage API](/workers/runtime-apis/nodejs/asynclocalstorage/):

...example

```

Think clearer with code?

irvinebroque (commented):
@kodster28 this is good addition from @GregBrimble — is a fun tricky one to explain well. taking a crack but eyes helpful

content/workers/runtime-apis/bindings/_index.md
mhart: Or even, store the client in a global `Map` where the key is the env value(s) or hash thereof  🧠 

Just on the `AsyncLocalStorage` note, that's still only going to be per-request, so you'd still be creating "a new client instance for each request". Just would want to make that clear if we were providing code for that.

content/workers/runtime-apis/bindings/_index.md
kodster28:
```suggestion
When a new [version or deployment](/workers/configuration/versions-and-deployments/) only changes bindings - and not the code itself - your Worker generally has better performance. Isolates run with a dynamic set of bindings, meaning they do not need to restart with every change to bindings. So if an isolate is "hot" - or already responding to traffic - the isolate usually remains "hot" after non-code changes. 
```

content/workers/runtime-apis/bindings/_index.md
kodster28: Maybe lead with what you should do, then have the shouldn't do?

kodster28:
```suggestion
As a result, you must be careful when "polluting" global scope with derivatives of your bindings. Anything you create there might continue to exist despite making changes to any underlying bindings. Consider an external client instance which uses a secret API key accessed from `env`: if you put this client instance in a global scope and then make changes to the secret, a client instance using the original value might continue to exist. The correct approach would be to create a new client instance for each request. If you have more advanced needs, explore the [AsyncLocalStorage API](/workers/runtime-apis/nodejs/asynclocalstorage/), which provides a mechanism for exposing values down to child execution handlers.
```

kodster28 (approved):

content/workers/runtime-apis/bindings/_index.md
GregBrimble: I think @irvinebroque's  suggestions here were slightly better, so I made a tiny tweak and went with that. If there's something you feel is important in yours, @kodster28, feel free to just re-suggest!

content/workers/runtime-apis/bindings/_index.md
```diff
@@ -40,4 +40,38 @@ export default {
 }
 ```
 
-You can think of a binding as a permission and an API in one piece. With bindings, you never have to add secret keys or tokens to your Worker in order to access resources on your Cloudflare account — the permission is embedded within the API itself. The underlying secret is never exposed to your Worker's code, and therefore can't be accidentally leaked.
\ No newline at end of file
+You can think of a binding as a permission and an API in one piece. With bindings, you never have to add secret keys or tokens to your Worker in order to access resources on your Cloudflare account — the permission is embedded within the API itself. The underlying secret is never exposed to your Worker's code, and therefore can't be accidentally leaked.
```
jasnell, line 43: Is this "never have to" actually true? Are there cloudflare services for which there is no binding currently? I just always get nervous when I see absolute statements like this in docs.

content/workers/runtime-apis/bindings/_index.md
```diff
@@ -40,4 +40,38 @@ export default {
...
+
+    // ... do things with `client`
+  }
+}
+```
+
+Compared to this alternative, which might have surprising and unwanted behavior:
+
+```ts
+let client = undefined;
```
jasnell, line 66: nit: I would add a comment here like: `// DON'T DO THIS` to make it extra clear

content/workers/runtime-apis/bindings/_index.md
```diff
@@ -40,4 +40,38 @@ export default {
...
+export default {
+  fetch(request, env) {
+    client ??= new Client(env.MY_SECRET); // `client` here might not be updated when `env.MY_SECRET` changes, since it may already exist in global scope
+
+    // ... do things with `client`
+  }
+}
+```
+
+If you have more advanced needs, explore the [AsyncLocalStorage API](/workers/runtime-apis/nodejs/asynclocalstorage/), which provides a mechanism for exposing values down to child execution handlers.
```
jasnell, line 77: Hmm.. I need to verify but is it possible that the environment/bindings can be reloading while a request is being processed? If for instance, the request is pending some async i/o, can the bindings be updated on the existing isolate while that i/o is pending? If it can, then even using `AsyncLocalStorage` to propagate the `client` in this example could potentially be problematic by propagating the binding as it existed before the update! The most reliable approach is likely to just access the binding in exactly the execution scope that it is used.

GregBrimble: We've seen people do stuff like:

```ts
let thingOnGlobal;

export default {
  fetch(request, env, ctx) {
    thingOnGlobal ??= Math.random();
    env.THING_ON_ENV ??= Math.random();
    return Response.json({ env, thingOnGlobal });
  }
}
```

Here, `env.THING_ON_ENV` behaves very similarly to `thingOnGlobal`. They'll both get initialized with a value when the first request comes in and it'll stay in-memory. The only difference is that `env.THING_ON_ENV` will change whenever *anything* on `env` changes, but `thingOnGlobal` will only change when the isolate is evicted.

I tried this:


```ts
let thingOnGlobal;

export default {
  fetch(request, env, ctx) {
    thingOnGlobal ??= Math.random();
    env.THING_ON_ENV ??= Math.random();
    const initial = JSON.stringify({ env, thingOnGlobal });
    await new Promise((resolve) => setTimeout(resolve, 10000));
    const after = JSON.stringify({ env, thingOnGlobal });
    return new Response(`initial: ${initial}\nafter: ${after}`);
  }
}
```

and despite deploying with a change on `env` in those 10 seconds, never saw any change to `env`'s reported value across `initial` and `after`. I even raced another request in the meantime (which did reflect the new values), and could never get disagreement between `initial` and `after`.

To me, this makes sense and would have been what I expected. I'd have been surprised if `env` did hot-reload with some new values magically. My mental model is that something "calls" my `fetch` handler with the `request`, `env` and `ctx` properties as they were at the time of the invocation. I don't personally think of them as hot-reloadable objects — it's just that the next time my handler gets called, they might be passed a new instance of `env` (or always a new instance of `request` and `ctx`).

`WorkerEntrypoint`s are maybe a bit different though, since they're only "called" with `request`. `env` and `ctx` exist as props on the class. I tried it out, and got the same behavior as regular handlers, but I think there's an argument that they could be hot-reloadable. We just don't seem to do that today.

GregBrimble:
Not perfect, but I'm just going to merge this as an improvement on what's there today and we can follow up with additional PRs if folks have strong opinions and good suggestions :)

PR #14509, opened 2024-05-10
fix LoaderFunction example on deploy-a-remix-site.md

tgfjt:
fix an example source.
https://developers.cloudflare.com/pages/framework-guides/deploy-a-remix-site/


dario-piotrowicz (approved):
@tgfjt thanks so very much for the fix! it looks great! ❤️ 🚀

PR #14520, opened 2024-05-10
[Workers] Document received WebSocket message size limit

jclee:
We mention this limit on the Durable Objects limits page, but it's probably not a bad idea to have it in the WebSocket docs, too, since it applies to any Worker using the WebSocket API, and it's an easy detail to miss.

Not sure if it also might warrant a mention on the Workers limits page as well.


irvinebroque (approved):
ty!

kodster28:
fine to merge in @jclee?

PR #14529, opened 2024-05-12
[queues] correct `DurableObject` to `DurableObjectNamespace` & remove constructor block

helloimalastair:
In the `Env` declaration for the [Use Queues with Durable Objects](https://developers.cloudflare.com/queues/examples/use-queues-with-durable-objects/) example, the `YOUR_DO_CLASS` binding should be an instance of `DurableObjectNamespace` instead of `DurableObject`.

Constructor block for the `YourDurableObject` class contained an extra `}`, causing the `fetch()` to not be recognized. This block has been removed entirely, and the `env` been made `public` to match `state`

### DIFF
--- a/content/queues/examples/use-queues-with-durable-objects.md
+++ b/content/queues/examples/use-queues-with-durable-objects.md
@@ -48,7 +48,7 @@ filename: src/index.ts
 ---
 interface Env {
   YOUR_QUEUE: Queue;
-  YOUR_DO_CLASS: DurableObject;
+  YOUR_DO_CLASS: DurableObjectNamespace;
 }
 
 export default {
--- a/content/queues/examples/use-queues-with-durable-objects.md
+++ b/content/queues/examples/use-queues-with-durable-objects.md
@@ -79,7 +79,7 @@ export default {
 }
 
 export class YourDurableObject implements DurableObject {
-  constructor(public state: DurableObjectState, public env: Env) {}
+  constructor(private state: DurableObjectState, private env: Env) {}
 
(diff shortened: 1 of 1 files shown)


garysassano:
Thanks for the PR.

There's a small snag with the `fetch()` call. Because it's within an if statement, TypeScript needs a return statement outside the if to cover cases where the condition isn't met. Otherwise, it throws a `Function lacks ending return statement and return type does not include 'undefined'` error, as the function might not always return the expected `Promise<Response>`.

helloimalastair:
@garysassano should be fixed, now returns a 400

Maddy-Cloudflare (approved):

Maddy-Cloudflare (approved):

content/queues/examples/use-queues-with-durable-objects.md
```diff
@@ -74,17 +74,12 @@ export default {
...
 
 export class YourDurableObject implements DurableObject {
-  constructor(public state: DurableObjectState, env: Env) {
-      this.state = state;
-      // Ensure you pass your bindings and environment variables into
-      // each Durable Object when it is initialized
-      this.env = env;
-    }
-  }
+  constructor(public state: DurableObjectState, public env: Env) {}
```
Maddy-Cloudflare:
```suggestion
  constructor(private state: DurableObjectState, private env: Env) {}
```

Maddy-Cloudflare (changes requested):
Better to change the properties to be private.

Maddy-Cloudflare (approved):

PR #14616, opened 2024-05-16
[Roughtime] Update install instruction for getroughtime

toksdotdev:
`go get` doesn't work outside of a go module. 

For a hello world example, it might help to keep the step simple by installing the `getroughtime` client (without setting up a module).

This PR updates the instruction to do that.


lukevalenta (approved):
Looks good--thanks for the documentation cleanup!

kodster28 (approved):

PR #14660, opened 2024-05-20
[Pages] Remove preamble from other framework guides

kodster28:


marciocloudflare (approved):

dario-piotrowicz (approved):

dario-piotrowicz:
@kodster28 given this sort of change, I wonder if we should also remove `deployment` from the next.js sub-pages: 
Screenshot at 2024-05-21 11-00-16

what do you think? :slightly_smiling_face:

PR #14693, opened 2024-05-22
Update storage-options.md

garysassano:
Removed duplicate, misplaced footnote.


kodster28 (approved):
Ty for the find / fix, @garysassano

PR #14724, opened 2024-05-23
Fix disable_flag for Workers RPC

maxwellpeterson:
- The disable_flag was incorrectly set to the same value as the enable_flag
- no_rpc is the correct disable_flag value


kodster28 (approved):
Validated at https://github.com/cloudflare/workerd/blob/b3c613b9a5e3265a7e468e7c62d1a14df1e057df/src/workerd/io/compatibility-date.capnp#L412

PR #14731, opened 2024-05-24
Update deploy-a-hugo-site.md

yaardie:
Hugo supports different content formats.


kodster28 (approved):
thanks for the catch and fix, @yaardie

PR #14753, opened 2024-05-25
docs(update): clarify wrangler-configuration.md

sambacha:
When running the command `npx wrangler pages download config` the user may be surprised that the `wrangler.toml` file that is generated contains sensitive access credentials, like for GitHub. Add a note stating this explicitly.


WalshyDev:
It will download your env vars + bindings. If you store things like access tokens as env vars and not secrets (aka visible and not encrypted) then this is expected.
You should always be storing these as secrets - https://developers.cloudflare.com/pages/functions/bindings/#secrets

PR #14764, opened 2024-05-27
early-hints: Update "any" -> "all"

kevinji:
From the context, it would make the most sense for **all** of the attributes to be present, rather than **any**.


GregBrimble (commented):
Just made this slightly more explicit with what actually happens: https://github.com/cloudflare/workers-sdk/blob/64ccdd6a6777c5fd85116af0d660cb3ee2e1de4d/packages/pages-shared/asset-server/handler.ts#L380-L404

GregBrimble (approved):

kevinji:
@GregBrimble anything blocking this from being merged?

GregBrimble:
Sorry, lost track of this. Thanks for following up!

PR #14791, opened 2024-05-28
[Workers Logpush] fix log message array limit to correct value

markjmiller:


kodster28 (approved):
Assuming you're able to sign off on the tech accuracy here, @markjmiller

PR #14827, opened 2024-05-29
Remove smart placement as a limitation for gradual deployments

WalshyDev:


kodster28 (approved):
Random, but one of the best parts about writing (and especially at CF) is that you outlast some of these warnings and limitations in the docs.

Not normal, lol, usually you assume they'll outlast you at the company.

PR #14831, opened 2024-05-29
[Docs Site] Stop command output being copied to clipboard

KianNH:
Improves the 'Copy to clipboard' functionality on `sh`/`shell` codeblocks.

These codeblocks do not allow the user to select comments, and command output, like so:

image

However, clicking the 'Copy to clipboard' button would include that text:

```
npx wrangler vectorize create tutorial-index --dimensions=3 --metric=cosine

✅ Successfully created index 'tutorial-index'

[[vectorize]]
binding = "VECTORIZE_INDEX"
index_name = "tutorial-index"
```

Now, it will only copy the command to the clipboard:

```
npx wrangler vectorize create tutorial-index --dimensions=3 --metric=cosine
```

### DIFF
--- a/bin/prism.config.ts
+++ b/bin/prism.config.ts
@@ -55,6 +55,11 @@ Prism.languages.sh = {
       },
     },
   },
+
+  output: {
+    pattern: /.(?:.*(?:[\r\n]|.$))*/,
+    alias: "unselectable"
+  }
 };
 
 const originalGrammar = Prism.languages.powershell;
--- a/components/CodeCopy.vue
+++ b/components/CodeCopy.vue
@@ -26,8 +26,12 @@ function copyCode(e: MouseEvent) {
     lines.forEach((line) =>
       textLines.push((line as HTMLElement).innerText.trimEnd())
     );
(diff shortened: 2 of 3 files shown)


kodster28 (approved):
Did some testing -- 20-ish `sh` blocks -- and looks good enough to me

KianNH:
Since the 'extra newlines at the end' logic affects all codeblock languages, I also tested that examples which have intended newlines (i.e between lines of code) still function normally & that the change only affects newlines at EOF.

Arguably there are some things like the C standard, or crontabs, which do need a newline at the end - but that isn't too relevant for our documentation and this serves as a quick win.

pedrosousa:
@KianNH - Maybe we can do the same change for PowerShell code blocks? Recently it was updated to match the `sh` behavior (hover, copy), but unlike `sh` we actually replaced the original grammar, so this would break multi-line commands currently using `powershell` syntax highlighting.

Currently we have only one multi-line PS command in https://developers.cloudflare.com/fundamentals/api/how-to/make-api-calls/#using-powershell (bottom of the page).

PR #14861, opened 2024-05-31
Update delete-objects.md

XabierCloudflare:
Perhaps, we can mention users to use a third party tools like https://duplicator.com/cloudflare-r2-backup/#:~:text=Cloudflare%20R2%20is%20an%20object,about%20any%20egress%20bandwidth%20fees.

I do not want to include any third party tool, but we can mention that they exist. This can save customers from having issues by deleting objects by mistake. There are several tickets on CSUP with this kind of situations.


elithrar:
We shouldn't recommend "third party tools" without being specific. I think it's fair to say that if you delete objects (or buckets), they're gone, as it would be with _any_ object storage.

PR #14884, opened 2024-06-03
[Docs Site] Upgrade mermaid to v10

### LINKED ISSUE #14093: [Docs Site] Resolve mermaid/cytoscape dependency issues
### Example URL(s)

N/A

### Actual Behavior

CI was failing at https://github.com/cloudflare/cloudflare-docs/actions/runs/8726706847/job/23942576800?pr=14034#step:7:3627, but was fixed in https://github.com/cloudflare/cloudflare-docs/pull/14092

### Expected Behavior

Mermaid patched cytoscape 3.28.1 but isn't pinned, so the release of cytoscape 3.29.0 breaks our CI setup (since the patch added exports into package.json which can't be found).

To fix this we added overrides in package.json but that isn't a long term fix.

This may be a breakage that mermaid need to fix upstream and then we need to upgrade (ideally to v10).

### Additional information

Whilst we have a `package-lock.json` in the repository, the CI job that failed uses `pnpm install` and we don't have a `pnpm-lock.yaml`. We should pick either `npm` or `pnpm`, use that universally and make sure we're leveraging lockfiles properly.

KianNH:
Closes https://github.com/cloudflare/cloudflare-docs/issues/14093

### DIFF
--- a/components/Mermaid.vue
+++ b/components/Mermaid.vue
@@ -1,22 +1,25 @@
 <script setup lang="ts">
 import { onMounted, ref } from "vue";
-import Mermaid from "mermaid";
+import mermaid from "mermaid";
 
 const props = defineProps({
   mermaid: String,
 });
 
-const svg = ref("");
+const element = ref("");
 const render = () => {
   const theme = document.documentElement.getAttribute("theme");
   const mermaidTheme = theme === "dark" ? "dark" : "neutral";
-  Mermaid.mermaidAPI.initialize({ theme: mermaidTheme, startOnLoad: false });
+  mermaid.initialize({ theme: mermaidTheme, startOnLoad: false });
 
-  Mermaid.mermaidAPI.render(
-    `mermaid-${crypto.randomUUID()}`,
-    decodeURIComponent(props.mermaid?.replaceAll("+", "%20") ?? ""),
-    (g: string) => (svg.value = g)
-  );
+  mermaid
+    .render(
+      `mermaid-${crypto.randomUUID()}`,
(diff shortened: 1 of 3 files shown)


components/Mermaid.vue
```diff
@@ -1,22 +1,25 @@
...
-    `mermaid-${crypto.randomUUID()}`,
-    decodeURIComponent(props.mermaid?.replaceAll("+", "%20") ?? ""),
-    (g: string) => (svg.value = g)
-  );
+  mermaid
+    .render(
+      `mermaid-${crypto.randomUUID()}`,
+      decodeURIComponent(props.mermaid?.replaceAll("+", "%20") ?? "")
+    )
+    .then(({ svg }) => {
```
KianNH, line 20: This is the only functional change, [`render` is now async](https://github.com/mermaid-js/mermaid/blob/master/CHANGELOG.md#mermaidrender-is-async-and-doesnt-accept-callbacks).

Everything else is naming or formatting changes.

package.json
```diff
@@ -22,14 +22,13 @@
...
     "instantsearch.js": "^4.64.0",
-    "mermaid": "9.4.3",
+    "mermaid": "^10.9.1",
     "node-html-parser": "6.1.6",
     "vue": "^3.2.45"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "4.20240208.0",
     "@types/glob": "^8.0.1",
-    "@types/mermaid": "^9.2.0",
```
KianNH, line 32: Mermaid provides types in the `mermaid` package now, the `@types/mermaid` package is deprecated.

kodster28 (approved):

PR #14901, opened 2024-06-03
Update _limits_increase.md to include feasibility

thomasgauvin:
We should include within the limits increase partial that all limit increases are subject to feasibility. For instance, in https://developers.cloudflare.com/kv/platform/limits/, the limit blurb gives the impression that all limits can be increased upon request (I've received such customer requests). As such, there should be some indication within the blurb indicating that it is possible that a limit cannot be increased to right-size expectations.

### DIFF
--- a/content/workers/_partials/_limits_increase.md
+++ b/content/workers/_partials/_limits_increase.md
@@ -7,6 +7,6 @@ _build:
 
 {{<Aside type="note" header="Increase your limits">}}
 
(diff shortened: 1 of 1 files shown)


irvinebroque (approved):

irvinebroque:
@tanushree-sharma since we were talking about this

tanushree-sharma (approved):

PR #14927, opened 2024-06-04
Update context.md

BilalBudhani:
The code snippet is referring to an undefined variable, this change should fix it


kodster28 (approved):
Thanks for the catch (and fix), @BilalBudhani

PR #14938, opened 2024-06-04
Update _create-hyperdrive-config.md to document variable in bash script

thomasgauvin:
We should document the variable $NAME currently used in our scripts in Hyperdrive docs. When following the tutorial, it's not written that this variable needs to be replaced with a desired name for the Hyperdrive configuration, only that the connection string should be replaced. As a result, you end up with an error in your script and have to debug/read the error to figure out what to change. Noting down the need to replace $NAME (which I propose renaming to <NAME_OF_HYPERDRIVE_CONFIG> in the absence of env var setting) would help avoid errors in the tutorial flow.

Screenshot 2024-06-04 at 12 39 33 PM

I am open to setting $NAME as a variable but that opens us up to the split behavior of setting env vars in Windows vs MacOS.

### DIFF
--- a/content/hyperdrive/_partials/_create-hyperdrive-config.md
+++ b/content/hyperdrive/_partials/_create-hyperdrive-config.md
@@ -20,10 +20,10 @@ postgres://USERNAME:PASSWORD@HOSTNAME_OR_IP_ADDRESS:PORT/database_name
 
 Most database providers will provide a connection string you can directly copy-and-paste directly into Hyperdrive.
 
-To create a Hyperdrive configuration with the [Wrangler CLI](/workers/wrangler/install-and-update/), open your terminal and run the following command, pasting the connection string provided from your database host, or replacing `user`, `password`, `HOSTNAME_OR_IP_ADDRESS`, `port`, and `database_name` placeholders with those specific to your database:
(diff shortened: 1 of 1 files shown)


thomasgauvin:
CC @kodster28

elithrar:
Note that `$NAME` is not an environmental variable. It is just a placeholder. We need a convention for placeholders.

Customers can't set a `$NAME` env var (what would it be used for anyway?).

KianNH:
> We need a convention for placeholders.

The closest we have is the one used on the [Wrangler commands page](https://developers.cloudflare.com/workers/wrangler/commands/#how-to-run-wrangler-commands)

`wrangler <COMMAND> <SUBCOMMAND> [PARAMETERS] [OPTIONS]`

i.e: `wrangler d1 create <DATABASE_NAME> [OPTIONS]`

kodster28:
Howdy @thomasgauvin, for now, we do have conventions around that specified in [our style guide](https://developers.cloudflare.com/style-guide/formatting/code-conventions-and-format/#angle-brackets---and--).

But do agree that more descriptive placeholders are better

thomasgauvin:
> Note that `$NAME` is not an environmental variable. It is just a placeholder. We need a convention for placeholders.
> 
> Customers can't set a `$NAME` env var (what would it be used for anyway?).

They can set a local $NAME env var, then use it in the wrangler command. If we anticipated further use of the env var (like in a deployment commands), then it may make sense, but I believe that is not the case which is why I propose the placeholder pattern

thomasgauvin:
@kodster28 can I get an approval on this PR? Seems like it's already following the convention (this is how other technical docs do it as well, which is why I adopted this pattern)

kodster28 (approved):

PR #14946, opened 2024-06-05
Remove the outdated note of early hints

xuhdev:
Early hints are now widely supported: https://caniuse.com/mdn-http_status_103

Remove the note to prevent users to underestimate the value of  early hints.


kodster28 (approved):

kodster28:
Thanks for the catch + fix, @xuhdev.

PR #14947, opened 2024-06-05
Calls: Add content on datachannels and turn

renandincer:


content/calls/datachannels.md
```diff
@@ -0,0 +1,13 @@
+---
+pcx_content_type: get-started
+title: DataChannels
+weight: 8
+---
+
+# DataChannels
+
+Since Cloudflare Calls is basically a pub/sub server for WebRTC that can scale up to many subscribers per publisher, it's fit for arbitrary data besides media too.
```
rcccf, line 9: Could we also make it explicit that initial implementation:

1. only supports `ordered` and `reliable` mode (sort of like websocket but over webrtc);
2. data sent back to a "subscribed" channel (aka a pulled channel) will be ignored

third774 (approved):
LGTM, but we should update to use "allowlist"

third774:
> LGTM, but we should update to use "allowlist"

Oh, wait — nevermind, looks like we're using both here?

renandincer:
Rachel suggestion will come in next update.

@third774: yeah I was trying to use both in at least 1 place for searchability

PR #14957, opened 2024-06-05
[Workers] Adds New Python Examples

megaconfidence:
This PR adds new Python Examples for:
- [Rewrite links](https://developers.cloudflare.com/workers/examples/rewrite-links/)
- [Set security headers](https://developers.cloudflare.com/workers/examples/security-headers/)
- [Turnstile with Workers](https://developers.cloudflare.com/workers/examples/turnstile-html-rewriter/)
- [Using the Cache API](https://developers.cloudflare.com/workers/examples/cache-api/)
- [Using timingSafeEqual](https://developers.cloudflare.com/workers/examples/protect-against-timing-attacks/)


megaconfidence:
Please can you do a technical review @hoodmane. TY!

kodster28 (approved):
Pending a tech review

content/workers/examples/cache-api.md
```diff
@@ -97,4 +98,39 @@ export default {
 ```
 
 {{</tab>}}
-{{</tabs>}}
+{{<tab label="py">}}
+
+```py
+from pyodide.ffi import create_proxy
+from js import Response, Request, fetch, URL, caches
```
hoodmane: Dumb nit: maybe sort these 
```suggestion
from js import Response, Request, URL, caches, fetch
```

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -53,3 +58,34 @@ export default {
...
+
+    if len(auth_token) != len(secret):
+        return Response.new("Unauthorized", status=401)
+
+    encoder = TextEncoder.new()
+    a = encoder.encode(auth_token)
+    b = encoder.encode(secret)
+
+    if a.byteLength != b.byteLength:
+        return Response.new("Unauthorized", status=401)
```
hoodmane: Per the advice in the diff directly above:

> Since the time taken to encode the values may reveal the length of our secret value, you should check the length of the strings before encoding.

```suggestion
    if a.byteLength != b.byteLength:
        return Response.new("Unauthorized", status=401)

    encoder = TextEncoder.new()
    a = encoder.encode(auth_token)
    b = encoder.encode(secret)
```

content/workers/examples/cache-api.md
```diff
@@ -97,4 +98,39 @@ export default {
...
+        # If not in cache, get it from origin
+        response = await fetch(request)
+        # Must use Response constructor to inherit all of response's fields
+        response = Response.new(response.body, response)
+
+        # Cache API respects Cache-Control headers. Setting s-max-age to 10
+        # will limit the response to be in cache for 10 seconds s-maxage
+        # Any changes made to the response here will be reflected in the cached value
+        response.headers.append("Cache-Control", "s-maxage=10")
+        ctx.waitUntil(create_proxy(cache.put(cache_key, response.clone())))
```
dom96, line 129: Why is the `create_proxy` required here? Doesn't `cache.put` return a JS object?

content/workers/examples/cache-api.md
```diff
@@ -97,4 +98,39 @@ export default {
...
 {{</tab>}}
-{{</tabs>}}
+{{<tab label="py">}}
+
+```py
+from pyodide.ffi import create_proxy
+from js import Response, Request, fetch, URL, caches
+
+async def on_fetch(request, _env, ctx):
+    cache_url = URL.new(request.url)
```
dom96: Instead of the JS URL, could/should we use the Python urllib?

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -27,7 +31,8 @@ interface Environment {
 
 export default {
   async fetch(req: Request, env: Environment) {
-    if (!env.MY_SECRET_VALUE) return new Response("Missing secret binding", { status: 500 });
+    if (!env.MY_SECRET_VALUE)
+      return new Response("Missing secret binding", { status: 500 });
```
dom96: Style nit, but I think it's good to be explicit:

```suggestion
    if (!env.MY_SECRET_VALUE) {
      return new Response("Missing secret binding", { status: 500 });
    }
```

content/workers/examples/rewrite-links.md
```diff
@@ -98,5 +98,35 @@ export default {
...
+
+    class AttributeRewriter:
+        def __init__(self, attr_name):
+            self.attr_name = attr_name
+        def element(self, element):
+            attr = element.getAttribute(self.attr_name)
+            if attr:
+                element.setAttribute(self.attr_name, attr.replace(old_url, new_url))
+
+    rewriter = HTMLRewriter.new().on("a", create_proxy(AttributeRewriter("href"))).on("img", create_proxy(AttributeRewriter("src")))
```
dom96: Nit: wrap line

content/workers/examples/rewrite-links.md
```diff
@@ -98,5 +98,35 @@ export default {
...
+    res = await fetch(request)
+    content_type = res.headers["Content-Type"]
+
+    # If the response is HTML, it can be transformed with
+    # HTMLRewriter -- otherwise, it should pass through
+    if "text/html" in content_type:
```
dom96: should this be checking for startsWith like the JS code?

content/workers/examples/security-headers.md
```diff
@@ -193,5 +194,59 @@ export default {
...
 
+{{</tab>}}
+{{<tab label="py">}}
+
+```py
+from js import Response, fetch, Headers
+
+async def on_fetch(request):
+    default_security_headers = {
+        #Secure your application with Content-Security-Policy headers.
```
dom96: Style nit:

```suggestion
        # Secure your application with Content-Security-Policy headers.
```

Same below for each comment

content/workers/examples/turnstile-html-rewriter.md
```diff
@@ -82,6 +83,35 @@ export default {
...
+    class AppendOnID:
+        def __init__(self, name):
+            self.name = name
+        def element(self, element):
+            # You are using the `getAttribute` method here to retrieve the `id` or `class` of an element
+            if element.getAttribute("id") == self.name:
+                element.append(f'<div class="cf-turnstile" data-sitekey="{site_key}" data-theme="light"></div>', { "html": True })
+
+    # Instantiate the API to run on specific elements, for example, `head`, `div`
+    new_res = HTMLRewriter.new().on("head", create_proxy(Append())).on("div", create_proxy(AppendOnID("NAME_OF_ATTRIBUTE"))).transform(res)
```
dom96: wrap line

content/workers/examples/turnstile-html-rewriter.md
```diff
@@ -82,6 +83,35 @@ export default {
...
+        def element(self, element):
+            element.append('<scriptsrc="https://challenges.cloudflare.com/turnstile/v0/api.js"async defer></script>', {"html": True})
+
+    class AppendOnID:
+        def __init__(self, name):
+            self.name = name
+        def element(self, element):
+            # You are using the `getAttribute` method here to retrieve the `id` or `class` of an element
+            if element.getAttribute("id") == self.name:
+                element.append(f'<div class="cf-turnstile" data-sitekey="{site_key}" data-theme="light"></div>', { "html": True })
```
dom96: wrap line

content/workers/examples/turnstile-html-rewriter.md
```diff
@@ -82,6 +83,35 @@ export default {
...
+from pyodide.ffi import create_proxy
+from js import HTMLRewriter, fetch
+
+async def on_fetch(request, env):
+    site_key = env.SITE_KEY
+    res = await fetch(request)
+
+    class Append:
+        def element(self, element):
+            element.append('<scriptsrc="https://challenges.cloudflare.com/turnstile/v0/api.js"async defer></script>', {"html": True})
```
dom96: wrap line

dom96:
```suggestion
            element.append('<scriptsrc="https://challenges.cloudflare.com/turnstile/v0/api.js" async defer></script>', {"html": True})
```

dom96 (approved):
Awesome to see these examples, thank you!

content/workers/examples/cache-api.md
hoodmane: It's a `Promise` but these get wrapped into `Future`s in case the return value is a Python object that needs memory management. But this case is slightly unfortunate, especially because this leaks the `Future`. To handle it, we need a way to wrap the `Future` into a `Promise` that gets resolved to the result of the `Future` and destroys the `Future` when the `Future` is resolved...

content/workers/examples/cache-api.md
hoodmane: In this case the only thing we do is convert it back to a string so we don't need either.

content/workers/examples/cache-api.md
```diff
@@ -97,4 +98,39 @@ export default {
...
+
+    # Construct the cache key from the cache URL
+    cache_key = Request.new(cache_url.toString(), request)
```
hoodmane: `cache_url.toString()` should be `str(cache_url)`. But we don't need to do this at all since we can just not make a `URL` in the first place.

content/workers/examples/rewrite-links.md
hoodmane: agreed
```suggestion
    if content_type.startswith("text/html"):
```

content/workers/examples/security-headers.md
hoodmane: Ideally should be applying formatters and linters to our docs examples. It's not entirely clear how to make this work with the code examples inline in markdown. Of course it's possible to manually copy to separate file, apply transform, and copy back, but it'd be nice to have something more systematic.

content/workers/examples/cache-api.md
dom96: oh, that is unfortunate

PR #14969, opened 2024-06-06
docs: Update transactional-storage-api.md

paulmelero:
Unless I'm missing something, the `url` variable was not needed.

Also,I took the chance to format one option as inline code.


KimJ15:

PR #14975, opened 2024-06-06
docs: clarify `persist` values

benmccann:
"not present or true" is confusing because I don't know if "not" applies only to "present" or also to "true"


dario-piotrowicz (approved):
Thanks for the quick improvement @benmccann 🙂👍

kodster28 (approved):

PR #14992, opened 2024-06-07
[Workers] Add a section to Wrangler configuration for top-level only keys

petebacondarwin:
Some keys in the inheritable section are not actually inheritable.


petebacondarwin (commented):
I didn't add the `migrations` top-level key to this section here because I hope to move it to inheritable soon.

kodster28 (approved):

PR #15002, opened 2024-06-07
[Workers AI] Update from API with schemas!

craigsdennis:
Update from new API that includes the JSON input and output schemas


content/workers-ai/models/bart-large-cnn.md
```diff
@@ -8,8 +8,7 @@ model:
     id: "6f4e65d8-da0f-40d2-9aa4-db582a5a04fd"
     name: "Summarization"
     description: "Summarization is the task of producing a shorter version of a document while preserving its important information. Some models can extract text from the original input, while other models can generate entirely new text."
-  tags:
-    - "summarization"
+  tags: []
```
craigsdennis, line 11: For reviewers. This is here temporarily as we phase out tags. It is in the API as a patch, and we will lean into that. This will eventually disappear when it does in the API.

content/workers-ai/models/meta-llama-3-8b-instruct.md
```diff
@@ -1,28 +1,20 @@
...
     description: "Family of generative text models, such as large language models (LLM), that can be adapted for a variety of natural language tasks."
   tags: []
-  properties:
-    - property_id: "beta"
-      value: "true"
-    - property_id: "info"
-      value: "https://llama.meta.com"
-    - property_id: "terms"
-      value: "https://llama.meta.com/llama3/license/#"
+  properties: []
```
craigsdennis: My concern here is that this is an alias and it doesn't have properties. Maybe there shouldn't be a model entry?

content/workers-ai/models/mistral-7b-instruct-v0.1-vllm.md
```diff
@@ -0,0 +1,28 @@
+---
+model:
```
mchenco: this model should not be in docs, we just can't have it experimental as its serving traffic (we basically made the -vllm serve traffic for the original -v0.1, which was triton based

craigsdennis: This is how it will behave docs gen wise, let's get this out so it's not tribal knowledge. If there is a new property we introduce I can filter, otherwise...Also in the API, dash and wrangler.

kodster28 (approved):

content/workers-ai/models/mistral-7b-instruct-v0.1-vllm.md
kodster28: Yeah, if we're committing to auto-gen, then we can't half-way it.

If it's set up that way in the API, then it'll appear in the docs and will be an ENG problem.

content/workers-ai/models/detr-resnet-50.md
```diff
@@ -1,25 +1,24 @@
 ---
 model:
-  id: "cc34ce52-3059-415f-9a48-12aa919d37ee"
+  id: "6ade40e1-1646-4013-8da0-95abb3360e33"
   source: 1
-  name: "@cf/facebook/detr-resnet-50"
-  description: "DEtection TRansformer (DETR) model trained end-to-end on COCO 2017 object detection (118k annotated images)."
+  name: "@cf/meta/detr-resnet-50"
```
craigsdennis: @mchenco making sure this is correct. (Looks like we have a new model with the potentially same name? ID is different than what was there)

craigsdennis: Got this fixed, there was a dupe!

PR #15014, opened 2024-06-07
Adds information on "script will never generate a response" errors

mikenomitch:
Open questions:
* Is this is best place for this information. It feels too small for its own page, but also maybe too big for this specific spot. Any thoughts on where to best put it. As long as we can link to this code, I'm happy wherever.
* Should we add information about how libraries sometimes stick Promises in global state and we can't rely on that. I'm not sure exactly how to phrase it... "Some code may be relying on global cross-request state to resolve a Promise in another request. Workers Requests cannot rely on cross-request Promise resolution, so Request-blocking Promises should only get created within their own Request". Seems like a mouthful, and I'm not 100% positive on the specifics TBH.

Notes:
* I think ideally the WebSocket error has a different message. In that case, we could break that out into its own code, but we can do that whenever that message is changed.

refs https://github.com/cloudflare/workers-sdk/issues/5433

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -1301,6 +1301,7 @@
 /workers/observability/logpush/ /workers/observability/logging/logpush/ 301
 /workers/platform/tail-workers/ /workers/observability/logging/tail-workers/ 301
 /workers/observability/tail-workers/  /workers/observability/logging/tail-workers/ 301
+/workers/errors/script-will-never-generate-a-response/  /workers/observability/errors/#the-script-will-never-generate-a-response-errors 301
 /workers/learning/metrics-and-analytics/ /workers/observability/metrics-and-analytics/ 301
 /workers/learning/integrations/databases/ /workers/databases/connecting-to-databases/ 301
 /workers/learning/how-routing-works/ /workers/platform/routing/ 301
--- a/content/workers/observability/errors.md
+++ b/content/workers/observability/errors.md
@@ -38,7 +38,7 @@ Some requests may return a 1101 error with `The script will never generate a res
 
 #### Cause 1: Unresolved Promises
 
-This is most commonly caused by relying on a Promise that is never resolved or rejected, in order to return a Response. To debug, look for Promises within your code or dependencies' code that block a Response, and ensure they are resolved or rejected.
+This is most commonly caused by relying on a Promise that is never resolved or rejected, which is required to return a Response. To debug, look for Promises within your code or dependencies' code that block a Response, and ensure they are resolved or rejected.
 
 In browsers and other JavaScript runtimes, equivalent code will hang indefinitely, leading to both bugs and memory leaks. The Workers runtime throws an explicit error to help you debug.
 
--- a/content/workers/observability/errors.md
+++ b/content/workers/observability/errors.md
@@ -32,6 +32,50 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
 
 A Worker cannot call itself or another Worker more than 16 times. In  order to prevent infinite loops between Workers, the [`CF-EW-Via`](/fundamentals/reference/http-request-headers/#cf-ew-via) header's value is an integer that indicates how many invocations are left. Every time a Worker is invoked, the integer will decrement by 1. If the count reaches zero, a [`1019`](#error-pages-generated-by-workers) error is returned.
 
+### "The script will never generate a response" errors
+
+Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned. This is most often a Response relying on a Promise that never gets resolved or rejected. To debug this problem, look for the creation of Promises within your own code or library code that block a Response, then and ensure their resolution. In other runtimes, this could result in indefinitely hanging code, but Workers error immediately to help with debugging.
+
+In the (admittedly contrived) example below, the Response relies on a Promise resolution that never happens. Uncommenting the `resolve` callback solves the issue.
+
+```js
+export default {
+	fetch(req) {
+		let response = new Response("Example response");
+		let { promise, resolve } = Promise.withResolvers();
+
+		// If this a resolve will never happen, the Workers runtime will
+		// recognize this and throw an error.
+
+		// setTimeout(resolve, 0)
+
+		return promise.then(() => response);
+	},
+};
+```
+
+A subset of these cases are related to open WebSocket connections. If a WebSocket is missing the proper code to close it's server-side connection, a `script will never generate a response` error is thrown. In the example below, the `'close'` event from the client is not properly handled with a `server.close()`, and the error is thrown. In order to avoid this, ensure that the WebSocket's server-side connection is properly closed via an event listener or other server-side logic.
+
+```js
+async function handleRequest(request) {
+	let webSocketPair = new WebSocketPair();
+	let [client, server] = Object.values(webSocketPair);
+	server.accept();
+
+	server.addEventListener('close', () => {
+    // This missing line would keep a WebSocket connection open indefinitely
+    // and results in "The script will never generate a response" errors
+
+		// server.close();
+	});
+
+	return new Response(null, {
+		status: 101,
+		webSocket: client,
+	});
+}
+```
+
 ## Errors on Worker upload
 These errors occur when a Worker is uploaded or modified.
 
--- a/content/workers/observability/errors.md
+++ b/content/workers/observability/errors.md
@@ -34,45 +34,61 @@ A Worker cannot call itself or another Worker more than 16 times. In  order to p
 
 ### "The script will never generate a response" errors
 
-Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned. This is most often a Response relying on a Promise that never gets resolved or rejected. To debug this problem, look for the creation of Promises within your own code or library code that block a Response, then and ensure their resolution. In other runtimes, this could result in indefinitely hanging code, but Workers error immediately to help with debugging.
+Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the Workers runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned.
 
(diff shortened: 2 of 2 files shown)


content/_redirects
```diff
@@ -1290,6 +1290,7 @@
 /workers/observability/logpush/ /workers/observability/logging/logpush/ 301
 /workers/platform/tail-workers/ /workers/observability/logging/tail-workers/ 301
 /workers/observability/tail-workers/  /workers/observability/logging/tail-workers/ 301
+/workers/errors/script-will-never-generate-a-response/  /workers/observability/errors/#the-script-will-never-generate-a-response-errors 301
```
mikenomitch, line 1304: This is to make a slightly nicer URL for error logs. Not 100% necessary, but seems nice to have a slightly easier reference link

content/workers/observability/errors.md
```diff
@@ -36,8 +36,52 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
...
 
+### "The script will never generate a response" errors
+
+Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned. This is most often a Response relying on a Promise that never gets resolved or rejected. To debug this problem, look for the creation of Promises within your own code or library code that block a Response, then and ensure their resolution. In other runtimes, this could result in indefinitely hanging code, but Workers error immediately to help with debugging.
+
+In the (admittedly contrived) example below, the Response relies on a Promise resolution that never happens. Uncommenting the `resolve` callback solves the issue.
```
mikenomitch: Debated adding "(admittedly contrived)" or not. I think it gives the docs some flavor/life, but understand if that's not the tone we want.

kodster28:
Two thoughts here:

1. Might make sense to reference w/in the table that this is a permutation of `1101` at [this line](https://github.com/cloudflare/cloudflare-docs/pull/15014/files#diff-07d64db4d448224bc0d6a0ab84c688c12f7c8ccbbf58af6ff01c152a6cba7b6dR20).
2. A helpful question to ask about the "new page vs section" debate is whether you can foresee folks grabbing this error message and copy/pasting into Google (which it seems like they would). If I'm making a correct assumption here, then you'd be better off with a new page b/c that has higher implicit authority than a section.

content/workers/observability/errors.md
```diff
@@ -36,8 +36,52 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
 
 A Worker cannot call itself or another Worker more than 16 times. In  order to prevent infinite loops between Workers, the [`CF-EW-Via`](/fundamentals/reference/http-request-headers/#cf-ew-via) header's value is an integer that indicates how many invocations are left. Every time a Worker is invoked, the integer will decrement by 1. If the count reaches zero, a [`1019`](#error-pages-generated-by-workers) error is returned.
 
+### "The script will never generate a response" errors
+
+Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned. This is most often a Response relying on a Promise that never gets resolved or rejected. To debug this problem, look for the creation of Promises within your own code or library code that block a Response, then and ensure their resolution. In other runtimes, this could result in indefinitely hanging code, but Workers error immediately to help with debugging.
```
irvinebroque:
```suggestion
Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the Workers runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned.

#### Cause 1: Unresolved Promises

This is most commonly caused by relying on a Promise that is never resolved or rejected, in order to return a Response. To debug, look for Promises within your code or dependencies' code that block a Response, and ensure they are resolved or rejected.

In browsers and other JavaScript runtimes, equivalent code will hang indefinitely, leading to both bugs and memory leaks. The Workers runtime throws an explicit error to help you debug.
```

content/workers/observability/errors.md
irvinebroque:
```suggestion
In the example below, the Response relies on a Promise resolution that never happens. Uncommenting the `resolve` callback solves the issue.
```

content/workers/observability/errors.md
```diff
@@ -36,8 +36,52 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
...
+
+In the (admittedly contrived) example below, the Response relies on a Promise resolution that never happens. Uncommenting the `resolve` callback solves the issue.
+
+```js
+export default {
+	fetch(req) {
+		let response = new Response("Example response");
+		let { promise, resolve } = Promise.withResolvers();
+
+		// If this a resolve will never happen, the Workers runtime will
```
irvinebroque:
```suggestion
		// If the promise is not resolved, the Workers runtime will
```

content/workers/observability/errors.md
```diff
@@ -36,8 +36,52 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
...
+		// recognize this and throw an error.
+
+		// setTimeout(resolve, 0)
```
irvinebroque: You can highlight a specific line in a code block, maybe worth doing here.

content/workers/observability/errors.md
```diff
@@ -36,8 +36,52 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
...
+
+		return promise.then(() => response);
+	},
+};
+```
+
+A subset of these cases are related to open WebSocket connections. If a WebSocket is missing the proper code to close it's server-side connection, a `script will never generate a response` error is thrown. In the example below, the `'close'` event from the client is not properly handled with a `server.close()`, and the error is thrown. In order to avoid this, ensure that the WebSocket's server-side connection is properly closed via an event listener or other server-side logic.
```
irvinebroque:
```suggestion
#### Cause 2: WebSocket connections that are never closed

If a WebSocket is missing the proper code to close it's server-side connection, the Workers runtime will throw a `script will never generate a response` error. In the example below, the `'close'` event from the client is not properly handled by calling `server.close()`, and the error is thrown. In order to avoid this, ensure that the WebSocket's server-side connection is properly closed via an event listener or other server-side logic.
```

irvinebroque (approved):

content/workers/observability/errors.md
isaac-mcfadyen: Also small typo fix: `it's` -> `its`

irvinebroque:
Ship it?

irvinebroque (approved):
Content looks great

should it live in this section? https://43dfd917.cloudflare-docs-7ou.pages.dev/workers/observability/errors/#runtime-errors

mikenomitch:
@kodster28 I didn't do either of your suggestions:
1. I think it might be too much detail to mention this at the top level in that table
2. I think we're okay with this on this one page. We'll be linking to this section via the error messages, so I think we should be good enough on discoverability. If we want to eventually break it off into its own thing, that seems okay to me, but I don't want to mess w the overall page arch too much.

If you're okay w that, any chance you can +1 so I can merge this? 🤞

content/_redirects
kodster28: Technically, there's also a small SEO benefit to better URLs

content/workers/observability/errors.md
```diff
@@ -32,6 +32,66 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
 
 A Worker cannot call itself or another Worker more than 16 times. In  order to prevent infinite loops between Workers, the [`CF-EW-Via`](/fundamentals/reference/http-request-headers/#cf-ew-via) header's value is an integer that indicates how many invocations are left. Every time a Worker is invoked, the integer will decrement by 1. If the count reaches zero, a [`1019`](#error-pages-generated-by-workers) error is returned.
 
+### "The script will never generate a response" errors
+
+Some requests may return a 1101 error with `The script will never generate a response` in the error message. This occurs when the Workers runtime detects that all the code associated with the request has executed and no events are left in the event loop, but a Response has not been returned.
+
+#### Cause 1: Unresolved Promises
+
+This is most commonly caused by relying on a Promise that is never resolved or rejected, in order to return a Response. To debug, look for Promises within your code or dependencies' code that block a Response, and ensure they are resolved or rejected.
```
kodster28:
```suggestion
This is most commonly caused by relying on a Promise that is never resolved or rejected, which is required to return a Response. To debug, look for Promises within your code or dependencies' code that block a Response, and ensure they are resolved or rejected.
```

kodster28 (approved):
One comment and small suggestion.

content/workers/observability/errors.md
```diff
@@ -32,6 +32,66 @@ Other `11xx` errors generally indicate a problem with the Workers runtime itself
...
+---
+filename: index.js
+highlight: [10]
+---
+async function handleRequest(request) {
+  let webSocketPair = new WebSocketPair();
+  let [client, server] = Object.values(webSocketPair);
+  server.accept();
+
+  server.addEventListener('close', () => {
```
RamIdeas, line 81: Is this correct? The paragraph above implies this should be `client.addEventListener`.

Maybe this section needs a once over as it's not very easy to understand cc @irvinebroque @petebacondarwin

Quicksilver0218: It doesn't work. I've tried this code and the error still exists.
```js
server.addEventListener("close", () => {
  console.log("This function will be run, but the error still exists.");
  server.close();
  client.close();
});
client.addEventListener("close", () => {
  console.log("This function will not be run.");
  server.close();
  client.close();
});
```

Quicksilver0218: I have tried the code but the error still exists. It seems that there are no ways to catch the error when using WebSocket. Here is my code for reproduction.

Worker:
```ts
export default {
        if (request.headers.get("Upgrade") !== "websocket")
            return new Response("Expected Upgrade: websocket", { status: 426 });
        const { 0: client, 1: server } = new WebSocketPair();
        server.accept();
        server.addEventListener("close", event => {
            console.log("Close event triggered");
            // All lines below are useless
            server.close();
            client.close();
            event.preventDefault();
            event.stopImmediatePropagation();
        });
        server.addEventListener("error", event => console.log("Never triggered"));
        client.addEventListener("close", event => console.log("Never triggered"));
        client.addEventListener("error", event => console.log("Never triggered"));
        return new Response(null, { status: 101, webSocket: client, });
    },
```

Node.js Client:
```js
new WebSocket("ws://localhost:8787").onopen = () => process.exit();
```
P.S. I have only tested with `wrangler dev`.

PR #15025, opened 2024-06-10
[R2] Add links to all presigned URL examples

KianNH:
Adds links to all of the examples that contain a presigned URL section based on Discord feedback (we have some SDK-specific instructions on their examples page but don't lead users to them).

image

cc @kodster28 for shortcode changes

### DIFF
--- a/content/r2/api/s3/presigned-urls.md
+++ b/content/r2/api/s3/presigned-urls.md
@@ -56,11 +56,7 @@ R2 currently supports the following methods when generating a presigned URL:
 
 Generate a presigned URL by referring to the following examples:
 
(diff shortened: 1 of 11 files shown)


kodster28 (approved):
@KianNH, mind adding a `show_descriptions=true` to the various resources on the `/use-cases/ai/` page?

kodster28 (commented):
Might be worth doing before merged or maybe not.... but either way, we'll likely want to circle back around on this... the tag should probably be `Presigned URLs` and then we'd maybe want to add a separate filter on the component to just pull those tags from the `R2` folder.

PR #15027, opened 2024-06-10
Update cron-trigger.md

alejandro-pigeon:


kodster28 (changes requested):
Howdy @openai-research, you're introducing a typo instead of fixing one?

PR #15062, opened 2024-06-12
[Docs Site] Tab Switcher bug fixes

KianNH:
Fixes:
  * issue when `default` tabs contained a space since we wasn't wrapping the selector in quotes
  * issue when the `label` on a `tab` wasn't lower-cased (we `lower`'d the labels in `tabs` but not in `tab`, meaning the IDs were different case)
  * issue with `no-code` labels adding an `active` class (we handle this in client-side JS)
  
Tested on:
  * [js-ts-py labels](https://kian-pcx-11272.cloudflare-docs-7ou.pages.dev/workers/examples/return-html/)
  * [no-code labels](https://kian-pcx-11272.cloudflare-docs-7ou.pages.dev/ai-gateway/configuration/caching/)
  * [spaces + punctuation in labels](https://kian-pcx-11272.cloudflare-docs-7ou.pages.dev/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/switch-organizations#switch-organizations-in-warp)
  * [tabs set as explicit default](https://kian-pcx-11272.cloudflare-docs-7ou.pages.dev/hyperdrive/configuration/connect-to-postgres/#postgresjs)

### DIFF
--- a/assets/events.ts
+++ b/assets/events.ts
@@ -130,7 +130,7 @@ export function tabs() {
           const tabId = parts.slice(0, parts.length - 1).join("-");
 
           const defaultTabLabel = wrappers[i].querySelector(
-            `a[data-link=${tabId}]`
+            `a[data-link="${tabId}"]`
           );
 
           (defaultTab as HTMLElement).style.display = "block";
--- a/layouts/shortcodes/tabs.html
+++ b/layouts/shortcodes/tabs.html
@@ -1,13 +1,16 @@
-{{- $text := .Get "labels" -}} {{- $labels := split $text " | " -}}
-{{ $unique_id := substr (md5 .Inner) 0 16 }}
+{{- $text := .Get "labels" -}}
+{{- $labels := split $text " | " -}}
+{{- $unique_id := lower (substr (md5 .Inner) 0 16) -}}
 
 <div class="tabs-wrapper" tab-wrapper-id="{{$unique_id}}">
 <nav role="navigation" class="tabs">
 <ul class="tab-active" block-id="{{$unique_id}}">
-{{ range $idx, $txt := $labels }} {{- $link := replace $txt "/" "-"}}
+{{ range $idx, $txt := $labels }}
+{{- $link := replace $txt "/" "-" -}}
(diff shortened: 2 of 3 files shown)


layouts/shortcodes/tabs.html
```diff
@@ -106,7 +109,7 @@
 </a>
 
 {{ else }}
-<a class="tab-label noCodeLabel {{ if eq $idx 0 }}active{{ end }}" href="#" data-link="tab-{{lower $link}}" data-id="{{lower $unique_id}}">
```
KianNH, line 109: `{{ if eq $idx 0 }}active{{ end }}` conflicts with adding a `default="true"` label that isn't the first label (it highlighted both as active). We already do `if idx === 0 then active` logic in JS.

layouts/shortcodes/tabs.html
```diff
@@ -1,13 +1,16 @@
...
+{{- $labels := split $text " | " -}}
+{{- $unique_id := lower (substr (md5 .Inner) 0 16) -}}
 
 <div class="tabs-wrapper" tab-wrapper-id="{{$unique_id}}">
 <nav role="navigation" class="tabs">
 <ul class="tab-active" block-id="{{$unique_id}}">
-{{ range $idx, $txt := $labels }} {{- $link := replace $txt "/" "-"}}
+{{ range $idx, $txt := $labels }}
+{{- $link := replace $txt "/" "-" -}}
+{{- $dataLink := printf "tab-%s" (lower $link) }}
```
KianNH, line 10: We lowercase IDs across the board for simplicity which means `label` on `tab` can now be the same casing, or different, to the `labels` on the `tabs`.

assets/events.ts
```diff
@@ -130,7 +130,7 @@ export function tabs() {
           const tabId = parts.slice(0, parts.length - 1).join("-");
 
           const defaultTabLabel = wrappers[i].querySelector(
-            `a[data-link=${tabId}]`
+            `a[data-link="${tabId}"]`
```
KianNH, line 133: This is since labels can contain spaces, like `rest api`.

kodster28 (approved):

PR #15071, opened 2024-06-13
[DNS] Clarify Multi-signer as the only DNSSEC option for outgoing zone transfers

dom-cf:
Currently the docs do not state which DNSSEC options are compatible with outgoing zone transfers.


ghost (approved):

RebeccaTamachiro:
Raised PCX-11428 and waiting a bit to merge this as there was a new reply on the original thread.

RebeccaTamachiro:
Thanks again for flagging this! Moving to https://github.com/cloudflare/cloudflare-docs/pull/15926

PR #15080, opened 2024-06-13
06-12 release notes for workers

garrettgu10:


data/changelogs/workers.yaml
```diff
@@ -5,6 +5,9 @@ productLink: "/workers/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+- publish_date: '2024-06-12'
+  description: |-
+    - Blob and Body objects now include a new `bytes()` method.
```
irvinebroque: can we clarify that this is part of standard and link to it?

garrettgu10: I don't think this is standard? Maybe @jasnell knows more about this

irvinebroque: if not standard then we need link and docs

KianNH: Seems to be a recent addition to the two standards: https://github.com/whatwg/fetch/commit/1085f4fbfeebda8d4fa5efcb6c5c7350953e524b & https://github.com/w3c/FileAPI/commit/85b18bc93d934bfafe0c46ad285adc457f5017df

ObsidianMinor (approved):

irvinebroque (approved):

PR #15081, opened 2024-06-13
Document the new compatibilityFlags query API

jasnell:
Preparing to land in the internal repo. Should be in production within the next few days


irvinebroque (approved):
Curious - does the API have TypeScript types, where the valid compat flags are represented as their own type?

jasnell:
No, we would need to define it. I think if we defined the type as `Record<string, boolean>` that would be sufficient and we wouldn't need to define a type for each individual flag. Should defer to folks who are more familiar with how the type definition system works tho.

KimJ15:

PR #15096, opened 2024-06-14
[ZT] WARP glossary entries

ranbel:


data/glossary/cloudflare-one.yaml
```diff
@@ -63,6 +67,14 @@ entries:
   associated_products:
     - DDoS Protection
 
+- term: device posture
+  general_definition: |-
+    evaluates how secure a device is by checking its serial number, OS version, installed certificates, or other attributes.
```
crwaters16: only add that doesn't lead with a noun

crwaters16 (approved):

PR #15112, opened 2024-06-15
Add alternative WARP download links

thibault-ml:
Microsoft AppCenter is down, so add alternative WARP download links from one.one.one.one


thibault-ml:
Not needed anymore, MS AppCenter is back up

PR #15124, opened 2024-06-17
[RPC] documentation update on fully consuming Response streams before use

draphy:
I have added a note to the documentation emphasizing the importance of fully consuming Response objects using methods like .text(), .json(), or .blob() before logging or further utilization. This ensures the response stream is not closed prematurely, particularly in asynchronous contexts such as queue processing.

issue: https://discord.com/channels/595317990191398933/1247472256267980811

tested: https://github.com/draphy/cloudflare-rpc-test

### DIFF
--- a/content/workers/runtime-apis/rpc/_index.md
+++ b/content/workers/runtime-apis/rpc/_index.md
@@ -218,9 +218,9 @@ Only [byte-oriented streams](https://developer.mozilla.org/en-US/docs/Web/API/St
 
 In all cases, ownership of the stream is transferred to the recipient. The sender can no longer read/write the stream after sending it. If the sender wishes to keep its own copy, it can use the [`tee()` method of `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee) or the [`clone()` method of `Request` or `Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone). Keep in mind that doing this may force the system to buffer bytes and lose the benefits of flow control.
 
-### Note:
-
+{{<Aside type="note" description="Fully consume response objects">}}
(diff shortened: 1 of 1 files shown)


content/workers/runtime-apis/rpc/_index.md
```diff
@@ -218,6 +218,10 @@ Only [byte-oriented streams](https://developer.mozilla.org/en-US/docs/Web/API/St
 
 In all cases, ownership of the stream is transferred to the recipient. The sender can no longer read/write the stream after sending it. If the sender wishes to keep its own copy, it can use the [`tee()` method of `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee) or the [`clone()` method of `Request` or `Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone). Keep in mind that doing this may force the system to buffer bytes and lose the benefits of flow control.
 
+{{<Aside type="note" description="Fully consume response objects">}}
+When handling `Response` objects, fully consume the response using `.text()`, `.json()`, or `.blob()` before logging or utilizing it further. This ensures the stream is not closed prematurely in asynchronous contexts like queue processing.
```
jasnell, line 222: hmm... I'm not entirely sure about this. In some scenarios even calling `.text()` or the other methods would not necessarily be enough to stop the stream from being closed. It's unlikely, but not guaranteed. There are also plenty of reasons why you may want to defer calling these as they end up fully consuming the stream.

draphy: @jasnell  Thank you for the feedback. I understand your concerns about the potential limitations and edge cases. You make a valid point that fully consuming the response may not be a guaranteed solution in all scenarios, and deferring consumption could be preferred in certain contexts for performance reasons. But in my case in an asynchronous context, this solution works fine, but yeah it's not universal, but it's worth it in my case for now. I am also looking forward to a better solution if any. With your insights, maybe I can update the documentation to provide more nuanced guidance addressing the potential trade-offs and limitations, rather than presenting it as a universal best practice.

KimJ15:

PR #15127, opened 2024-06-17
[Workers] Adds New Rust Examples

megaconfidence:
Adds Rust examples for 
- [Redirect](https://developers.cloudflare.com/workers/examples/redirect)
- [Read post](https://developers.cloudflare.com/workers/examples/read-post)


kodster28 (approved):
approved pending tech review

kodster28:
@megaconfidence, mind also adding in a page similar to https://github.com/cloudflare/cloudflare-docs/blob/production/content/workers/languages/typescript/examples.md within the Rust section itself (and filtered to `Rust`) that'll help highlight the examples you're adding?

irvinebroque (commented):
@kflansburg @zebp may have good feedback here

content/workers/examples/redirect.md
```diff
@@ -49,12 +50,27 @@ def on_fetch(request):
...
+{{<tab label="rs">}}
+
+```rs
+use worker::*;
+
+#[event(fetch)]
+async fn fetch(_req: Request, _env: Env, _ctx: Context) -> Result<Response> {
+    let destination_url = Url::parse("https://example.com")?;
+    let status_code = 301;
+    Response::redirect_with_status(destination_url, status_code)
```
kflansburg, line 63: It might be slightly more idiomatic to use `TryFrom`. 
```suggestion
    Response::redirect_with_status("https://example.com".try_into()?, status_code)
```

content/workers/examples/read-post.md
```diff
@@ -172,4 +173,46 @@ async def on_fetch(request):
...
+fn raw_html_response(html: &str) -> Result<Response> {
+    Response::from_html(html)
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+struct Payload {
+    msg: String,
+}
+
+async fn read_request_body(mut req: Request) -> String {
```
kflansburg, line 191: I think it would probably be best to avoid recommending that people use `unwrap`. We should be able to have this return `Result<String>` and use `?` in most cases here.

content/workers/examples/read-post.md
```diff
@@ -172,4 +173,46 @@ async def on_fetch(request):
...
 
 {{</tab>}}
-{{</tabs>}}
+{{<tab label="rs">}}
+
+```rs
+use serde::{Deserialize, Serialize};
+use worker::*;
+
+fn raw_html_response(html: &str) -> Result<Response> {
```
kflansburg, line 182: This method seems kind of unnecessary

kflansburg (commented):
This is a little bit tricky because `workers-rs` now has the `http` compiler option which uses different request / response types. Would it be easy to have Rust (http) as a variant here?

content/workers/examples/redirect.md
megaconfidence: That works, but wanted to keep the Rust implementation identical to the JS/TS/PY examples

content/workers/examples/read-post.md
megaconfidence: Totally agree. Only there for consistency with the other implementations

content/workers/examples/read-post.md
megaconfidence: Yes. I guess you're referring to the headers on the next line 
```rs
 let ctype = req.headers().get("content-type").unwrap().unwrap();
```

PR #15164, opened 2024-06-18
Update Wrangler KV commands in docs to remove ':'

thomasgauvin:


content/kv/reference/kv-commands.md
```diff
@@ -1,23 +1,23 @@
...
 
-Manage KV namespaces.
+## `kv namespace`
 
-The `kv:...` commands allow you to manage application data in the Cloudflare network to be accessed from Workers using KV.
+Manage KV namespaces.
 
 ### create
 
 Creates a new KV namespace.
```
elithrar, line 23: You need to specify which wrangler version no longer requires `kv:<THING>` - since users often have older versions of wrangler.

petebacondarwin: And then one must explain what the old command format is for those older versions. Perhaps in a legacy section at the bottom of the page?

thomasgauvin: Thoughts on added deprecation notices?

petebacondarwin (approved):

kodster28 (approved):

kodster28:
FYI, pushed up some changes to address the conflicts introduced by merging in #15309

PR #15202, opened 2024-06-19
aig adding custom cache key

kathayl:
AIG - adding new header for custom cache key

### DIFF
--- a/content/ai-gateway/configuration/caching.md
+++ b/content/ai-gateway/configuration/caching.md
@@ -113,9 +113,9 @@ curl https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/openai/chat/
 
 ### Custom cache key (cf-aig-cache-key)
 
-In order to override the default cache key, you can use the header **cf-aig-cache-key**. 
+Custom cache keys let you override the default cache key in order to precisely set the cacheability setting for any resource. To override the default cache key, you can use the header **cf-aig-cache-key**. 
 
-The first time you use the **cf-aig-cache-key** header, you will get a response from the provider. Then, each subsequent time the header is used, you will receive the cached response. If the **cf-cache-ttl** header is used, responses will be cached according to Cache Time To Live. However, if no **cf-cache-ttl** header is used, responses will be cached according to dashboard settings. If caching is not turned on for the gateway, responses will be cached for 5 minutes by default.
+When you use the **cf-aig-cache-key** header for the first time, you will receive a response from the provider. Subsequent requests with the same header will return the cached response. If the **cf-cache-ttl** header is used, responses will be cached according to the specified Cache Time To Live. Otherwise, responses will be cached according to the cache settings in the dashboard. If caching is not enabled for the gateway, responses will be cached for 5 minutes by default.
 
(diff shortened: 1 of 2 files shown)


content/ai-gateway/configuration/caching.md
```diff
@@ -110,3 +110,32 @@ curl https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/openai/chat/
    	 }
 '
 ```
+
+### Custom cache key (cf-aig-cache-key)
+
```
mchenco, line 115: can we get more info on why someone would want to set a custom cache key / why it matters?

content/ai-gateway/configuration/caching.md
```diff
@@ -110,3 +110,32 @@ curl https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/openai/chat/
...
+In order to override the default cache key, you can use the header **cf-aig-cache-key**. 
+
+The first time you use the header, you will get a response from the provider. Then, each subsequent time the header is used, you will receive the cached response. If no **cf-cache-ttl header** is used, responses will be cached for 5 minutes. If the **cf-cache-ttl header** is used, responses will be cached according to Cache TTL.
```
kodster28: Is `if no **cf-cache-ttl** header is used` part accurate? So setting the cache key overrides your default caching settings regardless?

content/ai-gateway/configuration/caching.md
```diff
@@ -110,3 +110,32 @@ curl https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/openai/chat/
...
+
+As an example, when submitting a request to OpenAI, include the header in the following manner:
+
+```bash
+---
+header: Request with custom cache key
+---
+
+curl https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/openai/chat/completions \
+  --header 'Authorization: Bearer $TOKEN' \
```
kodster28:
```suggestion
  --header 'Authorization: Bearer {openai_token}' \
```

kodster28 (approved):

content/ai-gateway/configuration/caching.md
kodster28: From another area of our docs, `custom cache keys let you precisely set the cacheability setting for any resource`

kodster28: Realize we actually need to add that info to another place as well 🙃 .

#15212

kodster28:
hyperlint review

content/ai-gateway/configuration/caching.md
kathayl: what i learned the order of operations is:
-if cache-ttl is set, that takes priority
-if it's not, then it's whatever the cache setting the user set in the dashboard is
-if the user had not turned on caching in the dashboard, then it's 5 minutes

How's this rewrite?

The first time you use the header, you will get a response from the provider. Then, each subsequent time the header is used, you will receive the cached response. If the cf-cache-ttl header is used, responses will be cached according to Cache TTL. However, if no cf-cache-ttl header is used, responses will be cached according to dashboard settings. If caching is not turned on for the gateway, responses will be cached for 5 minutes by default.

PR #15250, opened 2024-06-21
[CDN] Update ref-arch-cdn-figure2.svg

luishcardoso:
Hey team, how can we fix the typo ("Closet") in this Figure? Shall we replace this with a new one? Please let me now. Thanks!  PCX-10902


kodster28:
addressed by #15518

PR #15263, opened 2024-06-23
[Docs Site] bump dependencies, improve building on Windows

Cherry:
I was previously getting lots of weird errors when trying to build with Windows, and common solutions involved upgrading Vite.

This PR includes a few maintenance tasks on the docs site:

- I took the time to upgrade Vite and a bunch of other dependencies as much as possible while remaining compatible with Node 20 and each package's current version ranges. I tested this and anecdotally the Vite build feels faster, and I've not been able to find any regressions throughout.
  - Most dep bumps are minor, outside of Vite which is a major version upgrade from 4 -> 5
- Upgrades the `volta` definition for the Node.js version the latest v20-lts, since that seems to be what's used in CI here now.
- Adds a `.gitattributes` to better normalise line endings when committing from Windows vs Linux vs Mac, so no mixture of CRLF and LF files

If there's any further testing I can do for this PR here, please let me know.


KianNH (approved):
Tested this locally, and all the below is still fine:

* Mermaid diagrams
* Search
* Tabs
* JSON that the Pages dashboard depends on
* Lightbox JS
* Open in Playground examples

I'd still like @kodster28 to double-check, but the only thing for us to communicate to contributors is to do a fresh `npm ci` (mainly because of the glob import change, otherwise it would work albeit outdated).

Cherry:
Thanks @KianNH! I just updated this briefly again to bump a few more deps around TypeScript. Everything still seems to work fine from my testing. Let me know if I can do anything to assist getting this merged.

PR #15265, opened 2024-06-23
[Docs Site] use npm in workflows

Cherry:
A couple of workflows here were still using `pnpm` despite the repo using npm and shipping a `package-lock.json`.

This PR does the following:

- Removes `pnpm` usage in workflows
- Switches to using `npm ci` for package installs so that the lockfile is actually used - this should prevent any weird issues in future that come with transitive dependency updates
- Adds an `actions/cache` step to cache npm package installs which should speed up these workflows


GregBrimble (approved):
Thanks! LGTM but I will leave it to @KianNH as final arbitrator of docs' direction.

KianNH (approved):
Looks good to me, CI is happy.

PR #15267, opened 2024-06-23
[Docs Site]: minor TS improvements to assets

Cherry:
This PR makes the following minor improvements to many of the the `assets/*.ts` files:

- Transitions to more modern `querySelector` and `querySelectorAll` APIs
- With above, types the elements being queried more accurately
- Removes a bunch of unnecessary `as any` (etc.) type casting
- Better handles nullish elements when query selectors might fail

There should be no functional changes - just types improvements. This reduces the number of TypeScript errors reported (via `npx tsc --noEmit`) from ~63 to ~17.

### DIFF
--- a/assets/contents.ts
+++ b/assets/contents.ts
@@ -7,13 +7,13 @@ export function toc() {
   let target = document.querySelector('ul.DocsTableOfContents');
   let article = target && document.querySelector('article.DocsMarkdown');
 
-  if (article) {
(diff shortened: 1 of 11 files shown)


KianNH (approved):
Tested all of the non-types changes (like checking if `var` exists before using it) and couldn't find any change in behaviour or breakages.

PR #15270, opened 2024-06-23
[Docs Site] minor html fixes

Cherry:
This simply fixes up a couple minor HTML issues where tags weren't being closed/ordered correctly

### DIFF
--- a/layouts/shortcodes/spotlight-author.html
+++ b/layouts/shortcodes/spotlight-author.html
@@ -7,7 +7,7 @@
 <div class="author-container">
   <div class="author-details">{{- $details -}}</div>
   <div>
-      <span class="author-details">Written by: {{- $author -}}</span>
+      <span class="author-details">Written by: </span>{{- $author -}}
   </div>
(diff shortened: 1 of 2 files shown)


layouts/shortcodes/spotlight-author.html
```diff
@@ -7,7 +7,7 @@
 <div class="author-container">
   <div class="author-details">{{- $details -}}</div>
   <div>
-      <span class="author-details">Written by: </span>{{- $author -}}</span>
+      <span class="author-details">Written by: {{- $author -}}</span>
```
KianNH: It looks like this styles the `$author` with `author-details`
```suggestion
      <span class="author-details">Written by: </span>{{- $author -}}
```

Before:
image

After:
image

KianNH (approved):

layouts/shortcodes/spotlight-author.html
Cherry: Good catch, thanks

PR #15290, opened 2024-06-24
Add Hyperdrive instructions for Exograph

ramnivas:


thomasgauvin:
@ramnivas this may land better in Workers docs, since that seems to be the scope of exograph. Perhaps a `Automatically generate GraphQL APIs from your Postgres data models with Exograph` tutorial doc that lands here https://developers.cloudflare.com/workers/tutorials/. The current Hyperdrive docs are specific to connecting from Workers to databases directly

ramnivas:
@thomasgauvin Thanks for the review. I have now moved it to the location you suggested. Let me know how it looks.

KimJ15:

PR #13511, opened 2024-03-18
Clarify timer behavior

irvinebroque:
refs https://github.com/cloudflare/cloudflare-docs/pull/15292

PR #15292, opened 2024-06-24
Changelog for timers improvement

irvinebroque:
refs EW PR 8128, https://github.com/cloudflare/cloudflare-docs/pull/13511

### DIFF
--- a/data/changelogs/workers.yaml
+++ b/data/changelogs/workers.yaml
@@ -7,7 +7,10 @@ productAreaLink: /workers/platform/changelog/platform/
 entries:
 - publish_date: '2024-06-25'
   description: |-
-    - When you use timers (`Date.now()`, `performance.now()`) in your Worker, the timing information returned now more accurately reflects the time that your Worker's [handler](/workers/runtime-apis/handlers/) is invoked. Refer to the [Performance and timers docs](/workers/runtime-apis/performance/) for more details.
+    - When you use timers (`Date.now()`, `performance.now()`) in your Worker, the timing information returned now more accurately reflects the time that your Worker's [handler](/workers/runtime-apis/handlers/) is invoked.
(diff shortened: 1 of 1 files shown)


kodster28 (approved):

data/changelogs/workers.yaml
```diff
@@ -5,6 +5,9 @@ productLink: "/workers/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+- publish_date: '2024-06-25'
+  description: |-
+    - When you use timers (`Date.now()`, `performance.now()`) in your Worker, the timing information returned now more accurately reflects the time that your Worker's [handler](/workers/runtime-apis/handlers/) is invoked. Refer to the [Performance and timers docs](/workers/runtime-apis/performance/) for more details.
```
irvinebroque:
```suggestion
    - When you use timers (`Date.now()`, `performance.now()`) in your Worker, the timing information returned now more accurately reflects the time that your Worker's [handler](/workers/runtime-apis/handlers/) is invoked.
    - Previously, the timestamp returned by timers was "frozen" to when the top-level (global) scope of the Worker is entered. This made measuring the timing of I/O inaccurate, as the [startup time](/workers/platform/limits/#worker-startup-time) of your Worker was included in any attempt to measure the first I/O of an invocation of your Worker.
    - With this change, the timestamp returned by a timer (ex: `performance.now()`) is "frozen" to the time that the [handler](/workers/runtime-apis/handlers/) of your Worker is invoked.
    - Refer to the [Performance and timers docs](/workers/runtime-apis/performance/) for more details.
```

kentonv:
I had written the release note as:

> Fixed a bug which caused `Date.now()` to return skewed results if called before the first I/O of the first request after a Worker first started up. The value returned would be offset backwards by the amount of CPU time spent starting the Worker (compiling and running global scope). This misled people into thinking that the first I/O (e.g. first fetch()) was slower than it really was. Contrary to popular belief, this skew had nothing to do with Spectre mitigations; it was simply a longstanding bug.

Why rewrite it? The new text seems more confusing IMO.

irvinebroque:
Captured in changelog in another PR

PR #15293, opened 2024-06-24
[DDoS Protection] EU CMB support

patriciasantaana:
PCX-9318

### DIFF
--- a/content/data-localization/compatibility.md
+++ b/content/data-localization/compatibility.md
@@ -40,7 +40,7 @@ The table below provides a summary of the Data Localization Suite product's beha
 | Product | Geo Key Manager | Regional Services | Customer Metadata Boundary |
 | --- | --- | --- | --- |
 | Advanced Certificate Manager | ⚫️ | ⚫️ | ⚫️ |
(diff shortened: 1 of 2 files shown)


content/data-localization/compatibility.md
```diff
@@ -40,7 +40,7 @@ The table below provides a summary of the Data Localization Suite product's beha
 | Product | Geo Key Manager | Regional Services | Customer Metadata Boundary |
 | --- | --- | --- | --- |
 | Advanced Certificate Manager | ⚫️ | ⚫️ | ⚫️ |
-| Advanced DDoS Protection | ✅ | ✅ | 🚧{{<fnref num="3">}} |
+| Advanced DDoS Protection | ✅ | ✅ | ✅ |
```
pedrosousa, line 43: Nit: I don't think we're using footnote number 3 now. If that's the case, we can remove it from the bottom of the page.

pedrosousa (approved):

content/data-localization/compatibility.md
patriciasantaana: Addressed :)

PR #15305, opened 2024-06-24
Add instructions for properties of Durable Object errors

jclee:
Updating PR #14261 to address feedback and rebase; also adding entries for the workers and durable objects changelogs.


content/durable-objects/best-practices/error-handling.md
```diff
@@ -54,14 +62,13 @@ export default {
...
-      // It is recommended you retry requests to your Durable Object as long
-      // as the operation in your Durable Object is idempotent.
-      try {
-	      return await makeRequest(env, id);
-      } catch (e: any) {
-        // Retry still did not work, therefore return an error.
-        return new Response("server error", { status: 500 });
+      if (e.retryable) {
+        try {
+          return await makeRequest(env, id);
```
justin-mp: Is it possible to put something about backoff in the example?  People are going to copy this blindly, so having something about backoff is critical.  (A comment would do, but code would be better.)

content/durable-objects/best-practices/error-handling.md
```diff
@@ -25,7 +23,17 @@ A Durable Object stub is a client Object used to send requests to a remote Durab
...
+Any uncaught exceptions thrown by a Durable Object or its infrastructure will be propagated to the callsite of the client. Catching these exceptions allows you to retry creating the Durable Object stub and sending requests.
+
+{{<Aside type="note" header="Durable Object stubs">}}
+
+If any exception is thrown from a Durable Object stub, then the stub will need to be recreated prior to retrying requests. This is to provide E-order semantics as discussed in [create Durable Object stubs and send requests](/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/#create-durable-object-stubs-and-send-requests).
+
+JavaScript Errors with the property `.retryable` set to True are suggested to be retried if requests to the Durable Object are idempotent, or can be applied multiple times without changing the response. If requests are not idempotent, then you will need to decide what is best for your application.
+
+JavaScript Errors with the property `.overloaded` set to True should not be retried. If a Durable Object is overloaded, then retrying will worsen the overload and increase the overall error rate.
```
irvinebroque, line 34: Wonder if helpful to reference here at all?

https://developers.cloudflare.com/workers/observability/errors/
https://developers.cloudflare.com/workers/runtime-apis/web-standards/

content/durable-objects/best-practices/error-handling.md
jclee: Sure...  I know we have a note at the bottom suggesting that production code should use exponential backoff instead, but that's easy to miss.  I went ahead and added an attempt at randomized exponential backoff...  Hopefully it doesn't add too much complexity to the example?

content/durable-objects/best-practices/error-handling.md
jclee: I started adding info to `/workers/observability/errors/`, but then I noticed that there was already similar information in the [RPC error-handling](https://developers.cloudflare.com/workers/runtime-apis/rpc/error-handling/) section.  So I added info there and added a "Related resource" link from `/workers/observability/errors/` to `/workers/runtime-apis/rpc/error-handling/`.

Was there a particular place it would make sense to mention the behavior on the web-standards page?  I think having the additional properties on Error classes is probably non-standard, but I don't know if it makes sense to call it out at the same level as the `eval()` or `Date.now()` differences on that page.

justin-mp (approved):

content/durable-objects/best-practices/error-handling.md
```diff
@@ -10,12 +10,10 @@ Review how to handle errors (exceptions) generated by both your own Durable Obje
...
-- From code within the `constructor` method defined on your Durable Object class. The resulting exception will have a `.remote` property set to `True` in this case.
-- Generated by Durable Object's infrastructure due to transient internal errors, or because you are sending too many requests to a single Durable Object, or when too many requests are queued due to slow or excessive I/O (external API calls or storage operations) within an individual Durable Object.
-Some of the exceptions generated by Durable Object's infrastructure will also have the `.remote` property set to `True`.
+- Exceptions are thrown within the user code which implements a Durable Object class. The resulting exception will have a `.remote` property set to `True` in this case.
+- Generated by Durable Object's infrastructure due to transient internal errors, or because you are sending too many requests to a single Durable Object, or when too many requests are queued due to slow or excessive I/O (external API calls or storage operations) within an individual Durable Object. Some of the exceptions generated by Durable Object's infrastructure will also have the `.remote` property set to `True`.
```
vy-ton: Multiple `or...or` to list with `and` instead?

vy-ton: > Some of the exceptions generated by Durable Object's infrastructure will also have the `.remote` property set to `True`.

Can we share example scenarios?

content/durable-objects/best-practices/error-handling.md
```diff
@@ -25,53 +23,70 @@ A Durable Object stub is a client Object used to send requests to a remote Durab
...
+Any uncaught exceptions thrown by a Durable Object or its infrastructure will be propagated to the callsite of the client. Catching these exceptions allows you to retry creating the Durable Object stub and sending requests.
+
+{{<Aside type="note" header="Durable Object stubs">}}
+
+If any exception is thrown from a Durable Object stub, then the stub will need to be recreated prior to retrying requests. This is to provide E-order semantics as discussed in [create Durable Object stubs and send requests](/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/#create-durable-object-stubs-and-send-requests).
+
+JavaScript Errors with the property `.retryable` set to True are suggested to be retried if requests to the Durable Object are idempotent, or can be applied multiple times without changing the response. If requests are not idempotent, then you will need to decide what is best for your application.
+
+JavaScript Errors with the property `.overloaded` set to True should not be retried. If a Durable Object is overloaded, then retrying will worsen the overload and increase the overall error rate.
```
vy-ton, line 34: Can we update https://5545ebe5.cloudflare-docs-7ou.pages.dev/durable-objects/observability/troubleshooting/#durable-object-is-overloaded to mention the `overloaded` property?

content/durable-objects/best-practices/error-handling.md
```diff
@@ -10,12 +10,10 @@ Review how to handle errors (exceptions) generated by both your own Durable Obje
...
 ## How exceptions are thrown
 
-There are three ways in which a Durable Object can throw an exception:
+Durable Objects can throw exceptions in one of two ways:
 
-- From the code within the `fetch()` handler defined on your Durable Object class. The resulting exception will have a `.remote` property set to `True` in this case.
-- From code within the `constructor` method defined on your Durable Object class. The resulting exception will have a `.remote` property set to `True` in this case.
-- Generated by Durable Object's infrastructure due to transient internal errors, or because you are sending too many requests to a single Durable Object, or when too many requests are queued due to slow or excessive I/O (external API calls or storage operations) within an individual Durable Object.
-Some of the exceptions generated by Durable Object's infrastructure will also have the `.remote` property set to `True`.
+- Exceptions are thrown within the user code which implements a Durable Object class. The resulting exception will have a `.remote` property set to `True` in this case.
```
vy-ton: Is it possible for an exception to be both `.remote` and `.retryable`?

data/changelogs/durable-objects.yaml
```diff
@@ -5,6 +5,10 @@ productLink: "/durable-objects/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+  - publish_date: '2024-06-24'
+    description: |-
+      [Exceptions](/durable-objects/best-practices/error-handling) thrown from Durable Object internal operations and tunneled to the caller may now be populated with a `.retryable: true` property if the exception was likely due to a transient failure, or populated with an `.overloaded: true` property if the exception was due to overload.
```
vy-ton:
```suggestion
      [Exceptions](/durable-objects/best-practices/error-handling) thrown from Durable Object internal operations and tunneled to the caller may now be populated with a `.retryable: true` property if the exception was likely due to a transient failure, or populated with an `.overloaded: true` property if the exception was due to [overload](/durable-objects/observability/troubleshooting/#durable-object-is-overloaded).
```

added link for "overload" explanation

vy-ton (approved):

content/durable-objects/best-practices/error-handling.md
jclee: I think the main way that can happen is if one Durable Object calls another, the called DO throws a retryable exception, and the calling DO fails to handle the thrown exception.  In that case, the escaping exception would have both a retryable property, because the original exception was retryable, and a remote property because it escaped from user code.  I'm not sure if it's helpful to document this particular case, though?

content/durable-objects/best-practices/error-handling.md
jclee: It is probably worth noting that many of the other errors on that page are also going to have the overloaded property set, too. ("Your account is doing too many concurrent storage operations", "Your account is generating too much load on Durable Objects", and "Durable Object storage operation exceeded timeout".)  I'll try adding a mention at the top of the section, pointing to the best-practices error handling page.

I notice we have advice on the "Durable Object storage operation exceeded timeout" error that it's OK to retry failing `deleteAll()` operations, which somewhat conflicts with advising customers to not retry `overloaded` errors.

content/durable-objects/best-practices/error-handling.md
jclee: It looks like `.remote` would be true if the Durable Object exceeds CPU or memory limits.  I added that example to the paragraph.

content/durable-objects/best-practices/error-handling.md
vy-ton: No, I wouldn't include this scenario. Thanks for walk through

content/durable-objects/best-practices/error-handling.md
vy-ton: > I'll try adding a mention at the top of the section, pointing to the best-practices error handling page.

makes sense to me

> I notice we have advice on the "Durable Object storage operation exceeded timeout" error that it's OK to retry failing deleteAll() operations, which somewhat conflicts with advising customers to not retry overloaded errors.

I don't understand current docs stating `When this happens, note that each deleteAll() call does make progress and that it is safe to retry until it succeeds.`? -- Is it actually true?

jclee: > I don't understand current docs stating When this happens, note that each deleteAll() call does make progress and that it is safe to retry until it succeeds.? -- Is it actually true?

Yeah, skimming the code, it seems that `deleteAll()` deletes its rows in chunks, and is special in that it is not run as part of a transaction with other operations.  I expect this was to reduce the likelihood of timeout errors when deleting very large namespaces.  It looks like the documentation was originally added in PR #5068 .

vy-ton:
@jclee Can you resolve conflicts? - then we can merge this today

jclee:
> @jclee Can you resolve conflicts? - then we can merge this today

I've rebased, resolving the conflict between the workers changelog entry and the changelog update that went in yesterday.  But I think I am still blocked from merging, myself, due to needing approval from other code owners.  But maybe someone else has permission to merge?

vy-ton:
@irvinebroque or @kodster28 do you mind adding codeowner approval?

elithrar (approved):

irvinebroque (approved):

PR #15300, opened 2024-06-24
[Style guide] Table guidelines

kodster28:

PR #15306, opened 2024-06-24
[Chore] Remove table-wrap from Zaraz

kodster28:
Adding new guidelines in #15300 around tables + table-wrap.

It's not needed for most tables and causes some unexpected issues on mobile.


KianNH (approved):
Spot-checked on desktop & mobile, all looked good!

PR #15307, opened 2024-06-24
[Chore] Remove workers tablewrap where appropriate

kodster28:
Similar vibes as to #15306.

Adding new guidelines in #15300 around tables + table-wrap.

It's not needed for most tables and causes some unexpected issues on mobile. This removes table-wrap from small tables in the `/workers-ai/`, `/workers/`, and `/pages/` spaces. Used mobile preview while doing local dev and didn't see any issues.


irvinebroque (approved):

KianNH (approved):
LGTM - went through a handful on iOS and they all looked fine to me.

PR #15352, opened 2024-06-27
[DNS] Clarify DNSSEC options with Secondary DNS Override

dom-cf:
Secondary DNS Override will not work with Pre-signed DNSSEC, adding a warning to clarify.


Fattouche:
Thanks!

RebeccaTamachiro (approved):

PR #15369, opened 2024-06-27
[Style Guide] Strategy updates

crwaters16:


kodster28 (approved):
I might add some more and partialize it in a bit, but lgtm for now

PR #15371, opened 2024-06-27
ai: link to AI Gateway in sidebar

elithrar:
Could also alternatively be under Platform? The cross-link is important for discovery!


kodster28 (approved):
Moved under platform b/c the top-level nav is getting a bit crowded

PR #15373, opened 2024-06-27
Update connection tampering section in Radar glossary to resets and timeouts (#10618)

lukevalenta:
We also need to update the link to this glossary entry in https://blog.cloudflare.com/syria-iraq-algeria-exam-internet-shutdown.


content/radar/glossary.md
```diff
@@ -95,6 +95,19 @@ Domain Rankings is based on our anonymized and aggregated 1.1.1.1 resolver data,
 
 Trending domains are domains which are currently experiencing an increase in popularity. Domains Trending Today are domains spiking in popularity, reflecting increased interest potentially related to a particular event or a topic. Domains Trending This Week are domains that have steadily grown in popularity, reflecting an increase of their user base over the week.
 
+## Early TCP connection resets and timeouts
```
marwanfayed: Maybe "TCP early resets, and timeouts" or "TCP timeouts and early resets"? O/w I read as being possible to have an "early timeouts"

content/radar/glossary.md
```diff
@@ -95,6 +95,19 @@ Domain Rankings is based on our anonymized and aggregated 1.1.1.1 resolver data,
...
+
+On Cloudflare Radar’s [Security & Attacks page](https://radar.cloudflare.com/security-and-attacks), you can view statistics on resets and timeouts from a sample of TCP connections to Cloudflare’s servers, broken down by how far the connection progressed before termination. The plot lines are defined as follows:
+* **Mid-handshake (Post-SYN)**: Connection resets or timeouts after the server received only a single SYN packet.
+* **Immediately post-handshake (Post-ACK)**: Connection resets or timeouts after the server received both a SYN packet and an ACK packet, meaning the connection was successfully established.
```
marwanfayed: Some addition to say, "but before any data packets, such as TLS clienthello" or similar?

content/radar/glossary.md
lukevalenta: I think "TCP early resets and timeouts" might be appropriate actually, given that we only detect resets and timeouts when they occur early in the connection (within the first 10 packets).

content/radar/glossary.md
lukevalenta: The following bullet point (Post-PSH) describes what 'data' we're talking about ("The PSH flag indicates that the TCP packet contains data (such as a TLS Client Hello message) ready to deliver to the application."), so I'd left it out here to avoid being redundant. Happy to include it in the Post-ACK description too though--what do you think?

meddulla (approved):

PR #15406, opened 2024-06-29
[Speed] Update adding-captions.md

luishcardoso:
Hey team, let's clarify this feature is currently in Beta, per info already shown on the Stream > Videos dashboard. Thanks!


third774 (approved):

kodster28:
I'd prefer we didn't do this inline + zooming out a bit, we generally only label `feature` level things as beta in our docs (as opposed to options within a feature).

PR #15412, opened 2024-06-30
add missing backtick on section `Transitive loading of @types/node overrides @cloudflare/workers-types `

KeisukeYamashita:
## Why

Because it is malformed on the website.

Screenshot-2024-06-30 at 15 14 11@2x


WalshyDev (approved):
Thank you!

PR #15419, opened 2024-07-01
[WAF] Add general updates changelog

pedrosousa:
Addresses PCX-11132.


content/waf/change-log/managed-rulesets.md
pedrosousa, line 1: RSS-feed-only page in case some customers want to have a WAF changelog just for managed ruleset updates.

kodster28 (approved):

PR #15481, opened 2024-07-05
Update _universal-ssl-enable-full.md

jmesquita1:
I have come across some tickets where customer are confused as to why they are receiving USSL for full zones when their records are DNS-only. 

Small change to let them know that is expected behavior.


RebeccaTamachiro:
Thank you, @jmesquita1! Have raised PCX-12578 to clarify a doubt that came up. 
Let's talk there and update the PR if needed.

RebeccaTamachiro (approved):

PR #15486, opened 2024-07-05
[Workers] Adds tutorial to run Workerd in container

megaconfidence:
This PR adds a new tutorial to run Workerd in a Docker container.


irvinebroque (changes requested):
Would like to discuss first

irvinebroque:
A tutorial (if needed) to run `Wrangler` in a container makes sense to me as part of the Cloudflare docs.

A tutorial that is intended to show you how to _self-host_ `workerd` and run it directly in a container seems like it makes sense here: https://github.com/cloudflare/workerd/tree/main/docs

PR #13149, opened 2024-02-26
[DNS] Exposed DNS settings in dash

RebeccaTamachiro:
PCX-9776
PCX-10635

PR #15493, opened 2024-07-06
Update account-custom-nameservers.md

jmesquita1:
Spoke with the DNS team and I confirmed this behavior with them.

I thought it would be good to have this publicly documented as I had customers come to me with this question.


RebeccaTamachiro:
Holding back on this one in favor of #13149 
(Also created PCX-12870 to consolidate information around that topic)

PR #15496, opened 2024-07-06
[Pages] Correct header syntax typo in early-hints.md

Nathan2055:
This fixes a typo in early-hints.md where the incorrect attribute `as=stylesheet` is used instead of the correct `as=style` used elsewhere in the document.


kodster28 (approved):

kodster28:
Thanks for the catch and fix, @Nathan2055.

PR #15502, opened 2024-07-08
[Docs Site] Support variables within codeblocks in partials

KianNH:
When using `$1` in a partial, within a codeblock, the `render` shortcode receives something like this:

```html
data-code="%241"
```

This is because it is URL encoded by the [render-codeblock.html](https://github.com/cloudflare/cloudflare-docs/blob/a8c009efb1fa0e760b5d3620302af5885b97cf9b/layouts/_default/_markup/render-codeblock.html#L2) hook.

This change means users can use `$1, $n...` inside of codeblocks.


kodster28 (approved):
tested locally and lgtm

PR #15529, opened 2024-07-09
Add example of correct and incorrect O2O configuration to SFCC O2O Provider Guide

jthorne-cf:


content/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/provider-guides/salesforce-commerce-cloud.md
```diff
@@ -36,6 +36,26 @@ The `CNAME` provided by SFCC Business Manager will resemble `commcloud.prod-abcd
 | --- | --- | --- | --- |
 | `CNAME` | `<YOUR_HOSTNAME>` | `commcloud.prod-abcd-example-com.cc-ecdn.net` | Proxied |
 
+{{<Aside type="note">}}
```
kodster28: I'd recommend moving this to a separate section later in the page and then separating correct vs incorrect with headers. It's not immediately clear from scanning, which we'd want to do if we're ever providing an example of an incorrect version.

kodster28 (approved):

PR #15532, opened 2024-07-09
Adds information about local dev environment vars

mikenomitch:
### Summary

It took me a while to figure out the best way of doing dev-only environment variables, so I'm adding some additional information to the docs.

I'm not positive that there isn't some automatic "environment" value applied in dev mode. If there is, I should mention this instead of pointing people towards the `--env` flag. Or if we have some preferred/alternative way of doing this, let me know!

### DIFF
--- a/content/workers/configuration/environment-variables.md
+++ b/content/workers/configuration/environment-variables.md
@@ -71,6 +71,24 @@ SERVICE_X_DATA = { URL = "service-x-api.dev.example", MY_ID = 123 }
 API_HOST = "production.example.com"
 API_ACCOUNT_ID = "production_example_user"
 SERVICE_X_DATA = { URL = "service-x-api.prod.example", MY_ID = 456 }
+
+# used when providing "--env=local" flag to "wrangler dev"
+[env.local.vars]
+API_HOST = "localhost:3000"
+API_ACCOUNT_ID = "local_example_user"
+SERVICE_X_DATA = { URL = "localhost:4000", MY_ID = 789 }
+```
+
+For local development, you can specify an `environment` value via the `env` flag when running the development command like so `wrangler dev --env=local`. Note that you may want to modify your `package.json` if you include a `dev` script that calls `wrangler dev`.
+
+Alternatively, wrangler automatically uses values defined in a `.dev.vars` file located in the root directory of your worker when you run `wrangler dev`. These values will override any values specified in your `wrangler.toml` file. Nested variables are not supported in `.dev.vars`.
+
+```shell
+---
+filename: .dev.vars
+---
+API_HOST = "localhost:4000"
+API_ACCOUNT_ID = "local_example_user"
 ```
 
 ## Add environment variables via the dashboard
--- a/content/workers/configuration/environment-variables.md
+++ b/content/workers/configuration/environment-variables.md
@@ -83,7 +83,7 @@ API_HOST = "localhost:4000"
 API_ACCOUNT_ID = "local_example_user"
 ```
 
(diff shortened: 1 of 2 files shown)


content/workers/configuration/environment-variables.md
```diff
@@ -71,6 +71,24 @@ SERVICE_X_DATA = { URL = "service-x-api.dev.example", MY_ID = 123 }
...
+# used when providing "--env=local" flag to "wrangler dev"
+[env.local.vars]
+API_HOST = "localhost:3000"
+API_ACCOUNT_ID = "local_example_user"
+SERVICE_X_DATA = { URL = "localhost:4000", MY_ID = 789 }
+```
+
+For local development, you can specify an `environment` value via the `env` flag when running the development command like so `wrangler dev --env=local`. Note that you may want to modify your `package.json` if you include a `dev` script that calls `wrangler dev`.
+
+Alternatively, wrangler automatically uses values defined in a `.dev.vars` file located in the root directory of your worker when you run `wrangler dev`. These values will override any values specified in your `wrangler.toml` file. Nested variables are not supported in `.dev.vars`.
```
mikenomitch: > Nested variables are not supported in `.dev.vars`.

Is there a better way of expressing this?

irvinebroque (approved):
👍 think @petebacondarwin would be good 👁️ on this

petebacondarwin:
What you are describing here @mikenomitch is just another "named environment" no different to "staging" or "prod". Your proposal is to use this environment when doing local development. While this works, it is not ideal if you want to have secrets available in local dev, since you would not want to add these to your `wrangler.toml`.

The more common approach, I have seen in the past, is just to define your local vars at the top level and then override them in named environments. To prevent you from accidentally deploying these top level vars to production, Wrangler ensures that you have defined all the vars that are at the top level in any named environment that you are deploying. And then any secret vars that should not appear in the `wrangler.toml` go in the `.dev.vars` file.

Is this not a good workflow approach for you?

mikenomitch:
@petebacondarwin, that approach seems fine to me. I'm less concerned about the actual approach and more concerned that we explain to the user what they should do. For instance, I didn't know the interplay between top-level vars and environment-level vars until I just read what you wrote. I can add a bit about that!

As far as the actual approach, it feels a bit weird to me to override top level vars with named environments. Having API_HOST=localhost:3000 (or something like this) as the "default" value just feels somewhat "off". But I don't really feel strongly and can't quite put into words why, so I'm okay suggesting this to users if that's what we've been doing.

petebacondarwin:
The environment story is described here https://developers.cloudflare.com/workers/wrangler/configuration/#environments but perhaps not with the greatest clarity.

mikenomitch:
@petebacondarwin I changed it to lead with .dev.vars and only mention --env as a possibility briefly after (though I'm happy to remove this if you want). There's also a Local Development page that I thought could use this info too, so I added info about .dev.vars there as well.

content/workers/configuration/environment-variables.md
```diff
@@ -73,6 +73,18 @@ API_ACCOUNT_ID = "production_example_user"
...
+
+```shell
+---
+filename: .dev.vars
+---
+API_HOST = "localhost:4000"
+API_ACCOUNT_ID = "local_example_user"
+```
+
+Alternatively, you can specify local-only values in `wrangler.toml` and provide an `environment` value via the `env` flag when running the development command like so `wrangler dev --env=local`.
```
petebacondarwin: This looks good. I would like to be careful not to give the impression that the `local` name for the environment has any special meaning. It is just a convention you have adopted here.

petebacondarwin (approved):

content/workers/configuration/environment-variables.md
mikenomitch: Good point - will tweak wording

PR #15541, opened 2024-07-10
Adds illegal invocation error information

mikenomitch:
### Summary

Adds illegal invocation error information. We want to link to this information via this message in workderd: https://github.com/cloudflare/workerd/pull/2375#issuecomment-2218101724

### DIFF
--- a/content/workers/observability/errors.md
+++ b/content/workers/observability/errors.md
@@ -98,10 +98,9 @@ The error message `TypeError: Illegal invocation: function called with incorrect
 
 This is typically caused by calling a function that calls `this`, but the value of `this` has been lost.
 
-For example, if `obj.foo()` is a method, and within that method `this` references `obj`, executing the method via `obj.foo();` will work properly. However, assigning it to a variable, `let func = obj.foo;`, then calling it, `func();`, will result in `this` being undefined. This is because `this` is lost when the method is called as a standalone function. This is standard behavior in JavaScript.
+For example, given an `obj` object with the `obj.foo()` method which logic relies on `this`, executing the method via `obj.foo();` will make sure that `this` properly references the `obj` object. However, assigning the method to a variable, e.g.`const func = obj.foo;` and calling such variable, e.g. `func();` would result in `this` being `undefined`. This is because `this` is lost when the method is called as a standalone function. This is standard behavior in JavaScript. 
 
-In practice, this is often seen when destructuring Javscript objects that have functions that rely on the presence of `this`, such as `ctx`. Avoid destructuring or re-bind the function to
-the original context to avoid the error.
+In practice, this is often seen when destructuring runtime provided Javscript objects that have functions that rely on the presence of `this`, such as `ctx`. Avoid destructuring or re-bind the function to the original context to avoid the error.
 
 ```js
 ---
@@ -110,7 +109,7 @@ filename: index.js
 export default {
   async fetch(request, env, ctx) {
 
-    // destructuring ctx makes waitUntil lose its receiver
+    // destructuring ctx makes waitUntil lose its 'this' reference
     let { waitUntil } = ctx;
     // waitUntil errors, as it has no 'this'
     waitUntil(somePromise);
@@ -118,11 +117,11 @@ export default {
     // re-binding to ctx via apply, call, or bind avoids the error
     waitUntil.apply(ctx, [somePromise]);
     waitUntil.call(ctx, somePromise);
-    let reboundWaitUntil = waitUntil.bind(ctx)
-    reboundWaitUntil(someAsyncFunction)
+    const reboundWaitUntil = waitUntil.bind(ctx);
+    reboundWaitUntil(somePromise);
 
-    // not destructuring ctx at all avoids the error
-    ctx.waitUntil(someAsyncFunction)
+    // directly calling the method on ctx avoids the error
+    ctx.waitUntil(somePromise);
 
     return fetch(request);
   }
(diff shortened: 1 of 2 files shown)


jasnell (approved):

content/workers/observability/errors.md
```diff
@@ -92,6 +92,41 @@ async function handleRequest(request) {
 }
 ```
 
+### "Illegal invocation" errors
+
+The error message `TypeError: Illegal invocation: function called with incorrect \`this\` reference` can be a source of confusion.
```
irvinebroque:
```suggestion
The error message `TypeError: Illegal invocation: function called with incorrect this reference` can be a source of confusion.
```

content/workers/observability/errors.md
```diff
@@ -92,6 +92,41 @@ async function handleRequest(request) {
...
+
+This is typically caused by calling a member function of an object without the object itself as the receiver. For example, `let foo = obj.foo; foo();`. This is standard behavior in Javscript.
```
irvinebroque: I think we can be clearer here and use words people actually know, rather than "member function of an object without the object itself as the receiver"

irvinebroque: i.e. — when you destructure an object, like this:

```
let foo = obj.bar;
await foo();
```

...when you call `foo()`, the value of `this` is different than it would be if you called `obj.bar()`.

irvinebroque (approved):
Non-blocking, can iterate on — but I think with these the more explicit we can be about:

Don't do this:

Do this instead:

In a loud way — separate code examples, separate headings — the clearer these can be, and the less we'll need to explain in words (often using teminology that I'd argue most JS devs don't really know about)

content/workers/observability/errors.md
mikenomitch: Took another pass - LMK if you prefer it

mikenomitch:
@irvinebroque ah just saw your top level comment - I tried keeping it in a single code block just to keep it from making the long page even longer. But maybe two is preferable - will give it a try and see what I think.

content/workers/observability/errors.md
```diff
@@ -92,6 +92,43 @@ async function handleRequest(request) {
 }
 ```
 
+### "Illegal invocation" errors
+
+The error message `TypeError: Illegal invocation: function called with incorrect this reference` can be a source of confusion.
+
+This is typically caused by calling a function that calls `this`, but the value of `this` has been lost.
+
+For example, if `obj.foo()` is a method, and within that method `this` references `obj`, executing the method via `obj.foo();` will work properly. However, assigning it to a variable, `let func = obj.foo;`, then calling it, `func();`, will result in `this` being undefined. This is because `this` is lost when the method is called as a standalone function. This is standard behavior in JavaScript.
```
dario-piotrowicz:
```suggestion
For example, given an `obj` object with the `obj.foo()` method which logic relies on `this`, executing the method via `obj.foo();` will make sure that `this` properly references the `obj` object. However, assigning the method to a variable, e.g.`const func = obj.foo;` and calling such variable, e.g. `func();` would result in `this` being `undefined`. This is because `this` is lost when the method is called as a standalone function. This is standard behavior in JavaScript. 
```

content/workers/observability/errors.md
```diff
@@ -92,6 +92,43 @@ async function handleRequest(request) {
...
+
+In practice, this is often seen when destructuring Javscript objects that have functions that rely on the presence of `this`, such as `ctx`. Avoid destructuring or re-bind the function to
+the original context to avoid the error.
```
dario-piotrowicz:
```suggestion
In practice, this is often seen when destructuring runtime provided Javscript objects that have functions that rely on the presence of `this`, such as `ctx`. Avoid destructuring or re-bind the function to the original context to avoid the error.
```

content/workers/observability/errors.md
```diff
@@ -92,6 +92,43 @@ async function handleRequest(request) {
...
+
+```js
+---
+filename: index.js
+---
+export default {
+  async fetch(request, env, ctx) {
+
+    // destructuring ctx makes waitUntil lose its receiver
```
dario-piotrowicz: we never used the world `receiver` here, maybe we should keep this consistent with the above info?

```suggestion
    // destructuring ctx makes waitUntil lose its 'this' reference
```

content/workers/observability/errors.md
```diff
@@ -92,6 +92,43 @@ async function handleRequest(request) {
...
+    let { waitUntil } = ctx;
+    // waitUntil errors, as it has no 'this'
+    waitUntil(somePromise);
+
+    // re-binding to ctx via apply, call, or bind avoids the error
+    waitUntil.apply(ctx, [somePromise]);
+    waitUntil.call(ctx, somePromise);
+    let reboundWaitUntil = waitUntil.bind(ctx)
+    reboundWaitUntil(someAsyncFunction)
```
dario-piotrowicz:
```suggestion
    const reboundWaitUntil = waitUntil.bind(ctx);
    reboundWaitUntil(somePromise);
```

content/workers/observability/errors.md
```diff
@@ -92,6 +92,43 @@ async function handleRequest(request) {
...
+
+    // not destructuring ctx at all avoids the error
+    ctx.waitUntil(someAsyncFunction)
```
dario-piotrowicz:
```suggestion
    // directly calling the method on ctx avoids the error
    ctx.waitUntil(somePromise);
```

kodster28 (approved):
Looks like you have some comments from Dario as well though

PR #15573, opened 2024-07-11
[Workers AI] Model updates

kodster28:
### Summary

Updating workers ai models from api source


kodster28:
exposed props, closing

PR #15574, opened 2024-07-11
Deprecated model labels

### LINKED ISSUE #15562: Update `llama-2-7b-chat-int8` chat model to reflect its current deprecated status

- https://developers.cloudflare.com/workers-ai/models/llama-2-7b-chat-int8

My application recently stopped working due to the deprecation of the `llama-2-7b-chat-int8` model. I suggest that the model pages reflect this change automatically.

To handle the scenario where today's date is greater than or equal to the `planned_deprecation_date`, we need to use a date comparison within the template. We can transform the date string into a date object for comparison. As the templating language in this case does not directly support complex date manipulations, it would be more feasible if handled within the content management system or at the template rendering layer.

Let’s assume we can use Go's templating functions to help with this. The following changes will be required:

1. Modify the `_default/model.html` template to check whether today's date is greater than or equal to the `planned_deprecation_date`.

2. Use Go’s templating functions to perform the date comparison and conditionally render content.

### Additional information

_No response_

kodster28:
### Summary

Small tweaks to what's proposed in #15563.

Closes #15562

### DIFF
--- a/layouts/_default/model.html
+++ b/layouts/_default/model.html
@@ -18,28 +18,30 @@
       <article class="DocsMarkdown">
           <h1>{{ .Title }}</h1>
           <div class="DocsMarkdown-beta">
+          {{ $now := now }}
           {{ range .Params.model.properties }}
             {{ if and (eq .property_id "beta") (eq .value "true") }}
-              {{/* Cannot use heading-pill shortcode outside of content dir */}}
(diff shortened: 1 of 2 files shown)


KianNH (approved):
Tested that it renders fine on https://deprecated-model-labels.cloudflare-docs-7ou.pages.dev/workers-ai/models/llama-2-7b-chat-int8/

PR #15580, opened 2024-07-11
[Hugo] Expand external-resources-by-tag component

kodster28:
### Summary

`external-resources-by-tag` needs to be expanded to accept multiple inputs, hence we're:
1. Renaming it to be `external-resources`
2. Allowing it to accept `products` as a filtering input.

This will allow us to provide filtered apps and videos based on product names or arrays, instead of only being able to filter by tags.

### Additional work

Added new pages to `kv`, `durable objects`, and `Hono` docs to highlight related apps.


KianNH (approved):
Tested locally, looks good to me!

PR #15584, opened 2024-07-12
chore(redirects): updates redirects for settings

jacobbednarz:
301s a bunch of old URLs to the new operation ID path in the developer documentation.

In essence:

- `zone-settings-change-.+` => `zone-settings-edit-single-setting`
- `zone-settings-get-.+` => `zone-settings-get-single-setting`


jacobbednarz:
tried to test the redirects here however, it looks like they aren't deployed to the feature branch preview.

jacobbednarz:
@GregBrimble actually - this may not work here depending on the call order of the two separate repositories. are you able to confirm where these redirects _should_ live if it won't work as is?

GregBrimble:
`_redirects` only take effect at the lowest, static asset serving, level. If `/api` is being served by Pages Functions or another Worker route, that'll execute before `_redirects` even have an opportunity to execute. You'll need to handle the redirects in the application above this.

jacobbednarz:
assumed as much. i _suspect_ it is a different route so we'll need to do this further up.

PR #15595, opened 2024-07-12
[Rules] Update _index.md

luishcardoso:
Hello team

### Summary

This is for better readability. We already labeled it as beta at the product's page. Thanks!


pedrosousa:
We're following this approach in quite a few places, so this is something that we wouldn't want to change in a single place. I'll check internally if we want to change our approach about inline beta pills.

pedrosousa:
Inline pills removed here: #15705

PR #15597, opened 2024-07-12
Add O2O best practices to SFCC O2O Provider Guide

jthorne-cf:


kodster28 (approved):
Think this adds value for now.

In the medium term, should likely move this to a sub-page of the current doc b/c it's getting a bit long + break this out into separate sections on that new page.

PR #15621, opened 2024-07-15
[Network] Fix format error on grpc-connections docs

josephvoss:
Was looking at this page earlier today and noticed that this markdown list is double-indented, meaning it renders as a long string instead of the list. Removes the stray indentation so this list is rendered correctly.


kodster28 (approved):

kodster28:
Thanks for the catch and fix, @josephvoss

PR #15639, opened 2024-07-16
[Cache] Edit to cache keys docs

angelampcosta:


content/cache/how-to/cache-keys.md
```diff
@@ -7,16 +7,9 @@ meta:
...
-Host: www.cloudflare.com
-User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.90 Safari/537.36
-Accept: image/jpg
-```
-
-As we can see from the example, the default cache key includes:
+A default cache key includes:
 
 1.  Full URL:
     - scheme - not shown above, but could be HTTP or HTTPS.
```
pedrosousa: This part in `1.` relies on the example that is being deleted... should we remove any mentions to the example?

pedrosousa (approved):

PR #15681, opened 2024-07-18
[Crypto] Fixing description of timingSafeEqual

### LINKED ISSUE #15564: "Using timingSafeEqual" documentation is vulnerable to timing attack

- https://developers.cloudflare.com/workers/examples/protect-against-timing-attacks/

The documentation states "Since the time taken to encode the values may reveal the length of our secret value, you should check the length of the strings before encoding."

However, this introduces another timing attack: the timing of the response will be different if the attacker guesses the length of the secret or not. I would suggest encoding with a false secret of the same length of the value given by the attacker, and then comparing timingSafeEqual, but forcing an unauthorized response anyway in case this is a false secret.

### Additional information

_No response_

mschwarzl:


kodster28 (approved):

content/workers/examples/protect-against-timing-attacks.md
```diff
@@ -46,10 +47,6 @@ export default {
     const a = encoder.encode(authToken);
     const b = encoder.encode(env.MY_SECRET_VALUE);
 
-    if (a.byteLength !== b.byteLength) {
```
KianNH: Could you please explain why this check is being removed? Without this, a comparison between `a` and `à` will cause an exception to be thrown by `timingSafeEqual`.

KianNH (changes requested):

KianNH (approved):

Avazbek84 (commented):
Ооо

PR #15697, opened 2024-07-18
[ZT] Tunnel does not support public IPs

ranbel:
PCX-10434
PCX-11153


janani-cr (approved):
Thanks Ranbel, lets see what feedback we receive !

PR #15750, opened 2024-07-23
docs: add documentation for the --experimental-include-runtime flag (draft)

andyjessop:
### Summary

This adds documentation for the `wrangler types --experimental-include-runtime` flag that dynamically generates runtime types for any combination of `compatibility_date` and `compatibility_flags`.

The PR that introduces the feature: https://github.com/cloudflare/workers-sdk/pull/6295

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/languages/typescript/index.mdx
+++ b/src/content/docs/workers/languages/typescript/index.mdx
@@ -13,6 +13,36 @@ head:
 
 TypeScript is a first-class language on Cloudflare Workers. Cloudflare publishes type definitions to [GitHub](https://github.com/cloudflare/workers-types) and [npm](https://www.npmjs.com/package/@cloudflare/workers-types) (`npm install -D @cloudflare/workers-types`). All APIs provided in Workers are fully typed, and type definitions are generated directly from [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.
 
+### Generate types that match your Worker's configuration (experimental)
+
+Cloudflare continuously improves [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.
+Some updates change the TypeScript types that APIs return. For example, the [`urlsearchparams_delete_has_value_arg`](/workers/configuration/compatibility-dates/#urlsearchparams-delete-and-has-value-argument) compatibility flag adds optional arguments to some methods, in order to support new additions to the WHATWG URL standard API.
+
+This means the correct TypeScript types for your Worker depend on:
+
+1. Your Worker's [Compatibility Date](/workers/configuration/compatibility-dates/).
+2. Your Worker's [Compatibility Flags](/workers/configuration/compatibility-dates/#compatibility-flags).
+
+For example, if you have `compatibility_flags = ["nodejs_als"]` in your `wrangler.toml`, then the runtime will allow you to use the [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) class in your worker code, but you will not see this reflected in the type definitions in `@cloudflare/workers-types`.
+
+In order to solve this issue, and to ensure that your type definitions are always up-to-date with your compatibility settings, you can dynamically generate the runtime types (as of `wrangler 3.66.0`):
+
+```bash
+wrangler types --experimental-include-runtime
+```
+
+This will generate a `d.ts` file and (by default) save it to `.wrangler/types/runtime.d.ts`. You will be prompted in the command's output to add that file to your `tsconfig.json`'s `compilerOptions.types` array.
+
+If you would like to commit the file to git, you can provide a custom path. Here, for instance, the `runtime.d.ts` file will be saved to the root of your project:
+
+```bash
+wrangler types --experimental-include-runtime="./runtime.d.ts"
+```
+
+Note that the `--experimental-include-runtime` command replaces the need for the `@cloudflare/workers-types` package, so if you do have that package installed, you should remove it to avoid any potential for conflict.
+
+See [the full list of available flags](/workers/wrangler/commands/#types) for more details.
+
 ### Known issues
 
 #### Transitive loading of `@types/node` overrides `@cloudflare/workers-types`
--- a/src/content/docs/workers/languages/typescript/index.mdx
+++ b/src/content/docs/workers/languages/typescript/index.mdx
@@ -41,7 +41,7 @@ wrangler types --experimental-include-runtime="./runtime.d.ts"
 
 {{<Aside type="warning">}}
 
-To ensure that your types are always up-to-date, make sure to run `wrangler types --experimental-include-runtime` after any changes to your `wrangler.toml` file.
+To ensure that your types are always up-to-date, make sure to run `wrangler types --experimental-include-runtime` after any changes to your config file.
 
 {{</Aside>}}
 
@@ -49,17 +49,39 @@ See [the full list of available flags](/workers/wrangler/commands/#types) for mo
 
 #### Migrating from `@cloudflare/workers-types` to `wrangler types --experimental-include-runtime`
 
-The `@cloudflare/workers-types` package provides runtime types according to your compatibility date, and therefore it is superceeded by the `wrangler types --experimental-include-runtime` command.
+The `@cloudflare/workers-types` package provides runtime types according to your compatibility date, but is superceeded by the `wrangler types --experimental-include-runtime` command.
 
-Here are the steps to switch from `@cloudflare/workers-types` to using wrangler types with the experimental runtime inclusion:
+Here are the steps to switch from `@cloudflare/workers-types` to using `wrangler types` with the experimental runtime inclusion:
 
-##### 1. Uninstall `@cloudflare/workers-types`
+##### Uninstall `@cloudflare/workers-types`
 
-```bash
+<Tabs> <TabItem label="npm">
+
+```sh
 npm uninstall @cloudflare/workers-types
 ```
 
-##### 2. Generate runtime types using wrangler
+</TabItem> <TabItem label="yarn">
+
+```sh
+yarn remove @cloudflare/workers-types
+```
+
+</TabItem> <TabItem label="pnpm">
+
+```sh
+pnpm uninstall @cloudflare/workers-types
+```
+
+</TabItem> <TabItem label="bun">
+
+```sh
+bun remove @cloudflare/workers-types
+```
+
+</TabItem> </Tabs>
+
+##### Generate runtime types using wrangler
 
 ```bash
 wrangler types --experimental-include-runtime
@@ -67,7 +89,7 @@ wrangler types --experimental-include-runtime
 
 This will generate a `.d.ts` file, saved to `.wrangler/types/runtime.d.ts` by default.
 
-##### 3. Update your `tsconfig.json` to include the generated types
+##### Update your `tsconfig.json` to include the generated types
 
 ```json
 {
@@ -79,7 +101,7 @@ This will generate a `.d.ts` file, saved to `.wrangler/types/runtime.d.ts` by de
 
 Note that if you have specified a custom path for the runtime types file, you should use that in your `compilerOptions.types` array instead of the default path.
 
-##### 4. Update your scripts and CI pipelines
+##### Update your scripts and CI pipelines
 
 When switching to `wrangler types --experimental-include-runtime`, you'll want to ensure that your development process always uses the most up-to-date types. The main thing to remember here is that - regardless of your specific framework or build tools - you should run the `wrangler types --experimental-include-runtime` command before any development tasks that rely on TypeScript. This ensures your editor and build tools always have access to the latest types.
 
@@ -88,8 +110,8 @@ Most projects will have existing build and development scripts scripts, as well 
 ```json
 {
   "scripts": {
-    "dev": "your-existing-dev-command",
-    "build": "your-existing-build-command",
+    "dev": "existing-dev-command",
+    "build": "-existing-build-command",
     "type-check": "wrangler types --experimental-include-runtime && tsc"
   }
 }
@@ -97,7 +119,7 @@ Most projects will have existing build and development scripts scripts, as well 
 
 In CI, you may have a separate build and test commands. It is best practice to run `wrangler types --experimental-include-runtime` before other CI commands. For example:
 
-```text
+```yaml
 - run: npm run generate-types
 - run: npm run build
 - run: npm test
--- a/src/content/docs/workers/languages/typescript/index.mdx
+++ b/src/content/docs/workers/languages/typescript/index.mdx
@@ -39,9 +39,71 @@ If you would like to commit the file to git, you can provide a custom path. Here
 wrangler types --experimental-include-runtime="./runtime.d.ts"
 ```
 
-Note that the `--experimental-include-runtime` command replaces the need for the `@cloudflare/workers-types` package, so if you do have that package installed, you should remove it to avoid any potential for conflict.
+{{<Aside type="warning">}}
+
+To ensure that your types are always up-to-date, make sure to run `wrangler types --experimental-include-runtime` after any changes to your `wrangler.toml` file.
+
+{{</Aside>}}
 
 See [the full list of available flags](/workers/wrangler/commands/#types) for more details.
+
+#### Migrating from `@cloudflare/workers-types` to `wrangler types --experimental-include-runtime`
+
+The `@cloudflare/workers-types` package provides runtime types according to your compatibility date, and therefore it is superceeded by the `wrangler types --experimental-include-runtime` command.
+
+Here are the steps to switch from `@cloudflare/workers-types` to using wrangler types with the experimental runtime inclusion:
+
+##### 1. Uninstall `@cloudflare/workers-types`
+
+```bash
+npm uninstall @cloudflare/workers-types
+```
+
+##### 2. Generate runtime types using wrangler
+
+```bash
+wrangler types --experimental-include-runtime
+```
+
+This will generate a `.d.ts` file, saved to `.wrangler/types/runtime.d.ts` by default.
+
+##### 3. Update your `tsconfig.json` to include the generated types
+
+```json
+{
+  "compilerOptions": {
+    "types": ["./.wrangler/types/runtime"]
+  }
+}
+```
+
+Note that if you have specified a custom path for the runtime types file, you should use that in your `compilerOptions.types` array instead of the default path.
+
+##### 4. Update your scripts and CI pipelines
+
+When switching to `wrangler types --experimental-include-runtime`, you'll want to ensure that your development process always uses the most up-to-date types. The main thing to remember here is that - regardless of your specific framework or build tools - you should run the `wrangler types --experimental-include-runtime` command before any development tasks that rely on TypeScript. This ensures your editor and build tools always have access to the latest types.
+
+Most projects will have existing build and development scripts scripts, as well as some type-checking. In the example below, we're adding the `wrangler types --experimental-include-runtime` before the type-checking script in the project:
+
+```json
+{
+  "scripts": {
+    "dev": "your-existing-dev-command",
+    "build": "your-existing-build-command",
+    "type-check": "wrangler types --experimental-include-runtime && tsc"
+  }
+}
+```
+
+In CI, you may have a separate build and test commands. It is best practice to run `wrangler types --experimental-include-runtime` before other CI commands. For example:
+
+```text
+- run: npm run generate-types
+- run: npm run build
+- run: npm test
+```
+
+By integrating the `wrangler types --experimental-include-runtime` command into your workflow this way, you ensure that your development environment, builds, and CI processes always use the most accurate and up-to-date types for your Cloudflare Worker, regardless of your specific framework or tooling choices.
 
 ### Known issues
 
--- a/src/content/docs/workers/languages/typescript/index.mdx
+++ b/src/content/docs/workers/languages/typescript/index.mdx
@@ -6,7 +6,6 @@ sidebar:
 head:
   - tag: title
     content: Write Cloudflare Workers in TypeScript
-
 ---
 
 import { TabItem, Tabs } from "~/components";
@@ -41,11 +40,7 @@ If you would like to commit the file to git, you can provide a custom path. Here
 wrangler types --experimental-include-runtime="./runtime.d.ts"
 ```
 
-{{<Aside type="warning">}}
-
-To ensure that your types are always up-to-date, make sure to run `wrangler types --experimental-include-runtime` after any changes to your config file.
-
-{{</Aside>}}
+**Note: To ensure that your types are always up-to-date, make sure to run `wrangler types --experimental-include-runtime` after any changes to your config file.**
 
 See [the full list of available flags](/workers/wrangler/commands/#types) for more details.
 
@@ -95,9 +90,9 @@ This will generate a `.d.ts` file, saved to `.wrangler/types/runtime.d.ts` by de
 
 ```json
 {
-  "compilerOptions": {
-    "types": ["./.wrangler/types/runtime"]
-  }
+	"compilerOptions": {
+		"types": ["./.wrangler/types/runtime"]
+	}
 }
 ```
 
@@ -111,11 +106,11 @@ Most projects will have existing build and development scripts scripts, as well 
 
 ```json
 {
-  "scripts": {
-    "dev": "existing-dev-command",
-    "build": "-existing-build-command",
-    "type-check": "wrangler types --experimental-include-runtime && tsc"
-  }
+	"scripts": {
+		"dev": "existing-dev-command",
+		"build": "-existing-build-command",
+		"type-check": "wrangler types --experimental-include-runtime && tsc"
+	}
 }
 ```
 
@@ -177,7 +172,7 @@ For more information, refer to [this GitHub issue](https://github.com/cloudflare
 
 ### Resources
 
-* [TypeScript template](https://github.com/cloudflare/workers-sdk/tree/main/templates/worker-typescript)
-* [@cloudflare/workers-types](https://github.com/cloudflare/workers-types)
-* [Runtime APIs](/workers/runtime-apis/)
-* [TypeScript Examples](/workers/examples/?languages=TypeScript)
+- [TypeScript template](https://github.com/cloudflare/workers-sdk/tree/main/templates/worker-typescript)
+- [@cloudflare/workers-types](https://github.com/cloudflare/workers-types)
+- [Runtime APIs](/workers/runtime-apis/)
+- [TypeScript Examples](/workers/examples/?languages=TypeScript)
--- a/src/content/docs/workers/languages/typescript/index.mdx
+++ b/src/content/docs/workers/languages/typescript/index.mdx
@@ -119,11 +119,39 @@ Most projects will have existing build and development scripts scripts, as well 
 
 In CI, you may have a separate build and test commands. It is best practice to run `wrangler types --experimental-include-runtime` before other CI commands. For example:
 
+<Tabs> <TabItem label="npm">
+
 ```yaml
 - run: npm run generate-types
 - run: npm run build
 - run: npm test
 ```
+
+</TabItem> <TabItem label="yarn">
+
+```yaml
+- run: yarn generate-types
+- run: yarn build
+- run: yarn test
+```
+
+</TabItem> <TabItem label="pnpm">
+
+```yaml
+- run: pnpm run generate-types
+- run: pnpm run build
+- run: pnpm test
+```
+
+</TabItem> <TabItem label="bun">
+
+```yaml
+- run: bun run generate-types
+- run: bun run build
+- run: bun test
+```
+
+</TabItem> </Tabs>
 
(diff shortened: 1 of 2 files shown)


irvinebroque (commented):
Think there should be distinct section here:

https://developers.cloudflare.com/workers/languages/typescript/

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,32 @@ meta:
 
 TypeScript is a first-class language on Cloudflare Workers. Cloudflare publishes type definitions to [GitHub](https://github.com/cloudflare/workers-types) and [npm](https://www.npmjs.com/package/@cloudflare/workers-types) (`npm install -D @cloudflare/workers-types`). All APIs provided in Workers are fully typed, and type definitions are generated directly from [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.
 
+### Dynamic Runtime Types (Experimental)
+
+Your runtime APIs are dynamic - their exact configuration depends on:
```
irvinebroque: Think this should explain why to do this

content/workers/wrangler/commands.md
```diff
@@ -2816,6 +2822,20 @@ Generate types from bindings and module rules in configuration.
 wrangler types [<PATH>] [OPTIONS]
 ```
 
+{{<Aside type="warning">}}
+
+The `--experimental-runtime` flag dynamically generates runtime types according to the `compatibility_date` and `compatibility_flags` defined in your `wrangler.toml`.
+
+It is a replacement for the `@cloudflare/workers-types` package, so that package should be uninstalled to avoid any potential conflict.
```
irvinebroque:
```suggestion
It is a replacement for the [`@cloudflare/workers-types` package](https://www.npmjs.com/package/@cloudflare/workers-types), so that package should be uninstalled to avoid any potential conflict.
```

content/workers/wrangler/commands.md
```diff
@@ -2816,6 +2822,20 @@ Generate types from bindings and module rules in configuration.
...
+
+After running the command, you must add the path of the generated runtime types file to your `tsconfig` `compilerOptions.types` array.
```
irvinebroque:
```suggestion
After running the command, you must add the path of the generated runtime types file to the [`compilerOptions.types` field](https://www.typescriptlang.org/tsconfig/#types) in your project's [tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) file.
```

content/workers/wrangler/commands.md
```diff
@@ -2827,6 +2847,16 @@ wrangler types [<PATH>] [OPTIONS]
   - The name of the interface to generate for the environment object.
   - Not valid if the Worker uses the Service Worker syntax.
 
+- `--experimental-runtime` {{<type>}}string{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}} {{<prop-meta>}}(default: `./.wrangler/types/runtime.d.ts`){{</prop-meta>}}
+  - The path of the generated runtime types file.
```
irvinebroque: To me, `PATH` is already documented above, and this makes it sound a bit like there are two _different_ paths that you must specify?

Screenshot 2024-08-20 at 9 00 32 PM

This is really just a flag that changes the behavior of the command, but doesn't add new arguments, right?

So what I'd expect here is a description of what is different that happens when you use this flag.

Maybe could just move the aside above into this section, since that content is about this flag anyways?

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
 
 TypeScript is a first-class language on Cloudflare Workers. Cloudflare publishes type definitions to [GitHub](https://github.com/cloudflare/workers-types) and [npm](https://www.npmjs.com/package/@cloudflare/workers-types) (`npm install -D @cloudflare/workers-types`). All APIs provided in Workers are fully typed, and type definitions are generated directly from [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.
 
+### Dynamic Runtime Types (Experimental)
```
irvinebroque:
```suggestion
### Generate Types that match your Worker's configuration (Experimental)
```

Something like this to be more concrete?

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+There is a possibility that the types defined in `@cloudflare/workers-types` might not exactly match your project's runtime APIs. This is because runtime APIs are dynamic, and their exact configuration depends on:
```
irvinebroque:
```suggestion
Cloudflare continuously improves [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.

Some updates change the TypeScript types that APIs return. For example, the [`urlsearchparams_delete_has_value_arg`](/workers/configuration/compatibility-dates/#urlsearchparams-delete-and-has-value-argument) compatibility flag adds optional arguments to some methods, in order to support new additions to the WHATWG URL standard API.

This means the correct TypeScript types for your Worker depend on:
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+1. The `compatibility_date` defined in your `wrangler.toml`.
```
irvinebroque:
```suggestion
1. Your Worker's [Compatibility Date](/workers/configuration/compatibility-dates/)
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+2. Any `compatibility_flags` defined in your `wrangler.toml`.
```
irvinebroque:
```suggestion
2. Your Worker's [Compatibility Flags](/workers/configuration/compatibility-dates/#compatibility-flags)
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+For example, if you have `compatibility_flags = ["nodejs_als"]` in your `wrangler.toml`, then the runtime will allow you to use the [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) class in your worker code, but you will not see this reflected in the type definitions in `@cloudflare/workers-types`.
```
irvinebroque: To me this implies that if I set `compatibility_flags = ["nodejs_compat"]` — then I'll get types back for Node.js APIs (because ALS is a subset of this)

But that's not something that we support (we should say this explicitly!)

Is there maybe an example of a different compat flag that we could use here, for clarity?

And also...is there an explicit reason why `nodejs_compat` doesn't yet work but `nodejs_als` does?

irvinebroque:
```suggestion
For example, if you have `compatibility_flags = ["nodejs_als"]` in your `wrangler.toml`, then the [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) API can be used in your Worker, but you will not see this reflected in the type definitions in `@cloudflare/workers-types`.
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+In order to solve this issue, and to ensure that your type definitions are always up-to-date with your compatibility settings, you can dynamically generate the runtime types (as of `wrangler 3.66.1`):
```
irvinebroque:
```suggestion
To ensure that the type definitions you rely on match runtime behavior, as of `wrangler 3.66.1`, you can run a Wrangler command to generate TypeScript types based on your compatibility date and flags:
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+```bash
+wrangler types --experimental-runtime
+
+# You could also use the alias
+wrangler types --x-runtime
```
irvinebroque: I would suggest picking one and sticking with it in context of these docs. If the Wrangler docs on commands spell out both, that seems fine.

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+```
+
+This will generate a `d.ts` file and (by default) save it to `.wrangler/types/runtime.d.ts`. You will be prompted in the command's output to add that file to your `tsconfig.json`'s `compilerOptions.types` array.
```
irvinebroque:
```suggestion
This will generate a `d.ts` file and (by default) save it to `.wrangler/types/runtime.d.ts`. After running the command, you must add the path of the generated runtime types file to the [`compilerOptions.types` field](https://www.typescriptlang.org/tsconfig/#types) in your project's [tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) file.
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+If you would like to commit the file to git, you can provide a custom path. Here, for instance, the `runtime.d.ts` file will be saved to the root of your project:
+
+```bash
+wrangler types --x-include-runtime="./runtime.d.ts"
+```
+
+Note that the `--x-runtime` command replaces the need for the `@cloudflare/workers-types` package, so that package should be uninstalled to avoid any potential for conflict.
```
irvinebroque:
```suggestion
Note that the `--x-runtime` command replaces the need for the `@cloudflare/workers-types` package, you should uninstall that package to avoid conflicts.
```

content/workers/languages/typescript/_index.md
```diff
@@ -10,6 +10,73 @@ meta:
...
+
+### Type Checking (Experimental)
```
irvinebroque: Not reviewing below this since removing for now

content/workers/wrangler/commands.md
irvinebroque: OH! I get it. Okay we're going to need to be super explicit about this.

So with this flag, there _are_ two _different_ paths that you can specify, and each has their own default.

irvinebroque:
We should add a changelog entry to this that links back to these docs. (I know that Wrangler automatically generates changelog entries but they can get missed in the noise)

content/workers/wrangler/commands.md
dario-piotrowicz: is this assuming that the package is always installed? 🤔 

```suggestion
It is a replacement for the `@cloudflare/workers-types` package, so that package should not be present as a dependency to avoid any potential conflict.
```

content/workers/wrangler/commands.md
dario-piotrowicz: Are there plans to update the tsconfig automatically?

It's not the simplest file to deal with since is jsonc 😓  but we've done this in the past (e.g. [c3](https://github.com/cloudflare/workers-sdk/blob/419e9f0f12716f3741afe50ece642fdd95e1e3d1/packages/create-cloudflare/src/workers.ts#L32)) with good results

content/workers/wrangler/commands.md
dario-piotrowicz: yes, it took it a bit for me to click too, especially for newcomers I think it could be pretty confusing to understand what there are different d.ts files getting generated...

Also.... has a "compact" option been considered? could we just shove everything into a single `d.ts` file? I'm asking this because I am guessing that the userland d.ts file has little to no meaning without the runtime d.ts file right? so it's not like something that that valuable to have in userland/commit is it? (unless the expectation is that users are supposed to likely manually touch the file?)

content/workers/wrangler/commands.md
```diff
@@ -2827,6 +2847,16 @@ wrangler types [<PATH>] [OPTIONS]
...
+  - Leave the path blank to use the default option, e.g. `npx wrangler types --x-include-runtime`
+  - A custom path must be relative to the project root.
+  - A custom path must have a `d.ts` extension.
+
+- `--experimental-check` {{<type>}}boolean{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
```
dario-piotrowicz: I am not completely sure I understand the value of this flag, is it just like adding `&& tsc` at the end of the command?

content/workers/wrangler/commands.md
```diff
@@ -824,6 +824,12 @@ As of Wrangler v3.2.0, `wrangler dev` is supported by any Linux distributions pr
   - Specify Wrangler's logging level.
 - `--show-interactive-dev-session` {{<type>}}boolean{{</type>}} {{<prop-meta>}}(default: true if the terminal supports interactivity){{</prop-meta>}} {{<prop-meta>}}optional{{</prop-meta>}}
   - Show the interactive dev session.
+- `--experimental-types` {{<type>}}boolean{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
+  - Run `wrangler types` in the background.
```
dario-piotrowicz: I imagine this runs in a sort of watch mode and re runs when there are changes in the config file? should we clarify when it runs the command?

content/workers/wrangler/commands.md
```diff
@@ -824,6 +824,12 @@ As of Wrangler v3.2.0, `wrangler dev` is supported by any Linux distributions pr
...
+- `--experimental-types-runtime`  {{<type>}}string{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}} {{<prop-meta>}}(default: `./.wrangler/types/runtime.d.ts`){{</prop-meta>}}
+  - Alias for `wrangler types --x-runtime`
```
dario-piotrowicz: I think this could be clarified, I _think_ that this is saying that this runs `wrangler types --x-runtime` in the background? also since the flag is not shorthanded (i.e. it starts with `--experimental`) shouldn't the `wrangler types` one also be in its extended form? (i.e. not `--x`)

content/workers/wrangler/commands.md
```diff
@@ -824,6 +824,12 @@ As of Wrangler v3.2.0, `wrangler dev` is supported by any Linux distributions pr
...
+- `--experimental-types-check` {{<type>}}boolean{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}}
+  - Alias for `wrangler types -x-check`
```
dario-piotrowicz: same as above, I think this could be clarified

content/workers/wrangler/commands.md
```diff
@@ -2816,6 +2822,20 @@ Generate types from bindings and module rules in configuration.
 wrangler types [<PATH>] [OPTIONS]
 ```
 
+{{<Aside type="warning">}}
+
+The `--experimental-runtime` flag dynamically generates runtime types according to the `compatibility_date` and `compatibility_flags` defined in your `wrangler.toml`.
```
dario-piotrowicz: not sure, but it might be worth it to try to generically specify it as config file instead of `wrangler.toml` in case we'll want to push for `wrangler.json` etc.? 🤔 

```suggestion
The `--experimental-runtime` flag dynamically generates runtime types according to the `compatibility_date` and `compatibility_flags` defined in your [config file](https://aj-runtime-types.cloudflare-docs-7ou.pages.dev/workers/wrangler/configuration/).
```

@penalosa wdyt? (it kinda seems that `wrangler.toml` is alredy used everywhere anyways 😅😓)

content/workers/languages/typescript/_index.md
dario-piotrowicz: I'm also not that convinced that `--x-runtime` is that much shorter/better than `--experimental-runtime` I'd personally just go with the extended version 🤔

andyjessop: This is here in order to maintain consistency with other experimental flags in wrangler. As far as I know, they all have (or should have) an `--x` alias.

content/workers/wrangler/commands.md
andyjessop: > Also.... has a "compact" option been considered? could we just shove everything into a single d.ts file?

Something like that could be considered in the future, if we see user issues. The reason for the current implementation is that it's decoupled and minimal (and by default the runtime types are not committed to git as they're saved to a hidden folder), making the implementation as low impact as possible.

andyjessop: > I'm asking this because I am guessing that the userland d.ts file has little to no meaning without the runtime d.ts file right?

I don't think that's the case, because that's the current situation, right? Or rather, that's the situation without the runtime types flag. Users can just run `wrangler types` to generate their Env types.

content/workers/wrangler/commands.md
andyjessop: Maybe that could be worded differently? We mean to say that the user should uninstall it if they do have it currently installed. But I think we also want to state anyway that it is a replacement.

andyjessop:
> Are there plans to update the tsconfig automatically?

That could happen! We don't have plans for it right now, but if we have precedent for it in c3 then maybe it's something to consider.

content/workers/wrangler/commands.md
andyjessop: Yeah, we're removing that - we jumped the gun on that implementation and are now walking it back (the PR never landed and I'm just removing it from these draft docs).

content/workers/wrangler/commands.md
andyjessop: This, too, is being put on hold until we get more user feedback.

content/workers/wrangler/commands.md
dario-piotrowicz: > I don't think that's the case, because that's the current situation, right? Or rather, that's the situation without the runtime types flag. Users can just run wrangler types to generate their Env types.

Yes, you're right that's the current behavior, but it relies on `@cloudflare/workers-types`

So I was asking assuming that we don't want people to use `@cloudflare/workers-types`, without it a userland generated d.ts file would have little meaning (e.g. it could have a `MY_KV: KVNamespace` but no definition of what `KVNamespace` is) no?

content/workers/languages/typescript/_index.md
dario-piotrowicz: ah ok sorry I didn't know/wasn't sure about that 🙂 

if it's an established pattern then yeah I agree that we should just follow it 👍

content/workers/wrangler/commands.md
dario-piotrowicz: sure I am fine with whatever you feel more appropriate 🙂 

my concern with the existing version is just that we say that the package should be uninstalled, but the package could have not been installed in the first place (so just saying to unconditionally uninstall it feels incorrect and potentially confusing to me)

content/workers/wrangler/commands.md
andyjessop: I don't believe it does rely on the `@cloudflare/workers-types` package.

If I start a new project with c3, then remove the `@cloudflare/workers-types` package:

Screenshot 2024-08-23 at 12 32 08

Then add a KV namespace:

Screenshot 2024-08-23 at 12 32 12

When I run `wrangler types`, I still get the interface generated and saved to `workers-configuration.d.ts`:

Screenshot 2024-08-23 at 12 32 19

src/content/docs/workers/languages/typescript/index.mdx
```diff
@@ -6,13 +6,152 @@ sidebar:
...
+import { TabItem, Tabs } from "~/components";
+
 ## TypeScript
 
 TypeScript is a first-class language on Cloudflare Workers. Cloudflare publishes type definitions to [GitHub](https://github.com/cloudflare/workers-types) and [npm](https://www.npmjs.com/package/@cloudflare/workers-types) (`npm install -D @cloudflare/workers-types`). All APIs provided in Workers are fully typed, and type definitions are generated directly from [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.
 
+### Generate types that match your Worker's configuration (experimental)
+
+Cloudflare continuously improves [workerd](https://github.com/cloudflare/workerd), the open-source Workers runtime.
+Some updates change the TypeScript types that APIs return. For example, the [`urlsearchparams_delete_has_value_arg`](/workers/configuration/compatibility-dates/#urlsearchparams-delete-and-has-value-argument) compatibility flag adds optional arguments to some methods, in order to support new additions to the WHATWG URL standard API.
```
dario-piotrowicz: I'm not a huge fan of the wording here, `Some updates change the TypeScript types` makes it sound like workerd has built-in TypeScript support to me 😕 

What about something like:
```
Changes in workerd can introduce JavaScript API changes thus changing the respective TypeScript types.
```

src/content/docs/workers/languages/typescript/index.mdx
```diff
@@ -6,13 +6,152 @@ sidebar:
...
+
+This means the correct TypeScript types for your Worker depend on:
+
+1. Your Worker's [Compatibility Date](/workers/configuration/compatibility-dates/).
+2. Your Worker's [Compatibility Flags](/workers/configuration/compatibility-dates/#compatibility-flags).
+
+For example, if you have `compatibility_flags = ["nodejs_als"]` in your `wrangler.toml`, then the runtime will allow you to use the [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) class in your worker code, but you will not see this reflected in the type definitions in `@cloudflare/workers-types`.
```
dario-piotrowicz:
```suggestion
For example, if you have `compatibility_flags = ["nodejs_als"]` in your `wrangler.toml`, then the runtime will allow you to use the [`AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage) class in your worker code, but you will not see this reflected in the type definitions in `@cloudflare/workers-types`.
```

src/content/docs/workers/languages/typescript/index.mdx
```diff
@@ -6,13 +6,152 @@ sidebar:
...
+
+If you would like to commit the file to git, you can provide a custom path. Here, for instance, the `runtime.d.ts` file will be saved to the root of your project:
+
+```bash
+wrangler types --experimental-include-runtime="./runtime.d.ts"
+```
+
+**Note: To ensure that your types are always up-to-date, make sure to run `wrangler types --experimental-include-runtime` after any changes to your config file.**
+
+See [the full list of available flags](/workers/wrangler/commands/#types) for more details.
```
dario-piotrowicz, line 45:
```suggestion
See [the full list of available flags](/workers/configuration/compatibility-dates/#compatibility-flagss) for more details.
```

src/content/docs/workers/languages/typescript/index.mdx
```diff
@@ -6,13 +6,152 @@ sidebar:
...
+
+#### Migrating from `@cloudflare/workers-types` to `wrangler types --experimental-include-runtime`
+
+The `@cloudflare/workers-types` package provides runtime types according to your compatibility date, but is superceeded by the `wrangler types --experimental-include-runtime` command.
```
dario-piotrowicz: `according to your compatibility date` sounds very ambiguous to me and can give the impression the `@cloudflare/workers-types` does take into account the compat date from the config file

I think we should clarify that that's not the case

what about something like this for example?:
```suggestion
The `@cloudflare/workers-types` package provides runtime types according to the used [date entrypoint](https://github.com/cloudflare/workerd/tree/main/npm/workers-types#compatibility-dates), but is superceeded by the `wrangler types --experimental-include-runtime` command.
```

src/content/docs/workers/languages/typescript/index.mdx
```diff
@@ -6,13 +6,152 @@ sidebar:
...
+
+1. Your Worker's [Compatibility Date](/workers/configuration/compatibility-dates/).
+2. Your Worker's [Compatibility Flags](/workers/configuration/compatibility-dates/#compatibility-flags).
+
+For example, if you have `compatibility_flags = ["nodejs_als"]` in your `wrangler.toml`, then the runtime will allow you to use the [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) class in your worker code, but you will not see this reflected in the type definitions in `@cloudflare/workers-types`.
+
+In order to solve this issue, and to ensure that your type definitions are always up-to-date with your compatibility settings, you can dynamically generate the runtime types (as of `wrangler 3.66.0`):
+
+```bash
+wrangler types --experimental-include-runtime
```
dario-piotrowicz: I don't think that this is correct, as this would not work unless you have wrangler globally installed (which we discourage)

I think this should be `npx wrangler ...` (with all the other pm alternatives if needed)

(This applies to the other similar commands in this PR)

src/content/docs/workers/languages/typescript/index.mdx
```diff
@@ -6,13 +6,152 @@ sidebar:
...
+```
+
+This will generate a `.d.ts` file, saved to `.wrangler/types/runtime.d.ts` by default.
+
+##### Update your `tsconfig.json` to include the generated types
+
+```json
+{
+	"compilerOptions": {
+		"types": ["./.wrangler/types/runtime"]
```
dario-piotrowicz, line 94: Nit:
I am not sure what's better, putting this into `types` or simply including `.wrangler/types` in the tsconfig [`include` option](https://www.typescriptlang.org/tsconfig/#include)

e.g.:
```ts
	"include": ["worker-configuration.d.ts", ".wrangler/types", "src/**/*.ts"]
```

According to the [TypeScript docs](https://www.typescriptlang.org/tsconfig/#types) `types` seems more appropriate for packages 🤔 

`include` has the nice benefit that we could change the `d.ts` generation (file name, number of files, etc...) and things would still work for users without them needing to change their tsconfig

What do you think?

src/content/docs/workers/wrangler/commands.mdx
```diff
@@ -2592,13 +2593,36 @@ Generate types from bindings and module rules in configuration.
 wrangler types [<PATH>] [OPTIONS]
 ```
 
-- `PATH` string (default: `worker-configuration.d.ts`)
+{{<Aside type="warning">}}
+
+The `--experimental-include-runtime` flag dynamically generates runtime types according to the `compatibility_date` and `compatibility_flags` defined in your [config file](/workers/wrangler/configuration/).
+
+It is a replacement for the [`@cloudflare/workers-types` package](https://www.npmjs.com/package/@cloudflare/workers-types), so that package should be uninstalled to avoid any potential conflict.
```
dario-piotrowicz:
```suggestion
It is a replacement for the [`@cloudflare/workers-types` package](https://www.npmjs.com/package/@cloudflare/workers-types), so that package, if installed, should be uninstalled to avoid any potential conflict.
```

src/content/docs/workers/wrangler/commands.mdx
```diff
@@ -2592,13 +2593,36 @@ Generate types from bindings and module rules in configuration.
...
+  - The path to where **the `Env` types** for your Worker will be written.
+  - The path must have a `d.ts` extension.
 
 - `--env-interface` string (default: `Env`)
   - The name of the interface to generate for the environment object.
   - Not valid if the Worker uses the Service Worker syntax.
 
-{/* <!--TODO Add examples of DTS generated output --> */}
+- `--experimental-include-runtime` {{<type>}}string{{</type>}} {{<prop-meta>}}optional{{</prop-meta>}} {{<prop-meta>}}(default: `./.wrangler/types/runtime.d.ts`){{</prop-meta>}}
+  - The path to where the  **runtime types** file will be written.
```
dario-piotrowicz:
```suggestion
  - The path to where the **runtime types** file will be written.
```

irvinebroque (approved):

src/content/docs/workers/languages/typescript/index.mdx
andyjessop: Yep, I think that makes sense 👍

src/content/docs/workers/languages/typescript/index.mdx
andyjessop: This is actually for the `wrangler types` command flags, not the compatibility flags.

src/content/docs/workers/languages/typescript/index.mdx
andyjessop: What about something like this? To be more explicit.

```
The `@cloudflare/workers-types` package provides runtime types for each distinct [compatibility date](https://github.com/cloudflare/workerd/tree/main/npm/workers-types#compatibility-dates), and specified by the user in their `tsconfig.json`. But this package is superseded by the `wrangler types --experimental-include-runtime` command.
```

src/content/docs/workers/languages/typescript/index.mdx
andyjessop: Good call 👍

src/content/docs/workers/languages/typescript/index.mdx
andyjessop: I think the problem is that if you don't have an `include` field, then this addition will restrict the included files to only the runtime types declaration file.

src/content/docs/workers/languages/typescript/index.mdx
dario-piotrowicz: ahhh ok I see 😅👍

src/content/docs/workers/languages/typescript/index.mdx
dario-piotrowicz: good point 😕 :+1:

src/content/docs/workers/languages/typescript/index.mdx
dario-piotrowicz: sounds good to me 😄

dario-piotrowicz (approved):
Looks great to me! thanks for addressing my comments @andyjessop 😄

PR #15787, opened 2024-07-24
Document `wrangler.json` as an alternative to `wrangler.toml`

penalosa:
### Summary

This ensures that all `wrangler.toml` snippets in the docs have a corresponding `wrangler.json` tab, showing the equivalent config in JSON format (supported by Wrangler with the `-j` flag). The JSON config has a comment explaining how to make Wrangler recognise the format.

No additional work is required to add these tabs—they're generated automatically for every `toml` snippet that references `filename: wrangler.toml` (to ensure we don't accidentally break e.g. `cargo.toml` files).

Some other things have also been changed/improved:
- The tab system has been reworked to no longer require IDs, instead relying on child indexes in the DOM. These IDs were based on the hash of tab contents, which caused unexpected behaviour with multiple tabs on the same page with the same contents (as this change introduced in a few places)

### Screenshots (optional)

Screenshot 2024-07-24 at 21 22 07

### DIFF
--- a/assets/events.ts
+++ b/assets/events.ts
@@ -15,7 +15,7 @@ export function $focus(elem: HTMLElement, bool: boolean) {
   if (SEARCH_ID && SEARCH_ID.test(elem.id)) {
     SEARCH_INPUT = elem;
 
-    if(!elem.parentElement || !elem.parentElement.parentElement) return;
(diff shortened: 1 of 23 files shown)


irvinebroque (commented):
We'll talk about this one first before anything lands. Clever technical approach to automatically generating both versions.

KimJ15:

PR #15807, opened 2024-07-25
[ZVM] Zone IDs note

patriciasantaana:
### Summary

Mention the need to use zone tags when using the API on different versions

### Documentation checklist


content/version-management/how-to/versions.md
```diff
@@ -34,6 +34,12 @@ Your zone configurations are split up into two areas: **Global Configuration** a
     - Editable when not applied to a [read-only environment](/version-management/reference/read-only-environments/).
     - Applied when [associated with an environment](/version-management/how-to/environments/#change-environment-version).
 
+{{<Aside type="note">}}
+
+To use the API for a different version, you will need to use a different zone tag.
```
pedrosousa: If we're talking about the URL piece containing the zone identifier (`.../{zone_id}/...`), we call it "zone ID" in the docs instead of "zone tag" (which is an internal name).
If there's a different "zone tag" concept in Version Management that I don't know about, feel free to ignore. :)

patriciasantaana: Hey Pedro! Thanks for bringing this up. Just checked with the team and it is zone ID

ranbel (approved):

PR #15821, opened 2024-07-26
[Hyperdrive]Update metrics.md

luishcardoso:
Hi team, 

### Summary

The HIT, MISS as per https://developers.cloudflare.com/cache/concepts/cache-responses/#hit

Thanks!


elithrar (changes requested):
@nenizera - what motivated this change?

Hyperdrive isn't exposing HTTP response headers. The metrics response from the GraphQL endpoint are all lowercased.

PR #15830, opened 2024-07-26
Fix documented duration limit for bulk compute types

a-robinson:
Cron triggers, DO alarms, and Queue consumers all have a 15 minute duration limit that wasn't documented here.

I didn't add DO alarms to the "CPU time" column because it looks like alarms aren't actually given any more CPU time than normal DO invocations, the only thing that makes them unique is having the duration limit applied.

This section now renders as:

Screenshot 2024-07-26 at 1 25 55 PM

@maheshwarip @vy-ton @irvinebroque


a-robinson:
Re-pushed to capitalize "Alarm" for consistency, although I'm ambivalent about whether alarm/consumer should be capitalized.

kodster28 (approved):

a-robinson:
Thanks for the review!

Just in the interest of clarity, I don't have the ability to merge this, so it's still waiting on action from someone else.

tanushree-sharma:
Merged.

PR #15842, opened 2024-07-28
Update troubleshooting-cloudflare-5xx-errors.md

gpoulton:
Spent the last day trying to debug a case where blocking http traffic causes some browsers (eg Safari on iOS) to display CloudFlare's 522 page instead of retrying on https.  I'm mentioning it here in case it happens to other people.

### Summary

Added one line to the 522 section.

### Screenshots (optional)

n/a

### Documentation checklist


ngayerie:
Hello,
If the origin is blocking HTTP port 80 an HTTP Error 521 would be thrown then (as the connection between Cloudflare and the origin would be terminated instead of timing out).
The behavior you're describing is odd. 
Have you raised a Community post about it?
What was the fix in your case?

gpoulton:
Maybe it's some weirdness in the specific way my ISP (Ionos)'s firewall
works? The fix was to enable port 80 on the firewall.

cheers,
Gareth

KimJ15:

PR #15857, opened 2024-07-29
Pcx 11620

jason-cf:
### Summary

- YT channel renamed
- Craig reshot this video, so just replacing the id with the new one

### DIFF
--- a/data/external-resources/videos.yml
+++ b/data/external-resources/videos.yml
@@ -98,14 +98,14 @@ entries:
   difficulty: Advanced
   content_type: 🎥 Video
   pcx_content_type: tutorial
-  - link: https://www.youtube.com/watch?v=CHfKeFakGAI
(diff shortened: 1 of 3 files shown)


marciocloudflare (approved):

data/external-resources/videos.yml
```diff
@@ -97,4 +97,15 @@ entries:
   updated: 2023-12-14
   difficulty: Advanced
   content_type: 🎥 Video
+  pcx_content_type: tutorial
+  - link: https://www.youtube.com/watch?v=CHfKeFakGAI
```
pedrosousa:
```suggestion
- link: https://www.youtube.com/watch?v=CHfKeFakGAI
```

jason-cf: thanks @pedrosousa

pedrosousa: I think the lines after this one must have a 2-space indent, but let's see if it builds as it is.

PR #15970, opened 2024-08-04
images: fix redirect to pricing

elithrar:
Fix a missing trailing slash that breaks the redirects.

cc @kodster28 - can we better link the `_redirects` file so this doesn't happen? Not the first time.


elithrar:
- redirects to `/images/platform/pricing/`: https://silverlock-fix-images-pricin.cloudflare-docs-7ou.pages.dev/images/pricing/
- does not: https://silverlock-fix-images-pricin.cloudflare-docs-7ou.pages.dev/images/pricing

Do we need another redirect here?

KianNH (approved):

PR #15983, opened 2024-08-05
[Chore] Update wrangler generate to c3

kodster28:
### Summary

Update an outdated command with a new version (and use the favored approach to Rust, as discussed in #15834).

**Update**: Have fixed the examples that I pushed earlier, by combo of correctly setting the `-- --template <GITHUB_URL>` option + adding `@latest` to all the commands.

### DIFF
--- a/content/cloudflare-one/policies/access/external-evaluation.md
+++ b/content/cloudflare-one/policies/access/external-evaluation.md
@@ -29,7 +29,7 @@ You can set up External Evaluation rules using any API service, but to get start
 1. Open a terminal and clone our example project.
 
    ```sh
-   $ npm create cloudflare my-worker --template=https://github.com/cloudflare/workers-access-external-auth-example
+   $ npm create cloudflare my-worker -- --template=https://github.com/cloudflare/workers-access-external-auth-example
    ```
 
 2. Go to the project directory.
--- a/content/cloudflare-one/policies/access/external-evaluation.md
+++ b/content/cloudflare-one/policies/access/external-evaluation.md
@@ -29,7 +29,7 @@ You can set up External Evaluation rules using any API service, but to get start
 1. Open a terminal and clone our example project.
 
    ```sh
-   $ npm create cloudflare@latest my-worker -- --template=https://github.com/cloudflare/workers-access-external-auth-example
+   $ npm create cloudflare@latest my-worker -- --template https://github.com/cloudflare/workers-access-external-auth-example
    ```
 
 2. Go to the project directory.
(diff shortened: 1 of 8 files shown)


content/cloudflare-one/policies/access/external-evaluation.md
```diff
@@ -20,15 +20,16 @@ You can set up External Evaluation rules using any API service, but to get start
...
+- Install [Node.js](https://nodejs.org/en/)
 - Application protected by Access
 
 ### 1. Create a new Worker
 
 1. Open a terminal and clone our example project.
 
    ```sh
-   $ wrangler generate my-worker https://github.com/cloudflare/workers-access-external-auth-example
+   $ npm create cloudflare my-worker --template=https://github.com/cloudflare/workers-access-external-auth-example
```
elithrar: This doesn't work. It just silently ignores the `--template` and asks you whatever it wants to ask you.

```
➜  npm create cloudflare my-worker --template=https://github.com/cloudflare/workers-access-external-auth-example

using create-cloudflare version 2.0.8

╭ Create an application with Cloudflare Step 1 of 3
│
├ Where do you want to create your application?
│ dir my-worker
│
├ What type of application do you want to create?
│ type "Hello World" script
│
├ Do you want to use TypeScript?
│ typescript yes
│
├ Copying files from "simple" template
│
╰ Application created
```

elithrar (changes requested):

content/cloudflare-one/policies/access/external-evaluation.md
kodster28: Okay, should be updated by the latest 2 commits.

elithrar: `--template` isn't a documented flag here.

image

elithrar: From the source it looks like it is - https://github.com/cloudflare/workers-sdk/blob/eb6f249510a9d36a91d459a2c5918f65b74d2c36/packages/create-cloudflare/src/helpers/args.ts#L176-L188 - but it's not exposed in the help text and is swallowed up.

Correcting to use `npm create cloudflare use-a-template -- --template=https://github.com/cloudflare/workers-access-external-auth-example` with `--` to pass the flag to the `create-cloudflare` SDK instead of `npm` yields no change.


image

kodster28: Okay, these should be addressed now (combo of fixing the arg structure + specifying `@latest` across all our `npm create cloudflare` instances)

KianNH (approved):

elithrar (approved):

PR #15985, opened 2024-08-05
[KV] Update KV specific wrangler docs to match global wrangl…

thomasgauvin:
https://developers.cloudflare.com/kv/reference/kv-commands/ has gotten out of date from https://developers.cloudflare.com/workers/wrangler/commands/

Mostly around the absence of --local and --persist-to CLI args.


thomasgauvin:
Thanks for the example @kodster28, I will continue this PR with this approach

kodster28:
> Thanks for the example @kodster28, I will continue this PR with this approach

Nice!! Let me also add applying this to other doc areas as a potential onboarding task for some of my new folks.

KimJ15:

PR #16002, opened 2024-08-06
[Docs] Fix API links

### LINKED ISSUE #15967: Broken API docs links
The following API doc links are broken: <ul><li>https://developers.cloudflare.com/api/operations/vectorize-update-vectorize-index</li><li>https://developers.cloudflare.com/api/operations/radar-get-connection-tampering-summary</li><li>https://developers.cloudflare.com/api/operations/radar-get-connection-tampering-timeseries-group</li></ul>

pedrosousa:
### Summary

Fixes API links in Dev Docs.
Also adds Radar changelog about endpoint changes.
Closes #15967.

### Notes

API reference links don't need to include the hostname (in fact, you'll get a Vale warning if you do), **unless** the YAML file is being directly used elsewhere outside Dev Docs.
Links in changelog RSS feeds (namely in [Radar's feed](https://developers.cloudflare.com/radar/changelog/index.xml)) will always include the hostname, so that they don't get broken outside the Dev Docs context.

Changelog preview:
https://pedro-issue15967-docs-fix-ap.cloudflare-docs-7ou.pages.dev/radar/changelog/

### DIFF
--- a/data/changelogs/radar.yaml
+++ b/data/changelogs/radar.yaml
@@ -5,6 +5,13 @@ productLink: "/radar/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+- publish_date: '2024-06-27'
+  title: Change TCP connection tampering API endpoints to TCP Resets Timeouts
+  description: |-
+    * Changed the connection tampering summary and timeseries API endpoints to
+    TCP resets timeouts [summary](/api/operations/radar-get-tcp-resets-timeouts-summary)
+    and [timeseries](/api/operations/radar-get-tcp-resets-timeouts-timeseries-group),
+    respectively.
 - publish_date: '2023-11-27'
   title: Add more meta information's
   description: |-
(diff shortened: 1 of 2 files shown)


meddulla (approved):

data/changelogs/radar.yaml
```diff
@@ -13,84 +20,83 @@ entries:
...
 - publish_date: '2023-10-31'
   title: Add new Layer 3 direction parameter
   description: |-
     * Added a `direction` parameter to all Layer 3 endpoints. Use together with `location` parameter to filter by origin or
-      target location [timeseries groups](https://developers.cloudflare.com/api/operations/radar-get-attacks-layer3-timeseries-group-by-vector).
+      target location [timeseries groups](/api/operations/radar-get-attacks-layer3-timeseries-group-by-vector).
 - publish_date: '2023-09-08'
   title: Add Connection Tampering endpoints
   description: |-
-    * Added Connection Tampering [summary](https://developers.cloudflare.com/api/operations/radar-get-connection-tampering-summary)
```
lukevalenta, line 38: As this is the change log, perhaps we should keep the old `connection-tampering` endpoint? Or is this change needed for the link checker?

lukevalenta (approved):
One non-blocking comment but looks good otherwise. Thanks!

data/changelogs/radar.yaml
pedrosousa: We'll have to either update the URLs so the link checker doesn't complain, or use something like monospaced text instead of real links if we want to show the URLs of the old endpoints. My suggestion is that we update the link here, since the previous URLs actually contained `connection-tampering` and should be easy to identify.

kodster28 (approved):

PR #16026, opened 2024-08-07
[ZT] WARP requires PowerShell

ranbel:
PCX-10928


dh-cf (commented):
Ranbel - what/who is the origin of this request to change? We shouldn't be adding Powershell as a requirement in the dev docs.

Never mind - I just saw the PCX ticket number, I'll check there.

PR #16027, opened 2024-08-07
[SSL/TLS] Remove resolved caveat in total-tls/enable.md

mitch292:
### Summary

Total TLS now allows users to change the CA without first disabling the feature.


mitch292:
If we could wait to merge this until next week that'd be great.

kodster28 (approved):

kodster28:
Looks like it works... but still shows an error message?

image

mitch292:
@kodster28 - Sorry about that we had rolled back the change temporarily, but it's back out now.

PR #16083, opened 2024-08-12
Use early return, set satisfies for exported handler

danawoodman:
### Summary

A few "cleanup" items for the typescript example here that (I believe) make the code a bit simpler and idiomatic, specifically using an "early return" rather than an if/else and separating logical blocks with newlines.

This change also means you don't have to pre-define `img` and give it the type `Buffer` which was wrong (should be `ArrayBuffer | null`)

Formatted with Prettier default formatting (which I believe is the convention)

### Documentation checklist

### DIFF
--- a/content/browser-rendering/get-started/screenshots.md
+++ b/content/browser-rendering/get-started/screenshots.md
@@ -97,7 +97,7 @@ export default {
 
 		return new Response(img, { headers: { "content-type": "image/jpeg" } });
 	},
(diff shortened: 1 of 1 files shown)


danawoodman:
ah crap i changed the wrong snippet, one second 🤦

danawoodman:
Ok all fixeeeeeeeed

KimJ15:

PR #16091, opened 2024-08-13
r2: list() max objects

elithrar:
cc @markdembo 

Clarify how list() works in a Workers binding.


elithrar:
Superseded by #16374

PR #16121, opened 2024-08-14
[Docs Site] Fix wrong CSS selector

jsparkdev:
### Summary

By default, checkboxes are readonly. This causes the `input:read-only` CSS selector to be applied. 

Changing the selector to `input[readonly]` will only apply to inputs with the readonly attribute, not checkboxes and radio buttons.

sc

### DIFF
--- a/src/input.css
+++ b/src/input.css
@@ -5,7 +5,7 @@ select {
 	border-width: 2px;
 }
 
-input:read-only {
+input[readonly] {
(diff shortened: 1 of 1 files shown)


KianNH:
Thanks for this @jsparkdev! We're having an issue with CI from forks at the moment so I'll get this change pushed up in another PR but we appreciate your contribution!

KianNH:
This was merged in and should be live soon - thanks again! ❤️

PR #16152, opened 2024-08-14
Update find-account-and-zone-ids.mdx

jasgud:
### Summary

Added note for users who have no workers created.


kodster28:
Howdy @jasgud, we're working internally to fix this to surface the Account ID on the Quickstart page, so going to close out the contribution for now.

PR #16155, opened 2024-08-14
[Docs Site] Increase heap size for publish job

KianNH:
### Summary

Test if 12 GB provides any meaningful performance improvements over 4 GB


kodster28 (approved):

KianNH:
Didn't make a noticeable difference, 3x'd the heap and it built 7 seconds quicker - will shelf in favour of the patch.

PR #16163, opened 2024-08-14
[Chore] Add redirect for sitemap + customize behavior

kodster28:
### Summary

Starlight makes the -- to me at least -- somewhat odd decision to host our sitemap at `/sitemap-index.xml`, which then has all our entries at `/sitemap-0.xml`.

Adding a redirect + some config logic around adding `lastMod` to entries within the sitemap.

Tried following the customization instructions at https://docs.astro.build/en/guides/integrations-guide/sitemap/#_top.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -10,6 +10,7 @@ import rehypeAutolinkHeadings from 'rehype-autolink-headings';
 import { h } from 'hastscript';
 import { readdir } from "fs/promises";
 import icon from "astro-icon";
+import sitemap from '@astrojs/sitemap';
 
 async function autogenSections() {
 	const sections = (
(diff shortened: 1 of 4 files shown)


astro.config.mjs
```diff
@@ -136,5 +137,13 @@ export default defineConfig({
 		}),
 		liveCode({ layout: "~/components/live-code/Layout.astro" }),
 		icon(),
+    sitemap({
+      serialize(item) {
+        item.changefreq = 'daily';
```
kodster28:
```suggestion
```

astro.config.mjs
```diff
@@ -136,5 +137,13 @@ export default defineConfig({
...
+        item.lastmod = new Date();
+        item.priority = 0.9;
```
kodster28:
```suggestion
```

KianNH (approved):

PR #16168, opened 2024-08-14
Update wrangler D1 docs to be consistent with CLI

simonha9:
### Summary

Updating wrangler d1 docs to be consistent - the current description of the `--file` flag is misrepresented.  The CLI mentions that this switch is used for files to be **ingested** which may cause confusion as to expected behaviour.

```
npx wrangler@latest d1 execute --help
Need to install the following packages:
wrangler@3.71.0
Ok to proceed? (y) y
wrangler d1 execute <database>

Execute a command or SQL file

POSITIONALS
  database  The name or binding of the DB  [string] [required]

GLOBAL FLAGS
  -j, --experimental-json-config  Experimental: support wrangler.json  [boolean]
  -c, --config                    Path to .toml configuration file  [string]
  -e, --env                       Environment to use for operations and .env files  [string]
  -h, --help                      Show help  [boolean]
  -v, --version                   Show version number  [boolean]

OPTIONS
  -y, --yes         Answer "yes" to any prompts  [boolean]
      --local       Execute commands/files against a local DB for use with wrangler dev  [boolean]
      --remote      Execute commands/files against a remote DB for use with wrangler dev  [boolean]
      --file        A .sql file to ingest  [string]
      --command     A single SQL statement to execute  [string]
      --persist-to  Specify directory to use for local persistence (for --local)  [string]
      --json        Return output as clean JSON  [boolean] [default: false]
      --preview     Execute commands/files against a preview D1 DB  [boolean] [default: false]
```

### Documentation checklist

- [x ] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.

### DIFF
--- a/src/content/docs/workers/wrangler/commands.mdx
+++ b/src/content/docs/workers/wrangler/commands.mdx
@@ -293,7 +293,7 @@ You must provide either `--command` or `--file` for this command to run successf
 * `--command` string optional
   * The SQL query you wish to execute.
 * `--file` string optional
-  * Path to the SQL file you wish to execute.
+  * A .sql file to ingest.
 * `-y, --yes` boolean optional
   * Answer `yes` to any prompts.
 * `--local` boolean(default: true) optional


dario-piotrowicz (commented):
Thanks @simonha9 😃 

Looks good to me, but I would suggest, if you're making the docs consistent with the CL could you check and update the other flags too?

It seems like there are a few still that are inconsistent, such as `--command`, `--json` and `preview` 🙂 

(we should really look into automating this somehow! 🤔)

dario-piotrowicz:
I also think that `--batch-size` should be removed as it's deprecated and hidden in the CLI
([a while back](https://github.com/cloudflare/workers-sdk/pull/5696/files#diff-000731c5322f551427f9c0bcaa8a22fc21229d916a6727318fd11c4f357ad09eR36))

cc. @geelen

simonha9:
Hey! great idea I will comb through these - yeah totally agree, would rather have the customer-facing docs and CLI both generated from the same place Ideally but since they're different repos this is kind of a pain.

KimJ15 (approved):
LGTM

KimJ15:
@dario-piotrowicz I think we merge this PR today with future efforts to improve CLI in a future PR. Is that accurate?

simonha9:
Actually there's some internal work on this, I'm not sure if this is still relevant so I'm happy to close for now as it would be better maintained by more active docs-updaters.

cc @Oxyjun

Oxyjun:
Thanks @simonha9, let's revisit when it's more appropriate.

PR #16172, opened 2024-08-15
Use a real dark for the dark accent

KTibow:
### Summary

Right now, the sidebar has really low contrast compared to either the reference implementation or the old sidebar. This is because the "dark" color is actually white. This sets the dark color to a real dark color. (It was derived from Material algorithms at high contrast levels.)

### Screenshots (optional)

Before:
image
After:
image
(That was a concept image and wasn't actually tested, I'll update this if it gets tested)

### Documentation checklist

### DIFF
--- a/tailwind.config.mjs
+++ b/tailwind.config.mjs
@@ -1,6 +1,6 @@
 import starlightPlugin from "@astrojs/starlight-tailwind";
 
-const accent = { 200: "#F6821F", 600: "#F6821F", 900: "#F0F0F0" };
+const accent = { 200: "#230C00", 600: "#F6821F", 900: "#F0F0F0" };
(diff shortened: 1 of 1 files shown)


kodster28:
Thanks for contributing! Upon inspection, looks like that turns it pretty dark red / orange so it's illegible.

Screenshot 2024-08-15 at 8 05 14 AM


Let me convert this into an issue and then we can look at it from our side.

KTibow:
~~I believe this would be fixed if the icon was set to use the mode-independent `--sl-color-text-accent` (the same color the background is). However this would require further updating the Tailwind config to make it a color.~~ Disregard the previous, I didn't see it affected more than icons.

PR #16175, opened 2024-08-15
Add undocumented CSP header for WAF’s JS challenge

nisbet-hubbard:
### Summary
The added CSP header is required for WAF’s JS challenge to appear in browsers at all, but is heretofore undocumented.

### Documentation checklist


pedrosousa:
Hi @nisbet-hubbard 

I've checked internally and you actually cannot set your own CSP in challenge pages, as we already say in the [Turnstile documentation](https://developers.cloudflare.com/turnstile/reference/content-security-policy/). I'll create a separate PR to also add this information to the Challenges page.

nisbet-hubbard:
> you actually cannot set your own CSP in challenge pages

@pedrosousa, this PR isn't about setting CSP _on_ challenge pages. The point is without these CSPs on the _protected_ page, a visitor wouldn't even see a challenge and is met with a white screen.

pedrosousa:
I believe you are talking about Turnstile widgets, not Cloudflare challenge pages.

The CSP requirements for Turnstile are already documented in the Turnstile docs:
https://developers.cloudflare.com/turnstile/reference/content-security-policy/

nisbet-hubbard:
Well, nope. We don't use any Turnstile at the moment. The issue here concerns pages where a custom firewall rule is set to do 'JS Challenge'. 

But turns out (non-interactive) JS Challenge requires its own different CSP to show up at all. Hence this PR.

pedrosousa:
Thank you for clarifying.

In fact, we do have [a note in the Rules troubleshooting section](https://developers.cloudflare.com/rules/reference/troubleshooting/#interaction-between-cloudflare-challenges-and-rules-features) about possible issues between Rules features and challenges. We now link to that section from the Challenges page.

Make sure you exclude the specific challenge URI paths from any custom CSP you're setting.

nisbet-hubbard:
Whenever we use your WAF's JS Challenge action on any page, we have to add the following CSP on the page's headers or face a white screen of death:
```html
script-src 'self' 'unsafe-eval' 'unsafe-inline' https://challenges.cloudflare.com; frame-src 'self' https://challenges.cloudflare.com; worker-src 'self' blob:
```
I can't find this mentioned anywhere else, which is the reason I opened this PR.

Could you please confirm whether this CSP requirement is actually mentioned somewhere in your docs?

nisbet-hubbard:
Fixed by #17786.

PR #16201, opened 2024-08-15
[Docs Site] Use sidebar.label if present for group labels

KianNH:
### Summary

Use sidebar.label if present for group labels

### DIFF
--- a/src/components/overrides/Sidebar.astro
+++ b/src/components/overrides/Sidebar.astro
@@ -74,7 +74,7 @@ async function handleGroup(group: Group): Promise<SidebarEntry> {
 
 	const frontmatter = indexPage.data;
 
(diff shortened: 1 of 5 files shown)


kodster28:
Hmmm. pulling down locally, I'm not seeing any customization if I set the label under sidebar.

Not sure if I'm just misundersatnding here though

KianNH:
> Hmmm. pulling down locally, I'm not seeing any customization if I set the label under sidebar.
> 
> Not sure if I'm just misundersatnding here though

This was just for group labels, i.e renaming the `Get Started` group in the Workers sidebar, but it's probably a good opportunity to add things like `groupLabel` (and let `label` instead override the hardcoded `Overview` as well as add something like `hideIndex`).

kodster28 (approved):

PR #16217, opened 2024-08-16
[Pages] Nuxt related resources

kodster28:
### Summary

Based on recent contributions from @atinux, adding a section pulling in the demos + a tutorial at the bottom of the Nuxt page.

We'll have a gallery treatment coming shortly so these are more visible from the top-level of the site, but this improves the immediate experience :).

image


dario-piotrowicz (approved):
Looks great to me! 🚀

crwaters16 (approved):

PR #16251, opened 2024-08-17
Fix: scroll generation issue when opening littlefoot

### LINKED ISSUE #16250: Problem generating scroll when opening littlefoot
### Example URL(s)

https://developers.cloudflare.com/pages/platform/limits/

### Actual Behavior

When opening littlefoot, a scroll is created, so we can't immediately check the content of littlefoot
dd

### Expected Behavior

When opening littlefoot, scroll is not generated and the content of littlefoot should be visible.

### Additional information

I checked the css style and it seems that `overflow: auto` of `.sl-markdown-content table:not(:where(.not-content *))` is the problem

b0unt9:
Close #16250

### DIFF
--- a/src/components/overrides/MarkdownContent.astro
+++ b/src/components/overrides/MarkdownContent.astro
@@ -91,6 +91,12 @@ import '@astrojs/starlight/style/markdown.css';
 		opacity: 1;
 	}
 
+	/* Fixes the problem of scrollbar appearing when littlefoot is opened by removing overflow property */
+	.sl-markdown-content :global(table:not(:where(.not-content *))) {
+		display: block;
+		border-spacing: 0;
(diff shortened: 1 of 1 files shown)


KimJ15:
@KianNH

KianNH:
Hey there! This doesn't seem to fix the issue of littlefoot extending the size of the table & requiring scrolling, but rather just stops the text within the tooltip from overflowing:

image

The issue is primarily due to the use of littlefoot vs. something like Tippy or Floating UI which are 'smarter' about these (i.e https://developers.cloudflare.com/rules/snippets/#availability).

I'll keep the issue open so we can work towards replacing littlefoot with one of those libraries, thanks for the contribution!

PR #16266, opened 2024-08-19
[Bots] Fix sequence-rules partial variables

KianNH:
### Summary

Fixes `sequence-rules` partial.

### Screenshots (optional)

Before:
image

After:
image


janrueth (commented):
👍

marciocloudflare (approved):

PR #16269, opened 2024-08-19
chore: fixup a few codeblock langs

Cherry:
### Summary

Fixes up a few minor codeblocks that were causing Astro warnings due to using invalid langs.


kodster28 (approved):
TY!

PR #16274, opened 2024-08-19
[Chore] Re-add metadata to individual pages

kodster28:
### Summary

Re-adds several bits of metadata that were lost during the mgiration


src/components/overrides/Head.astro
```diff
@@ -13,6 +13,7 @@ if (currentSection) {
   const product = await getEntry("products", currentSection)
 
   if (product) {
+    console.log(product)
```
KianNH:
```suggestion
```

KianNH (approved):

PR #16396, opened 2024-08-22
Update sql-statements.mdx

ShaheenFattoe8:
Updated to reflect `ON DELETE CASCADE` behavior when foreign key enforcement is deferred (`PRAGMA defer_foreign_keys = on`).

### DIFF
--- a/src/content/docs/d1/reference/sql-statements.mdx
+++ b/src/content/docs/d1/reference/sql-statements.mdx
@@ -129,7 +129,7 @@ Allows you to defer the enforcement of [foreign key constraints](/d1/build-with-
 
 This does not disable foreign key enforcement outside of the current transaction: if you have not resolved outstanding foreign key violations at the end of your transaction, it will fail with a `FOREIGN KEY constraint failed` error.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/d1/reference/sql-statements.mdx
```diff
@@ -129,6 +129,8 @@ Allows you to defer the enforcement of [foreign key constraints](/d1/build-with-
 
 This does not disable foreign key enforcement outside of the current transaction: if you have not resolved outstanding foreign key violations at the end of your transaction, it will fail with a `FOREIGN KEY constraint failed` error.
 
+Note that setting `PRAGMA defer_foreign_keys = on` does not prevent `ON DELETE CASCADE` actions from being executed. While foreign key enforcement is deferred, `ON DELETE CASCADE` operations will still be active during a transaction.
```
vy-ton: - try to combine sentences since they echo each other
- mention that this behavior is similar to sqlite (worth testing with sqlite to confirm)

vy-ton (approved):

PR #16406, opened 2024-08-22
[Docs Site] Refine baseSchema for improved MDX intellisense.

KianNH:
### Summary

Adds known values and descriptions to a subset of properties.

### DIFF
--- a/src/schemas/base.ts
+++ b/src/schemas/base.ts
@@ -20,6 +20,7 @@ export const baseSchema = z.object({
 			z.literal("concept"),
 			z.literal("reference"),
 			z.literal("reference-architecture"),
+			z.literal("reference-architecture-diagram"),
 			z.literal("tutorial"),
 			z.literal("api"),
 			z.literal("troubleshooting"),
--- a/src/schemas/base.ts
+++ b/src/schemas/base.ts
@@ -6,16 +6,62 @@ const spotlightAuthorDetails = z
 		author_bio_link: z.string().url(),
 		author_bio_source: z.string(),
 	})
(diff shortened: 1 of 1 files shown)


src/schemas/base.ts
```diff
@@ -6,16 +6,62 @@ const spotlightAuthorDetails = z
...
+			z.literal("faq"),
+			z.literal("integration-guide"),
+			z.literal("changelog"),
+			z.literal("configuration"),
+			z.literal("navigation"),
+			z.literal("example"),
+			z.literal("learning-unit"),
+			z.literal("design-guide"),
+		])
+		.catch((ctx) => ctx.input)
```
KianNH, line 36: This pattern allows us to provide a strict set of values to Zod (and therefore the JSON schema), but still fall back to using the original input as the value.

kodster28 (approved):

src/schemas/base.ts
```diff
@@ -6,16 +6,62 @@ const spotlightAuthorDetails = z
...
-	pcx_content_type: z.string().optional(),
+	pcx_content_type: z
+		.union([
+			z.literal("overview"),
+			z.literal("get-started"),
+			z.literal("how-to"),
+			z.literal("concept"),
+			z.literal("reference"),
+			z.literal("reference-architecture"),
+			z.literal("tutorial"),
```
kodster28, line 24:
```suggestion
      z.literal("reference-architecture-diagram"),
			z.literal("tutorial"),
```

kodster28 (approved):

PR #16427, opened 2024-08-23
Start accepted vocab

kodster28:
### Summary

Per Hyperlint style guidelines, add in an acceptable vocab list


.vale.ini
```diff
@@ -2,6 +2,7 @@ StylesPath = .github/styles
 MinAlertLevel = suggestion
 IgnoredScopes = code, tt, img, url, a
 SkippedScopes = script, style, pre, figure, code
+Vocab = cloudflare-docs
```
bllchmbrs: The issue is here, will follow up in a second

kodster28: Nice!!! TY.

Think Kim / Claire will be super jazzed about this hitting.

bllchmbrs: Nevermind, looks like you got it!

bllchmbrs: See here: https://github.com/cloudflare/cloudflare-docs/pull/16429

You can add that test file and use that to build out your vocabulary, test all the exceptions and so on. Nothing you can't get locally, but it makes it easy to see all of the rules.

bllchmbrs:
The reason why the tests aren't failing @ https://github.com/cloudflare/cloudflare-docs/pull/16427/commits/51028bee9ac1d67c8325fde8fec707a26e541bb6

Is that the `.vale.ini` still doesn't look at mdx files.

Edit: you don't need reject, accept is just fine.

bllchmbrs:
@kodster28 , small config issue. The fix + explanation is demo'd in this PR: https://github.com/cloudflare/cloudflare-docs/pull/16429

kodster28:
> @kodster28 , small config issue. The fix + explanation is demo'd in this PR: #16429

Nice, TYSM for your help!

haleycode (approved):

bllchmbrs:
Nice!

On every commit, you get an update. re: **line 7** Feel free to bump up that error message with a bit more context - Hyperlint will automatically rewrite those for folks.

PR #16429, opened 2024-08-23
[do not merge] Style Guide Setup

bllchmbrs:
@kodster28 , you'll see the change necessary here. You have to add "vale" as a style so that it uses the vocabulary. It's a Vale default style to be able to do it.

When using a `style_guide_test.md` file. It will flag all issues **but will not do suggestions**. Those only come up on 'real' PRs. This allows you to iterate on the style guide very easily (and quickly) and see the results / see test cases and share them with others.


kodster28:
TY!

PR #16472, opened 2024-08-27
[SSL] Update change-encryption-mode-api.mdx

ngayerie:
### Summary

Update encryption modes to add the API origin_pull option


RebeccaTamachiro:
Addressed within #16471

PR #16514, opened 2024-08-28
[Gateway] Global indicator feeds

maxvp:
PCX-13266


src/content/docs/security-center/indicator-feeds/get-started.mdx
```diff
@@ -6,10 +6,9 @@ sidebar:
 head:
   - tag: title
     content: Get started
-
 ---
 
-import { Render } from "~/components"
```
mhrdadsaber68: >

mhrdadsaber68: خوب

src/content/docs/security-center/indicator-feeds.mdx
```diff
@@ -0,0 +1,58 @@
...
+   | --------------- | -------- | ------------------- | ------ |
+   | Indicator Feeds | in       | _Threat Intel Feed_ | Block  |
+
+5. Select **Create policy**.
+
+For more information on creating Gateway policies, refer to [DNS policies](/cloudflare-one/policies/gateway/dns-policies/).
+
+### Publicly available feeds
+
+Cloudflare provides some feeds to Gateway users without the need to establish a provider relationship.
```
pedrosousa: Nit: This information could appear earlier in the "Custom Indicator Feeds" page, because users may feel that they will always need to create a custom indicator feed first.
I'm assuming that these would be immediately available (if they require contacting someone at Cloudflare, it probably makes sense as it is).

src/content/docs/security-center/indicator-feeds.mdx
```diff
@@ -0,0 +1,58 @@
+---
+pcx_content_type: concept
+title: Custom Indicator Feeds
+sidebar:
+  order: 8
+head:
+  - tag: title
+    content: Custom Indicator Feeds
```
pedrosousa: I don't think we need to set this specifically, since it's the same as the page title.

pedrosousa (approved):

PR #16547, opened 2024-08-29
Fix Prisma migrate file format

henrytoone:
### Summary

When using PowerShell on Windows, the command in the docs to generate a migration file produces a file in UTF-16 format which wrangler is unable to migrate properly.
Updating the command to use the output flag instead produces a file in UTF-8 which runs fine with the migrate command

Solution found in: https://github.com/prisma/prisma/issues/23702
Prisma release with the fix: https://github.com/prisma/prisma/releases/tag/5.12.1

### Screenshots (optional)

Error produced when file is in UTF-16
Error produced when file is in UTF-16


Oxyjun (approved):

KianNH:
Thanks for catching this!

PR #16630, opened 2024-09-04
do docs update

elithrar:
This PR revamps the "early" Durable Objects experience by:

- [ ] Updating the get started guide to default to RPC
- [ ] The "Accessing Durable Objects" page will lay out RPC vs HTTP why/when with examples
- [ ] Clarify the terms - namespaces, stubs, Durable Object (singular), Durable Objects (plural), instance(s), etc across the docs
- [ ] A new doc on scaling and data modelling using DOs (horizontal scale)
- [ ] Make clear how storage per DO is private within that DO
- [ ] Provide end-to-end examples using the RPC API (today we link out to the RPC API docs in the Workers docs)

This will help better intro DOs to more users, get them using the newer APIs, and set us up for introducing the SQL API as it matures.


elithrar:
Tracking in https://github.com/cloudflare/cloudflare-docs/pull/16903

PR #16671, opened 2024-09-05
[ZT] Overview audit

maxvp:
PCX-13039


src/content/docs/cloudflare-one/connections/connect-devices/agentless/dns/index.mdx
```diff
@@ -3,17 +3,6 @@ pcx_content_type: navigation
...
-
-:::caution
-
-
-Deploying Gateway DNS filtering using static IP addresses may prevent users from connecting to public Wi-Fi networks through captive portals.
-
-
-:::
-
-<DirectoryListing />
```
pedrosousa, line 19: If you're not setting up redirects, I'd keep at least the `<DirectoryListing>` component so that the page is not empty.
The same would apply to other occurrences in this PR.

pedrosousa (approved):

PR #16685, opened 2024-09-06
Added supported providers on costs

daisyfaithauma:
- Removed the word 'traces' on obsevability index file
- Added supported providers on cost page

### DIFF
--- a/src/content/docs/ai-gateway/observability/costs.mdx
+++ b/src/content/docs/ai-gateway/observability/costs.mdx
@@ -1,0 +1,33 @@
+---
+title: Costs
+pcx_content_type: reference
+---
+
+## Supported Providers
+
+AI Gateway currently supports cost metrics from the following providers:
+
+- Anthropic
+- Azure OpenAI
+- Cohere
+- Google AI Studio
+- Groq
+- Mistral
+- OpenAI
+- Perplexity
+- Replicate
+- Workers AI
+
+Cost metrics are only available for endpoints where the models return token data and the model name in their responses.
+
+:::note[Note]
+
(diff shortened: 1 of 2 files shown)


src/content/docs/ai-gateway/observability/costs.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: Costs
+pcx_content_type: reference
+---
+
+## Supported Providers
+
+AI Gateway currently supports cost metrics from the following providers:
+
+- Anthropic
```
kodster28, line 10: Links over to the providers pages?

src/content/docs/ai-gateway/observability/costs.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+:::caution[Caution]
+
+Providers may introduce new models or change their pricing. If you notice outdated cost data or are using a model not yet supported by our cost tracking, please [submit a request](https://forms.gle/cuXu1QnQCrSNkkaS8)
+
+:::
+
+## Custom Costs
+
+AI Gateway allows users to set custom costs when operating under special pricing agreements or negotiated rates. Custom costs can be applied at the request level, and when applied, they will override the default or public model costs.
```
kodster28, line 37: How? Cross-link or instructions?

daisyfaithauma: This will be edited to redirect to the custom costs. Just finishing on its PR

src/content/docs/ai-gateway/observability/costs.mdx
daisyfaithauma: I was nit actually sure of this, that is why I went with just a list. Some providers like OpenAI are partially supported and I think it would be confusing. The provider pages jus shows the parameters and an example

kodster28 (approved):
Thanks for the context!

Issue #14883, opened 2024-06-03
Clarify smart placement and D1 bindings


brettwillis:
### Existing documentation URL(s)

https://developers.cloudflare.com/workers/configuration/smart-placement/#d1

### What changes are you suggesting?

The [docs](https://github.com/cloudflare/cloudflare-docs/blob/e06c9df3b34f835be13d1e7ca20220f78476e38c/content/workers/configuration/smart-placement.md?plain=1#L38) state:

> Workers with a D1 binding will always be placed in a data center near the location of the D1 database they are bound to. Subrequests to other back-end services are ignored by Smart Placement in this case.

It is unclear if "always" in this case means *always when Smart Placement is on* or just plain *always* (with or without smart placement). Can you clarify if the worker is always colocated with the D1 binding regardless if smart placement is configured? Or is it only colocated with the D1 binding when smart placement is on?

Furthermore, does the "one percent (1%) of requests are not routed with Smart Placement" still apply when a D1 binding is forcing the placement?

gshirescf:
cc: @harshil1712

Oxyjun:
The whole chapter is going through edits, this is now being tracked here: https://github.com/cloudflare/cloudflare-docs/pull/16697

Oxyjun:
Relevant PR has been merged, closing this issue.

PR #16697, opened 2024-09-06
[Workers] Update description of how smart placement works

maxwellpeterson:
- We recently overhauled how smart placement works, and relevant docs need to be updated to reflect changes to the system

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/configuration/smart-placement.mdx
+++ b/src/content/docs/workers/configuration/smart-placement.mdx
@@ -11,8 +11,6 @@ sidebar:
 
 By default, [Workers](/workers/) and [Pages Functions](/pages/functions/) are invoked in a data center closest to where the request was received. If you are running back-end logic in a Worker, it may be more performant to run that Worker closer to your back-end infrastructure rather than the end user. Smart Placement automatically places your workloads in an optimal location that minimizes latency and speeds up your applications.
 
-You may benefit from Smart Placement if you are making multiple round trips to a centralized database, API or origin server in a Worker.
-
 ## Background
 
 The following example demonstrates how moving your Worker close to your back-end services could decrease application latency:
@@ -27,28 +25,11 @@ The issue is the time that it takes the Worker to perform multiple round trips t
 
 ## Understand how Smart Placement works
 
-Smart Placement is enabled on a per-Worker basis. Once enabled, fetch requests (also known as subrequests) from your Worker are analyzed regularly. The Smart Placement algorithm determines the optimal placement to minimize the round-trip time (RTT) between the Worker and the back-end service the Worker is communicating with.
+Smart Placement is enabled on a per-Worker basis. Once enabled, Smart Placement analyzes the [request duration](/workers/observability/metrics-and-analytics/#request-duration) of the Worker in different Cloudflare locations around the world on a regular basis. Smart Placement decides where to run the Worker by comparing the estimated request duration in the location closest to where the request was received (the default location where the Worker would run) to a set of candidate locations around the world. For each candidate location, Smart Placement considers the performance of the Worker in that location as well as the network latency added by forwarding the request to that location. If the estimated request duration in the best candidate location is significantly faster than the location where the request was received, the request will be forwarded to that candidate location. Otherwise, the Worker will run in the default location closest to where the request was received.
 
-Smart Placement is only active for Workers that **make more than one roundtrip** to back-end infrastructure. If your Worker does less than one subrequest on average, Smart Placement will run the Worker at the data center closest to the user.
-
-Smart Placement is a best-effort attempt. Smart Placement will not take action unless it is more performant than the default (which is running the Worker at the data center closest to the user).
+Smart Placement only considers candidate locations where the Worker has previously run, since the estimated request duration in each candidate location is based on historical data from the Worker running in that location. This means that Smart Placement cannot run the Worker in a location that it does not normally receive traffic from.
 
 Smart Placement only affects the execution of [fetch event handlers](/workers/runtime-apis/handlers/fetch/). Workers without a fetch event handler will be ignored by Smart Placement. For Workers with both fetch and non-fetch event handlers, Smart Placement will only affect the execution of the fetch event handler.
-
-### D1
-
-Workers with a [D1](/d1/) binding will always be placed in a data center near the location of the D1 database they are bound to. Subrequests to other back-end services are ignored by Smart Placement in this case.
-
-### Unsupported back-end services
-
-There are some back-end services that are not considered by the Smart Placement algorithm:
-
-- **Globally distributed services**: If the services that your Worker communicates with are geo-distributed in many regions (for example, CDNs, distributed databases, distributed APIs), Smart Placement is not a good fit. We automatically rule these out of the Smart Placement optimization.
-
-  - Examples: Google APIs, services using Fastly or Akamai's CDN.
-
-- **Analytics or logging services**: Requests to analytics or logging services should not be in the critical path of your application. [`waitUntil()`](/workers/runtime-apis/context/#waituntil) should be used so that the response back to users is not blocked when instrumenting your code. Since `waitUntil()` does not impact the request duration from a user’s perspective, we automatically rule analytics and logging services out of the Smart Placement optimization.
-  - Examples: New Relic, Datadog, Tinybird, Grafana, Amplitude, Honeycomb.
 
 ## Enable Smart Placement
 
@@ -67,9 +48,9 @@ To enable Smart Placement via Wrangler:
    mode = "smart"
    ```
 
-3. Send some initial traffic (approximately 20-30 requests) to your Worker. It takes a few minutes after you have sent traffic to your Worker for Smart Placement to take effect.
+3. Wait for Smart Placement to analyze your Worker. This process may take up to 15 minutes.
 
-4. View your Worker's [request duration analytics](/workers/observability/metrics-and-analytics/).
+4. View your Worker's [request duration analytics](/workers/observability/metrics-and-analytics/#request-duration).
 
 ### Enable Smart Placement via the dashboard
 
@@ -80,8 +61,8 @@ To enable Smart Placement via the dashboard:
 3. In **Overview**,select your Worker.
 4. Select **Settings** > **General**.
 5. Under **Placement**, choose **Smart**.
-6. Send some initial traffic (approximately 20-30 requests) to your Worker. It takes a few minutes after you have sent traffic to your Worker for Smart Placement to take effect.
-7. View your Worker's [request duration analytics](/workers/observability/metrics-and-analytics/)
+6. Wait for Smart Placement to analyze your Worker. Smart Placement requires consistent traffic to the Worker from multiple locations around the world to make a placement decision. The analysis process may take up to 15 minutes.
+7. View your Worker's [request duration analytics](/workers/observability/metrics-and-analytics/#request-duration)
 
 ## Observability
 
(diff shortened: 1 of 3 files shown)


src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -27,29 +25,12 @@ The issue is the time that it takes the Worker to perform multiple round trips t
...
-
-Smart Placement is only active for Workers that **make more than one roundtrip** to back-end infrastructure. If your Worker does less than one subrequest on average, Smart Placement will run the Worker at the data center closest to the user.
+Smart Placement is enabled on a per-Worker basis. Once enabled, Smart Placement analyzes the [request duration](/workers/observability/metrics-and-analytics/#request-duration) of the Worker in different cities around the world on a regular basis. Smart Placement decides where to run the Worker by comparing the estimated request duration in the city closest to where the request was received (the default location where the Worker would run) to a set of candidate cities around the world. For each candidate city, Smart Placement considers the performance of the Worker in that city as well as the network latency added by forwarding the request to that city. If the estimated request duration in the best candidate city is significantly faster than the city where the request was received, the request will be forwarded to that candidate city.
```
irvinebroque: "Cloudflare location" instead of "candidate city" maybe? Consistency about how we talk about locations on our network?

https://www.cloudflare.com/network/

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -27,29 +25,12 @@ The issue is the time that it takes the Worker to perform multiple round trips t
...
 
-Smart Placement is a best-effort attempt. Smart Placement will not take action unless it is more performant than the default (which is running the Worker at the data center closest to the user).
```
irvinebroque, line 34: This still seems relevant? Like saying that we only move where your Worker runs if it makes things faster?

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -27,29 +25,12 @@ The issue is the time that it takes the Worker to perform multiple round trips t
...
+Smart Placement only considers candidate cities where the Worker has previously run, since the estimated request duration in each candidate city is based on historical data from the Worker running in that city. This means that Smart Placement cannot run the Worker in a city that it does not normally receive traffic from.
```
irvinebroque: What's the case where someone using this has to worry about this, or actually hits it? Maybe something I don't understand.

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -67,9 +48,9 @@ To enable Smart Placement via Wrangler:
    mode = "smart"
    ```
 
-3. Send some initial traffic (approximately 20-30 requests) to your Worker. It takes a few minutes after you have sent traffic to your Worker for Smart Placement to take effect.
+3. Wait for Smart Placement to analyze your Worker. This process may take up to 15 minutes.
```
irvinebroque, line 51: We still need to say _something_ about the need to send traffic to your Worker, in order for it to be placed, no?

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -98,10 +79,9 @@ curl -X GET https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/workers/s
 Possible placement states include:
 
 - _(not present)_: The Worker has not been analyzed for Smart Placement yet.
-- `INSUFFICIENT_INVOCATIONS`: Not enough requests for Smart Placement to make a placement decision.
-- `NO_VALID_HOSTS`: The Worker does not send subrequests to back-end services supported by Smart Placement.
-- `INSUFFICIENT_SUBREQUESTS`: The Worker does not send enough subrequests to valid back-end services.
-- `SUCCESS`: The Worker has been successfully analyzed and will be optimized by Smart Placement.
+- `SUCCESS`: The Worker was successfully analyzed and will be optimized by Smart Placement.
+- `INSUFFICIENT_INVOCATIONS`: The Worker has not received enough requests to make a placement decision.
+- `UNSUPPORTED_APPLICATION`: Smart Placement began optimizing the Worker and measured the results, which showed that Smart Placement made the Worker slower. In response, Smart Placement reverted to default placement and will not analyze the Worker again until it's redeployed. This state is rare and accounts for less that 1% of Workers with Smart Placement enabled.
```
irvinebroque: The name of this state makes it sound like we're saying we don't support your Worker...but in practice there are legit use cases where it's just...better for the Worker to run everywhere? So it's not really that the Worker is unsupported?

irvinebroque (commented):
Nice one, will leave to @vy-ton for ✅ — wonder if the states could get clearer

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: I was thinking about people developing small hobby/POC applications, where the developer is the only one sending traffic to the Worker. In this case, the Worker will always run in the colo closest to the developer, and can't be placed anywhere else (even if they want it to be). Placement hints will solve this problem. For production workloads where there is a consistent variety of traffic sources, this isn't a problem.

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: The Worker needs to receive traffic, but it's unlikely that one developer can generate the required traffic pattern on their own (without something similar to crossbow or other routing tricks). Currently, we require invocation data in four different cities for a Worker to be "placed," which is hard to generate synthetically. If their Worker receives consistent traffic, most people are best off just waiting instead of trying to coerce the placement algorithm. If their Worker doesn't receive consistent traffic, they're likely out of luck until we ship placement hints (which I really want to do).

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: This state covers the case where smart placement _thinks_ that running the Worker remotely is faster, but for some reason it turns out not to be. If smart placement thinks it's faster to run the Worker everywhere then absolutely it will do that. But sometimes expectations don't match reality, and this state covers that mismatch. I went with "unsupported" here since our model isn't sophisticated enough to understand these cases, but could theoretically do so with improvements (support) in the future.

There's an argument that this state is rare enough that it's not worth exposing publicly and risking unnecessary confusion.

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: Updated to use "Cloudflare location" or "location" instead of "city"

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: I folded this part into the first paragraph, but added another sentence to make it extra clear:
> If the estimated request duration in the best candidate location is significantly faster than the location where the request was received, the request will be forwarded to that candidate location. Otherwise, the Worker will run in the default location closest to where the request was received.

src/content/docs/workers/configuration/smart-placement.mdx
irvinebroque: oh nvm i misread your commit 🤦

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -80,8 +61,8 @@ To enable Smart Placement via the dashboard:
 3. In **Overview**,select your Worker.
 4. Select **Settings** > **General**.
```
vy-ton, line 62: Under `Runtime` section

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -98,10 +79,9 @@ curl -X GET https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/workers/s
 Possible placement states include:
 
 - _(not present)_: The Worker has not been analyzed for Smart Placement yet.
-- `INSUFFICIENT_INVOCATIONS`: Not enough requests for Smart Placement to make a placement decision.
-- `NO_VALID_HOSTS`: The Worker does not send subrequests to back-end services supported by Smart Placement.
-- `INSUFFICIENT_SUBREQUESTS`: The Worker does not send enough subrequests to valid back-end services.
-- `SUCCESS`: The Worker has been successfully analyzed and will be optimized by Smart Placement.
+- `SUCCESS`: The Worker was successfully analyzed and will be optimized by Smart Placement.
+- `INSUFFICIENT_INVOCATIONS`: The Worker has not received enough requests to make a placement decision.
```
vy-ton: We should describe our heuristic for "enough request" at a high-level

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -98,10 +79,9 @@ curl -X GET https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/workers/s
 Possible placement states include:
 
 - _(not present)_: The Worker has not been analyzed for Smart Placement yet.
-- `INSUFFICIENT_INVOCATIONS`: Not enough requests for Smart Placement to make a placement decision.
-- `NO_VALID_HOSTS`: The Worker does not send subrequests to back-end services supported by Smart Placement.
-- `INSUFFICIENT_SUBREQUESTS`: The Worker does not send enough subrequests to valid back-end services.
-- `SUCCESS`: The Worker has been successfully analyzed and will be optimized by Smart Placement.
+- `SUCCESS`: The Worker was successfully analyzed and will be optimized by Smart Placement.
```
vy-ton: This indicates that we selected remote execution? Let's be explicit.

src/content/docs/workers/configuration/smart-placement.mdx
vy-ton: I'd be in favor of sharing directly that this was local execution?

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -98,10 +79,9 @@ curl -X GET https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/workers/s
 Possible placement states include:
```
vy-ton, line 79: Can users observe when the last placement analysis was ran?

maxwellpeterson: No, not currently. It wouldn't be too hard to expose this in the dashboard but would require some simple design work.

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: We moved it under "Runtime" in the new Workers UI, which isn't public yet. For now this is still accurate.

src/content/docs/workers/configuration/smart-placement.mdx
maxwellpeterson: In this state, we will use remote execution if it's faster than local execution. If local execution is indeed optimal, then we'll stick to using that. I added a sentence here to clarify.

src/content/docs/workers/configuration/smart-placement.mdx
```diff
@@ -80,8 +61,8 @@ To enable Smart Placement via the dashboard:
 3. In **Overview**,select your Worker.
 4. Select **Settings** > **General**.
 5. Under **Placement**, choose **Smart**.
-6. Send some initial traffic (approximately 20-30 requests) to your Worker. It takes a few minutes after you have sent traffic to your Worker for Smart Placement to take effect.
-7. View your Worker's [request duration analytics](/workers/observability/metrics-and-analytics/)
+6. Wait for Smart Placement to analyze your Worker. This process may take up to 15 minutes.
```
vy-ton: @maxwellpeterson Think we need a note callout that explains the observed traffic needed to trigger remote execution.

Otherwise, might get feedback that "Smart Placement isn't working". This was definitely easier to observe for subrequest optimization

maxwellpeterson: Added a sentence about the traffic requirements to this step

Oxyjun (approved):

Oxyjun:
Closes https://github.com/cloudflare/cloudflare-docs/issues/14883

PR #16752, opened 2024-09-10
Anni/pages build image update

aninibread:
### Summary

Updated Pages 'Supported Languages and Tools' page:
- Changed header name to 'Build System' to better reflect the contents of the page, and match with the setting name on the dashboard
- Surfaced supported languages and tools tables higher
- Added instructions for how to override default versions

### Documentation checklist


src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
+---
+pcx_content_type: concept
+title: Build system
```
jahands: Where did we end up on the convo of calling it **Build system** vs **Build image**?

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+rss: https://github.com/cloudflare/cloudflare-docs/commits/production/content/pages/_partials/_platform-language-support-and-tools.atom
+---
+
+import {
+	PagesBuildEnvironment,
+	PagesBuildEnvironmentLanguages,
+	PagesBuildEnvironmentTools,
+} from "~/components";
+
+Cloudflare Pages' build environment has broad support for a variety of languages, such as Ruby, Node.js, Python, PHP, and Go.
```
kodster28: The content is generally fine... but I'd prefer some of this separated out into different pages. Happy to hop in and help if this makes sense.

Where I'd probably want this is something like:

```
Build system
|
-- Overview (Everything top of page thru Frameworks)
-- V2 Build system
-- Overriding default versions (maybe put into overview)
-- Build environment
```

src/content/docs/pages/configuration/build-system.mdx
irvinebroque: Think this page specifically describes the Build image and that seems clearest? (separately need to figure out naming of Workers CI/CD / Builds etc.)

ex github documents this way:

https://github.com/actions/runner-images

src/content/docs/pages/configuration/build-system.mdx
irvinebroque: ^ agree

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+If you need to use a [specific version](/pages/configuration/build-system/#overriding-default-versions) of a language, (for example, Node.js or Ruby) you can specify it by providing an associated environment variable in your build configuration, or setting the relevant file in your source code.
```
irvinebroque: This sentence/paragraph is odd because all it tells you is to link to content on the same page?

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+## Supported languages and tools
+
+Cloudflare Pages offers a [v1 and v2 build system](/pages/configuration/build-system/#v2-build-system) with different language and tooling versions supported. You can [override the default versions](/pages/configuration/build-system/#overriding-default-versions) of languages and tools in both build systems.
```
irvinebroque: Why is it important to lead with the v1 vs v2 piece? if I am reading these docs coming to pages for the first time, I just use v2 and don't need to think about it?

Would zoom out and start with the basics (which you kind of have below)

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+### Languages and runtime
```
irvinebroque: This section needs an example of _how_ to specify the version of one of these tools

i.e. — it implies that `NODE_VERSION` is the environment variable I would set to `21.7.1` — but helpful to spell it out

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+In the following table, review the preinstalled versions for languages and runtime included in the Cloudflare Pages' build system, and the environment variables and/or files available for overriding the preinstalled version:
```
irvinebroque: How can we get the table itself to be more self explanatory instead of this?

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+<PagesBuildEnvironmentLanguages />
+
+:::note[Any version]
+Under Supported versions, "Any version" refers to support for all versions of the language or tool including versions newer than the Default version.
```
irvinebroque: I don't understand?

Clearer instead to give an example of overriding?

- the default version is x
- you can override it to a newer version of that language runtime by doing <this>

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+:::
+
+### Tools
+
+In the following table, review the preinstalled versions for tools included in the Cloudflare Pages' build system, and the environment variables available for overriding the preinstalled version:
```
irvinebroque: Docs shouldn't tell you to review a table

```suggestion
The following package managers and tools are preinstalled in the Pages build image:
```

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+<PagesBuildEnvironmentTools />
+
+:::note[Any version]
+Under Supported versions, "Any version" refers to support for all versions of the language or tool including versions newer than the Default version.
```
irvinebroque: Same feedback as above, show rather than tell

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+:::
+
+### Frameworks
+
+If you want to set a specific version of a framework your Cloudflare Pages project is using, Pages will respect your package manager of choice during your build process. For example, if you use Gatsby, your `package.json` should indicate a version of the `gatsby` npm package, which will be installed using `npm install` as your project builds on Cloudflare Pages.
```
irvinebroque:
```suggestion
To use a specific version of a framework, specify it in the project's package manager configuration file.

For example, if you use Gatsby, your `package.json` should include the following:

 <show code>
  
When your build starts, if not already cached (link to build caching docs), version <x> of Gatsby will b installed using `npm install`
```

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+## V2 build system
```
irvinebroque: should be just own page

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+### Limitations
+
+Here are some limitations with the v2 build system:
+
+- Specifying Node.js versions as codenames (for example, `hydrogen` or `lts/hydrogen`).
+- Detecting Yarn version from `yarn.lock` file version.
+- Detecting pnpm version detection based `pnpm-lock.yaml` file version.
+- Detecting Node.js and package managers from `package.json` -> `"engines"`.
+- `pipenv` and `Pipfile` support.
```
irvinebroque: These aren't specific enough, ex, what does:

> Detecting Yarn version from `yarn.lock` file version.

Mean? That this won't work?

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+## Overriding Default Versions
+
+To override default versions of languages and tools in the build system, you can either set the desired version through environment variables or by adding files to your project.
+
+To set the version using **environment variables**, you can:
+
+1. Find the environment variable name for the language or tool on [this page](/pages/configuration/build-system/#supported-languages-and-tools).
```
irvinebroque: Isn't this content right above? Why is this separate from that?

src/content/docs/pages/configuration/build-system.mdx
```diff
@@ -0,0 +1,89 @@
...
+2. Add the environment variable on the dashboard by going to **Settings** > **Environmnet variables** in your Pages project, or [add the environment variable via Wrangler](/workers/configuration/environment-variables/#add-environment-variables-via-wrangler).
+
+Or, to set the version by **adding a file** to your project, you can:
+
+1. Find the environment variable name for the language or tool on [this page](/pages/configuration/build-system/#supported-languages-and-tools).
+2. Add the specified file name to the root directory of your project, and add the desired version number as the contents of the file.
+
+For example, if you were previously relying on the default version of Node.js in the v1 build system, to migrate to v2, you must specify that you need Node.js `12.18.0` by setting a `NODE_VERSION = 12.18.0` environment variable or by adding a `.node-version` or `.nvmrc` file to your project with `12.18.0` added as the contents to the file.
```
irvinebroque: I think you have example, but in not quite right place, see other comment above.

irvinebroque (commented):
Similar to Workers seems like this is deserving of whole section, but progress over perfection. See comments.

src/content/docs/pages/configuration/build-system.mdx
aninibread: I like a better organized build system section in Pages too, but will probably break that change out as a separate PR. In that change I can also loop in other pages that are build system related like "git integration", "build caching"... etc.

aninibread: Happy to pull this into a separate PR that considers that the bigger picture of what the Builds section in Pages should look like. I think in this section it could include other pages too like "build configurations", "build caching" etc. I think that may be too big of a change to fit all into this one PR.

src/content/docs/pages/configuration/build-system.mdx
aninibread: this is more detailed instructions on how it could be done

src/content/docs/pages/configuration/build-system.mdx
aninibread: this is what we had before, happy to revisit this section in a separate PR.

src/content/docs/pages/configuration/build-system.mdx
irvinebroque: `Build Image` Seems most specific:

Screenshot 2024-09-16 at 11 11 56 AM

```suggestion
title: Build Image
```

irvinebroque (approved):

GregBrimble (approved):
technical (not content) approval

PR #16789, opened 2024-09-12
[WAF, other] Remove line breaks after Render

pedrosousa:
### Summary

We no longer need to explicitly add line breaks (`<br/>`) after the `Render` component.
In docs-in-dash and especially when above the fold, removing these breaks helps save some vertical space.

Updating a few tiles in this PR.

### DIFF
--- a/src/content/docs/rules/index.mdx
+++ b/src/content/docs/rules/index.mdx
@@ -6,85 +6,125 @@ sidebar:
 head:
   - tag: title
     content: Cloudflare Rules
-
 ---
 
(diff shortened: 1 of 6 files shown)


src/content/docs/rules/index.mdx
```diff
@@ -6,85 +6,125 @@ sidebar:
...
+	LinkTitleCard,
+	Plan,
+	RelatedProduct,
+	Render,
+} from "~/components";
 
 <Plan type="all" />
 
-<Render file="rules-definition" /> <br/>
+<Render file="rules-definition" />
```
pedrosousa, line 22: Only real change: Removed the line break.

marciocloudflare (approved):

PR #16794, opened 2024-09-12
[DNS] Remove outdated domains list screenshot

RebeccaTamachiro:
### Summary

Zone lists in the Dash home have changed, so this PR removes an instance where we were showing an outdated screenshot.
Ended up also fixing some notes that ended up like `:::undefined` and were not rendering.

PCX-12903


daisyfaithauma (approved):
LGTM

PR #16834, opened 2024-09-14
[Docs Site] remove unnecessary DOMContentLoaded listeners

Cherry:
### Summary

Just some very minor optimisations by not using `DOMContentLoaded` events. Since all these scripts are processed by Astro, they get `defer`-ed as `type=module`, so they're guaranteed to only run when all DOM nodes are accessible.


KimJ15:
@KianNH

KianNH (approved):

PR #16839, opened 2024-09-14
[Docs Site] format astro files

Cherry:
### Summary

Formats .astro files with repo rules, to slowly get this all consistent.


src/pages/plans.astro
```diff
@@ -96,12 +107,32 @@ import { ProductFeatures } from "~/components";
...
-	<p>This page is meant to be a companion to our <a href='https://www.cloudflare.com/plans/'>Plans and Pricing page</a>, but is not meant to be comprehensive. In the event of any inconsistency between this page and the Plans and Pricing page, the Plans and Pricing page controls.</p>
-    <p>Refer to the following locations for more information:</p>
-    <ul>
-        <li><strong>Developer products</strong>: Refer to the <a href='https://www.cloudflare.com/plans/developer-platform/'>Developer Platform pricing page</a> or individual <a href='/products/'>product docs</a> to learn more about plans and pricing.</li>
-        <li><strong>Zero Trust products</strong>: Refer to the <a href='https://www.cloudflare.com/plans/zero-trust-services/'>Zero Trust pricing</a> page.</li>
-        <li><strong>Other products</strong>: Refer to the individual <a href='/products/'>product docs</a> to learn more about plans and pricing.</li>
-    </ul>
-	
+	<p>
+		This page is meant to be a companion to our <a
```
GregBrimble, line 111: I don't love the split tags personally, but we can change that in a follow-up PR once we get everything standardized in the first place.

GregBrimble (approved):

src/pages/plans.astro
Cherry: fwiw I agree. Splitting them makes sense once you're passing 3+ attributes, but just for content feels arbitrary. Can all be tweaked in future as you said - this is mostly just default prettier rules + tabs: https://github.com/cloudflare/cloudflare-docs/blob/production/.prettierrc.mjs

KimJ15:
@KianNH

PR #16860, opened 2024-09-16
[WAF] Clarify interaction between challenges and Rules features

pedrosousa:
### Summary

Related to #16175.
We already had some content about the interaction between Challenges and Rules features in the Rules documentation, but we were missing a reference to this in the Challenges page.


daisyfaithauma (approved):
lgtm

PR #16865, opened 2024-09-16
Cancel in-progress publish jobs when a new one comes in

penalosa:


KianNH (approved):

pedrosousa:
These changes are now killing publish jobs in `production` if a new PR gets merged. If we have sufficient merges in a row, we'll be actually delaying updates in the live docs site. 

Should we exclude `production` from these concurrency checks?

PR #16889, opened 2024-09-17
[Ref Arch] Move image captions below images

kodster28:
### Summary

Removes unused files from previous CI + find/replace captions for Ref Arch diagrams


kodster28:
Should non-prettier it.

PR #16890, opened 2024-09-17
[Ref Arch] Update caption formatting (no prettier)

kodster28:
### Summary

We no longer support captions as part of image pre-processing using Hugo markdown hooks. Changing the formatting to account for this.


kodster28:
@KianNH just keeps preventing me from inflating my commit count :(

PR #16891, opened 2024-09-17
[Docs Site] Add figcaption from img title attribute

KianNH:
### Summary

Adds `figcaption` elements below an `img` using the `title` attribute which is the string after the path: `foo`

Before:
image

After:
image


kodster28 (approved):
LGTM

PR #16898, opened 2024-09-17
[SSL] Update bundling-methodologies.mdx

ngayerie:
### Summary

Clarify which value for the bundle_method parameter needs to be used in the API


chris-martinelli (commented):
removed whitespace, looks good otherwise!

RebeccaTamachiro (approved):

PR #16903, opened 2024-09-18
Docs feature branch: SQLite in DO

vy-ton:
TODO: 

- [x] Overview
- [x] @vy-ton Pricing, Limits
- [x] Get started walkthrough
- [x] @vy-ton (New) Best Practices > Query embedded SQL database: code examples for different SQL tasks
- [x] @vy-ton API > Transactional Storage: add SQL API
- [x] PITR
- [x] Reference > Durable Objects migrations: add `new_sqlite_classes`
- [x] @vy-ton (New) Platform: D1 vs DO comparison
- [x] @vy-ton Changelog

TODO unrelated to SQLite-in-DO:

- [ ] Refactor "Best Practices" to "Build" and combine both https://developers.cloudflare.com/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/ and https://developers.cloudflare.com/durable-objects/best-practices/access-durable-objects-from-a-worker/ into a clearer doc that talks about RPC vs. fetch and how to communicate with a DO. @elithrar 
- [ ] Example of communicating with a DO in a different script (using `script_name` in the binding)
- [ ] Add end-to-end examples of using the RPC API @elithrar 

Later PR: 
- [ ] (New) Examples > Use the SQL API 
- [ ] (New) Examples > SQL schema migrations
- [ ] - [ ] Add examples of RPC (default; recommended) vs fetch (WebSockets) and clarify differences between each approach (and that both can be used) @elithrar 
- [ ] New `/durable-objects/build/scaling-durable-objects/` doc that talks about how to scale horizontally, data model, and manage DOs. @elithrar


vy-ton:
We're going to use `sqlite-in-do` as the docs feature branch. Please open any docs PR against this branch so we can merge it pre-launch.

> (New) Examples > Use the SQL API

@harshil1712 Could you work on adding an example or even multiple? Kenton's example in the public blog should be available in docs as well.

src/content/docs/durable-objects/get-started/walkthrough.mdx
```diff
@@ -218,6 +218,24 @@ tag = "v1" # Should be unique for each entry
 new_classes = ["MyDurableObject"] # Array of new classes
 ```
 
+### 6.a Optional: Configure new Durable Object class for SQL storage
```
elithrar: IMO we should break this down into two tutorials:

- The existing walkthrough
- A copy of it that uses the SQL API (only) with a big "beta" note at the top

That way we don't mix the two worlds + users understand the differences (beta, limits, etc)

src/content/docs/durable-objects/get-started/walkthrough.mdx
elithrar, line 1: @vy-ton I can rewrite this if you want - I had started this in https://github.com/cloudflare/cloudflare-docs/pull/16630/files#diff-4944f04235c81c2c87330186cbd12b899f755ae573dd5f0edfceb377b75c315e

- The existing guide is rewritten to use RPC throughout
- We copy that over and replace the storage part with SQL parts and a beta notice (rest stays the same!) - can use partials/a template to make it really easy.

vy-ton: @elithrar SGTM. Current tutorial doesn't introduce DO storage at all, so that needs to be added to both KV and SQL versions.

src/content/partials/durable-objects/durable-objects-sql.mdx
```diff
@@ -0,0 +1,88 @@
...
+  this.sql.exec("SELECT * FROM artist ORDER BY artistname ASC;").one();
+
+  // returns { artistid: 123, artistname: 'Alice' }
+  let oneRow = this.sql.exec("SELECT * FROM artist WHERE artistname = ?;", "Alice").one()
+```
+
+Cursor behavior:
+```ts
+  let cursor = this.sql.exec("SELECT * FROM artist ORDER BY artistname ASC;");
+  let { value: firstRow } = cursor[Symbol.iterator]().next()
```
vy-ton: @vy-ton to update code examples once workerd is released with https://github.com/cloudflare/workerd/pull/2753

vy-ton:
@joshthoward @kentonv @justin-mp SQLite in DO docs ready for initial pass. Included updated DO glossary from Matt.

I still have: 
- [TODO] Platform: D1 vs DO comparison

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+---
+
+import { Render } from "~/components";
+
+Durable Objects are a powerful compute API that provides a compute with storage building block. Each Durable Object has private, transactional and strongly consistent storage attached. By default, key-value storage is available. SQL storage(/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) is available on a new beta version of Durable Objects.
+
+You should use Durable Objects storage to persist state durably on-disk that needs to survive beyond the lifetime of a Durable Object, that is, eviction of a Durable Object from memory. A Durable Object is alive based on the request traffic incoming to the instance. There are also normal operations like [code deployments](/workers/configuration/versions-and-deployments/) that trigger a Durable Object to reset. For these reasons, Durable Objects storage is recommended for persisting [in-memory state](/durable-objects/reference/in-memory-state/).
+
+## Access storage
```
vy-ton, line 15: @vy-ton TODO: fold in blog content on fast synchronous writes and N+1 selects here. Link blog as well

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -6,9 +6,11 @@ sidebar:
 
 ---
 
-The Transactional Storage API allows you to achieve consistent key-value storage.
+import { Render } from "~/components";
 
-Durable Objects gain access to a persistent Transactional Storage API via `state`, the first parameter passed to the Durable Object constructor.
+The Transactional Storage API allows Durable Objects to access consistent key-value storage or (beta) [SQL storage](/durable-objects/api/transactional-storage-api/#sqlexec). A Durable Object's attached storage is private to its unique instance and cannot be accessed by other objects.
```
joshthoward: Maybe it's worth renaming "Transactional Storage API" to "Storage API" and then referring to SQLite as a capability of the Storage API in addition to KV.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+<code>ctx.storage.sql.databaseSize()</code> : number
+
+#### Returns
+The current SQLite database size in bytes.
+
+```ts
+  let size = ctx.storage.sql.databaseSize();
+```
+
+### Point in time recovery
```
joshthoward, line 298: Should note that these methods are not available in local development mode because we don't keep a persistent log of change locally.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -39,14 +41,20 @@ export class Counter {
 
 ## Methods
 
+:::note[SQLite in Durable Objects Beta]
+
+New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. SQL storage is [opt-in during beta](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-create-durable-object-class-migration) when creating a Durable Object class; otherwise, a Durable Object class has the standard, private key-value storage. Objects with SQL storage can use the [SQL API](/durable-objects/api/transactional-storage-api/#sqlexec).
```
joshthoward: We should probably mention that key-value storage stores different data in a _cf_KV table which is queryable via SQL. Not that this would be preferable, but I could imagine customers expecting that somehow SQL and KV have the same underlying data accessed by different APIs which is incorrect. CosmosDB is an existing product which gives you different APIs for the same data.

src/content/docs/durable-objects/get-started/walkthrough.mdx
joshthoward: +1

src/content/docs/durable-objects/platform/limits.mdx
```diff
@@ -10,24 +10,38 @@ import { Render } from "~/components"
...
+| Key size                                 | 2 KiB (2048 bytes)                                               | Key and value combined cannot exceed 2 MB                        |
+| Value size                               | 128 KiB (131072 bytes)                                           | Key and value combined cannot exceed 2 MB                        |
+| WebSocket message size                   | 1 MiB (only for received messages)                               | 1 MiB (only for received messages)                               |
+| CPU per request                          | 30s (including WebSocket messages) <sup>3</sup>                  | 30s (including WebSocket messages) <sup>3</sup>                  |
+
+<sup>1</sup> New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. When creating a Durabe Object class, users can [opt-in to using SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-create-durable-object-class-migration).
+
+<sup>2</sup> Durable Objects both bills and measures storage based on a gigabyte <br/> (1 GB = 1,000,000,000 bytes) and not a gibibyte (GiB). <br/>
```
joshthoward, line 27: We actually use gibibytes

src/content/docs/durable-objects/platform/limits.mdx
```diff
@@ -10,24 +10,38 @@ import { Render } from "~/components"
...
+
+<sup>3</sup> Each incoming HTTP request or WebSocket *message* resets the remaining available CPU time to 30 seconds. This allows the Durable Object to consume up to 30 seconds of compute after each incoming network request, with each new network request resetting the timer. If you consume more than 30 seconds of compute between incoming network requests, there is a heightened chance that the individual Durable Object is evicted and reset.
+
+<sup>4</sup> Will be raised to 10GB for general availability.
```
joshthoward, line 31: 10 GiB

src/content/docs/durable-objects/reference/durable-objects-migrations.mdx
```diff
@@ -94,3 +94,21 @@ Note that `.toml` files do not allow line breaks in inline tables (the `{key = "
 
 
 :::
+
+### Enable SQL storage on create Durable Object class migration
+
+:::note[SQLite in Durable Objects Beta]
+
+New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. SQL storage is opt-in during beta; otherwise, a Durable Object class has the standard, private key-value storage. Objects with SQL storage can use the [SQL API](/durable-objects/api/transactional-storage-api/#sqlexec).
```
joshthoward: We should make it clear that SQL storage is opt-in only for new classes at first deploy. I would do this here (rather than just below) and in the change log.

src/content/partials/workers/transactional_storage_api_pricing.mdx
```diff
@@ -18,10 +18,36 @@ Durable Objects do not have to use the Transactional Storage API, but if your co
 
 <sup>2</sup>  List operations are billed by read request units, based on the amount of data examined. For example, a list request that returns a combined 80 KB of keys and values will be billed 20 read request units. A list request that does not return anything is billed for 1 read request unit.
 
-<sup>3</sup>  Each alarm write is billed as a single write request unit.
+<sup>3</sup>  Each `setAlarm` is billed as a single write request unit.
```
joshthoward: `getAlarm` will also be a read.

src/content/partials/workers/transactional_storage_api_pricing.mdx
```diff
@@ -18,10 +18,36 @@ Durable Objects do not have to use the Transactional Storage API, but if your co
...
+
+For [Durable Objects classes with SQL access](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-create-durable-object-class-migration) via `ctx.storage.sql` the following pricing is used instead:
+
+|                             | Workers Paid plan          |
+| ----------------------------| -------------------------- |
+| Rows reads <sup>1,2</sup>   | First 25 billion / month included + $0.001 / million rows |
+| Rows written <sup>1,2</sup> | First 50 million / month included + $1.00 / million rows |
+| SQL Stored data             | 5 GB-month, + $0.20/ GB-month |
+
+<sup>1</sup> Rows read and rows written included limits and rates match D1, Cloudflare's serverless SQL database, [pricing](/d1/platform/pricing/).
```
joshthoward: Either here or on the D1 vs. DO comparison page, I would explain the rationale that platform limits and costs are not intended to be a decision point between DO and D1.

src/content/docs/durable-objects/platform/limits.mdx
vy-ton: @joshthoward CF billing converts to GB on invoices I've seen. Can we use GB everywhere?

For D1 max database size, we also use 10GB

src/content/docs/durable-objects/api/transactional-storage-api.mdx
kentonv: I agree, it's especially weird to use "Transactional" in the title when we meanwhile tell people they don't need to worry about transactions because it all happens implicitly...

src/content/docs/durable-objects/api/transactional-storage-api.mdx
kentonv: > We should probably mention that key-value storage stores different data in a _cf_KV table which is queryable via SQL.

It is not queryable by the app. We explicitly block access. From the developer's point of view, the DO has two different storage areas: KV and SQL.

I'm not against documenting that KV is actually implemented by a hidden table in SQL under the hood but that's more of a detail.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+* `query`: string 
+  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings.
+* `bindings`: any[] Optional
+  * Optional variable number of arguments that correspond to the `?` placeholders in `query`.
+
+#### Returns 
+A cursor (`SqlStorageCursor`) to iterate over query row results as objects with the following supported methods:
+  
+* `raw()`: Iterator
+  * Returns cursor of query rows as arrays of row values.
```
kentonv: For the moment it returns an iterator, not a full cursor.

```suggestion
  * Returns an iterator over the same query results, but representing each row as an array (with no column names) rather than an object.
```

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+* `columnNames()`: string[]
+  * Returns an array of column name strings. Often used with `raw()` to find index of specific column(s).
+*  `rowsRead()`: number
+  * Returns number of rows read by the executed SQL `query` based on cursor consumption. Metric is the same one used for [SQL billing](TODO link).
+*  `rowsWritten()`: number
+  * Returns number of rows written by the executed SQL `query` based on curosr consumption. Metric is the same one used for [SQL billing](TODO link).
```
kentonv:
```suggestion
*  `rowsRead()`: number
  * Returns the number of rows read so far as part of this SQL query. This may increase as you iterate the cursor. The final value is used for [SQL billing](TODO link).
*  `rowsWritten()`: number
  * Returns the number of rows written so far as part of this SQL query. This may increase as you iterate the cursor. The final value is used for [SQL billing](TODO link).
```

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+A cursor to iterate over query row results as objects with the [same supported methods as `sql.exec`](/durable-objects/api/transactional-storage-api/#returns)
+
+#### Examples
+
+```ts
+  let stmt = this.sql.prepare("SELECT * FROM artist WHERE artistid = ?;");
+  let result = stmt(123);
+  console.log([...result.raw()]); // prints [[123, 'Alice' ]]
+```
+
```
kentonv: We removed `prepare()`.

```suggestion
```

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+#### Parameters
+* `query`: string 
+  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings.
+* `bindings`: any[] Optional
+  * Optional variable number of arguments that correspond to the `?` placeholders in `query`.
+
+#### Returns 
+A cursor (`SqlStorageCursor`) to iterate over query row results as objects with the following supported methods:
+  
+* `raw()`: Iterator
```
kentonv, line 257: Need to add `next()`, `toArray()`, and `one()` here.

src/content/changelogs/durable-objects.yaml
```diff
@@ -5,6 +5,12 @@ productLink: "/durable-objects/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+  - publish_date: "2024-09-26"
+    description: |-
+      New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. When creating a Durabe Object class, users can [opt-in to using SQL storage](durable-objects/best-practices/access-durable-objects-storage/#sql-storage) in order to access Transactional Storage [SQL API](/durable-objects/api/transactional-storage-api/#sqlexec).
```
justin-mp: spelling: Durabe -> Durable

justin-mp: "Transactional Storage SQL API" is a mouthful of a term.  We now have two backends: one that provides only the "Storage API"  and one that provides both "Storage API" and the "SQL API".  It's probably good to explain that.

src/content/changelogs/durable-objects.yaml
```diff
@@ -5,6 +5,12 @@ productLink: "/durable-objects/"
...
+
+      During the initial beta, billing is not enabled for SQL queries (rows read and rows written) and SQL storage. SQLite-backed Durable Objects will incur [charges for requests and duration](/durable-objects/platform/pricing/#billing-metrics). We plan to enable [SQL billing](/durable-objects/platform/pricing/#transactional-storage-api-billing) in the first half of 2025 with advance notice. SQL billing metrics will be added to GraphQL analytics in the future as well.
```
justin-mp: As far as I'm aware, billing is not enabled for queries and storage of any kind for SQLite-backed DOs.  I'd just drop the SQL from all of these sentences.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  let size = ctx.storage.sql.databaseSize();
+```
+
+### Point in time recovery
+
+For [Durable Objects classes with SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-create-durable-object-class-migration), the following Storage API methods are available to restore a Durable Object's embedded SQLite database to any point in time in the past 30 days.
+
+<code>ctx.storage.getCurrentBookmark()</code> : Promise\<string>
+<code>ctx.storage.getBookmarkForTime(timestamp: number | Date)</code>: Promise\<string>
+<code>ctx.storage.onNextSessionRestoreBookmark(bookmark: string)</code>: Promise\<string>
```
kentonv, line 312: These need to be fleshed out I suppose. Here's some text:

> The point-in-time-recovery (PITR) API represents points in times using "bookmarks". A bookmark is a mostly alphanumeric string like `0000007b-0000b26e-00001538-0c3e87bb37b3db5cc52eedb93cd3b96b`. Bookmarks are designed to be lexically comparable: a bookmark representing an earlier point in time compares less than one representing a later point, using regular string comparison.

`getCurrentBookmark()`:

> Returns a bookmark representing the current point in time in the object's history.

`getBookmarkForTime()`:

> Returns a bookmark representing approximately the given point in time, which must be within the last 30 days. If the timestamp is represented as a number, it is converted to a date as if using `new Date(timestamp)`.

`onNextSessionRestoreBookmark()`:

> Instructs that the next time the Durable Object restarts, it should restore its storage to exactly match what it contained at the given bookmark. After calling this, the application should typically invoke `ctx.abort()` to restart the Durable Object, thus completing the point-in-time recovery.
>
> This method returns a special bookmark representing the point in time immediately before the recovery takes place (even though that point in time is still technically in the future). Thus, after the recovery completes, it can be undone by performing a second recovery to this bookmark.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+#### Returns
+The current SQLite database size in bytes.
+
+```ts
+  let size = ctx.storage.sql.databaseSize();
+```
+
+### Point in time recovery
+
+For [Durable Objects classes with SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-create-durable-object-class-migration), the following Storage API methods are available to restore a Durable Object's embedded SQLite database to any point in time in the past 30 days.
```
kentonv: May want to clarify:

> These methods simultaneously apply to both the object's SQL storage and its traditional key-value storage.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
justin-mp: curosr->cursor

src/content/partials/durable-objects/durable-objects-sql.mdx
```diff
@@ -0,0 +1,88 @@
...
+[SQL API](/durable-objects/api/transactional-storage-api/#sqlexec) examples below use the following SQL schema:
+
+```ts
+export class MyDurableObject extends DurableObject{
+  sql: SqlStorage
+	constructor(ctx: DurableObjectState, env: Env) {
+		super(ctx, env);
+		this.sql = ctx.storage.sql;
+
+    this.sql.exec(`CREATE TABLE IF NOT EXISTS artist(
```
kentonv, line 16: It looks like you're trying to pass two SQL statements as separate parameters to this `sql.exec`, but that's not how it works. You can pass two statements as part of a single string, separated by a `;` inside the string.

src/content/partials/durable-objects/durable-objects-sql.mdx
```diff
@@ -0,0 +1,88 @@
...
+---
+
+import { Details } from "~/components"
+
+[SQL API](/durable-objects/api/transactional-storage-api/#sqlexec) examples below use the following SQL schema:
+
+```ts
+export class MyDurableObject extends DurableObject{
+  sql: SqlStorage
+	constructor(ctx: DurableObjectState, env: Env) {
```
kentonv: Some lines here are indented with tabs, others with spaces.

src/content/partials/durable-objects/durable-objects-sql.mdx
```diff
@@ -0,0 +1,88 @@
...
+  let resultsArray3 = this.sql.exec("SELECT * FROM artist;").toArray();
+```
+
+Convert query results to an array of row values arrays:
+```ts
+  // Returns [[123,"Alice"],[456,"Bob"],[789,"Charlie"]]
+  let resultsArray = [...this.sql.exec("SELECT * FROM artist;").raw()];	
+
+  // Returns ["artistid","artistname"]
+  let columnNameArray = Array.from(this.sql.exec("SELECT * FROM artist;").columnNames);
```
kentonv: `columnNames` is already an array, no need for `Array.from`.

```suggestion
  let columnNameArray = this.sql.exec("SELECT * FROM artist;").columnNames;
```

src/content/partials/durable-objects/durable-objects-sql.mdx
kentonv: If we're going to show an example of `next()` I suppose it had better check the `done` property, since the statement could return zero results.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
justin-mp, line 257: We need to document that the `raw()` iterator and the cursor iterate over the same values.  It's possible to use both at the same time!

src/content/partials/workers/transactional_storage_api_pricing.mdx
kentonv: For old-style DOs I think `getAlarm` is free.

For SQLite-backed DOs I think we have now decided to make it free, since it is read implicitly in a lot of places.

src/content/partials/workers/transactional_storage_api_pricing.mdx
```diff
@@ -18,10 +18,36 @@ Durable Objects do not have to use the Transactional Storage API, but if your co
...
+
+### SQL storage billing 
+
+:::note[SQLite in Durable Objects Beta]
+
+New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. When creating a Durabe Object class, users can [opt-in to using SQL storage](/durable-objects/best-practices/access-durable-objects-storage/#sql-storage).
+
+During the initial beta, billing is not enabled for SQL queries (rows read and rows written) and SQL storage. SQLite-backed Durable Objects will incur [charges for requests and duration](/durable-objects/platform/pricing/#billing-metrics). We plan to enable SQL billing in the first half of 2025 with advance notice.
+
+You can introspect rows read and rows written using `ctx.storage.sql().rowsRead()` and `ctx.storage.sql().rowsWritten()` on the [cursor returned](/durable-objects/api/transactional-storage-api/#returns) by a SQL query. [`ctx.storage.sql.databaseSize()`](/durable-objects/api/transactional-storage-api/#sqldatabaseSize) returns the current SQL database size for a Durable Object. SQL billing metrics will be added to GraphQL analytics in the future.
```
kentonv: `rowsRead`, `rowsWritten`, and `databaseSize` are properties, not methods.

```suggestion
You can introspect rows read and rows written using `cursor.rowsRead` and `cursor.rowsWritten` on the [cursor returned](/durable-objects/api/transactional-storage-api/#returns) by a SQL query. [`ctx.storage.sql.databaseSize`](/durable-objects/api/transactional-storage-api/#sqldatabaseSize) returns the current SQL database size for a Durable Object. SQL billing metrics will be added to GraphQL analytics in the future.
```

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+<code>ctx.storage.sql.exec(query: string, ...bindings: any[])</code> : SqlStorageCursor
+
+#### Parameters
+* `query`: string 
+  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings.
+* `bindings`: any[] Optional
+  * Optional variable number of arguments that correspond to the `?` placeholders in `query`.
+
+#### Returns 
+A cursor (`SqlStorageCursor`) to iterate over query row results as objects with the following supported methods:
```
justin-mp: We need to make it clear that `SqlStorageCursor` is both a JavaScript Iterator and Iterable so that you can iterate on the cursor directly.  Otherwise, folks will think you need to call `exec(...).raw()` to get an iterator.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  
+* `raw()`: Iterator
+  * Returns cursor of query rows as arrays of row values.
+* `columnNames()`: string[]
+  * Returns an array of column name strings. Often used with `raw()` to find index of specific column(s).
+*  `rowsRead()`: number
+  * Returns number of rows read by the executed SQL `query` based on cursor consumption. Metric is the same one used for [SQL billing](TODO link).
```
justin-mp: I think we can make this easier to understand by making it less terse.  How about:

Returns number of rows read by the executed SQL `query`.  The number of rows will increase as you iterate the the results of the cursor. The metric is the same one used for [SQL billing](TODO link).

We can do the same thing for rowsWritten.

src/content/partials/durable-objects/durable-objects-sql.mdx
kentonv: Note: GitHub has hidden two comments immediately before this one, don't forget to expand them:

Screenshot from 2024-09-23 16-20-29

(People often miss it...)

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+*  `rowsWritten()`: number
+  * Returns number of rows written by the executed SQL `query` based on curosr consumption. Metric is the same one used for [SQL billing](TODO link).
+
+#### Examples
+
+<Render file="durable-objects-sql" />
 
+### sql.prepare
```
justin-mp: `sql.prepare` requires the experimental flag and it's considered to be a public API.  We shouldn't document it.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+```ts
+  let stmt = this.sql.prepare("SELECT * FROM artist WHERE artistid = ?;");
+  let result = stmt(123);
+  console.log([...result.raw()]); // prints [[123, 'Alice' ]]
+```
+
+### sql.databaseSize
+
+<code>ctx.storage.sql.databaseSize()</code> : number
```
justin-mp: `sql.databaseSize` is a property, not a function.  Please remove the `()`.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+#### Returns
+The current SQLite database size in bytes.
+
+```ts
+  let size = ctx.storage.sql.databaseSize();
```
justin-mp: Just:
```
let size = ctx.storage.sql.databaseSize;
```

src/content/docs/durable-objects/api/transactional-storage-api.mdx
justin-mp: We're missing `.one()` and `.toArray()`.

When documenting `.toArray()`, we need to explain that if you've already started iterating, `toArray` will give you the remaining portion of the Cursor.

src/content/docs/durable-objects/api/transactional-storage-api.mdx
```diff
@@ -218,7 +226,82 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+For [Durable Objects classes with SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-create-durable-object-class-migration), the following Storage API methods are available to restore a Durable Object's embedded SQLite database to any point in time in the past 30 days.
+
+<code>ctx.storage.getCurrentBookmark()</code> : Promise\<string>
+<code>ctx.storage.getBookmarkForTime(timestamp: number | Date)</code>: Promise\<string>
+<code>ctx.storage.onNextSessionRestoreBookmark(bookmark: string)</code>: Promise\<string>
+
+```ts
+  let now = new Date();
+  // restore to 2 days from current date and time
```
justin-mp: How about just "restore to 2 days ago"?

src/content/docs/durable-objects/api/transactional-storage-api.mdx
vy-ton: `All Storage API methods work with SQL storage. Key-value methods like `get()`, `put()`, `delete()`, or `list()` store data in a hidden SQLite table. ` is mentioned in this same note callout

src/content/changelogs/durable-objects.yaml
vy-ton: I've been framing as top-level `Storage API` which has KV API methods and SQL API methods.  Depending on backend, you have access to certain API methods.

Thoughts?

src/content/docs/durable-objects/api/transactional-storage-api.mdx
vy-ton: renamed to `Storage API`

src/content/changelogs/durable-objects.yaml
justin-mp: A top-level "Storage API" with "Key-Value API" methods and "SQL API" methods would be fine if we're consistent about the distinction.

elithrar: ^ agree

src/content/docs/durable-objects/platform/limits.mdx
vy-ton: discussed with Josh, we will use `GB` in docs even if edgeworker configures `GiB`. Users can get a tiny bit more which isnt an issue

src/content/docs/durable-objects/platform/limits.mdx
vy-ton: see https://github.com/cloudflare/cloudflare-docs/pull/16903#discussion_r1773712888

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
+---
+{}
+---
+
+## SQL in Durable Objects vs D1
```
vy-ton, line 5: cc @joshthoward @justin-mp

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+<code>ctx.storage.sql.exec(query: string, ...bindings: any[])</code> : SqlStorageCursor
+
+#### Parameters
+* `query`: string 
+  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings.
+* `bindings`: any[] Optional
+  * Optional variable number of arguments that correspond to the `?` placeholders in `query`.
+
+#### Returns 
+A cursor (`SqlStorageCursor`) to iterate over query row results as objects. Cursor is a JavaScript [Iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol), which supports iteration using `for (let row of cursor)`. Cursor is also a JavaScript [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol), which supports iteration using `cursor.next()`.
```
vy-ton: cc @justin-mp @kentonv for another pass of `sql.exec` api docs

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+Cloudflare Workers offers a SQLite-backed serverless database product - [D1](/d1/). How should you compare [SQLite in Durable Objects](/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) and D1?
+
+**D1 should be viewed as a managed database product.**
```
justin-mp:
```suggestion
**D1 is a managed database product.** 
```

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+{}
+---
+
+## SQL in Durable Objects vs D1
+
+Cloudflare Workers offers a SQLite-backed serverless database product - [D1](/d1/). How should you compare [SQLite in Durable Objects](/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) and D1?
+
+**D1 should be viewed as a managed database product.** 
+
+D1 fits into a familiar architecture for developers, where application servers communicate with a database over the network. Application servers are typically Workers; however, D1 also supports external, non-Worker access via [HTTP API](https://developers.cloudflare.com/api/operations/cloudflare-d1-create-database).
```
justin-mp: Given that it doesn't seem possible to point to all of the D1 HTTP API methods as a group, it's probably better to point to querying the database, as that's actually a data plane operation.

```suggestion
D1 fits into a familiar architecture for developers, where application servers communicate with a database over the network. Application servers are typically Workers; however, D1 also supports external, non-Worker access via an [HTTP API](https://developers.cloudflare.com/api/operations/cloudflare-d1-query-database). 
```

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+## SQL in Durable Objects vs D1
+
+Cloudflare Workers offers a SQLite-backed serverless database product - [D1](/d1/). How should you compare [SQLite in Durable Objects](/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) and D1?
+
+**D1 should be viewed as a managed database product.** 
+
+D1 fits into a familiar architecture for developers, where application servers communicate with a database over the network. Application servers are typically Workers; however, D1 also supports external, non-Worker access via [HTTP API](https://developers.cloudflare.com/api/operations/cloudflare-d1-create-database). 
+
+D1 aims for a "batteries included" feature set, including HTTP API, [database schema management](/d1/reference/migrations/#_top), [data import/export](/d1/build-with-d1/import-export-data/), database query [insights](/d1/observability/metrics-analytics/#query-insights), and upcoming built-in read replication.
```
justin-mp: We shouldn't mention unreleased features in documentation.

```suggestion
D1 aims for a "batteries included" feature set, including the above HTTP API, [database schema management](/d1/reference/migrations/#_top), [data import/export](/d1/build-with-d1/import-export-data/), database query and [insights](/d1/observability/metrics-analytics/#query-insights). 
```

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+**D1 should be viewed as a managed database product.** 
+
+D1 fits into a familiar architecture for developers, where application servers communicate with a database over the network. Application servers are typically Workers; however, D1 also supports external, non-Worker access via [HTTP API](https://developers.cloudflare.com/api/operations/cloudflare-d1-create-database). 
+
+D1 aims for a "batteries included" feature set, including HTTP API, [database schema management](/d1/reference/migrations/#_top), [data import/export](/d1/build-with-d1/import-export-data/), database query [insights](/d1/observability/metrics-analytics/#query-insights), and upcoming built-in read replication. 
+
+With D1, your application code and SQL database queries are not colocated which can impact application performance. Workers has [Smart Placement](/workers/configuration/smart-placement/#_top) to dynamically run your Worker in the best location to reduce total Worker request latency, considering everything your Worker talks to, including D1.
```
justin-mp:
```suggestion
With D1, your application code and SQL database queries are not colocated which can impact application performance. If performance is a concern with D1, Workers has [Smart Placement](/workers/configuration/smart-placement/#_top) to dynamically run your Worker in the best location to reduce total Worker request latency, considering everything your Worker talks to, including D1.
```

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+**SQLite in Durable Objects is a lower-level compute with storage building block for distributed systems.** By design, Durable Objects are accessed with Workers-only.
+
+Durable Objects require a bit more effort, but in return, give you more flexibility and control. With Durable Objects, you have two pieces of code that run in different places: a front-end Worker which routes incoming requests from the Internet to a unique Durable Object, and the Durable Object itself, which runs on the same machine as the SQLite database.
```
justin-mp:
```suggestion
Durable Objects require a bit more effort, but in return, give you more flexibility and control. With Durable Objects, you must implement two pieces of code that run in different places: a front-end Worker which routes incoming requests from the Internet to a unique Durable Object, and the Durable Object itself, which runs on the same machine as the SQLite database.  You get to choose what runs where and it may be that your application benefits from running some application business logic right next to the database.
```

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+With D1, your application code and SQL database queries are not colocated which can impact application performance. Workers has [Smart Placement](/workers/configuration/smart-placement/#_top) to dynamically run your Worker in the best location to reduce total Worker request latency, considering everything your Worker talks to, including D1.
+
+**SQLite in Durable Objects is a lower-level compute with storage building block for distributed systems.** By design, Durable Objects are accessed with Workers-only.
+
+Durable Objects require a bit more effort, but in return, give you more flexibility and control. With Durable Objects, you have two pieces of code that run in different places: a front-end Worker which routes incoming requests from the Internet to a unique Durable Object, and the Durable Object itself, which runs on the same machine as the SQLite database. 
+ 
+You may need to carefully consider where to run different application code, and you may need to build some of your own database tooling that exists out-of-the-box with D1. But because you are in full control, you can tailor the solution to your application's needs and potentially achieve more.
```
justin-mp:
```suggestion
With SQLite in Durable Objects, you may also need to build some of your own database tooling that comes out-of-the-box with D1.
```

src/content/partials/workers/storage_api_pricing.mdx
```diff
@@ -0,0 +1,55 @@
+---
+{}
+
+---
+
+The Durable Objects [Storage API](/durable-objects/api/storage-api) is only accessible from within Durable Objects.
+
+Durable Objects do not have to use the Storage API, but if your code does call methods on `ctx.storage`, it will additional charges.
```
justin-mp: "it will additional charges" seems to be missing a verb.

src/content/partials/workers/storage_api_pricing.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+### Key-value storage billing
```
justin-mp: If we're using "Key-Value API" and "SQL API" as the two kinds of APIs offered by DOs, then this section needs a new name because it's not how the Key-Value API will be charged with SQLite-backed DOs.

vy-ton: What if the parent section was called "Storage billing" (no API)?

justin-mp: Reading through more of this PR, I think what we're trying to say is that we have two storage backends (Key-Value Storage and SQL Storage) and two storage APIs (Key-Value API and SQL API, collectively called the Storage API) with the Key-Value Storage backend supporting the Key-Value API and the SQL Storage backend supporting both the Key-Value API and the SQL API.

It gets very muddy when we use phrases like "Storage SQL API methods".  Maybe we should ban the term "Storage SQL API" and use only "SQL API"?

src/content/docs/durable-objects/api/storage-api.mdx
justin-mp:
```suggestion
A cursor (`SqlStorageCursor`) to iterate over query row results as objects. `SqlStorageCursor` is a JavaScript [Iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol), which supports iteration using `for (let row of cursor)`. `SqlStorageCursor` is also a JavaScript [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol), which supports iteration using `cursor.next()`.
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+Returned cursor supports the following methods:
```
justin-mp:
```suggestion
`SqlStorageCursor` supports the following methods:
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  
+* `next()`
+  * Returns an object representing the next value of the cursor. Object has `done` and `value` properties. `done` is set to `false` when a next value is present, and `value` is set to the next row object in the query result. `done` is set to `true` when the entire cursor is consumed, and no `value` is set.
```
justin-mp:
```suggestion
  * Returns an object representing the next value of the cursor. The returned object has  `done` and `value` properties adhering to the JavaScript [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol). `done` is set to `false` when a next value is present, and `value` is set to the next row object in the query result. `done` is set to `true` when the entire cursor is consumed, and no `value` is set. 
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+A cursor (`SqlStorageCursor`) to iterate over query row results as objects. Cursor is a JavaScript [Iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol), which supports iteration using `for (let row of cursor)`. Cursor is also a JavaScript [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol), which supports iteration using `cursor.next()`.
+
+Returned cursor supports the following methods:
+  
+* `next()`
+  * Returns an object representing the next value of the cursor. Object has `done` and `value` properties. `done` is set to `false` when a next value is present, and `value` is set to the next row object in the query result. `done` is set to `true` when the entire cursor is consumed, and no `value` is set. 
+* `toArray()`
+  * Iterates through remaining cursor value(s) and returns an array of returned row objects. 
+* `one()`
+  * Returns a row object if query result has exactly one row. If query result has zero or more than one row, `one()` returns an error.
```
justin-mp:
```suggestion
  * Returns a row object if query result has exactly one row. If query result has zero rows or more than one row, `one()` throws an exception. 
```

src/content/partials/workers/storage_api_pricing.mdx
vy-ton: > Reading through more of this PR, I think what we're trying to say is that we have two storage backends (Key-Value Storage and SQL Storage) and two storage APIs (Key-Value API and SQL API, collectively called the Storage API) with the Key-Value Storage backend supporting the Key-Value API and the SQL Storage backend supporting both the Key-Value API and the SQL API.

Yes, this is what I'm trying to convey.

> It gets very muddy when we use phrases like "Storage SQL API methods". Maybe we should ban the term "Storage SQL API" and use only "SQL API"?

I'm good to stick with `SQL API` for clarity. `Storage SQL API` was trying to explicitly convey the `Storage API > SQL API` hierarchy but it doesnt work

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  * Returns an Iterator over the same query results, with each row as an array of column values (with no column names) rather than an object. 
+  * Returned Iterator supports `next()`, `toArray()`, and `one()` methods above.
+  * Returned cursor and `raw()` iterator iterate over the same query results and can be combined.
+
+```ts
+  let cursor = this.sql.exec("SELECT * FROM artist ORDER BY artistname ASC;");
+  let result = cursor.raw().next();
+
+  if (!result.done) {
+    console.log(result.value); // prints [ artistid: 123, artistname: 'Alice' ]
```
justin-mp:
```suggestion
    console.log(result.value); // prints [ 123, 'Alice' ]
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+* `one()`
+  * Returns a row object if query result has exactly one row. If query result has zero or more than one row, `one()` returns an error. 
+* `raw()`: Iterator
+  * Returns an Iterator over the same query results, with each row as an array of column values (with no column names) rather than an object. 
+  * Returned Iterator supports `next()`, `toArray()`, and `one()` methods above.
+  * Returned cursor and `raw()` iterator iterate over the same query results and can be combined.
+
+```ts
+  let cursor = this.sql.exec("SELECT * FROM artist ORDER BY artistname ASC;");
+  let result = cursor.raw().next();
```
justin-mp: You may want to call this rawResult.

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+  if (!result.done) {
+    console.log(result.value); // prints [ artistid: 123, artistname: 'Alice' ]
+  } else {
+    // query returned zero results
+  }
+
+  console.log(cursor.toArray()); // prints [{ artistid: 456, artistname: 'Bob' },{ artistid: 789, artistname: 'Charlie' }]
```
justin-mp, line 272: I get that this example is showing multiple things, but it seems kinda weird to have cursor.toArray() in this example.  It would be better to have individual examples.

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+```
+
+* `columnNames()`: string[]
```
justin-mp: columnNames is a property, not a method.

```suggestion
* `columnNames`: string[]
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  * Consumes the returned cursor and returns an array of column name strings. Often used with `raw()` to find index of specific column(s).
```
justin-mp:
```suggestion
  * The column names of the query in the order they appear in each row array returned by the `raw` iterator.
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+*  `rowsRead()`: number
```
justin-mp: `rowsRead` is a property, not a method.

```suggestion
*  `rowsRead`: number
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  * Returns the number of rows read so far as part of this SQL `query`. This may increase as you iterate the cursor. The final value is used for [SQL billing](/durable-objects/platform/pricing/#sql-storage-billing).
+*  `rowsWritten()`: number
```
justin-mp: `rowsWritten is a property, not a method.

```suggestion
*  `rowsWritten`: number
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+    // query returned zero results
+  }
+
+  console.log(cursor.toArray()); // prints [{ artistid: 456, artistname: 'Bob' },{ artistid: 789, artistname: 'Charlie' }]
+```
+
+* `columnNames()`: string[]
+  * Consumes the returned cursor and returns an array of column name strings. Often used with `raw()` to find index of specific column(s).
+*  `rowsRead()`: number
+  * Returns the number of rows read so far as part of this SQL `query`. This may increase as you iterate the cursor. The final value is used for [SQL billing](/durable-objects/platform/pricing/#sql-storage-billing).
```
justin-mp:
```suggestion
  * The number of rows read so far as part of this SQL `query`. This may increase as you iterate the cursor. The final value is used for [SQL billing](/durable-objects/platform/pricing/#sql-storage-billing).
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+*  `rowsWritten()`: number
+  * Returns the number of rows written so far as part of this SQL `query`. This may increase as you iterate the cursor. The final value is used for [SQL billing](/durable-objects/platform/pricing/#sql-storage-billing).
```
justin-mp:
```suggestion
  * The number of rows written so far as part of this SQL `query`. This may increase as you iterate the cursor. The final value is used for [SQL billing](/durable-objects/platform/pricing/#sql-storage-billing).
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+#### Returns
+The current SQLite database size in bytes.
+
+```ts
+  let size = ctx.storage.sql.databaseSize;
+```
+
+### Point in time recovery
+
+For [Durable Objects classes with SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-new-durable-object-class-migration), the following point-in-time-recovery (PITR) API methods are available to restore a Durable Object's embedded SQLite database to any point in time in the past 30 days. These methods apply to the entire SQLite database contents, including both the object's stored SQL data and stored key-value data using the key-value `put()` API. PITR API is not supported in local development because a durable log of data changes is not stored locally.
```
justin-mp:
```suggestion
For [Durable Objects classes with SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-new-durable-object-class-migration), the following point-in-time-recovery (PITR) API methods are available to restore a Durable Object's embedded SQLite database to any point in time in the past 30 days. These methods apply to the entire SQLite database contents, including both the object's stored SQL data and stored key-value data using the key-value `put()` API. The PITR API is not supported in local development because a durable log of data changes is not stored locally. 
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+```ts
+  let size = ctx.storage.sql.databaseSize;
+```
+
+### Point in time recovery
+
+For [Durable Objects classes with SQL storage](/durable-objects/reference/durable-objects-migrations/#enable-sql-storage-on-new-durable-object-class-migration), the following point-in-time-recovery (PITR) API methods are available to restore a Durable Object's embedded SQLite database to any point in time in the past 30 days. These methods apply to the entire SQLite database contents, including both the object's stored SQL data and stored key-value data using the key-value `put()` API. PITR API is not supported in local development because a durable log of data changes is not stored locally. 
+
+The point-in-time-recovery (PITR) API represents points in times using "bookmarks". A bookmark is a mostly alphanumeric string like `0000007b-0000b26e-00001538-0c3e87bb37b3db5cc52eedb93cd3b96b`. Bookmarks are designed to be lexically comparable: a bookmark representing an earlier point in time compares less than one representing a later point, using regular string comparison.
```
justin-mp:
```suggestion
The PITR API represents points in times using "bookmarks". A bookmark is a mostly alphanumeric string like `0000007b-0000b26e-00001538-0c3e87bb37b3db5cc52eedb93cd3b96b`. Bookmarks are designed to be lexically comparable: a bookmark representing an earlier point in time compares less than one representing a later point, using regular string comparison.
```

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -216,9 +224,102 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+
+  * Returns a bookmark representing the current point in time in the object's history.
+
+<code>ctx.storage.getBookmarkForTime(timestamp: number | Date)</code>: Promise\<string>
+
+  * Returns a bookmark representing approximately the given point in time, which must be within the last 30 days. If the timestamp is represented as a number, it is converted to a date as if using `new Date(timestamp)`.
+
+<code>ctx.storage.onNextSessionRestoreBookmark(bookmark: string)</code>: Promise\<string>
+
+  * Instructs that the next time the Durable Object restarts, it should restore its storage to exactly match what it contained at the given bookmark. After calling this, the application should typically invoke `ctx.abort()` to restart the Durable Object, thus completing the point-in-time recovery.
```
justin-mp:
```suggestion
  * Configure the Durable Object so that the next time it restarts, it should restore its storage to exactly match what the storage contained at the given bookmark. After calling this, the application should typically invoke `ctx.abort()` to restart the Durable Object, thus completing the point-in-time recovery.
```

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -0,0 +1,69 @@
...
+title: Access Durable Objects Storage
+pcx_content_type: concept
+sidebar:
+  order: 4
+
+---
+
+import { Render } from "~/components";
+
+Durable Objects are a powerful compute API that provides a compute with storage building block. Each Durable Object has private, transactional and strongly consistent storage attached. By default, key-value storage is available. SQL storage(/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) is available on a new beta version of Durable Objects.
```
justin-mp: I think the "SQL Storage" link in the second sentence is broken.

justin-mp:
```suggestion
Durable Objects are a powerful compute API that provides a compute with storage building block. Each Durable Object has it's own private, transactional and strongly consistent storage. By default, key-value storage is available. SQL storage(/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) is available on a new beta version of Durable Objects.
```

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+You should use Durable Objects storage to persist state durably on-disk that needs to survive beyond the lifetime of a Durable Object, that is, eviction of a Durable Object from memory. A Durable Object is alive based on the request traffic incoming to the instance. There are also normal operations like [code deployments](/workers/configuration/versions-and-deployments/) that trigger a Durable Object to reset. For these reasons, Durable Objects storage is recommended for persisting [in-memory state](/durable-objects/reference/in-memory-state/).
```
justin-mp: I don't understand what this paragraph is trying to tell me.

I _think_ it's trying to make the distinction between in-memory and on-disk storage, but that doesn't follow from the first paragraph.  Perhaps the first paragraph needs to mention that DOs persist across requests?

src/content/partials/durable-objects/durable-objects-sql.mdx
```diff
@@ -0,0 +1,109 @@
...
+  console.log(remainingRows); // prints [{ artistid: 456, artistname: 'Bob' },{ artistid: 789, artistname: 'Charlie' }]
+```
+
+Returned cursor and `raw()` iterator iterate over the same query results:
+```ts
+  let cursor = this.sql.exec("SELECT * FROM artist ORDER BY artistname ASC;");
+  let result = cursor.raw().next();
+
+  if (!result.done) {
+    console.log(result.value); // prints [ artistid: 123, artistname: 'Alice' ]
```
justin-mp:
```suggestion
    console.log(result.value); // prints [ 123, 'Alice' ]
```

src/content/partials/workers/storage_api_pricing.mdx
justin-mp: Trying to refine this a bit more: I wonder if rather than using "Key-Value Storage" and "SQL Storage", we used "Key-Value Backend" and "SQL Backend" (or "SQLite Backend") so that the word storage is only associated with APIs.

src/content/docs/durable-objects/api/storage-api.mdx
vy-ton: > Returned cursor and `raw()` iterator iterate over the same query results and can be combined.

Wanted this example embedded here to highlight the above bullet.

There are individual examples for different tasks below on this page

src/content/docs/durable-objects/api/storage-api.mdx
vy-ton: I see the below in workers-types. Is this changing with the updated APIs?

```
declare abstract class SqlStorageCursor {
  ...
  get columnNames(): string[];
  ...
}```

File: src/content/partials/workers/storage_api_pricing.mdx (hunk 39 above)
Comment (vy-ton) [reply src/content/partials/workers/storage_api_pricing.mdx]: @justin-mp Some examples to confirm we like the direction before I change files:

- "users opt-in to SQL storage backend when deploying new Durable Object class"
- "Storage API provides access to Durable Objects' attached storage. Storage API has KV APIs and SQL APIs. Only classes with a SQL storage backend can access SQL APIs."
- "DOs are billed for Storage API usage. Pricing depends on storage backend type, either key-value or SQLite."

Terms:
- storage backend
- Storage API encompassing KV API, SQL API, and PITR API

File: src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx (hunk 56 above)
Comment (vy-ton) [reply src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx]: Reworked, lmk if it helped.

File: src/content/partials/workers/storage_api_pricing.mdx (hunk 39 above)
Comment (elithrar) [reply src/content/partials/workers/storage_api_pricing.mdx]: Storage backend or "storage subsystem" (what we used for D1) work for me.

Comment (justin-mp) [reply src/content/partials/workers/storage_api_pricing.mdx]: I think storage backend works better here.  To summarize, we have:

* 2 storage backends:
  * Key-value backend
  * SQLite backend
* 1 Storage API with 3 sub-APIs:
  * KV API (supported by both storage backends)
  * SQL API (supported by only the SQLite backend)
  * PITR API (supported by only the SQLite backend)

File: src/content/partials/durable-objects/durable-objects-vs-d1.mdx [hunk 58]
```diff
@@ -0,0 +1,23 @@
...
+{}
+---
+
+## SQL in Durable Objects vs D1
+
+Cloudflare Workers offers a SQLite-backed serverless database product - [D1](/d1/). How should you compare [SQLite in Durable Objects](/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) and D1?
+
+**D1 is a managed database product.** 
+
+D1 fits into a familiar architecture for developers, where application servers communicate with a database over the network. Application servers are typically Workers; however, D1 also supports external, non-Worker access via an [HTTP API](https://developers.cloudflare.com/api/operations/cloudflare-d1-query-database).
```
joshthoward: nit: I think it's a missed opportunity not to highlight the framework support that D1 has and that DO probably cannot have.

src/content/partials/workers/storage_api_pricing.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+<sup>5</sup>  Durable Objects will be billed for stored data until the data is removed. Once the data is removed, the object will be cleaned up automatically by the system.
+
+Requests that hit the [Durable Objects in-memory cache](/durable-objects/reference/in-memory-state/) or that use the [multi-key versions of `get()`/`put()`/`delete()` methods](/durable-objects/api/storage-api/) are billed the same as if they were a normal, individual request for each key.
+
+### SQL storage billing 
+
+:::note[SQLite in Durable Objects Beta]
+
+New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. When deploying a new Durable Object class, users can [opt-in to using SQL storage](/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) in order to access [Storage SQL API methods](/durable-objects/api/storage-api/#sqlexec).
```
joshthoward: nit: The new beta version...

src/content/docs/workers/platform/storage-options.mdx
```diff
@@ -200,3 +200,13 @@ To get started with Vectorize:
 - [Create your first vector database](/vectorize/get-started/intro/).
 - Combine [Workers AI and Vectorize](/vectorize/get-started/embeddings/) to generate, store and query text embeddings.
 - Learn more about [how vector databases work](/vectorize/reference/what-is-a-vector-database/).
+
+<Render file="durable-objects-vs-d1" product="durable-objects" />
+
+:::note[SQLite in Durable Objects Beta]
+
+New beta version of Durable Objects is available where each Durable Object has a private, embedded SQLite database. When deploying a new Durable Object class, users can opt-in to using SQL storage in order to access [Storage SQL API methods](/durable-objects/api/storage-api/#sqlexec). Otherwise, a Durable Object class has the standard, private key-value storage.
```
joshthoward: nit: The new beta version...

joshthoward (approved):

justin-mp:
I have no idea where this comment went, but @vy-ton asked

>I see the below in workers-types. Is this changing with the updated APIs?
>
> ```
> declare abstract class SqlStorageCursor {
>  ...
>  get columnNames(): string[];
>  ...
> }```

This defines `columnNames` as a read-only property, despite looking like a function.  https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -0,0 +1,69 @@
...
+sidebar:
+  order: 4
+
+---
+
+import { Render } from "~/components";
+
+Durable Objects are a powerful compute API that provides a compute with storage building block. Each Durable Object has it's own private, transactional and strongly consistent storage. By default, key-value storage is available. SQL storage(/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) is available on a new beta version of Durable Objects.
+
+A Durable Object's [in-memory state](/durable-objects/reference/in-memory-state/) is preserved as long as the Durable Object is not evicted from memory. Inactive Durable Objects with no incoming request traffic can be evicted. There are normal operations like [code deployments](/workers/configuration/versions-and-deployments/) that trigger Durable Objects to restart and lose their in-memory state. For these reasons, you should use Durable Objects Storage API to persist state durably on-disk that needs to survive eviction or restart of Durable Objects.
```
justin-mp:
```suggestion
A Durable Object's [in-memory state](/durable-objects/reference/in-memory-state/) is preserved as long as the Durable Object is not evicted from memory. Inactive Durable Objects with no incoming request traffic can be evicted. There are normal operations like [code deployments](/workers/configuration/versions-and-deployments/) that trigger Durable Objects to restart and lose their in-memory state. For these reasons, you should use Durable Objects Storage API to persist state durably on disk that needs to survive eviction or restart of Durable Objects. 
```

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -0,0 +1,69 @@
...
+title: Access Durable Objects Storage
+pcx_content_type: concept
+sidebar:
+  order: 4
+
+---
+
+import { Render } from "~/components";
+
+Durable Objects are a powerful compute API that provides a compute with storage building block. Each Durable Object has it's own private, transactional and strongly consistent storage. By default, key-value storage is available. SQL storage(/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) is available on a new beta version of Durable Objects.
```
Oxyjun:
```suggestion
Durable Objects are a powerful compute API that provides a compute with storage building block. Each Durable Object has its own private, transactional and strongly consistent storage. By default, key-value storage is available. SQL storage(/durable-objects/best-practices/access-durable-objects-storage/#sql-storage) is available on a new beta version of Durable Objects.
```

src/content/glossary/durable-objects.yaml
```diff
@@ -1,16 +1,52 @@
...
+  general_definition: |-
+    A container for a collection of Durable Objects that all share the same Durable Object (class) definition. A single Namespace can have (tens of) millions of Durable Objects. Metrics are scoped per Namespace.
+
+- term: "Durable Object"
+  general_definition: |-
+    An individual Durable Object. A Durable Object is globally unique (referenced by ID), provides a global point of coordination for all methods/requests sent to it, and has private, persistent storage that is not shared with other Durable Objects within a Namespace.
+
+- term: "Stub"
+  general_definition: |-
+    An object that refers to a unique Durable Object within a Namespace and allows you to call into that Durable Object via RPC methods or the `fetch` API. - e.g. `let stub = env.MY_DURABLE_OBJECT.get(id)`
```
Oxyjun:
```suggestion
    An object that refers to a unique Durable Object within a Namespace and allows you to call into that Durable Object via RPC methods or the `fetch` API. For example, `let stub = env.MY_DURABLE_OBJECT.get(id)`
```

src/content/glossary/durable-objects.yaml
```diff
@@ -1,16 +1,52 @@
...
+  general_definition: |-
+    The strongly consistent (serializable) [key-value storage API](/durable-objects/api/storage-api/) for persisting data within each Durable Object instance. State stored within a unique DO instance is "private" to that Durable Object, and not accessible from other Durable Objects.
+
+- term: "SQL Storage API"
+  general_definition: |-
+    The strongly consistent SQL API for querying and persisting data within each SQL-enabled Durable Object instance. State stored within a unique DO instance is "private" to that Durable Object, and not accessible from other Durable Objects. 
+
+- term: migration
+  general_definition: |-
+    a Durable Object migration is a mapping process from a class name to a runtime state.
```
Oxyjun:
```suggestion
    A Durable Object migration is a mapping process from a class name to a runtime state.
```

src/content/glossary/durable-objects.yaml
```diff
@@ -1,16 +1,52 @@
...
+    Initiate a Durable Object migration when you need to:
+
+    - Create a new Durable Object class.
+    - Rename a Durable Object class.
+    - Delete a Durable Object class.
+    - Transfer an existing Durable Objects class.
+
+- term: Alarm
+  general_definition: |-
+    a Durable Object alarm is a mechanism that allows you to schedule the Durable Object to be woken up at a time in the future.
```
Oxyjun:
```suggestion
    A Durable Object alarm is a mechanism that allows you to schedule the Durable Object to be woken up at a time in the future.
```

src/content/partials/durable-objects/durable-objects-sql.mdx
```diff
@@ -0,0 +1,109 @@
...
+  let columnNameArray = this.sql.exec("SELECT * FROM artist;").columnNames.toArray();
+```
+
+Get first row object of query results:
+```ts
+  // Returns {"artistid":123,"artistname":"Alice"}
+  let firstRow = this.sql.exec("SELECT * FROM artist ORDER BY artistname DESC;").toArray()[0];
+```
+
+Check if query results has exactly one row:
```
Oxyjun:
```suggestion
Check if query results have exactly one row:
```

src/content/partials/durable-objects/durable-objects-vs-d1.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+D1 fits into a familiar architecture for developers, where application servers communicate with a database over the network. Application servers are typically Workers; however, D1 also supports external, non-Worker access via an [HTTP API](https://developers.cloudflare.com/api/operations/cloudflare-d1-query-database). 
+
+D1 aims for a "batteries included" feature set, including the above HTTP API, [database schema management](/d1/reference/migrations/#_top), [data import/export](/d1/build-with-d1/import-export-data/), and [database query insights](/d1/observability/metrics-analytics/#query-insights). 
+
+With D1, your application code and SQL database queries are not colocated which can impact application performance. If performance is a concern with D1, Workers has [Smart Placement](/workers/configuration/smart-placement/#_top) to dynamically run your Worker in the best location to reduce total Worker request latency, considering everything your Worker talks to, including D1.
+
+**SQLite in Durable Objects is a lower-level compute with storage building block for distributed systems.** By design, Durable Objects are accessed with Workers-only.
```
Oxyjun:
```suggestion
**SQLite in Durable Objects is a lower-level compute with storage building block for distributed systems.**

By design, Durable Objects are accessed with Workers-only.
```

Oxyjun (approved):

PR #16931, opened 2024-09-18
[GitHub] Separate preview and production actions

kodster28:
### Summary

We're running into a "fun" mix of issues due to:
- Dogfooding Workers Assets (which cause our build times to get stretched out b/c we're simultaneously publishing to Pages). This'll be partially solved by #16905, but think it'll still be hanging out as an issue.
- Moving from preview / production builds in Pages to a GitHub action, which was a limitation due to the size of our new site in Astro and the Pages CI pipeline.
- Concurrency checks that can cause `Production` merges to get either queued or cancelled behind each other. This definitely makes sense for `Preview` builds and was added in #16865.
- However, our team also has to publish pretty quickly and canceling deploys to production can be frustrating / confusing
image

### Proposal

Separate out our preview / production branch workflows into separate actions:
- *Preview* -> Keeps concurrency checks and Workers assets (only runs on branches)
- *Production* -> Removes concurrency and Workers assets publishing (and only triggers based on PRs being [merged into production](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges))

Open to convo / thoughts, @GregBrimble / @penalosa / @KianNH / @pedrosousa.


.github/workflows/publish-preview.yml
```diff
@@ -0,0 +1,48 @@
...
+      - run: npx astro build
+        env:
+          NODE_OPTIONS: --max-old-space-size=4096
+      - run: npx wrangler pages deploy --project-name cloudflare-docs dist
+        name: Deploy to Cloudflare Pages
+        env:
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      - run: npm run build:worker && npx wrangler versions upload -c ./wrangler-workers.toml
+        if: github.ref != 'refs/heads/production'
```
GregBrimble: Don't need this any more!

GregBrimble (approved):

.github/workflows/publish-production.yml
```diff
@@ -0,0 +1,40 @@
+on:
+  pull_request:
+    types:
+      - closed
```
penalosa: Should this be pushes to the `production` branch?

kodster28: Think that gets us there with:

```
on:
  pull_request:
    branches:
      - production
    types:
      - closed
...
    if: github.event.pull_request.merged == true 
```

penalosa: As in, could this just be for the `push` event to the `production` branch? Is there a reason it needs to be `pull_request` triggered?

penalosa:
Does the `publish.yml` workflow also need to be deleted?

kodster28:
> Does the `publish.yml` workflow also need to be deleted?

ahhhhh..... yep. You mean you didn't want even more workflows taking up our org bandwidth 😅 ?

penalosa (approved):

KianNH (approved):

.github/workflows/publish-production.yml
kodster28: > As in, could this just be for the `push` event to the `production` branch? Is there a reason it needs to be `pull_request` triggered?

Ah, that's fair feedback (apologies, should've seen before merging). Will PR it to be 

```
on:
  push:
    branches:
      - 'production'
```

PR #16975, opened 2024-09-20
[Docs Site] Add default title to PowerShell codeblocks

KianNH:
### Summary

Add default title to PowerShell codeblocks


pedrosousa (approved):
Tested that:
- The default title appears
- We can override the default title

PR #16978, opened 2024-09-20
[Docs] Update PowerShell code blocks

pedrosousa:
### Summary

Updates PowerShell code blocks based on a recent internal discussion.
Removing the titles because code block title/headers will be set automatically to "PowerShell" when #16975 gets merged.


daisyfaithauma (approved):
I think this new Powershell approach looks better!

PR #17007, opened 2024-09-22
[Calls / TURN] Add note about RTCPeerConnection structure

benfoxall:
### Summary

Maybe this doc change could be clearer, but here's the issue I was having:

I was stuck for quite a while trying to pass on turn credentials to an `RTCPeerConnection` via [y-webrtc](https://github.com/yjs/y-webrtc) & [simple-peer](https://github.com/feross/simple-peer), it seems that I was passing along the ice credentials incorrectly.

Here's how I ended up modifying the credentials:
```js
const TURN_ID = 'secret_id';
const TOKEN = 'secret_token';

// Fetch TURN credentials from the API (don't worry, I'm using a worker in the real thing!)
const response = await fetch(`https://rtc.live.cloudflare.com/v1/turn/keys/${TURN_ID}/credentials/generate`, {
  method: 'POST',
  headers: {
    'Authorization': `Bearer ${TOKEN}`,
  },
  body: JSON.stringify({ ttl: 86400 }),
});

// Parse the response
const config = await response.json();

try {
  // Attempt to create a new RTCPeerConnection using the returned config
  const pc = new RTCPeerConnection(config);
} catch (e) {
  // Error occurs here because config.iceServers is not iterable
  console.error('Error constructing RTCPeerConnection:', e);
  // Error message: TypeError: Failed to construct 'RTCPeerConnection': 
  // Failed to read the 'iceServers' property from 'RTCConfiguration': 
  // The object must have a callable @@iterator property.
}

// Fix the issue by wrapping iceServers in an array
config.iceServers = [config.iceServers]; // Ensure iceServers is an array

try {
  // Attempt to create the RTCPeerConnection again with the fixed config
  const pc = new RTCPeerConnection(config);
  console.log('RTCPeerConnection successfully created.');
} catch (e) {
  // This block won't be called as the iceServers issue has been resolved
  console.error('Unexpected error constructing RTCPeerConnection:', e);
}
```

### Documentation checklist

### DIFF
--- a/src/content/docs/calls/turn/generate-credentials.mdx
+++ b/src/content/docs/calls/turn/generate-credentials.mdx
@@ -47,7 +47,7 @@ The JSON response below can then be passed on to your front-end application:
 }
 ```
 
-Use `username` and `credential` as follows when instantiating the `RTCPeerConnection`:
+Use `username` and `credential` as follows when instantiating the `RTCPeerConnection` (note, that `iceServers` is an array or objects):
 
 ```js
 const myPeerConnection = new RTCPeerConnection({
--- a/src/content/docs/calls/turn/generate-credentials.mdx
+++ b/src/content/docs/calls/turn/generate-credentials.mdx
@@ -47,7 +47,7 @@ The JSON response below can then be passed on to your front-end application:
 }
 ```
 
-Use `username` and `credential` as follows when instantiating the `RTCPeerConnection` (note, that `iceServers` is an array or objects):
+Use `username` and `credential` as follows when instantiating the `RTCPeerConnection` (note, `iceServers` is now an array):
 
 ```js
 const myPeerConnection = new RTCPeerConnection({


KimJ15:
@dcpena reassigning this PR to you.

nils-ohlmeier:
I think the pseudo code already shows an array. And as far as I can remember the change for iceServers (not the plural here) to be an array is quite old, see https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#browser_compatibility

The word 'now' suggests to me that this is something new, which I don't think is the case. If we remove the word 'now' from the PR I have no problem with the extra hint about it being an array.

@renandincer what do you think?

renandincer:
let's ship it! thanks for the contribution @benfoxall

renandincer (approved):

benfoxall:
>  the word 'now' 

I meant this as in "it was an object in the credential response, but now it's an array", tbh the wording isn't that great, but thanks for merging 😅

PR #17014, opened 2024-09-23
[BYOIP] Service bindings and MT with CDN guide

RebeccaTamachiro:
### Summary

PCX-10710: Magic Transit + CDN Reverse Proxy

### Documentation checklist


src/content/docs/byoip/service-bindings/magic-transit-with-cdn.mdx
```diff
@@ -0,0 +1,209 @@
...
+</TabItem>
+</Tabs>
+
+:::note
+As you create the necessary DNS records, [Total TLS](/ssl/edge-certificates/additional-options/total-tls/) can help making sure that you have SSL/TLS certificates in place for all your hostnames.
+:::
+
+While the DNS record proxy status and address map will determine how Cloudflare's authoritative DNS responds to requests for your hostnames, the IP addresses specified in `A`/`AAAA` records will determine [how Cloudflare reaches the configured origin](/fundamentals/concepts/how-cloudflare-works/#how-cloudflare-works-as-a-reverse-proxy).
+
+<Details header="Example" >
```
ricardomacas, line 185: In this example, the origin IP (.150) is different from the public facing IP (.100) that is bound to CDN.

However, the most common use case is when these are the same and the IP is "transparently" proxied. In that case, the IP address listed on the record is the same as that in the binding (with the same externally bound IP as the origin IP, in this case, setting the CDN service binding in .150 and using the same .150 in the DNS). 

Even if the example is of a different IP (it allows the reader to better understand the role of each field) it would still be worth to add a note that transparent proxying (preserving the origin IP) is possible by simply binding the CDN or Spectrum service to the same IP as the origin IP set in DNS (this will not cause any loop).

RebeccaTamachiro: Thanks! I've added lines 203-213 below to address this.

daisyfaithauma (approved):

PR #17077, opened 2024-09-24
Mchoi/wpartialasides

Meahaa:
### Summary

Remove aside titles from remaining partials with folders with w titles

### DIFF
--- a/src/content/partials/waf/managed-lists.mdx
+++ b/src/content/partials/waf/managed-lists.mdx
@@ -13,5 +13,5 @@
 <br />
 
 :::caution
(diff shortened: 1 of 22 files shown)


src/content/partials/waf/managed-lists.mdx
```diff
@@ -12,6 +12,6 @@
 
 <br />
 
-:::caution[<a id="vpn-important-remarks"></a>Important remarks about the Cloudflare VPNs Managed List]
```
pedrosousa, line 15: We need to keep the anchor because this aside is the target of a link in a different page.

Meahaa: makes sense! I'll add it to the body content then since I'm removing the headers

Meahaa (commented):
update managed-lists

Meahaa:
duplicate

PR #17120, opened 2024-09-26
Apply beta badge to all of builds' docs

GregBrimble:
with my change

image

vs. without

image


irvinebroque (approved):
TIL possible, ty!

PR #17133, opened 2024-09-26
[WAF] fix IPv6 range sizes in custom-lists.mdx

favonia:
### Summary

The documentation does not match the actual API behavior. Attempting to add an IPv6 CIDR range as large as `/4` (as documented) will get the error 10035: `invalid value for ip at position 0: filters.api.IPv6 CIDRs should have a prefix length from /12 to /64 inclusive`.

### Documentation checklist


pedrosousa (approved):

pedrosousa:
Thank you for the fix, @favonia!

PR #17142, opened 2024-09-26
[SSL] Add callout about min TLS version and R2 custom domains

RebeccaTamachiro:
### Summary

Supersedes #15932 
Call out that minimum TLS setup from SSL dash/API does not apply to R2 custom domains

### DIFF
--- a/src/content/docs/ssl/edge-certificates/additional-options/minimum-tls.mdx
+++ b/src/content/docs/ssl/edge-certificates/additional-options/minimum-tls.mdx
@@ -13,9 +13,7 @@ For example, if TLS 1.1 is selected, visitors attempting to connect using TLS 1.
 
 :::note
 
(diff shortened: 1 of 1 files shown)


src/content/docs/ssl/edge-certificates/additional-options/minimum-tls.mdx
```diff
@@ -35,6 +33,10 @@ If you have an [Advanced Certificate Manager](/ssl/edge-certificates/advanced-ce
 
 ## Setup
 
+:::caution
+The Minimum TLS version that you set up following these steps does not apply to [R2](/r2/) custom domains. To control the TLS version for R2 custom domains refer to the [custom domains documentation](/r2/buckets/public-buckets/#minimum-tls-version).
```
dcpena:
```suggestion
The Minimum TLS version that you set up following these steps does not apply to [R2](/r2/) custom domains. To control the TLS version for R2 custom domains, refer to the [custom domains documentation](/r2/buckets/public-buckets/#minimum-tls-version).
```

dcpena (approved):

PR #17181, opened 2024-09-27
Remove some alpha/beta info from D1 wrangler command docs

a-robinson:
The "Report bugs" notes are generally only used for beta products, but D1 has been GA since the spring, so I don't think it's really needed anymore (unless you'd want to make the claim that all products should have a note like that).

And the Export command doesn't really need a caution about alpha DBs, given that alpha DBs are fully deprecated at this point (https://developers.cloudflare.com/d1/platform/alpha-migration/)

@vy-ton


vy-ton (approved):

vy-ton:
@Oxyjun Can you stamp approval and merge?

KianNH (approved):

PR #17189, opened 2024-09-28
Fix typo in vector tutorial

andreaselia:
### Summary

Vector documentation tutorial was using `DATABASE` in the delete method instead of `DB` like elsewhere.


KianNH (approved):
Thanks for fixing this!

PR #17196, opened 2024-09-29
Clarify node.js vs edge runtime in Next.js <> Pages docs

irvinebroque:
Provide more guidance for which adapter to use when, and what will/won't work.


petebacondarwin (approved):

KimJ15 (approved):
Also approved

PR #17197, opened 2024-09-29
Callout @opennextjs/cloudflare from Next.js Pages SSR docs

irvinebroque:
When you land on this page:

Screenshot 2024-09-29 at 12 22 57 PM

We should let you know that `@opennextjs/cloudflare` exists, and give you links and context that lets you choose what is right for your use case.


petebacondarwin (approved):

KimJ15 (approved):
LGTM

PR #17240, opened 2024-09-30
[Docs Site] Don't show breadcrumbs on /products/

KianNH:
### Summary

Don't show breadcrumbs on /products/

### DIFF
--- a/src/components/overrides/PageTitle.astro
+++ b/src/components/overrides/PageTitle.astro
@@ -19,6 +19,10 @@ const updated = Astro.props.entry.data.updated;
 const badge = Astro.props.entry.data.sidebar?.badge;
 const summary = Astro.props.entry.data.summary;
 
+const slug = Astro.props.entry.slug;
+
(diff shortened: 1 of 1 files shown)


src/components/overrides/PageTitle.astro
```diff
@@ -19,6 +19,10 @@ const updated = Astro.props.entry.data.updated;
 const badge = Astro.props.entry.data.sidebar?.badge;
 const summary = Astro.props.entry.data.summary;
 
+const slug = Astro.props.entry.slug;
+
+const showBreadcrumbs = slug !== "products";
```
GregBrimble: Tiny nit just to make it easier to add any other pages in the future.

```suggestion
const showBreadcrumbs = !["products"].includes(slug);
```

GregBrimble (approved):

src/components/overrides/PageTitle.astro
KianNH: Sounds good to me!

PR #17243, opened 2024-09-30
Clarify target and infra app setup

asamborski:
### Summary

Clarifying target and app creation setup. Specifically calling out: 
- Target hostnames are not used in DNS address resolution.
- Targets cannot have the same IP address and VNET pairings.
- Users must switch to the VNET the target is located within before they execute an SSH connection request.
- Clarify details around Gateway SSH proxy certificate vs. other short-lived certificates.
- Adding in specific section for split tunnel gotcha for allowing connections to addresses under RFC 1918.

### DIFF
--- a/src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
+++ b/src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
@@ -20,8 +20,6 @@ import { Tabs, TabItem, Badge, Render } from "~/components";
 
 2. In the **Private Networks** tab for the tunnel, enter the private IP address of your server (or a range that includes the server IP).
 
-<Render file="tunnel/warp-to-tunnel-route-ips" />
-
 ## 2. Set up the client
 
 To connect your devices to Cloudflare:
--- a/src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
+++ b/src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
@@ -20,6 +20,8 @@ import { Tabs, TabItem, Badge, Render } from "~/components";
 
 2. In the **Private Networks** tab for the tunnel, enter the private IP address of your server (or a range that includes the server IP).
 
+<Render file="tunnel/warp-to-tunnel-route-ips" />
+
 ## 2. Set up the client
 
 To connect your devices to Cloudflare:
@@ -62,7 +64,7 @@ Next, configure your SSH server to trust the Cloudflare SSH CA. This allows Acce
 
 ## 8. Connect as a user
 
-Users can use any SSH client to connect to the target, as long as they are logged into the WARP client on their device. Users do not need to modify any SSH configs on their device. For example, to SSH from a terminal:
+Users can use any SSH client to connect to the target, as long as they are logged into the WARP client on their device. If the target is located within a particular virtual network, ensure that the WARP client is connected to that virtual network before initiating the connection. Users do not need to modify any SSH configs on their device. For example, to SSH from a terminal:
 
 ```sh
 ssh <username>@<target IP>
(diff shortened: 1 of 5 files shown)


src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
```diff
@@ -20,6 +20,8 @@ import { Tabs, TabItem, Badge, Render } from "~/components";
 
 2. In the **Private Networks** tab for the tunnel, enter the private IP address of your server (or a range that includes the server IP).
 
+<Render file="tunnel/warp-to-tunnel-route-ips" />
```
asamborski: @ranbel - Calling this change to your attention specifically. I think we need to highlight what's in this file to users, and I'd love your feedback on if this is the right block to reuse or if we should make any changes to format/copy.

src/content/partials/cloudflare-one/ssh/ssh-proxy-ca.mdx
```diff
@@ -5,6 +5,11 @@
 
 import { Render } from "~/components"
 
+:::note
```
asamborski: @ranbel - this is duplicate copy from [here](https://developers.cloudflare.com/cloudflare-one/policies/gateway/network-policies/ssh-logging/#2-generate-a-gateway-ssh-proxy-ca). Should we make this a partial for reuse?

src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
ranbel: This file is already in Step 3 below :) https://github.com/cloudflare/cloudflare-docs/pull/17243/files#diff-28fa65d3a5179b85d5746360aea83b05ae9ea7d44e205e4b5f06fb790893d901R33

src/content/partials/cloudflare-one/ssh/ssh-proxy-ca.mdx
ranbel: yes, that's a good idea

src/content/partials/cloudflare-one/access/add-target.mdx
```diff
@@ -13,14 +13,14 @@ To create a new target:
...
 	<Details header="Format restrictions">
 			- Case insensitive
 			- Contain no more than 255 characters
 			- Contain only alphanumeric characters, `-`, or `.` (no spaces allowed)
 			- Start and end with an alphanumeric character
 	</Details>
-4. In **IP addresses**, enter the private IPv4 and/or IPv6 address of the target resource. If the IP address overlaps across multiple private networks, select the [virtual network](/cloudflare-one/connections/connect-networks/private-net/cloudflared/tunnel-virtual-networks/) where the resource is located.
+4. In **IP addresses**, enter the private IPv4 and/or IPv6 address of the target resource. If the IP address overlaps across multiple private networks, select the [virtual network](/cloudflare-one/connections/connect-networks/private-net/cloudflared/tunnel-virtual-networks/) where the resource is located. Targets may not point to the same IP address and virtual network pairings.
```
ranbel: Is this just saying that you can't duplicate targets? What are users trying to accomplish when they try this configuration?

asamborski: Yes, users can't duplicate targets. This could happen if a user wanted to assign a different hostname to the same address to differentiate via policy or in the future, if they wanted to define a target per protocol. Neither of these cases are supported by design, which is why I thought we may want to call out no support for duplicates upfront.

src/content/docs/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access.mdx
asamborski: Oh whoops, I don't know how I missed that! Will remove!

ranbel (approved):

PR #17252, opened 2024-10-01
[Workers Assets] fix and clarify `not_found_handling` documentation

emily-shen:
### Summary

1. Fix example and description for the assets binding (see this [discord comment](https://discord.com/channels/595317990191398933/779390076219686943/1290268605342289941) for more information).
2. Clarify that you need to use the asset binding to invoke 'not_found_handling` from a user worker, also based on community feedback.


src/content/docs/workers/static-assets/binding.mdx
```diff
@@ -61,7 +61,7 @@ In the example above, assets would be available through `env.ASSETS`.
 
 Your dynamic code can make new, or forward incoming, requests to your project's static assets using the assets binding.
 
-Take the following example that configures a Worker to return a response under all requests headed for `/api/`. Otherwise, the Worker will fallback to returning static assets.
+Take the following example that configures a Worker to return a response under all requests headed for `/api/`. It then uses the asset binding to invoke behaviour set by [`not_found_handling`](/workers/static-assets/routing/#routing-configuration) to catch requests that do not match.
```
emily-shen: We try serving assets first, so this is not 'falling back' exactly - the binding is used for not_found_handling only.

src/content/docs/workers/static-assets/routing.mdx
```diff
@@ -25,7 +25,7 @@ In this example, request to `example.com/blog` serves the `blog.html` file.
 
 ![A request to `example.com/blog` serves the `/blog.html` file. ](~/assets/images/workers/platform/assets/workers-assets-serve-asset.png)
 
-2. If there are no assets that match the current route requested then the Worker will be invoked. If there is no Worker, [`not_found_handling`](/workers/static-assets/routing/#2-not_found_handling) will be evaluated. By default, a null-body 404-status response will be served.
+2. If there are no assets that match the current route requested then the Worker will be invoked. If there is no Worker, [`not_found_handling`](/workers/static-assets/routing/#2-not_found_handling) will be evaluated. By default, a null-body 404-status response will be served. To invoke `not_found_handling` behavior _with_ a Worker, you must use the [asset binding](/workers/static-assets/binding/#runtime-api-reference).
```
GregBrimble:
```suggestion
2. If there is no Worker script, [`not_found_handling`](/workers/static-assets/routing/#2-not_found_handling) will be evaluated. By default, a null-body 404-status response will be served.

If a Worker is configured, and there are no assets that match the current route requested, the Worker will be invoked. The Worker can then "fall back" to `not_found_handling` asset behavior, by passing the incoming request through to the [asset binding](/workers/static-assets/binding/#runtime-api-reference).
```

src/content/docs/workers/static-assets/routing.mdx
```diff
@@ -43,7 +43,7 @@ Forcing or dropping trailing slashes on request paths (e.g. `example.com/page/`
 
 #### 2. `not_found_handling`
 
-In the event a request does not match a static asset, and there is no Worker script (or that Worker script calls `fetch()` on [the assets binding](/workers/static-assets/binding/)), `not_found_handling` determines how Cloudflare will construct a response.
+`not_found_handling` determines how Cloudflare will construct a response if a request does not match a static asset, and there is no Worker script. If you do have a Worker script, this behavior **needs to be manually invoked** using the [assets binding](/workers/static-assets/binding/#runtime-api-reference).
```
GregBrimble: I personally don't like this change. I prefer what's there before. Users don't "invoke the behavior", and nor does "need to be invoked". It's just a thing that happens if they choose to do `fetch()`.

```suggestion
In the event a request does not match a static asset, and there is no Worker script (or that Worker script calls `fetch()` on [the assets binding](/workers/static-assets/binding/)), `not_found_handling` determines how Cloudflare will construct a response.
```

src/content/docs/workers/static-assets/binding.mdx
```diff
@@ -73,8 +73,7 @@ export default {
 			// TODO: Add your custom /api/* logic here.
 			return new Response("Ok");
 		}
-		// Otherwise, serve the static assets.
-		// Without this, the Worker will error and no assets will be served.
+		// The asset binding invokes the behavior set by `not_found_handling`.
```
GregBrimble:
```suggestion
		// Because the incoming request is being passed through to the assets binding, and because no asset matched this request, this will evaluate `not_found_handling` behavior.
```

src/content/docs/workers/static-assets/binding.mdx
GregBrimble:
```suggestion
Take the following example that configures a Worker script to return a response under all requests headed for `/api/`. Otherwise, the Worker script will fall back to serving assets. In this case, the incoming request is known to have already not matched against any static asset, so [`not_found_handling`](/workers/static-assets/routing/#routing-configuration) behavior will be evaluated.
```

src/content/docs/workers/static-assets/routing.mdx
emily-shen: fair enough - i saw some people confused why not_found_handling behaviour wasn't 'just working' with a user woker, but with the additional clarification elsewhere I'm satisfied :)

src/content/docs/workers/static-assets/binding.mdx
emily-shen: just changed to "This will evaluate `not_found_handling` behavior", would prefer to keep the script comments shorter.

src/content/docs/workers/static-assets/binding.mdx
emily-shen: tweaked a bit to make it super duper explicit about what's happening

GregBrimble (approved):

Issue #17011, opened 2024-09-23
[Platform RFC] Run `prettier --write` on merge


GregBrimble:
A couple of assumptions:

1. When making an edit, starting from a consistently formatted page makes creating that edit easier.
2. Not everyone is making these changes in an IDE with Prettier installed (e.g. GitHub online directly)
3. We don't want to block edits which don't conform to formatting rules. We do want to allow them to be made without any additional friction.
4. We don't love reviewing PRs which come from contributors which have an auto-formatter installed on their IDE, so inadvertently pull in unrelated formatting changes with their PRs.
5. Consistent formatting is useful to both contributors, and to our readers (esp. for code snippets).

I propose making a (non-trivial) effort to run `prettier --write` across the codebase, and then, in order to maintain that consistency going forward, run `prettier --write` again on every merge atop the contributor's content commits. Couple of ways we could technically do this, and I don't really mind about the specifics. Just wanted to start the conversation here about actually trying to uphold some consistency here.

Cherry:
Huge +1 to this from me (as per some of my recent PRs), and relates partially to https://github.com/cloudflare/cloudflare-docs/issues/7798, and pushes way more in that direction.

https://autofix.ci/ is a great modern solution for making this as frictionless as possible for contributors.

Varecha:
![Screenshot_20240801-204348](https://github.com/user-attachments/assets/c1e6777e-f905-48c5-bc2e-c29ab824144a)

kodster28:
I think we'll need to close this one out as partially completed.

We do enforce prettier for `.astro`, `.ts` and other dev-type files.

Content type files are too fraught to go with here, b/c they cause sporadic failures in the build that are often a mystery to contributors.

PR #17262, opened 2024-10-01
[Docs Site] core formatting checks

Cherry:
### Summary

- Formats remaining files left over from previous formatting ones
- Adds a "core" formatting check for JS/TS, etc. that's part of the docs site

This also adds some scripts for formatting other content, but none of that runs by default yet and will likely need to wait for https://github.com/cloudflare/cloudflare-docs/issues/17011


ghost (commented):
Great

KianNH (approved):

package.json
```diff
@@ -12,7 +12,11 @@
 		"check:functions": "npx tsc --noEmit -p ./functions/tsconfig.json",
 		"check:worker": "npx tsc --noEmit -p ./worker/tsconfig.json",
 		"dev": "npx astro dev",
-		"format": "npx prettier --write \"**/*.{js,jsx,ts,tsx,mjs,astro,css,json,yaml,yml}\"",
+		"format": "npm run format:core && npm run format:content && npm run format:data",
```
KianNH:
```suggestion
		"format": "npm run format:core && npm run format:data",
```

KianNH: We're still discussing if we want to use Prettier (or alternative formatters) for MD/MDX at the moment so I think we can remove this (for users of npm run format) but keep the format:content check around for our testing.

KianNH (approved):

package.json
Cherry: Sure, sounds good!

PR #17264, opened 2024-10-01
SQLite in DO clarifications:

vy-ton:
- deleteAll() behavior
- multi-statement queries with sql.exec

### DIFF
--- a/src/content/docs/durable-objects/api/storage-api.mdx
+++ b/src/content/docs/durable-objects/api/storage-api.mdx
@@ -100,7 +100,9 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
 * <code>deleteAll(options <Type text='Object' /> <MetaInfo text='optional' />)</code>: <Type text='Promise' />
 
-  * Deletes all keys and associated values, effectively deallocating all storage used by the Durable Object. In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted. `deleteAll()` does not proactively delete [Alarms](/durable-objects/api/alarms/). Use [`deleteAlarm()`](/durable-objects/api/alarms/#deletealarm) to delete an alarm.
+  * Deletes all stored data, effectively deallocating all storage used by the Durable Object. For Durable Objects with a key-value storage backend, `deleteAll()` removes all keys and associated values for an individual Durable Object. For Durable Objects with a [SQLite storage backend](/durable-objects/best-practices/access-durable-objects-storage/#sqlite-storage-backend), `deleteAll()` removes the entire contents of a Durable Object's private SQLite database.  
+  * In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted. 
+  * `deleteAll()` does not proactively delete [Alarms](/durable-objects/api/alarms/). Use [`deleteAlarm()`](/durable-objects/api/alarms/#deletealarm) to delete an alarm.
 
 #### Supported options
 
@@ -245,7 +247,7 @@ SQL API methods accessed with `ctx.storage.sql` are only allowed on [Durable Obj
 
 #### Parameters
 * `query`: <Type text ='string' />
-  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings.
+  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings. Multiple SQL statements, separated with a semicolen, can be executed in the `query`; any parameter bindings are applied to the last SQL statement in the `query`.
 * `bindings`: <Type text='any[]' /> <MetaInfo text='Optional' />
   * Optional variable number of arguments that correspond to the `?` placeholders in `query`.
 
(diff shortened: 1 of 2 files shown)


src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -100,7 +100,9 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
 * <code>deleteAll(options <Type text='Object' /> <MetaInfo text='optional' />)</code>: <Type text='Promise' />
 
-  * Deletes all keys and associated values, effectively deallocating all storage used by the Durable Object. In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted. `deleteAll()` does not proactively delete [Alarms](/durable-objects/api/alarms/). Use [`deleteAlarm()`](/durable-objects/api/alarms/#deletealarm) to delete an alarm.
+  * Deletes all stored data, effectively deallocating all storage used by the Durable Object. For Durable Objects with a key-value storage backend, `deleteAll()` removes all keys and associated values for an individual Durable Object. For Durable Objects with a [SQLite storage backend](/durable-objects/best-practices/access-durable-objects-storage/#sqlite-storage-backend), `deleteAll()` removes the entire contents of a Durable Object's private SQLite database.  
+  * In the event of a failure while the `deleteAll()` operation is still in flight, it may be that only a subset of the data is properly deleted.
```
kentonv: This is actually no longer true with the sqlite backend. Either everything is deleted or nothing is.

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -245,7 +247,7 @@ SQL API methods accessed with `ctx.storage.sql` are only allowed on [Durable Obj
 
 #### Parameters
 * `query`: <Type text ='string' />
-  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings.
+  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings. Multiple SQL statements, separated with a semicolen, can be executed in the `query`; any parameter bindings are applied to the last SQL statement in the `query`.
```
kentonv:
```suggestion
  * The SQL query string to be executed. `query` can contain `?` placeholders for parameter bindings. Multiple SQL statements, separated with a semicolon, can be executed in the `query`; any parameter bindings are applied to the last SQL statement in the `query`.
```

kentonv: The returned cursor is also only for the last statement in the query.

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -41,6 +41,9 @@ export class Counter extends DurableObject {
   }
 }
 ```
+### Removing a Durable Object's storage
+
+To effectively "delete" a Durable Object, you should first use [`deleteAll()`](/durable-objects/api/storage-api/#deleteall) to remove any attached storage for a Durable Object. Calling `deleteAll()` ensures that a Durable Object will not be [billed for Storage API usage](/durable-objects/platform/pricing/#storage-api-billing). Once storage is deleted and a Durable Object is no longer active due to incoming requests, then the Durable Object is effectively deleted.
```
kentonv:
```suggestion
A Durable Object fully ceases to exist if, when it shuts down, its storage is empty. One a DO ceases to exist, it no longer incurs any billing for storage. If you never write to a Durable Object's storage at all (including setting alarms), then storage remains empty, and so the DO will no longer exist once it shuts down. But if you ever write to storage, then you must explicitly call [`storage.deleteAll()`](/durable-objects/api/storage-api/#deleteall) to cause storage to become empty again. Note that it is not sufficient to simply delete the specific data that you wrote (e.g. deleting a key or dropping a table), as some metadata may remain. The only way to remove all traces of storage is to call `deleteAll()`.
```

vy-ton: worked in your suggestion with minor edits

Oxyjun (approved):

src/content/docs/durable-objects/best-practices/access-durable-objects-storage.mdx
```diff
@@ -41,6 +41,11 @@ export class Counter extends DurableObject {
   }
 }
 ```
+### Removing a Durable Object's storage
+
+A Durable Object fully ceases to exist if, when it shuts down, its storage is empty. If you never write to a Durable Object's storage at all (including setting alarms), then storage remains empty, and so the Durable Object will no longer exist once it shuts down.
+
+However if you ever write using [Storage API](/durable-objects/api/storage-api/), including setting alarms, then you must explicitly call [`storage.deleteAll()`](/durable-objects/api/storage-api/#deleteall) to empty storage. It is not sufficient to simply delete the specific data that you wrote, such as deleting a key or dropping a table, as some metadata may remain. The only way to remove all storage is to call `deleteAll()`. Calling `deleteAll()` ensures that a Durable Object that ceases to exist will not be billed for storage.
```
kentonv:
```suggestion
However if you ever write using [Storage API](/durable-objects/api/storage-api/), including setting alarms, then you must explicitly call [`storage.deleteAll()`](/durable-objects/api/storage-api/#deleteall) to empty storage. It is not sufficient to simply delete the specific data that you wrote, such as deleting a key or dropping a table, as some metadata may remain. The only way to remove all storage is to call `deleteAll()`. Calling `deleteAll()` ensures that a Durable Object will not be billed for storage.
```

"that ceases to exist" as a qualifier here feels circular since `deleteAll()` itself is what allows/causes the durable object to cease to exist.

kentonv (approved):

PR #17286, opened 2024-10-02
[Docs Site] Render Mermaid diagrams at build-time

KianNH:
### Summary

Render Mermaid diagrams at build-time.


KianNH:
Replaced by https://github.com/cloudflare/cloudflare-docs/pull/17287 due to CI instability & tracking down the true root cause.

PR #17288, opened 2024-10-02
[Images] Upload an image via a Worker

dcpena:
Fixed an incorrect code example and added a new page for uploading an image with a worker to address PCX-13703.

### DIFF
--- a/src/components/models/code/TextToImageCode.astro
+++ b/src/components/models/code/TextToImageCode.astro
@@ -23,14 +23,14 @@ export default {
       prompt: "cyberpunk cat",
     };
 
-    const response = await env.AI.run(
+    const stream = await env.AI.run(
       "${name}",
       inputs
     );
 
     return new Response(response, {
       headers: {
-        "content-type": "image/png",
+        "content-type": "image/jpg",
       },
     });
   },
--- a/src/content/docs/images/upload-images/upload-file-worker.mdx
+++ b/src/content/docs/images/upload-images/upload-file-worker.mdx
@@ -29,7 +29,7 @@ const response = await fetch(API_URL, {
 
 ## Upload from AI generated images
 
-To upload an AI generated image, refer to the example below.
+You can use an AI Worker to generate an image and then upload that image to store it in Cloudflare Images. For more information about using Workers AI to generate an image, refer to the [SDXL-Lightning Model](https://developers.cloudflare.com/workers-ai/models/stable-diffusion-xl-lightning).
 
 ```ts
 const API_URL = "https://api.cloudflare.com/client/v4/accounts/<ACCOUNT_ID>/images/v1";
--- a/src/content/docs/images/upload-images/upload-file-worker.mdx
+++ b/src/content/docs/images/upload-images/upload-file-worker.mdx
@@ -29,7 +29,7 @@ const response = await fetch(API_URL, {
 
 ## Upload an AI generated image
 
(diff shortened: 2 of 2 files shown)


src/content/docs/images/upload-images/upload-file-worker.mdx
```diff
@@ -0,0 +1,15 @@
...
+
+You can use a Worker to upload your image to Cloudflare Images.
+
+Refer to the example below or refer to the [Workers documentation](/workers/) for more information.
+
+```ts
+const image = await fetch("./myimage.png");
+const bytes = await image.bytes();
+formData.append('file', new File([bytes], 'myimage.png');
+```
```
dochne: We should extend this so it's a bit less of a stub!  This guides them through the weird bit, but an example that they can copy/paste the majority of is always preferable!

```ts
const API_URL = "https://api.cloudflare.com/client/v4/accounts/<ACCOUNT_ID>/images/v1";
const TOKEN = "YOUR_TOKEN_HERE";

const image = await fetch("https://example.com/image.png");
const bytes = await image.bytes();

const formData = new FormData();
formData.append('file', new File([bytes], 'image.png'));

const response = await fetch(API_URL, {
  method: 'POST',
  headers: {
    'Authorization': `Bearer ${TOKEN}`,
  },
  body: formData,
});
```

dochne:
I think it's worth us adding an example here phrased as "Uploading from AI Generated Images" with this code in it:
```ts
const API_URL = "https://api.cloudflare.com/client/v4/accounts/<ACCOUNT_ID>/images/v1";
const TOKEN = "YOUR_TOKEN_HERE";

const stream = await env.AI.run(
  "@cf/bytedance/stable-diffusion-xl-lightning",
  {
    prompt: YOUR_PROMPT_HERE
  }
);
const bytes = await (new Response(stream)).bytes();

const formData = new FormData();
formData.append('file', new File([bytes], 'image.jpg');

const response = await fetch(API_URL, {
  method: 'POST',
  headers: {
    'Authorization': `Bearer ${TOKEN}`,
  },
  body: formData,
});
```

src/content/docs/images/upload-images/upload-file-worker.mdx
```diff
@@ -0,0 +1,15 @@
+---
+pcx_content_type: how-to
+title: Upload an image using a Worker
```
dochne: image

Looking at the titles of our current options, "Upload via Worker" seems like it'd fit better

src/content/docs/images/upload-images/upload-file-worker.mdx
dcpena: Thanks so much for this! I definitely agree that an example that walks a user through _and_ is something they can copy/paste is preferable! Updated

src/content/docs/images/upload-images/upload-file-worker.mdx
dcpena: Done

src/components/models/code/TextToImageCode.astro
```diff
@@ -23,14 +23,14 @@ export default {
       prompt: "cyberpunk cat",
     };
 
-    const response = await env.AI.run(
+    const stream = await env.AI.run(
       "${name}",
       inputs
     );
 
     return new Response(response, {
```
pedrosousa, line 31: Should we also change `response` here to `stream`?

pedrosousa (approved):

PR #17298, opened 2024-10-03
Update dns-in-google-sheets.mdx to add caching in NSLookup

beltofte:
Adding internal caching of the results on the NSLookup function to limit the number of the DNS resolver requests and speed up the results - especially in larger Google Sheets.

### Summary

The NSLookup function for Google Sheets provided in the original documentation page does not support internal caching of the results. This can lead to issues in Google Sheets with many calls to the function - this includes error 400 from the https://cloudflare-dns.com/dns-query if you use it too much. 

It is possible to disable caching and control the cache TTL via arguments in the NSLookup function.

### Documentation checklist


src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
```diff
@@ -68,6 +89,10 @@ function NSLookup(type, domain) {
 
   var outputString = outputData.join(',');
 
+  if (usecache == true) {
+    cache.put(cacheBinKey, outputString, cachettl);
+  }
```
hunts, line 98: The `Answer` records have TTLs. If we use `cacheTTL = 1800` as the cap, we should adjust the effective TTL to respect the TTL of the record set.

E.g

```
  for (var i in response.Answer) {
    outputData.push(response.Answer[i].data);
    cacheTTL = Math.min(cacheTTL, response.Answer[i].TTL);
  }
```

src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
```diff
@@ -22,10 +22,31 @@ function NSLookup(type, domain) {
...
+
+  if (typeof cachettl != "number") {
+    throw new Error('Only numeric values allowed in 4 cache ttl');
+  }
+
   type = type.toUpperCase();
+  domain = domain.toLowerCase();
+
+  if (usecache == true) {
```
hunts: Can be `if (usecache)` as `usecache` is validated to be boolean.

src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
hannes-cf: The default should be to use the `Answer` record's TTL. So let's add your suggestion.

Also renaming `cacheTTL` to `minCacheTTL` to make it clear.

src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
```diff
@@ -12,7 +12,7 @@ import { Details } from "~/components"
 1.1.1.1 works directly inside Google Sheets. To get started, create a [Google Function](https://developers.google.com/apps-script/guides/sheets/functions) with the following code:
 
 ```js
-function NSLookup(type, domain) {
+function NSLookup(type, domain, usecache = true, cachettl = 1800) {
```
hannes-cf: The default value for `usecache` should be false, since correctness should be preferred over performance. Users who want to change that are free to do so.

src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
```diff
@@ -89,10 +89,15 @@ function NSLookup(type, domain, usecache = true, cachettl = 1800) {
...
-  if (usecache == true) {
+  if (usecache) {
     cache.put(cacheBinKey, outputString, cachettl);
   }
-  
+
+  for (var i in response.Answer) {
+    outputData.push(response.Answer[i].data);
+    minCacheTTL = Math.min(minCacheTTL, response.Answer[i].TTL);
+  }
```
hunts, line 98: It seems this is in the wrong place. What it should do is to replace the for loop starts at line #86.

Also if we use `cachettl` as the "minimal" TTL, the math supposed to use `Math.max()` instead of `Math.min()` now, e.g.

```
cachettl = Math.max(cachettl, response.Answer[i].TTL);
```

In that case, the the default value I think can be much lower than 1800, e.g. use 30 seconds instead.

beltofte: Implemented the changes in the latest commit. Including removing the extra for loop and moving minCacheTTL to loop starting at line #86. Also changed the default value of cachettl from 1800 to 30 seconds.

hunts:
Sorry for the confusing, I think our TTL calculation is still not right. This would be the changes I'd like to make before we can merge it.

Basically, use the minimal TTL from records, but clamp the lower bound at the minCacheTTL (which is 30 seconds). Also replaced `var` with `const` /`let` keywords by the way.

```
index df02454ac..480cd4be6 100644
--- a/src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
+++ b/src/content/docs/1.1.1.1/other-ways-to-use-1.1.1.1/dns-in-google-sheets.mdx
@@ -12,7 +12,7 @@ import { Details } from "~/components"
 1.1.1.1 works directly inside Google Sheets. To get started, create a [Google Function](https://developers.google.com/apps-script/guides/sheets/functions) with the following code:
 
 ```js
-function NSLookup(type, domain, usecache = false, cachettl = 30) {
+function NSLookup(type, domain, usecache = false, minCacheTTL = 30) {
 
   if (typeof type == 'undefined') {
     throw new Error('Missing parameter 1 dns type');
@@ -26,43 +26,44 @@ function NSLookup(type, domain, usecache = false, cachettl = 30) {
     throw new Error('Only boolean values allowed in 3 use cache');
   }
 
-  if (typeof cachettl != "number") {
-    throw new Error('Only numeric values allowed in 4 cache ttl');
+  if (typeof minCacheTTL != "number") {
+    throw new Error('Only numeric values allowed in 4 min cache ttl');
   }
 
   type = type.toUpperCase();
   domain = domain.toLowerCase();
 
+  let cache = null;
   if (usecache) {
     // Cache key and hash
     cacheKey = domain + "@" + type;
     cacheHash = Utilities.base64Encode(cacheKey);
     cacheBinKey = "nslookup-result-" + cacheHash;
 
-    var cache = CacheService.getScriptCache();
-    var cachedResult = cache.get(cacheBinKey);
+    cache = CacheService.getScriptCache();
+    const cachedResult = cache.get(cacheBinKey);
     if (cachedResult != null) {
       return cachedResult;
     }
   }
 
-  var url = 'https://cloudflare-dns.com/dns-query?name=' + encodeURIComponent(domain) + '&type=' + encodeURIComponent(type);
-  var options = {
+  const url = 'https://cloudflare-dns.com/dns-query?name=' + encodeURIComponent(domain) + '&type=' + encodeURIComponent(type);
+  const options = {
     muteHttpExceptions: true,
     headers: {
       accept: "application/dns-json"
     }
   };
 
-  var result = UrlFetchApp.fetch(url, options);
-  var rc = result.getResponseCode();
-  var resultText = result.getContentText();
+  const result = UrlFetchApp.fetch(url, options);
+  const rc = result.getResponseCode();
+  const resultText = result.getContentText();
 
   if (rc !== 200) {
     throw new Error(rc);
   }
 
-  var errors = [
+  const errors = [
     { name: "NoError", description: "No Error"}, // 0
     { name: "FormErr", description: "Format Error"}, // 1
     { name: "ServFail", description: "Server Failure"}, // 2
@@ -75,24 +76,25 @@ function NSLookup(type, domain, usecache = false, cachettl = 30) {
     { name: "NotAuth", description: "Not Authorized"} // 9
   ];
 
-  var response = JSON.parse(resultText);
+  const response = JSON.parse(resultText);
 
   if (response.Status !== 0) {
     return errors[response.Status].name;
   }
 
-  var outputData = [];
-  var minCacheTTL = cachettl;
+  const outputData = [];
+  let cacheTTL = 0;
 
-  for (var i in response.Answer) {
+  for (const i in response.Answer) {
     outputData.push(response.Answer[i].data);
-    minCacheTTL = Math.max(minCacheTTL, response.Answer[i].TTL);
+    const ttl = response.Answer[i].TTL;
+    cacheTTL = Math.min(cacheTTL || ttl, ttl);
   }
 
-  var outputString = outputData.join(',');
+  const outputString = outputData.join(',');
 
   if (usecache) {
-    cache.put(cacheBinKey, outputString, minCacheTTL);
+    cache.put(cacheBinKey, outputString, Math.max(cacheTTL, minCacheTTL));
   }
 
   return outputString;
```

beltofte:
Thanks @hunts. I have committed the patch, and at the same time renamed usecache to useCache, so all variables / constants are following lowerCamelCase.

hunts (approved):
Thanks for the quick updates. Looks good to me now.

RebeccaTamachiro (approved):

PR #17312, opened 2024-10-04
[Docs Site] Re-render Mermaid diagrams when theme changes

KianNH:
### Summary

Re-render Mermaid diagrams when theme changes.

### DIFF
--- a/src/components/overrides/Head.astro
+++ b/src/components/overrides/Head.astro
@@ -175,7 +175,45 @@ if (Astro.props.entry.data.external_link) {
 <script>
 	import mermaid from "mermaid";
 
-	mermaid.initialize({ startOnLoad: true, theme: "neutral" });
+	const diagrams = document.querySelectorAll<HTMLPreElement>("pre.mermaid");
+
+	diagrams.forEach((diagram) =>
+		diagram.setAttribute("data-diagram", diagram.textContent as string),
+	);
+
+	async function render() {
+		if (diagrams) {
+			const theme =
+				document.documentElement.getAttribute("data-theme") === "light"
+					? "neutral"
+					: "dark";
+
+			for (const diagram of diagrams) {
+				const def = diagram.getAttribute("data-diagram") as string;
+
+				mermaid.initialize({ startOnLoad: false, theme });
+				await mermaid
+					.render(`mermaid-${crypto.randomUUID()}`, def)
+					.then(({ svg }) => (diagram.innerHTML = svg));
+
+				diagram.setAttribute("data-processed", "true");
+
+				const { height, width } = diagram.getBoundingClientRect();
+
+				diagram.setAttribute(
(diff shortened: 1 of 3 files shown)


src/components/overrides/Head.astro
```diff
@@ -175,7 +175,38 @@ if (Astro.props.entry.data.external_link) {
 <script>
 	import mermaid from "mermaid";
 
-	mermaid.initialize({ startOnLoad: true, theme: "neutral" });
+	const diagrams = document.querySelectorAll<HTMLPreElement>("pre.mermaid");
+
+	diagrams.forEach((diagram) =>
+		diagram.setAttribute("data-diagram", diagram.textContent as string),
```
KianNH: We have to store the original Mermaid diagram definition otherwise we'd lose it and be unable to re-render the diagram.

src/components/overrides/Head.astro
```diff
@@ -175,7 +175,38 @@ if (Astro.props.entry.data.external_link) {
...
+					? "neutral"
+					: "dark";
+
+			for (const diagram of diagrams) {
+				const def = diagram.getAttribute("data-diagram") as string;
+
+				mermaid.initialize({ startOnLoad: false, theme });
+				await mermaid
+					.render(`mermaid-${crypto.randomUUID()}`, def)
+					.then(({ svg }) => (diagram.innerHTML = svg));
```
KianNH: `render` instead of `run` is an intentional choice, we'd like to wait until we've got a new SVG before replacing the current one to prevent a small flicker of having no diagram.

src/components/overrides/Head.astro
```diff
@@ -175,7 +175,38 @@ if (Astro.props.entry.data.external_link) {
...
+
+				diagram.setAttribute("data-processed", "true");
```
KianNH: Mermaid would add this on it's own when using `initialize` or `run`.

GregBrimble (approved):
Helpful comments! Not sure the `data-processed` is needed — none of the other examples on mermaid.js.org have it — but whatevs :)

KianNH:
> Helpful comments! Not sure the `data-processed` is needed — none of the other examples on mermaid.js.org have it — but whatevs :)

It's normally only relevant when using [the `initialize`/`run` functions](https://github.com/mermaid-js/mermaid/blob/6e0fe035525f1c033d58161f78a1ffe7c9e17f8b/packages/mermaid/src/mermaid.ts#L85-L104), but we apply some CSS to `pre.mermaid:not([data-processed])` for FOUC/noscript purposes.

image

There's still some layout shift, since the size of the definiton in the `pre` isn't representative of the rendered diagram, but it's a little better than nothing.

I'd prefer server-side rendering but that requires Playwright & makes builds pretty flaky (OOM and/or timeouts).

PR #17350, opened 2024-10-06
[Docs Site] chore: optimize SVGs

Cherry:
This PR optimizes all of the SVGs in the repo currently, saving a total of ~33.12MB across all of the docs!

Before: ~68.44 MB of total SVGs
After: ~35.33 MB of total SVGs

Astro's image service doesn't process SVGs today, so it just passes them through entirely unoptimized. Many of the SVGs in this repo are exported from Illustrator and the likes which massively inflates their sizes, sometimes into the megabytes. Some pages with multiple SVGs can end up loading over a MB of SVG content which not only takes longer at a network level, but also longer for the browser to parse and then render.

I originally made this an Astro integration plugin, but it was taking a minute or two _per SVG_ on some of the more large and complex ones, so adding tens of minutes to the build every time didn't make sense. I've instead moved it to a generic script in a new `scripts` folder. The long-term goal is to run these automatically and then push back to the repo any time a new SVG file is added, but for now, they can just be run manually every so often.

It might make more sense to build these optimised versions to another folder, and then somehow tweak Astro/rehype's file loading to pull from there first if exists, before the raw sources, I'm not sure.


securitypedant:
Interesting work. All the reference architecture SVGs are exported from Figma, not Illustrator. Some of our docs have 15+ SVGs in them. Is there a process we can add into our release of new documents to optimize our images?

Cherry:
I'm not sure if Figma has a more web-optimised export honestly - that's something worth looking into.

But once this lands, ideally eventually in PRs the optimisation will run automatically and transparently to contributors (out of scope for this PR, and probably something more suited in addition to #17011), but in the meantime, if folks will just run `npm run script:optimize-svgs` locally, it'll optimise all of the SVGs for faster downloads/parsing on the web.

Or, someone can do a fly-by pass optimisation PR every so often to crunch any SVGs that didn't get optimised.

GregBrimble (approved):
Thank you for your contribution! This'll save our end-users a lot of network bytes!

KianNH (approved):

PR #17364, opened 2024-10-07
Clarifiy that allow_custom_ports flag affects WebSockets

rnbrady:
Clarify that creating a WebSocket client with a call to `new WebSocket(url)` will also obey the `allow_custom_ports` or `ignore_custom_ports` flags, which I have verified through testing.

I haven't created an issue as this is not a "significant change" as described in CONTRIBUTING.md.


KianNH (approved):
Thanks for adding this!

PR #17403, opened 2024-10-08
[Homepage] Replace DevPlat cards with video

KianNH:
### Summary

Replace DevPlat cards with video

### Screenshots (optional)

Before:
image

After:
image


src/components/homepage/FeaturedContentSection.astro
```diff
@@ -47,17 +48,35 @@ const { title, text, image, cards, imagePosition } = props.parse(Astro.props);
...
 		{
-			cards.map((card) => (
-				<Card title={card.title}>
-					<span set:html={marked.parse(card.text)} />
-					<br />
-					<LinkCard title={card.cta.title} href={card.cta.href} />
-				</Card>
-			))
+			title === "Developer Platform" ? (
+				<div class="[&>article]:gap-0">
```
KianNH, line 55: We don't want a title on this card, and we need to override the gap because of the now-empty `<p>` tag still causing it to be calculated.

KimJ15 (approved):
Chatted with @abstergail and i'm fine to add this now. She has a few recommendations, then we can merge.

Oxyjun (approved):

pedrosousa (approved):

PR #17411, opened 2024-10-08
Update reference documentation for Worker accessible DO interfaces

joshthoward:
### Documentation checklist

- [ ] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has be
en adhered to.

Screenshot 2024-10-08 at 1 46 03 PM
Screenshot 2024-10-08 at 1 45 56 PM

### DIFF
--- a/src/content/docs/durable-objects/api/alarms.mdx
+++ b/src/content/docs/durable-objects/api/alarms.mdx
@@ -2,7 +2,7 @@
 title: Alarms
 pcx_content_type: concept
 sidebar:
-  order: 4
+  order: 6
 ---
 
 import { Type } from "~/components";
@@ -37,16 +37,13 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 
 ### getAlarm
 
-- `getAlarm()`: <Type text ='number | null' />
+- <code>getAlarm()</code>: <Type text ='number | null' />
 
   - If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
 
 ### setAlarm
 
-- <code>
-  	setAlarm(scheduledTimeMs <Type text="number" />)
-  </code>
-  : <Type text="void" />
+- <code> setAlarm(scheduledTimeMs <Type text="number" />)</code>: <Type text="void" />
 
   - Set the time for the alarm to run at in number of milliseconds elapsed since the UNIX epoch.
 
--- a/src/content/docs/durable-objects/api/alarms.mdx
+++ b/src/content/docs/durable-objects/api/alarms.mdx
@@ -21,7 +21,7 @@ Notably:
 
 - Each Durable Object instance is able to schedule a single alarm at a time by calling `setAlarm()`.
 - Alarms have guaranteed at-least-once execution and are retried automatically when the `alarm()` handler throws.
-- Retries are performed using exponential backoff starting at a two second delay from the first failure with up to six retries allowed.
+- Retries are performed using exponential backoff starting at a 2 second delay from the first failure with up to 6 retries allowed.
 
 :::note[How are alarms different from Cron Triggers?]
 
@@ -63,7 +63,7 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 
   - Called by the system when a scheduled alarm time is reached.
 
-  - The `alarm()` handler has guaranteed at-least-once execution and will be retried upon failure using exponential backoff, starting at two second delays for up to six retries. Retries will be performed if the method fails with an uncaught exception.
+  - The `alarm()` handler has guaranteed at-least-once execution and will be retried upon failure using exponential backoff, starting at 2 second delays for up to 6 retries. Retries will be performed if the method fails with an uncaught exception.
 
   - This method can be `async`.
 
--- a/src/content/docs/durable-objects/api/alarms.mdx
+++ b/src/content/docs/durable-objects/api/alarms.mdx
@@ -2,8 +2,7 @@
 title: Alarms
 pcx_content_type: concept
 sidebar:
-  order: 1
-
+  order: 4
 ---
 
 import { Type } from "~/components";
@@ -20,17 +19,15 @@ Durable Objects alarms allow you to schedule the Durable Object to be woken up a
 
 Notably:
 
-* Each Durable Object instance is able to schedule a single alarm at a time by calling `setAlarm()`.
-* Alarms have guaranteed at-least-once execution and are retried automatically when the `alarm()` handler throws.
-* Retries are performed using exponential backoff starting at a two second delay from the first failure with up to six retries allowed.
+- Each Durable Object instance is able to schedule a single alarm at a time by calling `setAlarm()`.
+- Alarms have guaranteed at-least-once execution and are retried automatically when the `alarm()` handler throws.
+- Retries are performed using exponential backoff starting at a two second delay from the first failure with up to six retries allowed.
 
 :::note[How are alarms different from Cron Triggers?]
-
 
 Alarms are more fine grained than [Cron Triggers](/workers/configuration/cron-triggers/#cron-triggers). A Worker can have up to three Cron Triggers configured at once, but it can have an unlimited amount of Durable Objects, each of which can have an alarm set.
 
 Alarms are directly scheduled from within your Durable Object. Cron Triggers, on the other hand, are not programmatic. [Cron Triggers](/workers/configuration/cron-triggers/#cron-triggers) execute based on their schedules, which have to be configured through the Cloudflare dashboard or API.
-
 
 :::
 
@@ -40,94 +37,81 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 
 ### getAlarm
 
+- `getAlarm()`: <Type text ='number | null' />
 
-
-* `getAlarm()`: <Type text ='number | null' />
-
-  * If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
-
-
+  - If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
 
 ### setAlarm
 
+- <code>
+  	setAlarm(scheduledTimeMs <Type text="number" />)
+  </code>
+  : <Type text="void" />
 
-
-* <code>setAlarm(scheduledTimeMs <Type text='number'/>)</code>: <Type text='void' />
-
-  * Set the time for the alarm to run at in number of milliseconds elapsed since the UNIX epoch.
-
-
+  - Set the time for the alarm to run at in number of milliseconds elapsed since the UNIX epoch.
 
 ### deleteAlarm
 
+- `deleteAlarm()`: <Type text='void' />
 
+  - Unset the alarm if there is a currently set alarm.
 
-* `deleteAlarm()`: <Type text='void' />
-
-  * Unset the alarm if there is a currently set alarm.
-
-  * Calling `deleteAlarm()` inside the `alarm()` handler may prevent retries on a best-effort basis, but is not guaranteed.
-
-
+  - Calling `deleteAlarm()` inside the `alarm()` handler may prevent retries on a best-effort basis, but is not guaranteed.
 
 ## Handler methods
 
 ### alarm
 
+- `alarm()`: <Type text='void' />
 
+  - Called by the system when a scheduled alarm time is reached.
 
-* `alarm()`: <Type text='void' />
+  - The `alarm()` handler has guaranteed at-least-once execution and will be retried upon failure using exponential backoff, starting at two second delays for up to six retries. Retries will be performed if the method fails with an uncaught exception.
 
-  * Called by the system when a scheduled alarm time is reached.
-
-  * The `alarm()` handler has guaranteed at-least-once execution and will be retried upon failure using exponential backoff, starting at two second delays for up to six retries. Retries will be performed if the method fails with an uncaught exception.
-
-  * This method can be `async`.
-
-
+  - This method can be `async`.
 
 ## Example
 
 This example shows how to both set alarms with the `setAlarm(timestamp)` method and handle alarms with the `alarm()` handler within your Durable Object.
 
-* The `alarm()` handler will be called once every time an alarm fires.
-* If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine.
-* Following a short delay, the `alarm()` handler will run from the beginning on the other machine.
+- The `alarm()` handler will be called once every time an alarm fires.
+- If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine.
+- Following a short delay, the `alarm()` handler will run from the beginning on the other machine.
 
 ```js
-import { DurableObject } from 'cloudflare:workers';
+import { DurableObject } from "cloudflare:workers";
 
 export default {
-  async fetch(request, env) {
-    let id = env.ALARM_EXAMPLE.idFromName("foo");
-    return await env.ALARM_EXAMPLE.get(id).fetch(request);
-  },
+	async fetch(request, env) {
+		let id = env.ALARM_EXAMPLE.idFromName("foo");
+		return await env.ALARM_EXAMPLE.get(id).fetch(request);
+	},
 };
 
 const SECONDS = 1000;
 
 export class AlarmExample extends DurableObject {
-  constructor(ctx, env) {
-    this.ctx = ctx;
-    this.storage = ctx.storage;
-  }
-  async fetch(request) {
-    // If there is no alarm currently set, set one for 10 seconds from now
-    let currentAlarm = await this.storage.getAlarm();
-    if (currentAlarm == null) {
-      this.storage.setAlarm(Date.now() + 10 * SECONDS);
-    }
-  }
-  async alarm() {
-    // The alarm handler will be invoked whenever an alarm fires.
-    // You can use this to do work, read from the Storage API, make HTTP calls
-    // and set future alarms to run using this.storage.setAlarm() from within this handler.
-  }
+	constructor(ctx, env) {
+		this.ctx = ctx;
+		this.storage = ctx.storage;
+	}
+	async fetch(request) {
+		// If there is no alarm currently set, set one for 10 seconds from now
+		let currentAlarm = await this.storage.getAlarm();
+		if (currentAlarm == null) {
+			this.storage.setAlarm(Date.now() + 10 * SECONDS);
+		}
+	}
+	async alarm() {
+		// The alarm handler will be invoked whenever an alarm fires.
+		// You can use this to do work, read from the Storage API, make HTTP calls
+		// and set future alarms to run using this.storage.setAlarm() from within this handler.
+	}
 }
 ```
 
 ## Related resources
 
-* Understand how to [use the Alarms API](/durable-objects/examples/alarms-api/) in an end-to-end example.
-* Read the [Durable Objects alarms announcement blog post](https://blog.cloudflare.com/durable-objects-alarms/).
-* Review the [Storage API](/durable-objects/api/storage-api/) documentation for Durable Objects.
+- Understand how to [use the Alarms API](/durable-objects/examples/alarms-api/) in an end-to-end example.
+- Read the [Durable Objects alarms announcement blog post](https://blog.cloudflare.com/durable-objects-alarms/).
+- Review the [Storage API](/durable-objects/api/storage-api/) documentation for Durable Objects.
--- a/src/content/docs/durable-objects/api/alarms.mdx
+++ b/src/content/docs/durable-objects/api/alarms.mdx
@@ -37,15 +37,19 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 
 ### getAlarm
 
-- <code>getAlarm()</code>: <Type text ='number | null' />
+- <code>getAlarm()</code>: <Type text="number | null" />
 
   - If there is an alarm set, then return the currently set alarm time as the number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
 
 ### setAlarm
 
-- <code> setAlarm(scheduledTimeMs <Type text="number" />)</code>: <Type text="void" />
+- <code>
+  	{" "}
+  	setAlarm(scheduledTimeMs <Type text="number" />)
+  </code>
+  : <Type text="void" />
 
-  - Set the time for the alarm to run.  Specify the time as the number of milliseconds elapsed since the UNIX epoch.
+  - Set the time for the alarm to run. Specify the time as the number of milliseconds elapsed since the UNIX epoch.
 
 ### deleteAlarm
 
@@ -72,7 +76,7 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 This example shows how to both set alarms with the `setAlarm(timestamp)` method and handle alarms with the `alarm()` handler within your Durable Object.
 
 - The `alarm()` handler will be called once every time an alarm fires.
-- If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine.
+- If an unexpected error terminates the Durable Object, the `alarm()` handler may be re-instantiated on another machine.
 - Following a short delay, the `alarm()` handler will run from the beginning on the other machine.
 
 ```js
--- a/src/content/docs/durable-objects/api/alarms.mdx
+++ b/src/content/docs/durable-objects/api/alarms.mdx
@@ -45,7 +45,7 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 
 - <code> setAlarm(scheduledTimeMs <Type text="number" />)</code>: <Type text="void" />
 
-  - Set the time for the alarm to run at in number of milliseconds elapsed since the UNIX epoch.
+  - Set the time for the alarm to run.  Specify the time as the number of milliseconds elapsed since the UNIX epoch.
 
 ### deleteAlarm
 
--- a/src/content/docs/durable-objects/api/alarms.mdx
+++ b/src/content/docs/durable-objects/api/alarms.mdx
@@ -39,7 +39,7 @@ Alarms can be used to build distributed primitives, like queues or batching of w
 
 - <code>getAlarm()</code>: <Type text ='number | null' />
 
-  - If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
+  - If there is an alarm set, then return the currently set alarm time as the number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
 
 ### setAlarm
 
--- a/src/content/docs/durable-objects/api/id.mdx
+++ b/src/content/docs/durable-objects/api/id.mdx
@@ -11,7 +11,7 @@ import { Tabs, TabItem } from "~/components";
 
 The `DurableObjectId` interface refers to a new or existing Durable Object instance. This interface is most frequently used by [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to obtain a stub for submitting requests to the Durable Object class.
 
-Note that creating an ID for a Durable Object instance does not invoke the constructor of the Durable Object. This is done lazily when calling [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to create a [`DurableObjectStub`](/durable-objects/api/stub) from a `DurableObjectId`. This ensures that objects are not constructed unless their methods can be directly invoked.
+Note that creating an ID for a Durable Object instance does not create the Durable Object. The Durable Object is created lazily after calling [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to create a [`DurableObjectStub`](/durable-objects/api/stub) from a `DurableObjectId`. This ensures that objects are not constructed until they are actually accessed.
 
 :::note[`DurableObjectId`]
 
@@ -53,7 +53,7 @@ A 64 digit hex string.
 ```js
 const id1 = env.MY_DURABLE_OBJECT.newUniqueId();
 const id2 = env.MY_DURABLE_OBJECT.newUniqueId();
-console.assert(!id1.equals(id2), "This should always be true");
+console.assert(!id1.equals(id2), "Different unique ids should never be equal.");
 ```
 
 #### Parameters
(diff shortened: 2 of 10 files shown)


src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
+---
+title: Durable Object ID
```
vy-ton, line 2: Thoughts on combining `ID` and `Stub` pages?

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+:::note[`newUniqueId`]
+
+IDs created by `newUniqueId` will result in lower latencies when getting a [`DurableObjectStub`](../stub) relative to [`idFromName`](../namespace/#idFromName).
+
+:::
+
+#### Parameters
+
+An optional object with the key `jurisdiction` and value of a [jurisdiction](../../reference/data-location/#restrict-durable-objects-to-a-jurisdiction) string.
```
vy-ton: Jurisdiction can also be specified with `idFromName`

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+`get` obtains a [`DurableObjectStub`](../stub) from a [`DurableObjectId`](../id) which can be used to invoke methods on a Durable Object instance.
+
+```js
+const id = env.MY_DURABLE_OBJECT.newUniqueId();
+const stub = env.MY_DURABLE_OBJECT.get(id);
+```
+
+#### Parameters
+
+A required [`DurableObjectId`](../id) and an optional object with the key `locationHint` and value of a [locationHint](../../reference/data-location/#provide-a-location-hint) string.
```
vy-ton: @Oxyjun can help identify the docs components to use.

My preference:
- bullet for each parameter to make clear the parameter count and required vs optional

One day, I dream this could all be auto-generated from code comments

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+A required [jurisdiction](../../reference/data-location/#restrict-durable-objects-to-a-jurisdiction) string.
+
+#### Return value
+
+A `DurableObjectNamespace` scoped to a particular geographic jurisdiction.
+
+## Related resources
+
+- [Durable Objects: Easy, Fast, Correct – Choose Three](https://blog.cloudflare.com/durable-objects-easy-fast-correct-choose-three/).
```
vy-ton, line 167: link to related `Best practices` section

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
+---
+title: Durable Object Namespace
+pcx_content_type: concept
+sidebar:
+  order: 1
```
vy-ton, line 5: Preferred order:

1. Namespace
2. ID
3. Stub
4. Storage
5. Websockets
6. Alarms
7. WebGPU

image

vy-ton (commented):
Initial feedback focused on `DurableObjectNamespace`, can do another pass after we finalize format

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+---
+
+import { Tabs, TabItem } from "~/components";
+
+## Description
+
+The `DurableObjectNamespace` interface is used to obtain a reference to a new or existing Durable Object instance. The interface is accessible from the fetch handler on a Cloudflare Worker via the `env` parameter, which is the standard interface for reference bindings declared in `wrangler.toml`.
+
+There are several functions of this interface that can be used to create an ID for a Durable Object instance. Note that creating an ID for a Durable Object instance does not invoke the constructor of the Durable Object. This is done lazily when calling [`DurableObjectNamespace::get`](../namespace/#get) to create a [`DurableObjectStub`](../stub) from a `DurableObjectId`. This ensures that objects are not constructed unless their methods can be directly invoked.
```
vy-ton: > `DurableObjectNamespace::get`

@Oxyjun For consistency, we should decide how to reference method calls in text. Sometimes its `get()`, etc

src/content/docs/durable-objects/api/id.mdx
joshthoward: I was thinking that we could keep them separate for consistency. It keeps the side bar navigation clean on the RHS and I'd expect the side bar navigation to be cleaned up further on the LHS as we continue down this path.

src/content/docs/durable-objects/api/namespace.mdx
joshthoward: I couldn't get that working. I think that we should add this capability.

src/content/docs/durable-objects/api/namespace.mdx
joshthoward: I would prefer that we do not use `get()` because this looks like we are calling a function with no parameters when it does in fact have required parameters. `DurableObjectNamespace::get` is common in C++, Rust, and Java, but it might be confusing in the JS community (I don't actually know). `DurableObjectNamespace.get` could also work.

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+pcx_content_type: concept
+sidebar:
+  order: 1
+---
+
+import { Tabs, TabItem } from "~/components";
+
+## Description
+
+The `DurableObjectNamespace` interface is used to obtain a reference to a new or existing Durable Object instance. The interface is accessible from the fetch handler on a Cloudflare Worker via the `env` parameter, which is the standard interface for reference bindings declared in `wrangler.toml`.
```
cacheburner: `s/reference bindings/referencing bindings`

src/content/docs/durable-objects/api/namespace.mdx
cacheburner: nit: `s/several functions/several methods`

That said, I found it a bit confusing to read, so maybe: "_This interface defines several methods that can be used to create an ID for a Durable Object instance._"

Then we can link to the `Methods` section below in the word "methods".

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+  ...
+}
+
+// Worker
+export default {
+  async fetch(request, env) {
+    // Every unique ID refers to an individual instance of the Durable Object class
+    const id = env.MY_DURABLE_OBJECT.idFromName("foo");
+
+    // A stub is a client Object used to invoke methods on the Durable Object instance.
```
cacheburner: Hmmm, maybe we should say `A 'stub' is...`, since it's more of a concept, not just the variable?

It might also help to slightly rephrase with "invoke methods defined by the Durable Object instance."

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+#### Return value
+
+A [`DurableObjectId`](../id) refering to an instance of a Durable Object class.
+
+### `newUniqueId`
+
+#### Description
+
+creates a `DurableObjectId` which refers to an individual instance of the
```
cacheburner: Let's start this off with `newUniqueId` since we do that in the other descriptions.

src/content/docs/durable-objects/api/stub.mdx
```diff
@@ -0,0 +1,114 @@
...
+export default {
+	async fetch(request, env) {
+		// Every unique ID refers to an individual instance of the Durable Object class
+		const id = env.MY_DURABLE_OBJECT.idFromName("foo");
+
+		// A stub is a client used to invoke methods on the Durable Object instance
+		const stub = env.MY_DURABLE_OBJECT.get(id);
+
+		// Methods on the Durable Object are invoked via the stub
+		const rpcResponse = stub.sayHello();
```
cacheburner, line 40: Hmm, I wonder if this particular example might be misleading to those unfamiliar with RPC.

The method `sayHello()` isn't marked `async`, and we aren't `await`ing it on line 40 where we call it, so to the casual observer this looks like a synchronous call to the `sayHello()` method.

Maybe we can improve here by having the Worker prefix the resulting string?

```
// Some comment about how this is a REMOTE call, and we want to do more processing before returning it
// so we're going to await.
const rpcResponse = await stub.sayHello();
return new Response(`Message from DurableObject ${rpcResponse}`);
```

src/content/docs/durable-objects/api/stub.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+```js
+const id = env.MY_DURABLE_OBJECT.idFromName("foo");
+const stub = env.MY_DURABLE_OBJECT.get(id);
+console.assert(stub.name === "foo", "This should always be true");
+```
+
+## Related resources
+
+- [Durable Objects: Easy, Fast, Correct – Choose Three](https://blog.cloudflare.com/durable-objects-easy-fast-correct-choose-three/).
```
cacheburner, line 110: We should probably link something else here, since some customers may not be interested in the storage API at all. Maybe something that covers RPC or error handling?

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+pcx_content_type: concept
+sidebar:
+  order: 2
+---
+
+import { Tabs, TabItem } from "~/components";
+
+## Description
+
+The `DurableObjectId` interface refers to a new or existing Durable Object instance. This interface is most frequently used to by [`DurableObjectNamespace::get`](./namespace.mdx) to obtain a stub for submitting requests to the Durable Object class.
```
cacheburner: `s/used to by/used by`

cacheburner: > submitting requests to the Durable Object class.

I think we should change "to the Durable Object class" -> "to a Durable Object instance".

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+Note that creating an ID for a Durable Object instance does not invoke the constructor of the Durable Object. This is done lazily when calling [`DurableObjectNamespace::get`](../namespace/#get) to create a [`DurableObjectStub`](../stub) from a `DurableObjectId`. This ensures that objects are not constructed unless their methods can be directly invoked.
+
+:::note[`DurableObjectId`]
+
+If you are experiencing an issue with a particular Durable Object instance the `DurableObjectId` corresponding to this particular instance is very helpful for obtaining a resolution through Cloudflare support. It is advisable to log a `DurableObjectId` from a Worker and include it in support requests where applicable.
```
cacheburner: I think we should drop "very" from "very helpful".

This paragraph was a bit tricky to read, could we shorten it? Perhaps:


_If you are experiencing an issue with a particular Durable Object instance, consider logging the `DurableObjectId` from your Worker and including it in your Cloudflare support request._

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+### `toString`
+
+#### Description
+
+`toString` converts a `DurableObjectId` to a 64 digit hex string. This string is useful for logging purposes or storing the `DurableObjectId` elsewhere, e.g. in a session cookie. This string can be used to reconstruct a `DurableObjectId` via `DurableObjectNamespace::idFromString`.
+
+```js
+// Create a new unique ID
+const id = env.MY_DURABLE_OBJECT.newUniqueId();
+// Save the unique ID elsewhere, e.g. a session cookie via id.toString()
```
cacheburner: This example seems a bit awkward to me because it never actually calls `toString()`. When I read the **Parameters** section below and saw "None" I was confused because the last thing I read was `idFromString()`. It might be useful to include a legitimate call to `id.toString()`.

src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -40,94 +37,81 @@ Alarms can be used to build distributed primitives, like queues or batching of w
...
-* `getAlarm()`: <Type text ='number | null' />
-
-  * If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
-
-
+  - If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
 
 ### setAlarm
 
+- <code>
```
cacheburner, line 46: The formatting of this is pretty messed up here https://joshthoward-do-docs.cloudflare-docs-7ou.pages.dev/durable-objects/api/alarms/#setalarm. Probably need to use backticks.

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -3,85 +3,51 @@ title: Data location
...
 
-Because the jurisdiction is encoded permanently in the Durable Object ID, it is possible to have the same name represent different Durable Objects in different jurisdictions. For example: `OBJECT_NAMESPACE.idFromName('my-name')` and `OBJECT_NAMESPACE.jurisdiction('eu').idFromName('my-name')` represent different Durable Objects. They will have their own transient (in-memory) and persistent state, and will likely run in different geographical regions.
-
-This may be counterintuitive at first, but it would be impossible to enforce two different non-overlapping jurisdictions for a single name. The key insight to remember is that Durable Object namespaces operate on IDs, not names, and the jurisdiction is a permanent part of the ID.
+:::note[logging]
 
+A [`DurableObjectId`](../../api/id) will be logged outside of the specified jurisdiction for billing and debugging purposes.
```
cacheburner: What does this mean?

src/content/docs/durable-objects/api/alarms.mdx
Oxyjun: Line breaks were breaking the format. Have adjusted in my own commit.

src/content/docs/durable-objects/api/id.mdx
Oxyjun: Adding fix

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: Have realigned with preferred ordering.

src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -20,18 +19,16 @@ Durable Objects alarms allow you to schedule the Durable Object to be woken up a
 
 Notably:
 
-* Each Durable Object instance is able to schedule a single alarm at a time by calling `setAlarm()`.
-* Alarms have guaranteed at-least-once execution and are retried automatically when the `alarm()` handler throws.
-* Retries are performed using exponential backoff starting at a two second delay from the first failure with up to six retries allowed.
+- Each Durable Object instance is able to schedule a single alarm at a time by calling `setAlarm()`.
+- Alarms have guaranteed at-least-once execution and are retried automatically when the `alarm()` handler throws.
+- Retries are performed using exponential backoff starting at a two second delay from the first failure with up to six retries allowed.
```
justin-mp: nit: I don't know what our policy is for numbers, but "2s" or even "2 seconds" seems better than "two seconds".  Same with 6/six later on.

src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -40,94 +37,78 @@ Alarms can be used to build distributed primitives, like queues or batching of w
...
 
+- <code>getAlarm()</code>: <Type text ='number | null' />
 
-
-* `getAlarm()`: <Type text ='number | null' />
-
-  * If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
-
-
+  - If there is an alarm set, then return the currently set alarm time in number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
```
justin-mp:
```suggestion
  - If there is an alarm set, then return the currently set alarm time as the number of milliseconds elapsed since the UNIX epoch. Otherwise, return `null`.
```

src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -40,94 +37,78 @@ Alarms can be used to build distributed primitives, like queues or batching of w
...
 
+- <code> setAlarm(scheduledTimeMs <Type text="number" />)</code>: <Type text="void" />
 
-
-* <code>setAlarm(scheduledTimeMs <Type text='number'/>)</code>: <Type text='void' />
-
-  * Set the time for the alarm to run at in number of milliseconds elapsed since the UNIX epoch.
-
-
+  - Set the time for the alarm to run at in number of milliseconds elapsed since the UNIX epoch.
```
justin-mp:
```suggestion
  - Set the time for the alarm to run.  Specify the time as the number of milliseconds elapsed since the UNIX epoch.
```

src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -40,94 +37,78 @@ Alarms can be used to build distributed primitives, like queues or batching of w
...
 
 ## Example
 
 This example shows how to both set alarms with the `setAlarm(timestamp)` method and handle alarms with the `alarm()` handler within your Durable Object.
 
-* The `alarm()` handler will be called once every time an alarm fires.
-* If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine.
-* Following a short delay, the `alarm()` handler will run from the beginning on the other machine.
+- The `alarm()` handler will be called once every time an alarm fires.
+- If an unexpected error terminates the Durable Object, the `alarm()` handler will be re-instantiated on another machine.
```
justin-mp: "will be" or "may be"?  With SQLite-backed DOs, re-instantiating the DO will likely be on the same metal.

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+  order: 2
+---
+
+import { Tabs, TabItem } from "~/components";
+
+## Description
+
+The `DurableObjectId` interface refers to a new or existing Durable Object instance. This interface is most frequently used by [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to obtain a stub for submitting requests to the Durable Object class.
+
+Note that creating an ID for a Durable Object instance does not invoke the constructor of the Durable Object. This is done lazily when calling [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to create a [`DurableObjectStub`](/durable-objects/api/stub) from a `DurableObjectId`. This ensures that objects are not constructed unless their methods can be directly invoked.
```
justin-mp:
```suggestion
Note that creating an ID for a Durable Object instance does not create the Durable Object. The Durable Object is created lazily after calling [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to create a [`DurableObjectStub`](/durable-objects/api/stub) from a `DurableObjectId`. This ensures that objects are not constructed until they are actually accessed.
```

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+### `equals`
+
+#### Description
+
+`equals` is used to compare equality between two instances of `DurableObjectId`.
+
+```js
+const id1 = env.MY_DURABLE_OBJECT.newUniqueId();
+const id2 = env.MY_DURABLE_OBJECT.newUniqueId();
+console.assert(!id1.equals(id2), "This should always be true");
```
justin-mp:
```suggestion
console.assert(!id1.equals(id2), "Different unique ids should never be equal.");
```

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+## Properties
+
+### `name`
+
+#### Description
+
+`name` is an optional property of a `DurableObjectId`, which returns the name that was used to create the `DurableObjectId` via [`DurableObjectNamespace::idFromName`](/durable-objects/api/namespace/#idfromname). This value is undefined if the `DurableObjectId` was constructed using [`DurableObjectNamespace::newUniqueId`](/durable-objects/api/namespace/#newuniqueid).
+
+```js
+const id1 = env.MY_DURABLE_OBJECT.newUniqueId();
```
justin-mp: In this block of code, it would be better to call the variables `uniqueId` and `fromNameId` so that it's clear what you're comparing later.

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+const id2 = env.MY_DURABLE_OBJECT.idFromName("foo");
+console.assert(id1.name === undefined, "This should always be true");
```
justin-mp:
```suggestion
console.assert(id1.name === undefined, "unique ids have no name");
```

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+console.assert(id2.name === "foo", "This should always be true");
```
justin-mp:
```suggestion
console.assert(id2.name === "foo", "name matches parameter to idFromName");
```

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,177 @@
...
+  order: 1
+---
+
+import { Tabs, TabItem } from "~/components";
+
+## Description
+
+The `DurableObjectNamespace` interface is used to obtain a reference to a new or existing Durable Object instance. The interface is accessible from the fetch handler on a Cloudflare Worker via the `env` parameter, which is the standard interface when referencing bindings declared in `wrangler.toml`.
+
+This interface defines several [methods](/durable-objects/api/namespace/#methods) that can be used to create an ID for a Durable Object instance. Note that creating an ID for a Durable Object instance does not invoke the constructor of the Durable Object. This is done lazily when calling [`DurableObjectNamespace::get`](/durable-objects/api/namespace/#get) to create a [`DurableObjectStub`](/durable-objects/api/stub) from a `DurableObjectId`. This ensures that objects are not constructed unless their methods can be directly invoked.
```
justin-mp: I added suggestions to the copy of this in id.mdx.  Whatever text we use for this description in id.mdx, we should use here too.

src/content/docs/durable-objects/api/namespace.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+#### Return value
+
+A [`DurableObjectStub`](/durable-objects/api/stub) referring to an instance of a Durable Object class.
+
+### `jurisdiction`
+
+#### Description
+
+`jurisdiction` creates a subnamespace from a namespace where all Durable Object instance IDs and references created from that subnamespace will be restricted to the specified [jurisdiction](/durable-objects/reference/data-location/#restrict-durable-objects-to-a-jurisdiction).
```
justin-mp, line 150: Dumb question: what happens if the objects aren't in the jurisdiction of the subnamespace?

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -100,7 +100,7 @@ Each method is implicitly wrapped inside a transaction, such that its results ar
 
 * <code>deleteAll(options <Type text='Object' /> <MetaInfo text='optional' />)</code>: <Type text='Promise' />
 
-  * Deletes all stored data, effectively deallocating all storage used by the Durable Object. For Durable Objects with a key-value storage backend, `deleteAll()` removes all keys and associated values for an individual Durable Object. For Durable Objects with a [SQLite storage backend](/durable-objects/best-practices/access-durable-objects-storage/#sqlite-storage-backend), `deleteAll()` removes the entire contents of a Durable Object's private SQLite database.  
+  * Deletes all stored data, effectively deallocating all storage used by the Durable Object. For Durable Objects with a key-value storage backend, `deleteAll()` removes all keys and associated values for an individual Durable Object. For Durable Objects with a [SQLite storage backend](/durable-objects/best-practices/access-durable-objects-storage/#sqlite-storage-backend), `deleteAll()` removes the entire contents of a Durable Object's private SQLite database.
```
justin-mp:
```suggestion
  * Deletes all stored data, effectively deallocating all storage used by the Durable Object. For Durable Objects with a key-value storage backend, `deleteAll()` removes all keys and associated values for an individual Durable Object. For Durable Objects with a [SQLite storage backend](/durable-objects/best-practices/access-durable-objects-storage/#sqlite-storage-backend), `deleteAll()` removes the entire contents of a Durable Object's private SQLite database, including both SQL data and key-value data.
```

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -3,85 +3,51 @@ title: Data location
...
-
-
-Durable Object IDs will be logged outside of the specified jurisdiction for billing and debugging purposes.
-
-
-:::
+| Parameter | Location                     |
+| --------- | ---------------------------- |
+| eu        | The European Union           |
+| fedramp   | The United States of America |
```
justin-mp: fedramp doesn't necessarily mean "USA", just that the data is stored in FedRAMP-compliant data centers.  It's possible for FedRAMP locations to be located outside of the USA.  It's better to say FedRAMP-compliant data centers or something here and link to our FedRAMP documentation.

justin-mp (commented):
I like how we have all of our APIs laid out by type then by methods.  It makes it much easier to find documentation given a method name.

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: I don't have the domain knowledge to advise on this one. I'll leave it as-is for now.

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: Not documenting `idFromName` for now - please feel free to ping me if it's ready to be added in :D

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: @MellowYarker how would you rephrase?

My (limited) understanding of a stub is a function that is used to replace actual functions (usually for unit testing)

src/content/docs/durable-objects/reference/data-location.mdx
Oxyjun: @vy-ton  could you advise?

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+## Properties
+
+### `name`
+
+#### Description
+
+`name` is an optional property of a `DurableObjectId`, which returns the name that was used to create the `DurableObjectId` via [`DurableObjectNamespace::idFromName`](/durable-objects/api/namespace/#idfromname). This value is undefined if the `DurableObjectId` was constructed using [`DurableObjectNamespace::newUniqueId`](/durable-objects/api/namespace/#newuniqueid).
+
+```js
+const id1 = env.MY_DURABLE_OBJECT.newUniqueId();
```
Oxyjun:
```suggestion
const uniqueId = env.MY_DURABLE_OBJECT.newUniqueId();
```

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+const id2 = env.MY_DURABLE_OBJECT.idFromName("foo");
```
Oxyjun:
```suggestion
const fromNameId = env.MY_DURABLE_OBJECT.idFromName("foo");
```

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+console.assert(id1.name === undefined, "unique ids have no name");
```
Oxyjun:
```suggestion
console.assert(uniqueId.name === undefined, "unique ids have no name");
```

src/content/docs/durable-objects/api/id.mdx
```diff
@@ -0,0 +1,84 @@
...
+console.assert(id2.name === "foo", "name matches parameter to idFromName");
```
Oxyjun:
```suggestion
console.assert(fromNameId.name === "foo", "name matches parameter to idFromName");
```

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: @vy-ton could you advise?

src/content/docs/durable-objects/reference/data-location.mdx
Oxyjun:
```suggestion
| fedramp   | FedRAMP-compliant data centers |
```

src/content/docs/durable-objects/api/id.mdx
Oxyjun: Replaced `id1` with `uniqueId`
Replaced `id2` with `fromNameId`

src/content/docs/durable-objects/reference/data-location.mdx
Oxyjun: For now, I've simply replaced it with "FedRamp", as that seems to be consistent with https://developers.cloudflare.com/r2/reference/data-location/#available-jurisdictions.

src/content/docs/durable-objects/api/namespace.mdx
justin-mp: The `stub` here is this kind of stub: https://en.wikipedia.org/wiki/Stub_(distributed_computing)

In RPC systems like ours, the `stub` is the RPC client for the RPC server (the DO in this case).

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: Added link to https://developers.cloudflare.com/durable-objects/best-practices/access-durable-objects-from-a-worker/

src/content/docs/durable-objects/api/namespace.mdx
Oxyjun: @vy-ton @joshthoward I can help with this, but usually the parameters need some sort of name for the components to be useful. Could you advise?

E.g. general format is 

`- PARAMETER NAME( ) <TYPE> <Required/Optional>`

src/content/docs/durable-objects/api/stub.mdx
joshthoward: I'm going to punt on this for now. We should think about how we direct customers after we have more of the docs cleaned up.

src/content/docs/durable-objects/reference/data-location.mdx
joshthoward: This is to clarify that we are processing metadata outside of the jurisdiction.

src/content/docs/durable-objects/api/namespace.mdx
joshthoward: This is answered a bit further down. `subnamespace.get` will throw on an ID in a different jurisdiction.

src/content/docs/durable-objects/api/namespace.mdx
joshthoward: I am trying to follow MDN docs. We can standardize this a bit more in the future but I'm leaving this as 
"{A required| An optional} <type> <purpose>"

justin-mp (approved):

elithrar (approved):

vy-ton (approved):

src/content/docs/durable-objects/api/namespace.mdx
cacheburner: Oh, I just meant say _\`stub\`_ instead of _stub_ (in the comment). It was just a nitpick.

PR #17429, opened 2024-10-09
[Docs Site] Add hashes back into entrypoint filenames, disable hoisting.

KianNH:
### Summary

Add hashes back into entrypoint filenames, disable hoisting.

The `hoisted.js` file is present in the `<head>` of every page built, but the content hashes were changing as the scripts imports were identical but the ordering was not.

[`directRenderScript`](https://docs.astro.build/en/reference/configuration-reference/#experimentaldirectrenderscript) will be the default in Astro 5 and removes the need for hoisting in the first place, so we're happy to enable it now.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -104,6 +104,7 @@ export default defineConfig({
 	experimental: {
 		contentIntellisense: true,
 		contentLayer: true,
+		directRenderScript: true,
 	},
 	server: {
 		port: 1111,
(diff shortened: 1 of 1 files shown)


WalshyDev (approved):

penalosa (approved):

KianNH:
Tested that footnotes/Mermaid (loaded in `Head.astro`) & other JS like `<script>` tag on components/pages still work.

PR #17471, opened 2024-10-10
Add reference documentation for DurableObjectState

joshthoward:
### Summary

Add reference documentation for DurableObjectState. A lot of this is taken from the existing WebSockets documentation. I will shortly refactor those docs to operate more like a guide for using WebSockets with and without hibernation. 

### Screenshots (optional)

Note that the RHS bar needs some fixes to the CSS. Text is being wrapped when it shouldn't be. 

Screenshot 2024-10-10 at 12 30 18 PM

### Documentation checklist


src/content/docs/durable-objects/api/state.mdx
```diff
@@ -0,0 +1,282 @@
...
+  order: 4
+---
+
+import { Tabs, TabItem } from "~/components";
+
+## Description
+
+The `DurableObjectState` interface is accessible as an instance property on the Durable Objects class. This interface encapsulates methods that modify the state of a Durable Object, for example which WebSockets are attached to a Durable Object instance or how the runtime should handle concurrent Durable Object requests.
+
+The `DurableObjectState` interface is different that the [Storage API](/durable-objects/api/storage-api) in that it does not have top level methods which manipulate persistant application data. These methods are instead encapsulated in the [`DurableObjectStorage`](/durable-objects/api/storage-api) interface and accessed by [`DurableObjectState::storage`](/durable-objects/api/state/#storage).
```
lambrospetrou:
```suggestion
The `DurableObjectState` interface is different than the [Storage API](/durable-objects/api/storage-api) in that it does not have top level methods which manipulate persistant application data. These methods are instead encapsulated in the [`DurableObjectStorage`](/durable-objects/api/storage-api) interface and accessed by [`DurableObjectState::storage`](/durable-objects/api/state/#storage).
```

src/content/docs/durable-objects/api/state.mdx
```diff
@@ -0,0 +1,282 @@
...
+
+<Tabs> <TabItem label="JavaScript" icon="seti:javascript">
+
+```js
+import { DurableObject } from "cloudflare:workers";
+
+// Durable Object
+export class MyDurableObject extends DurableObject {
+  // DurableObjectState is accessible via the ctx instance property
+	constructor(ctx, env) {
```
lambrospetrou, line 24: Why are we naming it `ctx` and not `state`? There is a mismatch I feel here between the name and the type.

src/content/docs/durable-objects/api/state.mdx
```diff
@@ -0,0 +1,282 @@
...
+}
+```
+
+</TabItem> </Tabs>
+
+## Methods
+
+### `waitUntil`
+
+`waitUntil` waits until the promise which is passed as a parameter resolves and can extends an event context up to 30 seconds after the last client disconnects.
```
lambrospetrou:
```suggestion
`waitUntil` waits until the promise which is passed as a parameter resolves and extends an event context up to 30 seconds after the last client disconnects.
```
Do people understand what an "event context" is? I don't really.

src/content/docs/durable-objects/api/state.mdx
```diff
@@ -0,0 +1,282 @@
...
+
+- A required `WebSocket`.
+
+#### Return values
+
+- An `Array<string>` of tags.
+
+### `abort`
+
+`abort` is used to forcibly reset a Durable Object instance. A JavaScript Error with the message passed as a parameter will be logged. This error is not able to be caught within the applciation code.
```
lambrospetrou:
```suggestion
`abort` is used to forcibly reset a Durable Object instance. A JavaScript `Error` with the message passed as a parameter will be logged. This error is not able to be caught within the application code.
```

src/content/docs/durable-objects/api/state.mdx
joshthoward: It's effectively how long we keep the DO running in memory. Do you have an idea on how to better frame this?

src/content/docs/durable-objects/api/state.mdx
joshthoward: This is for consistency with WorkerEntrypoints. It's documented elsewhere, but that documentation should be moved here.

cacheburner:
Looks like the page preview failed to build? https://github.com/cloudflare/cloudflare-docs/actions/runs/11279062020/job/31368913837?pr=17471

src/content/docs/durable-objects/api/state.mdx
Oxyjun:
```suggestion
The `DurableObjectState` interface is different from the [Storage API](/durable-objects/api/storage-api) in that it does not have top-level methods which manipulate persistent application data. These methods are instead encapsulated in the [`DurableObjectStorage`](/durable-objects/api/storage-api) interface and accessed by [`DurableObjectState::storage`](/durable-objects/api/state/#storage).
```

src/content/docs/durable-objects/api/state.mdx
Oxyjun: Let's add this to the DO glossary, and have it hover in-line. Could you help with the definition, in a way that users would understand?

image

Oxyjun: Glossary definition has been updated, the hover text will display the correct definition.

Oxyjun (approved):

PR #17487, opened 2024-10-11
[Docs Site] Cache Astro assets in ci.yml

KianNH:
### Summary

Cache Astro assets (resized images) like we do in publish-production and publish-preview.

Explicit NPM cache step is unnecessary, v4 will handle this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,19 +20,13 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 22
-
-      - name: Get npm cache directory
-        id: npm-cache-dir
-        shell: bash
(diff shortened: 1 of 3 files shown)


haleycode (approved):

GregBrimble:
Pages just caches the entire `.astro` directory. Any reason not to do that here? https://developers.cloudflare.com/pages/configuration/build-caching/#frameworks

> Explicit NPM cache step is unnecessary, v4 will handle this.

TIL. Good addition, GitHub!

KianNH:
> Pages just caches the entire `.astro` directory. Any reason not to do that here? https://developers.cloudflare.com/pages/configuration/build-caching/#frameworks

No particular reason! Currently Astro only caches assets between runs but with Astro 5 on the horizon there'll be more things in there related to incremental builds so I'll move it over to the whole folder.

PR #17513, opened 2024-10-14
Document Wrangler's OS & Node support policy

penalosa:
### Summary

This adds a documented OS and Node version support policy for Wrangler, which we didn't previously have explicitly documented.

The Node version support policy is potentially controversial.

### DIFF
--- a/src/content/docs/workers/wrangler/install-and-update.mdx
+++ b/src/content/docs/workers/wrangler/install-and-update.mdx
@@ -21,9 +21,9 @@ Wrangler is installed locally into each of your projects. This allows you and yo
 
 <Details header="Wrangler System Requirements">
 
-Wrangler supports the [Current, Active, and Maintenance](https://nodejs.org/en/about/previous-releases) versions of Node.js. We aim not to break older versions of Node.js, but occasionally will due to features that are unavailable in older version of Node.js or security patches.
+We support running the Wrangler CLI with the [Current, Active, and Maintenance](https://nodejs.org/en/about/previous-releases) versions of Node.js. Your Worker will always be executed in `workerd`, the open source Cloudflare Workers runtime.
 
-For most Wrangler features to work, your operating system needs to support `workerd`, the open source Cloudflare Workers runtime. `workerd`'s OS support policy is listed in it's [GitHub README](https://github.com/cloudflare/workerd?tab=readme-ov-file#running-workerd).
+Wrangler is only supported on macOS 13.5+, Windows 11, and Linux distros that support glib 2.31. This follows [`workerd`'s OS support policy](https://github.com/cloudflare/workerd?tab=readme-ov-file#running-workerd).
 
 </Details>
 
--- a/src/content/docs/workers/wrangler/install-and-update.mdx
+++ b/src/content/docs/workers/wrangler/install-and-update.mdx
@@ -18,6 +18,16 @@ To install [Wrangler](https://github.com/cloudflare/workers-sdk/tree/main/packag
 
 Wrangler is installed locally into each of your projects. This allows you and your team to use the same Wrangler version, control Wrangler versions for each project, and roll back to an earlier version of Wrangler, if needed.
 
+
+<Details header="Wrangler System Requirements">
+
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/wrangler/install-and-update.mdx
```diff
@@ -14,10 +14,20 @@ Wrangler is a command-line tool for building with Cloudflare developer products.
...
 
 Wrangler is installed locally into each of your projects. This allows you and your team to use the same Wrangler version, control Wrangler versions for each project, and roll back to an earlier version of Wrangler, if needed.
 
+
+<Details header="Wrangler System Requirements">
+
+Wrangler supports the [Current, Active, and Maintenance](https://nodejs.org/en/about/previous-releases) versions of Node.js. We aim not to break older versions of Node.js, but occasionally will due to features that are unavailable in older version of Node.js or security patches.
+
+For most Wrangler features to work, your operating system needs to support `workerd`, the open source Cloudflare Workers runtime. `workerd`'s OS support policy is listed in it's [GitHub README](https://github.com/cloudflare/workerd?tab=readme-ov-file#running-workerd).
```
irvinebroque: Can we just state this upfront here, rather than linking somewhere else? And simplify?

This doesn't include Windows, should it?

src/content/docs/workers/wrangler/install-and-update.mdx
```diff
@@ -14,10 +14,20 @@ Wrangler is a command-line tool for building with Cloudflare developer products.
...
+To install [Wrangler](https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler), ensure you have [Node.js](https://nodejs.org/en/) and [npm](https://docs.npmjs.com/getting-started) installed, preferably using a Node version manager like [Volta](https://volta.sh/) or [nvm](https://github.com/nvm-sh/nvm). Using a version manager helps avoid permission issues and allows you to change Node.js versions.
 
 Wrangler is installed locally into each of your projects. This allows you and your team to use the same Wrangler version, control Wrangler versions for each project, and roll back to an earlier version of Wrangler, if needed.
 
+
+<Details header="Wrangler System Requirements">
+
+Wrangler supports the [Current, Active, and Maintenance](https://nodejs.org/en/about/previous-releases) versions of Node.js. We aim not to break older versions of Node.js, but occasionally will due to features that are unavailable in older version of Node.js or security patches.
```
irvinebroque: Can we be explicit that we are talking about the requirements of the _CLI_ and not the requirements of your application?

Just thinking about how people can easily assume their app is running Node.js.

irvinebroque:
```suggestion
Wrangler supports the [Current, Active, and Maintenance](https://nodejs.org/en/about/previous-releases) versions of Node.js.
```

Would just leave it at this

penalosa:
@irvinebroque I've reworded—how does that look?

irvinebroque (approved):

Cherry (commented):
Up to this point, wrangler has advertised support for Node.js 16+.

While I’m generally in favour of a change like this, allowing wrangler to target newer ecosystem features, I feel that a v4 needs to be cut to drop Node.js 16 support and move over to this new policy moving forward?

DaniFoldi:
Hi :wave:,

I have to agree with James here. This is a solution we've discussed on Discord before, and even though it's the one I'd choose personally (as opposed to bumping major just for minimum _supported_ nodejs even when nothing would break, which is marginally safer but look at the adoption of ts 4.9 vs ts 5.0 even though 4.9 was no less breaking), it's one that is not compatible with the current version's policy - 16.13.0 has been explicitly mentioned in a lot of docs.

Fortunately, if the only breaking changes are to use a more recent nodejs version, and hopefully an esbuild bump (and the natural removal of some deprecated features), it will be a much easier migration for most people 🤞and the aforementioned version fragmentation won't be as bad. The longer deprecated v3 features are kept in the code/docs, the worse the migration will be, when the time comes.

penalosa:
Closing for now. We'll still do this, but it'll be part of the v4 updates: https://github.com/cloudflare/workers-sdk/pull/7334

PR #17522, opened 2024-10-14
[ZT] Update minimum MTU

ranbel:
PCX-13754


cf-rhett:
The information is correct but it looks a bit busy in my opinion. I'd almost feel like we could put 1360 and have another page with details around the differences between the MTU for each particular

cf-rhett (approved):

KianNH (approved):

PR #17524, opened 2024-10-14
add Wrangler 'find_additional_modules' documentation

petebacondarwin:
### Summary

Adds documentation for missing Wrangler configuration

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/wrangler/bundling.mdx
+++ b/src/content/docs/workers/wrangler/bundling.mdx
@@ -3,24 +3,28 @@ pcx_content_type: configuration
 title: Bundling
 head: []
 description: Review Wrangler's default bundling.
-
 ---
 
 By default, Wrangler bundles your Worker code using [`esbuild`](https://esbuild.github.io/). This means that Wrangler has built-in support for importing modules from [npm](https://www.npmjs.com/) defined in your `package.json`. To review the exact code that Wrangler will upload to Cloudflare, run `npx wrangler deploy --dry-run --outdir dist`, which will show your Worker code after Wrangler's bundling.
 
 :::note
 
-We recommend using Wrangler's inbuilt bundling, but we understand there are cases where you will need more flexibility. We have built an escape hatch in the form of [Custom Builds](/workers/wrangler/custom-builds/), which lets you run your own build before Wrangler's built-in one. 
+Wrangler's inbuilt bundling usually provides the best experience, but we understand there are cases where you will need more flexibility.
+You can provide `rules` and set `find_additional_modules` in your configuration to control which files are included in the deployed Worker but not bundled into the entry-point file.
+Furthermore, we have an escape hatch in the form of [Custom Builds](/workers/wrangler/custom-builds/), which lets you run your own build before Wrangler's built-in one.
 :::
 
-## Files which will not be bundled
+## Including non-JavaScript modules
 
-Bundling your Worker code takes multiple modules and bundles them into one. Sometimes, you might have modules that should not be inlined directly into the bundle. For example, instead of bundling a Wasm file into your JavaScript Worker, you would want to upload the Wasm file as a separate module that can be imported at runtime. Wrangler supports this for the following file types:
+Bundling your Worker code takes multiple modules and bundles them into one file.
+Sometimes, you might have modules that cannot be inlined directly into the bundle.
+For example, instead of bundling a Wasm file into your JavaScript Worker, you would want to upload the Wasm file as a separate module that can be imported at runtime.
+Wrangler supports this for the following file types:
 
-* `.txt`
-* `.html`
-* `.bin`
-* `.wasm` and `.wasm?module`
+- `.txt`
+- `.html`
+- `.bin`
+- `.wasm` and `.wasm?module`
 
 Refer to [Bundling configuration](/workers/wrangler/configuration/#bundling) to customize these file types.
 
@@ -37,16 +41,29 @@ import wasm from "./example.wasm"; // Where `example.wasm` is a file in your loc
 const instance = await WebAssembly.instantiate(wasm); // Instantiate Wasm modules in global scope, not within the fetch() handler
 
 export default {
-  fetch(request) {
-    const result = instance.exports.exported_func();
-  },
+	fetch(request) {
+		const result = instance.exports.exported_func();
+	},
 };
 ```
 
 :::caution
 
-Cloudflare Workers does not support `WebAssembly.instantiateStreaming()`. 
+Cloudflare Workers does not support `WebAssembly.instantiateStreaming()`.
 :::
+
+## Find additional modules
+
+By setting `find_additional_modules` to `true` in your configuration file, Wrangler will traverse the file tree below `base_dir`.
+Any files that match the `rules` you define will also be included as unbundled, external modules in the deployed Worker.
+
+This approach is useful for supporting lazy loading of large or dynamically imported JavaScript files:
+
+- Normally a large lazy imported file (e.g. `await import("./large-dep.mjs")`) would be bundled directly into your entrypoint, making lazy loading less effective.
+  If matching rule is added to `rules`, then this file would only be loaded and executed at runtime when it is actually imported.
+- Previously, variable based dynamic imports (e.g. ``await import(`./lang/${language}.mjs`)``) would always fail at runtime, as Wrangler would have no way of knowing which modules to include in the upload.
+  Providing a rule that matches all these files (e.g. ` { type = "EsModule", globs = ["./land/**/*.mjs"], fallthrough = true }), will ensure this module is available at runtime.
+- "Partial bundling" is supported when `find_additional_modules` is `true`, and a source file matches one of the configured `rules`, since Wrangler will then treat it as "external" and not try to bundle it into the entry-point file.
 
 ## Conditional exports
 
@@ -56,7 +73,7 @@ Wrangler respects the [conditional `exports` field](https://nodejs.org/api/packa
 
 :::caution
 
-Disabling bundling is not recommended in most scenarios. Use this option only when deploying code pre-processed by other tooling. 
+Disabling bundling is not recommended in most scenarios. Use this option only when deploying code pre-processed by other tooling.
 :::
 
 If your build tooling already produces build artifacts suitable for direct deployment to Cloudflare, you can opt out of bundling by using the `--no-bundle` command line flag: `npx wrangler deploy --no-bundle`. If you opt out of bundling, Wrangler will not process your code and some features introduced by Wrangler bundling (for example minification, and polyfills injection) will not be available.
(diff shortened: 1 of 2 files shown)


src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -37,17 +41,30 @@ import wasm from "./example.wasm"; // Where `example.wasm` is a file in your loc
...
+
+By setting `find_additional_modules` to `true` in your configuration file, Wrangler will traverse the file tree below `base_dir`.
+Any files that match the `rules` you define will also be included as unbundled, external modules in the deployed Worker.
+
+This approach is useful for supporting lazy loading of large or dynamically imported JavaScript files:
+
+- Normally, a large lazy-imported file (for example, `await import("./large-dep.mjs")`) would be bundled directly into your entrypoint, reducing the effectiveness of the lazy loading.
+  If matching rule is added to `rules`, then this file would only be loaded and executed at runtime when it is actually imported.
+- Previously, variable based dynamic imports (for example, ``await import(`./lang/${language}.mjs`)``) would always fail at runtime because Wrangler had no way of knowing which modules to include in the upload.
+  Providing a rule that matches all these files, such as { type = "EsModule", globs = ["./land/**/*.mjs"], fallthrough = true }, will ensure this module is available at runtime.
```
GregBrimble:
```suggestion
  Providing a rule that matches all these files, such as { type = "ESModule", globs = ["./land/**/*.mjs"], fallthrough = true }, will ensure this module is available at runtime.
```

src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -37,17 +41,30 @@ import wasm from "./example.wasm"; // Where `example.wasm` is a file in your loc
...
 
+## Find additional modules
+
+By setting `find_additional_modules` to `true` in your configuration file, Wrangler will traverse the file tree below `base_dir`.
+Any files that match the `rules` you define will also be included as unbundled, external modules in the deployed Worker.
+
+This approach is useful for supporting lazy loading of large or dynamically imported JavaScript files:
+
+- Normally, a large lazy-imported file (for example, `await import("./large-dep.mjs")`) would be bundled directly into your entrypoint, reducing the effectiveness of the lazy loading.
+  If matching rule is added to `rules`, then this file would only be loaded and executed at runtime when it is actually imported.
```
GregBrimble, line 63:
```suggestion
  If a matching rule is added to `rules`, then this file would only be loaded and executed in the runtime when it is actually imported.
```

src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -37,17 +41,30 @@ import wasm from "./example.wasm"; // Where `example.wasm` is a file in your loc
...
+- Previously, variable based dynamic imports (for example, ``await import(`./lang/${language}.mjs`)``) would always fail at runtime because Wrangler had no way of knowing which modules to include in the upload.
```
GregBrimble, line 64:
```suggestion
- Without `find_additional_modules`, variable-based dynamic imports (for example, ``await import(`./lang/${language}.mjs`)``) would always fail at runtime because Wrangler had no way of knowing which modules to include in the upload.
```

src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -3,24 +3,28 @@ pcx_content_type: configuration
...
+Furthermore, we have an escape hatch in the form of [Custom Builds](/workers/wrangler/custom-builds/), which lets you run your own build before Wrangler's built-in one.
 :::
 
-## Files which will not be bundled
+## Including non-JavaScript modules
 
-Bundling your Worker code takes multiple modules and bundles them into one. Sometimes, you might have modules that should not be inlined directly into the bundle. For example, instead of bundling a Wasm file into your JavaScript Worker, you would want to upload the Wasm file as a separate module that can be imported at runtime. Wrangler supports this for the following file types:
+Bundling your Worker code takes multiple modules and bundles them into one file.
+Sometimes, you might have modules that cannot be inlined directly into the bundle.
+For example, instead of bundling a Wasm file into your JavaScript Worker, you would want to upload the Wasm file as a separate module that can be imported at runtime.
```
GregBrimble, line 21:
```suggestion
Bundling your Worker code takes multiple modules and bundles them into one file.
However, sometimes you might have modules that cannot be inlined directly into JavaScript file.
In that case, Wrangler will discover these files and include them in your Worker separately.
For example, instead of bundling a Wasm file into your JavaScript Worker, Wrangler will automatically upload the Wasm file as a separate module that can be imported at runtime.
```

GregBrimble (approved):
I know it's not strictly related to your change, but I wonder if it's worth just calling out somewhere on this page that automatic bundling is what makes TypeScript possible. It's not actually obvious when we say stuff like "TypeScript is a first-class language on Cloudflare Workers" elsewhere in the docs.

This is especially important if we're making module discovery more accessible. It should be really clear that these modules need to be given to us as ESM/CJS/Wasm/Text or Data.

src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -3,24 +3,28 @@ pcx_content_type: configuration
...
 
 By default, Wrangler bundles your Worker code using [`esbuild`](https://esbuild.github.io/). This means that Wrangler has built-in support for importing modules from [npm](https://www.npmjs.com/) defined in your `package.json`. To review the exact code that Wrangler will upload to Cloudflare, run `npx wrangler deploy --dry-run --outdir dist`, which will show your Worker code after Wrangler's bundling.
 
 :::note
 
-We recommend using Wrangler's inbuilt bundling, but we understand there are cases where you will need more flexibility. We have built an escape hatch in the form of [Custom Builds](/workers/wrangler/custom-builds/), which lets you run your own build before Wrangler's built-in one. 
+Wrangler's inbuilt bundling usually provides the best experience, but we understand there are cases where you will need more flexibility.
+You can provide `rules` and set `find_additional_modules` in your configuration to control which files are included in the deployed Worker but not bundled into the entry-point file.
+Furthermore, we have an escape hatch in the form of [Custom Builds](/workers/wrangler/custom-builds/), which lets you run your own build before Wrangler's built-in one.
```
irvinebroque, line 14:
```suggestion
If you need more control, you can use [Custom Builds](/workers/wrangler/custom-builds/), which lets you configure your own build command that runs before Wrangler's built-in one.
```

src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -3,24 +3,28 @@ pcx_content_type: configuration
...
 description: Review Wrangler's default bundling.
-
 ---
 
 By default, Wrangler bundles your Worker code using [`esbuild`](https://esbuild.github.io/). This means that Wrangler has built-in support for importing modules from [npm](https://www.npmjs.com/) defined in your `package.json`. To review the exact code that Wrangler will upload to Cloudflare, run `npx wrangler deploy --dry-run --outdir dist`, which will show your Worker code after Wrangler's bundling.
 
 :::note
 
-We recommend using Wrangler's inbuilt bundling, but we understand there are cases where you will need more flexibility. We have built an escape hatch in the form of [Custom Builds](/workers/wrangler/custom-builds/), which lets you run your own build before Wrangler's built-in one. 
+Wrangler's inbuilt bundling usually provides the best experience, but we understand there are cases where you will need more flexibility.
```
irvinebroque, line 12:
```suggestion
Wrangler's built-in bundling usually provides the best experience, but we understand there are cases where you will need more flexibility.
```

src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -37,17 +41,30 @@ import wasm from "./example.wasm"; // Where `example.wasm` is a file in your loc
...
 :::
 
+## Find additional modules
+
+By setting `find_additional_modules` to `true` in your configuration file, Wrangler will traverse the file tree below `base_dir`.
+Any files that match the `rules` you define will also be included as unbundled, external modules in the deployed Worker.
+
+This approach is useful for supporting lazy loading of large or dynamically imported JavaScript files:
+
+- Normally, a large lazy-imported file (for example, `await import("./large-dep.mjs")`) would be bundled directly into your entrypoint, reducing the effectiveness of the lazy loading.
```
irvinebroque, line 62: Even if you upload these as separate modules, don't we still have an issue?

https://github.com/cloudflare/workerd/issues/2372

Or am I misunderstanding the problem there?

cc @jasnell

src/content/docs/workers/wrangler/configuration.mdx
```diff
@@ -978,6 +995,15 @@ export default {
...
 ```
 
+### Find additional modules
+
+Normally Wrangler will only include additional modules that are statically imported in your source code as in the example above.
+By setting `find_additional_modules` to `true` in your configuration file, Wrangler will traverse the file tree below `base_dir`.
+Any files that match `rules` will also be included as unbundled, external modules in the deployed Worker.
+`base_dir` defaults to the directory containing your `main` entrypoint.
+
+See https://developers.cloudflare.com/workers/wrangler/bundling/ for more details and examples.
```
irvinebroque, line 1005:
```suggestion
See [the developer docs on bundling](/workers/wrangler/bundling/) for more details and examples.
```

irvinebroque (approved):
ty @petebacondarwin for docs, few notes re: dynamic imports behavior at runtime

src/content/docs/workers/wrangler/bundling.mdx
petebacondarwin: There is definitely still a cost, since as I understand it, we do parse all modules at start up. The main difference, unless I am mistaken is that we don't execute the code in the lazy module until it is imported.
Despite this, we found that next-on-pages start up time was improved dramatically by splitting out chunks from the main entry-point, so there is some benefit.
Also, once we update the implementation in workerd this will become significantly improved, so things can only get better if you use this approach.

src/content/docs/workers/wrangler/bundling.mdx
petebacondarwin:
```suggestion
Bundling your Worker code takes multiple modules and bundles them into one file.
However, sometimes you might have modules that cannot be inlined directly into the JavaScript file.
In that case, Wrangler can discover these files and include them in your Worker separately.
For example, instead of bundling a Wasm file into your JavaScript Worker, Wrangler can upload the Wasm file as a separate module that can be imported at runtime.
```

petebacondarwin:
> I know it's not strictly related to your change, but I wonder if it's worth just calling out somewhere on this page that automatic bundling is what makes TypeScript possible. It's not actually obvious when we say stuff like "TypeScript is a first-class language on Cloudflare Workers" elsewhere in the docs.
> 
> This is especially important if we're making module discovery more accessible. It should be really clear that these modules need to be given to us as ESM/CJS/Wasm/Text or Data.

Actually, I think an even better solution would be for us to be able to run non-ESM JavaScript/TypeScript additional modules through esbuild to ensure they are also in the right state for consumption by workerd.

GregBrimble:
> Actually, I think an even better solution would be for us to be able to run non-ESM JavaScript/TypeScript additional modules through esbuild to ensure they are also in the right state for consumption by workerd.

There's definitely some code lying around somewhere for a "process entrypoint"/"bundle no-bundle" sort of thing which could be used. Interesting idea. And I guess `--no-bundle` still works as the opt-out. Yeah, could be explored!

irvinebroque:
Thank you Pete!!!

PR #17562, opened 2024-10-16
clarify that wildcards are not allowed in redirect-to-custom-domain.mdx

fry69:
As far as I can tell wildcards (`*`) are not supported for bulk redirect rules, so mentioning the `*` so prominently confuses a bit.

The `<deploy-id>` placeholder is debatable. If possible please offer a better suggestion. Or please explain if I overlooked something, I am still learning the Cloudflare API.


ToriLindsay:
👋 @fry69  Thank you for the feedback about how we can make improvements with[ developers.cloudflare.com](http://developers.cloudflare.com/) documentation. Closing this as a duplicate of #19283

PR #17576, opened 2024-10-16
[Docs Site] Fix footnotes with multiple references

KianNH:
### Summary

Fix footnotes with multiple references

### DIFF
--- a/src/scripts/footnotes.ts
+++ b/src/scripts/footnotes.ts
@@ -9,7 +9,7 @@ if (footnotes) {
 		const content = note.querySelector("p") as HTMLParagraphElement;
 
 		const fnrefs = document.querySelectorAll<HTMLAnchorElement>(
-			`#${note.id.replace("fn", "fnref")}`,
+			`a[id^='${note.id.replace("fn", "fnref")}']`,
 		);
 
 		for (const fnref of fnrefs) {


Oxyjun (approved):

GregBrimble:
Got an example of where this is used?

KianNH:
> Got an example of where this is used?

https://developers.cloudflare.com/durable-objects/platform/limits/ vs. https://kian-pcx-14165.cloudflare-docs-7ou.pages.dev/durable-objects/platform/limits/

src/scripts/footnotes.ts
```diff
@@ -9,7 +9,7 @@ if (footnotes) {
 		const content = note.querySelector("p") as HTMLParagraphElement;
 
 		const fnrefs = document.querySelectorAll<HTMLAnchorElement>(
-			`#${note.id.replace("fn", "fnref")}`,
+			`a[id^='${note.id.replace("fn", "fnref")}']`,
```
GregBrimble, line 12: Oh, does duplicate ones do something like `id="blah-2"`? Is that why you're switching to `^=`?

GregBrimble (approved):

src/scripts/footnotes.ts
KianNH: Yep! You get `user-content-fnref-2` and `user-content-fnref-2-2` 😢

PR #17657, opened 2024-10-20
Add missing product to footnote Renders

mhart:
### Summary

A couple of `Render` components are missing the right `product` attributes. This adds them in.


pedrosousa:
It seems that the footnote definition must be in the same partial where it is referenced (after your changes, the footnote was still not being rendered). I made a few adjustments to your proposal and checked locally that it worked.

pedrosousa (approved):

PR #17670, opened 2024-10-21
[DO] Update WebSocket Hibernation example

harshil1712:
### Summary

This example adds the `serializeAttachment` and `deserialzieAttachemt` APIs to the existing example. It also shows how to send messages with the following condition:
- Send message to the client
- Send message to all the connected clients
- Send message to all the connected clients except sender


src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
```diff
@@ -32,146 +29,225 @@ import { DurableObject } from "cloudflare:workers";
...
-    // total number of connections.
-    ws.send(`[Durable Object] message: ${message}, connections: ${this.ctx.getWebSockets().length}`);
-  }
-
-  async webSocketClose(ws, code, reason, wasClean) {
-    // If the client closes the connection, the runtime will invoke the webSocketClose() handler.
-    ws.close(code, "Durable Object is closing WebSocket");
-  }
+	// Keep track of all WebSocket connections
+	sessions = new Map();
```
cacheburner: This is lost every time we hibernate FYI. You would probably want to reconstruct it whenever the `constructor()` runs.

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
```diff
@@ -32,146 +29,225 @@ import { DurableObject } from "cloudflare:workers";
...
+		// and allowing the WebSocket to send and receive messages.
+		// Unlike `ws.accept()`, `state.acceptWebSocket(ws)` informs the Workers Runtime that the WebSocket
+		// is "hibernatable", so the runtime does not need to pin this Durable Object to memory while
+		// the connection is open. During periods of inactivity, the Durable Object can be evicted
+		// from memory, but the WebSocket connection will remain open. If at some later point the
+		// WebSocket receives a message, the runtime will recreate the Durable Object
+		// (run the `constructor`) and deliver the message to the appropriate handler.
+		this.ctx.acceptWebSocket(server);
+
+		// Keep a copy of value in memory to survive hibernation.
```
cacheburner: I'm confused by this comment? Hibernation means evicting and losing all in-memory state. When the DO "hibernates" and then comes back, `this.sessions` is set to `{}`.

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
```diff
@@ -32,146 +29,225 @@ import { DurableObject } from "cloudflare:workers";
...
+
+		return new Response(null, {
+			status: 101,
+			webSocket: client,
+		});
+	}
+
+	async webSocketMessage(sender, message) {
+		// Upon receiving a message, get the session associated with the WebSocket connection.
+		const session = this.sessions.get(sender);
```
cacheburner: Every time we hibernate and then receive a new message, `this.sessions.get(sender)` will not return the associated session because `this.sessions` is lost whenever we hibernate.

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
```diff
@@ -32,146 +29,225 @@ import { DurableObject } from "cloudflare:workers";
...
+
+		// If it is a new connection, generate a new ID for the session.
```
cacheburner: As per above comment, this doesn't actually happen on a "per connection basis". Since `this.sessions` is cleared whenever we're evicted, the `session.id` will always not exist the first time we get a message on a pre-existing connection.

I think what you really want to do is something more like:

1. When we receive a message, check the `sessions` map. If we have an associated session there, great, use that.
2. If we don't have a session, check if we've already serialized something on the WS, i.e. `sender.deserializeAttachment()` and check if we have an ID, if we do, create the session and set it on `sessions` `this.sessions.set(sender, sendersID);`.

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
```diff
@@ -32,146 +29,225 @@ import { DurableObject } from "cloudflare:workers";
...
+		}
+
+		// Upon receiving a message from the client, the server replies with the same message,
+		// and the total number of connections with the "[Durable Object]: " prefix
+		sender.send(
+			`[Durable Object] message: ${message}, from: ${session.id}. Total connections: ${this.ctx.getWebSockets().length}`,
+		);
+
+		// Send a message to all WebSocket connections, loop over all the connected WebSockets.
+		this.ctx.getWebSockets().forEach((ws) => {
```
cacheburner, line 84: Just FYI calling `getWebSockets()` will unhibernate _all_ your websocket connections. In other words, when the DO hibernates, all connections are also "hibernated", and when the DO wakes back up, we only load your connections back into memory when you need them.

This means that if you call `getWebSockets()` (bc you wanna broadcast), you load all your WS connections back into memory. If you _don't_ broadcast, and instead only reply to the WS that sent the DO a message, then only that DO is brought back into DO memory.

I don't think we actually explain this well in the docs, but yeah, just something you might wanna know!

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
cacheburner: Alternatively, in the DOs `constructor()` you could just do something like `this.ctx.getWebSockets()` and then for each load it into `sessions`.

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
```diff
@@ -32,146 +29,225 @@ import { DurableObject } from "cloudflare:workers";
...
+	}
+
+	async webSocketMessage(sender, message) {
+		// Upon receiving a message, get the session associated with the WebSocket connection.
+		const session = this.sessions.get(sender);
+
+		// If it is a new connection, generate a new ID for the session.
+		if (!session.id) {
+			session.id = crypto.randomUUID();
+			sender.serializeAttachment({
```
cacheburner: You probably want to actually set the `session.id` when you first create the websocket in `fetch()`.

cacheburner:
I think we should make the same changes for the non-hibernation example. New devs might look at regular WS example, then see the Hibernation example and be hesitant to use hibernation because it looks significantly more complicated.

If we don't do that, we should _at least_ modify the description of the example, because currently it says "_This example is similar to the [Build a WebSocket server](https://update-do-examples.cloudflare-docs-7ou.pages.dev/durable-objects/examples/websocket-server/) example, but uses the WebSocket Hibernation API"_.

src/content/docs/durable-objects/examples/websocket-hibernation-server.mdx
joshthoward: What Milan said, but you probably want something like
```
	constructor(ctx: DurableObjectState, env: Env) {
    super(ctx, env);
    this.sessions = new Map();
    this.ctx.getWebSockets().forEach((ws) => {
      const { ...session } = ws.deserializeAttachment();
      this.sessions.set(ws, { ...session });
    });
	}
```
You'll also have to `ws.serializeAttachment` every time that you update sessions.

PR #17698, opened 2024-10-22
[Durable Objects] Document transaction API changes with SQLite.

kentonv:
Document that:
* The existing `storage.transaction()` API applies to SQL queries.
* There is also `storage.transactionSync()` for synchronous transactions.

### DIFF
--- a/src/content/docs/durable-objects/api/storage-api.mdx
+++ b/src/content/docs/durable-objects/api/storage-api.mdx
@@ -191,7 +191,21 @@ The `put()` method returns a `Promise`, but most applications can discard this p
 
 * `txn`
 
-  * Provides access to the `put()`, `get()`, `delete()` and `list()` methods documented above to run in the current transaction context. In order to get transactional behavior within a transaction closure, you must call the methods on the `txn` Object instead of on the top-level `state.storage` Object.<br/><br/>Also supports a `rollback()` function that ensures any changes made during the transaction will be rolled back rather than committed. After `rollback()` is called, any subsequent operations on the `txn` Object will fail with an exception. `rollback()` takes no parameters and returns nothing to the caller.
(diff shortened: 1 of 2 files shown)


lambrospetrou (approved):

src/content/docs/durable-objects/api/storage-api.mdx
```diff
@@ -191,7 +191,21 @@ The `put()` method returns a `Promise`, but most applications can discard this p
...
+  * Provides access to the `put()`, `get()`, `delete()` and `list()` methods documented above to run in the current transaction context. In order to get transactional behavior within a transaction closure, you must call the methods on the `txn` Object instead of on the top-level `ctx.storage` Object.<br/><br/>Also supports a `rollback()` function that ensures any changes made during the transaction will be rolled back rather than committed. After `rollback()` is called, any subsequent operations on the `txn` Object will fail with an exception. `rollback()` takes no parameters and returns nothing to the caller.
+
+  * When using [the SQLite-backed storage engine](/durable-objects/best-practices/access-durable-objects-storage/#sqlite-storage-backend), the `txn` object is obsolete. Any storage operations performed directly on the `ctx.storage` object, including SQL queries using [`ctx.storage.sql.exec()`](#sqlexec), will be considered part of the transaction.
+
+### transactionSync
+
```
vy-ton, line 199: Can we add a working example for reference?

Oxyjun (approved):

kentonv:
I added an example but in retrospect it's a bad example because as written it doesn't actually need transactionSync(): we automatically guarantee atomicity of queries performed with no await in between.

I need a different example where the thrown exception actually rolls something back. That's the part you can only do with transactionSync().

kentonv:
I reverted the example, so this is back to what was reviewed and approved previously. I'm just going to merge this as-is without an example rather than let it be delayed further.

kentonv:
For the record this was my example: 3718c36b2a87e093f8b75e8893fb6e46703531a6

But again, it's a bad example because it doesn't actually benefit from the use of `transactionSync()` -- the code would work fine without it, by just relying on automatic write coalescing in DO.

PR #17735, opened 2024-10-23
[ZT] Update WARP changelog

ranbel:
PCX-14355

### DIFF
--- a/src/content/changelogs/warp.yaml
+++ b/src/content/changelogs/warp.yaml
@@ -5,6 +5,36 @@ productLink: "/cloudflare-one/connections/connect-devices/warp/"
 productArea: Cloudflare One
 productAreaLink: /cloudflare-one/changelog/
 entries:
+  - publish_date: "2024-10-23"
+    title: WARP client for macOS (version 2024.10.279.1)
+    description: |-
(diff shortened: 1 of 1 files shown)


src/content/changelogs/warp.yaml
```diff
@@ -5,6 +5,35 @@ productLink: "/cloudflare-one/connections/connect-devices/warp/"
...
+        - Fixed an issue where SSH sessions and other application connections over TCP or UDP could drop when a device that is using MASQUE changes its primary network interface.
+        - Fixed an issue to ensure the Cloudflare root certificate (or custom certificate) is installed in the trust store if not already there.
+
+      Known issues:
+        - Cloudflare is investigating temporary networking issues on macOS 15 (Sequoia) that affect some users and may occur on any version of the WARP client.
+
+  - publish_date: "2024-10-23"
+    title: WARP client for Windows (version 2024.10.279.1)
+    description: |-
+      A new beta release for the Windows WARP client is now available in the [App Center](https://install.appcenter.ms/orgs/cloudflare/apps/1.1.1.1-windows/distribution_groups/beta). This release contains minor fixes and improvements.
```
dcpena, line 23: Is it intentional that this section and the date are the same as the entry above it? Looks like the only difference between the two is the Known Issues?

ranbel: Yes, most of the updates and the version numbers are the same. App Center link and known issues are different.

dcpena (approved):

PR #17770, opened 2024-10-24
script_name

elithrar:
Clarify how `script_name` can be used.

### DIFF
--- a/src/content/docs/workflows/build/workers-api.mdx
+++ b/src/content/docs/workflows/build/workers-api.mdx
@@ -116,10 +116,34 @@ compatibility_date = "2024-10-22"
 [[workflows]]
 # name of your workflow
 name = "workflows-starter"
-# binding name env.MYWORKFLOW
+# binding name env.MY_WORKFLOW
 binding = "MY_WORKFLOW"
 # this is class that extends the Workflow class in src/index.ts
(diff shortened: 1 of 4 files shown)


LuisDuarte1 (approved):

src/content/docs/workflows/build/workers-api.mdx
```diff
@@ -116,12 +116,36 @@ compatibility_date = "2024-10-22"
...
-# binding name env.MYWORKFLOW
+# binding name env.MY_WORKFLOW
 binding = "MY_WORKFLOW"
 # this is class that extends the Workflow class in src/index.ts
 class_name = "MyWorkflow"
 ```
 
+### Cross-script calls
+
+You can also bind to a Workflow that is defined in a different Worker script from the script your Workflow definition is in. To do this, provide the `script_name` key with the name of the script to the `[[workflows]]` binding definition in your `wrangler.toml` configuration.
```
Oxyjun, line 127:
```suggestion
You can also bind to a Workflow that is defined in a different Worker script. To do this, provide the `script_name` key with the name of the script to the `[[workflows]]` binding definition in your `wrangler.toml` configuration.
```

Oxyjun (approved):

PR #17829, opened 2024-10-28
[Calls] Added changelog file and entries

dcpena:
Added changelog file and entries to address PCX-13087.

### DIFF
--- a/src/content/changelogs/calls.yaml
+++ b/src/content/changelogs/calls.yaml
@@ -10,9 +10,13 @@ entries:
     description: |-
       Cloudflare Calls TURN service is generally available and helps address common challenges with real-time communication. For more information, refer to the [blog post](https://blog.cloudflare.com/webrtc-turn-using-anycast/) or [TURN documentation](/calls/turn/).
   - publish_date: "2024-04-04"
-    title: Cloudflare Calls open beta and Orange Meets availability
+    title: Orange Meets availability
     description: |-
-      Cloudflare Calls is in open beta and available from the Cloudflare Dashboard. Orange Meets, Cloudflare's internal video conferencing app, is open source and available for use from [Github](https://github.com/cloudflare/orange?cf_target_id=40DF7321015C5928F9359DD01303E8C2). 
+      Orange Meets, Cloudflare's internal video conferencing app, is open source and available for use from [Github](https://github.com/cloudflare/orange?cf_target_id=40DF7321015C5928F9359DD01303E8C2). 
+  - publish_date: "2024-04-04"
+    title: Cloudflare Calls open beta
+    description: |-
+      Cloudflare Calls is in open beta and available from the Cloudflare Dashboard.
   - publish_date: "2022-09-27"
     title: Cloudflare Calls closed beta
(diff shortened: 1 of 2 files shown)


src/content/changelogs/calls.yaml
```diff
@@ -0,0 +1,19 @@
...
+link: "/calls/changelog/"
+productName: Calls
+productLink: "/calls/"
+productArea: Developer platform
+productAreaLink: /workers/platform/changelog/platform/
+entries:
+  - publish_date: "2024-09-25"
+    title: TURN service is generally available (GA)
+    description: |-
+      Cloudflare Calls TURN service is generally available and helps address common challenges with real-time communication such connectivity, privacy, performance, and scalability. For more information, refer to the [blog post](https://blog.cloudflare.com/webrtc-turn-using-anycast/) or [TURN documentation](/calls/turn/).
```
nils-ohlmeier: I would prefer if we limit the list of attributes to "such as connectivity and privacy".

It depends on what we are attributing here. Any real-time communication which works without TURN has better performance. If we compare our TURN service to other TURN services, then yes our scales better and should be better in performance. But that distinction is not very clear to me here. Thus I would suggest to be a little bit more humble.

renandincer:
```suggestion
      Cloudflare Calls TURN service is generally available and helps address common challenges with real-time communication. For more information, refer to the [blog post](https://blog.cloudflare.com/webrtc-turn-using-anycast/) or [TURN documentation](/calls/turn/).
```

src/content/changelogs/calls.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - publish_date: "2024-04-04"
+    title: Cloudflare Calls open beta and Orange Meets availability
```
renandincer: Should these be two separate points on the same day?

dcpena: Done!

nils-ohlmeier (approved):

patriciasantaana (approved):

src/content/changelogs/calls.yaml
```diff
@@ -0,0 +1,23 @@
...
+    description: |-
+      Cloudflare Calls TURN service is generally available and helps address common challenges with real-time communication. For more information, refer to the [blog post](https://blog.cloudflare.com/webrtc-turn-using-anycast/) or [TURN documentation](/calls/turn/).
+  - publish_date: "2024-04-04"
+    title: Orange Meets availability
+    description: |-
+      Orange Meets, Cloudflare's internal video conferencing app, is open source and available for use from [Github](https://github.com/cloudflare/orange?cf_target_id=40DF7321015C5928F9359DD01303E8C2). 
+  - publish_date: "2024-04-04"
+    title: Cloudflare Calls open beta
+    description: |-
+      Cloudflare Calls is in open beta and available from the Cloudflare Dashboard.
```
patriciasantaana, line 19:
```suggestion
      Cloudflare Calls is in open beta and available from the Cloudflare dashboard.
```

PR #17895, opened 2024-10-31
[ZT] The max number of Egress policies per account

kyouheicf:
PCX-14475


maxvp:
Hi @kyouheicf, I added egress and resolver policy limits with #17888.

PR #17926, opened 2024-11-01
Added docs for cache: no-store

tewaro:
### Summary

Update Documentation for adding `cache: no-store`.

### Documentation checklist

### DIFF
--- a/src/content/compatibility-dates/cache-no-store.md
+++ b/src/content/compatibility-dates/cache-no-store.md
@@ -1,0 +1,39 @@
+---
+_build:
+	publishResources: false
+	render: never
+	list: never
+
+name: "Enable `cache: no-store` HTTP standard API"
+sort_date: "2024-11-11"
+enabled_date: "2024-11-11"
+enable_flag: "cache_option_enabled"
+disable_flag: "cache_option_disabled"
+---
+
+When you enable the `cache_option_enabled` compatibility flag, you can specify a value for the `cache` property of the Request interface.
+When this compatibility flag is not enabled, or `cache_option_disabled` is set, the Workers runtime will throw an `Error` saying `The 'cache' field on
+'RequestInitializerDict' is not implemented.`
+
+When this flag is enabled you can instruct Cloudflare not to cache the response from a subrequest you make from your Worker using the [`fetch()` API](/workers/runtime-apis/fetch/):
+
+The only cache option enabled with `cache_option_enabled` is `'no-store'`.
+Specifying any other value will cause the Workers runtime to throw a `TypeError` with the message `Unsupported cache mode: <the-mode-you-specified>`.
+
+When `no-store` is specified:
+* All requests have the headers `Pragma: no-cache` and `Cache-Control: no-cache` are set on them.
+
+* Subrequests to origins not hosted by Cloudflare bypasses Cloudflare's cache.
+
+Examples using `cache: 'no-store'`:
+
+```js
+const response = await fetch("https://example.com", { cache: 'no-store'});
+```
+
+The cache value can also be set on a `Request` object.
+
+```js
+const request = new Request("https://example.com", { cache: 'no-store'});
+const response = await fetch(request);
+```
--- a/src/content/docs/workers/examples/cache-using-fetch.mdx
+++ b/src/content/docs/workers/examples/cache-using-fetch.mdx
@@ -466,3 +466,13 @@ async function handleRequest(request) {
 ```
 
 </TabItem> </Tabs>
+
+## Using the HTTP Cache API
+
+The `cache` mode can be set in `fetch` options.
+Currently Workers only support the `no-store` mode for controlling the cache.
+When `no-store` is supplied the cache is bypassed on the way to the origin and the request is not cacheable.
+
+```js
+fetch(request, { cache: 'no-store'});
+```
--- a/src/content/docs/workers/runtime-apis/fetch.mdx
+++ b/src/content/docs/workers/runtime-apis/fetch.mdx
@@ -75,9 +75,12 @@ async function eventHandler(event) {
 * [`resource`](https://developer.mozilla.org/en-US/docs/Web/API/fetch#resource) Request | string | URL
 
 * `options` options
+	* `cache` `undefined | 'no-store'` optional
+		* Standard HTTP `cache` header. Only `cache: 'no-store'` is supported.
+		Any other `cache` header will result in a `TypeError` with the message `Unsupported cache mode: <attempted-cache-mode>`.
+			* For all requests this forwards the `Pragma: no-cache` and `Cache-Control: no-cache` headers to the origin.
+			* For requests to origins not hosted by Cloudflare, `no-store` bypasses the use of Cloudflare's caches.
   * An object that defines the content and behavior of the request.
-
-
 
 ***
 
--- a/src/content/docs/workers/runtime-apis/request.mdx
+++ b/src/content/docs/workers/runtime-apis/request.mdx
@@ -62,7 +62,10 @@ let request = new Request(input, options)
 
 An object containing properties that you want to apply to the request.
 
+* `cache` `undefined | 'no-store'` optional
 
+	* Standard HTTP `cache` header. Only `cache: 'no-store'` is supported.
+	Any other cache header will result in a `TypeError` with the message `Unsupported cache mode: <attempted-cache-mode>`.
 
 * `cf` RequestInitCfProperties optional
 
@@ -180,7 +183,7 @@ All properties of an incoming `Request` object (the request you receive from the
 
   :::caution
 
-  If the response is a redirect and the redirect mode is set to `follow` (see below), then all headers will be forwarded to the redirect destination, even if the destination is a different hostname or domain. This includes sensitive headers like `Cookie`, `Authorization`, or any application-specific headers. If this is not the behavior you want, you should set redirect mode to `manual` and implement your own redirect policy. Note that redirect mode defaults to `manual` for requests that originated from the Worker's client, so this warning only applies to `fetch()`es made by a Worker that are not proxying the original request. 
+  If the response is a redirect and the redirect mode is set to `follow` (see below), then all headers will be forwarded to the redirect destination, even if the destination is a different hostname or domain. This includes sensitive headers like `Cookie`, `Authorization`, or any application-specific headers. If this is not the behavior you want, you should set redirect mode to `manual` and implement your own redirect policy. Note that redirect mode defaults to `manual` for requests that originated from the Worker's client, so this warning only applies to `fetch()`es made by a Worker that are not proxying the original request.
   :::
 
 * `method` string read-only


src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,24 @@
...
+When this flag is enabled, the cache header becomes available within the [`fetch()` API](/workers/runtime-apis/fetch/). The option for `no-store` is enabled.
+
+For a cross-zone request this simply specifies the headers `Pragma: no-cache` and `Cache-Control: no-cache` to the origin.
+
+For grey-cloud subrequests this bypasses the Cloudflare's cache.
+
+Example:
+
+```js
+const response = await fetch("https://example.com", { cache: 'no-store'});
```
mhart, line 31: Suggestion: add an example for `new Request()` too? We support it there too right?

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,24 @@
...
+	list: never
+
+name: "Enable `cache: no-store` HTTP standard API"
+sort_date: "2024-11-11"
+enabled_date: "2024-11-11"
+enable_flag: "cache_option_enabled"
+disable_flag: "cache_option_disabled"
+---
+
+When this flag is enabled, the cache header becomes available within the [`fetch()` API](/workers/runtime-apis/fetch/). The option for `no-store` is enabled.
```
irvinebroque: What this enables is this:

https://developer.mozilla.org/en-US/docs/Web/API/Request/cache

The `cache` property of `Request`.

I see where you're going describing it in terms of headers, but "the header becomes available" not quite right.

We should be precise here and elsewhere:

```suggestion
When you enable the `cache_option_enabled` compatibility flag, you can specify a value for the `cache` property of the `Request` interface. When this compatibility flag is not enabled, or `cache_option_disabled` is set, the Workers runtime will throw an error if you attempt to specify the `cache` property.

For example, when this flag is enabled you can instruct Cloudflare not to cache the response from a subrequest you make from your Worker, using the [`fetch()` API](/workers/runtime-apis/fetch/):

<example using no-store>

...then show how this maps to headers, as you do below on line 16
```

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,15 @@ async function handleRequest(request) {
 ```
 
 </TabItem> </Tabs>
+
+## Using the HTTP Cache API
+
+Currently we only support the `no-store` header for controlling cache.
```
irvinebroque: This is out of context, need to back up here

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,15 @@ async function handleRequest(request) {
...
+When `no-store` is supplied the cache is bypassed both on the way to the origin and on the way back.
+For cross-zone requests this simply forwards the `Pragma: no-cache` and `Cache-Control: no-cache` headers to the origin.
+This is because the cache is on the origin side for cross-zone requests.
+For grey-cloud requests this bypasses the use of Cloudflare's caches.
+
+```js
+fetch(event.request, { cache: 'no-store'});
```
irvinebroque: Why does this use `event.request`? That seems like leftover service worker thing?

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,15 @@ async function handleRequest(request) {
 ```
 
 </TabItem> </Tabs>
+
+## Using the HTTP Cache API
+
+Currently we only support the `no-store` header for controlling cache.
+When `no-store` is supplied the cache is bypassed both on the way to the origin and on the way back.
+For cross-zone requests this simply forwards the `Pragma: no-cache` and `Cache-Control: no-cache` headers to the origin.
```
irvinebroque: This example content shouldn't be our reference material here.

This kind of content should be here:

https://developers.cloudflare.com/workers/runtime-apis/fetch/

src/content/docs/workers/runtime-apis/request.mdx
```diff
@@ -62,7 +62,9 @@ let request = new Request(input, options)
 
 An object containing properties that you want to apply to the request.
 
+* `cache` `undefined | 'no-store'` optional
 
+	* Standard HTTP `cache` header. Only `cache: no-store` is supported.
```
irvinebroque: Can we explain what the behavior here is if you attempt to set a value other than `no-store`?

irvinebroque (commented):
@mikenomitch FYI

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,24 @@
...
+
+For a cross-zone request this simply specifies the headers `Pragma: no-cache` and `Cache-Control: no-cache` to the origin.
+
+For grey-cloud subrequests this bypasses the Cloudflare's cache.
```
mikenomitch: We sometimes use grey cloud in external docs, but for the most part I think we shouldn't. We know what it means internally, but average new user won't.

src/content/compatibility-dates/cache-no-store.md
mikenomitch: +1 to the wording change. Was a bit hard to follow exactly what you can now do originally.

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,24 @@
...
+name: "Enable `cache: no-store` HTTP standard API"
+sort_date: "2024-11-11"
+enabled_date: "2024-11-11"
+enable_flag: "cache_option_enabled"
+disable_flag: "cache_option_disabled"
+---
+
+When this flag is enabled, the cache header becomes available within the [`fetch()` API](/workers/runtime-apis/fetch/). The option for `no-store` is enabled.
+
+For a cross-zone request this simply specifies the headers `Pragma: no-cache` and `Cache-Control: no-cache` to the origin.
```
mikenomitch: style nitpick: I'd replace "simply" with "just" or just drop the word.

mikenomitch: Also, I think this reads slightly off "specifies headers to the origin" - specifies headers in the request to the origin?

src/content/docs/workers/examples/cache-using-fetch.mdx
mikenomitch: +1

Maybe you start by explaining how the cache normally works (Brendan's next comment), and then explain the header?

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,15 @@ async function handleRequest(request) {
...
+This is because the cache is on the origin side for cross-zone requests.
+For grey-cloud requests this bypasses the use of Cloudflare's caches.
```
mikenomitch: Noting the other "grey-cloud" use (see above)

src/content/compatibility-dates/cache-no-store.md
tewaro: do we want to do the same with cross-zone?

src/content/docs/workers/examples/cache-using-fetch.mdx
tewaro: Same question here about cross-zone.

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+name: "Enable `cache: no-store` HTTP standard API"
+sort_date: "2024-11-11"
+enabled_date: "2024-11-11"
+enable_flag: "cache_option_enabled"
+disable_flag: "cache_option_disabled"
+---
+
+When you enable the `cache_option_enabled` compatibility flag, you can specify a value for the `cache` property of the Request interface.
+When this compatibility flag is not enabled, or `cache_option_disabled` is set, the Workers runtime will throw an `Error` saying `The 'cache' field on
+'RequestInitializerDict is not implemented.`
```
mikenomitch: nit - I think missing a ' after RequestInitializerDict (or added one before it inadvertantly

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+Examples using `cache: 'no-store'`.
+
+```js
+const response = await fetch("https://example.com", { cache: 'no-store'});
+```
+
+You can also create a Response object to pass around:
+
+```js
+const request = new Response("https://example.com", { cache: 'no-store'});
```
mhart: This should be `Request`, not `Response`

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+enabled_date: "2024-11-11"
+enable_flag: "cache_option_enabled"
+disable_flag: "cache_option_disabled"
+---
+
+When you enable the `cache_option_enabled` compatibility flag, you can specify a value for the `cache` property of the Request interface.
+When this compatibility flag is not enabled, or `cache_option_disabled` is set, the Workers runtime will throw an `Error` saying `The 'cache' field on
+'RequestInitializerDict is not implemented.`
+
+For example, when this flag is enabled you can instruct Cloudflare not to cache the response from a subrequest you make from your Worker, using the [`fetch()` API](/workers/runtime-apis/fetch/):
```
mikenomitch: The "For example" feels a bit odd here because that's more or less the only thing you can do at this point right. Maybe just drop that?

And then add comma after "when this flag is enabled"

Also, I think the comma before "using" can be dropped?

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+
+The only cache option enabled with `cache_option_enabled` is `'no-store'`.
+Any other value will cause the Workers runtime to throw a `TypeError` with the message `Unsupported cache mode: <the-mode-you-specified>`.
```
mikenomitch: nit - maybe: "Specifying any other value will throw a throw a `TypeError` with the message `Unsupported cache mode: <the-mode-you-specified>`."

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+
+When `no-store` is specified:
+* For a cross-zone request the headers `Pragma: no-cache` and `Cache-Control: no-cache` are sent to the origin.
```
mikenomitch: I try to make the bullet points read well with the header

If you read it like "When no-store is specified for a cross zone", the "for" sounds kind of weird

So I'd do:

When `no-store` is specified:
* All requests have the headers `Pragma: no-cache` and `Cache-Control: no-cache` set
* Subrequests to origins outside of Cloudflare bypass Cloudflare's cache

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+
+* For subrequests to origins not hosted by Cloudflare, `no-store` bypasses Cloudflare's cache in addition to specifying the above headers.
+
+Examples using `cache: 'no-store'`.
+
+```js
+const response = await fetch("https://example.com", { cache: 'no-store'});
+```
+
+You can also create a Response object to pass around:
```
mikenomitch: Maybe just: "The cache value can also be set on Request"

"to pass around" isn't doing much here

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+Any other value will cause the Workers runtime to throw a `TypeError` with the message `Unsupported cache mode: <the-mode-you-specified>`.
+
+When `no-store` is specified:
+* For a cross-zone request the headers `Pragma: no-cache` and `Cache-Control: no-cache` are sent to the origin.
+
+* For subrequests to origins not hosted by Cloudflare, `no-store` bypasses Cloudflare's cache in addition to specifying the above headers.
+
+Examples using `cache: 'no-store'`.
+
+```js
```
mikenomitch, line 30: Maybe add a little text about this one specifically, so it's similar to to the one below.

Maybe just: "The 'cache' have can be passed into the options for `fetch`"

src/content/compatibility-dates/cache-no-store.md
```diff
@@ -0,0 +1,39 @@
...
+
+The only cache option enabled with `cache_option_enabled` is `'no-store'`.
+Any other value will cause the Workers runtime to throw a `TypeError` with the message `Unsupported cache mode: <the-mode-you-specified>`.
+
+When `no-store` is specified:
+* For a cross-zone request the headers `Pragma: no-cache` and `Cache-Control: no-cache` are sent to the origin.
+
+* For subrequests to origins not hosted by Cloudflare, `no-store` bypasses Cloudflare's cache in addition to specifying the above headers.
+
+Examples using `cache: 'no-store'`.
```
mikenomitch: This isn't a sentence, so we shouldn't have a period. Maybe our lowest lever of header would render well? Or just replace the period with a ":"?

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,13 @@ async function handleRequest(request) {
 ```
 
 </TabItem> </Tabs>
+
+## Using the HTTP Cache API
+
+The `cache` mode can be set on fetches.
```
mikenomitch: "can be set in `fetch` options" maybe? - A little more specific

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,13 @@ async function handleRequest(request) {
...
+Currently we only support the `no-store` mode for controlling the cache.
```
mikenomitch: s/we only support/Workers only supports

src/content/docs/workers/examples/cache-using-fetch.mdx
```diff
@@ -466,3 +466,13 @@ async function handleRequest(request) {
...
+When `no-store` is supplied the cache is bypassed both on the way to the origin and on the way back.
```
mikenomitch: What does "on the way back" mean in this case?

src/content/docs/workers/runtime-apis/fetch.mdx
```diff
@@ -75,6 +75,12 @@ async function eventHandler(event) {
 * [`resource`](https://developer.mozilla.org/en-US/docs/Web/API/fetch#resource) Request | string | URL
 
 * `options` options
+	* `cache` `undefined | 'no-store'` optional
+		* Standard HTTP `cache` header. Only `cache: 'no-store'` is supported.
+		Any other `cache` header will result in a `TypeError` with the message `Unsupported cache mode: <attempted-cache-mode>`.
+			* For cross-zone requests this simply forwards the `Pragma: no-cache` and `Cache-Control: no-cache` headers to the origin.
+			This is because the cache is on the origin side for cross-zone requests.
+			* For requests to origins not hosted by Cloudflare, `no-store` bypasses the use of Cloudflare's caches.
```
mikenomitch, line 82: Its a bit unclear reading this if non-Cloudflare get the headers or not. Reading above it sounds like it does.

mikenomitch (changes requested):

src/content/docs/workers/runtime-apis/fetch.mdx
tewaro: They get the headers.

tewaro: I will fix the wording.

mikenomitch (approved):
LGTM now - thanks for making the changes

PR #17936, opened 2024-11-01
[ZT] Terraform examples (device enrollment, infra access)

ranbel:
PCX-6717

### DIFF
--- a/src/content/partials/cloudflare-one/access/create-service-token.mdx
+++ b/src/content/partials/cloudflare-one/access/create-service-token.mdx
@@ -57,10 +57,10 @@ import { Tabs, TabItem } from '~/components';
 
 5. Read the Client ID and Client Secret:
 		```sh
-		terraform output example_service_token_client_id
+		terraform output -raw example_service_token_client_id
 		```
 		```sh
-		terraform output example_service_token_client_secret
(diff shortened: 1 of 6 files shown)


src/content/partials/cloudflare-one/access/create-service-token.mdx
```diff
@@ -16,5 +20,47 @@
...
+		```sh
+		terraform apply
+		```
+
+5. Read the Client ID and Client Secret:
+		```sh
+		terraform output example_service_token_client_id
+		```
+		```sh
+		terraform output example_service_token_client_secret
```
jacobbednarz: minor fix here to include `-raw` otherwise the value is quoted (for strings) which will throw off templating or replacement if it is used.

before

```
$ terraform output token
"c001646ec59187c602b3713fce272b12ea7f84d7b40017bf224cade6ca985ce8"
```

after

```
$ terraform output token
c001646ec59187c602b3713fce272b12ea7f84d7b40017bf224cade6ca985ce8
```

~~~suggestion
		terraform output -raw example_service_token_client_id
		```
		```sh
		terraform output -raw example_service_token_client_secret
~~~

daisyfaithauma (approved):
LGTM

PR #17952, opened 2024-11-04
fix node:crypto API support

vicb:
The `node:crypto` keys API are not available.

While those exist, they would throw an `ERR_METHOD_NOT_IMPLEMENTED` error.

Note that we have plans to implement them in the future, see https://github.com/cloudflare/workerd/issues/3009 for more details.

Most diffs in the PR are due to the formatting fix, the actual changes are l80+

/cc @jasnell


GregBrimble (approved):

vicb:
Thanks Greg

PR #18004, opened 2024-11-05
Update caching.mdx

kathayl:
caching updates

### Documentation checklist


kathayl:
I've updated the page with 
1) the new updated headers
2) the minimum time to live of 60 seconds and max of 1 month
3) changed the example from milliseconds to seconds

Daisy, can you 
1) figure out how the best way to say that the old headers still work but we have new names. While the old headers will still work, we recommend updating to the new headers to ensure compatibility with future releases:
- cf-cache-ttl has been replaced by cf-aig-cache-ttl
- cf-skip-cache has been replaced by cf-aig-skip-cache

2) need to add this note in about caching behavior
_Note: Cache in AI Gateway is volatile. If two identical requests are sent simultaneously, the first request may not cache in time for the second to use it, causing the second request to retrieve data from the original source._

src/content/docs/ai-gateway/reference/limits.mdx
```diff
@@ -11,6 +11,22 @@ The following limits apply to gateway configurations, logs, and related features
...
+| Maximum Logpush jobs                                            | 4 per account                       |
+| Maximum gateway name length                                     | 64 characters                       |
+
+<sup>1</sup> If you have reached 10 million logs stored per gateway, new logs
+will stop being saved. To continue saving logs, you must delete older logs in
+that gateway to free up space or create a new gateway.
+
+<sup>2</sup> If you have reached 100,000 logs stored per account, accross all
+gateways, new logs will stop being saved. To continue saving logs, you must
+Delete older logs.
```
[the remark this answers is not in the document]
pedrosousa: The merge conflict marker is actually in the next line (29).
FYI: @bllchmbrs - Missed by one line, and that's why the explanation mentions that there's no visible conflict marker.

src/content/docs/ai-gateway/reference/limits.mdx
```diff
@@ -31,6 +47,7 @@ gateways, new logs will stop being saved. To continue saving logs, you must
 delete older logs. Refer to [Auto Log
 Cleanup](/ai-gateway/observability/logging/#auto-log-cleanup) for more details
 on how to automatically delete logs.
+>>>>>>> 914c70d6b867c652d10b92bd7fb44a284bafacea
```
pedrosousa: This line should also be removed.

src/content/docs/ai-gateway/reference/limits.mdx
```diff
@@ -11,6 +11,22 @@ The following limits apply to gateway configurations, logs, and related features
 
 | Feature                                                         | Limit                               |
 | --------------------------------------------------------------- | ----------------------------------- |
+<<<<<<< HEAD
```
pedrosousa: You must remove this line.

daisyfaithauma (approved):

PR #18115, opened 2024-11-12
Update managing-dynamic-ip-addresses.mdx

SunsetQuest:
Remove 'DNS-O-Matic' as it is no longer available and add 'Dynamic-DNS-Record-Updater-For-Cloudflare'.

This was a Dynamic DNS updater I created because there is nothing for Windows anymore.  The existing item is no longer available for download as it was purchased by another company.  I think this will make people's lives easier - at least windows users.   Thank you.

Here is a possible replacement tool
https://github.com/SunsetQuest/Dynamic-DNS-Record-Updater-For-CloudFlare/blob/master/README.md


RebeccaTamachiro:
Hi @SunsetQuest 👋 Thanks for reaching out.
Unfortunately, I will not be able to list your updater in our docs, but you can share it with other Cloudflare users via our [Community](https://community.cloudflare.com/).

SunsetQuest:
@RebeccaTamachiro - No concerns. It's advisable to update the page on managing dynamic IP addresses at Cloudflare's developer documentation. Currently, there is no Dynamic DNS updater available for Windows users without installing Perl. Additionally, the DNS-O-Matic service is no longer accessible.

PR #18125, opened 2024-11-12
remove incorrect information

DonEstefan:
### Summary

The "use-application-dns.net" canary domain is only used to disable DNS over HTTPS (DoH), not to disable ECH (see [documentation](https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet)). Originally, Firefox required DoH to be enabled in order for ECH to function. Since Firefox 129, ECH is enabled even if DoH is disabled (see [here](https://wiki.mozilla.org/Security/Encrypted_Client_Hello#Dependency_on_DoH ))

Other browsers have always ignored the canary domain (see [here](https://www.chromium.org/developers/dns-over-https/#faq)).


angelampcosta:
hi @DonEstefan This is being tracked internally under a ticket we’ll be working from. Closing this out since the work will be handled elsewhere.

PR #18133, opened 2024-11-12
Product Directory Refresh

hturan:
### Summary
This PR gives a visual and functional refresh to the Products page (https://developers.cloudflare.com/products), including iconography, product descriptions, and the ability to search.

### Screenshots (optional)
Screenshot 2024-11-12 at 15 43 53
Screenshot 2024-11-12 at 15 44 02
Screenshot 2024-11-12 at 15 44 16
Screenshot 2024-11-12 at 15 45 29
Screenshot 2024-11-12 at 18 31 27

### Documentation checklist


src/components/ProductCatalog.tsx
hturan, line 1: Lots of room for us to de-dupe this and `ModelCatalog` in the future. Left it be for now until we tackle a third instance of this pattern.

src/pages/products.astro
```diff
@@ -1,181 +1,49 @@
 ---
-import StarlightPage from "@astrojs/starlight/components/StarlightPage.astro";
 import { getCollection, type CollectionEntry } from "astro:content";
-import { ThreeCardGrid, LinkCard } from "~/components";
+// @ts-ignore virtual module
+import iconCollection from "virtual:astro-icon";
```
hturan, line 4: I got 90% of the way through this work and then realised we have no good way to render icon / SVGs in `react`, and `astro-icon` does some interesting integration/optmimization under the hood. This code here replicates a little of what the `Icon` component does (https://github.com/natemoo-re/astro-icon/blob/531db17ec5367e22df7958d8f7abdeb880d8342b/packages/core/components/Icon.astro#L92) but just passes optimised SVG path data (as a string) through to React to render.

hturan: I originally had this in the `jsx` component, but due to how Astro works it was trying to import `node` modules into the client-side and breaking.

hturan: I also tried directly importing the icons (using `import` URLs), but you can't style SVG when used as an `img` `src`.

KimJ15 (approved):

src/pages/products.astro
```diff
@@ -1,181 +1,49 @@
...
-	}
-
-	getGroupFilters()?.forEach((input) => {
-		input.addEventListener("change", filterProducts);
-	});
-
-	function initProducts() {
-		const params = new URLSearchParams(window.location.search);
-
-		const groups = params.get("product-group");
```
KianNH, line 164: It looks like we're missing this functionality now - we rely on it on the homepage using links like `/products/?product-group=Developer+platform`

src/components/index.ts
```diff
@@ -48,7 +48,6 @@ export { default as ResourcesBySelector } from "./ResourcesBySelector.astro";
 export { default as RuleID } from "./RuleID.astro";
 export { default as SpotlightAuthorDetails } from "./SpotlightAuthorDetails.astro";
 export { default as Stream } from "./Stream.astro";
-export { default as ThreeCardGrid } from "./ThreeCardGrid.astro";
```
KianNH, line 51: Can we delete the component file itself too? I think the products grid was the only usage of it.

src/pages/products.astro
```diff
@@ -1,181 +1,49 @@
...
-									href={product.data.product.url}
-								/>
-							</div>
-						);
-					})
-				}
-			</ThreeCardGrid>
-		</div>
-	</div>
+	<ProductCatalog products={products} client:load />
```
KianNH: CI is failing due to a type error on `products` here: https://github.com/cloudflare/cloudflare-docs/actions/runs/11820899830/job/32934322983?pr=18133#step:6:46

KianNH (changes requested):
Awesome work, the styling is perfect! This looks great.

I found a few small issues which are hopefully quick fixes.

src/pages/products.astro
hturan: Ah, sorry! Totally missed this, my bad. I'll add it via stetting initial state in React.

src/pages/products.astro
hturan: Thanks! I've made a more specific `ProductData` type for the `ProductCatalog` (including icons and groups).

KianNH (approved):
❤️

PR #18155, opened 2024-11-13
[ZT] - Give More information into how to format tokens

rderewianko:
### Summary
When using a service token, CloudFlare Zero Trust formats the token to include  
`CF-Access-Client-Id: `
`CF-Access-Client-Secret: `

If you directly paste these into a Cloudflare Warp `mdm.xml` it will fail as a invalid token. 

I've added a bit of clarity to this to describe what the value expects.

### Documentation checklist

- [ x] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.


ranbel:
Thanks for the suggestion! I swapped the notes for a concrete example. `CF-Access-Client-Id:` and `CF-Access-Client-Secret: ` don't appear if you create the token using the API/Terraform.

ranbel (approved):

rderewianko:
Great thank you!

I'd suspect most Endpoint engineers aren't using terraform/api to get the token when they're first starting a deployment.

PR #18159, opened 2024-11-13
[ZT] Refresh JumpCloud IdP

ranbel:
PCX-10685


kennyj42 (approved):
assuming we're doing SCIM stuff next?

ranbel:
> assuming we're doing SCIM stuff next?

Yup, SCIM will be a separate PR.

patriciasantaana (approved):

PR #18218, opened 2024-11-15
Updates pricing and limits in Workers Static Assets

korinne:
Includes updates to pricing information for storing assets on Workers, as well as file size limits


WalshyDev:
We should probably make an Assets section in the limits page and add there - https://developers.cloudflare.com/workers/platform/limits/

korinne:
@WalshyDev Good idea! I'll make an update to the Limits page in a separate PR, but still want to have something in the Assets-specific docs.

WalshyDev (approved):

GregBrimble (approved):

PR #18262, opened 2024-11-19
add Workers AI footnote, until we have proper explanations

kbitr:
add Workers AI footnote, until we have proper explanations


angelampcosta:
Done in https://github.com/cloudflare/cloudflare-docs/pull/18387

PR #18284, opened 2024-11-20
Add global flags to wrangler commands

vicb:


vicb:
@petebacondarwin @penalosa This PR adds the global flags to [all the commands](https://common-flags.cloudflare-docs-7ou.pages.dev/workers/wrangler/commands/) to make them easily discoverable.

src/content/partials/workers/wrangler-commands/hyperdrive.mdx
```diff
@@ -39,6 +39,10 @@ wrangler hyperdrive create <CONFIG_NAME> [OPTIONS]
 - `--swr` <Type text="number" /> <MetaInfo text="optional" />
   - Stale While Revalidate - Indicates the number of seconds cache may serve the response after it becomes stale, cannot be set when caching is disabled.
 
+The wrangler global flags can also be used:
```
petebacondarwin: Perhaps add this sentence to the reusable block too? I think it is repeated in every case.

src/content/partials/workers/wrangler-commands/hyperdrive.mdx
```diff
@@ -39,6 +39,10 @@ wrangler hyperdrive create <CONFIG_NAME> [OPTIONS]
...
+
+<Render file="wrangler-commands/global-flags" product="workers" />
```
petebacondarwin, line 42: Is `Render` the standard approach to doing this? I don't know much about these docs.

petebacondarwin (approved):

src/content/partials/workers/wrangler-commands/hyperdrive.mdx
vicb: I haven't include the sentence because it's not use in [the first Note](https://github.com/cloudflare/cloudflare-docs/pull/18284/files#diff-b3a0c1efcf8a9ecb6417993f4f5d3af693262c82e8e8c17cd50dc030028d4ad3L47) (describing the global flags) but we can come with a text that works everywhere.

src/content/partials/workers/wrangler-commands/hyperdrive.mdx
vicb: `Render` is how this file (`hypderdrive.mdx`) is rendered into `commands.mdx` so I think it's the right thing to do.

_(I first googled for including .mdx and a common approach seems to be transclusion but it needs to be enabled in the build and I don't think we do - at least we never `import from '*.mdx'`)_

petebacondarwin (approved):

vicb:
rebased

ToriLindsay (approved):

PR #18302, opened 2024-11-20
updating compatibility matrix

aninibread:
### Documentation checklist


irvinebroque (approved):
non-blocking but seems helpful to the reader if we have link to both the Pages docs for the functionality we are talking about, and the Workers functionality

(will also make be a forcing function to make sure we have right docs in place)

PR #18416, opened 2024-11-26
workflows: update type params docs

elithrar:
Updates:

- [x] Document the additional support for type params on the `Workflow` type per https://github.com/cloudflare/workerd/pull/3167
- [x] Add additional clarifying text that this does NOT validate the type itself and does not propagate the type information into the `WorkflowEvent`
- [x]  Reference the partial in both the events/types docs and the Workers API doc

### DIFF
--- a/src/content/docs/workflows/build/workers-api.mdx
+++ b/src/content/docs/workflows/build/workers-api.mdx
@@ -179,7 +179,7 @@ An ID is automatically generated, but a user-provided ID can be specified (up to
 // Create a new Workflow instance with your own ID and pass params to the Workflow instance
 let instance = await env.MY_WORKFLOW.create({
 	id: myIdDefinedFromOtherSystem,
-	params: { "hello": "world }
+	params: { "hello": "world" }
 })
 return Response.json({
 	id: instance.id,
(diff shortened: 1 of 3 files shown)


src/content/docs/workflows/build/workers-api.mdx
```diff
@@ -171,13 +171,16 @@ Create (trigger) a new instance of the given Workflow.
...
-An ID is automatically generated, but a user-provided ID can be specified (up to 64 characters [^1]). This can be useful when mapping Workflows to users, merchants or other identifiers in your system.
+An ID is automatically generated, but a user-provided ID can be specified (up to 64 characters [^1]). This can be useful when mapping Workflows to users, merchants or other identifiers in your system. You can also provide a JSON object as the `params` property, allowing you to pass data for the Workflow instance to act on as its [`WorkflowEvent`](/workflows/build/events-and-parameters/).
 
 ```ts
-// Create a new Workflow instance with your own ID:
-let instance = await env.MY_WORKFLOW.create({ id: myIdDefinedFromOtherSystem })
+// Create a new Workflow instance with your own ID and pass params to the Workflow instance
+let instance = await env.MY_WORKFLOW.create({
+	id: myIdDefinedFromOtherSystem,
+	params: { "hello": "world }
```
Skye-31:
```suggestion
	params: { "hello": "world" }
```

dcpena (approved):

PR #18470, opened 2024-11-28
[Rules] Update formatting of Bulk Redirects

pedrosousa:
### Summary

- Use `Type` and `MetaInfo` components.
- Partially switch from 1st-level bullets to headings to facilitate linking to a specific parameter.


src/content/docs/rules/url-forwarding/bulk-redirects/reference/parameters.mdx
```diff
@@ -5,54 +5,78 @@ sidebar:
...
+import { Type } from "~/components";
+
 A URL redirect has a source URL, a target URL, a status code, and some additional parameters that affect its URL matching behavior and runtime behavior.
 
-The URL redirect parameters are the following:
+## Source URL
+
+API field: `source_url` <Type text="String" />
+
+The URL string that the incoming request URL must match for the redirect to be applied. This property is mandatory. The maximum length of the source URL is 32 KB.
```
pedrosousa, line 16: The parameter descriptions did not change, except for some minor formatting and adding links to other parameters on the same page (which we can now do).

src/content/docs/rules/url-forwarding/bulk-redirects/reference/parameters.mdx
```diff
@@ -5,54 +5,78 @@ sidebar:
...
+
+The value must be a valid URL. The final target URL depends on the values of the [**Preserve query string**](#preserve-query-string) and [**Preserve path suffix**](#preserve-path-suffix) parameters.
+
+For more information on the supported URL components, refer to [Supported URL components](/rules/url-forwarding/bulk-redirects/reference/url-components/).
+
+## Subpath matching
+
+API field: `subpath_matching` <Type text="Boolean" />
+
+If `true`, the current redirect will apply the subpath matching algorithm to the request URL when determining if there is a match for the current URL redirect. The default value is `false`.
```
Oxyjun:
```suggestion
If `true`, the current redirect will apply the subpath matching algorithm to the request URL when determining if there is a match for the current URL redirect. The default value is `false`.
```
Nit: Personally, it may make sense for the default value to be specified higher up (maybe using the MetaInfo component?) I.e. a logical progression of "The default is false. But if true, then..."

Oxyjun (approved):

PR #18479, opened 2024-11-29
Allow building with pnpm

millette:
### Summary

Added a few dependencies (that were automatically pulled by npm) so the docs can be built with pnpm.


kodster28:
Closing this out as it's gotten stale (and hasn't come up from other users of the docs).

PR #18484, opened 2024-11-29
[Cache] Update cache.mdx

luishcardoso:
Hello team,

### Summary

Per the style guide. Links without '  ' within it.

Thanks!


pedrosousa:
Hi @nenizera 
Can I ask where you found this recommendation? 
Monospaced formatting is something we use often when linking to functions, fields, parameters, and properties in many different pages. 

If the style guide is recommending something different, we should review it.

ToriLindsay:
Hi @nenizera - I checked the style guide and wasn't able to find this either. Can you let us know where you see it? Thanks!

PR #18495, opened 2024-12-02
Add better messaging on Vitest Pool Workers and nodejs_compat

andyjessop:
### Summary

Improves the messaging surrounding `nodejs_compat` and Vitest Pool Workers as a result of [this PR](https://github.com/cloudflare/workers-sdk/pull/7388) which automatically injects `nodejs_compat` flag into the pool Worker.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/testing/vitest-integration/isolation-and-concurrency.mdx
+++ b/src/content/docs/workers/testing/vitest-integration/isolation-and-concurrency.mdx
@@ -63,7 +63,7 @@ The Workers Vitest pool works by running code inside a Cloudflare Worker that Vi
 
 Using Vitest Pool Workers may cause your Worker to behave differently when deployed than during testing as the `nodejs_compat` flag is enabled by default. This means that Node.js-specific APIs and modules are available when running your tests. However, Cloudflare Workers do not support these Node.js APIs in the production environment unless you specify this flag in your Worker configuration.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/testing/vitest-integration/isolation-and-concurrency.mdx
```diff
@@ -56,12 +56,12 @@ In this model, a single `workerd` process is started with a single Worker for al
...
 
 :::caution
 
 
-Using the pool may cause your Worker to behave differently when deployed than during testing, as Node-style resolution and additional polyfills will be available to your Worker's source code and dependencies too.
+Using Vitest Pool Workers may cause your Worker to behave differently when deployed than during testing as the `nodejs_compat` flag is enabled by default. This means that Node.js-specific APIs and modules are available when running your tests. However, Cloudflare Workers do not support these Node.js APIs in the production environment unless you specify this flag in your Worker configuration.
```
irvinebroque, line 64: Can we explicitly link people at this point to add the flag

And maybe give basic example? Like — in your tests, you'll see that your Worker's code can use xyz node.js API. But when you deploy, unless flag is enabled, this wont' work, so your tests can give you false positive.

Think this is really close but showing vs. telling will really help to clarify

andyjessop: Agreed, I'll make that update. Thanks for the input @irvinebroque 👍

andyjessop: I've added a more detailed explanation with code example.

irvinebroque (approved):

PR #18508, opened 2024-12-02
WC-2955 Add documentation for serve_directly usage

WillTaylorDev:
### Summary

Provides documentation for the serve_directly configuration option for Worker assets.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/static-assets/binding.mdx
+++ b/src/content/docs/workers/static-assets/binding.mdx
@@ -17,6 +17,8 @@ import {
 	Tabs,
 } from "~/components";
 
+import { WranglerConfig } from "~/components";
+
 Configuring a Worker with assets requires specifying a [directory](/workers/static-assets/binding/#directory) and, optionally, an [assets binding](/workers/static-assets/binding/), in your Worker's `wrangler.toml` file. The [assets binding](/workers/static-assets/binding/) allows you to dynamically fetch assets from within your Worker script (e.g. `env.ASSETS.fetch()`), similarly to how you might with a make a `fetch()` call with a [Service binding](/workers/runtime-apis/bindings/service-bindings/http/).
 
 Only one collection of static assets can be configured in each Worker.
@@ -24,8 +26,6 @@ Only one collection of static assets can be configured in each Worker.
 ## `directory`
 
 The folder of static assets to be served. For many frameworks, this is the `./public/`, `./dist/`, or `./build/` folder.
-
-import { WranglerConfig } from "~/components";
 
 <WranglerConfig>
 
@@ -59,6 +59,23 @@ _headers
 ```
 
 Now Wrangler will not upload these files as client-side assets when deploying the Worker.
+
+## `experimental_serve_directly`
+
+Controls whether assets will be served first on a matching request. `experimental_serve_directly = true` ([default](/workers/static-assets/routing/#default-behavior)) will serve any static asset matching a request, while `experimental_serve_directly = false` will unconditionally [invoke your Worker script](/workers/static-assets/routing/#invoking-worker-script-ahead-of-assets).
+
+<WranglerConfig>
+
+```toml title="wrangler.toml"
+name = "my-worker"
+compatibility_date = "2024-09-19"
+main = "src/index.ts"
+ # The following configuration unconditionally invokes the Worker script at
+ # `src/index.ts`, which can programatically fetch assets via the ASSETS binding
+assets = { directory = "./public/", binding = "ASSETS", experimental_serve_directly = false }
+```
+
+</WranglerConfig>
 
 ## `binding`
 
--- a/src/content/docs/workers/static-assets/compatibility-matrix.mdx
+++ b/src/content/docs/workers/static-assets/compatibility-matrix.mdx
@@ -40,7 +40,7 @@ We plan to bridge the gaps between Workers and Pages and provide ways to migrate
 | **Static Assets**                                                                   |         |         |
 | [Early Hints](/pages/configuration/early-hints/)                                    | ❌      | ✅      |
 | [Custom HTTP headers for static assets](/pages/configuration/headers/)              | 🟡 [^1] | ✅      |
-| [Middleware](/pages/functions/middleware/)                                          | 🟡 [^2] | ✅      |
+| [Middleware](/workers/static-assets/binding/#experimental_serve_directly)           | ✅ [^2] | ✅      |
 | [Redirects](/pages/configuration/redirects/)                                        | 🟡 [^3] | ✅      |
 | [Smart Placement](/workers/configuration/smart-placement/)                          | ✅      | ✅      |
 | **Observability**                                                                   |         |         |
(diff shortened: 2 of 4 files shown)


src/content/docs/workers/static-assets/binding.mdx
```diff
@@ -37,11 +37,26 @@ assets = { directory = "./public/" }
 
 </WranglerConfig>
 
-## `binding`
+## `serve_directly`
 
-Configuring the optional [binding](/workers/runtime-apis/bindings) gives you access to the collection of assets from within your Worker script.
+Controls whether assets will be served first on a matching request. `serve_directly = true` ([default](/workers/static-assets/routing/#default-behavior)) will serve any static asset matching a request, while `serve_directly = false` will unconditionally [invoke your Worker script](/workers/static-assets/routing/#invoking-worker-script-ahead-of-assets).
```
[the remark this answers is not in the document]
WillTaylorDev: I can't seem to get this to be happy -- lowercasing the path doesn't seem to link correctly (locally). Is there a way to disable this check?

src/content/docs/workers/static-assets/routing.mdx
```diff
@@ -33,6 +37,58 @@ In this example, request to `example.com/api` doesn't match a static asset so th
...
+
+```js
+export default {
+  async fetch(
+    request: Request,
+    env: Env,
+    ctx: ExecutionContext
+  ): Promise<Response> {
+    const url = new URL(request.url);
+    if (url.pathname === "/supersecret.html") {
```
GregBrimble: Since we do `html_handling = "auto"` by default, this actually won't be the final path for this asset. Want to instead make it an image (`.jpg`) or something like that?

WillTaylorDev: Oh good call! Html handling is always something I forget 😅

src/content/docs/workers/static-assets/routing.mdx
```diff
@@ -33,6 +37,58 @@ In this example, request to `example.com/api` doesn't match a static asset so th
...
 ![A request to `example.com/blog` runs the Worker.](~/assets/images/workers/platform/assets/workers-assets-invoke-worker.png)
 
+## Invoking Worker Script Ahead of Assets
+
+Users may desire to run code before assets are served, much like [middleware in Pages](/pages/functions/middleware/). [`serve_directly`](/workers/static-assets/binding/#serve_directly) is provided as a configuration within wrangler.toml. When disabled, this option forces a user script to be invoked regardless of a request matching an asset.
+
+Take the following directory structure, wrangler.toml, and user Worker code:
+
+```
+|---- src
```
WalshyDev: https://starlight.astro.build/components/file-tree/

src/content/docs/workers/static-assets/routing.mdx
```diff
@@ -33,6 +37,58 @@ In this example, request to `example.com/api` doesn't match a static asset so th
 
 ![A request to `example.com/blog` runs the Worker.](~/assets/images/workers/platform/assets/workers-assets-invoke-worker.png)
 
+## Invoking Worker Script Ahead of Assets
+
+Users may desire to run code before assets are served, much like [middleware in Pages](/pages/functions/middleware/). [`serve_directly`](/workers/static-assets/binding/#serve_directly) is provided as a configuration within wrangler.toml. When disabled, this option forces a user script to be invoked regardless of a request matching an asset.
```
GregBrimble: I have a personal aversion to using the word "middleware" here, esp. when linking it to Pages Functions'. If we can get away without linking to Pages Functions' middleware here, I'd love to. And even better if we can avoid the word altogether, imo. But product should make the call.

```suggestion
Users may wish to run code before assets are served. This is often the case when implementing authentication, logging, personalization, internationalization, or other similar functions. [`serve_directly`](/workers/static-assets/binding/#serve_directly) is a configuration option available in `wrangler.toml` which controls this behavior. When disabled, `serve_directly = false` will invoke the Worker script regardless of any assets that would ordinarily otherwise match.
```

src/content/docs/workers/static-assets/compatibility-matrix.mdx
```diff
@@ -85,7 +85,7 @@ We plan to bridge the gaps between Workers and Pages and provide ways to migrate
 
 [^2]: Similar to <sup>3</sup>, to customize the HTTP headers that are returned by static assets, you can use [Service bindings](/workers/runtime-apis/bindings/service-bindings/) to connect a Worker in front of the Worker with assets.
 
-[^3]: If you need to run a Worker before serving static assets, you can create a separate Worker that acts as middleware, and then use [Service bindings](/workers/runtime-apis/bindings/service-bindings/) to forward the request to the Worker with assets. We plan to explore additional configuration to support more complex routing in the future.
+[^3]: Middleware can be configured via the `[serve_directly](/workers/static-assets/binding/#serve_directly)` option, but is charged as a normal Worker invocation. We plan to explore options for free middleware in the future.
```
GregBrimble:
```suggestion
[^3]: Middleware can be configured via the `[serve_directly](/workers/static-assets/binding/#serve_directly)` option, but is charged as a normal Worker invocation. We plan to explore additional related options in the future.
```

GregBrimble: We should ✅ the matrix, link it to `serve_directly` rather than Pages Functions' thing, and yeah, update this footnote as you have.

src/content/docs/workers/static-assets/binding.mdx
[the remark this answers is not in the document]
GregBrimble: You just resolve the conversation :)

src/content/docs/workers/static-assets/routing.mdx
WillTaylorDev: I _knew_ using `middleware` would garner conversation 😆

GregBrimble (approved):
Hold off until everything is live in prod. And yeah, maybe let product have a pass. But LGTM!

WillTaylorDev:
Drafting for now-- floating around internally for feedback.

src/content/docs/workers/static-assets/routing.mdx
```diff
@@ -33,6 +37,62 @@ In this example, request to `example.com/api` doesn't match a static asset so th
 
 ![A request to `example.com/blog` runs the Worker.](~/assets/images/workers/platform/assets/workers-assets-invoke-worker.png)
 
+## Invoking Worker Script Ahead of Assets
+
+Users may wish to run code before assets are served. This is often the case when implementing authentication, logging, personalization, internationalization, or other similar functions. [`serve_directly`](/workers/static-assets/binding/#serve_directly) is a configuration option available in `wrangler.toml` which controls this behavior. When disabled, `serve_directly = false` will invoke the Worker script regardless of any assets that would have otherwise matched.
```
irvinebroque: "you" rather than "users" - write like you are talking to a person

irvinebroque: "Worker" rather than "Worker script"

GregBrimble: We were struggling with this a bit, so we're inconsistent in a few places right now.

A Worker can own some static assets. This is like how a Worker can own the implementation of a Durable Object. We didn't want to make it seem like you were connecting to an external resource (e.g. KV or R2) because that's not really an accurate model for how this actually works. We wanted it to be clear that a Worker was made up of compute server-side modules (a.k.a. script) and/or static assets.

When it came to talking about precedence of these two within a Worker, describing "static assets" was easy enough. But we couldn't find a good name for the script piece.

Please do make some suggestions. We're all ears.

irvinebroque: "your Worker's code" ?

korinne (approved):
Looks great!

PR #18510, opened 2024-12-02
update API documentation links

jacobbednarz:
🚨 Don't merge 🚨 

Updates all previous documentation links with the new API schema site structure.


jacobbednarz:
link checks will fail until we cut over the API documentation site. happy to wait for that and re-run that particular check but is expected for now.

pedrosousa:
We should probably also update:

- Target URLs of redirects pointing to the current API reference site. For example:

https://github.com/cloudflare/cloudflare-docs/blob/28308f02cc0f351986dbd21cee01c3e6c89dcd68/public/_redirects#L647

- Links in YAML files like changelogs also pointing to the current site. For example:

https://github.com/cloudflare/cloudflare-docs/blob/28308f02cc0f351986dbd21cee01c3e6c89dcd68/src/content/changelogs/r2.yaml?plain=1#L31


cc: @crwaters16

pedrosousa:
Also noticed that API URLs with parentheses aren't included in this PR yet. Example:
`/api/operations/custom-hostname-for-a-zone-delete-custom-hostname-(-and-any-issued-ssl-certificates)`

Probably related to the regex you used to capture existing links?

Btw, some of these URLs have escaped parentheses characters in the MDX source. For example:
`/api/operations/zone-analytics-\(-deprecated\)-get-dashboard`

pedrosousa:
Some (apparently normal?) API links still need to be updated — found 174 search results for `/api/operations/` across different documentation sections, including changelogs, notifications, and partials. Example:
https://github.com/cloudflare/cloudflare-docs/blob/761df1a70a61fbca94c74e22906c3ea46be623df/src/content/docs/cloudflare-one/identity/idp-integration/entra-id.mdx?plain=1#L209

Maybe rerunning the script will fix these occurrences?

jacobbednarz:
> Probably related to the regex you used to capture existing links?

there wasn't a regex used here; it was taking the existing URLs and mapping them to the current documentation URLs. if it isn't picked up, one of those are missing.

> Also noticed that API URLs with parentheses aren't included in this PR yet. Example:
/api/operations/custom-hostname-for-a-zone-delete-custom-hostname-(-and-any-issued-ssl-certificates)

this one is due to workaround being used to not escape the parentheses (it's called an automatic link). in the source it is laid out as 

```
To delete a custom hostname and any issued certificates using the API, send a [`DELETE` request](</api/operations/custom-hostname-for-a-zone-delete-custom-hostname-(-and-any-issued-ssl-certificates)>).
```

while valid, this doesn't get picked up as the URL is wrapped with `<>` violating the word boundary restriction we have on the replacement to prevent it from overlapping replacements. i'll update it to allow optional `<>` to wrap.

> Btw, some of these URLs have escaped parentheses characters in the MDX source. For example:
/api/operations/zone-analytics-\(-deprecated\)-get-dashboard

these won't automatically get picked up as that is not in the source of the openapi schema which was used for mapping. we can manually update these ones.

src/pages/search.astro
```diff
@@ -318,7 +318,7 @@ import "instantsearch.css/themes/satellite.css";
 			container: "#hits",
 			templates: {
 				item(hit, { html, components }) {
-					if (hit.path.startsWith("/api/operations/")) {
+					if (hit.path.startsWith("/api/resources/")) {
```
jacobbednarz, line 321: @KianNH does this still do what you need it to do here?

crwaters16 (approved):

src/pages/search.astro
KianNH: Looks good to me - I'm not sure if this code actually does anything since we don't search the API docs from DevDocs anymore anyhow.

KianNH (approved):

PR #18567, opened 2024-12-04
[worker] Document response stream disconnected error

jackatbancast:
### Summary

This change documents the `responseStreamDisconnected` event `outcome` seen in CloudFlare worker logs and error metrics.

Previously, this outcome doesn't appear to be documented, but presents as an error in metrics and presents as a non-error entry in the CloudFlare Worker Logs. This change aims to reduce confusion going forward.

Changes introducing `responseStreamDisconnected`: https://github.com/cloudflare/workerd/pull/2503

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/observability/errors.mdx
+++ b/src/content/docs/workers/observability/errors.mdx
@@ -231,10 +231,6 @@ Runtime errors will occur within the runtime, do not throw up an error page, and
 | `Memory limit`<br/>`would be exceeded`<br/> `before EOF` | Trying to read a stream or buffer that would take you over the [memory limit](/workers/platform/limits/#memory). |
 | `daemonDown`                                             | A temporary problem invoking the Worker.                                                                         |
 
-### Response Stream Disconnected
-
-A `responseStreamDisconnected` event `outcome` occurs when one end of the connection hangs up during the deferred proxying stage of a Worker request flow. This is regarded as an error for request metrics, and presents in logs as a non-error log entry. It commonly appears for longer lived connections such as WebSockets.
-
 ## Identify errors: Workers Metrics
 
 To review whether your application is experiencing any downtime or returning any errors:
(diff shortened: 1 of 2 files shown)


irvinebroque:
@jclee

irvinebroque (approved):

jclee:
> @jclee

Looks reasonable to me, FWIW.

src/content/docs/workers/observability/errors.mdx
```diff
@@ -231,6 +231,10 @@ Runtime errors will occur within the runtime, do not throw up an error page, and
 | `Memory limit`<br/>`would be exceeded`<br/> `before EOF` | Trying to read a stream or buffer that would take you over the [memory limit](/workers/platform/limits/#memory). |
 | `daemonDown`                                             | A temporary problem invoking the Worker.                                                                         |
 
+### Response Stream Disconnected
+
+A `responseStreamDisconnected` event `outcome` occurs when one end of the connection hangs up during the deferred proxying stage of a Worker request flow. This is regarded as an error for request metrics, and presents in logs as a non-error log entry. Commonly appears for longer lived connections such as websockets.
```
ToriLindsay:
```suggestion
A `responseStreamDisconnected` event `outcome` occurs when one end of the connection hangs up during the deferred proxying stage of a Worker request flow. This is regarded as an error for request metrics, and presents in logs as a non-error log entry. It commonly appears for longer lived connections such as WebSockets.
```

ToriLindsay:
@jackatbancast @jclee 
This has been added as a sub-heading under "Runtime errors". Is this the best placement for it in [the article](https://developers.cloudflare.com/workers/observability/errors/#identify-errors-workers-metrics)? Would it make more sense to add it to the error message table that is directly above it? (See screenshot)
Screenshot 2025-01-06 at 3 45 03 PM

jackatbancast:
@ToriLindsay thank you for the feedback and the suggestions.

I'd initially considered adding this to the table mentioned, but I don't believe it quite fits there as we don't get an error message in the logs, instead getting an `outcome` that looks like an error through the metrics interface.

This being said, I'm not sure if this page is the best place for this content at all given that it is not an error log, but I thought it would be best to make it slightly more discoverable.

I'm happy to defer to guidance on where this would be best placed.

ToriLindsay:
@jackatbancast Thanks for the context!
In that case, would it make more sense to put it in the section directly below: "[Identify errors: Workers Metrics](https://developers.cloudflare.com/workers/observability/errors/#identify-errors-workers-metrics)" ?
Screenshot 2025-01-07 at 11 35 45 AM

ToriLindsay:
@GregBrimble @jasnell Would either of you mind confirming that the location that I moved this info to in the docs makes sense? 
(Tldr: The contributor confirmed that it didn't belong under "Runtime errors", where he originally had it, so I moved it to "Identify errors: Workers Metrics" instead because I thought it might make more sense there.)

jackatbancast:
@ToriLindsay, thank you for making these changes. I was planning to pick this back up today after closing off other priorities. I think the change makes sense from my side, thank you for actioning your feedback.

PR #18584, opened 2024-12-05
Remove obsolete challenge actions

migueldemoura:
These actions have either never been used or are obsolete and about to be removed.


rupert648 (approved):

pedrosousa:
Hi @migueldemoura 
Do you have any suggestions regarding the WAF FAQ entries where we mention these challenge actions?
[#1](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/waf/troubleshooting/faq.mdx?plain=1#L166), [#2](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/waf/troubleshooting/faq.mdx?plain=1#L172), [#3](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/waf/troubleshooting/faq.mdx?plain=1#L183)

migueldemoura:
Happy to adjust those, but the tl;dr is that they were never useful and just outright misleading as we never log the overwhelming majority of challenge failures to security events.

pedrosousa:
We also have a couple of occurrences in the following page, which we can probably delete if they no longer make sense:
https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/logs/reference/security-fields.mdx?plain=1

migueldemoura:
Made a few adjustments, feel free to change them! I also noticed that we don't mention every kind of challenge type when referring to the action names, but I guess that can be fixed in a follow-up.

pedrosousa (approved):

PR #18596, opened 2024-12-05
Updates Python Workers examples to use new Python SDK API

dom96:
The new Python SDK is now available in the latest version of Wrangler. It provides Python native implementations of classes like `Response`.

This PR modifies most of our Python examples to use the new APIs in the SDK.


src/content/docs/workers/examples/bulk-redirects.mdx
```diff
@@ -88,14 +89,14 @@ async def on_fetch(request):
...
-    url = URL.new(request.url)
-    location = redirect_map.get(url.pathname, None)
+    url = urlparse(request.url)
+    location = redirect_map.get(url.path, None)
 
     if location:
         return Response.redirect(location, 301)
 
-    # If request not in map, return the original request
-    return fetch(request)
```
dom96: Something that confused me while I was going through these examples: this line ends up creating an infinite loop of requests back to the worker. What's the intention behind this? Is there something I missed?

I changed it here to return a 404 as it seems to make most sense for this example, but there are other examples that use the same pattern. I left those largely alone.

danlapid: See https://developers.cloudflare.com/workers/configuration/routing/custom-domains/#interaction-with-routes for an explanation, this change should probably be reverted.

src/content/docs/workers/examples/aggregate-requests.mdx
```diff
@@ -61,19 +61,21 @@ export default {
 </TabItem> <TabItem label="Python" icon="seti:python">
 
 ```py
-from js import Response, fetch, Headers, JSON, Promise
+from cloudflare.workers import Response, fetch
+import asyncio
```
irvinebroque, line 65: This can be deployed even when packages not enabled?

src/content/docs/workers/examples/bulk-redirects.mdx
dom96: Thanks. These examples should really call this pattern out. Getting an infinite loop on examples like this is really not a great dev experience when playing with these examples locally.

Maybe something for another PR

dom96:
- [x] The validator version that's currently in prod doesn't support these changes yet, so we need to wait for it to be updated before merging this.

src/content/docs/workers/examples/aggregate-requests.mdx
dom96: Yeah

irvinebroque (commented):
Where is the actual SDK documented? Showing what modules and methods are available on it?

ToriLindsay (approved):

megaconfidence (approved):

src/content/docs/workers/examples/aggregate-requests.mdx
```diff
@@ -61,19 +61,21 @@ export default {
...
 
 async def on_fetch(request):
     # some_host is set up to return JSON responses
     some_host = "https://jsonplaceholder.typicode.com"
     url1 = some_host + "/todos/1"
     url2 = some_host + "/todos/2"
 
-    responses = await Promise.all([fetch(url1), fetch(url2)])
-    results = await Promise.all(map(lambda r: r.json(), responses))
+    responses = await asyncio.gather(*[fetch(url1), fetch(url2)])
```
hoodmane:
```suggestion
    responses = await asyncio.gather(fetch(url1), fetch(url2))
```

src/content/docs/workers/examples/aggregate-requests.mdx
```diff
@@ -61,19 +61,21 @@ export default {
...
+    results = await asyncio.gather(*map(lambda r: r.json(), responses))
```
hoodmane:
```suggestion
    results = await asyncio.gather(*(r.json() for r in responses))
```

src/content/docs/workers/examples/aggregate-requests.mdx
```diff
@@ -61,19 +61,21 @@ export default {
...
 
-    headers = Headers.new({"content-type": "application/json;charset=UTF-8"}.items())
-    return Response.new(JSON.stringify(results), headers=headers)
+    headers = {"content-type": "application/json;charset=UTF-8"}
+    return Response(json.dumps(results), headers=headers)
```
hoodmane: It would nice to add a `Response.json()` method. That could also set the content type.

src/content/docs/workers/examples/alter-headers.mdx
```diff
@@ -80,28 +80,27 @@ export default {
...
-from js import Response, fetch
+from workers import Response, fetch
 
 async def on_fetch(request):
     response = await fetch("https://example.com")
 
-    # Clone the response so that it's no longer immutable
-    new_response = Response.new(response.body, response)
+    # Grab the response headers so they can be modified
+    new_headers = response.headers
```
hoodmane, line 89: What is `response.headers`?

src/content/docs/workers/examples/alter-headers.mdx
```diff
@@ -80,28 +80,27 @@ export default {
...
+        del new_headers["x-header-to-delete"]
+    if "x-header2-to-delete" in new_headers:
+        del new_headers["x-header2-to-delete"]
 
     # Adjust the value for an existing header
-    new_response.headers.set("x-header-to-change", "NewValue")
+    new_headers["x-header-to-change"] = "NewValue"
 
-    return new_response
+    return Response(response.body, headers=new_headers)
```
hoodmane, line 103: What would happen if we just returned the original response? Since it looks like we modified its headers in place?

dom96:
> Where is the actual SDK documented? Showing what modules and methods are available on it?

Nowhere as of right now, but we do have some code to generate markdown from the Python modules that we can get into the docs eventually.

src/content/docs/workers/examples/aggregate-requests.mdx
dom96: we do have this

src/content/docs/workers/examples/modify-response.mdx
```diff
@@ -132,12 +133,12 @@ async def on_fetch(request):
...
 
     # Change status and statusText, but preserve body and headers
-    response = Response.new(original_response.body, status=500, statusText="some message", headers=original_response.headers)
+    response = Response(original_response.body, status=500, statusText="some message", headers=original_response.headers)
 
     # Change response body by adding the foo prop
     original_body = await original_response.json()
-    original_body.foo = "bar"
-    response = Response.new(JSON.stringify(original_body), response)
+    original_body["foo"] = "bar"
```
hoodmane: Maybe `original_body` isn't the best name for something we're going to mutate.

src/content/docs/workers/examples/protect-against-timing-attacks.mdx
```diff
@@ -68,29 +68,30 @@ export default {
...
-        return Response.new("Unauthorized", status=401)
+        return Response("Unauthorized", status=401)
 
     if a.byteLength != b.byteLength:
-        return Response.new("Unauthorized", status=401)
+        return Response("Unauthorized", status=401)
 
     encoder = TextEncoder.new()
     a = encoder.encode(auth_token)
     b = encoder.encode(secret)
```
hoodmane, line 89: Using `TextEncoder` here is pretty weird I think. I like the following slightly better:
```py
    a = to_js(auth_token.encode())
    b = to_js(auth_token.encode())
```

hoodmane: Not really relevant to the current PR though I understand.

src/content/docs/workers/examples/security-headers.mdx
```diff
@@ -232,23 +232,23 @@ async def on_fetch(request):
     blocked_headers = ["Public-Key-Pins", "X-Powered-By" ,"X-AspNet-Version"]
 
     res = await fetch(request)
-    new_headers = Headers.new(res.headers)
+    new_headers = res.headers
```
hoodmane, line 235: Again, calling this `new_headers` but seemingly (?) modifying the original ones in place.

src/content/docs/workers/languages/python/examples.mdx
```diff
@@ -99,7 +101,7 @@ async def on_fetch(request, env):
 ## Query a D1 Database
 
 ```python
-from js import Response
+from workers import Response
```
hoodmane, line 105: So does `Response.json()` just work? Maybe we should replace the example above that does it manually?

hoodmane (approved):
Looks good to me aside from a bunch of minor comments. I'm most concerned about the semantics around mutation, but it's probably not that important.

src/content/docs/workers/examples/aggregate-requests.mdx
hoodmane: Should we use it here?

src/content/docs/workers/examples/alter-headers.mdx
dom96: It's https://pyodide.org/en/stable/usage/api/python-api/http.html#pyodide.http.FetchResponse.headers so dict[str, str]

src/content/docs/workers/examples/alter-headers.mdx
dom96: we don't modify them in-place, we get a new instance of the headers when we access `headers`.

src/content/docs/workers/examples/security-headers.mdx
dom96: nope, `headers` returns a new instance of headers as a dict

src/content/docs/workers/languages/python/examples.mdx
dom96: you mean that the above example uses to_js in the `Response.json`? I guess I missed that, fixed it.

PR #18617, opened 2024-12-06
Changed value for resolved to allow DNS over TLS for systemd-resolved

makuru-org:
### Summary

Please also update the website: https://one.one.one.one

Changed value for resolved to allow DNS over TLS for systemd-resolved as described below.
https://man.archlinux.org/man/resolved.conf.5#OPTIONS

### DIFF
--- a/src/content/docs/1.1.1.1/setup/linux.mdx
+++ b/src/content/docs/1.1.1.1/setup/linux.mdx
@@ -51,6 +51,12 @@ sudo <EDITOR> /etc/systemd/resolved.conf
 
 ```txt
 [Resolve]
+DNS=1.1.1.1
+```
+To use DNS over TLS, add `#one.one.one.one` as in the following example:
+
+```txt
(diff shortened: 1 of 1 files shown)


src/content/docs/1.1.1.1/setup/linux.mdx
```diff
@@ -53,6 +53,12 @@ sudo <EDITOR> /etc/systemd/resolved.conf
 [Resolve]
 DNS=1.1.1.1
 ```
+To use DNS over TLS, add `#one.one.one.one` as in the following example:
+
+```txt
+[Resolve]
+DNS=1.1.1.1#one.one.one.one
```
SebastiaanYN, line 60: I believe you also need to set `DNSOverTLS=true` for it to use DoT. Would be good to include that line in the example to prevent confusion

RebeccaTamachiro (approved):

PR #18634, opened 2024-12-09
Added notes for cache request hostnames

ToriLindsay:
Added notes - Do not use dynamic or invalid hostnames in cache requests. This can cause DNS cache failures and latency.

### DIFF
--- a/src/content/docs/workers/runtime-apis/cache.mdx
+++ b/src/content/docs/workers/runtime-apis/cache.mdx
@@ -51,12 +51,13 @@ await myCache.match(request);
 ```
 When using the cache API, avoid overriding the hostname in cache requests, as this can lead to unnecessary DNS lookups and cache inefficiencies. Always use the hostname that matches the domain associated with your Worker.
 
-```async fetch(request, env, ctx) {
-    // avoid overriding the request URL or passing a new request into the cache API
-    request.url = "https://some-overridden-value.com";
+```js
+async fetch(request, env, ctx) {
+// avoid overriding the request URL or passing a new request into the cache API
+request.url = "https://some-overridden-value.com";
 
-    let myCache = await caches.open('custom:cache');
-    await myCache.match(request);
+let myCache = await caches.open('custom:cache');
+await myCache.match(request);
 },
 ```
 ***
--- a/src/content/docs/workers/runtime-apis/cache.mdx
+++ b/src/content/docs/workers/runtime-apis/cache.mdx
@@ -49,7 +49,16 @@ You may create and manage additional Cache instances via the [`caches.open`](htt
 let myCache = await caches.open('custom:cache');
 await myCache.match(request);
 ```
+When using the cache API, avoid overriding the hostname in cache requests, as this can lead to unnecessary DNS lookups and cache inefficiencies. Always use the hostname that matches the domain associated with your Worker.
 
+```async fetch(request, env, ctx) {
+    // avoid overriding the request URL or passing a new request into the cache API
+    request.url = "https://some-overridden-value.com";
+
+    let myCache = await caches.open('custom:cache');
+    await myCache.match(request);
+},
+```
 ***
 
 ## Headers
@@ -105,14 +114,6 @@ cache.put(request, response);
 
 
 The `stale-while-revalidate` and `stale-if-error` directives are not supported when using the `cache.put` or `cache.match` methods.
-
-
-:::
-
-:::note
-To avoid unnecessary DNS lookups and cache inefficiencies, do not override the hostname in cache requests. Always use the hostname that matches the domain associated with your Worker.
-:::
-
 #### Parameters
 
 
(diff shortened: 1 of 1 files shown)


Oxyjun (approved):

src/content/docs/workers/runtime-apis/cache.mdx
```diff
@@ -109,6 +109,10 @@ The `stale-while-revalidate` and `stale-if-error` directives are not supported w
 
 :::
 
+:::note
+Do not use dynamic or invalid hostnames in cache requests. This can cause DNS cache failures and latency.
```
mikenomitch: I think we should be more explicit. We should say that the hostname used in the cache request should match your hostname.

And then instead of "DNS cache failures" I'd say "a DNS lookup" which will significantly increase latency.

mikenomitch: I think if they use "request.headers.host" that'd be the correct hostname to use in all cases? I could be wrong about this though

EDIT: Err... we don't need to specify this since they'll just be passing the whole request object in.

I might say "Do not override the hostname in the request or use a hostname that differs from your own"?

irvinebroque (changes requested):
Needs more context — show vs. just tell

src/content/docs/workers/runtime-apis/cache.mdx
ToriLindsay:
```suggestion
To avoid unnecessary DNS lookups and cache inefficiencies, do not override the hostname in cache requests. Always use the hostname that matches the domain associated with your Worker.
```

src/content/docs/workers/runtime-apis/cache.mdx
```diff
@@ -156,6 +160,10 @@ The `stale-while-revalidate` and `stale-if-error` directives are not supported w
...
```
ToriLindsay:
```suggestion
To avoid unnecessary DNS lookups and cache inefficiencies, do not override the hostname in cache requests. Always use the hostname that matches the domain associated with your Worker.
```

ToriLindsay:
@mikenomitch @irvinebroque Thank you for the feedback! I updated the note with the details from Mike. An example could also be included if you think that's necessary to show it more explicitly? Let me know your thoughts.

irvinebroque (commented):
Same feedback as previously — show vs. tell

src/content/docs/workers/runtime-apis/cache.mdx
```diff
@@ -49,7 +49,17 @@ You may create and manage additional Cache instances via the [`caches.open`](htt
 let myCache = await caches.open('custom:cache');
 await myCache.match(request);
 ```
+When using the cache API, avoid overriding the hostname in cache requests, as this can lead to unnecessary DNS lookups and cache inefficiencies. Always use the hostname that matches the domain associated with your Worker.
 
+```js
+async fetch(request, env, ctx) {
```
mikenomitch: Indentation on this JS is a bit off. Everything within the brackets should be indented once.

Though actually, this also isn't valid fetch code. You would need to `return` the response from `myCache.match(request)` in order for it to be valid.

But... that doesn't really make sense because you might have a cache miss and you'd have to handle that (see this full code - https://developers.cloudflare.com/workers/examples/cache-api/)

I think since we don't want to show all of that code in here, what we should do is just drop the fetch() bits completely? It might be a tad less clear where `request` is coming from in this case, but I think it's fine.

So I'd just do 

```
// avoid overriding the URL in request or passing a new request into the cache API
request.url = "https://some-overridden-value.com/";

let myCache = await caches.open('custom:cache');
let response = await myCache.match(request);
```

If it'll never work then I'd add this above the last line "// will never work because URL doesn't match Worker's domain"

src/content/docs/workers/runtime-apis/cache.mdx
```diff
@@ -49,7 +49,17 @@ You may create and manage additional Cache instances via the [`caches.open`](htt
 let myCache = await caches.open('custom:cache');
 await myCache.match(request);
 ```
+When using the cache API, avoid overriding the hostname in cache requests, as this can lead to unnecessary DNS lookups and cache inefficiencies. Always use the hostname that matches the domain associated with your Worker.
```
mikenomitch, line 53: Should we wrap this all in `:::note`?

mikenomitch (approved):
Looks good!

ToriLindsay:
@irvinebroque Ready for your re-review. I believe I need your approval as well for this to be unblocked. Thank you!

src/content/docs/workers/runtime-apis/cache.mdx
```diff
@@ -49,7 +49,17 @@ You may create and manage additional Cache instances via the [`caches.open`](htt
 let myCache = await caches.open('custom:cache');
 await myCache.match(request);
 ```
+:::note
+When using the cache API, avoid overriding the hostname in cache requests, as this can lead to unnecessary DNS lookups and cache inefficiencies. Always use the hostname that matches the domain associated with your Worker.
+
+```js
+// avoid overriding the URL in request or passing a new request into the cache API
```
irvinebroque:
```suggestion
// Do not do this
```

it was ambiguous when i read this what was meant by the comment — assume it's trying to tell people not to do this

irvinebroque (commented):
Seems like we should suggest what people _should_ do instead?

ToriLindsay:
> Seems like we should suggest what people _should_ do instead?

@irvinebroque Makes sense. I updated it again. Ready for your review when you have a chance. Thank you!

irvinebroque (approved):

PR #18709, opened 2024-12-12
Update BGE emebedding models to list vector dimensions

hturan:
This PR updates the descriptions for BGE embedding models to list vector dimensions in the descriptions.


craigsdennis (approved):
ty!

PR #18819, opened 2024-12-17
Adds Whisper model

craigsdennis:
### Summary

Adds Whisper model by using:

```bash
node ./bin/fetch-ai-models.js
```


KimJ15 (approved):
You have a double "the" on line 47. Approved if fixed!

PR #18821, opened 2024-12-18
R2-2317 Document SSE-C Availability

helloimalastair:
### Summary

Retroactively document when SSE-C support within R2 was rolled out to all customers

### DIFF
--- a/src/content/changelogs/r2.yaml
+++ b/src/content/changelogs/r2.yaml
@@ -5,6 +5,9 @@ productLink: "/r2/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+  - publish_date: "2024-12-03"
+    description: |-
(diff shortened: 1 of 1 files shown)


src/content/changelogs/r2.yaml
```diff
@@ -5,6 +5,9 @@ productLink: "/r2/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+  - publish_date: "2024-12-03"
+    description: |-
+      - [Server-side Encryption with Customer-Provided Keys](r2/examples/ssec) is now available to all users via the Workers and S3-compatible APIs.
```
WalshyDev:
```suggestion
      - [Server-side Encryption with Customer-Provided Keys](/r2/examples/ssec) is now available to all users via the Workers and S3-compatible APIs.
```

dcpena (approved):

PR #18914, opened 2024-12-23
[CEP] Update configuring-custom-pages-error-and-challenge.mdx

luishcardoso:
Hello team,

### Summary

Customers can collapse CEPs locally as well to see character size by running our collapsify tool.

Thanks!


ngayerie:
I'll defer to @nikitacano , but I think as the collapsify repo is archived we should not mention it here..

nikitacano:
We are not using this legacy version of collapsify at the edge anymore, so we shouldn't point customers to it as results might differ.

PR #18933, opened 2024-12-25
[ZT] macOS Big Sur or newer need to manually trust

kyouheicf:
It need to be fixed from Ventura to Big Sur because of this release.

https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes#Security

>**Security**
>**New Features**
>- macOS Big Sur 11 beta improves system security by requiring an administrator password when a certificate trust settings change is made in the admin trust domain. Running as the root user alone is no longer sufficient to modify certificate trust. User trust domain settings continue to require confirmation by entering the password for the user’s account. This change may affect you if one of the following is true:
>   - You have written scripts which call /usr/bin/security add-trusted-cert -d ... as root.
>   - Your process runs as root and calls the SecTrustSettingsSetTrustSettings function to trust a certificate.

Ref. [security add-trusted-cert asks pas… | Apple Developer Forums](https://forums.developer.apple.com/forums/thread/671582)


kyouheicf:
PCX-15148

kokolocomotion1 (approved):

maxvp (approved):

PR #18934, opened 2024-12-25
[ZT] Change ProgramData to the upper case

kyouheicf:
`%ProgramData%` path does not work for the device posture as the file check.

Can you change it to `%PROGRAMDATA%` so that we can use it for the device posture as the file check?


kyouheicf:
PCX-15149

ranbel (approved):

PR #19105, opened 2025-01-09
Clarify the ability of using Custom Rules to act on truncated payloads

Vortexmind:
As per title

### DIFF
--- a/src/content/docs/waf/managed-rules/index.mdx
+++ b/src/content/docs/waf/managed-rules/index.mdx
@@ -52,6 +52,4 @@ At the zone level, you can only deploy each WAF managed ruleset once. At the [ac
 
 Cloudflare analyzes the body of incoming requests up to a certain maximum size that varies according to your Cloudflare plan. For Enterprise customers, the maximum body size is 128 KB, while for other plans the limit is lower. This means that the behavior of specific managed rules that analyze request bodies can vary according to your current Cloudflare plan.
 
-:::note
-If included in your plan, you can use the [HTTP request body fields](/ruleset-engine/rules-language/fields/http-request-body/) such as `http.request.body.truncated` or `http.request.headers.truncated` to create [Custom Rules](/waf/custom-rules/) and apply appropriate actions to requests that have been truncated due to the maximum body size.
-:::
+If included in your plan, you can use [request body fields](/ruleset-engine/rules-language/fields/http-request-body/) such as `http.request.body.truncated` or `http.request.headers.truncated` in [custom rules](/waf/custom-rules/) that apply appropriate actions to requests that have not been fully analyzed by Cloudflare due to the maximum body size.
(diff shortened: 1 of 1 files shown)


pedrosousa:
Hi @Vortexmind 

As stated in the documentation, the (plan-dependent) maximum body size [only applies to the values of Cloudflare body fields](https://developers.cloudflare.com/ruleset-engine/rules-language/fields/http-request-body/#:~:text=The%20maximum%20body%20size%20applies%20only%20to%20the%20values%20of%20HTTP%20body%20fields) — the origin server will always receive the complete request body, without any truncation.

Truncation info is useful to understand why a given Cloudflare rule with a body field in its expression may not be triggering. For example, if a rule matches a body containing a specific string, it may not match a request if that string does not occur in the piece of the request body that is actually evaluated by Cloudflare (i.e., if it appears in the bytes that got truncated).

Since the truncation of body values in Cloudflare fields has no impact on what the origin server receives, what kind of rule would you create based on truncated body values?

Vortexmind:
Hi @pedrosousa - thanks for your reply. If I am not misunderstanding, in scenarios where security is paramount I could decide to block a request if the flag shows that truncation has occurred, to avoid a scenario where the offending payload was in the part that was not analyzed. Thanks, Paolo

pedrosousa:
Got it, thanks for clarifying. I will slightly rephrase your proposal because I had a different understanding from what I read.

src/content/docs/waf/managed-rules/index.mdx
```diff
@@ -51,3 +51,7 @@ At the zone level, you can only deploy each WAF managed ruleset once. At the [ac
 ## Important remarks
 
 Cloudflare analyzes the body of incoming requests up to a certain maximum size that varies according to your Cloudflare plan. For Enterprise customers, the maximum body size is 128 KB, while for other plans the limit is lower. This means that the behavior of specific managed rules that analyze request bodies can vary according to your current Cloudflare plan.
+
+:::note
+If included in your plan, you can use the [HTTP request body fields](/ruleset-engine/rules-language/fields/http-request-body/) such as `http.request.body.truncated` or `http.request.headers.truncated` to create [Custom Rules](/waf/custom-rules/) and apply appropriate actions to requests that have been truncated due to the maximum body size.
+:::
```
pedrosousa:
```suggestion
If included in your plan, you can use [request body fields](/ruleset-engine/rules-language/fields/http-request-body/) such as `http.request.body.truncated` or `http.request.headers.truncated` in [custom rules](/waf/custom-rules/) that apply appropriate actions to requests that have not been fully analyzed by Cloudflare due to the maximum body size.
```

pedrosousa (commented):
Added a suggested edit.

pedrosousa (approved):

PR #19130, opened 2025-01-10
Clarify overlapping event rules

### LINKED ISSUE #17949: Add documentation regarding conflicting event notification rules

- https://developers.cloudflare.com/r2/buckets/event-notifications/

Provide documentation regarding when event notification rules conflict with other event notification rules

### Additional information

e.g. when adding an event notification rule on some bucket for some queue, you can not have multiple event notifications configured for the same action, e.g. 'PutObject'. It will give you an error mentioning that there are conflicting rules.

harshil1712:
### Summary

Add notes to clarify overlapping event rules. Fixes #17949

### DIFF
--- a/src/content/docs/r2/buckets/event-notifications.mdx
+++ b/src/content/docs/r2/buckets/event-notifications.mdx
@@ -232,3 +232,6 @@ Queue consumers receive notifications as [Messages](/queues/configuration/javasc
 ## Notes
 
 - Queues [per-queue message throughput](/queues/platform/limits/) is currently 5,000 messages per second. If your workload produces more than 5,000 notifications per second, we recommend splitting notification rules across multiple queues.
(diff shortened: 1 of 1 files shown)


bthwaites:
This isn't quite accurate. You would be allowed to configure one rule with PutObject, prefix `a/`, and another with CopyObject, prefix `a/`. The key point is that we will reject any rule that overlaps with another rule such that multiple rules may trigger from the same R2 request. I'd suggest something like this:

- Rule prefixes and suffixes for the same R2 action may not overlap in such a way that could cause multiple rules to trigger at once. For example, a rule with PutObject and prefix `*` and a rule with PutObject and prefix `a/` would both match an object named `a/b`, and therefore may not be configured at the same time.

What do you think @jonesphillip ?

harshil1712:
Thank you @bthwaites. I checked with @jonesphillip and have made the changes.

jonesphillip (approved):

Oxyjun (approved):

PR #19208, opened 2025-01-14
[Workers] Clarify service binding pricing for deprecated usage models

maxwellpeterson:
- Customer was reading current documentation as "subrequests ... are charged as if they are requests from the Internet"
  - This isn't true, we never charge for subrequests
- Reword to clarify that _incoming_ requests from service bindings are what we're charging for
- Remove dead links to old usage model docs

### Documentation checklist


irvinebroque (approved):
Great change, thank you for catching

PR #19253, opened 2025-01-16
[AIG]add filter table

daisyfaithauma:
### Summary

- Added table with all filters

### Documentation checklist


kathayl:
think this filter list is super helpful. would be worth linking to elsewhere. like i'm thinking it could be helpful in the observability section too so people know how they can filter and look through their logs...
--> should it also be on this page? https://developers.cloudflare.com/ai-gateway/observability/logging/

pending the decision to the above question, i think we should def link to whereever the table is from this section where we talk about filtering to delete logs!
https://aig-filters.cloudflare-docs-7ou.pages.dev/ai-gateway/observability/logging/#manual-deletion

kathayl (approved):

PR #19323, opened 2025-01-21
Update compat language to make nodejs_compat with polyfills the defau…

thomasgauvin:

PR #19327, opened 2025-01-21
adjust db docs for no default nodejs compat

thomasgauvin:


thomasgauvin:
This attempts to reuse partials introduced in https://github.com/cloudflare/cloudflare-docs/pull/19323, only merge this PR when 19323 is merged

PR #19334, opened 2025-01-21
[Docs Site] Shrink codeblock buttons on mobile to desktop size

KianNH:
### Summary

EC sets the copy button to be `2.5rem` on mobile and `2rem` on desktop, this change means it will always be `2rem`.

### Screenshots (optional)

Before:
image

After:
image

### DIFF
--- a/src/code.css
+++ b/src/code.css
@@ -1,0 +1,8 @@
+/*
+    This is normally set to 2.5rem the user is unable to hover (i.e mobile)
+    and 2rem otherwise, we would like it to always be 2rem.
+*/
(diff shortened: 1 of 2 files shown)


WalshyDev (approved):

src/code.css
```diff
@@ -0,0 +1,8 @@
+/*
+    This is normally set to 2.5rem the user is unable to hover (i.e mobile)
```
KianNH:
```suggestion
    This is normally set to 2.5rem if the user is unable to hover (i.e mobile)
```

PR #19335, opened 2025-01-21
[Docs Site] Restore analytics events for various elements

KianNH:
### Summary

Restore analytics events for various elements, using their original event names.

The new `track` utility provided in `~/util/zaraz` will log the event to the console in the event that `window.zaraz` is undefined, such as in local development or preview builds.

image

### DIFF
--- a/src/components/Stream.astro
+++ b/src/components/Stream.astro
@@ -6,7 +6,12 @@ interface Props {
 	moreVideosLink: string;
 }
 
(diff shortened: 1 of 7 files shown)


kodster28 (approved):
LGTM. 

The logic for `external link clicks` is a touch overactive on previews (b/c technically all links are external to our regular site when you're on a preview URL)... but shouldn't impact anything.

PR #19341, opened 2025-01-22
Adding changelog for node compat improvements

mikenomitch:
Adds Changelog for new Node APIs and docs for them

### Summary

Adds a changelog for 3 additional node modules we now support and related docs.

### Screenshots (optional)
Screenshot 2025-01-28 at 10 59 25 AM
Screenshot 2025-01-28 at 10 59 29 AM
Screenshot 2025-01-28 at 11 00 21 AM

### DIFF
--- a/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
+++ b/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
@@ -3,7 +3,6 @@ title: Support for Node.js DNS, Net, and Timer APIs in Workers
 description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compat.
 products:
   - workers
-	- pages
 date: 2025-01-22T13:00:00Z
 ---
 
@@ -53,7 +52,7 @@ Note that [`net.Server`](https://nodejs.org/api/net.html#class-netserver) is not
 
 #### node:dns
 
-You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution via DNS lookups using
+You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution via [DNS over HTTPS](/1.1.1.1/encryption/dns-over-https/) using
 [Cloudflare DNS](https://www.cloudflare.com/application-services/products/dns/) at 1.1.1.1.
 
 <TypeScriptExample filename="index.ts">
@@ -61,7 +60,7 @@ You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution vi
 import dns from "node:dns";
 
 dns.lookup("example.org", (_err: any, address: string, ipFamily: number) =>
-	console.log(`address: ${address} family: IPv${ipFamily}`);
+	console.log(`address: ${address} family: IPv${ipFamily}`));
 ````
 
 </TypeScriptExample>
--- a/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
+++ b/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
@@ -17,31 +17,32 @@ When using a Worker with the [`nodejs_compat`](/workers/runtime-apis/nodejs/) co
 
 #### node:net
 
-You can use [`net.Socket`](https://nodejs.org/api/net.html#class-netsocket) to create a direct connection to a server via a TCP socket.
-This can be useful when connecting to databases such as [MySQL](https://www.mysql.com/) or [PostgreSQL](https://www.postgresql.org/),
-though if possible, using [Hyperdrive Connection Pooling](/hyperdrive/configuration/how-hyperdrive-works/#connection-pooling) is preferable.
+This makes it possible to connect to databases such as [MySQL](https://www.mysql.com/), [PostgreSQL](https://www.postgresql.org/),
+[Redis](https://redis.io/), or [MongoDB](https://github.com/mongodb/mongo). Though for production use, we recommend that you connect
+using TLS, which can be done with the [`cloudflare:sockets`](/workers/runtime-apis/tcp-sockets/#connect)
+module, or prefereably by using [Hyperdrive](/hyperdrive), which includes
+[connection pooling](/hyperdrive/configuration/how-hyperdrive-works/#connection-pooling)
+and [query caching](/hyperdrive/configuration/how-hyperdrive-works/#query-caching).
 
 <TypeScriptExample filename="index.ts">
 ```ts
-import net from 'node:net';
+import net from "node:net";
 
-const exampleIP = '127.0.0.1';
+const exampleIP = "127.0.0.1";
 
 export default {
-	async fetch(req): Promise<Response> {
-		const socket = new net.Socket();
-		socket.connect(4000, exampleIP, function () {
-			console.log('Connected');
-		});
+  async fetch(req): Promise<Response> {
+    const socket = new net.Socket();
+    socket.connect(4000, exampleIP, function () {
+      console.log("Connected");
+    });
 
-    	socket.write('Hello, Server!');
-    	socket.end();
+    socket.write("Hello, Server!");
+    socket.end();
 
-    	return new Response('Wrote to server', { status: 200 });
-    },
-
+    return new Response("Wrote to server", { status: 200 });
+  },
 } satisfies ExportedHandler;
-
 ````
 </TypeScriptExample>
 
@@ -77,16 +78,15 @@ and [`setImmediate`](https://nodejs.org/api/timers.html#setimmediatecallback-arg
 
 <TypeScriptExample filename="index.ts">
 ```ts
-import timers from 'node:timers';
+import timers from "node:timers";
 
 console.log("first");
 timers.setTimeout(() => {
-console.log("last");
+  console.log("last");
 }, 10);
 
 timers.setTimeout(() => {
-console.log("next");
+  console.log("next");
 });
-
 ```
 </TypeScriptExample>
--- a/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
+++ b/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
@@ -1,14 +1,15 @@
 ---
-title: Improved Node.js compatibility
-description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compatibility.
+title: Support for Node.js DNS, Net, and Timer APIs in Workers
+description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compat.
 products:
   - workers
+	- pages
 date: 2025-01-22T13:00:00Z
 ---
 
 import { Render, PackageManagers, TypeScriptExample } from "~/components";
 
-When using a Worker with [nodejs_compatability](/workers/runtime-apis/nodejs/) enabled, you can now use the following Node.js APIs:
+When using a Worker with the [`nodejs_compat`](/workers/runtime-apis/nodejs/) compatibility flag enabled, you can now use the following Node.js APIs:
 
 - [`node:net`](/workers/runtime-apis/nodejs/net/)
 - [`node:dns`](/workers/runtime-apis/nodejs/dns/)
@@ -19,8 +20,6 @@ When using a Worker with [nodejs_compatability](/workers/runtime-apis/nodejs/) e
 You can use [`net.Socket`](https://nodejs.org/api/net.html#class-netsocket) to create a direct connection to a server via a TCP socket.
 This can be useful when connecting to databases such as [MySQL](https://www.mysql.com/) or [PostgreSQL](https://www.postgresql.org/),
 though if possible, using [Hyperdrive Connection Pooling](/hyperdrive/configuration/how-hyperdrive-works/#connection-pooling) is preferable.
-
-These functions use [`connect`](/workers/runtime-apis/tcp-sockets/#connect) functionality from the built-in `cloudflare:sockets` module.
 
 <TypeScriptExample filename="index.ts">
 ```ts
@@ -49,7 +48,7 @@ export default {
 Additionally, you can now use other APIs incliding [`net.BlockList`](https://nodejs.org/api/net.html#class-netblocklist) and
 [`net.SocketAddress`](https://nodejs.org/api/net.html#class-netsocketaddress).
 
-Note that the [`net.Server`](https://nodejs.org/api/net.html#class-netserver) module is not supported.
+Note that [`net.Server`](https://nodejs.org/api/net.html#class-netserver) is not supported.
 
 #### node:dns
 
@@ -60,20 +59,20 @@ You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution vi
 ```ts
 import dns from "node:dns";
 
-dns.lookup("example.org", (_err: any, address: string, ipFamily: number) => {
-	console.log("address: %j family: IPv%s", address, ipFamily);
+dns.lookup("example.org", (_err: any, address: string, ipFamily: number) =>
+	console.log(`address: ${address} family: IPv${ipFamily}`);
 ````
 
 </TypeScriptExample>
 
-All `node:dns` functions are available.
+All `node:dns` functions are available, except `lookup`, `lookupService`, and `resolve` which throw "Not implemented" errors when called.
 
 #### node:timers
 
 You can use [`node:timers`](https://nodejs.org/api/timers.html) to schedule functions to be called at some future period of time.
 
 This includes [`setTimeout`](https://nodejs.org/api/timers.html#settimeoutcallback-delay-args) for calling a function after a delay,
-[`setInterval`](https://nodejs.org/api/timers.html#clearintervaltimeout) for calling a function repeatedly,
+[`setInterval`](https://nodejs.org/api/timers.html#setintervalcallback-delay-args) for calling a function repeatedly,
 and [`setImmediate`](https://nodejs.org/api/timers.html#setimmediatecallback-args) for calling a function in the next iteration of the event loop.
 
 <TypeScriptExample filename="index.ts">
@@ -91,8 +90,3 @@ console.log("next");
 
 ```
 </TypeScriptExample>
-
-Note that due to [security-based restrictions on timers](/workers/reference/security-model/#step-1-disallow-timers-and-multi-threading) in Workers,
-timers are limited to returning the time of the last I/O. This means that while setTimeout, setInterval, and setImmediate will defer your function execution
-until after other events have run, they will not delay them for the full time specified.
-```
--- a/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
+++ b/src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
@@ -1,0 +1,98 @@
+---
+title: Improved Node.js compatibility
+description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compatibility.
+products:
+  - workers
+date: 2025-01-22T13:00:00Z
+---
+
+import { Render, PackageManagers, TypeScriptExample } from "~/components";
+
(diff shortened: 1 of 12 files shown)


src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,96 @@
...
+	console.log("last");
+}, 10);
+
+timers.setTimeout(() => {
+	console.log("next");
+});
+```
+
+Note that due to [security-based restrictions on timers](/workers/reference/security-model/#step-1-disallow-timers-and-multi-threading) in Workers,
+timers are limited to returning the time of the last I/O. This means that while setTimeout and setInterval will defer your function execution
```
anonrig:
```suggestion
timers are limited to returning the time of the last I/O. This means that while setTimeout, setInterval and setImmediate will defer your function execution
```

anonrig (approved):

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+const exampleIP = '127.0.0.1';
+
+export default {
+	async fetch(req): Promise<Response> {
+		const socket = new net.Socket();
+		socket.connect(4000, exampleIP, function () {
+			console.log('Connected');
+		});
+
+    	socket.write('Hello, Server!');
```
vicb: nit: formatting looks off - `pnpm format` should help

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+} satisfies ExportedHandler;
+
+````
+</TypeScriptExample>
+
+Additionally, you can now use other APIs incliding [`net.BlockList`](https://nodejs.org/api/net.html#class-netblocklist) and
+[`net.SocketAddress`](https://nodejs.org/api/net.html#class-netsocketaddress).
+
+Note that the [`net.Server`](https://nodejs.org/api/net.html#class-netserver) module is not supported.
```
vicb: I don't think this is a "module"

```suggestion
Note that [`net.Server`](https://nodejs.org/api/net.html#class-netserver) is not supported.
```

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution via DNS lookups using
+[Cloudflare DNS](https://www.cloudflare.com/application-services/products/dns/) at 1.1.1.1.
+
+<TypeScriptExample filename="index.ts">
+```ts
+import dns from "node:dns";
+
+dns.lookup("example.org", (_err: any, address: string, ipFamily: number) => {
+	console.log("address: %j family: IPv%s", address, ipFamily);
```
vicb:
```suggestion
dns.lookup("example.org", (_err: any, address: string, ipFamily: number) => 
	console.log(`address: ${address} family: IPv${ipFamily}`);
```

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+````
+
+</TypeScriptExample>
+
+All `node:dns` functions are available.
```
vicb: not sure this is correct, @anonrig?

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+#### node:timers
+
+You can use [`node:timers`](https://nodejs.org/api/timers.html) to schedule functions to be called at some future period of time.
+
+This includes [`setTimeout`](https://nodejs.org/api/timers.html#settimeoutcallback-delay-args) for calling a function after a delay,
```
vicb: The links doesn't look they match the functions

src/content/docs/workers/runtime-apis/nodejs/dns.mdx
```diff
@@ -0,0 +1,23 @@
...
+<Render file="nodejs-compat-howto" />
+
+You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution via DNS lookups using
+[Cloudflare DNS](https://www.cloudflare.com/application-services/products/dns/) at 1.1.1.1.
+
+```js
+import dns from "node:dns";
+
+dns.lookup("example.org", (err, address, ipFamily) => {
+	console.log("address: %j family: IPv%s", address, ipFamily);
```
vicb: The same fix as above should be applied here

src/content/docs/workers/runtime-apis/nodejs/dns.mdx
```diff
@@ -0,0 +1,23 @@
...
+});
+```
+
+Note that DNS requests will execute a subrequest, counts for your [Worker's subrequest limit](/workers/platform/limits/#subrequests).
```
vicb: wrap this in a `:::note` ?

src/content/docs/workers/runtime-apis/nodejs/dns.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+The full `node:dns` API is documented in the [Node.js documentation for `node:dns`](https://nodejs.org/api/dns.html).
```
vicb, line 28: maybe callout the functions that are not implemented yet?
(if any but IIRC there are a few, @anonrig ?)

src/content/docs/workers/runtime-apis/nodejs/net.mdx
```diff
@@ -0,0 +1,45 @@
...
+pcx_content_type: configuration
+title: Net
+---
+
+import { Render, TypeScriptExample } from "~/components";
+
+<Render file="nodejs-compat-howto" />
+
+You can use [`node:net`](https://nodejs.org/api/net.html) to create a direct connection to servers via a TCP sockets
+with [`net.Socket`](https://nodejs.org/api/net.html#class-netsocket). This can be useful when connecting to databases such as MySQL or PostgreSQL, though if possible, using
```
vicb: mongoDB, redis?

src/content/docs/workers/runtime-apis/nodejs/net.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+} satisfies ExportedHandler;
+
+```
+</TypeScriptExample>
+
+Additionally, other APIs such as [`net.BlockList`](https://nodejs.org/api/net.html#class-netblocklist)
+and [`net.SocketAddress`](https://nodejs.org/api/net.html#class-netsocketaddress) are available.
+
+Note that the [`net.Server`](https://nodejs.org/api/net.html#class-netserver) module is not supported by Workers.
```
vicb: The fix above should be applied here too

src/content/docs/workers/runtime-apis/nodejs/timers.mdx
```diff
@@ -0,0 +1,36 @@
...
+}, 10);
+
+timers.setTimeout(() => {
+	console.log('next');
+});
+
+```
+</TypeScriptExample>
+
+Note that due to [security-based restrictions on timers](/workers/reference/security-model/#step-1-disallow-timers-and-multi-threading) in Workers,
```
vicb: wrap in `:::note`?

src/content/docs/workers/runtime-apis/nodejs/timers.mdx
```diff
@@ -0,0 +1,36 @@
...
+	console.log('last');
+}, 10);
+
+timers.setTimeout(() => {
+	console.log('next');
+});
+
+```
+</TypeScriptExample>
+
```
vicb, line 30: I think we should callout that the `node:timers` are "not interoperable" with the global timer functions because this behavior is different from Node.

We could also say "not *fully* interoperable" but we would then need to give details on what is vs what is not which could bring confusion.

For example (not sure we want to add this here):

```ts
import timers from 'node:timer';

// no-op, the timer is not cancelled
clearTimeout(timers.setTimeout(...));

// the timer is cancelled
timers.clearTimeout(setTimeout(...));
```

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: Improved Node.js compatibility
+description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compatibility.
+products:
+  - workers
```
vicb: also pages?

src/content/docs/workers/runtime-apis/nodejs/assert.mdx
```diff
@@ -1,39 +1,35 @@
 ---
 pcx_content_type: configuration
-title: assert
-
+title: Assert
```
vicb: Those changes (across multiple files in this PR) look wrong.
Module names are case sensitive and lower case.

```suggestion
title: assert
```

vicb (commented):
Awesome work 🎉

It will be really helpful for our users!

Added some comment inline.

src/content/docs/workers/runtime-apis/nodejs/assert.mdx
mikenomitch: We had the module names split between upper and lower case - so I tried unifying them all - debated between going all lower or all upper and decided on upper, but I can ask others what they think

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: Improved Node.js compatibility
```
irvinebroque: Think title can get more specific

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compatibility.
```
irvinebroque:
```suggestion
description: Node.js APIs from the node:dns, node:net, and node:timers modules are now available when using nodejs_compat.
```

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+products:
+  - workers
+date: 2025-01-22T13:00:00Z
+---
+
+import { Render, PackageManagers, TypeScriptExample } from "~/components";
+
+When using a Worker with [nodejs_compatability](/workers/runtime-apis/nodejs/) enabled, you can now use the following Node.js APIs:
```
irvinebroque:
```suggestion
When using a Worker with the [`nodejs_compat`](/workers/runtime-apis/nodejs/) compatibility flag enabled, you can now use the following Node.js APIs:
```

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+- [`node:net`](/workers/runtime-apis/nodejs/net/)
+- [`node:dns`](/workers/runtime-apis/nodejs/dns/)
+- [`node:timers`](/workers/runtime-apis/nodejs/timers/)
+
+#### node:net
+
+You can use [`net.Socket`](https://nodejs.org/api/net.html#class-netsocket) to create a direct connection to a server via a TCP socket.
+This can be useful when connecting to databases such as [MySQL](https://www.mysql.com/) or [PostgreSQL](https://www.postgresql.org/),
+though if possible, using [Hyperdrive Connection Pooling](/hyperdrive/configuration/how-hyperdrive-works/#connection-pooling) is preferable.
```
irvinebroque: This doesn't quite make sense to me since you need to connect to a DB securely using `tls`?

And we wouldn't ever actually recommend that someone connect to a DB from a Worker without Hyperdrive?

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+These functions use [`connect`](/workers/runtime-apis/tcp-sockets/#connect) functionality from the built-in `cloudflare:sockets` module.
```
irvinebroque: Why is this relevant to the reader? Is it because the use of `connect()` under the hood changes something about the APIs?

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+</TypeScriptExample>
+
+Additionally, you can now use other APIs incliding [`net.BlockList`](https://nodejs.org/api/net.html#class-netblocklist) and
+[`net.SocketAddress`](https://nodejs.org/api/net.html#class-netsocketaddress).
+
+Note that the [`net.Server`](https://nodejs.org/api/net.html#class-netserver) module is not supported.
+
+#### node:dns
+
+You can use [`node:dns`](https://nodejs.org/api/dns.html) for name resolution via DNS lookups using
```
irvinebroque: Do you want to explain somewhere that this is DoH?

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
```diff
@@ -0,0 +1,98 @@
...
+timers.setTimeout(() => {
+console.log("next");
+});
+
+```
+</TypeScriptExample>
+
+Note that due to [security-based restrictions on timers](/workers/reference/security-model/#step-1-disallow-timers-and-multi-threading) in Workers,
+timers are limited to returning the time of the last I/O. This means that while setTimeout, setInterval, and setImmediate will defer your function execution
+until after other events have run, they will not delay them for the full time specified.
```
irvinebroque: This seems like more of a detail for the docs page about the API than the changelog post

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
anonrig: They are available, but 3 of those functions are stub. lookup, lookupService and resolve. Ref: https://github.com/cloudflare/workerd/blob/main/src/node/internal/internal_dns.ts#L37

We can implement them in the future, if needed.

src/content/docs/workers/runtime-apis/nodejs/dns.mdx
anonrig: Responded to the other review.

src/content/docs/workers/runtime-apis/nodejs/net.mdx
anonrig:
```suggestion
Note that the [`net.Server`](https://nodejs.org/api/net.html#class-netserver) class is not supported by Workers.
```

src/content/docs/workers/runtime-apis/nodejs/timers.mdx
anonrig: > I think we should callout that the node:timers are "not interoperable" with the global timer functions because this behavior is different from Node.

IMHO: This should be mentioned in nodejs docs about node:timers, not in cloudflare.

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
vicb: Yep, I think we should clearly indicate that those functions are throwing for now

src/content/docs/workers/runtime-apis/nodejs/timers.mdx
vicb: Well that is the exact issue here, this problem is specific to Cloudflare and **does not exist on Node**

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
mikenomitch: Yeah good point - I might keep it in the docs so they know about that as a non-Node option, but not here

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
mikenomitch: One was wrong (fixed now), but the others are good - the arguments in the function name get added to the URL, so it looks off

src/content/changelogs-next/2025-01-22-nodejs-compat-improvements.mdx
mikenomitch: +1 - keeping in docs but removing from here

src/content/docs/workers/runtime-apis/nodejs/timers.mdx
vicb: That's why I think we should add a callout - we are not compatible with Node.js here.

mikenomitch: Do we have a thread on this generally? I'm not sure I fully understand what's going on here.

On the example provided, so in a Node.js environment the timer could be cancelled in the first example `clearTimeout(timers.setTimeout(...));` but for us it is not cancelled?

If so, yeah I think that deserves a callout if behavior differs from Node.js.

Any other cases to mention?

vicb: There was some discussion on the [PR](https://github.com/cloudflare/workerd/pull/3346)

What doesn't work on cloudflare that would work in Node:

```js
import timers from "node:timers":

// no-op on cf, works on Node
clearTimeout(timers.setTimeout(...));
clearInterval(timers.setInterval(...));
timers.active(setTimeout(...));
timers.active(setInterval(...)):

// throws on cf, works with Node
const timeout = setTimeout(...); // or setInterval(...) 
timeout.refresh();
timeout.ref();
timeout.unref();
timeout.hasref();
timeout.close();
```

Also we might want to mention that the following will throw:

```
timers.enroll(...);
```

I think this should be an exhaustive list.

@anonrig please correct me if this is wrong - I think you will implement some of the first list (`// no-op on cf, works on Node`). But this PR is already in prod so I think it's worth documenting the state as of today and trim the list when changes are made.

vicb: > On the example provided, so in a Node.js environment the timer could be cancelled in the first example `clearTimeout(timers.setTimeout(...));` but for us it is not cancelled?

Your understanding is correct.

anonrig: setTimeout, setImmediate and setInterval doesn't return Timeout/Immediate classes on cloudflare because the web spec says that it should return an ID. Whereas, Node.js doesn't implement this. But this fact doesn't change with the new changes. We are not changing the existing behavior, we are just adding new modules. I don't think this changelog is the place to list every single differences that we have.

I recommend keeping this post as simple as possible, and just mention the behavior that we have added. IMHO: Other information should be mentioned in the documentation (not on changelog)

mikenomitch: Okay, I updated the node:timers docs to include a Note here. I didn't go super in depth though FWIW.

Kept out of the changelog since this changelog post is already super long and felt docs-y.

vicb: Oh... You probably got confused by the title of the PR (can it be updated?)
But we are now commenting on the documentation for the timers module, similar to https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/

anonrig (approved):

irvinebroque (approved):

src/content/docs/workers/runtime-apis/nodejs/timers.mdx
mikenomitch: Oh, no I gotcha - Just being overly clear on where I added it :)

PR #19350, opened 2025-01-22
[Docs Site] Decode HTML entities and strip Markdown from page descriptions

KianNH:
### Summary

Before:

```
You can use zaraz.track() anywhere inside the &#x3C;body> tag of a page.

Write Workers in 100% Rust using the [`workers-rs`crate](https://github.com/cloudflare/workers-rs)
```

After:

```
You can use zaraz.track() anywhere inside the <body> tag of a page.

Write Workers in 100% Rust using the workers-rs crate
```


src/components/overrides/Head.astro
```diff
@@ -139,7 +139,7 @@ if (currentSection) {
 	}
 }
 
-Astro.props.entry.data.description ??= await getPageDescription(
+Astro.props.entry.data.description = await getPageDescription(
```
KianNH, line 142: We used to do `if (entry.data.description) return entry.data.description;` in `getPageDescription` but that didn't do anything - because we never used this function if that property existed.

That was a bug, but also fine since we never would have changed it. Now we are transforming it so removed `??`.

src/util/description.ts
```diff
@@ -1,25 +1,37 @@
 import type { CollectionEntry } from "astro:content";
 import { parse } from "node-html-parser";
 import { entryToString } from "./container";
-/*
-    1. If there is a `description` property in the frontmatter, return that.
-    2. If there is a `<p>...</p>` element in the HTML, return that.
-    3. Return `undefined` to signal to consumers there is no suitable description.
-*/
+import { remark } from "remark";
+import strip from "strip-markdown";
```
KianNH, line 5: This could technically be done in a lot more lines with `marked` but most of their discussions recommend using other libraries, like `strip-markdown`.

src/util/description.ts
```diff
@@ -1,25 +1,37 @@
...
-	if (!html) return undefined;
+		if (!html) return undefined;
 
-	const dom = parse(html);
-	const description = dom.querySelector(":root > p");
+		const dom = parse(html);
+		const paragraph = dom.querySelector(":root > p");
 
-	if (description) return description.innerText;
+		if (paragraph) description = he.decode(paragraph.innerText);
```
KianNH, line 33: `&#x3C;body>` -> `<body>`

src/util/description.ts
```diff
@@ -1,25 +1,37 @@
...
+ */
 export async function getPageDescription(
 	entry: CollectionEntry<"docs">,
 	locals: any,
 ) {
-	if (entry.data.description) return entry.data.description;
+	let description = undefined;
+
+	if (entry.data.description) {
+		const file = await remark().use(strip).process(entry.data.description);
```
KianNH, line 22:
```
[`workers-rs` crate](https://github.com/cloudflare/workers-rs)` -> `workers-rs crate`
```

kodster28 (approved):
Output lgtm + this should only affect descriptions so it's a low-risk change.

PR #19369, opened 2025-01-22
[Style Guide] Document using JSX in partials

KianNH:
### Summary

Documents using JSX in partials

### DIFF
--- a/src/content/docs/style-guide/components/render.mdx
+++ b/src/content/docs/style-guide/components/render.mdx
@@ -102,12 +102,13 @@ params:
 
 Generic content.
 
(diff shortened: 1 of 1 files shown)


pedrosousa:
Hi Kian! A couple of requests:
- Can we expand the example a bit more? Not everyone in the team is well versed in JSX and how to create these optional blocks from start to finish. Even if we need to repeat ourselves a bit with content already covered in other pages... If the example also covered the use of the `Markdown` component it would be great!
- Can we add an `if/then/else` example, probably using a `condition ? true : false` conditional?

IreneSAvila (commented):
#17108

KianNH:
Superseded by https://github.com/cloudflare/cloudflare-docs/pull/19721

PR #19390, opened 2025-01-23
[Docs Site] Adopt Cloudflare styling for badge component

KianNH:
### Summary

Adopt Cloudflare styling for badge component. This overrides all `variant` and `size` combinations, so all badges will be blue in light mode and orange in dark mode.

There is an escape hatch for existing components that leveraged gray badges, where `class="sl-badge gray"` can be used.

We will adopt the rest of the hues from the official component and map these to Starlight's variants in the future.

### Screenshots (optional)

Before:
image

After:
image


kodster28 (approved):

pedrosousa:
Should we define an extra color besides `gray` for deprecated features such as [firewall rules](https://developers.cloudflare.com/firewall/)?

We could use the following colors from the dashboard pill for "Previous version" (couldn't find a "Deprecated" pill):

Light mode:
image
```
background-color: rgb(254, 204, 200);
color: rgb(90, 8, 1);
```

Dark mode:
image
(this one seems a bit too strong, maybe we could make it lighter)
```
background-color: rgb(252, 87, 74);
color: rgb(29, 29, 29);
```

KianNH:
Added the per-variant colours

image

image

image

pedrosousa (approved):

PR #19444, opened 2025-01-27
[Docs Site] Set eslint in CI to use filter_mode: nofilter

KianNH:
### Summary

Set eslint in CI to use filter_mode: nofilter. This stops errors from getting past reviewdog which would later be caught by `npx eslint .` locally.


KianNH:
Error is successfully picked up

image

WalshyDev (approved):

PR #19446, opened 2025-01-27
[Docs Site] Fix unmatched CODEOWNERS entries

KianNH:
### Summary

Fixes unmatched CODEOWNERS entries

### DIFF
--- a/bin/validate-codeowners.ts
+++ b/bin/validate-codeowners.ts
@@ -1,0 +1,34 @@
+import { spawnSync } from "node:child_process";
+import { loadOwners, matchPattern } from "codeowners-utils";
+
+let owners = await loadOwners(process.cwd());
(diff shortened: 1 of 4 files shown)


deadlypants1973 (approved):

bin/validate-codeowners.ts
WalshyDev, line 1: planning to add as a workflow on codeowner change?

WalshyDev (approved):

bin/validate-codeowners.ts
KianNH: I guess we could but I also don't want it to be "Congratulations, you're trying to touch this file but we deleted the folder on L372 so you gotta go fix that too". I might just make it a scheduled job like our existing anchor link stuff.

PR #19452, opened 2025-01-27
Fix incorrect expression for Zone Lockdown example

connorgurney:
This is a minor change that ensures that the expression matches the example on the page. Given how little I’ve changed, it naturally adheres to the [documentation style guide](https://developers.cloudflare.com/style-guide/).


pedrosousa:
Thank you for the fix, @connorgurney-personal!

pedrosousa (approved):

PR #19471, opened 2025-01-27
[Docs Site] Rewrite /api/ links to point at production

KianNH:
### Summary

Rewrite /api/ links to point at production so they work when browsing from `astro dev` or in preview URLs.


astro.config.ts
```diff
@@ -14,6 +14,7 @@ import rehypeMermaid from "./src/plugins/rehype/mermaid.ts";
 import rehypeAutolinkHeadings from "./src/plugins/rehype/autolink-headings.ts";
 import rehypeExternalLinks from "./src/plugins/rehype/external-links.ts";
 import rehypeHeadingSlugs from "./src/plugins/rehype/heading-slugs.ts";
+import rehypeRewriteUrls from "~/plugins/rehype/rewrite-urls.ts";
```
KianNH:
```suggestion
import rehypeRewriteUrls from "./src/plugins/rehype/rewrite-urls.ts";
```

KianNH:
Will abandon this approach and instead look at something like Astro middleware that is only registered in `dev`.

PR #19512, opened 2025-01-29
[Rules] Update functions.mdx

ngayerie:
### Summary

Adding an example for the url_decode with the u option
SPM-2470


pedrosousa (approved):

pedrosousa:
Ideally this would live in the regular expressions documentation, but we don't have a section about the regex supported [flags](https://docs.rs/regex/latest/regex/#grouping-and-flags) and Unicode patterns for now.

PR #19578, opened 2025-01-30
Adds streaming HTMLRewriter content docs and changelog

mikenomitch:
### Summary

Adds documentation and a changelog for recent updates to HTMLRewriter that allow for streamed content.

Screenshot 2025-01-30 at 3 34 42 PM

### DIFF
--- a/src/content/changelogs-next/2025-02-03-html-rewriter-streaming.mdx
+++ b/src/content/changelogs-next/2025-02-03-html-rewriter-streaming.mdx
@@ -1,0 +1,40 @@
+---
+title: Rewrite HTML with streaming content
+description: HTMLRewiter now supports streamed content for more efficient replacement of HTML elements
+products:
+  - workers
+date: 2025-02-03T3:00:00Z
+---
+
+import { Render, TypeScriptExample } from "~/components";
+
+[`HTMLRewriter`](/workers/runtime-apis/html-rewriter) now supports replacing HTML with content from a stream.
+
+This allows you to process HTML more efficiently, as content does not have to be loaded into memory before replacements are made.
+
+[`Response`](/workers/runtime-apis/response/) and [`ReadableStream`](/workers/runtime-apis/streams/readablestream/) values can now
+be passed as [`Content`](/workers/runtime-apis/html-rewriter/#global-types) into methods duch as
+`replace`, `append`, and `prepend`.
+
+<TypeScriptExample filename="index.ts">
+```ts
+class ElementRewriter {
+	async element(element: any) {
+		// able to replace elements while streaming content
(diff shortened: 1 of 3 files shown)


npaun (approved):

irvinebroque (approved):

src/content/changelogs-next/2025-02-03-html-rewriter-streaming.mdx
```diff
@@ -0,0 +1,40 @@
...
+title: Rewrite HTML with streaming content
+description: HTMLRewiter now supports streamed content for more efficient replacement of HTML elements
+products:
+  - workers
+date: 2025-02-03T3:00:00Z
+---
+
+import { Render, TypeScriptExample } from "~/components";
+
+[`HTMLRewriter`](/workers/runtime-apis/html-rewriter) now supports replacing HTML with content from a stream.
```
irvinebroque: speak 2nd person — you can now...

src/content/changelogs-next/2025-02-03-html-rewriter-streaming.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+This allows you to process HTML more efficiently, as content does not have to be loaded into memory before replacements are made.
```
irvinebroque: before and after?

src/content/changelogs-next/2025-02-03-html-rewriter-streaming.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: Rewrite HTML with streaming content
```
irvinebroque: title hard but something about

transform streams of html content

or on the fly or something

irvinebroque (commented):
Think we can do a bit more here to speak directly to use case

ex:

- I have some site, some origin
- I have a Worker in front
- In my Worker, I need to transform the HTML response — for example, I need to insert a `<script>` tag into the page
- Before — I had to fetch the entire response object from the origin, and only after getting the whole response, I could pass that response into `HTMLRewriter`. This made things slower, because it prevented streaming the bytes of the response back to the client, as I receive them from the origin
- After — I can pass the response directly in, and HTMLRewriter can immediately start parsing and handling rewriting content as it receives bytes of the response, and stream that back to the client immediately

src/content/changelogs-next/2025-02-03-html-rewriter-streaming.mdx
mikenomitch: I think it's just before because you would've await'd response.body previously, then passed in the string

mikenomitch:
@irvinebroque - gave it another pass

PR #19585, opened 2025-01-31
Update Calls openapi spec

astroza:
### Summary

Add the new capabilities available in Calls API and some examples as reference to use them

### DIFF
--- a/public/calls/static/calls-api-2024-05-21.yaml
+++ b/public/calls/static/calls-api-2024-05-21.yaml
@@ -22,6 +22,11 @@ paths:
             type: string
           required: true
           description: WebRTC application ID
+        - in: query
+          name: thirdparty
+          schema:
+            type: boolean
+          description: Session is intended to connect to an ICE-lite peer like a third party SFU/server
       responses:
         "201":
           description: Created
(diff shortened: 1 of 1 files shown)


public/calls/static/calls-api-2024-05-21.yaml
```diff
@@ -75,6 +80,61 @@ paths:
...
+                      a=rtpmap:96 VP8/90000
+                      ...
+                    type: offer
+                  tracks:
+                    - location: local
+                      mid: "0"
+                      trackName: mic-1
+                      bidirectionalMediaStream: true
+                      kind: "audio"
+              push_track_without_an_offer:
```
nils-ohlmeier, line 105: This is a little confusing as I initially thought you would have to set an option called `push_track_without_an_offer`. Maybe we can add an additional note here, saying that the not including `sdp` is what triggers this?

nils-ohlmeier (approved):
Cool. Thank you for getting this up-to-date!

public/calls/static/calls-api-2024-05-21.yaml
astroza: There is a description below about generating Calls generating a SDP for you to push a local track

PR #19596, opened 2025-01-31
[Docs Site] Support paths with and without trailing slash in local dev

KianNH:
### Summary

This was highlighted by a chance in our sidebar logic - we shouldn't "support" links like `/workers` and should direct people to `/workers/` as all links and redirects written should use the trailing slash.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -44,7 +44,6 @@ const runLinkCheck = process.env.RUN_LINK_CHECK || false;
 // https://astro.build/config
 export default defineConfig({
 	site: "https://developers.cloudflare.com",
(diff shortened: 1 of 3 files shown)


kodster28 (approved):
Oooh, ty!

This is very annoying b/c otherwise we sometimes have to do double redirects for each version.

I should really set aside time quarterly to review our config options 😓

GregBrimble:
Consider [`html_handling = "force-trailing-slash"`](https://developers.cloudflare.com/workers/static-assets/routing/#html_handling--auto-trailing-slash--force-trailing-slash--drop-trailing-slash--none) also.

PR #19500, opened 2025-01-28
[Workers] Properly spell `compatibility`

vil02:
### Summary

This PR fixes a typo: `compatability` -> `compatibility`. I did not find other occurrences of `compatability`.

I decided not to create an issue for such a _small_ change. If really needed, please let me know.

PR #19609, opened 2025-01-31
[Workflows] Proprely spell `separate`

vil02:
### Summary

This PR fixes a typo. There are no other occurrences of `seperate`.

Similar to #19500.


src/content/docs/workflows/build/rules-of-workflows.mdx
```diff
@@ -102,7 +102,7 @@ Otherwise, your entire Workflow might not be as durable as you might think, and
 ```ts
 export class MyWorkflow extends WorkflowEntrypoint {
 	async run(event: WorkflowEvent<Params>, step: WorkflowStep) {
-		// 🔴 Bad: you're calling two seperate services from within the same step. This might cause
+		// 🔴 Bad: you're calling two separate services from within the same step. This might cause
```
pedrosousa:
```suggestion
		// 🔴 Bad: you are calling two separate services from within the same step. This might cause
```
Let's also remove the contraction.

pedrosousa (approved):

PR #19675, opened 2025-02-03
[Docs Site] Rename /changelog-next/ to /changelog/, /changelog/ to /release-notes/

KianNH:
### Summary

Rename /changelog-next/ to /changelog/, /changelog/ to /release-notes/


kodster28 (approved):

irvinebroque (approved):

GregBrimble (approved):

src/pages/release-notes/index.xml.ts
```diff
@@ -77,9 +77,9 @@ export const GET: APIRoute = async (context) => {
 	});
 
 	return rss({
-		title: `Cloudflare product changelog`,
+		title: `Cloudflare release notes`,
 		description: `Updates to various Cloudflare products.`,
-		site: "https://developers.cloudflare.com/changelog/",
+		site: "https://developers.cloudflare.com/release-notes/",
```
pedrosousa, line 82: Many different products have individual `changelog` sections. Should we also rename (and redirect) these to `release-notes`?

kodster28: I'll DM you about this.

Broadly though, I think the way we're conceptualizing of how we'd organize this is that release notes would move into the newer "changelog" format, so I'd think (and hope) we can save ourselves the redirects here.

PR #19718, opened 2025-02-04
Made json first in wrangler component

ToriLindsay:
Request from @Oxyjun  (Thank you!)

The `<WranglerConfig>` component is a component which automatically creates a `json` equivalent for the `wrangler.toml` file. However, the component currently shows the `wrangler.toml` tab before `wrangler.json` (i.e. when a user lands on a page, they'll see `wrangler.toml` unless they explicitly choose to switch to `wrangler.json`.
It may be better to show `wrangler.json` first, if we're moving away from using `.toml` (for example, C3 now creates `wrangler.json` by default, so for new users, this will be the more natural option going forward.

### DIFF
--- a/src/components/WranglerConfig.astro
+++ b/src/components/WranglerConfig.astro
@@ -45,10 +45,10 @@ if (language === "toml") {
 ---
 
 <Tabs syncKey="wranglerConfig">
-	<TabItem label="wrangler.toml" icon="setting">
-		<Code lang="toml" code={toml} />
-	</TabItem>
 	<TabItem label="wrangler.json" icon="seti:json">
 		<Code lang="jsonc" code={json} />
+	</TabItem>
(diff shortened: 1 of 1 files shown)


ToriLindsay:
@KianNH Would you mind reviewing?
I assume this is all I needed to do based on Jun's suggestion (in description)? Thanks!

KianNH (approved):
Yep, that's all there is to it!

Before: https://developers.cloudflare.com/ai-gateway/integrations/aig-workers-ai-binding/#2-connect-your-worker-to-workers-ai

image

After: https://tori-pcx15588-make-json-firs.cloudflare-docs-7ou.pages.dev/ai-gateway/integrations/aig-workers-ai-binding/#2-connect-your-worker-to-workers-ai

image

PR #19742, opened 2025-02-04
D1 changelog read permission fix

vy-ton:


src/content/changelogs/d1.yaml
```diff
@@ -5,6 +5,13 @@ productLink: "/d1/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+  - publish_date: "2025-02-04"
+    title: Fixed bug with D1 read-only access via UI and /query REST API.
+    description: |-
+      A bug with D1 permissions, which allowed users with read-only roles via the UI and users with read-only API tokens via the `/query` [REST API](/api/resources/d1/subresources/database/methods/query/) to execute queries that modified databases, is fixed. UI actions via the `Tables` tab, such as creating and deleting tables, were incorrectly allowed with read-only access. However, UI actions via the `Console` tab were not affected by this bug and correctly required write access.
```
vy-ton: @rozenmd Does wrangler use the CF role? If yes, would wrangler have been impacted by this bug

joshthoward: Wrangler would have also been impacted here since they have user scoped tokens.

vy-ton: wranger user scoped tokens are not the same as API tokens though right?

rozenmd: I think wrangler explicitly asks for everything in edit mode, so not an issue

Oxyjun:
```suggestion
      Fixed a bug with D1 permissions which allowed users with read-only roles via the UI and users with read-only API tokens via the `/query` [REST API](/api/resources/d1/subresources/database/methods/query/) to execute queries that modified databases. UI actions via the `Tables` tab, such as creating and deleting tables, were incorrectly allowed with read-only access. However, UI actions via the `Console` tab were not affected by this bug and correctly required write access.
```

Oxyjun (approved):

PR #19873, opened 2025-02-10
thomasgauvin WIP separate wrangler docs

thomasgauvin:


kodster28:
Howdy @irvinebroque, know you had some thoughts on changes like this in the past.

Not sure if here or internally is the best place to have that convo, but wanted to flag.

irvinebroque (changes requested):
Please keep within Workers for now

PR #19631, opened 2025-01-31
[Docs Site] Refactor changelog collection handling

KianNH:
### Summary

Refactor changelog collection handling

PR #19901, opened 2025-02-11
[Docs Site] Load WARP releases into WARP changelog

KianNH:
### Summary

This loads WARP releases into the `entries` of the existing `src/content/changelogs/warp.yaml`.

They will appear in the global changelog, global RSS feed, `/cloudflare-one/changelog/warp/` and `/cloudflare-one/changelog/warp/index.xml`.

The code is duplicated since all of the code in these areas is duplicated - we were going to tackle this in https://github.com/cloudflare/cloudflare-docs/pull/19631/ but shelved that work in favour of changelog-next.

This also fixes the duplication of `date` headers in `ProductChangelog`, so multiple releases for the same day (i.e one for Windows and one for macOS) do not cause 3 identical headings.


KianNH:
Postponing this work until we figure out how to merge release notes and changelogs.

PR #19910, opened 2025-02-11
[Ruleset Engine] Add query param filtering to Fields reference

pedrosousa:
### Summary

Follows a similar strategy to the one in [`ProductCatalog.tsx`](https://github.com/cloudflare/cloudflare-docs/blob/production/src/components/ProductCatalog.tsx).

Adds support for linking to one or more field categories (filtered on page load).


src/components/FieldCatalog.tsx
```diff
@@ -79,6 +92,7 @@ const FieldCatalog = ({ fields }: { fields: Fields }) => {
 								type="checkbox"
 								className="mr-2"
 								value={category}
+								checked={filters.categories.includes(category ?? "")}
```
pedrosousa: Had a TS error here because `category` could be `undefined`. Not sure if there's a better way to handle it (maybe by updating the `Fields` type at the top of the file?).

src/content/fields/index.yaml
```diff
@@ -784,15 +784,15 @@ entries:
 
   - name: cf.tls_client_extensions_sha1
     data_type: String
-    categories: [Request]
+    categories: [Request, SSL/TLS]
```
pedrosousa, line 787: Added a new category so we can filter by mTLS fields + other SSL/TLS fields.
This "double category" is usually expressed in the docs as "the available `cf.tls_*` fields".

src/content/docs/waf/custom-rules/use-cases/require-specific-headers.mdx
```diff
@@ -3,7 +3,7 @@ pcx_content_type: configuration
 title: Require specific HTTP headers
 ---
 
-Many organizations qualify traffic based on the presence of specific HTTP request headers. Use the Rules language [HTTP request header fields](/ruleset-engine/rules-language/fields/reference/) to target requests with specific headers.
+Many organizations qualify traffic based on the presence of specific HTTP request headers. Use the Rules language [HTTP request header fields](/ruleset-engine/rules-language/fields/reference/?field-category=Headers&search-term=http.request) to target requests with specific headers.
```
pedrosousa, line 6: We don't have a category just for request headers, so here we filter by the `Headers` category and the `http.request` search term.

src/components/FieldCatalog.tsx
```diff
@@ -57,6 +57,21 @@ const FieldCatalog = ({ fields }: { fields: Fields }) => {
 		return true;
 	});
 
+	useEffect(() => {
+		// On component load, check for deep-links to categories in the query param
+		const params = new URLSearchParams(window.location.search);
+		const categories = params.getAll("field-category");
```
pedrosousa, line 63: Supports filtering by more than one category (and some links in this PR use this possibility).

src/content/docs/learning-paths/mtls/mtls-app-security/index.mdx
```diff
@@ -63,7 +63,7 @@ Use the values from the previous step.
 
 mTLS is verified and checked in the [Cloudflare WAF phase](/waf/reference/phases/). This is done by creating WAF [Custom Rules](/waf/custom-rules/) using the dynamic fields.
 
-All Client Certificate details can be found in the [`cf.tls_*`](/ruleset-engine/rules-language/fields/reference/) fields in the [Cloudflare Ruleset Engine](/ruleset-engine/).
+All Client Certificate details can be found in the [`cf.tls_*`](/ruleset-engine/rules-language/fields/reference/?field-category=mTLS&field-category=SSL/TLS) fields in the [Cloudflare Ruleset Engine](/ruleset-engine/).
```
pedrosousa, line 66: Filters by two categories.

src/components/FieldCatalog.tsx
KianNH: You could remove `optional()` from `categories` in `src/schemas/fields.ts` if every field is expected to have this, or change `.map((field) => field.categories)` on L28 to `.map((field) => field.categories ?? [])`.

The empty array will be removed by the `flat()` call - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#description

KianNH (approved):
Nice work!

PR #19944, opened 2025-02-12
[Fundamentals] Move Trace request & update redirects

dcpena:
Moved Trace a request outside of Security and added redirects.


public/_redirects
```diff
@@ -614,10 +614,10 @@
 /fundamentals/basic-tasks/prevent-ddos-attacks-external/ /fundamentals/security/prevent-ddos-attacks-external/ 301
 /fundamentals/basic-tasks/protect-your-origin-server/ /fundamentals/security/protect-your-origin-server/ 301
 /fundamentals/basic-tasks/pci-scans/ /fundamentals/security/pci-scans/ 301
-/fundamentals/basic-tasks/trace-request/ /fundamentals/security/trace-request/ 301
-/fundamentals/basic-tasks/trace-request/how-to/ /fundamentals/security/trace-request/how-to/ 301
-/fundamentals/basic-tasks/trace-request/limitations/ /fundamentals/security/trace-request/limitations/ 301
-/fundamentals/basic-tasks/trace-request/changelog/ /fundamentals/security/trace-request/changelog/ 301
+/fundamentals/basic-tasks/trace-request/ /fundamentals/trace-request/ 301
```
pedrosousa, line 617: We've had the new URLs live for at least a couple of days and we've had more than 3,000 pageviews so far, so I'd also redirect the `/fundamentals/security/trace-request/*` URLs to one level above.
I'll add these redirects in a separate commit.

public/_redirects
pedrosousa, line 1: We also need to update a bunch of URLs in different files still pointing to the old location (under `/security/`). I'll also update these in a separate commit.

pedrosousa (approved):

kodster28 (approved):

PR #19985, opened 2025-02-14
[ZT] Import ZT org into Terraform

ranbel:
part of PCX-12683

### DIFF
--- a/src/content/partials/learning-paths/zero-trust/create-zero-trust-org.mdx
+++ b/src/content/partials/learning-paths/zero-trust/create-zero-trust-org.mdx
@@ -15,13 +15,17 @@ To create a Zero Trust organization:
 
 ## (Optional) Manage Zero Trust in Terraform
 
-You can use the [Cloudflare Terraform provider](https://registry.terraform.io/providers/cloudflare/cloudflare/4.40.0) to manage your Zero Trust organization alongside your other IT infrastructure. To get started with Terraform, refer to our [Terraform tutorial series](/terraform/tutorial/).
+You can use the [Cloudflare Terraform provider](https://registry.terraform.io/providers/cloudflare/cloudflare/latest) to manage your Zero Trust organization alongside your other IT infrastructure. To get started with Terraform, refer to our [Terraform tutorial series](/terraform/tutorial/).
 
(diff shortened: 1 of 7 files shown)


src/content/partials/learning-paths/zero-trust/create-zero-trust-org.mdx
```diff
@@ -3,8 +3,52 @@
...
+
+## Sign up for Zero Trust
+
+To create a Zero Trust organization:
 
 <Render file="choose-team-name" product="cloudflare-one" />
+
+## (Optional) Manage Zero Trust in Terraform
+
+You can use the [Cloudflare Terraform provider](https://registry.terraform.io/providers/cloudflare/cloudflare/4.40.0) to manage your Zero Trust organization alongside your other IT infrastructure. To get started with Terraform, refer to our [Terraform tutorial series](/terraform/tutorial/).
```
pedrosousa: We now have version 4.52.0, so linking to a specific version will make the link outdated soon (we're still making minor changes to version 4.x).
I'd link to our provider's [landing page](https://registry.terraform.io/providers/cloudflare/cloudflare/latest), and the new tab names hopefully will clarify to which version the configurations we provide will apply.

src/content/partials/learning-paths/zero-trust/create-zero-trust-org.mdx
```diff
@@ -3,8 +3,56 @@
...
+
+1. Add the following permission to your [`cloudflare_api_token`](https://registry.terraform.io/providers/cloudflare/cloudflare/4.40.0/docs/resources/api_token):
+	- `Access: Organizations, Identity Providers, and Groups Write`
+
+2. Add the [`cloudflare_zero_trust_access_organization`](https://registry.terraform.io/providers/cloudflare/cloudflare/4.40.0/docs/resources/zero_trust_access_organization) resource:
+
+	```terraform
+	resource "cloudflare_zero_trust_access_organization" "<your-team-name>" {
+		account_id                         = var.cloudflare_account_id
+		name                               = "<your-team-name>"
```
tcerqueira-cf: When the account is first configured in the UI, this value seems to be set to `<your-team-name>.cloudflareaccess.com`. I checked this on 2 lab accounts via the API and the value was set as such. In either case, the import will still work with `<your-team-name>`, however, it will show a diff when running the plan (as expected). It might be worth checking with the Access team what the proper value here should be.

In either case, I tested the configuration and it works, even after the name change.

maxvp (approved):

PR #19988, opened 2025-02-14
Workflows: add info that in local dev, pause and resume methods are not available

bruxodasilva:
### Summary

Add info that for Workflows, in local dev, pause and resume methods are not available

Relates to issue [8128](https://github.com/cloudflare/workers-sdk/issues/8128)

Closes WOR-506


src/content/docs/workers/local-development.mdx
```diff
@@ -48,14 +48,16 @@ npx wrangler dev
...
-| Workflows                           | ✅                  | ❌                   |
+| Workflows                           | ✅[^3]              | ❌                   |
 
 With any bindings that are not supported locally, you will need to use the [`--remote` command](#develop-using-remote-resources-and-bindings) in wrangler, such as `wrangler dev --remote`.
 
 [^1]: Using Workers AI always accesses your Cloudflare account in order to run AI models and will incur usage charges even in local development.
 
 [^2]: Using Vectorize always accesses your Cloudflare account to run queries, and will incur usage charges even in local development.
 
+[^3]: `pause()` and `resume()` are not available in local development.
```
irvinebroque: If they're not in local dev, and Workflows doesn't work in remote dev, that means...you just have to write code ship to prod and trust it will work? Or am I missing something?

irvinebroque (commented):
Seems much more discoverable here, no?

https://docs.sentry.io/platforms/javascript/guides/cloudflare/

If I call one of these unsupported methods locally, how do I know it's not supposed to work, without reading the docs?

src/content/docs/workers/local-development.mdx
sidharthachatterjee: @irvinebroque No, you're correct. `pause` and `resume` are not supported locally. However, writing and running a `WorkflowEntrypoint` is. `wrangler dev` on a project that exports a `WorkflowEntrypoint` works and that's what I would classify as local development to assert if something works before shipping to production. 

Pause and resume are unsupported at the moment due to a miniflare limitation (killing the engine DO kills all threads unless I'm mistaken). Pause and resume are also often _not_ in the happy path for p90 Workflows use cases so we are happy with that being a foot note for now until we are unblocked.

sidharthachatterjee:
> Seems much more discoverable here, no?
> 
> https://docs.sentry.io/platforms/javascript/guides/cloudflare/
> 
> If I call one of these unsupported methods locally, how do I know it's not supposed to work, without reading the docs?

Guessing link is incorrect. Using them locally throws an Error documenting that they are unsupported at the moment.

kodster28:
This has conflicts, so closing out.

Feel free to re-open as a new PR if still required.

PR #19996, opened 2025-02-14
[D1] Adding PRAGMA optimize for D1

Oxyjun:
### Documentation checklist


src/content/docs/d1/best-practices/use-indexes.mdx
```diff
@@ -82,6 +82,10 @@ This will return output resembling the below:
 
 Note that you cannot modify this table, or an existing index. To modify an index, [delete it first](#remove-indexes) and [create a new index](#create-an-index) with the updated definition.
 
+## Run `PRAGMA optimize`
```
vy-ton: - Section needs to appear under `create an index`
- Need to explain to user why they need to run `pragma optimize` - explain the sqlite behavior. `D1 uses SQLite's query engine, which ...`

src/content/partials/d1/use-pragma-statements.mdx
```diff
@@ -411,3 +411,11 @@ PRAGMA defer_foreign_keys = off
 ```
 
 Refer to the [foreign key documentation](/d1/sql-api/foreign-keys/) to learn more about how to work with foreign keys.
+
+### `PRAGMA optimize`
+
+Attempts to optimize all schemas in a database by running various maintenance operations.
```
vy-ton: Need to connect `optimize all schemas` with query optimization

src/content/docs/d1/best-practices/use-indexes.mdx
Oxyjun: How much detail do we need to go into explaining the `PRAGMA optimize` behaviour? My understanding of how it works is:

- Obtains statistics on how often indexes are used and drops unused indexes
- Runs `VACUUM` command if necessary to take up less storage 
- Updates internal statistics for future optimizations

I need to check if these are accurate, and whether we want to have it in our own docs rather than referencing out.

@justin-mp do you have any more insight into this?

src/content/partials/d1/use-pragma-statements.mdx
```diff
@@ -411,3 +411,11 @@ PRAGMA defer_foreign_keys = off
 ```
 
 Refer to the [foreign key documentation](/d1/sql-api/foreign-keys/) to learn more about how to work with foreign keys.
+
+### `PRAGMA optimize`
+
+Attempts to [optimize all schemas]() in a database by running various maintenance operations. We recommend running this command after [creating an index](/d1/best-practices/use-indexes/).
```
justin-mp: Really, it should be after any schema change, though after `CREATE INDEX` is probably the most important.

src/content/partials/d1/use-pragma-statements.mdx
```diff
@@ -411,3 +411,11 @@ PRAGMA defer_foreign_keys = off
...
+
+Alternatively, to see all optimizations that would have been performed without actually executing them, run `PRAGMA optimize(-1)`.
```
justin-mp: The way D1 implements `PRAGMA optimize`, debug mode won't work since we constrain `PRAGMA optimize` to not return any rows.

As far as I can tell from reading the source code, debug mode returns a list of `ANALYZE schema_name.table_name` statements that `PRAGMA optimize` would have otherwise run.

@vy-ton I guess we should get debug mode to work, shouldn't we?  Sorry for missing that detail.  It's not documented at all!  I had to read the source to figure it out.

src/content/docs/d1/best-practices/use-indexes.mdx
justin-mp: Do you have a link documenting where `PRAGMA optimize` drops unused indices and runs `VACUUM`?

According to https://www.sqlite.org/lang_analyze.html and https://www.sqlite.org/pragma.html#pragma_optimize, and a cursory scan of the source code, SQLite does none of these things!

From my understanding, `PRAGMA optimize` essentially runs the `ANALYZE` command on all of the tables in the database.  The `ANALYZE` command updates an internal table statistics table.  That's it!

The reason why this is helpful is that the query planner accesses the table statistics table to make better query decisions.

As for the other points:
* For `VACUUM`, we separately auto-vacuum the database.
* DO/D1 doesn't touch the customer's schema, so the customer is in complete control of their indices.

src/content/partials/d1/use-pragma-statements.mdx
Oxyjun: Thanks, I'll add that in.

src/content/partials/d1/use-pragma-statements.mdx
Oxyjun: Can I check which section you're referring to here?

Update: resolved offline.

src/content/docs/d1/best-practices/use-indexes.mdx
Oxyjun: Ah okay, thanks for the explanation - I think I was misunderstanding `ANALYZE` command 😓

https://www.sqlite.org/pragma.html#pragma_optimize -> this section also lists the condition in which `ANALYZE` is run, but I wasn't sure what it did when that condition was **not** met. I.e: Does `PRAGMA optimize` do anything other than run `ANALYZE`?

justin-mp: `PRAGMA optimize` decides if to run `ANALYZE` for each table, and if it's time to run `ANALYZE, `PRAGMA optimize` will run `ANALYZE` with a limit so that the command doesn't take took long.   In the common case (nothing has changed and the table sizes are within 10x of the size they were last time `ANALYZE` was run), `PRAGMA optimize` does nothing.

Oxyjun: Thank you!

marciocloudflare (approved):

src/content/partials/d1/use-pragma-statements.mdx
vy-ton: > I guess we should get debug mode to work, shouldn't we?

@justin-mp Would be nice to close out the `pragma optimize` loop but I wouldn't rank it high given other dev week priorities. let's make sure to at least create a tracking ticket.

src/content/release-notes/d1.yaml
```diff
@@ -5,11 +5,18 @@ productLink: "/d1/"
 productArea: Developer platform
 productAreaLink: /workers/platform/changelog/platform/
 entries:
+  - publish_date: "2025-02-19"
+    title: D1 supports `PRAGMA optimize`
+    description: |-
+      D1 now supports `PRAGMA optimize` command, which can increase your database performance. Refer to [`PRAGMA optimize`](/d1/sql-api/sql-statements/#pragma-optimize) for more information.
```
vy-ton:
```suggestion
      D1 now supports `PRAGMA optimize` command, which can improve database query performance. It is recommended to run this command after a schema change (for example, after creating an index). Refer to [`PRAGMA optimize`](/d1/sql-api/sql-statements/#pragma-optimize) for more information.
```

same change can apply to DO release notes

vy-ton (approved):

src/content/partials/d1/use-pragma-statements.mdx
vy-ton: @Oxyjun Given debug mode doesnt work, I would mention that `PRAGMA optimize(-1)` explicitly is not supported yet

justin-mp: > > I guess we should get debug mode to work, shouldn't we?
> 
> @justin-mp Would be nice to close out the `pragma optimize` loop but I wouldn't rank it high given other dev week priorities. let's make sure to at least create a tracking ticket.

See https://github.com/cloudflare/workerd/pull/3595 for the fix.

PR #20018, opened 2025-02-15
docs: fix dynamic url redirects permission (#1)

techieforfun:
### Summary

I'm using Terraform to configure API Token permissions, but it gives an error for the `Single Redirect Write` permission. After checking the output of

```hcl
data "cloudflare_api_token_permission_groups" "all" {}
```

I found that the correct permission to use is `Dynamic URL Redirects Write`.

### Documentation checklist

- [x ] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.


pedrosousa:
Thank you for your PR, @techieforfun. 
We're addressing this change and some more in #20669.

PR #20022, opened 2025-02-17
Update CORS docs to consume environment variables correctly

helblinglilly:
### Summary

The code sample currently at https://developers.cloudflare.com/cloudflare-one/identity/authorization-cookie/cors/ would always result in an error at runtime as the environment variables are not consumed correctly.

### Screenshots (optional)

https://developers.cloudflare.com/workers/configuration/environment-variables/#add-environment-variables-via-wrangler

### Documentation checklist

### DIFF
--- a/src/content/docs/cloudflare-one/identity/authorization-cookie/cors.mdx
+++ b/src/content/docs/cloudflare-one/identity/authorization-cookie/cors.mdx
@@ -176,7 +176,7 @@ Open `/src/index.js` and delete the existing code and paste in the following exa
 const originalAPIHostname = "api.mysite.com";
 
 export default {
-	async fetch(request) {
+	async fetch(request, env) {
 		// Change just the host. If the request comes in on example.com/api/name, the new URL is api.mysite.com/api/name
 		const url = new URL(request.url);
 		url.hostname = originalAPIHostname;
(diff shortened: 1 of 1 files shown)


ranbel:
Hi @helblinglilly, thanks for the PR! The current code example does not make any assumptions about where you're storing the Client ID and Client Secret. If we want to change the example to consume env variables, I recommend adding a step that explains how to store the Client ID and Client Secret as env variables (potentially after [Step 1](https://github.com/cloudflare/cloudflare-docs/pull/20022/files#diff-57ab364a012765a3b690457b5ff85fc5e9ea9af131b71baad5ce95279ad5c057R129)).

helblinglilly:
Hi @ranbel, thanks for taking a look!

Sorry if I'm misunderstanding something - the code might not make any assumptions _yet_ but under this section: https://developers.cloudflare.com/cloudflare-one/identity/authorization-cookie/cors/#4-configure-the-worker sub-point 5 the docs go on to configure env variables.

I interpreted that as "let's write the worker and populate its values afterwards" - implying that the code snippet would not need modification between these steps.

Feel free to close this PR if you disagree or if I'm in the wrong part of the docs here 😅

ranbel:
Ah, you're right! I didn't see that we configure env variables in a later step. In that case this PR LGTM.

ranbel (approved):

PR #20025, opened 2025-02-17
Add tutorial

harshil1712:
### Summary

Adds MotherDuck tutorial for Pipelines

### DIFF
--- a/src/content/docs/pipelines/tutorials/query-data-with-motherduck/index.mdx
+++ b/src/content/docs/pipelines/tutorials/query-data-with-motherduck/index.mdx
@@ -1,5 +1,5 @@
 ---
-updated: 2025-02-17
+updated: 2025-02-24
 difficulty: Intermediate
 content_type: 📝 Tutorial
 pcx_content_type: tutorial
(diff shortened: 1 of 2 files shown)


src/content/docs/pipelines/tutorials/query-data-with-motherduck/index.mdx
```diff
@@ -0,0 +1,499 @@
...
+```
+
+The `src/index.ts` file does the following:
+
+- Checks if the request is a `POST` request to the `/api/clickstream` endpoint.
+- Extracts the data from the request body.
+- Sends the data to your Pipeline.
+- Returns a response to the client.
+
+You now have a working application that sends clickstream data to your pipeline. In the next step, you will create a pipeline to receive the data.
```
maheshwarip: Typo here? We've already created a pipeline, so at this point, shouldn't we just move on to connecting the R2 bucket to Motherduck?

src/content/docs/pipelines/tutorials/query-data-with-motherduck/index.mdx
```diff
@@ -0,0 +1,499 @@
...
+
+## 6. Ingest data to R2
```
maheshwarip: Hmm I'm not sure what this step is for? We don't actually need the Pipeline ID, and we don't use it anywhere right?

src/content/docs/pipelines/tutorials/query-data-with-motherduck/index.mdx
```diff
@@ -0,0 +1,499 @@
...
+
+```ts title="worker-configuration.d.ts"
+interface Env {
+	PIPELINES_ID: string;
+}
+```
+
+The `fetch` handler in the `src/index.ts` file uses the `PIPELINES_ID` environment variable to send the clickstream data to your pipeline.
+
+## 7. Test the application
```
maheshwarip: I'd add a caveat here. While running this server locally, data won't actually be sent to the Pipeline. But we can verify that it is working.

Additionally: I'd put the deploy step here instead of step 10. Local workers can't actually send data to a pipeline over a worker binding: so the e-commerce site has to be deployed first.

src/content/docs/pipelines/tutorials/query-data-with-motherduck/index.mdx
```diff
@@ -0,0 +1,499 @@
...
+You can connect the bucket to MotherDuck in several ways, which you can learn about from the [MotherDuck documentation](https://motherduck.com/docs/integrations/cloud-storage/cloudflare-r2/). In this tutorial, you will connect the bucket to MotherDuck using the MotherDuck dashboard.
+
+1. Log in to the MotherDuck dashboard and select your profile.
+2. Navigate to the **Secrets** page.
+3. Select the **Add Secret** button and enter the following information:
+
+   - **Secret Name**: `Clickstream pipeline`
+   - **Secret Type**: `Cloudflare R2`
+   - **Access Key ID**: `ACCESS_KEY_ID` (replace with the Access Key ID you obtained in the previous step)
+   - **Secret Access Key**: `SECRET_ACCESS_KEY` (replace with the Secret Access Key you obtained in the previous step)
```
maheshwarip: We don't actually obtain the secret access key above :) let's add a link to the R2 documentation for how to obtain API keys?

maheshwarip (approved):

PR #20042, opened 2025-02-17
In the config doc mention that there are toml<->json online converters

dario-piotrowicz:
### Summary

The changes here simply inform users that there are online converters from toml to json and viceversa

Not something hugely important but I figured it could be small useful tidbit for users

### DIFF
--- a/src/content/docs/workers/wrangler/configuration.mdx
+++ b/src/content/docs/workers/wrangler/configuration.mdx
@@ -16,7 +16,14 @@ Wrangler optionally uses a configuration file to customize the development and d
 
 :::note
 
-As of Wrangler v3.91.0 Wrangler supports both JSON (`wrangler.json` or `wrangler.jsonc`) and TOML (`wrangler.toml`) for its configuration file. Prior to that version, only `wrangler.toml` was supported. The format of Wrangler's configuration file is exactly the same across both languages, except that the syntax is `JSON` rather than `TOML`. Throughout this page and the rest of Cloudflare's documentation config snippets are provided as both JSON and TOML.
+As of Wrangler v3.91.0 Wrangler supports both JSON (`wrangler.json` or `wrangler.jsonc`) and TOML (`wrangler.toml`) for its configuration file.
+
+Prior to that version, only `wrangler.toml` was supported.
+
+The format of Wrangler's configuration file is exactly the same across both languages, only the syntax differs.
+You can also easily switch between the two, there are many online converters for that such as: [toml2jsonConverter](https://esakat.github.io/toml2jsonConverter).
+
+Throughout this page and the rest of Cloudflare's documentation config snippets are provided as both JSON and TOML.
 
 :::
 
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/wrangler/configuration.mdx
```diff
@@ -16,7 +16,14 @@ Wrangler optionally uses a configuration file to customize the development and d
 
 :::note
 
-As of Wrangler v3.91.0 Wrangler supports both JSON (`wrangler.json` or `wrangler.jsonc`) and TOML (`wrangler.toml`) for its configuration file. Prior to that version, only `wrangler.toml` was supported. The format of Wrangler's configuration file is exactly the same across both languages, except that the syntax is `JSON` rather than `TOML`. Throughout this page and the rest of Cloudflare's documentation config snippets are provided as both JSON and TOML.
+As of Wrangler v3.91.0 Wrangler supports both JSON (`wrangler.json` or `wrangler.jsonc`) and TOML (`wrangler.toml`) for its configuration file.
+
+Prior to that version, only `wrangler.toml` was supported.
+
+The format of Wrangler's configuration file is exactly the same across both languages, only the syntax differs.
+You can also easily switch between the two, there are many online converters for that such as: [toml2jsonConverter](https://esakat.github.io/toml2jsonConverter).
```
ToriLindsay:
```suggestion
You can also easily switch between the two. There are many online converters for that such as [toml2jsonConverter](https://esakat.github.io/toml2jsonConverter).
```
Also, just to confirm - This is something we trust directing people to (i.e. it will always be safe and accurate)? If you're not 100% sure, I would suggest changing it to just something like "You can use one of the many available online converters to easily switch between the two."

src/content/docs/workers/wrangler/configuration.mdx
```diff
@@ -16,7 +16,14 @@ Wrangler optionally uses a configuration file to customize the development and d
...
+
+Throughout this page and the rest of Cloudflare's documentation config snippets are provided as both JSON and TOML.
```
ToriLindsay:
```suggestion
Throughout this page and the rest of Cloudflare's documentation, configuration snippets are provided as both JSON and TOML.
```

ToriLindsay (approved):

src/content/docs/workers/wrangler/configuration.mdx
dario-piotrowicz: > This is something we trust directing people to (i.e. it will always be safe and accurate)?

We're not affiliated with them in any way.... so I think that you're actually right, this seems to be inviting unnecessary risk 😕 

I wanted to provide a link so that people can very conveniently just get to a valid tool in one click... but I agree with you that it's most likely not worth the (very small but existent) risk 😕

dario-piotrowicz: Text updated, thank you very much for the suggestion 🙂

dario-piotrowicz:
Thank you very much for the review @ToriLindsay 🫶

PR #20121, opened 2025-02-20
[Docs Site] Add various eslint plugins

KianNH:
### Summary

Adds various eslint plugins such as Tailwind, React Hooks, JSON, YAML and rules from unicorn.


kodster28:
Howdy @KianNH, how exactly do I evaluate this?

More of just a go through and spot check the components / values affected? Or something specific we'd be worried about here?

KianNH:
Going to shelve this since in hindsight I'm not too happy with the Unicorn rules setup

PR #20141, opened 2025-02-20
[Images] Images Bindings

dcpena:
Added docs for Images Bindings to Images and Workers docs.

### DIFF
--- a/src/content/docs/images/transform-images/bindings.mdx
+++ b/src/content/docs/images/transform-images/bindings.mdx
@@ -38,7 +38,7 @@ Within your Worker code, you can interact with this binding by using `env.IMAGES
 
 ### `.transform()`
 
-- Defines how an image should be optimized and manipulated through [`fetch()` options](images/transform-images/transform-via-workers/#fetch-options) such as `width`, `height`, and `blur`.
+- Defines how an image should be optimized and manipulated through [`fetch()` options](/images/transform-images/transform-via-workers/#fetch-options) such as `width`, `height`, and `blur`.
 
 ### `.draw()`
 
@@ -48,7 +48,7 @@ Within your Worker code, you can interact with this binding by using `env.IMAGES
 
 ### `.output()`
 
-* Defines the [output format](images/transform-images/) for the transformed image such as AVIF, WebP, and JPEG.
+* Defines the [output format](/images/transform-images/) for the transformed image such as AVIF, WebP, and JPEG.
 
 For example, to rotate, resize, and blur an image, then output the image as AVIF:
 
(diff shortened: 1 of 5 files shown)


src/content/docs/images/transform-images/bindings.mdx
```diff
@@ -0,0 +1,94 @@
+---
+pcx_content_type: how-to
+title: Bind to Workers
+sidebar:
+  order: 4
+---
+
+A [binding](/workers/runtime-apis/bindings/) connects your [Worker](/workers/) to external resources on the Developer Platform, like [Images](/images/transform-images/transform-via-workers/), [R2 Buckets](/r2/buckets/), or [KV Namespaces](/kv/concepts/kv-namespaces/).
```
maxvp: Nit: Is `R2 Buckets` a proper name?

maxvp (approved):

src/content/docs/images/transform-images/bindings.mdx
```diff
@@ -0,0 +1,94 @@
+---
+pcx_content_type: how-to
+title: Bind to Workers
```
irvinebroque: elsewhere in docs "Workers API"

src/content/docs/images/transform-images/bindings.mdx
dcpena: ah, good catch. Fixed!

src/content/docs/images/transform-images/bindings.mdx
dcpena: Fixed

src/content/docs/images/transform-images/draw-overlays.mdx
```diff
@@ -68,6 +68,46 @@ The `draw` property is an array. Overlays are drawn in the order they appear in
...
+When [interacting with Images through a binding](/images/transform-images/bindings/), the Images API supports a `.draw()` method.
+
+The accepted options for the overlaid image are `opacity`, `repeat`, `top`, `left`, `bottom`, and `right`.
+
+```js
+// Fetch image and watermark
+const img = await fetch('https://example.com/image.png');
+const watermark = await fetch('https://example.com/watermark.png');
+ 
+const response = await env.IMAGES.input(img.body)
```
kristianfreeman, line 82: All the image bindings calls need to be wrapped in parens like in `bindings.mdx` - will error otherwise. Like this:

```ts
const response = (
  await env.IMAGES.input(img.body)
    .transform({ width: 1024 })
    .draw(watermark.body, { "opacity": 0.25, "repeat": true })
    .output({ format: "image/avif" })
).response();
```

kristianfreeman (changes requested):

src/content/docs/images/transform-images/draw-overlays.mdx
dcpena: Thanks for calling this out! Fixed

kristianfreeman (approved):

PR #19956, opened 2025-02-13
[DDoS Protection] flagged traffic clarification

patriciasantaana:
### Summary

Superseded by #20145 

### Documentation checklist

- [ ] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.

PR #20145, opened 2025-02-20
[DDoS Protection] Flagged traffic clarification

patriciasantaana:
### Summary

Supersedes #19956 

### Documentation checklist


src/content/docs/ddos-protection/managed-rulesets/adaptive-protection.mdx
```diff
@@ -70,10 +66,17 @@ To view traffic flagged by L3/4 Adaptive DDoS Protection rules:
...
+3. Filter by rule ID.
 
 You may also obtain information about flagged traffic through [Logpush](/logs/about/) or the [GraphQL API](/analytics/graphql-api/).
 
+To determine if it is safe to enable an adaptive rule in mitigation:
+
+- If no packets match the rule (in the last 7 days or 24 hours), you should consider changing the rule from `log` to `block`.
+- If packets do match the rule, verify that the traffic matches and decide if it is valid. 
+
+The default rule action for `log` with a sensitivity set to `high` will only show packets with suspected attack traffic over internal `high` thresholds in your logs. For instance, if you set the threshold to `medium` or `low` then only packets over those thresholds will be logged.
```
patriciasantaana:
```suggestion
The default rule action for `log` with a sensitivity set to `high` will only show packets with suspected attack traffic over internal `high` thresholds in your logs. For instance, if you set the threshold to `medium` or `low`, then only packets over those thresholds will be logged.
```

src/content/docs/ddos-protection/managed-rulesets/adaptive-protection.mdx
```diff
@@ -70,10 +66,17 @@ To view traffic flagged by L3/4 Adaptive DDoS Protection rules:
...
 
 You may also obtain information about flagged traffic through [Logpush](/logs/about/) or the [GraphQL API](/analytics/graphql-api/).
 
+To determine if it is safe to enable an adaptive rule in mitigation:
+
+- If no packets match the rule (in the last 7 days or 24 hours), you should consider changing the rule from `log` to `block`.
+- If packets do match the rule, verify that the traffic matches and decide if it is valid. 
+
+The default rule action for `log` with a sensitivity set to `high` will only show packets with suspected attack traffic over internal `high` thresholds in your logs. For instance, if you set the threshold to `medium` or `low`, then only packets over those thresholds will be logged.
+
```
patriciasantaana:
```suggestion
To determine if an adaptive rule fits your traffic in a way that will only mitigate attack traffic and will not cause false positives, review the traffic that is _Logged_ by the adaptive rules.

:::note
You may not see any traffic matching the adaptive rules. This can be because there was no deviation from your traffic profile, so you may want to increase the time range and look for any _Logged_ traffic. Another reason why you may not see _Logged_ traffic by the adaptive rules is that there was not sufficient traffic volume to generate a traffic profile for your zone.
:::

If you do see traffic that was _Logged_ by the adaptive rules, use the dashboard to determine if the traffic matches the characteristics of legitimate users or that of attack traffic. As each Internet property is unique, understanding if the traffic is legitimate requires your understanding of how your legitimate traffic looks. For example, the user agent, source country, headers, query string for HTTP requests, and protocols, ports for L3/4 traffic. 

- In cases where you are certain that the rule is only flagging attack traffic, you should consider creating an override and enabling that rule with a [Managed Challenge](/waf/reference/cloudflare-challenges/#managed-challenge-recommended) or `Block` action.
- In cases where you see legitimate traffic being flagged, you should lower the sensitivity level of the rule and observe the flagged traffic. You can continue reducing the sensitivity level until you reach a point where legitimate traffic is not flagged. Then, you should create an override to enable the rule with a mitigation action.
- If the rule is still flagging legitimate traffic you can consider using the expression filters to condition the rules to exclude certain types of traffic.

The default rule action for `log` with a sensitivity set to `high` will only show packets or requests with suspected attack traffic over internal `high` thresholds in your logs. For instance, if you set the threshold to `medium` or `low`, then only packets over those thresholds will be logged.

```

src/content/docs/ddos-protection/managed-rulesets/adaptive-protection.mdx
```diff
@@ -70,10 +66,24 @@ To view traffic flagged by L3/4 Adaptive DDoS Protection rules:
...
 
 You may also obtain information about flagged traffic through [Logpush](/logs/about/) or the [GraphQL API](/analytics/graphql-api/).
 
+To determine if an adaptive rule fits your traffic in a way that will only mitigate attack traffic and will not cause false positives, review the traffic that is _Logged_ by the adaptive rules.
+
+:::note
+You may not see any traffic matching the adaptive rules. This can be because there was no deviation from your traffic profile, so you may want to increase the time range and look for any _Logged_ traffic. Another reason why you may not see _Logged_ traffic by the adaptive rules is that there was not sufficient traffic volume to generate a traffic profile for your zone.
+:::
+
+If you do see traffic that was _Logged_ by the adaptive rules, use the dashboard to determine if the traffic matches the characteristics of legitimate users or that of attack traffic. As each Internet property is unique, understanding if the traffic is legitimate requires your understanding of how your legitimate traffic looks. For example, the user agent, source country, headers, query string for HTTP requests, and protocols, ports for L3/4 traffic.
```
maxvp:
```suggestion
If you do see traffic that was _Logged_ by the adaptive rules, use the dashboard to determine if the traffic matches the characteristics of legitimate users or that of attack traffic. As each Internet property is unique, understanding if the traffic is legitimate requires your understanding of how your legitimate traffic looks. For example, the user agent, source country, headers, query string for HTTP requests, and protocols and ports for L3/4 traffic. 
```

maxvp (approved):

PR #20173, opened 2025-02-21
[Docs Site] Load filters from search params in ModelCatalog

KianNH:
### Summary

This enables loading filters from search parameters, like `/products/` and `/ruleset-engine/rules-language/fields/reference/`, such as `?search=facebook&tasks=Summarization`

### Screenshots (optional)

image

### DIFF
--- a/src/components/ModelCatalog.tsx
+++ b/src/components/ModelCatalog.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import ModelInfo from "./models/ModelInfo";
 import ModelBadges from "./models/ModelBadges";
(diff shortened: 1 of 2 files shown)


kodster28 (approved):
So this is an improvement... but for both this / the fields grid, it looks like we aren't updating the URL when the filters change.

Can track as a separate ask.

pedrosousa:
> So this is an improvement... but for both this / the fields grid, it looks like we aren't updating the URL when the filters change.
> 
> Can track as a separate ask.

+1
Yes, we're still missing this part. Having the possibility of clearing the search box (usually using an "X" at the end of the input box) could also be handy.

PR #20227, opened 2025-02-24
[Docs Site] Add "show more" button to changelogs

KianNH:
### Summary

Add "show more" button to changelogs

### DIFF
--- a/src/pages/changelog/index.astro
+++ b/src/pages/changelog/index.astro
@@ -75,7 +75,7 @@ const props = {
 	<div class="flex items-center justify-center">
 		<button
 			id="changelogs-next-button"
-			class="h-12 rounded bg-cl1-brand-orange px-6 font-medium text-cl1-black"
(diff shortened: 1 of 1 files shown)


bllchmbrs:
@KianNH - We (Hyperlint) just hit a little bug where we inadvertently started checks on PRs without files that we can review (`md`,`mdx`). The bug is fixed but this PR hit it. If you push a new change again, it should go away, but if not. you can safely ignore Hyperlint! Sorry for the inconvenience.

src/pages/changelog/index.astro
```diff
@@ -68,6 +72,14 @@ const props = {
 			);
 		})
 	}
+	<div class="flex items-center justify-center">
+		<button
```
kodster28, line 76: Not getting the mouse seeing the `click on hover` animation.

Not a blocker! But figured I'd call it out.

kodster28 (approved):

PR #20232, opened 2025-02-24
[Docs Site] Track Linux releases with WARP script

KianNH:
### Summary

Track Linux releases with WARP script.

Adds "no items" logic for tracks that don't yet have any releases, but will in the future.

### DIFF
--- a/bin/fetch-warp-releases.js
+++ b/bin/fetch-warp-releases.js
@@ -2,7 +2,14 @@ import fs from "fs";
 import YAML from "yaml";
 import { marked } from "marked";
 
-const tracks = ["windows/ga", "windows/beta", "macos/ga", "macos/beta"];
+const tracks = [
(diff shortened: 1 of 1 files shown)


haleycode (approved):

bin/fetch-warp-releases.js
```diff
@@ -11,11 +18,23 @@ const linesToRemove = [
...
+			if (!data.items) {
+				console.warn(
+					`${track} has no releases: ${JSON.stringify(data, null, 2)}`,
+				);
+
+				return;
+			}
+
 			data.items.forEach((item) => {
+				if (track.startsWith("noble-intel")) {
```
WalshyDev, line 34: that's fun..  surprised api doesn't just use linux

WalshyDev (approved):

PR #20348, opened 2025-02-27
[DO] Clarifying DO creation behaviour with location hint

Oxyjun:
### Documentation checklist

### DIFF
--- a/src/content/docs/durable-objects/reference/data-location.mdx
+++ b/src/content/docs/durable-objects/reference/data-location.mdx
@@ -72,7 +72,9 @@ let durableObjectStub = OBJECT_NAMESPACE.get(id, { locationHint: "enam" });
 ```
 
 :::caution
-If the location hint cannot be respected when creating a Durable Object, the Durable Object creation will fail. 
+If the location hint cannot be respected when creating a Durable Object, the Durable Object creation will fail.
+
+This is a [retryable](/durable-objects/best-practices/error-handling/) error. If this happens, retry the operation.
 :::
 
 ### Supported locations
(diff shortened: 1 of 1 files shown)


src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -72,8 +72,9 @@ let durableObjectStub = OBJECT_NAMESPACE.get(id, { locationHint: "enam" });
 ```
 
 :::caution
+If the location hint cannot be respected when creating a Durable Object, the Durable Object creation will fail.
 
-Hints are a best effort and not a guarantee. Unlike with jurisdictions, Durable Objects will not necessarily be instantiated in the hinted location, but instead instantiated in a data center selected to minimize latency from the hinted location.
+This is a [retryable](/durable-objects/best-practices/error-handling/) error. If this happens, retry the operation.
```
Frederik-Baetens, line 77: I'd take this out of a caution block, since this is a good thing. I'd just get rid of this block together, and add it to one of the paragraphs above. E.g.:

"To manually create Durable Objects in another location, provide an optional locationHint parameter to get(). The Durable Object is guaranteed to be created in the specified region. ..."

This seems better to me, because the "If the location hint cannot be respected" failure condition doesn't really exist in our codebase as it stands. We don't go over all possible locations in a region to store your Durable Object. We just try a location, and if it fails, your request fails in the same way that any Durable Object request can fail. That's also why I would drop the bit about a retryable operation. Not all errors you get when specifying a location hint are retryable. E.g. you can get an overloaded exception back, just like you can without a location hint, and those should generally not be retried.

Oxyjun:
I've had a look at the original request this PR, and I think that this PR is actually good to close. Thanks all!

PR #20472, opened 2025-03-03
[Docs Site] Add LLM Resources into sidebar for DevPlat

KianNH:
### Summary

Add LLM Resources into sidebar for Developer Platform products

### Screenshots (optional)

image

### DIFF
--- a/src/content/docs/workers/llms.txt.md
+++ b/src/content/docs/workers/llms.txt.md
@@ -1,6 +1,0 @@
----
-title: llms.txt
-sidebar:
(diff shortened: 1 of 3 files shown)


kodster28 (approved):
There's a small risk to internal feedback on this b/c it's adding something to every product sidebar in a space... but I'd rather get that feedback retroactively.

This makes things a lot better + standardized as a result of a lot of feedback we've received, as well as minimizing the level of impact b/c it's at the bottom of each product nav

PR #20535, opened 2025-03-05
Calls: Update API diagram to Mermaid and Improve DataChannel page

renandincer:


src/content/docs/calls/datachannels.mdx
```diff
@@ -3,10 +3,32 @@ pcx_content_type: get-started
...
+```
+
+DataChannels on Cloudflare Calls can scale up to many subscribers per publisher, there is no limit to the number of subscribers per publisher.
+
+### How to use DataChannels
+
+1. Create a two Calls sessions, one for the publisher and one for the subscribers.
+2. Create a DataChannel by calling /datachannels/new with the location set to "local" and the dataChannelName set to the name of the DataChannel.
+3. Create a DataChannel by calling /datachannels/new with the location set to "remote" and the sessionId set to the sessionId of the publisher.
+4. Use the DataChannel to send and receive data.
```
nils-ohlmeier: Since our data channels are not bi-directional yet I think this should be worded better. Maybe something like "Use the DataChannel on the publisher side to send data and receive the data on the subscriber side".

nils-ohlmeier (approved):
One clarification suggested.

src/content/docs/calls/datachannels.mdx
renandincer: yep! fair. fixing.

```suggestion
4. Use the DataChannel to send data from the publisher to the subscribers.
```

renandincer: appreciate the review!

PR #20539, opened 2025-03-05
Enumerate Node.js APIs on main Node.js docs page

irvinebroque:
We don't currently have one place where you can look at-a-glance at what Node.js APIs are supported in Workers.

The compatibility matrix here is very cool, but too detailed for a quick reference on what is or isn't supported:

https://workers-nodejs-compat-matrix.pages.dev/

The clearest thing we can do is to provide a simple list of all non-deprecated, non-experimental APIs from Node.js (pulled this straight from the Node.js docs, in the same alphabetical order) — and state clearly which are supported, which we have partial support for, and which are not supported. And then link to our own relevant docs page.

There are a few here that are currently supported, but we haven't yet added a docs page for. We'll need to follow up on this, so that the sidebar includes all of the Node.js APIs we currently support in the Workers Runtime.

### DIFF
--- a/src/content/docs/workers/runtime-apis/nodejs/index.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/index.mdx
@@ -12,14 +12,11 @@ When you write a Worker, you may need to import packages from [npm](https://www.
 Cloudflare Workers provides a subset of Node.js APIs in two forms:
 
 1. As built-in APIs provided by the Workers Runtime
-2. As polyfills that [Wrangler](/workers/wrangler/) adds to your Worker's code
-
-You can view which APIs are supported using the [Node.js compatibility matrix](https://workers-nodejs-compat-matrix.pages.dev).
+2. As polyfill shim implementations that [Wrangler](/workers/wrangler/) adds to your Worker's code, allowing it to import the module, but calling API methods will throw errors.
 
 ## Get Started
 
-To enable built-in Node.js APIs and add polyfills, you need to add the `nodejs_compat` compatibility flag to your wrangler configuration. This also enables `nodejs_compat_v2` as long as your compatibility date is 2024-09-23 or later. [Learn more about the Node.js compatibility flag and v2](/workers/configuration/compatibility-flags/#nodejs-compatibility-flag).
-
+To enable built-in Node.js APIs and add polyfills, add the `nodejs_compat` compatibility flag to your [wrangler configuration file](/workers/wrangler/configuration/), and ensure that your Worker's [compatibility date](/workers/configuration/compatibility-dates/) is 2024-09-23 or later. [Learn more about the Node.js compatibility flag and v2](/workers/configuration/compatibility-flags/#nodejs-compatibility-flag).
 
 <WranglerConfig>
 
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/runtime-apis/nodejs/index.mdx
```diff
@@ -30,38 +27,78 @@ compatibility_date = "2024-09-23"
...
+| Performance hooks                                                                    | 🟡 partially supported                    |
+| Process                                                                              | 🟢 supported                              |
+| Query strings                                                                        | 🟢 supported                              |
+| Readline                                                                             | ⚪ not supported                          |
+| Stream                                                                               | 🟢 supported                              |
+| [String decoder](/workers/runtime-apis/nodejs/string-decoder/)                       | 🟢 supported                              |
+| Test runner                                                                          | ⚪ not supported                          |
+| [Timers](/workers/runtime-apis/nodejs/timers/)                                       | 🟢 supported                              |
+| TLS/SSL                                                                              | ⚪ not supported                          |
```
anonrig:
```suggestion
| TLS/SSL                                                                              | 🟢 supported                          |
```

irvinebroque:
@anonrig @jasnell any edits? this accurate?

We can have auto-generated page too — but need something in our own docs that's digestable at a glance

src/content/docs/workers/runtime-apis/nodejs/index.mdx
```diff
@@ -30,38 +27,78 @@ compatibility_date = "2024-09-23"
...
 </WranglerConfig>
 
-## Built-in Node.js Runtime APIs
-
-The following APIs from Node.js are provided directly by the Workers Runtime when either `nodejs_compat` or `nodejs_compat_v2` are enabled:
+## Supported Node.js APIs
+
+The runtime APIs from Node.js listed below as "🟢 supported" are currently natively supported in the Workers Runtime.
+
+[Deprecated, experimental, and not yet stable APIs from Node.js](https://nodejs.org/docs/latest/api/documentation.html#stability-index) are not included as part of the list below:
```
anonrig:
```suggestion
[Deprecated or experimental APIs from Node.js](https://nodejs.org/docs/latest/api/documentation.html#stability-index) are not included as part of the list below:
```

anonrig (approved):

src/content/docs/workers/runtime-apis/nodejs/index.mdx
jasnell: For modules like tls and net where we intentionally have only partial support, how do we want to express it?

jasnell (approved):

src/content/docs/workers/runtime-apis/nodejs/index.mdx
irvinebroque: `🟡 partially supported ` + link to the page in our docs that explains (think we need a page for this)

irvinebroque:
```suggestion
| TLS/SSL                                                                              | 🟡 partially supported                         |
```

PR #20549, opened 2025-03-05
Update agents "routeAgentRequest" doc

deathbyknowledge:
### Summary

Update the `routeAgentRequest` description, as it's inaccurate. I tried to make it work for a while with the name of the JavaScript class but couldn't make it work. I had a look at the [routing helper code](
https://github.com/threepointone/partyserver/blob/main/packages/partyserver/src/index.ts#L112-L124) and it is using the `key` of the `env` object used by the agent (effectively the binding name given in the wrangler config).

Perhaps worth adding a note in the config docs too since users might be used to naming their bindings in uppercase?

### Documentation checklist


whoiskatrin:
if you fix the conflict, @deathbyknowledge, happy to merge

whoiskatrin (approved):

PR #20557, opened 2025-03-05
[Docs Site] Create TutorialsGrid component

KianNH:
### Summary

Create TutorialsGrid component

### Screenshots (optional)

image

### DIFF
--- a/src/components/TutorialsGrid.astro
+++ b/src/components/TutorialsGrid.astro
@@ -35,7 +35,7 @@ const productEntry = await getEntry(product);
 
 if (!productEntry) {
 	throw new Error(
-		`[ListTutorials] Unable to find product YAML for ${product.id}.`,
+		`[TutorialsGrid] Unable to find product YAML for ${product.id}.`,
 	);
 }
 
--- a/src/components/TutorialsGrid.astro
+++ b/src/components/TutorialsGrid.astro
@@ -1,0 +1,151 @@
+---
+import { z } from "astro:schema";
+import {
+	getCollection,
+	getEntry,
+	reference,
+	type CollectionEntry,
+} from "astro:content";
+import type { ComponentProps } from "astro/types";
+import { Badge } from "@astrojs/starlight/components";
+
+type Props = z.infer<typeof props>;
+type Badge = ComponentProps<typeof Badge>;
+
+const props = z.object({
+	product: reference("products").optional(),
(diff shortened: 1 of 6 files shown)


src/components/TutorialsGrid.astro
```diff
@@ -0,0 +1,151 @@
...
+		id: Astro.url.pathname.split("/").filter(Boolean)[0],
+		collection: "products",
+	};
+}
+
+const productEntry = await getEntry(product);
+
+if (!productEntry) {
+	throw new Error(
+		`[ListTutorials] Unable to find product YAML for ${product.id}.`,
```
kodster28:
```suggestion
		`[TutorialsGrid] Unable to find product YAML for ${product.id}.`,
```

src/content/docs/workers/tutorials/index.mdx
```diff
@@ -7,12 +7,12 @@ sidebar:
   order: 4
 ---
 
-import { GlossaryTooltip, ListTutorials } from "~/components";
+import { GlossaryTooltip, TutorialsGrid } from "~/components";
```
kodster28: Maybe we launch this on an area that's a bit lower viz than the main `/workers/`?

Thinking something like `/ai-gateway/` and `/workers-ai/`?

src/components/TutorialsGrid.astro
kodster28:
```suggestion
		`[TutorialsGrid] Unable to find product YAML for ${product.id}.`,
```

src/components/TutorialsGrid.astro
```diff
@@ -0,0 +1,151 @@
...
+					<p class="!mt-auto">
+						{badges.map((badge) => (
+							<Badge {...badge} />
+						))}
+					</p>
+				</a>
+			);
+		})
+	}
+</div>
```
kodster28, line 152: Feedback on the look / design based on previous things I've heard and experienced.

- Any way we're able to more clearly indicate which is an external resource vs hosted in our docs? Specific case for videos b/c some are hosted on YouTube. I suppose we could also solve this another way, i.e., centralizing video files somewhere and creating standalone pages for each Stream / YouTube entry.
- Is there a better way to organize the pills so the info is less jumbled? I find myself sorta glazing over all of them b/c it's hard to pick out which one applies to which situation.

PR #20214, opened 2025-02-24
Workers react framework guide

ToriLindsay:
WIP

Fixes #19653

PR #20584, opened 2025-03-06
Add Workers Assets + Vue framework guide

emily-shen:
### Summary

Essentially the same as #20214 :) 

cc @ToriLindsay - I've added the section on bindings back in, as I think its useful to have. If it looks okay we can add it to the React guide as well?

### DIFF
--- a/src/content/docs/workers/frameworks/framework-guides/vue.mdx
+++ b/src/content/docs/workers/frameworks/framework-guides/vue.mdx
@@ -64,6 +64,6 @@ You can serve static assets in your Vue application by [placing them in the `./p
 
 <Render file="workers-assets-routing-summary" />
 
-## Bindings
+## Use bindings with Vue
 
-Your new project also contains a Worker (at `./server/index.ts`), which you can use as a backend API for your Vue application. From this Worker, you can use bindings to access resources on the Cloudflare Developer Platform. Learn how to [configure Workers bindings](/workers/runtime-apis/bindings/).
+Your new project also contains a Worker at `./server/index.ts`, which you can use as a backend API for your Vue application. While your Vue application cannot directly access Workers bindings, it can interact with them through this Worker. You can make [`fetch()` requests](/workers/runtime-apis/fetch/) from your Vue application to the Worker, which can then handle the request and use bindings. Learn how to [configure Workers bindings](/workers/runtime-apis/bindings/).
--- a/src/content/docs/workers/frameworks/framework-guides/vue.mdx
+++ b/src/content/docs/workers/frameworks/framework-guides/vue.mdx
@@ -1,0 +1,69 @@
+---
+pcx_content_type: how-to
+title: Vue
+head: []
+description: Create a Vue application and deploy it to Cloudflare Workers with Workers Assets.
+---
+
+import {
+	Badge,
+	Description,
+	InlineBadge,
+	Render,
+	PackageManagers,
+} from "~/components";
+
+In this guide, you will create a new [Vue](https://vuejs.org/) application and deploy to Cloudflare Workers (with the new [<InlineBadge preset="beta" /> Workers Assets](/workers/static-assets/)).
+
+## 1. Set up a new project
+
+Use the [`create-cloudflare`](https://www.npmjs.com/package/create-cloudflare) CLI (C3) to set up a new project. C3 will create a new project directory, use code from the official Vue template, and provide the option to deploy instantly.
+
+To use `create-cloudflare` to create a new Vue project with <InlineBadge preset="beta" /> Workers Assets, run the following command:
+
+<PackageManagers
+	type="create"
+	pkg="cloudflare@latest my-vue-app"
+	args={"--framework=vue --platform=workers"}
+/>
+
+<Render
+	file="c3-post-run-steps"
+	product="workers"
+	params={{
+		category: "web-framework",
+		framework: "Vue",
+	}}
+/>
+
+After setting up your project, change your directory by running the following command:
+
+```sh
+cd my-vue-app
+```
+
+## 2. Develop locally
+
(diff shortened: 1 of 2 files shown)


ToriLindsay:
> ### Summary
> Essentially the same as #20214 :)
> 
> cc @ToriLindsay - I've added the section on bindings back in, as I think its useful to have. If it looks okay we can add it to the React guide as well?

Hey @emily-shen ! 
I had removed Bindings from the React framework guide based on @petebacondarwin 's recommendation. But I think that's just because React is different from the other frameworks. (See Pete's comments in #20214 ). Assuming Vue is more like the other frameworks, then it would make sense to include Bindings. However, perhaps you can use the same text that is in the other framework guides for consistency? I'll go ahead and make that suggestion here inline as well, while I review.

src/content/docs/workers/frameworks/framework-guides/vue.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Static assets
+
+You can serve static assets in your Vue application by [placing them in the `./public/` directory](https://vite.dev/guide/assets#the-public-directory). This can be useful for resource files such as images, stylesheets, fonts, and manifests.
+
+<Render file="workers-assets-routing-summary" />
+
+## Bindings
+
+Your new project also contains a Worker (at `./server/index.ts`), which you can use as a backend API for your Vue application. From this Worker, you can use bindings to access resources on the Cloudflare Developer Platform. Learn how to [configure Workers bindings](/workers/runtime-apis/bindings/).
```
ToriLindsay:
```suggestion
Your Vue application cannot directly access Workers bindings, but it can interact with them through a Worker backend API at `./server/index.ts`. You can make [`fetch()` requests](/workers/runtime-apis/fetch/) from your Vue application to the Worker, which can then handle the request and connect to bindings. Learn how to [configure Workers bindings](/workers/runtime-apis/bindings/).
```
@emily-shen
To answer your comment, I removed Bindings from the React framework guide per @petebacondarwin 's recommendation, because React is different from the other frameworks in that it doesn't connect with Workers directly (if I understand correctly) and it's simpler without mentioning Bindings. But it seems like maybe Vue is similar to React in that way? 
If so, here is what I had written for React (adapted for Vue) before I saw Pete's comment about deleting it. Let me know what you think. If this works, it's up to you / @petebacondarwin if you think we should add it into both guides? (And also Gatsby + other similar frameworks?)

src/content/docs/workers/frameworks/framework-guides/vue.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+---
+
+## Static assets
+
+You can serve static assets in your Vue application by [placing them in the `./public/` directory](https://vite.dev/guide/assets#the-public-directory). This can be useful for resource files such as images, stylesheets, fonts, and manifests.
+
+<Render file="workers-assets-routing-summary" />
+
+## Bindings
```
ToriLindsay:
```suggestion
## Connect Vue with Workers bindings
```

ToriLindsay:
> ### Summary
> Essentially the same as #20214 :)
> 
> cc @ToriLindsay - I've added the section on bindings back in, as I think its useful to have. If it looks okay we can add it to the React guide as well?

Answered as inline suggestion

src/content/docs/workers/frameworks/framework-guides/vue.mdx
emily-shen: Yup the React and Vue templates are _very_ similar so whatever we end up with should apply to both. I meant to chip in the react PR but saw it was getting merged, thought it would be easier to followup here - sorry about the confusion 😅 

It's definitely simpler to omit bindings, but imo it's also less complete. We _could've_ created a template that didn't have a Worker and only React/Vue, but since we _did_ include a Worker for a backend, it seems like we should explain that here in the docs, along with a bit of info about how to set up bindings. The suggestion you have about sounds great to me for that purpose.

This wouldn't apply to Gatsby etc. though, because those template _don't_ include a Worker.

ToriLindsay: @emily-shen 
Awesome. Thanks for the clarification! If we have time on this, I'll wait for @petebacondarwin to confirm that we can add it to React as well so we're all aligned.

emily-shen: A nit about the phrasing of "Workers backend API" - someone could confuse it with a Workers _platform_ API, rather than a worker being used for a backend? 

what about:

```suggestion
Your new project also contains a Worker at `./server/index.ts`, which you can use as a backend API for your Vue application. While your Vue application cannot directly access Workers bindings, it can interact with them through this Worker. You can make [`fetch()` requests](/workers/runtime-apis/fetch/) from your Vue application to the Worker, which can then handle the request and connect to bindings. Learn how to [configure Workers bindings](/workers/runtime-apis/bindings/).
```

petebacondarwin: This section looks good. Feel free to roll it out to the React doc too. Either in this PR or we/I can create a new PR for React.

petebacondarwin (approved):

emily-shen:
Updated the react doc here: [64a5a5c](https://github.com/cloudflare/cloudflare-docs/pull/20584/commits/64a5a5cc8532b3d2d2794fe5d84335d4cbb976a9)

petebacondarwin (approved):

ToriLindsay (approved):
@emily-shen Approved.

emily-shen:
@ToriLindsay added the section to the react template already in [64a5a5c](https://github.com/cloudflare/cloudflare-docs/pull/20584/commits/64a5a5cc8532b3d2d2794fe5d84335d4cbb976a9) :)

Thanks for all your help!

I'll merge this next Tuesday when our next c3 release is out

ToriLindsay (approved):

PR #20592, opened 2025-03-06
[Docs Site] Re-design /search/ and add view more results footer to searchbox

KianNH:
### Summary

Re-design /search/ and add view more results footer to searchbox

### Screenshots (optional)

image


kodster28 (approved):
🚢 🚢 🚢 

Tested current experience and seems like:
1. Improving what currently looks broken
2. Providing an outlet for folks who want `/search` for full SERP, but still conservative so we're not harming the on-page search experience.

PR #20640, opened 2025-03-07
[Chore] Tag cleanup

kodster28:
### Summary

Remove unused tags / convert some to `products` value where appropriate


patriciasantaana (approved):

src/content/docs/d1/examples/query-d1-from-python-workers.mdx
```diff
@@ -3,7 +3,6 @@ type: example
 summary: Learn how to query D1 from a Python Worker
 tags:
```
kodster28:
```suggestion
languages:
```

PR #20688, opened 2025-03-10
[Docs Site] WIP - Add Page Tag infra

kodster28:


src/components/PageTags.astro
```diff
@@ -0,0 +1,39 @@
...
+import { componentToString } from "~/util/container";
+import { tagsValues } from "~/schemas/tags";
+
+type Props = z.infer<typeof props>;
+
+const props = z.object({
+	tags: z.enum(tagsValues).array().optional(),
+});
+
+const parsedValues = props.safeParse(Astro.props);
```
kodster28: Right now, can only get this to either pass or fail.

Maybe I'm just trying to be too cute with zod. Is there a better way to make it filter out the unused values?

PR #20782, opened 2025-03-13
Update public-buckets.mdx with `partial` zone restriction

kszarlej:
### Summary

It seems that R2 custom domain doesn't work when the record is created within a zone that uses partial DNS setup.

Tested using the web console and using the [r2_bucket_custom_domain](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/r2_custom_domain) Terraform resource. The error in both cases is non-descriptive - Terraform fails with saying that `zone_id` is `incorrect`, and webconsole configures the domain, shows it as green in the R2 bucket settings, but it doesn't work. 

On a non-partial domain both TF resource and webconsole works. I used a subdomain delegation setup


Oxyjun:
Thanks for this PR @kszarlej :D

I realised we're running into a merge conflict, so I've spun another PR to add in the change you're trying to make.

Now tracking here https://github.com/cloudflare/cloudflare-docs/pull/22011

PR #20786, opened 2025-03-13
[Chore] Remove 'content_type' property b/c we can infer it

kodster28:
### Summary

Should result in no changes to the customer-facing experience. Removes `content_type` as a frontmatter property, inferring this from the collection it comes from.

This only affects the `ListTutorials` component.

### DIFF
--- a/templates/tutorial-template.mdx
+++ b/templates/tutorial-template.mdx
@@ -1,7 +1,6 @@
 ---
 updated: YYYY-MM-DD
 difficulty: Beginner | Intermediate | Expert
-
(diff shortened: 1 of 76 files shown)


templates/tutorial-template.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 updated: YYYY-MM-DD
 difficulty: Beginner | Intermediate | Expert
-content_type: 📝 Tutorial
+
```
KianNH:
```suggestion
```

KianNH (approved):

PR #20792, opened 2025-03-13
Clarify Wrangler esbuild update policy

penalosa:
### Documentation checklist

### DIFF
--- a/src/content/docs/workers/wrangler/bundling.mdx
+++ b/src/content/docs/workers/wrangler/bundling.mdx
@@ -7,6 +7,13 @@ description: Review Wrangler's default bundling.
 
 By default, Wrangler bundles your Worker code using [`esbuild`](https://esbuild.github.io/). This means that Wrangler has built-in support for importing modules from [npm](https://www.npmjs.com/) defined in your `package.json`. To review the exact code that Wrangler will upload to Cloudflare, run `npx wrangler deploy --dry-run --outdir dist`, which will show your Worker code after Wrangler's bundling.
 
+<details>
+<summary>`esbuild` version</summary>
+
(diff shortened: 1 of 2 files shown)


src/content/docs/workers/wrangler/bundling.mdx
```diff
@@ -7,6 +7,13 @@ description: Review Wrangler's default bundling.
 
 By default, Wrangler bundles your Worker code using [`esbuild`](https://esbuild.github.io/). This means that Wrangler has built-in support for importing modules from [npm](https://www.npmjs.com/) defined in your `package.json`. To review the exact code that Wrangler will upload to Cloudflare, run `npx wrangler deploy --dry-run --outdir dist`, which will show your Worker code after Wrangler's bundling.
 
+<details>
+<summary>`esbuild` version</summary>
+
+Wrangler uses `esbuild` v0.24. We periodically update the `esbuild` version included with Wrangler, and since `esbuild` is a pre-1.0.0 tool, this may sometimes include breaking changes to how bundling works. In particular, we may bump the `esbuild` version in a Wrangler minor version.
```
lrapoport-cf:
```suggestion
Wrangler uses `esbuild`. We periodically update the `esbuild` version included with Wrangler, and since `esbuild` is a pre-1.0.0 tool, this may sometimes include breaking changes to how bundling works. In particular, we may bump the `esbuild` version in a Wrangler minor version.
```
i'd suggest not calling out the specific version here, that way we don't need to update this location every time we bump esbuild while still in pre-1.0.0

i also think this is worth calling out in the migration guide so it's up front info for people who are migrating

lrapoport-cf (approved):

PR #20875, opened 2025-03-17
[API Shield] Add `cf-llm` managed label

pedrosousa:
### Summary

Addresses PCX-15177.

### DIFF
--- a/src/content/docs/api-shield/management-and-monitoring/endpoint-labels.mdx
+++ b/src/content/docs/api-shield/management-and-monitoring/endpoint-labels.mdx
@@ -5,10 +5,9 @@ title: Endpoint labeling service
 sidebar:
   order: 1
   label: Labeling service
-
 ---
 
-import { Render } from "~/components"
+import { Render } from "~/components";
 
 API Shield's labeling service will help you organize your endpoints and address vulnerabilities in your API. The labeling service comes with managed and user-defined labels.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/api-shield/management-and-monitoring/endpoint-labels.mdx
```diff
@@ -46,6 +45,8 @@ Use managed labels to identify endpoints by use case. Cloudflare may automatical
 
 `cf-rss-feed`: Add this label to endpoints that expect traffic from RSS clients.
 
+`cf-llm`: Services that are (partially) powered by Large Language Model (LLM).
```
caley-b, line 48: This might be a newbie question, so apologies in advance: Are we needing to callout the partial aspect via parentheses? I'm in favor of dropping the parentheses. Feel free to ignore this comment and proceed if there's a reason the parentheses make sense in this circumstance.

caley-b (changes requested):
Everything looks good. Just want insight into the use of parentheses on one line.

src/content/docs/api-shield/management-and-monitoring/endpoint-labels.mdx
pedrosousa: Actually this is a very specific case: the descriptions currently match exactly the label tooltips in the dashboard. My original proposal was a bit different, but the PM asked me to match the dashboard description, as it was previously done for other labels.

caley-b: Ah, gotcha! Thanks for the quick insight.

caley-b (approved):

PR #20977, opened 2025-03-20
[Terraform/Advanced Topics] improve documentation in `import-cloudflare-resources.mdx`

gui-wf:
### Summary
Simplify the documentation about using cf-terraforming to generate and import some resources.

#### Why
With this changes, I hope to make the documentation easier to follow, more compact and to bring a bit more knowledge to the reader - simple use case for a bit of shell scripting to automate things.

#### How/What was done

This PR changes the explanation of importing resources to the TF state:
- From a very repetitive and dull manual process of running multiple by-hand `terraform import ...` commands **to** saving the newline-separated output of `cf-terraforming import` into a file.
- Then we instruct the reader to run the file as a shell script.

### DIFF
--- a/src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
+++ b/src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
@@ -212,27 +212,29 @@ To fix this, you must import the real state of those resources from Cloudflare i
 
 When you run `cf-terraforming import ...`, you will obtain a list of `terraform import ...` commands that you must run manually afterward to import those resources into Terraform state. This is currently a manual process, but it may be automated in the future.
 
+One way to ease the process of executing all the `terraform import`, is saving the result from `cf-terraforming import ...` into a shell script file.
+
+> **NOTE**
+> You can also just let cf-terraforming output directly to stdout (standard output, your terminal) and copy and run each of the command lines that `cf-terraforming import ...` generates for you
+>
+> These command lines will look something like:
+>
+> ```sh
+> terraform import cloudflare_record.terraform_managed_resource_3c0b456bc2aa443089c5f40f45f51b31 1109d899a5ff5fd74bc01e581693685b/3c0b456bc2aa443089c5f40f45f51b31
+> ```
+
 1. Run the following command:
 
    ```sh
-   cf-terraforming import --resource-type "cloudflare_record" --email $CLOUDFLARE_EMAIL --key $CLOUDFLARE_API_KEY --zone $CLOUDFLARE_ZONE_ID
+   cf-terraforming import --resource-type "cloudflare_record" --email $CLOUDFLARE_EMAIL --key $CLOUDFLARE_API_KEY --zone $CLOUDFLARE_ZONE_ID > to-import.sh
    ```
 
-2. Copy each `terraform import ...` command included in the output and run it. Terraform will import each resource individually into Terraform state.
+   This new file - `to-import.sh` - will have valid the necessary commands `terraform import ...` to import into the TF. state all the previously generated resources. 
 
-For example, if the output of the first command (`cf-terraforming import ...`) contained the following `terraform` commands:
+2. Now, you can just run the new `to-import.sh` file, like below. Terraform will import each resource individually into Terraform state.
 
-```txt
-terraform import cloudflare_record.terraform_managed_resource_3c0b456bc2aa443089c5f40f45f51b31 1109d899a5ff5fd74bc01e581693685b/3c0b456bc2aa443089c5f40f45f51b31
-terraform import cloudflare_record.terraform_managed_resource_5e10399a590a45279f09aa8fb1163354 1109d899a5ff5fd74bc01e581693685b/d09d916d059aa9fc8cb54bdd49deea5f
-terraform import cloudflare_record.terraform_managed_resource_de1cb74bae184b569bb7f83fefe72248 1109d899a5ff5fd74bc01e581693685b/8d6ec0d02c5b22212ff673782c816ef8
-terraform import cloudflare_record.terraform_managed_resource_5799bb01054843eea726758f935d2aa2 1109d899a5ff5fd74bc01e581693685b/3766b952a2dda4c47e71952aeef33c77
-```
-
-You would run each command individually in the terminal:
-
-```sh
-terraform import cloudflare_record.terraform_managed_resource_3c0b456bc2aa443089c5f40f45f51b31 1109d899a5ff5fd74bc01e581693685b/3c0b456bc2aa443089c5f40f45f51b31
+```bash
+bash to-import.sh
 ```
 
 ```txt output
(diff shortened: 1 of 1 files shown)


src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
```diff
@@ -212,27 +212,35 @@ To fix this, you must import the real state of those resources from Cloudflare i
 
 When you run `cf-terraforming import ...`, you will obtain a list of `terraform import ...` commands that you must run manually afterward to import those resources into Terraform state. This is currently a manual process, but it may be automated in the future.
 
+One way to ease the process of executing all the `terraform import`, is saving the result from `cf-terraforming import ...` into a shell script file.
+
+> **NOTE**
+> You can also just let cf-terraforming output directly to stdout (standard output, your terminal) and copy and run each of the command lines that `cf-terraforming import ...` generates for you
```
jacobbednarz, line 218: there is an alternative here with `--modern-import-block` with will output the idempotent `import` blocks instead which can then live alongside your side. they look like: 

```hcl
import {
  to = "cloudflare_record.terraform_managed_resource_3c0b456bc2aa443089c5f40f45f51b31"
  id = "1109d899a5ff5fd74bc01e581693685b/3c0b456bc2aa443089c5f40f45f51b31"
}
```

full docs: https://developer.hashicorp.com/terraform/language/import

jacobbednarz, line 218: we should avoid words like "just" and shortened words where we can be more explicit.

```suggestion
> You can also let cf-terraforming output directly to standard output in your terminal and copy and run each of the command lines that `cf-terraforming import ...` generates for you
```

src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
```diff
@@ -212,27 +212,35 @@ To fix this, you must import the real state of those resources from Cloudflare i
...
+
 1. Run the following command:
 
    ```sh
-   cf-terraforming import --resource-type "cloudflare_record" --email $CLOUDFLARE_EMAIL --key $CLOUDFLARE_API_KEY --zone $CLOUDFLARE_ZONE_ID
+   cf-terraforming import --resource-type "cloudflare_record" --email $CLOUDFLARE_EMAIL --key $CLOUDFLARE_API_KEY --zone $CLOUDFLARE_ZONE_ID > to-import.sh
    ```
 
-2. Copy each `terraform import ...` command included in the output and run it. Terraform will import each resource individually into Terraform state.
+   This new file - `to-import.sh` - will have valid the necessary commands `terraform import ...` to import into the TF. state all the previously generated resources.
```
jacobbednarz, line 232: this sentence doesn't really read well for me. couple of stand outs:

- don't shorten to "TF", let's use the correct names
- we should clean up the usages of `-` and the period in the middle of the sentence.

src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
```diff
@@ -212,27 +212,35 @@ To fix this, you must import the real state of those resources from Cloudflare i
 
 When you run `cf-terraforming import ...`, you will obtain a list of `terraform import ...` commands that you must run manually afterward to import those resources into Terraform state. This is currently a manual process, but it may be automated in the future.
 
+One way to ease the process of executing all the `terraform import`, is saving the result from `cf-terraforming import ...` into a shell script file.
```
jacobbednarz, line 215: we're getting ready to revamp this entire tutorial with the goal of less bash. i don't think we _really_ want to be adding more right now. cc @jhutchings1

src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
gui-wf: @jacobbednarz yeah, it is an alternative. But depending on what people want to do, it might be less clean.
For the purpose of having a backup of the cloudflare settings for example. 

Can I add as an alternative? And not replace entirely?

gui-wf: @jacobbednarz 
My intention with `stdout` is because it's also a word that is widely used in tools, compilers, code and Linux.
What if I flip that into "standard output (stdout)"?

src/content/docs/terraform/advanced-topics/import-cloudflare-resources.mdx
gui-wf: @jacobbednarz From my recent use of cf-terraforming, I haven't found a way to not script the use of it. There is currently no way to generate all the resources with the tool in just one command. There was before an `all` argument that you could give to it but there is not now.
So I believe that scripting with shell script is a well known and widely accepted way of scripting simple small things vs using something like Python or other scripting languages. 

Also, I think that we should have a reference to the [Terraform tool's `-generate-config-out`](https://developer.hashicorp.com/terraform/language/import/generating-configuration) argument since that also has a somewhat similar result to cf-terraforming. What do you think?

Related to this: Ruleset can't be fully generated as of yet as per [this issue](https://github.com/cloudflare/cf-terraforming/issues/801). The fix a person found was to import the ruleset before generating the rest of it (with the rules). So both ways, with the modern import block or the command way can be acceptable for different use cases.

jhutchings1: @gui-baeta Thanks for all your work here! We've been working on a bunch of changes around cf-terraforming recently, so this is timely. My broader question on this is whether we think there's a good reason not to default our guidance to focus on the `--modern-import-block` options? Asking folks to run so many bash commands seems like it adds a bit of complexity. I know I struggled with this step in my own testing.

gui-wf: @jhutchings1 Yeah, it can get really convoluted and confusing really quickly. 
I agree.

I have been testing more throughout the day and I guess that the problem I and others detected on that cf-terraforming issue I reference, should be fixed in cf-terraforming and/or the terraform provider. 
The use of shell commands, even if they are only to interact with terraform with its basic commands (in this case the import via commandline), depending on the context it can be a dirty fix to the end goal which is to be able to use cf-terraforming directly and arrive at a *just works* tool solution.

So, for this PR, documenting the use of `--modern-import-block` would be the best approach? I can do that, if that sounds good to you guys.

jhutchings1: I think documenting --modern-import-block is the most sensible approach, yes. Thank you for leaning in like you have. Really appreciate the effort you've put in here! cc: @dcpena

dcpena:
Will be closing this due to inactivity. If this work still needs to be done, I'd recommend opening a new PR to be better in sync with production.

PR #20994, opened 2025-03-20
Rules: Add URL Rewrite Template for Normalizing Encoded Forward Slashes (%2F)

nikitacano:
### Summary

This PR introduces a new Transform Rules template that allows customers to normalize encoded forward slashes (%2F) in the request path to standard slashes (/). This helps align Cloudflare’s request handling with origin server behavior, as many web servers automatically decode %2F into /.

The template uses the url_decode(http.request.uri.path) function to ensure consistency in URL-based rule evaluation, logging, and request processing. This change does not alter Cloudflare’s default behavior, which follows [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) by preserving encoded reserved characters. Instead, it provides an optional one-click solution for users who need it.

### DIFF
--- a/src/content/docs/rules/transform/examples/normalize-encoded-slash.mdx
+++ b/src/content/docs/rules/transform/examples/normalize-encoded-slash.mdx
@@ -1,0 +1,40 @@
+---
+pcx_content_type: example  
+summary: Create a rewrite URL rule (part of Transform Rules) to normalize encoded forward slashes (`%2F`) in the request path to standard slashes (`/`).  
+products:  
+  - Transform Rules  
+operation:  
+  - Rewrite URL  
+title: Normalize encoded slashes in URL path  
+description: Create a rewrite URL rule (part of Transform Rules) to normalize encoded forward slashes (`%2F`) in the request path to standard slashes (`/`).  
+---
+
+import { Example } from "~/components";
+
+## Why normalize `%2F`?
+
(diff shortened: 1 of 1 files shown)


src/content/docs/rules/transform/examples/normalize-encoded-slash.mdx
```diff
@@ -0,0 +1,40 @@
...
+  - Rewrite URL  
+title: Normalize encoded slashes in URL path  
+description: Create a rewrite URL rule (part of Transform Rules) to normalize encoded forward slashes (`%2F`) in the request path to standard slashes (`/`).  
+---
+
+import { Example } from "~/components";
+
+## Why normalize `%2F`?
+
+Different web servers and applications handle encoded forward slashes (`%2F`) in URLs differently. Cloudflare follows [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), which specifies that `%2F` should **not** be automatically normalized to `/` because `/` is a reserved character in URLs, and decoding it might change the intended meaning of the path.
```
pedrosousa:
```suggestion
Different web servers and applications handle encoded forward slashes (`%2F`) in URLs differently. Cloudflare follows [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), which specifies that `%2F` **should not** be automatically normalized to `/` because `/` is a reserved character in URLs, and decoding it might change the intended meaning of the path.
```

pedrosousa (approved):

PR #20998, opened 2025-03-20
[Docs Site] Only show external link arrow if element has no image children

KianNH:
### Summary

Only show external link arrow if element has no image children

### DIFF
--- a/src/plugins/rehype/external-links.ts
+++ b/src/plugins/rehype/external-links.ts
@@ -1,9 +1,22 @@
 import rehypeExternalLinks, { type Options } from "rehype-external-links";
+import type { Element } from "hast";
+
+function hasImgChild(node: Element): boolean {
+	return node.children.some(
+		(child) => child.type === "element" && child.tagName === "img",
+	);
+}
+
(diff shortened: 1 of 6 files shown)


src/plugins/rehype/external-links.ts
```diff
@@ -1,9 +1,22 @@
 import rehypeExternalLinks, { type Options } from "rehype-external-links";
+import type { Element } from "hast";
+
+function hasImgChild(node: Element): boolean {
```
GregBrimble, line 4: Hard to imagine ever intentionally linking over more than just a single image child, but I guess in those cases you also probably don't want `↗`. Fair enough!

GregBrimble (approved):

kodster28 (approved):

src/plugins/rehype/external-links.ts
KianNH: I admittedly stole it from `rehype-external-links` themselves - they have it in their tests: https://github.com/rehypejs/rehype-external-links/blob/b0ada86b178e135d55368ffa47b915db17219716/test.js#L388-L398

PR #20999, opened 2025-03-20
Don't add arrow to external image links

GregBrimble:
### Summary

`rel` attribute is preserved. `↗` text is dropped.

### Screenshots (optional)

image


GregBrimble:
Better version here: #20998

PR #21001, opened 2025-03-20
[Docs Site] Add AvailableChangelogFeeds component

KianNH:
### Summary

Add AvailableChangelogFeeds component


src/components/AvailableChangelogFeeds.astro
```diff
@@ -0,0 +1,76 @@
...
+	This feed contains entries for all Cloudflare products: <RSSButton />
+</p>
+
+<AnchorHeading depth={3} title="Area-specific feeds" />
+{
+	productsByGroup.map(([group, products]) => (
+		<>
+			<AnchorHeading depth={4} title={group} />
+			{group === "Application security" && (
+				<Aside type="note" title="DDoS and WAF ruleset feeds">
```
kodster28, line 67: I'd move this to below the list of products for the area.

src/util/products.ts
```diff
@@ -0,0 +1,10 @@
+import { getCollection } from "astro:content";
+
+export const products = await getCollection("products");
```
kodster28: If we started from the `changelogs` collection, would that help us at all here?

To make sure we aren't surfacing products that don't have any entries on the changelog itself.

kodster28 (approved):

PR #21034, opened 2025-03-21
[Docs Site] Adopt code block design

KianNH:
### Summary

Adopt code block design

### Screenshots (optional)

Before:
image

After:
image


kodster28 (approved):
Like, I'd maybe say flip the button text in dark mode to be black (it doesn't match the top nav bar one).

PR #21035, opened 2025-03-21
make sure that the wrangler team owns the framework guides (both for workers and pages)

dario-piotrowicz:
The frameworks team has been merged with the wrangler team, including me and James Culveyhouse, the wrangler team as a whole is working on the framework integrations so me and James should not be specifically set as the owners of these files but the wrangler team in general should


kodster28 (approved):

dario-piotrowicz:
Thanks for the approval @kodster28 😄

PR #21046, opened 2025-03-21
[Release notes] Stop pulling Wrangler release notes from GitHub API

kodster28:
### Summary

With the Wrangler team more consistently writing up [summarized changelog entries](https://developers.cloudflare.com/changelog/2025-03-13-wrangler-v4/), think it's time to re-think having the [Wrangler release notes](https://developers.cloudflare.com/workers/platform/changelog/wrangler/) ported into docs.

The reason why is b/c the GitHub API usually ends up failing about `8 builds` a day as a result of [rate limiting](https://github.com/cloudflare/cloudflare-docs/actions/runs/13993938229/job/39184210591#step:6:1007). Given that the information is also already clearly available in GitHub (with an RSS feed there too), there doesn't appear to be much added value in also surfacing them in docs.

Theoretically, I suppose we could rewrite the component.... but there's a lot of other investment that needs to go into other things changelog so I'd rather spend the time there.

cc: @GregBrimble / @mikenomitch, fyi that this is an option we're considering. Open to having a convo if I'm missing pieces related to ☝️ .

### Additional context on timing

This has been more painful recently and -- with Dev Week coming up -- we don't want to be blocking any docs previews or publishes from going live.

### DIFF
--- a/public/__redirects
+++ b/public/__redirects
@@ -1461,6 +1461,7 @@
 /workers/learning/fetch-event-lifecycle/ /workers/runtime-apis/fetch-event/ 301
 /workers/learning/getting-started/ /workers/get-started/quickstarts/ 301
 /workers/learning/profiling-workers/ /workers/reference/how-workers-works/ 301
+/workers/platform/changelog/wrangler/ https://github.com/cloudflare/workers-sdk/releases 301
 /workers/platform/scripts/ /api/resources/workers/subresources/scripts/methods/list/ 301
 /workers/platform/services/ /workers/runtime-apis/bindings/service-bindings/ 301
 /workers/platform/web-assembly/ /workers/platform/webassembly/ 301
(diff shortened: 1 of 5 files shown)


public/__redirects
```diff
@@ -1461,6 +1461,7 @@
 /workers/learning/fetch-event-lifecycle/ /workers/runtime-apis/fetch-event/ 301
 /workers/learning/getting-started/ /workers/get-started/quickstarts/ 301
 /workers/learning/profiling-workers/ /workers/reference/how-workers-works/ 301
+/workers/platform/changelog/wrangler/ https://github.com/cloudflare/workers-sdk/releases 301
```
kodster28, line 1464: Note -> Add github rss redirect too

KianNH (approved):

PR #21063, opened 2025-03-22
Don't limit input image pixels

GregBrimble:
### Summary

Large gifs are failing the build: https://github.com/cloudflare/cloudflare-docs/actions/runs/14011261801/job/39231343261?pr=21062

https://docs.astro.build/en/reference/configuration-reference/#imageservice

Maybe we could look into adding an adhoc optimization pass like we do for SVG minification, but this'll do for now.


GregBrimble:
Added in #21064

PR #21081, opened 2025-03-24
Test PR - Minor Tweak

mikenomitch:
Testing build automation

### DIFF
--- a/src/content/docs/cloudflare-one/policies/gateway/http-policies/tls-decryption.mdx
+++ b/src/content/docs/cloudflare-one/policies/gateway/http-policies/tls-decryption.mdx
@@ -115,6 +115,35 @@ When FIPS compliance is enabled, Gateway will only choose [FIPS-compliant cipher
 
 FIPS-compliant traffic defaults to [HTTP/3](/cloudflare-one/policies/gateway/http-policies/http3/). To enforce HTTP policies for UDP traffic, you must turn on the [Gateway proxy for UDP](/cloudflare-one/policies/gateway/http-policies/http3/#enable-http3-inspection).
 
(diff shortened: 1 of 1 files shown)


src/pages/index.astro
```diff
@@ -61,6 +61,7 @@ const topCards = [
 			{ text: "Workers", href: "/workers/" },
 			{ text: "Pages", href: "/pages/" },
 			{ text: "R2", href: "/r2/" },
+			{ text: "Containers", href: "/containers/" },
```
kodster28, line 64: This is a space constrained area, so you'd want to remove an entry here as well

src/content/docs/containers/get-started.mdx
```diff
@@ -0,0 +1,290 @@
...
+Now, deploy your Worker:
+
+```sh
+wrangler deploy
+```
+
+When you run this command, several things happen:
+
+- Wrangler builds your image using Docker.
+- Wrangler pushes you image to the Cloudchamber Image Registry, which is automatically
```
johnspurlock, line 91: `you image` -> `your image`

src/content/docs/containers/get-started.mdx
```diff
@@ -0,0 +1,290 @@
...
+```
+
+When the Durable Object instance is created, it starts a new container
+using the `startAndWaitForPort` method, which calls `this.ctx.container.start`.
+
+The `fetch` method calls the `proxyFetch` helper, which passes a request
+to the server running on the open port.
+
+In this case, the Workers request will be proxied to the Golang server
+listeining on port 8080.
```
johnspurlock, line 210: `listeining` -> `listening`

PR #21088, opened 2025-03-24
Bump redirects-in-workers@0.0.7 and raise limits

GregBrimble:
Increased limits to save @kodster28's sanity. This does come at a cost, especially for the dynamic rules.

Expect cold start requests to take longer, the larger the `__redirects` file you have. Expect **every request** to take longer, the more dynamic rules you have.


kodster28 (approved):

kodster28:
🙏 .

Think we'll still keep pruning them / potentially moving some of those dynamic, historical ones over into bulk redirects.

PR #21100, opened 2025-03-24
[Hyperdrive] explain d1 & hyperdrive

thomasgauvin:
required depth for review: low

### DIFF
--- a/src/content/docs/hyperdrive/reference/faq.mdx
+++ b/src/content/docs/hyperdrive/reference/faq.mdx
@@ -15,6 +15,18 @@ Hyperdrive connects to your database using [Cloudflare's IP address ranges](http
 
 You can use this to configure restrictions in your database firewall to restrict the IP addresses that can access your database.
 
+### Does Hyperdrive support connecting to D1 databases?
+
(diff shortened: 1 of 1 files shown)


src/content/docs/hyperdrive/reference/faq.mdx
```diff
@@ -15,6 +15,18 @@ Hyperdrive connects to your database using [Cloudflare's IP address ranges](http
 
 You can use this to configure restrictions in your database firewall to restrict the IP addresses that can access your database.
 
+### Does Hyperdrive support connecting to D1 databases?
+
+Hyperdrive does not support [D1](/d1) because D1 provides fast connectivity from Workers by design.
+
+Hyperdrive is designed to speed up connectivity to traditional, regional SQL databases.
```
sejoker: `regional SQL databases` I suggest we explain in more details what we mean here.

thomasgauvin:
```suggestion
Hyperdrive is designed to speed up connectivity to traditional, regional SQL databases such as PostgreSQL. These databases are typically accessed using database drivers that communicate over TCP/IP.
```

sejoker (approved):

PR #21116, opened 2025-03-25
Update index.mdx

mattfysh:
### Documentation checklist

### DIFF
--- a/src/content/docs/cloudflare-one/connections/connect-networks/do-more-with-tunnels/local-management/index.mdx
+++ b/src/content/docs/cloudflare-one/connections/connect-networks/do-more-with-tunnels/local-management/index.mdx
@@ -9,7 +9,7 @@ sidebar:
 import { DirectoryListing } from "~/components"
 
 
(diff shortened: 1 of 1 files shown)


mattfysh:
Remotely-managed tunnels should NOT be recommended. The cloudflare API does not allow you to manage ingress rules individually, so it is completely useless for deployment automation where stacks define their own ingress

It might even be an idea to recommend the kubernetes cloudflare operator instead, which does a much better job at integrating with mature deployment pipelines via its `TunnelBinding` config, perhaps once https://github.com/adyanth/cloudflare-operator/issues/127 is resolved

jcsf:
Hi @mattfysh,

Thank you for your contribution! However, we're shifting away from locally managed tunnels due to their various challenges, especially in high availability setups. Ensuring that all cloudflared instances share the same configuration can be problematic with local management. We believe that using remotely managed tunnels is a more effective solution for achieving consistency.

Could you please open an issue on the [cloudflared](https://github.com/cloudflare/cloudflared) project, and tag me there? It would be helpful to detail your use case and specify any features you feel are lacking in the Cloudflare Tunnel API that would help you better utilize remotely managed tunnels.

Best regards

PR #21120, opened 2025-03-25
fix non-existing pages commands in the generated wrangler configuration section

dario-piotrowicz:
the `wrangler pages build` and `wrangler pages build-env` as missing the `functions` keyword

As you can see the `wrangler pages build-env` command is not recognized:
Screenshot 2025-03-25 at 11 31 38

While `wrangler pages functions build-env` is:
Screenshot 2025-03-25 at 11 31 52


dario-piotrowicz:
Note that the `wrangler pages functions` commands are barely documented as far as I can tell, so maybe we should remove these 2 entries from the list altogether?

kodster28 (approved):

kodster28:
Appears to have been addressed in another PR.

PR #21153, opened 2025-03-25
[ZT] Multi-user updates

ranbel:
PCX-16296 Fast user switching
PCX-15855 DNS logging


src/content/docs/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-multiuser.mdx
```diff
@@ -114,3 +115,13 @@ flowchart TB
 		regexists-. "No" .->reg
 ```
 
+### Fast user switching
+
+:::note
+Requires [multi-user mode](#enable-multi-user-mode).
+:::
+
+[Fast user switching](https://learn.microsoft.com/windows/win32/shell/fast-user-switching) is a Windows feature that allows users to switch accounts without logging out. With fast user switching, multiple users may be logged in to the device and generating network traffic. The WARP client will attribute all traffic to the user who has the interactive desktop. For example, if user A is logged in and fast user switches to user B, traffic from both accounts will appear to come from user B. Now assume that user B logs out and there is no [pre-login registration](/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-prelogin/); WARP will continue to attribute traffic to user B since they still have the active registration.
```
cf-rhett: @kokolocomotion1 does this make sense to you? It does to me since I've been in the code but just wanted to make sure it makes sense generally

kokolocomotion1: Would add in a couple things like this:

[Fast user switching](https://learn.microsoft.com/windows/win32/shell/fast-user-switching) is a Windows feature that allows users to switch between Windows accounts without logging out. With fast user switching, multiple users may be logged in to the device and generating network traffic. The WARP client will attribute all traffic to the user who has the active WARP registration. For example, if user A is logged in and fast user switches to user B who has an active WARP registration, traffic from both accounts will appear to come from user B. Now assume that user B logs out and there is no [Windows pre-login registration](/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-prelogin/) for user A; WARP will continue to attribute traffic to user B since they still have the active registration.

patriciasantaana (approved):

PR #21198, opened 2025-03-27
Explain how to do local dev when you have DOs with a framework

emily-shen:
### Documentation checklist

### DIFF
--- a/src/content/docs/durable-objects/platform/known-issues.mdx
+++ b/src/content/docs/durable-objects/platform/known-issues.mdx
@@ -3,12 +3,7 @@ pcx_content_type: concept
 title: Known issues
 ---
 
-import {
-	GlossaryTooltip,
-	Steps,
-	WranglerConfig,
-	PackageManagers,
-} from "~/components";
+import { GlossaryTooltip, Steps } from "~/components";
 
 Durable Objects is generally available. However, there are some known issues.
 
@@ -41,74 +36,3 @@ The Workers editor in the [Cloudflare dashboard](https://dash.cloudflare.com/) a
 Currently, when developing locally (using `npx wrangler dev`), Durable Object [alarm methods](/durable-objects/api/alarms) may fail after a hot reload (if you edit the code while the code is running locally).
 
 To avoid this issue, when using Durable Object alarms, close and restart your `wrangler dev` command after editing your code.
-
-## Local development with web frameworks
-
-When developing a Pages or Workers Assets application, you may have to define your Durable Object in a separate Worker in order to develop locally.
-This only applies to frameworks that use a development server that runs in Node.js. We are working to improve this experience - for example, frameworks that use the Cloudflare Vite plugin do not have this issue.
-
-For now, you can set up your project like this to enable local development. In this example, we assume you have a pre-existing Worker that you want to add a DO binding to.
-
-<Steps>
-1. Create a second Worker with your Durable Object code.
-
-    ```ts title="src/index.ts"
-    export class MyDurableObject extends DurableObject {
-    	// Your DO code goes here
-    }
-
-    export default {
-    	fetch() {
-    		// Doesn't have to do anything, but a DO cannot be the default export
-    			return new Response("Hello, world!");
-    	},
-    };
-    ```
-
-2.  Create a **new** Wrangler config file for this Worker.
-
-        <WranglerConfig>
-        ```json
-        {
-        	"name": "external-do-worker",
-        	"main": "src/index.ts",
-        	"compatibility_date": "XXXX-XX-XX"
-        }
-        ```
-        </WranglerConfig>
-
-3.  Update your **original** Durable Object bindings in your pre-existing Wrangler config file to include the `script_name` field. This should be the same as the `name` field in the new Worker's config.
-
-    <WranglerConfig>
-
-        ```json
-        {
-        	...
-        	"durable_objects": {
-        		"bindings": [
-        			{
-        				"name": "BINDING",
-        				"class_name": "MyDurableObject",
-        				"script_name": "external-do-worker",
-        			},
-        		],
-        	},
-        	...
-        },
-        ```
-
-        </WranglerConfig>
-
-4.  Develop locally
-
-        If you are not using RPC with your Durable Object, you can run a separate Wrangler dev session alongside your framework development server.
-
-        		Otherwise, you can build your application and run both Workers in the same Wrangler dev session.
-
-        	  If you are using Pages:
-        <PackageManagers type="exec" pkg="wrangler" args="pages dev -c path/to/pages/wrangler.jsonc -c path/to/external-do-worker/wrangler.jsonc" />
-
-        If you are using Workers with Assets:
-        <PackageManagers type="exec" pkg="wrangler" args="dev -c path/to/workers-assets/wrangler.jsonc -c path/to/external-do-worker/wrangler.jsonc" />
-
-</Steps>
--- a/src/content/docs/durable-objects/platform/known-issues.mdx
+++ b/src/content/docs/durable-objects/platform/known-issues.mdx
@@ -1,10 +1,14 @@
 ---
 pcx_content_type: concept
 title: Known issues
-
 ---
 
-import { GlossaryTooltip } from "~/components";
+import {
+	GlossaryTooltip,
+	Steps,
+	WranglerConfig,
+	PackageManagers,
+} from "~/components";
 
 Durable Objects is generally available. However, there are some known issues.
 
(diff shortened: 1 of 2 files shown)


src/content/docs/durable-objects/platform/known-issues.mdx
```diff
@@ -36,4 +40,75 @@ The Workers editor in the [Cloudflare dashboard](https://dash.cloudflare.com/) a
...
+    ```
+
+2.  Create a new Wrangler config file for this Worker.
+
+        <WranglerConfig>
+        ```json
+        {
+        	"name": "external-do-worker",
+        	"main": "src/index.ts",
+        	"compatibility_date": "XXXX-XX-XX"
```
vicb: Do we need to declare the DO bindings here? (cf https://github.com/cloudflare/workers-sdk/pull/8697/files#diff-3329f9859e87a732ec3417dca84fbef318e91a9f4d51599598ea9196df1d5bf1R9)

src/content/docs/durable-objects/platform/known-issues.mdx
```diff
@@ -36,4 +40,75 @@ The Workers editor in the [Cloudflare dashboard](https://dash.cloudflare.com/) a
...
+        </WranglerConfig>
+
+4.  Develop locally
+
+        If you are not using RPC with your Durable Object, you can run a separate Wrangler dev session alongside your framework development server.
+
+        		Otherwise, you can build your application and run both Workers in the same Wrangler dev session.
+
+        	  If you are using Pages:
+        <PackageManagers type="exec" pkg="wrangler" args="pages dev -c path/to/pages/wrangler.json -c path/to/external-do-worker/wrangler.json" />
```
vicb:
```suggestion
        <PackageManagers type="exec" pkg="wrangler" args="pages dev -c path/to/pages/wrangler.jsonc -c path/to/external-do-worker/wrangler.jsonc" />
```

src/content/docs/durable-objects/platform/known-issues.mdx
```diff
@@ -36,4 +40,75 @@ The Workers editor in the [Cloudflare dashboard](https://dash.cloudflare.com/) a
...
+
+        If you are using Workers with Assets:
+        <PackageManagers type="exec" pkg="wrangler" args="dev -c path/to/workers-assets/wrangler.json -c path/to/external-do-worker/wrangler.json" />
```
vicb:
```suggestion
        <PackageManagers type="exec" pkg="wrangler" args="dev -c path/to/workers-assets/wrangler.jsonc -c path/to/external-do-worker/wrangler.jsonc" />
```

src/content/docs/durable-objects/platform/known-issues.mdx
emily-shen: no, because the worker doesn't bind to the DO here (the fetch is just a no-op). it's the users original worker that binds to the DO.

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
...
+        }
+
+        export default {
+        	fetch() {
+        			// Doesn't have to do anything, but a DO cannot be the default export
+        			return new Response("Hello, world!");
+        	},
+        };
+        ```
+        That Worker also needs a Wrangler config that looks like this:
```
CarmenPopoviciu:
```suggestion
        That Worker also needs a Wrangler configuration file that looks like this:
```

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
 - [Workers AI bindings](/workers-ai/get-started/workers-wrangler/#2-connect-your-worker-to-workers-ai)
 
   <Render file="ai-local-usage-charges" product="workers" />
+
+- <a name="durable-objects"></a>[Durable Object bindings](/durable-objects/api/)
```
CarmenPopoviciu: are we adding this link here just so we can specifically link to this section? feels a bit odd that DOs get this and other bindings not. Is linking to `Supported bindings` not enough?

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
...
+
+  - To use a Durable Object binding with `getPlatformProxy`, always specify a [`script_name`](/workers/wrangler/configuration/#durable-objects).
+
+    For example, you might have the following binding in a file read by `getPlatformProxy`.
```
CarmenPopoviciu:
```suggestion
    For example, you might have the following binding in a Wrangler configuration file read by `getPlatformProxy`.
```

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
...
+        	"compatibility_date": "XXXX-XX-XX"
+        }
+        ```
+        </WranglerConfig>
+
+        If you are not using RPC with your Durable Object, you can run a separate Wrangler dev session alongside your framework development server.
+
+        Otherwise, you can build your application and run both Workers in the same Wrangler dev session.
+
+        If you are using Pages:
```
CarmenPopoviciu:
```suggestion
        If you are using Pages run:
```

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
...
+        <PackageManagers type="exec" pkg="wrangler" args="pages dev -c path/to/pages/wrangler.jsonc -c path/to/external-do-worker/wrangler.jsonc" />
+
+        If you are using Workers with Assets:
```
CarmenPopoviciu:
```suggestion
        If you are using Workers with Assets run:
```

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
...
+        ```json
+        {
+        	"name": "external-do-worker",
+        	"main": "src/index.ts",
+        	"compatibility_date": "XXXX-XX-XX"
+        }
+        ```
+        </WranglerConfig>
+
+        If you are not using RPC with your Durable Object, you can run a separate Wrangler dev session alongside your framework development server.
```
CarmenPopoviciu, line 433:
```suggestion
        If you are not using RPC with your Durable Object, you can run a separate `wrangler dev` session alongside your framework development server.
```
not 100% sure tho...

CarmenPopoviciu, line 433: I wonder if we're not confusing users with too many options here. Is there one option that works for both rpc/non-rpc?

src/content/docs/workers/wrangler/api.mdx
emily-shen: Yeah this is just so I can link to it from the warning log in wrangler - we don't need this for anything else and it's not visible to users so I think it's okay to have it for just the one.

src/content/docs/workers/wrangler/api.mdx
emily-shen: Might be a pages dev command so I'll leave it as is

emily-shen: There's a trade off with each unfortunately. The first option lets you have hot reloading etc. by using the frameworks dev server, but no RPC. The second requires you build each time but allows RPC.

src/content/docs/workers/wrangler/api.mdx
```diff
@@ -399,3 +386,56 @@ The bindings supported by `getPlatformProxy` are:
...
+
+        Otherwise, you can build your application and run both Workers in the same Wrangler dev session.
+
+        If you are using Pages:
```
CarmenPopoviciu: do we really need to spell out how to run the dev cmd for Pages vs Workers + Assets? Feels a bit outside the scope of this docs. Maybe link to where this specific local dev doc lives (tried to find it but couldn't, but I know we have a section abt multi-worker local dev)?

emily-shen: Yeah the problem is we don't have multi dev docs (@penalosa had a draft up but I'm not sure if there's a blocker). And the order you specify the configs in matters, so I thought it would be easier to just put it in.

src/content/docs/workers/wrangler/api.mdx
CarmenPopoviciu: Right, but again, the only reason we're adding it is because another page in our docs wants to link to that specific bit. So we are making decisions about how this page is semantically structured, based on how other docs pages interact with it, and that doesn't feel right. Even if we were to be consistent and convert all items in the bindings list into URI fragments, that wouldn't make much sense in the context of this page either, since barely any of those items do anything else other than link to another page. DOs are the only exception, and that is because we are now making these changes. 

I won't die on this hill, and I understand that we are trying to link users to the relevant section with zero friction, but personally I don't think this is the way to go, for this particular case, for the reasons I've explained. I think linking to `Supported bindings` should be enough.

emily-shen: got rid of the anchor link + updated the wrangler PR as well

src/content/docs/workers/wrangler/api.mdx
CarmenPopoviciu: oh, I was so convinced we had the docs up somewhere. Sorry for the confusion.

CarmenPopoviciu (approved):

CarmenPopoviciu (approved):

PR #21227, opened 2025-03-28
[R2] Adding workflow for using Rclone to upload to R2

Oxyjun:
### Summary

1. Adding new workflow for using Rclone to upload multiple objects to R2.
2. Recommending using Rclone upload method over dashboard by making it appear first in the page.
3. Adding a note to explain dashboard only allows single object upload at a time, and that Rclone is better suited for multiple object upload.

### Documentation checklist


Oxyjun:
CC: @harshil1712 could you check if the steps for configuring / uploading via Rclone is accurate? I tried it myself, but want to double check.

marciocloudflare (approved):

src/content/docs/r2/objects/upload-objects.mdx
```diff
@@ -5,16 +5,79 @@ sidebar:
...
+```
+</TabItem><TabItem label="Windows">
+To install on a Windows device, download it from the official [Rclone website](https://rclone.org/downloads/).
+</TabItem>
+</Tabs>
+
+:::note
+Use Rclone if you wish to upload multiple objects to R2.
+
+Cloudflare dashboard only allows you to upload one object at a time.
```
harshil1712: This is not correct. I was able to upload multiple objects with drag & drop and even going through the File select option.

src/content/docs/r2/objects/upload-objects.mdx
```diff
@@ -5,16 +5,79 @@ sidebar:
   order: 1
 ---
 
+import { Steps, Tabs, TabItem } from "~/components"
+
 You can upload objects to your bucket from the Cloudflare dashboard or using the Wrangler.
 
+## Upload objects via Rclone
+
+Rclone is a command-line tool which manages files on cloud storage. You can use Rclone to upload objects to R2. To use Rclone, install it onto your machine:
```
harshil1712: We should link to the docs and let the user decide the method based on the OS. Since there are multiple methods for a single OS: https://rclone.org/install/

src/content/docs/r2/objects/upload-objects.mdx
```diff
@@ -5,16 +5,79 @@ sidebar:
   order: 1
 ---
 
+import { Steps, Tabs, TabItem } from "~/components"
+
 You can upload objects to your bucket from the Cloudflare dashboard or using the Wrangler.
 
+## Upload objects via Rclone
```
harshil1712, line 12: We already have example for `rclone`. Should we reference it instead?
https://developers.cloudflare.com/r2/examples/rclone/

src/content/docs/r2/objects/upload-objects.mdx
Oxyjun: Yup, I was getting confused with `wrangler object put`. Will fix!

src/content/docs/r2/objects/upload-objects.mdx
Oxyjun: I think the example could do with clearer instructions. I'll link the example, but might not remove what I've written for now (the example doesn't capture info on how to configure Rclone). Does that sound sensible?

harshil1712: Should we then add these steps to that example and refer it here? The example already contains relevant links and calls out things that are important. We can also separate the `Upload and retrieve objects` into separate sections.
If we still want to show the code, we can have the code block and for other instructions send the user to the examples page.

harshil1712 (approved):

aninibread (approved):

PR #21258, opened 2025-03-31
Added a hono version of workers examples

coffee-mug:
### Summary

Updated all examples in documentation for Cloudflare Workers to include code for the example, written with hono. 

### Documentation checklist


src/content/docs/workers/examples/bulk-origin-proxy.mdx
```diff
@@ -74,6 +74,40 @@ export default {
 } satisfies ExportedHandler;
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from 'hono';
+
+type Bindings = {};
```
yusukebe: Is this `Bindings` necessary?

src/content/docs/workers/examples/bulk-origin-proxy.mdx
```diff
@@ -74,6 +74,40 @@ export default {
...
+app.all('*', async (c) => {
+  const url = new URL(c.req.url);
+  
+  // Check if incoming hostname is a key in the ORIGINS object
+  if (url.hostname in ORIGINS) {
+    const target = ORIGINS[url.hostname];
+    url.hostname = target;
+    
+    // If it is, proxy request to that third party origin
+    return fetch(url.toString(), c.req.raw);
```
yusukebe: This seems to be for advanced users, but you can use the Proxy Helper to remove unnecessary headers. For example, by removing `Accept-Encoding`, you can prevent the response from the origin from being incorrectly encoded. The implementation is as follows.

https://github.com/honojs/hono/blob/main/src/helper/proxy/index.ts

Example:

```ts
import { Hono } from 'hono'
import { proxy } from 'hono/proxy'

//... 

const app = new Hono()

app.all('*', async (c) => {
  const url = new URL(c.req.url)

  // Check if incoming hostname is a key in the ORIGINS object
  if (url.hostname in ORIGINS) {
    const target = ORIGINS[url.hostname]
    url.hostname = target

    // If it is, proxy request to that third party origin
    return proxy(url, c.req.raw)
  }

  // Otherwise, process request as normal
  return proxy(c.req.raw)
})
```

src/content/docs/workers/examples/bulk-redirects.mdx
```diff
@@ -99,4 +99,44 @@ async def on_fetch(request):
...
+const redirectMap = new Map([
+  ["/bulk1", `https://${externalHostname}/redirect2`],
+  ["/bulk2", `https://${externalHostname}/redirect3`],
+  ["/bulk3", `https://${externalHostname}/redirect4`],
+  ["/bulk4", "https://google.com"],
+]);
+
+// Middleware to handle redirects
+app.use('*', async (c, next) => {
+  const path = new URL(c.req.url).pathname;
```
yusukebe: You can get the path by `c.req.path`.

```ts
const path = c.req.path
```

src/content/docs/workers/examples/cache-api.mdx
yusukebe, line 1: It's not bad to introduce primitive implementations, but since Hono has Cache middleware, I think it's better to use that.

https://hono.dev/docs/middleware/builtin/cache

src/content/docs/workers/examples/cache-post-request.mdx
```diff
@@ -147,4 +147,71 @@ async def on_fetch(request, _, ctx):
...
+
+```ts
+import { Hono } from 'hono';
+
+interface Env {}
+
+const app = new Hono<{ Bindings: Env }>();
+
+// Helper function to create SHA-256 hash
+async function sha256(message: string): Promise<string> {
```
yusukebe: You can use `sha256` function from `hono/utils/crypto`:

```ts
import { sha256 } from 'hono/utils/crypto'
```

src/content/docs/workers/examples/cache-tags.mdx
```diff
@@ -104,6 +102,50 @@ export default {
...
+
+```ts
+import { Hono } from "hono";
+
+type Bindings = {};
+
+const app = new Hono<{ Bindings: Bindings }>();
+
+app.all("*", async (c) => {
+	try {
```
yusukebe: It's better not to use `try`, and you can handle errors with [`app.onError()`](https://hono.dev/docs/api/hono#error-handling):

```ts
app.onError((err, c) => {
  return c.json({ error: err.message }, 500)
})
```

src/content/docs/workers/examples/cache-tags.mdx
```diff
@@ -104,6 +102,50 @@ export default {
...
+		const url = new URL(c.req.url);
+		const params = url.searchParams;
+		const tags = params.has("tags") ? params.get("tags").split(",") : [];
```
yusukebe: You don't have to use `URL` and `SearchParams`. You can use `c.req.query()`:

```ts
const tags = c.req.query('tags')?.split(',') ?? []
const uri = c.req.query('uri') ?? ''
```

src/content/docs/workers/examples/conditional-response.mdx
```diff
@@ -163,4 +163,91 @@ async def on_fetch(request):
...
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from 'hono';
+import { HTTPException } from 'hono/http-exception';
+
+// Define the RequestWithCf interface to add Cloudflare-specific properties
+interface RequestWithCf extends Request {
```
yusukebe: If the types are set correctly, you don't have to define CF-specific `Request` object.

```json
{
  "compilerOptions": {
    "types": [
      "./worker-configuration"
    ]
  },
}
```

src/content/docs/workers/examples/cors-header-proxy.mdx
```diff
@@ -328,6 +328,149 @@ export default {
...
+// but does not implement CORS
+const API_URL = "https://examples.cloudflareworkers.com/demos/demoapi";
+
+// The endpoint you want the CORS reverse proxy to be on
+const PROXY_ENDPOINT = "/corsproxy/";
+
+const app = new Hono<{ Bindings: Bindings }>();
+
+// Demo page handler
+app.get('*', async (c) => {
```
yusukebe: You should use `next`:

```ts
app.get('*', async (c, next) => {
  const url = new URL(c.req.url)

  // Only handle non-proxy requests with this handler
  if (url.pathname.startsWith(PROXY_ENDPOINT)) {
    return next()
  }
```

src/content/docs/workers/examples/openai-sdk-streaming.mdx
yusukebe, line 1: Using [Streaming Helper](https://hono.dev/docs/helpers/streaming) is cool!

```ts
import { Hono } from 'hono'
import { streamText } from 'hono/streaming'
import OpenAI from 'openai'

interface Env {
  OPENAI_API_KEY: string
}

const app = new Hono<{ Bindings: Env }>()

app.get('*', async (c) => {
  const openai = new OpenAI({
    apiKey: c.env.OPENAI_API_KEY,
  })

  const chatStream = await openai.chat.completions.create({
    model: 'gpt-4o-mini',
    messages: [{ role: 'user', content: 'Tell me a story' }],
    stream: true,
  })

  return streamText(c, async (stream) => {
    for await (const message of chatStream) {
      await stream.write(message.choices[0].delta.content || '')
    }
    stream.close()
  })
})

export default app
```

src/content/docs/workers/examples/security-headers.mdx
yusukebe, line 1: We have [Secure Headers Middleware](https://hono.dev/docs/middleware/builtin/secure-headers) for this purpose. Using the middleware is Hono-way.

src/content/docs/workers/examples/signing-requests.mdx
```diff
@@ -244,6 +244,214 @@ export default {
...
+  const key = await crypto.subtle.importKey(
+    "raw",
+    secretKeyData,
+    { name: "HMAC", hash: "SHA-256" },
+    false,
+    ["sign", "verify"]
+  );
+  
+  // Make sure the request has the verification parameter
+  if (!url.searchParams.has("verify")) {
```
yusukebe: You can use `c.req.query()`:

```ts
c.req.query('verify')
```

src/content/docs/workers/examples/signing-requests.mdx
```diff
@@ -244,6 +244,214 @@ export default {
...
+    return c.text("Missing query parameter", 403);
+  }
+  
+  // Extract timestamp and signature
+  const [timestamp, hmac] = url.searchParams.get("verify")!.split("-");
```
yusukebe: You can use `c.req.query()`.

src/content/docs/workers/examples/signing-requests.mdx
```diff
@@ -244,6 +244,214 @@ export default {
...
+  const assertedTimestamp = Number(timestamp);
+  
+  // Recreate the data that should have been signed
+  const dataToAuthenticate = `${url.pathname}${assertedTimestamp}`;
```
yusukebe: You can use `c.req.path` instead of `url.pathname`.

src/content/docs/workers/examples/signing-requests.mdx
```diff
@@ -244,6 +244,214 @@ export default {
...
+  // Check if the signature has expired
+  if (Date.now() / 1000 > assertedTimestamp + EXPIRY) {
+    return c.text(
+      `URL expired at ${new Date((assertedTimestamp + EXPIRY) * 1000)}`,
+      403
+    );
+  }
+  
+  // If verification passes, proxy the request to example.com
+  return fetch(new URL(url.pathname, "https://example.com"), c.req.raw);
```
yusukebe: Use `c.req.path`. If so, you don't have to create the `URL` object.

src/content/docs/workers/examples/websockets.mdx
yusukebe, line 1: Let's use [WebSocket Helper](https://hono.dev/docs/helpers/websocket)!

```ts
import { Hono } from 'hono'
import { upgradeWebSocket } from 'hono/cloudflare-workers'

const app = new Hono()

app.get(
  '*',
  upgradeWebSocket((c) => {
    return {
      onMessage(event, ws) {
        console.log('Received message from client:', event.data)
        ws.send(`Echo: ${event.data}`)
      },
      onClose: () => {
        console.log('WebSocket closed:', event)
      },
      onError: () => {
        console.error('WebSocket error:', event)
      },
    }
  })
)
```

src/content/docs/workers/examples/ab-testing.mdx
```diff
@@ -137,4 +137,71 @@ async def on_fetch(request):
...
+const app = new Hono();
+
+const NAME = "myExampleWorkersABTest";
+
+// Middleware to handle A/B testing logic
+app.use('*', async (c) => {
+  const url = new URL(c.req.url);
+  
+  // Enable Passthrough to allow direct access to control and test routes
+  if (url.pathname.startsWith("/control") || url.pathname.startsWith("/test")) {
```
yusukebe: You can use `c.req.path` instead of `url.pathname`.

src/content/docs/workers/examples/basic-auth.mdx
yusukebe, line 1: Use [Basic Auth Middleware](https://hono.dev/docs/middleware/builtin/basic-auth)!

src/content/docs/workers/examples/cors-header-proxy.mdx
yusukebe, line 1: We may use the CORS middleware, but I think the API `https://examples.cloudflareworkers.com/demos/demoapi` is not working correctly, so I can't try to confirm that it works correctly. So, it's better to leave it.

src/content/docs/workers/examples/bulk-origin-proxy.mdx
coffee-mug: Thanks a lot, I didn't know Hono had a proxy built-in

src/content/docs/workers/examples/cache-post-request.mdx
coffee-mug: Got it, thanks a lot

src/content/docs/workers/examples/signing-requests.mdx
coffee-mug: Cool, thank you for catching that, updated

src/content/docs/workers/examples/websockets.mdx
coffee-mug: Done!

src/content/docs/workers/examples/bulk-redirects.mdx
coffee-mug: Ok, changed, thanks

src/content/docs/workers/examples/cache-api.mdx
coffee-mug: Right, and that also showcases hono capabilities to abstract lot of that work away. Got it, thanks, updated

src/content/docs/workers/examples/cache-tags.mdx
coffee-mug: Did not know about app.onError, cool! Done

src/content/docs/workers/examples/cache-tags.mdx
coffee-mug: Indeed, sorry for the repetition

src/content/docs/workers/examples/conditional-response.mdx
coffee-mug: Changed

src/content/docs/workers/examples/cors-header-proxy.mdx
coffee-mug: Thanks for catching that!

src/content/docs/workers/examples/openai-sdk-streaming.mdx
coffee-mug: Woaw cool!

src/content/docs/workers/examples/basic-auth.mdx
coffee-mug: wow how did I miss this, thanks

src/content/docs/workers/examples/cors-header-proxy.mdx
coffee-mug: To leave it as is?

src/content/docs/workers/examples/basic-auth.mdx
```diff
@@ -328,6 +328,58 @@ async fn fetch(req: Request, env: Env, _ctx: Context) -> Result<Response> {
...
+
+```ts
+/**
+ * Shows how to restrict access using the HTTP Basic schema with Hono.
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication
+ * @see https://tools.ietf.org/html/rfc7617
+ */
+
+import { Hono } from 'hono';
+import { auth } from 'hono/basic-auth';
```
yusukebe: The usage of Basic Auth Middleware is like the following:

```ts
import { Hono } from 'hono'
import { basicAuth } from 'hono/basic-auth'

// Define environment interface
interface Env {
  Bindings: {
    USERNAME: string
    PASSWORD: string
  }
}

const app = new Hono<Env>()

// Public homepage - accessible to everyone
app.get('/', (c) => {
  return c.text('Anyone can access the homepage.')
})

// Admin route - protected with Basic Auth
app.get('/admin/*', async (c, next) => {
  return basicAuth({ username: c.env.USERNAME, password: c.env.PASSWORD })(c, next)
})

export default app
```

Please update the code on the docs referring to this!

src/content/docs/workers/examples/cors-header-proxy.mdx
yusukebe: Yes. Let's leave it as is in this PR.

src/content/docs/workers/examples/basic-auth.mdx
coffee-mug: I referred to the example in the doc to restrict to a specific route and method [Basic Auth](https://hono.dev/docs/middleware/builtin/basic-auth)

`const app = new Hono()

app.get('/auth/page', (c) => {
  return c.text('Viewing page')
})

app.delete(
  '/auth/page',
  basicAuth({ username: 'hono', password: 'acoolproject' }),
  (c) => {
    return c.text('Page deleted')
  }
)`

What's the difference with returning the basicAuth function from the handler?

yusukebe: The syntax in the following is okay:

```ts
app.delete('/auth/page', basicAuth({ username: 'hono', password: 'acoolproject' }), (c) => {
  return c.text('Page deleted')
})
```

But, if you want to use variables in `c.env.USERNAME` and `c.env.PASSWORD`, you can't access `c` with the syntax above. So you have to write it as "callback-style":

```ts
app.delete(
  '/auth/page',
  async (c, next) => {
    return basicAuth({ username: c.env.USERNAME, password: c.env.PASSWORD })(c, next)
  },
  (c) => {
    return c.text('Page deleted')
  }
)
```

coffee-mug: Ok makes sense indeed thank you so much @yusukebe, fixed it in the example.

src/content/docs/workers/examples/103-early-hints.mdx
```diff
@@ -136,4 +136,48 @@ def on_fetch(request):
...
+  return c.body(CSS, {
+    headers: {
+      "content-type": "text/css",
+    },
+  });
+});
+
+// Serve HTML with early hints
+app.get('*', (c) => {
+  return c.body(HTML, {
```
yusukebe: This is like a nitpick, but you can use `c.html()`:

```ts
return c.html(HTML, {
  headers: {
    link: '</test.css>; rel=preload; as=style',
  },
})
```

yusukebe:
Hi @coffee-mug 

I still want you to fix some things and will comment on them. Please wait a moment.

src/content/docs/workers/examples/ab-testing.mdx
```diff
@@ -137,4 +137,72 @@ async def on_fetch(request):
     return fetch(urlunparse(url))
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from "hono";
```
yusukebe, line 143: I think the following is best.

* Use `app.all()` instead of `c.req.path.startsWith()`. It's Hono-way.
* Fixed the logics.
* Return the response with `c.body()` to use context values.

```diff
diff --git a/src/content/docs/workers/examples/ab-testing.mdx b/src/content/docs/workers/examples/ab-testing.mdx
index a4c8dfe16..22fedf721 100644
--- a/src/content/docs/workers/examples/ab-testing.mdx
+++ b/src/content/docs/workers/examples/ab-testing.mdx
@@ -148,23 +148,23 @@ const app = new Hono();
 const NAME = "myExampleWorkersABTest";

 // Middleware to handle A/B testing logic
-app.use("*", async (c) => {
-       // Enable Passthrough to allow direct access to control and test routes
-       if (c.req.path.startsWith("/control") || c.req.path.startsWith("/test")) {
-               return fetch(c.req.raw);
-       }
+
+// Enable Passthrough to allow direct access to control and test routes
+app.all("/control/*", (c) => fetch(c.req.raw));
+app.all("/test/*", (c) => fetch(c.req.raw));
+
+app.all("*", async (c) => {
+       const url = new URL(c.req.url);

        // Determine which group this requester is in
        const abTestCookie = getCookie(c, NAME);

        if (abTestCookie === "control") {
                // User is in control group
-               c.req.path = "/control" + c.req.path;
-               return fetch(url);
+               url.pathname = "/control" + c.req.path;
        } else if (abTestCookie === "test") {
                // User is in test group
-               url.pathname = "/test" + url.pathname;
-               return fetch(c.req.url);
+               url.pathname = "/test" + c.req.path;
        } else {
                // If there is no cookie, this is a new client
                // Choose a group and set the cookie (50/50 split)
@@ -172,17 +172,10 @@ app.use("*", async (c) => {

                // Update URL path based on assigned group
                if (group === "control") {
-                       c.req.path = "/control" + c.req.path;
+                       url.pathname = "/control" + c.req.path;
                } else {
-                       c.req.path = "/test" + c.req.path;
+                       url.pathname = "/test" + c.req.path;
                }
-
-               // Fetch from origin with modified path
-               const res = await fetch(c.req.url);
-
-               // Create a new response to avoid immutability issues
-               const newResponse = new Response(res.body, res);
-
                // Set cookie to enable persistent A/B sessions
                setCookie(c, NAME, group, {
                        path: "/",
@@ -191,15 +184,11 @@ app.use("*", async (c) => {
                        // httpOnly: true,
                        // sameSite: 'strict',
                });
+       }

-               // Copy the Set-Cookie header to the response
-               newResponse.headers.set(
-                       "Set-Cookie",
-                       c.res.headers.get("Set-Cookie") || "",
-               );
+       const res = await fetch(url);

-               return newResponse;
-       }
+       return c.body(res.body!, res);
 });

 export default app;
```

src/content/docs/workers/examples/accessing-the-cloudflare-object.mdx
```diff
@@ -56,6 +56,32 @@ export default {
 } satisfies ExportedHandler;
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from "hono";
+
+type Bindings = {};
```
yusukebe: This `Bindings` is unnecessary.

```diff
diff --git a/src/content/docs/workers/examples/accessing-the-cloudflare-object.mdx b/src/content/docs/workers/examples/accessing-the-cloudflare-object.mdx
index 98c703c4c..2c3d21d87 100644
--- a/src/content/docs/workers/examples/accessing-the-cloudflare-object.mdx
+++ b/src/content/docs/workers/examples/accessing-the-cloudflare-object.mdx
@@ -61,9 +61,7 @@ export default {
 ```ts
 import { Hono } from "hono";

-type Bindings = {};
-
-const app = new Hono<{ Bindings: Bindings }>();
+const app = new Hono();

 app.get("*", async (c) => {
        // Access the raw request to get the cf object
```

src/content/docs/workers/examples/aggregate-requests.mdx
```diff
@@ -58,6 +58,34 @@ export default {
 } satisfies ExportedHandler;
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from 'hono';
+
+type Bindings = {};
```
yusukebe: This is unnecessary.

src/content/docs/workers/examples/block-on-tls.mdx
```diff
@@ -69,6 +69,45 @@ export default {
...
+```ts
+import { Hono } from "hono";
+
+type Bindings = {};
+
+const app = new Hono<{ Bindings: Bindings }>();
+
+// Middleware to check TLS version
+app.use("*", async (c, next) => {
+	try {
```
yusukebe: Using `app.onError()` instead of `try`-`catch` is better since it's more Hono-way.

```diff
diff --git a/src/content/docs/workers/examples/block-on-tls.mdx b/src/content/docs/workers/examples/block-on-tls.mdx
index 5fd55f1c8..662c0c715 100644
--- a/src/content/docs/workers/examples/block-on-tls.mdx
+++ b/src/content/docs/workers/examples/block-on-tls.mdx
@@ -74,31 +74,29 @@ export default {
 ```ts
 import { Hono } from "hono";

-type Bindings = {};
-
-const app = new Hono<{ Bindings: Bindings }>();
+const app = new Hono();

 // Middleware to check TLS version
 app.use("*", async (c, next) => {
-       try {
-               // Access the raw request to get the cf object with TLS info
-               const request = c.req.raw;
-               const tlsVersion = request.cf?.tlsVersion;
-
-               // Allow only TLS versions 1.2 and 1.3
-               if (tlsVersion !== "TLSv1.2" && tlsVersion !== "TLSv1.3") {
-                       return c.text("Please use TLS version 1.2 or higher.", 403);
-               }
+       // Access the raw request to get the cf object with TLS info
+       const request = c.req.raw;
+       const tlsVersion = request.cf?.tlsVersion;

-               // Continue to the next handler if TLS version is acceptable
-               return next();
-       } catch (err: any) {
-               // Handle errors (especially in preview mode where request.cf might not exist)
-               console.error(
-                       "request.cf does not exist in the previewer, only in production",
-               );
-               return c.text(`Error in workers script: ${err.message}`, 500);
+       // Allow only TLS versions 1.2 and 1.3
+       if (tlsVersion !== "TLSv1.2" && tlsVersion !== "TLSv1.3") {
+               return c.text("Please use TLS version 1.2 or higher.", 403);
        }
+
+       // Continue to the next handler if TLS version is acceptable
+       await next();
+});
+
+app.onError((err, c) => {
+       // Handle errors (especially in preview mode where request.cf might not exist)
+       console.error(
+               "request.cf does not exist in the previewer, only in production",
+       );
+       return c.text(`Error in workers script: ${err.message}`, 500);
 });

 app.get("/", async (c) => {
```

src/content/docs/workers/examples/bulk-origin-proxy.mdx
```diff
@@ -74,6 +74,39 @@ export default {
...
+	if (url.hostname in ORIGINS) {
+		const target = ORIGINS[url.hostname];
+		url.hostname = target;
+
+		// If it is, proxy request to that third party origin
+		return proxy(url, c.req.raw);
+	}
+
+	// Otherwise, process request as normal
+	return fetch(c.req.raw);
```
yusukebe: You can use the `proxy()` here as well, and it's better because `proxy()` can remove unnecessary headers. 

```ts
return proxy(c.req.raw)
```

In other places, you can use `proxy()` instead of `fetch`, but I think it's not necessary to replace all `fetch`s.

src/content/docs/workers/examples/bulk-redirects.mdx
```diff
@@ -99,4 +99,44 @@ async def on_fetch(request):
...
+  const location = redirectMap.get(path);
```
yusukebe: That is not fixed. Use the following:

```ts
const location = redirectMap.get(c.req.path);
```

src/content/docs/workers/examples/geolocation-hello-world.mdx
```diff
@@ -137,4 +137,76 @@ async def on_fetch(request):
...
+		timezone: string;
+		// Add other CF properties as needed
+	};
+}
+
+const app = new Hono();
+
+app.get("*", (c) => {
+	// Cast the raw request to include Cloudflare-specific properties
+	const request = c.req.raw as RequestWithCf;
```
yusukebe: You don't need casting if the `tsconfig.json` is set properly.

```ts
const request = c.req.raw;
```

src/content/docs/workers/examples/basic-auth.mdx
```diff
@@ -328,6 +328,52 @@ async fn fetch(req: Request, env: Env, _ctx: Context) -> Result<Response> {
...
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+/**
+ * Shows how to restrict access using the HTTP Basic schema with Hono.
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication
+ * @see https://tools.ietf.org/html/rfc7617
+ */
+
+import { Hono } from 'hono';
```
yusukebe: The logic is wrong. Please rewrite the code to this:

```ts
import { Hono } from "hono";
import { basicAuth } from "hono/basic-auth";

// Define environment interface
interface Env {
	Bindings: {
		USERNAME: string;
		PASSWORD: string;
	};
}

const app = new Hono<Env>();

// Public homepage - accessible to everyone
app.get("/", (c) => {
	return c.text("Anyone can access the homepage.");
});

// Admin route - protected with Basic Auth
app.get(
	"/admin",
	async (c, next) => {
		const auth = basicAuth({
			username: c.env.USERNAME,
			password: c.env.PASSWORD,
		});
		return await auth(c, next);
	},
	(c) => {
		// Success! User is authenticated
		return c.text("🎉 You have private access!", 200, {
			"Cache-Control": "no-store",
		});
	},
);

export default app;
```

src/content/docs/workers/examples/logging-headers.mdx
```diff
@@ -62,6 +62,37 @@ async fn fetch(req: HttpRequest, _env: Env, _ctx: Context) -> Result<Response> {
...
+  console.log('Headers spread:', [...c.req.raw.headers]);
+  
+  // 3. Using Object.fromEntries to convert to an object
+  console.log('Headers as Object:', Object.fromEntries(c.req.raw.headers));
+  
+  // 4. Hono's built-in header accessor (for individual headers)
+  console.log('User-Agent:', c.req.header('User-Agent'));
+  
+  // 5. Using c.req.headers to get all headers
+  console.log('All headers from Hono context:', c.req.headers);
```
yusukebe: `c.req.headers` should be `c.req.header()`:

```ts
console.log('All headers from Hono context:', c.req.header());
```

src/content/docs/workers/examples/modify-request-property.mdx
```diff
@@ -184,4 +184,64 @@ async def on_fetch(request):
...
+const app = new Hono();
+
+app.all('*', async (c) => {
+  /**
+   * Example someHost is set up to return raw JSON
+   */
+  const someHost = "example.com";
+  const someUrl = "https://foo.example.com/api.js";
+
+  try {
```
yusukebe: You don't have to `try`-`catch` because Hono handles errors by default.

src/content/docs/workers/examples/modify-request-property.mdx
```diff
@@ -184,4 +184,64 @@ async def on_fetch(request):
...
+        "Content-Type": "application/json",
+        "X-Example": "bar"
+      },
+      // Change a Cloudflare feature on the outbound response
+      cf: { apps: false },
+    };
+
+    // Create a new request
+    // First create a clone of the original request with the new properties
+    const requestClone = new Request(c.req.raw, newRequestInit);
```
yusukebe: I think it's good not to separate `newRequestInit` properties because if you write them in `new Request,` the type inference will be enabled.

CleanShot 2025-04-03 at 19 32 55@2x

```ts
const requestClone = new Request(c.req.raw, {
  // Change method
  method: 'POST',
  // Change body
  body: JSON.stringify({ bar: 'foo' }),
  // Change the redirect mode
  redirect: 'follow' as RequestRedirect,
  // Change headers, note this method will erase existing headers
  headers: {
    'X-Example': 'bar',
  },
  // Change a Cloudflare feature on the outbound response
  cf: { apps: false },
})
```

src/content/docs/workers/examples/signing-requests.mdx
```diff
@@ -244,6 +244,213 @@ export default {
 } satisfies ExportedHandler<Env>;
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Buffer } from "node:buffer";
```
yusukebe, line 250: This code is not working and throwing errors. Can you correct it?

src/content/docs/workers/examples/turnstile-html-rewriter.mdx
```diff
@@ -95,6 +95,120 @@ export default {
 } satisfies ExportedHandler<Env>;
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from 'hono';
+import { html } from 'hono/html';
```
yusukebe: This line is unnecessary.

src/content/docs/workers/examples/turnstile-html-rewriter.mdx
```diff
@@ -95,6 +95,120 @@ export default {
...
+
+interface Env {
+  SITE_KEY: string;
+  TURNSTILE_ATTR_NAME?: string;
```
yusukebe: You may have to add the following type definition since it uses `c.env.SECRET_KEY` after here.

```ts
SECRET_KEY?: string
```

src/content/docs/workers/examples/turnstile-html-rewriter.mdx
```diff
@@ -95,6 +95,120 @@ export default {
...
+  // Verify the token with Turnstile API
+  const verifyResult = await fetch(
+    'https://challenges.cloudflare.com/turnstile/v0/siteverify',
+    {
+      method: 'POST',
+      body: verifyFormData
+    }
+  );
+  
+  const outcome = await verifyResult.json();
```
yusukebe: Adding the generics will remove the type error.

CleanShot 2025-04-03 at 19 42 30@2x

CleanShot 2025-04-03 at 19 42 46@2x




```ts
const outcome = await verifyResult.json<{ success: boolean }>()
```

src/content/docs/workers/examples/cache-api.mdx
```diff
@@ -132,4 +132,34 @@ async def on_fetch(request, _env, ctx):
     return response
 ```
 
+</TabItem> <TabItem label="Hono" icon="seti:typescript">
+
+```ts
+import { Hono } from "hono";
+import type { Context } from "hono";
```
yusukebe: Please remove this line.

src/content/docs/workers/examples/cache-api.mdx
```diff
@@ -132,4 +132,34 @@ async def on_fetch(request, _env, ctx):
...
+import { cache } from "hono/cache";
+
+type Env = {
```
yusukebe: This definition is not necessary.

coffee-mug:
Hello @yusukebe I am sorry for the late reply, fixes embarked, may I let you confirm and merge please?

src/content/docs/workers/examples/basic-auth.mdx
yusukebe: @coffee-mug 

Can you see [this comment](https://github.com/cloudflare/cloudflare-docs/pull/21258/files#r2026696112) and fix it?

yusukebe:
@coffee-mug 

Thanks! I've left a comment: https://github.com/cloudflare/cloudflare-docs/pull/21258/files#r2041988024

src/content/docs/workers/examples/basic-auth.mdx
coffee-mug: Woaw, i am really sorry for that, should be fixed and conform to your example now.

yusukebe (approved):
LGTM!

kodster28 (approved):

PR #21288, opened 2025-04-01
[Workers] Service binding URL requirement

### LINKED ISSUE #18581: Clarify HTTP Service Binding Needs Valid URL

https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/http/

When calling a bound worker's fetch method, it is not clear and not mentioned what the URL should be. When thinking about it, the worker is **bound**, suggesting there should be no need for a URL.

I have run in to the same issue as these
https://community.cloudflare.com/t/service-binding-to-other-worker-not-working/559030
https://community.cloudflare.com/t/service-binding-pages-function-to-worker-cant-call-path/647192

The solution appears to be to add any valid URL with a hostname (it will be ignored anyway).
I am suggesting mentioning this in the docs, for the docs only give examples of passing through or cloning requests with internal, already-valid URLs

### Additional information

_No response_

ToriLindsay:
Fixes #18581

### DIFF
--- a/src/content/docs/workers/runtime-apis/bindings/service-bindings/http.mdx
+++ b/src/content/docs/workers/runtime-apis/bindings/service-bindings/http.mdx
@@ -56,3 +56,20 @@ export default {
 	},
 };
 ```
+
+:::note
+
+If you create a new request manually, rather than forwarding an existing one, you need to provide a valid URL with a hostname. For example:
+
+```js
+export default {
+  async fetch(request, env) {
+    // provide a valid URL
+    let newRequest = new Request("https://valid-url.com", { method: "GET" });
+    let response = await env.WORKER_B.fetch(newRequest);
+    return response;
+  }
+};
+```
+
+:::
--- a/src/content/docs/workers/runtime-apis/bindings/service-bindings/http.mdx
+++ b/src/content/docs/workers/runtime-apis/bindings/service-bindings/http.mdx
@@ -59,7 +59,7 @@ export default {
 
 :::note
 
-If you construct a new request manually, rather than forwarding an existing one, you need to ensure that you are providing a valid and fully-qualified URL with a hostname. For example:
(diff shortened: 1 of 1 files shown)


ToriLindsay:
@GregBrimble Would you mind checking this for accuracy and that it fixes #18581 ? I just took my best guess. Also, let me know if you think it needs more of an acknowledgement that (as I understand it) the URL is fairly arbitrary and will be ignored, but is needed anyway.

src/content/docs/workers/runtime-apis/bindings/service-bindings/http.mdx
```diff
@@ -56,3 +56,20 @@ export default {
 	},
 };
 ```
+
+:::note
+
+If you create a new request manually, rather than forwarding an existing one, you need to provide a valid URL with a hostname. For example:
```
GregBrimble:
```suggestion
If you construct a new request manually, rather than forwarding an existing one, you need to ensure that you are providing a valid and fully-qualified URL with a hostname. For example:
```

Take em or leave em

GregBrimble (approved):

GregBrimble:
It's arbitrary, in the sense that a user can provide any URL at all. But it still matters as a legitimate value that the callee will receive and can act on if it chooses to.

src/content/docs/workers/runtime-apis/bindings/service-bindings/http.mdx
```diff
@@ -56,3 +56,20 @@ export default {
 	},
 };
 ```
+
+:::note
+
+If you construct a new request manually, rather than forwarding an existing one, you need to ensure that you are providing a valid and fully-qualified URL with a hostname. For example:
```
ToriLindsay:
```suggestion
If you construct a new request manually, rather than forwarding an existing one, ensure that you provide a valid and fully-qualified URL with a hostname. For example:
```

daisyfaithauma (approved):

PR #21326, opened 2025-04-02
durable objects: FURTHER clarify blockConcurrencyWhile

elithrar:


Oxyjun (approved):

src/content/docs/durable-objects/api/state.mdx
```diff
@@ -73,12 +73,21 @@ The request context for a Durable Objects extends at least 60 seconds after the
...
+* Another use case is executing `async` operations based on the current state of the Durable Object and using `blockConcurrencyWhile` to prevent that state from changing while yielding the event loop.
+* If the callback throws an exception, the object will be terminated and reset. This ensures that the object cannot be left stuck in an uninitialized state if something fails unexpectedly.
+* To avoid this behavior, enclose the body of your callback in a `try...catch` block to ensure it cannot throw an exception.
 
 To help mitigate deadlocks there is a 30 second timeout applied when executing the callback. If this timeout is exceeded, the Durable Object will be reset. It is best practice to have the callback do as little work as possible to improve overall request throughput to the Durable Object.
 
+:::note
+
+You should only need `blockConcurrencyWhile` if you are making additional, asynchronous calls (such as to another API or service) and cannot tolerate those other requests changing the internal state of the Durable Object when the event loop is yielded.
```
lambrospetrou: It's not necessarily those other requests that will change the internal state of the DO, but most likely other incoming requests to the DO?

```suggestion
You should only need `blockConcurrencyWhile` if you are making additional, asynchronous calls (such as to another API or service), and cannot tolerate other requests processed by the Durable Object changing its internal while the event loop is yielded from the original request.
```

PR #21518, opened 2025-04-08
Add "Build a Remote MCP client" to Guides section

dinasaur404:
### Summary

In the Guides section of the Agents docs, adding a page for building a remote MCP client which redirects to https://github.com/cloudflare/ai/tree/main/demos/mcp-client.


irvinebroque (approved):
We should make good docs for this one day or at some point add here, but now not yet time as so much changing

PR #21589, opened 2025-04-10
[D1] Update e-commerce tutorial

harshil1712:
### Summary

Adds retry logic to the tutorial

### DIFF
--- a/src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
+++ b/src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
@@ -552,9 +552,79 @@ For development purposes, you can also execute the insert statements on the loca
 npx wrangler d1 execute fast-commerce --command "INSERT INTO products (id, name, description, price, inventory, category) VALUES (1, 'Fast Ergonomic Chair', 'A comfortable chair for your home or office', 100.00, 10, 'Furniture'), (2, 'Fast Organic Cotton T-shirt', 'A comfortable t-shirt for your home or office', 20.00, 100, 'Clothing'), (3, 'Fast Wooden Desk', 'A wooden desk for your home or office', 150.00, 5, 'Furniture'), (4, 'Fast Leather Sofa', 'A leather sofa for your home or office', 300.00, 3, 'Furniture'), (5, 'Fast Organic Cotton T-shirt', 'A comfortable t-shirt for your home or office', 20.00, 100, 'Clothing')"
 ```
 
+## Step 6: Add retry logic
 
+To make the application more resilient, you can add retry logic to the API routes. Create a new file called `retry.ts` in the `src` directory.
 
-## Step 6: Update the API routes
+```ts
+export interface RetryConfig {
+  maxRetries?: number;
+  initialDelay?: number;
+  maxDelay?: number;
+  backoffFactor?: number;
+}
+
+const shouldRetry = (error: unknown): boolean => {
+  const errMsg = error instanceof Error ? error.message : String(error);
+  return (
+    errMsg.includes("Network connection lost") ||
+    errMsg.includes("storage caused object to be reset") ||
+    errMsg.includes("reset because its code was updated")
+  );
+};
+
+export const defaultRetryConfig: RetryConfig = {
+  maxRetries: 3,
+  initialDelay: 100,
+  maxDelay: 1000,
+  backoffFactor: 2,
+};
+
+export async function withRetry<T>(
+  operation: () => Promise<T>,
+  config: RetryConfig = defaultRetryConfig
+): Promise<T> {
+  const maxRetries = config.maxRetries ?? defaultRetryConfig.maxRetries!;
+  const initialDelay = config.initialDelay ?? defaultRetryConfig.initialDelay!;
+  const maxDelay = config.maxDelay ?? defaultRetryConfig.maxDelay!;
+  const backoffFactor =
+    config.backoffFactor ?? defaultRetryConfig.backoffFactor!;
+
+  let lastError: Error | unknown;
+  let delay = initialDelay;
+
+  for (let attempt = 0; attempt <= maxRetries; attempt++) {
+    try {
+      const result = await operation();
+      return result;
+    } catch (error) {
+      lastError = error;
+
+      if (!shouldRetry(error) || attempt === maxRetries) {
+        throw error;
+      }
+
+      // Wait for the calculated delay
+      await new Promise((resolve) => setTimeout(resolve, delay));
+
+      // Calculate next delay with exponential backoff
+      delay = Math.min(delay * backoffFactor, maxDelay);
+    }
+  }
+
+  throw lastError;
+}
+```
+
+The `withRetry` function is a utility function that retries a given operation with exponential backoff. It takes a configuration object as an argument, which allows you to customize the number of retries, initial delay, maximum delay, and backoff factor. It will only retry the operation if the error is due to a network connection loss, storage reset, or code update.
+
+Next, update the `src/index.ts` file to import the `withRetry` function and use it in the API routes.
+
+```ts
+import { withRetry } from "./retry";
+```
+
+## Step 7: Update the API routes
 
 Update the API routes to connect to the D1 database.
 
@@ -574,7 +644,8 @@ app.post('/api/product', async (c) => {
 
 	const { id } = product;
 
-	try {
+	return await withRetry(async() =>{ 
+		try{
 		// Check if the product exists
 		const { results } = await session.prepare('SELECT * FROM products where id = ?').bind(id).run();
 		if (results.length === 0) {
@@ -620,7 +691,7 @@ app.post('/api/product', async (c) => {
 	} catch (e) {
 		console.error(e);
 		return c.json({ message: 'Error upserting product' }, 500);
-	}
+	}});
 });
 
 ```
@@ -651,7 +722,8 @@ app.get('/api/products', async (c) => {
 
 	const session = db.withSession(bookmark);
 
-	try {
+	return await withRetry(async() => {
+		try {
 		const { results } = await session.prepare('SELECT * FROM products').all();
 
 		const latestBookmark = session.getBookmark();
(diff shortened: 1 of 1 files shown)


src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -552,9 +552,79 @@ For development purposes, you can also execute the insert statements on the loca
...
+      // Wait for the calculated delay
+      await new Promise((resolve) => setTimeout(resolve, delay));
+
+      // Calculate next delay with exponential backoff
+      delay = Math.min(delay * backoffFactor, maxDelay);
+    }
+  }
+
+  throw lastError;
+}
```
alsuren: a bunch of things here:

1) (just a little cleanup) either use Partial<...> or Required<...> to make the types of config and defaultRetryConfig a bit nicer and avoid the ! when unpacking (I made RetryConfig fields mandatory and then made the function take Partial<RetryConfig>. You could alternatively keep RetryConfig as it is and make defaultRetryConfig: Required<RetryConfig>)
2) (again a little cleanup) I found the unpacking of config quite noisy, so I used ...spread. It might be that your way is easier to read for people who are not used to that way of doing things, in which case keep it as it is.
3) extracted out a sleep() helper because every good js project should have that somewhere in their codebase ;-) .

```suggestion
export interface RetryConfig {
  maxRetries: number;
  initialDelay: number;
  maxDelay: number;
  backoffFactor: number;
}

const shouldRetry = (error: unknown): boolean => {
  const errMsg = error instanceof Error ? error.message : String(error);
  return (
    errMsg.includes("Network connection lost") ||
    errMsg.includes("storage caused object to be reset") ||
    errMsg.includes("reset because its code was updated")
  );
};

export const defaultRetryConfig: RetryConfig = {
  maxRetries: 3,
  initialDelay: 100,
  maxDelay: 1000,
  backoffFactor: 2,
};

function sleep(milliseconds: number) {
  return new Promise((resolve) => setTimeout(resolve, milliseconds));
}

export async function withRetry<T>(
  operation: () => Promise<T>,
  config: Partial<RetryConfig> = defaultRetryConfig
): Promise<T> {
  const { maxRetries, initialDelay, maxDelay, backoffFactor } = {
    ...defaultRetryConfig,
    ...config,
  };

  let lastError: Error | unknown;
  let delay = initialDelay;

  for (let attempt = 0; attempt <= maxRetries; attempt++) {
    try {
      const result = await operation();
      return result;
    } catch (error) {
      lastError = error;

      if (!shouldRetry(error) || attempt === maxRetries) {
        throw error;
      }

      // Wait for the calculated delay
      await sleep(delay);

      // Calculate next delay with exponential backoff
      delay = Math.min(delay * backoffFactor, maxDelay);
    }
  }

  throw lastError;
}
```

Things that I've not done, but might be worth considering:

3) add some randomness, to avoid synchronising retries taking down the durable object repeatedly (sleep between delay and delay  * 2  by doing `sleep(delay * (1 + Random.random())` or something). - this one is actually pretty important, I think.
4) use "break;" rather than `throw error;` so there is one less way to exit the function. Not sure which is more readable though, so I left it.

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -620,7 +691,7 @@ app.post('/api/product', async (c) => {
 	} catch (e) {
 		console.error(e);
 		return c.json({ message: 'Error upserting product' }, 500);
```
alsuren: 1) this try/catch wants to stay *outside* the retry wrapper. As it's written here, the retry wrapper will never see a raised exception (just a `return c.json(..., 500)`) and so it won't retry anything.

2) is this function idempotent? In this case I think it's fine because it is (the difference between `{ message: 'Product updated' }` and `{ message: 'Product inserted' }` doesn't really matter), but in general I would be cautious about putting retries around functions that do writes, because you could get duplicated inserts and all sorts of nonsense going on.

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -666,7 +738,7 @@ app.get('/api/products', async (c) => {
 	} catch (e) {
 		console.error(e);
 		return c.json([]);
-	}
+	}})
```
alsuren: same as above: final catch wants to live outside the withRetry wrapper

This is just doing a read though, so retrying is more obviously safe.

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -697,24 +769,26 @@ app.get('/api/products/:id', async (c) => {
...
+					maxAge: 60 * 60, // 1 hour
+				});
+	
+			console.log(results);
+	
+			return c.json(results);
+		} catch (e) {
+			console.error(e);
+			return c.json([]);
+		}
```
alsuren: same again: move the try/catch outside the withRetry

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -474,17 +475,18 @@ Create a new D1 database by running the following command:
 
 ```sh
 npx wrangler d1 create fast-commerce
-```
+````
```
alsuren: what does 4 closing backticks do?

(here and above)

harshil1712: Ah, could have been the auto formatting on save, messing this up. Nice catch!

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -552,77 +554,158 @@ For development purposes, you can also execute the insert statements on the loca
...
+      delay = Math.min(delay * backoffFactor, maxDelay);
+    }
+  }
+
+  throw lastError;
+}
+```
 
-Update the API routes to connect to the D1 database.
+The `withRetry` function is a utility function that retries a given operation with exponential backoff. It takes a configuration object as an argument, which allows you to customize the number of retries, initial delay, maximum delay, and backoff factor. It will only retry the operation if the error is due to a network connection loss, storage reset, or code update.
```
alsuren, line 627: I feel like we want some kind of warning about the pitfalls of adding retries to distributed systems. I would prefer to link out to a reputable source or something, but my google-fu is failing me. One example from a quick search is this section:

> ## Common Challenges with Retry Pattern in Microservices
> Implementing the Retry Pattern in microservices can present several challenges:
>
> * Retry Storms: When many services retry simultaneously, it can lead to a surge of requests that exacerbates the problem rather than alleviating it. Use exponential backoff and jitter to spread out retry attempts and prevent simultaneous retries.
> * Overloading Services: Frequent retries can put additional load on the affected service, potentially causing further degradation or failures. Implement rate limiting and circuit breakers alongside retries to manage load and avoid overwhelming services.
> * Increased Latency: Retries can increase the overall latency of requests, as each retry adds additional delay. Balance retry attempts with acceptable latency thresholds and use appropriate backoff strategies to manage the delay impact.
> * Complexity in Configuration: Configuring retries with appropriate limits, backoff strategies, and conditions can become complex and error-prone. Use well-documented libraries or frameworks for retry logic and follow best practices to simplify configuration.
> * Dependency Management: If a microservice relies on multiple other services, managing retries across all dependencies can become challenging and may require coordinated retry policies. Implement consistent retry policies and use distributed tracing to monitor and manage retries across services.
> * Testing and Validation: Testing retry logic under various failure scenarios can be difficult, and improper testing may lead to unexpected behavior in production. Use comprehensive testing strategies, including simulated failures and load testing, to validate retry behavior and performance.
> * Data Consistency Issues: Retries can sometimes lead to inconsistencies, especially if the operation being retried involves data modifications. Implement idempotent operations where possible, ensuring that retries do not cause unintended side effects or data inconsistencies.

-- https://www.geeksforgeeks.org/retry-pattern-in-microservices/#:~:text=impact%20system%20performance.-,Common%20Challenges%20with%20Retry%20Pattern%20in%20Microservices,-Implementing%20the%20Retry

The data consistency one is the one that I jumped at, but someone who is new to this will probably benefit from seeing a big block of warnings like this so they don't go about blindly adding retries everywhere and getting themselves into trouble.

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -552,77 +554,158 @@ For development purposes, you can also execute the insert statements on the loca
...
+const sleep = (ms: number): Promise<void> => {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+};
+
+export const defaultRetryConfig: RetryConfig = {
+  maxRetries: 3,
+  initialDelay: 100,
+  maxDelay: 1000,
+  backoffFactor: 2,
+};
```
alsuren, line 588: (optional) maybe move this up next to the RetryConfig type definition?

I also noticed that we're not ever passing the RetryConfig argument in anywhere, but I suspect that it will be useful later for anyone who is copy-pasting this code, so let's keep it.

src/content/docs/d1/tutorials/using-read-replication-for-e-com/index.mdx
```diff
@@ -552,77 +554,158 @@ For development purposes, you can also execute the insert statements on the loca
...
+
+export async function withRetry<T>(
+  operation: () => Promise<T>,
+	config: Partial<RetryConfig> = defaultRetryConfig,
```
alsuren, line 592: what's going on with the indentation here? Can you run this through an auto-formatter again?

alsuren (approved):
I think this looks okay as long as it still renders okay and you've checked that it works.

I will let you figure out your formatting issues and check it yourself. 

(I wonder if we should have formatting checked in CI to avoid having people's autoformatters fighting with each other? Not a thing for right now though)

Oxyjun (approved):

PR #21609, opened 2025-04-10
[Docs Site] Add query prop to APIRequest and CURL

KianNH:
### Summary

Add query prop to APIRequest and CURL

### Screenshots (optional)

image

### DIFF
--- a/src/content/docs/style-guide/components/curl.mdx
+++ b/src/content/docs/style-guide/components/curl.mdx
@@ -23,6 +23,10 @@ import { CURL } from "~/components";
 	json={{
 		key: "va'l'ue",
 	}}
+	query={{
+		foo: "bar",
+		bar: "baz",
+	}}
 	code={{
 		mark: "value"
 	}}
@@ -75,6 +79,12 @@ JSON data to include in the request.
 **type:** `Record<string, any>`
 
 The FormData payload to send.
+
+### `query`
+
+**type:** `Record<string, string>`
+
+URL query parameters to append to the request URL.
 
 ### `code`
 
--- a/src/content/docs/style-guide/components/curl.mdx
+++ b/src/content/docs/style-guide/components/curl.mdx
@@ -82,7 +82,7 @@ The FormData payload to send.
 
 ### `query`
 
-**type:** `Record<string, string>`
+**type:** `Record<string, string | string[]>`
 
 URL query parameters to append to the request URL.
 
--- a/src/content/docs/style-guide/components/curl.mdx
+++ b/src/content/docs/style-guide/components/curl.mdx
@@ -25,10 +25,10 @@ import { CURL } from "~/components";
 	}}
 	query={{
 		foo: "bar",
-		bar: "baz",
+		bar: ["baz", "qux"],
 	}}
 	code={{
-		mark: "value"
+		mark: "value",
 	}}
 />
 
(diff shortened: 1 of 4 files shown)


kodster28 (approved):

pedrosousa:
Nit:
I don't think it will affect any of our cURL/API examples, but I noticed that repeated keys in the query string are not supported by the component (it uses the last value only), while being a valid query string.

Example:
```jsx
<CURL
  url="https://httpbin.org/anything"
  method="POST"
  json={{
    key: "va'l'ue",
  }}
  query={{
    foo: "bar",
    foo: "baz",
    bar: "baz",
  }}
  code={{
    mark: "value",
  }}
/>
```

Result:

```
curl "https://httpbin.org/anything?foo=baz&bar=baz" \
  --request POST \
  --json '{
    "key": "va'\''l'\''ue"
  }'
```


Expected result:

```
curl "https://httpbin.org/anything?foo=bar&foo=baz&bar=baz" \
  --request POST \
  --json '{
    "key": "va'\''l'\''ue"
  }'
```



MDN has a JavaScript example with a repeated key:

```js
const url = new URL("https://example.com?foo=1&bar=2");
const params = new URLSearchParams(url.search);

// Add a second foo parameter.
params.append("foo", 4);
console.log(params.toString()); // Prints 'foo=1&bar=2&foo=4'
```

KianNH:
> Nit: I don't think it will affect any of our cURL/API examples, but I noticed that repeated keys in the query string are not supported by the component (it uses the last value only), while being a valid query string.
> 
> Example:
> 
> ```js
> <CURL
>   url="https://httpbin.org/anything"
>   method="POST"
>   json={{
>     key: "va'l'ue",
>   }}
>   query={{
>     foo: "bar",
>     foo: "baz",
>     bar: "baz",
>   }}
>   code={{
>     mark: "value",
>   }}
> />
> ```
> 
> Result:
> 
> ```
> curl "https://httpbin.org/anything?foo=baz&bar=baz" \
>   --request POST \
>   --json '{
>     "key": "va'\''l'\''ue"
>   }'
> ```
> 
> Expected result:
> 
> ```
> curl "https://httpbin.org/anything?foo=bar&foo=baz&bar=baz" \
>   --request POST \
>   --json '{
>     "key": "va'\''l'\''ue"
>   }'
> ```
> 
> MDN has a JavaScript example with a repeated key:
> 
> ```js
> const url = new URL("https://example.com?foo=1&bar=2");
> const params = new URLSearchParams(url.search);
> 
> // Add a second foo parameter.
> params.append("foo", 4);
> console.log(params.toString()); // Prints 'foo=1&bar=2&foo=4'
> ```

Added support for this in https://github.com/cloudflare/cloudflare-docs/pull/21609/commits/52177d919a4636cbba3eee140e91a38670a0ae0f - something like `query: { foo: ["bar", "baz"] }` would now result in `?foo=bar&foo=baz`.

src/content/docs/style-guide/components/curl.mdx
```diff
@@ -76,6 +80,12 @@ JSON data to include in the request.
 
 The FormData payload to send.
 
+### `query`
+
+**type:** `Record<string, string>`
```
pedrosousa: @KianNH We should probably update the type here, since we now support `Array` values.

pedrosousa (approved):

src/content/docs/style-guide/components/curl.mdx
KianNH:
```suggestion
**type:** `Record<string, string | string[]>`
```

PR #21631, opened 2025-04-11
Vale updates

kodster28:
Re-adds some Vale styles in anticipation of turning Hyperlint automations on in the background.


KianNH (approved):

.github/styles/cloudflare/Contractions.yml
pedrosousa, line 1: @kodster28 - This check seems to be _promoting_ the use of contractions, from the warnings I'm getting through my local Vale extension (in VS Code).
image

PR #21668, opened 2025-04-14
[Zero Trust Devices] Deprecate several endpoints

marcinflare:
### Summary

Deprecate several Zero Trust Devices endpoints.

### Documentation checklist

### DIFF
--- a/src/content/release-notes/api-deprecations.yaml
+++ b/src/content/release-notes/api-deprecations.yaml
@@ -8,14 +8,16 @@ entries:
   - publish_date: "2025-04-14"
     title: "Zero Trust Devices"
     description: |-
-      End of life date: November 25, 2025
+      Deprecation date: April 14, 2025
+
+      End of life date: November 11, 2025
 
       We are changing the definition of Devices. Devices are going to represent the real-world machines while
       the relation between Users and Devices will be represented by a new concept - Registrations.
 
-      As a result multiple fields are moving from Device to Registration and we're deprecating the endpoints listed below.
+      As a result multiple fields are moving from Devices to Registrations and we are deprecating the endpoints listed below.
 
-      The deprecated endpoints are not supported on accounts with [multi-user mode](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-multiuser/#enable-multi-user-mode) enabled.
+      The deprecated endpoints are not supported on accounts with [multi-user mode](/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-multiuser/) enabled.
 
       Deprecated API:
 
--- a/src/content/release-notes/api-deprecations.yaml
+++ b/src/content/release-notes/api-deprecations.yaml
@@ -8,7 +8,7 @@ entries:
   - publish_date: "2025-04-14"
     title: "Zero Trust Devices"
     description: |-
-      End of life date: November 25, 2026
+      End of life date: November 25, 2025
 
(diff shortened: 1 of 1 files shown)


src/content/release-notes/api-deprecations.yaml
```diff
@@ -5,6 +5,36 @@ productLink: "/fundamentals/"
 productArea: Core platform
 productAreaLink: /fundamentals/reference/changelog/platform/
 entries:
+  - publish_date: "2025-04-14"
+    title: "Zero Trust Devices"
+    description: |-
+      End of life date: November 25, 2025
```
ranbel: Should we add a Deprecation date? Example: https://developers.cloudflare.com/fundamentals/api/reference/deprecations/#2025-03-20

src/content/release-notes/api-deprecations.yaml
```diff
@@ -5,6 +5,36 @@ productLink: "/fundamentals/"
...
+
+      We are changing the definition of Devices. Devices are going to represent the real-world machines while
+      the relation between Users and Devices will be represented by a new concept - Registrations.
+
+      As a result multiple fields are moving from Device to Registration and we're deprecating the endpoints listed below.
+
+      The deprecated endpoints are not supported on accounts with [multi-user mode](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-multiuser/#enable-multi-user-mode) enabled.
```
ranbel:
```suggestion
      The deprecated endpoints are not supported on accounts with [multi-user mode](/cloudflare-one/connections/connect-devices/warp/deployment/mdm-deployment/windows-multiuser/) enabled.
```

ranbel: Is the multi-user limitation already in effect, or will it happen after Nov 25?

src/content/release-notes/api-deprecations.yaml
```diff
@@ -5,6 +5,36 @@ productLink: "/fundamentals/"
...
 entries:
+  - publish_date: "2025-04-14"
+    title: "Zero Trust Devices"
+    description: |-
+      End of life date: November 25, 2025
+
+      We are changing the definition of Devices. Devices are going to represent the real-world machines while
+      the relation between Users and Devices will be represented by a new concept - Registrations.
+
+      As a result multiple fields are moving from Device to Registration and we're deprecating the endpoints listed below.
```
ranbel:
```suggestion
      As a result multiple fields are moving from Devices to Registrations and we are deprecating the endpoints listed below.
```

ranbel (approved):

src/content/release-notes/api-deprecations.yaml
marcinflare: It's already in effect.
These endpoints technically work but when multi-user mode is enabled they return registration IDs while the endpoint names / parameter names suggest that these are device IDs. That's why we're deprecating them.
Without multi-user mode the results are correct because IDs of registration and device are equal to each other, and there is 1-to-1 relation between device and registration.

src/content/release-notes/api-deprecations.yaml
marcinflare: I'll add it. I'll also move EOL date 2 weeks earlier to avoid US holidays.

marcinflare: I asked around and the suggestion was to:
- set publish_date to the EOL date - done
- treat deprecation and EOL dates the same way - I removed the deprecation date

PR #21713, opened 2025-04-15
Update reference.mdx

ratesangelo:
The Expect-CT header is deprecated. This header was used to enforce Certificate Transparency (CT) requirements, but it is no longer recommended for use. Chromium-based browsers, such as Google Chrome, deprecated the Expect-CT header starting from version 107, as they now enforce CT by default

### Documentation checklist


pedrosousa:
Hello @ratesangelo 
Thank you for the heads up. I've shared this information internally.
However, the "Add security headers" Managed Transform, when enabled, still adds the header, so we will keep the documentation as is for now.

PR #21732, opened 2025-04-15
[Docs site] Add last updated metadata

kodster28:
### Summary

Reporting need + re-create Hugo logic


src/components/overrides/Head.astro
```diff
@@ -12,6 +12,7 @@ const NOINDEX_PRODUCTS = ["email-security", "style-guide", "security"];
 const currentSection = Astro.url.pathname.split("/")[1].replaceAll(".", "");
 
 const entry = Astro.locals.starlightRoute.entry;
+const lastUpdated = Astro.locals.starlightRoute.lastUpdated
```
pedrosousa, line 15:
```suggestion
const lastUpdated = Astro.locals.starlightRoute.lastUpdated;
```

kodster28:
This is broken for the time being

PR #21804, opened 2025-04-17
[Gateway] Custom cert origin warning

maxvp:
PCX-17054

### DIFF
--- a/src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/custom-certificate.mdx
+++ b/src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/custom-certificate.mdx
@@ -21,7 +21,7 @@ You can upload up to five custom root certificates. If your organization require
 :::caution
 Custom certificates are limited to use between your users and the Gateway proxy. Gateway connects to origin servers using publicly trusted certificates, similar to how a browser validates secure websites.
 
-If your users need to connect to self-signed origin servers, set the [untrusted certificate action](/cloudflare-one/policies/gateway/http-policies/#untrusted-certificates) to _Pass through_ in an HTTP policy.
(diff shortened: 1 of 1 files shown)


src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/custom-certificate.mdx
```diff
@@ -19,7 +19,9 @@ Enterprise customers who do not wish to install a [Cloudflare certificate](/clou
 You can upload up to five custom root certificates. If your organization requires more than five certificates, contact your account team.
 
 :::caution
-Custom certificates are limited to use between your users and the Gateway proxy. Connections between Gateway and the origin server will use a Cloudflare certificate.
+Custom certificates are limited to use between your users and the Gateway proxy. Gateway connects to origin servers using publicly trusted certificates, similar to how a browser validates secure websites.
+
+If your users need to connect to self-signed origin servers, set the [untrusted certificate action](/cloudflare-one/policies/gateway/http-policies/#untrusted-certificates) to _Pass through_ in an HTTP policy.
```
caley-b: Since this is a conditional selector that only works if you have also selected the appropriate action from the Action dropdown, would it make sense to add "set the Action to _Allow_" here? This was the only action that instigated the Untrusted certificate action selector to enable for me.

caley-b (approved):
Approving with a small question/suggestion but leaving it up to you.

src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/custom-certificate.mdx
maxvp:
```suggestion
If your users need to connect to self-signed origin servers, create an HTTP Allow policy for the origin server with the [untrusted certificate action](/cloudflare-one/policies/gateway/http-policies/#untrusted-certificates) set to _Pass through_.
```

PR #21818, opened 2025-04-18
[Chore] Release notes updates

kodster28:
### Summary

For products that only use the top-level `/changelog` and don't have a use case for a separate avenue of updates, collapsing their previous `release-notes` into the new changelog experience.

This is a longer-term fix version of the same pain point as #21822


src/components/ProductChangelog.astro
```diff
@@ -0,0 +1,55 @@
+---
+import { getChangelogs, type GetChangelogsOptions } from "~/util/changelog";
```
kodster28, line 2: TODO -> Add in logic to prevent `hidden` entries from showing. Don't think this is exactly going to be a use case, but should be consistent b/c this could trip up folks.

KianNH (approved):

ranbel (approved):

PR #21008, opened 2025-03-20
[Workers] Add getting started guide for API and IaC

markjmiller:
### Summary

There are no official documented examples of how to properly upload a workers script with content and metadata. This is a non-starter for API users. Ideally these examples could live in the [API docs](https://developers.cloudflare.com/api/resources/workers/subresources/scripts/methods/update/) themselves, but we can't make custom examples there yet. This doc serves as a practical guide in the meantime that will unblock users.

### Documentation checklist

- [x] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.

PR #21849, opened 2025-04-21
[Bug] Fix githubcode bug for non-ts blocks

kodster28:
### Summary

PCX-17063


kodster28:
Pulls out bugfix from #21008

KianNH (approved):

PR #21872, opened 2025-04-22
Fix wrangler installation step in Next.js workers guide

mazn4:
### Summary

wrangler now installs as a devDependency (removed type="install")

### Documentation checklist


vicb:
You are right, `@opennextjs/cloudflare@latest` should not be installed as a dev dependency as we use the API.
It has already been fixed.

vicb (approved):

vicb:
Thanks for the PR, good catch!

PR #21968, opened 2025-04-24
fix(kv): Link to new nested API URLs for kv

shrink:
### Summary

I was browsing the kv documentation and encountered some 404s when clicking through to the API documentation. The problem looks to be caused by the introduction of nesting in the kv API documentation, e.g:

```diff
- /api/resources/kv/subresources/namespaces/methods/bulk_delete/
+ /api/resources/kv/subresources/namespaces/subresources/keys/methods/bulk_delete/
```

While making this PR, I found broken links in the kv release notes, so I fixed them... but not sure if editing release notes after the fact is bad. I can remove that change if it's not good.

### DIFF
--- a/src/content/docs/kv/api/delete-key-value-pairs.mdx
+++ b/src/content/docs/kv/api/delete-key-value-pairs.mdx
@@ -68,7 +68,7 @@ Calling the `delete()` method will remove the key and value from your KV namespa
 
 ### Delete data in bulk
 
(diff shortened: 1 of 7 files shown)


Oxyjun (approved):

Oxyjun:
@shrink Thank you so much for catching these links!

I've approved + enabled auto-merge.

PR #21975, opened 2025-04-25
[R2] Update Presigned URL docs

### LINKED ISSUE #19190: [R2] Clarify Presigned URL docs

- https://developers.cloudflare.com/r2/api/s3/presigned-urls/

Developers have reported issues understanding how to use Presigned URL for R2. One of the feedback we have got is the confusion on using Presigned URLs outside of the Workers environment. We need to audit the documentation and make any changes required.

From the developer:
```
Yeah, I did notice that after reading through the entire documentation and those linked pages. The confusing part is the *flow* of the documentation, not necessarily its contents. It went something like this:

1. Search for presigning of urls using R2, with the purpose being to let users upload from a browser, directly to R2
2. Find this page at the top of any kind of search result
3. Find a section dedicated to Workers, think "oh yeah, I'm using that - this is for me" (see image)
4. Become very confused and go down a rabbit hole of various issues, since this section, was, in fact, not for me

Another tidbit is that example here uses `aws4fetch` in the workers example, and none of the other examples use this client. The assumption then is that this is the only client that works with Workers, because:
- This is generally the case with fetch/web-compatible libraries compared to nodejs/other js ones
- The word "alternative" in the section title carries weight with it, signals that other approaches *do not work* when using Workers

I *think* thats the full extent of the confusion. Some of it is pedantic but may still be valuable to take a look over and see if you can clarify. I also think an example section dedicated to this use case (presign R2 url from worker, upload to it with cURL or something) could be really helpful, since it's not the first time someone has gone through the same process & issues as me here on Discord.
```

### Additional information

One can find more information in this Discord thread: https://discord.com/channels/595317990191398933/1327764577630421002/1328414674642931786

harshil1712:
### Summary

Clarifies Presigned URLs in R2. Fixes #19190 

### Documentation checklist

### DIFF
--- a/src/content/docs/r2/api/s3/presigned-urls.mdx
+++ b/src/content/docs/r2/api/s3/presigned-urls.mdx
@@ -50,16 +50,6 @@ R2 currently supports the following methods when generating a presigned URL:
 
 `POST`, which performs uploads via native HTML forms, is not currently supported.
 
-## Generate presigned URLs
-
-Generate a presigned URL by referring to the following examples:
-
-- [AWS SDK for Go](/r2/examples/aws/aws-sdk-go/#generate-presigned-urls)
-- [AWS SDK for JS v3](/r2/examples/aws/aws-sdk-js-v3/#generate-presigned-urls)
-- [AWS SDK for JS](/r2/examples/aws/aws-sdk-js/#generate-presigned-urls)
-- [AWS SDK for PHP](/r2/examples/aws/aws-sdk-php/#generate-presigned-urls)
-- [AWS CLI](/r2/examples/aws/aws-cli/#generate-presigned-urls)
-
 ## Presigned URL alternative with Workers
 
 A valid alternative design to presigned URLs is to use a Worker with a [binding](/workers/runtime-apis/bindings/) that implements your security policy.
@@ -72,14 +62,33 @@ A binding is defined in the Wrangler file of your Worker project's directory.
 
 :::
 
-A possible use case may be restricting an application to only be able to upload to a specific URL. With presigned URLs, your central signing application might look like the following JavaScript code running on Cloudflare Workers, workerd, or another platform.
+Refer to [Use R2 from Workers](/r2/api/workers/workers-api-usage/) to learn how to bind a bucket to a Worker and use the binding to interact with your bucket.
 
-If the Worker received a request for `https://example.com/uploads/dog.png`, it would respond with a presigned URL allowing a user to upload to your R2 bucket at the `/uploads/dog.png` path.
+## Generate presigned URLs
+
+Generate a presigned URL by referring to the following examples:
+
+- [AWS SDK for Go](/r2/examples/aws/aws-sdk-go/#generate-presigned-urls)
+- [AWS SDK for JS v3](/r2/examples/aws/aws-sdk-js-v3/#generate-presigned-urls)
+- [AWS SDK for JS](/r2/examples/aws/aws-sdk-js/#generate-presigned-urls)
+- [AWS SDK for PHP](/r2/examples/aws/aws-sdk-php/#generate-presigned-urls)
+- [AWS CLI](/r2/examples/aws/aws-cli/#generate-presigned-urls)
+
+### Example of generating presigned URLs
+
+A possible use case may be restricting an application to only be able to upload to a specific URL. With presigned URLs, your central signing application might look like the following JavaScript code running on Cloudflare Workers, workerd, or another platform (you might have to update the code based on the platform you are using).
+
+If the application received a request for `https://example.com/uploads/dog.png`, it would respond with a presigned URL allowing a user to upload to your R2 bucket at the `/uploads/dog.png` path.
+
+To create a presigned URL, you will need to either use a package that implements the signing algorithm or implement the signing algorithm yourself. In this example, the `aws4fetch` package is used. You also need to have an access key ID and secret access key. Refer to [R2 API tokens](/r2/api/tokens/) for more information.
 
 ```ts
 import { AwsClient } from "aws4fetch";
 
-const r2 = new AwsClient({
+// Create a new client
+// Replace with your own access key ID and secret access key
+// Make sure to store these securely and not expose them
+const client = new AwsClient({
 	accessKeyId: "",
 	secretAccessKey: "",
 });
@@ -98,6 +107,7 @@ export default {
 			return new Response("Missing a filepath", { status: 400 });
 		}
 
+		// Replace with your bucket name and account ID
 		const bucketName = "";
 		const accountId = "";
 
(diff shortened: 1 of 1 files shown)


src/content/docs/r2/api/s3/presigned-urls.mdx
```diff
@@ -72,14 +62,33 @@ A binding is defined in the Wrangler file of your Worker project's directory.
...
+- [AWS CLI](/r2/examples/aws/aws-cli/#generate-presigned-urls)
+
+### Example of generating presigned URLs
+
+A possible use case may be restricting an application to only be able to upload to a specific URL. With presigned URLs, your central signing application might look like the following JavaScript code running on Cloudflare Workers, workerd, or another platform (you might have to update the code based on the platform you are using).
 
-If the Worker received a request for `https://example.com/uploads/dog.png`, it would respond with a presigned URL allowing a user to upload to your R2 bucket at the `/uploads/dog.png` path.
+If the application received a request for `https://example.com/uploads/dog.png`, it would respond with a presigned URL allowing a user to upload to your R2 bucket at the `/uploads/dog.png` path.
+
+To create a presigned URL, you will need to either use a package that implements the signing algorithm or implement the signing algorithm yourself. In this example, the `aws4fetch` package is used. You also need to have an access key ID and secret access key. Refer to [R2 API tokens](/r2/api/tokens/) for more information.
```
Oxyjun:
```suggestion
To create a presigned URL, you will need to either use a package that implements the signing algorithm, or implement the signing algorithm yourself. In this example, the `aws4fetch` package is used. You also need to have an access key ID and a secret access key. Refer to [R2 API tokens](/r2/api/tokens/) for more information.
```

src/content/docs/r2/api/s3/presigned-urls.mdx
```diff
@@ -146,34 +160,41 @@ const signed = await r2.sign(
...
+const response = await fetch(signed.url, {
+	method: "PUT",
+	body: file,
+	headers: {
+		"If-Unmodified-Since": "Tue, 28 Sep 2021 16:00:00 GMT",
+	},
+});
+```
+
+Note that the caller has to add the same `If-Unmodified-Since` header to use the URL. The caller cannot omit the header or use a different header since the signature covers the headers. If the caller uses a different header, the presigned URL signature would not match, and they would receive a `403/SignatureDoesNotMatch`.
```
Oxyjun:
```suggestion
Note that the caller has to add the same `If-Unmodified-Since` header to use the URL. The caller cannot omit the header or use a different header, since the signature covers the headers. If the caller uses a different header, the presigned URL signature would not match, and they would receive a `403/SignatureDoesNotMatch`.
```

Oxyjun (approved):

src/content/docs/r2/api/s3/presigned-urls.mdx
```diff
@@ -72,14 +62,33 @@ A binding is defined in the Wrangler file of your Worker project's directory.
...
+
+- [AWS SDK for Go](/r2/examples/aws/aws-sdk-go/#generate-presigned-urls)
+- [AWS SDK for JS v3](/r2/examples/aws/aws-sdk-js-v3/#generate-presigned-urls)
+- [AWS SDK for JS](/r2/examples/aws/aws-sdk-js/#generate-presigned-urls)
+- [AWS SDK for PHP](/r2/examples/aws/aws-sdk-php/#generate-presigned-urls)
+- [AWS CLI](/r2/examples/aws/aws-cli/#generate-presigned-urls)
+
+### Example of generating presigned URLs
+
+A possible use case may be restricting an application to only be able to upload to a specific URL. With presigned URLs, your central signing application might look like the following JavaScript code running on Cloudflare Workers, workerd, or another platform (you might have to update the code based on the platform you are using).
```
Oxyjun:
```suggestion
A possible use case may be restricting an application to only be able to upload to a specific URL. With presigned URLs, your central signing application might look like the following JavaScript code running on Cloudflare Workers, `workerd`, or another platform (you might have to update the code based on the platform you are using).
```

src/content/docs/r2/api/s3/presigned-urls.mdx
```diff
@@ -128,12 +138,16 @@ export default {
 } satisfies ExportedHandler;
 ```
 
-Notice the total absence of any configuration or token secrets present in the Worker code. Instead, in your [Wrangler configuration file](/workers/wrangler/configuration/), you would create a [binding](/r2/api/workers/workers-api-usage/#3-bind-your-bucket-to-a-worker) to whatever bucket represents the bucket you will upload to. Additionally, authorization is handled in-line with the upload which can reduce latency.
+## Differences between presigned URLs and R2 binding
+
+When using the R2 binding, you will not need any token secrets in your Worker code. Instead, in your [Wrangler configuration file](/workers/wrangler/configuration/), you will create a [binding](/r2/api/workers/workers-api-usage/#3-bind-your-bucket-to-a-worker) to your R2 bucket. Additionally, authorization is handled in-line which can reduce latency.
```
Oxyjun:
```suggestion
- When using an R2 binding, you will not need any token secrets in your Worker code. Instead, in your [Wrangler configuration file](/workers/wrangler/configuration/), you will create a [binding](/r2/api/workers/workers-api-usage/#3-bind-your-bucket-to-a-worker) to your R2 bucket. Additionally, authorization is handled in-line, which can reduce latency.
```

src/content/docs/r2/api/s3/presigned-urls.mdx
```diff
@@ -128,12 +138,16 @@ export default {
...
 
-In some cases, Workers lets you implement certain functionality more easily. For example, if you wanted to offer a write-once guarantee so that users can only upload to a path once, with pre-signed URLs, you would need to sign specific headers and require the sender to send them. You can modify the previous Worker to sign additional headers:
+However, when using presigned URLs, you will need to create and use the token secrets in your Worker code.
```
Oxyjun:
```suggestion
- When using presigned URLs, you will need to create and use the token secrets in your Worker code.
```

src/content/docs/r2/api/s3/presigned-urls.mdx
```diff
@@ -128,12 +138,16 @@ export default {
...
+
+In some cases, Workers lets you implement certain functionality more easily. For example, if you wanted to offer a write-once guarantee so that users can only upload to a path once, with pre-signed URLs, you would need to sign specific headers and require the sender to send the same headers. You can modify the previous example to sign additional headers:
```
Oxyjun:
```suggestion
In some cases, R2 bindings let you implement certain functionality more easily. For example, if you wanted to offer a write-once guarantee so that users can only upload to a path once, with pre-signed URLs:

 - With R2 binding: You only need to pass the parameter once.
 - With presigned URLs: You need to sign specific headers and require the sender to send the same headers.
 
 You can modify the previous example to sign additional headers:
```

aninibread:
that looks good to me, thank you for the suggestions!

PR #22005, opened 2025-04-28
update C3 instructions for Pages projects

petebacondarwin:
### Summary

It is now required to pass `--platform=pages` to create a Pages project.

### Documentation checklist

### DIFF
--- a/src/content/partials/pages/c3-run-command-no-directory.mdx
+++ b/src/content/partials/pages/c3-run-command-no-directory.mdx
@@ -1,0 +1,8 @@
+---
+{}
+---
+
+import { PackageManagers } from "~/components";
+
+<PackageManagers type="create" pkg="cloudflare@latest" args="--platform=pages"/>
+
--- a/src/content/docs/pages/get-started/c3.mdx
+++ b/src/content/docs/pages/get-started/c3.mdx
@@ -25,9 +25,13 @@ Cloudflare provides a CLI command for creating new Workers and Pages projects â€
 
 Open a terminal window and run:
 
-<Render file="c3-run-command-no-directory" product="workers" />
+<Render file="c3-run-command-no-directory" product="pages" />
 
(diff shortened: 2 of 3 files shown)


lrapoport-cf:
@petebacondarwin thanks for putting this up! can you add the `--platform` flag into the [CLI arguments](https://developers.cloudflare.com/pages/get-started/c3/#cli-arguments) section as well?

petebacondarwin:
> @petebacondarwin thanks for putting this up! can you add the `--platform` flag into the [CLI arguments](https://developers.cloudflare.com/pages/get-started/c3/#cli-arguments) section as well?

Is that actually the right thing to do here?

The only options for `--platform` are `pages` and `workers`. But this doc is in the Pages section of the docs and is specifically about running C3 to generate a Pages project; and in the snippets I have hard coded the `--platform=pages` option.

lrapoport-cf:
> > @petebacondarwin thanks for putting this up! can you add the `--platform` flag into the [CLI arguments](https://developers.cloudflare.com/pages/get-started/c3/#cli-arguments) section as well?
> 
> Is that actually the right thing to do here?
> 
> The only options for `--platform` are `pages` and `workers`. But this doc is in the Pages section of the docs and is specifically about running C3 to generate a Pages project; and in the snippets I have hard coded the `--platform=pages` option.

hmm, good question. the CLI arguments being listed in this section conflates pages-specific concerns with more general c3 documentation, especially given the wording:
image

we might want to consider reorganizing some of the c3 documentation, but that's not within the scope of this PR, so we can table my comment for now and revisit later if makes sense

src/content/partials/pages/c3-run-command-no-directory.mdx
GregBrimble, line 1: Should this just be inlined? Why is it a partial?

GregBrimble (approved):

irvinebroque (approved):

src/content/partials/pages/c3-run-command-no-directory.mdx
petebacondarwin: No idea. I guess the original plan was that it would be reused for Pages and Workers?

PR #22028, opened 2025-04-28
[Docs Site] Fix head tags after latest Starlight update

KianNH:
### Summary

Fix head tags after latest Starlight update


src/components/overrides/Head.astro
```diff
@@ -161,6 +160,8 @@ metaTags.map((attrs) => {
 		attrs,
 	});
 });
+
+console.log(head);
```
kodster28:
```suggestion
```

kodster28 (approved):

PR #22056, opened 2025-04-29
[BRAPI]Rest API guide

daisyfaithauma:
### Documentation checklist

### DIFF
--- a/src/content/docs/browser-rendering/how-to/markdown-extraction.mdx
+++ b/src/content/docs/browser-rendering/how-to/markdown-extraction.mdx
@@ -1,0 +1,178 @@
+---
+title: Extracting blog post content as markdown using the markdown endpoint
+sidebar:
+  order: 4
+---
+
+This guide shows you how to capture the complete JSON output from Cloudflare's [`/markdown` API endpoint](/browser-rendering/rest-api/markdown-endpoint/).
+
+We are extracting the content of a blog post from the Cloudflare Blog: [Introducing AutoRAG on Cloudflare](https://blog.cloudflare.com/introducing-autorag-on-cloudflare/)
+
+## Prerequisites
+
+1. Cloudflare Account and API Token.
+
+   - [Create a token](/fundamentals/api/get-started/create-token/) with **Browser Rendering: Edit** permissions.
+   - You can do this under **My Profile → API Tokens → Create Token** on your [Cloudflare dashboard](https://dash.cloudflare.com/).
+   - Note your **Account ID** (from the dashboard homepage) and **API Token**.
+
+2. Command-line tools installed.
+
+   - cURL: a command-line tool for sending HTTP requests.
+     - macOS/Linux: usually preinstalled.
+     - Windows: available via WSL, Git Bash, or native Windows builds.
+
+## 1: Configure your environment variables
+
+Save your sensitive information into environment variables to avoid hardcoding credentials.
+
+```bash
+export CF_ACCOUNT_ID="your-cloudflare-account-id"
+export CF_API_TOKEN="your-api-token-with-edit-permissions"
+```
+
+## 2: Make the API Request and save the raw JSON
+
+Run this command to fetch the markdown representation of the AutoRAG blog post and store it into a local JSON file:
+
+```bash
+curl -s -X POST \
+  "https://api.cloudflare.com/client/v4/accounts/${CF_ACCOUNT_ID}/browser-rendering/markdown" \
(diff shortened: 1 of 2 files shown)


src/content/docs/browser-rendering/how-to/markdown-extraction.mdx
```diff
@@ -0,0 +1,178 @@
...
+## 1: Configure your environment variables
+
+Save your sensitive information into environment variables to avoid hardcoding credentials.
+
+```bash
+export CF_ACCOUNT_ID="your-cloudflare-account-id"
+export CF_API_TOKEN="your-api-token-with-edit-permissions"
+```
+
+## 2: Make the API Request and save the raw JSON
```
kodster28, line 34: Question about this tutorial..... if we're already running a Python script, couldn't we just have steps 2-4 included in the Python script as well?

Seems like it would simplify things a lot.

daisyfaithauma: I am not sure how to do that. Maybe you could guide me in the right direction?

kodster28: > I am not sure how to do that. Maybe you could guide me in the right direction?

Python has a `requests` library, which could be used to make the cURL request.

kodster28: I suppose the broader question is also.... _why_ Python here?

If we're using the REST API, all of the other examples are using the TypeScript SDK... so why would we change languages from what we provide in the rest of the docs?

omarmosid: I am a little confused too..
if our end goal is to generate an md file for a blog post, we could perhaps do it in a worker? 
And if that is the case, it would make more sense to use BR bindings (instead of the REST API). We could all the BR work in the worker and finally return the md file as the worker response.

If it's not possible to do it in a worker, instead of python, we should do it as a node script with the typescript SDK or directly calling the REST API through fetch. Lemme know if you need any help figuring this out

daisyfaithauma: @omarmosid I specifically did not do it with Workers because we do not have any guide of using our REST API endpoints. The goal is to show how to use our REST API endpoints.
Please assist in doing it through the fetch API.

omarmosid: Something like this? (using node)

```js
// saveMarkdown.js
const fs = require('fs');

async function downloadMarkdown() {
  const accountId = '<accountId>';
  const apiToken = '<apiToken>';
  const blogUrl = 'https://example.com';
  const endpoint = `https://api.cloudflare.com/client/v4/accounts/${accountId}/browser-rendering/markdown`;

  try {
    const response = await fetch(endpoint, {
      method: 'POST',
      headers: {
        'Authorization': `Bearer ${apiToken}`
      },
      body: JSON.stringify({ url: blogUrl })
    });

    if (!response.ok) {
      throw new Error(`Failed to fetch markdown: ${response.status} ${response.statusText}`);
    }

    const data = await response.json();
    const markdown = data.result;

    const filename = 'output.md';
    fs.writeFileSync(filename, markdown);
    console.log(`Markdown saved to ${filename}`);
  } catch (error) {
    console.error('Error:', error.message);
  }
}

downloadMarkdown();
```

kodster28: > @omarmosid I specifically did not do it with Workers because we do not have any guide of using our REST API endpoints. The goal is to show how to use our REST API endpoints. Please assist in doing it through the fetch API.

If it's something that fundamentally doesn't make sense to do via the REST API, shouldn't we be looking for another use case?

We don't want to promote an inefficient approach to a problem, even if it's illustrative.

meddulla:
Any reason why we cant have a `python` tab in all the REST examples? Eg. here https://developers.cloudflare.com/browser-rendering/rest-api/content-endpoint/, in the same place we have `curl` and `typescript sdk` . And if using `pyhon`, we can either use the `requests` or even better our [python sdk](https://github.com/cloudflare/cloudflare-python/blob/main/src/cloudflare/resources/browser_rendering/content.py). The REST api can of course be used from any language, so no issue with having python examples

kodster28:
Based on all this conversation, I think we'll want to circle up and take a bit of a different approach to this here.

Going to close this out and -- once we settle on something -- we can open a new PR around this.

PR #22100, opened 2025-04-30
Add warning for Service Worker deprecation

harshil1712:
### Summary

Updates a warning for Service Worker examples.

### Documentation checklist

### DIFF
--- a/src/content/docs/learning-paths/mtls/mtls-workers/index.mdx
+++ b/src/content/docs/learning-paths/mtls/mtls-workers/index.mdx
@@ -5,6 +5,8 @@ sidebar:
   order: 4
 ---
 
+import { Tabs, TabItem } from "~/components";
+
 :::note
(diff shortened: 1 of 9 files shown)


kristianfreeman (approved):
Warning might be a bit redundant here but no biggie: https://service-worker-callout.preview.developers.cloudflare.com/workers/reference/migrate-to-module-workers/

Is it worth breaking this out into a re-usable template, in case we have to update this later? I'm good w/ it merging as-is but worth asking @kodster28 and @KianNH maybe

kodster28 (approved):

irvinebroque (commented):
Why do we still have service workers examples in our docs at all?

harshil1712:
@irvinebroque, we have deprecated Service Workers, but we still support them. We might have users still using them, and so we need examples to support them. We should remove the examples once we completely stop the supporting Service Workers.

PR #22107, opened 2025-04-30
[web-analytics]Update index.mdx

luishcardoso:
Hello team, 

### Summary

Referring to the Web analytics overview docs here.

Thanks!


kodster28:
This is implicit from the current context of the page + we don't want to send someone out to another section of the docs.

PR #22192, opened 2025-05-03
[Workers] Fix migrate-from-pages redirect source path in _redirect

NEKOYASAN:
### Summary

Fixed an incorrect redirect setting for the `Migrate from Pages ` page whose URL was changed in #22064.

### Documentation checklist


GregBrimble:
Thank you, but this was intentional — before the other migration guides, this lived at the top of the static assets section, so I'm just trying to capture references of when it lived there.

NEKOYASAN:
I thought it was a mistake because I was using this URL when `migrate-from-pages` was under `migrations` and not directly under `static-assets`.
If the current one is also correct, then it seems to me that a redirect should be set up, since the URL that is used as the redirect source for this PR is also widely known.

GregBrimble:
It was moved under `migrations` four days ago, and then moved to `migration-guides` three days ago. I don't think many external places picked up the `migrations` location during that 24hrs, but if there's anywhere that does have a broken link because of this, we absolutely can include a redirect!

PR #22284, opened 2025-05-07
Simplify websocket destructuring example

sambostock:
### Summary

Instead of extracting values and destructuring them positionally, we can destructure by key, which is both shorter and clearer about what's going on.

### Documentation checklist


kodster28:
Howdy @sambostock, asked around internally and we're fine with the example as is.

Thanks for the suggestion though!

PR #22335, opened 2025-05-09
Vectorize: Make index naming requirement explicit

s-h-a-d-o-w:
### Summary

Names can be arbitrary strings more often than not (I'm not referring to cloudflare here but programming in general), so it doesn't follow from examples that other casing can't be used.

### DIFF
--- a/src/content/docs/vectorize/best-practices/create-indexes.mdx
+++ b/src/content/docs/vectorize/best-practices/create-indexes.mdx
@@ -11,7 +11,7 @@ Indexes are the "atom" of Vectorize. Vectors are inserted into an index and enab
 
 Creating an index requires three inputs:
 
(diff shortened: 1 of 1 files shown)


sejoker (approved):

src/content/docs/vectorize/best-practices/create-indexes.mdx
```diff
@@ -11,7 +11,7 @@ Indexes are the "atom" of Vectorize. Vectors are inserted into an index and enab
 
 Creating an index requires three inputs:
 
-- A name, for example `prod-search-index` or `recommendations-idx-dev`.
+- A kebab-cased name, for example `prod-search-index` or `recommendations-idx-dev`.
```
kodster28:
```suggestion
- A kebab-cased name, such as `prod-search-index` or `recommendations-idx-dev`.
```

PR #22375, opened 2025-05-12
Document WRANGLER_LOG_PATH system environment variable

benjreinhart:
I see this option in the Changelog and in the source, but not in the docs.

### Documentation checklist

I took some of the same language from the [Changelog entry](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/CHANGELOG.md#minor-changes-112).

### DIFF
--- a/src/content/docs/workers/wrangler/system-environment-variables.mdx
+++ b/src/content/docs/workers/wrangler/system-environment-variables.mdx
@@ -61,7 +61,7 @@ Wrangler supports the following environment variables:
 
 * `WRANGLER_LOG_PATH` <Type text="string" /> <MetaInfo text="optional" />
 
-  * A file or directory path where Wrangler will write its logs. When specifying a path to a file, the filename must end with `.log`. When specifying a directory or using the default location, a filename with a timestamp is used.
+  * A file or directory path where Wrangler will write debug logs. When specifying a path to a file, the filename must end with `.log`. When specifying a directory or using the default location, a filename with a timestamp is used.
 
 * `FORCE_COLOR` <Type text="string" /> <MetaInfo text="optional" />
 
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/wrangler/system-environment-variables.mdx
```diff
@@ -59,6 +59,10 @@ Wrangler supports the following environment variables:
 
   * Options for Logging levels are `"none"`, `"error"`, `"warn"`, `"info"`, `"log"` and `"debug"`. Levels are case-insensitive and default to `"log"`. If an invalid level is specified, Wrangler will fallback to the default. Logs can include requests to Cloudflare's API, any usage data being collected, and more verbose error logs.
 
+* `WRANGLER_LOG_PATH` <Type text="string" /> <MetaInfo text="optional" />
+
+  * A file or directory path where Wrangler will write its logs. When specifying a path to a file, the filename must end with `.log`. When specifying a directory or using the default location, a filename with a timestamp is used.
```
petebacondarwin: Nit: this not only writes log but it actually writes all the "debug" logs:

```suggestion
  * A file or directory path where Wrangler will write debug logs. When specifying a path to a file, the filename must end with `.log`. When specifying a directory or using the default location, a filename with a timestamp is used.
```

petebacondarwin: I found the explanation about the extension a bit confusing. Perhaps:

```suggestion
  * A file or directory path where Wrangler will write its logs. If the path ends in `.log` then this is treated as the file to write to; otherwise it is treated as a directory that will contain timestamped files.
```

petebacondarwin (approved):
Couple of minor suggestions

petebacondarwin (approved):

kodster28 (approved):

nichochar:
Wow congrats on the holobyte ben, huge

PR #22378, opened 2025-05-13
[R2] Adding Sip n Slurp migration strategy

Oxyjun:
### Summary

Adding a migration strategy using Sippy first, then Super Slurper, to effectively migrate all objects into an R2 bucket without any downtime.

### Documentation checklist

### DIFF
--- a/src/content/docs/r2/data-migration/index.mdx
+++ b/src/content/docs/r2/data-migration/index.mdx
@@ -57,7 +57,20 @@ Quickly and easily migrate data from other cloud providers to R2. Explore each o
 
 ## Migration strategies
 
-You can use a combination of Super Slurper and Sippy to effectively migrate all objects without any downtime.
+You can use a combination of Super Slurper and Sippy to effectively migrate all objects with minimal downtime.
 
-1. Activate Sippy to gradually copy objects from the previous bucket to an R2 bucket when they are requested (on-demand).
+### When the source bucket is actively being read from / written to
+
+1. Enable Sippy and start using the R2 bucket in your application.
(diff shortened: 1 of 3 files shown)


src/content/docs/r2/data-migration/index.mdx
```diff
@@ -54,3 +54,10 @@ Quickly and easily migrate data from other cloud providers to R2. Explore each o
 		</tr>
 	</tbody>
 </table>
+
+## Migration strategies
```
sdnts: Like we were discussing in our internal thread, I think it makes sense to document at least two strategies here, one of them might be more appropriate depending on how users use S3 / GCS etc:

1. For active buckets that are being read from and written to: Enable Sippy and start using the R2 bucket in your application -- objects will be copied over on demand as they are accessed, and new uploads will only go to R2. Then once the most-accessed objects are copied over, they can run Slurper (with "skip existing" enabled) so the remaining tail of objects are copied over.

2. Cases when the source bucket is not being read from as often, and is only receiving new uploads: Use Slurper to copy all objects (note that Slurper in this case might skip some objects if they are added after it lists them), then enable Sippy on your R2 bucket, and start using R2. New uploads will now go to the R2 bucket, and objects that were uploaded while Slurper was copying objects will be transferred on-demand when they're requested.

Oxyjun: Thanks! I've included the two strategies you've mentioned.

aninibread:
The content looks great! Let's have migration strategies as a separate page under "Data Migrations" to keep the Overview page a bit cleaner

marciocloudflare (approved):

PR #22383, opened 2025-05-13
[R2]Update object-lifecycles.mdx

luishcardoso:
The correct phrasing is "24-hour period" when used as a modifier (like an adjective) before a noun. For example, "a 24-hour period".

### Summary
The correct phrasing is "24-hour period" when used as a modifier (like an adjective) before a noun. For example, "a 24-hour period". 
Thanks!


Oxyjun (approved):

Oxyjun:
Thanks for the PR!

PR #22401, opened 2025-05-13
[WARP] Release notes

kodster28:
To validate:

Go to the [Stable releases page](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/download-warp/) in the preview and check both `Linux` and `macOS` to make sure the latest releases match what we're adding here.

### DIFF
--- a/src/content/warp-releases/linux/ga/2025.4.929.0.yaml
+++ b/src/content/warp-releases/linux/ga/2025.4.929.0.yaml
@@ -1,0 +1,45 @@
+releaseNotes: >-
+  This release contains two significant changes all customers should be aware
+  of:
+
+  1. All DNS traffic now flows inside the WARP tunnel. Customers are no longer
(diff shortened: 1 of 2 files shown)


kokolocomotion1 (commented):
There's extra spacing between sentences in Linux notes but I'm not sure if that will translate. otherwise this is all good, thank you!

src/content/warp-releases/linux/ga/2025.4.929.0.yaml
```diff
@@ -0,0 +1,45 @@
+releaseNotes: >-
```
pedrosousa, line 1: Nit: Ideally we'd always use either `>-` or `|-` consistently. My personal preference goes to `|-` because it would require less line breaks in the YAML file. More info at https://stackoverflow.com/a/21699210.

pedrosousa (approved):

PR #22461, opened 2025-05-15
Update index.yaml

simon-says:
### Summary

Adding http.host description that clarifies (lack of) interaction with Origin Rules

### DIFF
--- a/src/content/fields/index.yaml
+++ b/src/content/fields/index.yaml
@@ -12,6 +12,8 @@ entries:
     categories: [Request, URI]
     keywords: [request, uri, url, domain, client, visitor]
     summary: The hostname used in the full request URI.
+    description: |-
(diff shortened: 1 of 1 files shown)


src/content/fields/index.yaml
```diff
@@ -12,6 +12,8 @@ entries:
     categories: [Request, URI]
     keywords: [request, uri, url, domain, client, visitor]
     summary: The hostname used in the full request URI.
+    description: |-
+      `http.host` contains the host header from the original request. **If** you have **origin rules** that change the hostname, they will not be reflected in the `http.host` value
```
pedrosousa:
```suggestion
      The `http.host` field contains the `Host` header from the original request.
      
      If you have configured [Origin Rules](/rules/origin-rules/) that change the hostname, they will not be reflected in the `http.host` field value.
```

pedrosousa (approved):

PR #22494, opened 2025-05-16
[DO] Recommending restricting DurableObjectNamespace to jurisdiction, rather than DurableObjectID

Oxyjun:
### Documentation checklist


src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,31 +19,40 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
-Methods on a [`DurableObjectNamespace`](/durable-objects/api/namespace/) that take a [`DurableObjectId`](/durable-objects/api/id) as a parameter will throw an exception if the parameter is associated with a different jurisdiction. However, these methods will not throw an exception if the [`DurableObjectNamespace`](/durable-objects/api/namespace/) is not associated with a jurisdiction. The common case is that any valid [`DurableObjectId`](/durable-objects/api/id) can be used in the top-level namespace's methods.
+However, these methods will not throw an exception if the [`DurableObjectNamespace`](/durable-objects/api/namespace/) is not associated with a jurisdiction. The common case is that any valid [`DurableObjectId`](/durable-objects/api/id) can be used in the top-level namespace's methods.
 
 ```js
 const euSubnamespace = env.MY_DURABLE_OBJECT.jurisdiction("eu");
 const euId = euSubnamespace.idFromName(name);
 const stub = env.MY_DURABLE_OBJECT.get(euId);
 ```
 
+:::caution[Do not use `newUniqueId(jurisdiction)`]
```
vy-ton: let's frame this as what is the recommendation first

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,31 +19,40 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
 
 ```js
 const euId1 = env.MY_DURABLE_OBJECT.idFromName("my-name");
 const euId2 = env.MY_DURABLE_OBJECT.jurisdiction("eu").idFromName("my-name");
 console.assert(!euId1.equal(euId2), "This should always be true");
 ```
-
-Methods on a [`DurableObjectNamespace`](/durable-objects/api/namespace/) that take a [`DurableObjectId`](/durable-objects/api/id) as a parameter will throw an exception if the parameter is associated with a different jurisdiction. However, these methods will not throw an exception if the [`DurableObjectNamespace`](/durable-objects/api/namespace/) is not associated with a jurisdiction. The common case is that any valid [`DurableObjectId`](/durable-objects/api/id) can be used in the top-level namespace's methods.
+However, these methods will not throw an exception if the [`DurableObjectNamespace`](/durable-objects/api/namespace/) is not associated with a jurisdiction. The common case is that any valid [`DurableObjectId`](/durable-objects/api/id) can be used in the top-level namespace's methods.
```
vy-ton: This and paragraph above are written somewhat confusingly, can we try to clear up.
-  "As a consequence, it is possible to have the same name represent different IDs in different jurisdictions." seems like the key takeaway so start with it.
- "Methods on a [`DurableObjectNamespace`](/durable-objects/api/namespace/) that take a [`DurableObjectId`](/durable-objects/api/id) as a parameter will throw an exception if the parameter is associated with a different jurisdiction." is written very-technical, API first which is confusing

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,24 +19,41 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
 
 ```js
 const euSubnamespace = env.MY_DURABLE_OBJECT.jurisdiction("eu");
 const euId = euSubnamespace.newUniqueId();
-// or
+```
+
+This approach works with all of the [ID methods](/durable-objects/api/id/).
+
+:::caution[Do not use `newUniqueId(jurisdiction)`]
```
vy-ton: Callouts should be frames as the recommended approach so say `Use DurableObjectNamespace.jurisdiction`

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,24 +19,41 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
+
+When specifying a jurisdiction, Cloudflare recommends you first create a namespace restricted to a jurisdiction, as explained above.
```
vy-ton: include the example inline here too

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,24 +19,41 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
+
+Note that it is possible to specify a jurisdiction by creating an individual [`DurableObjectId`](/durable-objects/api/id) restricted to a jurisdiction.
+
+**However, Cloudflare does not recommend this approach.**
+
+This approach is only documented for completeness.
```
vy-ton: aim for callout to be shorter, so end it here.

The bottom 3 bullets dont make sense in the note

image

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,24 +19,41 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
-Durable Objects can be restricted to a specific jurisdiction either by creating a [`DurableObjectNamespace`](/durable-objects/api/namespace/) restricted to a jurisdiction, or by creating an individual [`DurableObjectId`](/durable-objects/api/id) restricted to a jurisdiction:
+Durable Objects can be restricted to a specific jurisdiction by creating a [`DurableObjectNamespace`](/durable-objects/api/namespace/) restricted to a jurisdiction:
 
 ```js
 const euSubnamespace = env.MY_DURABLE_OBJECT.jurisdiction("eu");
 const euId = euSubnamespace.newUniqueId();
-// or
+```
+
+This approach works with all of the [ID methods](/durable-objects/api/id/).
```
vy-ton: combine this with the paragraph starting with `Durable Objects can be...`

also make it easier for the user to understand what this means. Ex: `All Durable Object ID methods are valid on ids within a namespace restricted to a jurisdiction`

src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -19,24 +19,41 @@ A [`DurableObjectId`](/durable-objects/api/id) will be logged outside of the spe
...
 
 ```js
 const euId1 = env.MY_DURABLE_OBJECT.idFromName("my-name");
 const euId2 = env.MY_DURABLE_OBJECT.jurisdiction("eu").idFromName("my-name");
 console.assert(!euId1.equal(euId2), "This should always be true");
 ```
+:::
 
-Methods on a [`DurableObjectNamespace`](/durable-objects/api/namespace/) that take a [`DurableObjectId`](/durable-objects/api/id) as a parameter will throw an exception if the parameter is associated with a different jurisdiction. However, these methods will not throw an exception if the [`DurableObjectNamespace`](/durable-objects/api/namespace/) is not associated with a jurisdiction. The common case is that any valid [`DurableObjectId`](/durable-objects/api/id) can be used in the top-level namespace's methods.
+The common case is that any valid [`DurableObjectId`](/durable-objects/api/id) can be used in the top-level namespace's methods.
```
vy-ton: I would delete, this sentence isnt really helpful

src/content/docs/durable-objects/reference/data-location.mdx
Oxyjun: This will cause duplicate code example to appear in quick succession, can I check you're okay with that?

image

src/content/docs/durable-objects/reference/data-location.mdx
Oxyjun: To double check, is the codeblock below that also relevant, or should we delete that too?

src/content/docs/durable-objects/reference/data-location.mdx
vy-ton: Let's remove both code examples in the callout. You can inline the code for `env.MY_DURABLE_OBJECT.newUniqueId({ jurisdiction: "eu" });` in the sentence (shouldn't make code copy snippet if we don't recommend)

vy-ton:
image

The code at the bottom applies to the first bullet

src/content/docs/durable-objects/reference/data-location.mdx
vy-ton: Let's put `All [Durable Object ID methods](https://jun-do-jurisdiction-recommendation.preview.developers.cloudflare.com/durable-objects/api/id/) are valid on IDs within a namespace restricted to a jurisdiction.` in the bulleted list, then this code block can be under that bullet

vy-ton (approved):

deadlypants1973 (approved):

PR #22512, opened 2025-05-16
Access analytics changelog

cloudflare-noelle:
### Summary

Added a changelong entry for new access analytics in . This includes the new folder: src/assets/images/changelog/cf1-analytics/ for images


src/assets/images/changelog/cf1-analytics/test
```diff
@@ -0,0 +1 @@
+
```
kodster28: Delete?

src/content/changelog/cf1-analytics/access-analytics-v2
```diff
@@ -0,0 +1,20 @@
+---
+title: New Access Analytics in the Cloudflare One Dashboard
```
kennyj42: @kodster28 can we tag something with multiple products? It'd be great if this could also be flagged as an Access change

src/assets/images/changelog/cf1-analytics/test
cloudflare-noelle: yes, if possible. I couldnt figure out how to.

src/content/changelog/cf1-analytics/access-analytics-v2
kodster28: Let me hop in and help here.

kodster28: We already support this actually. 

Whatever the current folder is, we automatically tag with the `product`.

To add a new product tag, you add a YAML array to the frontmatter.

```
products:
  - access
```

kodster28 (approved):

PR #22539, opened 2025-05-19
[Workers] Update tcp-sockets.mdx

luishcardoso:
Hello team, 

### Summary

Let's link it to the _try...catch_ docs for more info and examples.

Thanks!

### DIFF
--- a/src/content/docs/workers/runtime-apis/tcp-sockets.mdx
+++ b/src/content/docs/workers/runtime-apis/tcp-sockets.mdx
@@ -143,7 +143,7 @@ const secureSocket = socket.startTls();
 
 ## Handle errors
 
-To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside [`try/catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) statement blocks. The following example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a `500` response:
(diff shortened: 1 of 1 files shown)


kodster28 (approved):

src/content/docs/workers/runtime-apis/tcp-sockets.mdx
```diff
@@ -143,7 +143,7 @@ const secureSocket = socket.startTls();
 
 ## Handle errors
 
-To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside `try..catch` blocks. The following example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a `500` response:
+To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside [`try/catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) statement blocks. The following example opens a connection to Google.com, initiates a HTTP request, and returns the response. If any of this fails and throws an exception, it returns a `500` response:
```
kodster28:
```suggestion
To handle errors when creating a new TCP socket, reading from a socket, or writing to a socket, wrap these calls inside [`try/catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) statement blocks. The following example opens a connection to Google.com, initiates a HTTP request, and returns the response. If this fails and throws an exception, it returns a `500` response:
```

PR #22554, opened 2025-05-19
Documents connect being shut off when outbound worker is used

mikenomitch:
### Summary

When an outbound workers is on, the connect() API no longer works. This adds a note to the documentation explaining this.

In the long-run, we may add an outboundConnect-like API to allow the platform user to handle outbound TCP or allow them to turn it on/off with a toggle. If we do that, then this should be reverted or changed.

### Screenshot

Screenshot 2025-05-19 at 9 49 20 AM

### Documentation checklist


angelampcosta (approved):

angelampcosta:
@mikenomitch can I merge the PR?

irvinebroque (approved):
does this mean you can't use hyperdrive when outbound workers are enabled?

glad we are documenting but this feels like just a bug, since afaik outbound workers were only ever intended to intercept `fetch()` in the first place.

maybe another example of where pipelines and complexity that falls out from that hurts us

@dinasaur404 fyi for context re: bringing outbound workers to workers

PR #22634, opened 2025-05-23
Update hello-world.mdx

sebst:
### Summary

`module` is not available on `WorkersScript`. Updated to Service Worker Syntax

### Documentation checklist


praneetloke:
@sebst I believe you just need to set the `mainModule` property to `worker.js` since that's the default module name. That's the only change you'll likely need in order to make the example work with ES6 modules.

https://github.com/pulumi/pulumi-cloudflare/issues/969#issuecomment-3079143875

PR #22681, opened 2025-05-27
[Workers]Update cache.mdx

luishcardoso:
Hello team,

### Summary

For a better understanding of our implementation, let's point it to our example on using the Cache API.

Thanks!


kodster28:
Don't think that's the right place to be linking (covered above in the page).

Also specific resources are already referenced in https://developers.cloudflare.com/workers/runtime-apis/cache/#related-resources

PR #22719, opened 2025-05-28
[Fundamentals & Rules] Moved Trace Request & added redirects

dcpena:
Moving Trace Request out of Fundamentals and into Rules for PCX-17591.


src/content/release-notes/trace.yaml
```diff
@@ -1,9 +1,9 @@
 ---
-link: "/fundamentals/trace-request/changelog/"
+link: "/rules/trace-request/changelog/"
 productName: Trace
-productLink: "/fundamentals/trace-request/"
+productLink: "/rules/trace-request/"
 productArea: Core platform
-productAreaLink: /fundamentals/reference/changelog/platform/
+productAreaLink: /rules/reference/changelog/platform/
```
pedrosousa:
```suggestion
productAreaLink: /fundamentals/reference/changelog/platform/
```
I'd keep the link as is, since according to the [`rules.yml` file](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/products/rules.yml#L7) the main area for Rules is "Core platform" (this link actually points to a page hidden from the navbar, but the page should exist).

pedrosousa (approved):

kodster28 (approved):

PR #22723, opened 2025-05-28
[cache] Update purge-everything.mdx

luishcardoso:
Hello team, 

### Summary

Can we warn customers here about potential impacts on their origin server and their application performance due to purge everything? 

Thanks!


angelampcosta:
He think this note is not necessary, since the page already states "Purging everything instantly clears all resources from your CDN cache in all Cloudflare data centers. Each new request for a purged resource returns to your origin server to validate the resource. If Cloudflare cannot validate the resource, Cloudflare fetches the latest version from the origin server and replaces the cached version."

PR #22728, opened 2025-05-28
Fix Durable Objects TTL example where long TTLs lead to infinite alar…

joshthoward:
…m invocations

### Documentation checklist

### DIFF
--- a/src/content/docs/durable-objects/examples/durable-object-ttl.mdx
+++ b/src/content/docs/durable-objects/examples/durable-object-ttl.mdx
@@ -12,6 +12,12 @@ import { TabItem, Tabs, GlossaryTooltip, WranglerConfig } from "~/components";
 
 A common feature request for Durable Objects is a Time To Live (TTL) for Durable Object instances. Durable Objects give developers the tools to implement a custom TTL in only a few lines of code. This example demonstrates how to implement a TTL making use of <GlossaryTooltip term="alarm">`alarms`</GlossaryTooltip>. While this TTL will be extended upon every new request to the Durable Object, this can be customized based on a particular use case.
 
+:::caution[Not available in local development]
+
+The TTL is extended upon every new fetch request to the Durable Object. It might be tempting to instead extend the TTL in the constructor of the Durable Object. This is not advised because the Durable Object's constructor will be called before invoking the alarm handler if the alarm wakes the Durable Object up from hibernation. This approach will naively result in the constructor continually extending the TTL without running the alarm handler.
+
+:::
+
 <Tabs> <TabItem label="JavaScript" icon="seti:javascript">
 
 ```js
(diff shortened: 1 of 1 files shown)


src/content/docs/durable-objects/examples/durable-object-ttl.mdx
```diff
@@ -12,28 +12,30 @@ import { TabItem, Tabs, GlossaryTooltip, WranglerConfig } from "~/components";
 
 A common feature request for Durable Objects is a Time To Live (TTL) for Durable Object instances. Durable Objects give developers the tools to implement a custom TTL in only a few lines of code. This example demonstrates how to implement a TTL making use of <GlossaryTooltip term="alarm">`alarms`</GlossaryTooltip>. While this TTL will be extended upon every new request to the Durable Object, this can be customized based on a particular use case.
 
+:::caution[Not available in local development]
```
vy-ton, line 15: Why does the title say `not available in local dev`?

vy-ton (approved):

vy-ton:
@Oxyjun as a separate task, can we also enhance the alarm docs with:

- [ ] warns against calling `.setAlarm()` in the constructor, at least without checking that the alarm is not already set
- [ ] clarify recommendation is because if DO is inactive then the constructor is invoked before the `alarm` handler. So if the constructor code calls `setAlarm`, then it could interfere with the already configured next alarm

src/content/docs/durable-objects/examples/durable-object-ttl.mdx
Frederik-Baetens: I think alarm support in local dev is limited/non-existent.

Oxyjun:
> @Oxyjun as a separate task, can we also enhance the alarm docs with:
> 
> * [ ]  warns against calling `.setAlarm()` in the constructor, at least without checking that the alarm is not already set
> * [ ]  clarify recommendation is because if DO is inactive then the constructor is invoked before the `alarm` handler. So if the constructor code calls `setAlarm`, then it could interfere with the already configured next alarm

Tracking this here: https://github.com/cloudflare/cloudflare-docs/pull/22746

src/content/docs/durable-objects/examples/durable-object-ttl.mdx
```diff
@@ -12,28 +12,30 @@ import { TabItem, Tabs, GlossaryTooltip, WranglerConfig } from "~/components";
...
+
+The TTL is extended upon every new fetch request to the Durable Object. It might be tempting to instead extend the TTL in the constructor of the Durable Object. This is not advised because the Durable Object's constructor will be called before invoking the alarm handler if the alarm wakes the Durable Object up from hibernation. This approach will naively result in the constructor continually extending the TTL without running the alarm handler.
```
Oxyjun:
```suggestion
The TTL is extended upon every new fetch request to the Durable Object. Cloudflare does not advise extending the TTL in the constructor of the Durable Object. This is because the Durable Object's constructor will be called before invoking the [alarm handler](/durable-objects/api/alarms/#handler-methods) if the alarm wakes the Durable Object from hibernation. This approach will naively result in the constructor continually extending the TTL without running the alarm handler.
```

src/content/docs/durable-objects/examples/durable-object-ttl.mdx
joshthoward: This was a copy/paste error. Thank you for catching @vy-ton.

joshthoward:
This PR is replaced by https://github.com/cloudflare/cloudflare-docs/pull/22761 because of the new security rules for our GitHub account.

PR #22758, opened 2025-05-29
[Chore] Add noindex to Wrangler 1 pages

kodster28:
### Summary

With the work in #22754, we're going to need to be much more careful about the "outdated, but still referenceable content in our docs".

An example of this is Wrangler 1 commands.

We still need these for reference though, so we'll be adding the `noindex` flag to them as a middle path.


src/content/docs/workers/wrangler/migration/v1-to-v2/wrangler-legacy/configuration.mdx
```diff
@@ -102,98 +100,90 @@ preview_id = "999..."
 
 To deploy this example Worker to the `helloworld` environment, you would run `wrangler publish --env helloworld`.
 
-***
+---
```
kodster28, line 103: This is all just prettier formatting.

caley-b (approved):
Live long and prosper to no index-ville.

PR #22775, opened 2025-05-30
[Firewall] Add noindex to entire tile

pedrosousa:
### Summary

Addresses PCX-17621.


pedrosousa:
We'll follow a different approach to de-emphasize certain content areas like Firewall Rules for Support AI.
Refer to the following PR for additional context: #22794.

PR #22832, opened 2025-06-02
[Docs Site] Add tags filter to /search/

KianNH:
### Summary

Add tags filter to /search/

Fixes the "update URL with params" logic to remove params when all refinements are unticked and support multiple params to be set at once now that we have multiple filters.

### DIFF
--- a/src/components/search/InstantSearch.tsx
+++ b/src/components/search/InstantSearch.tsx
@@ -34,6 +34,22 @@ function SearchBox(props: UseSearchBoxProps) {
 		}
 	}, []);
 
+	useEffect(() => {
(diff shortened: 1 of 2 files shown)


kodster28 (approved):
Is there a way to also sync the URL query params with the entered search term?

Maybe this is a future ask since it's like this in the current state too... but it feels more out of place with the tags / products getting appended to your current URL.

So you'd get something like, `/search/?q=test&tags=Hono`? It respects this logic when you load the page, but the URL doesn't update when you update the search term.

kodster28 (approved):
LGTM

PR #22929, opened 2025-06-05
[Magic Firewall] Increase Max Capture Time to 24 Hours

Maddy-Cloudflare:


deadlypants1973 (approved):

src/content/docs/magic-firewall/packet-captures/index.mdx
```diff
@@ -10,6 +10,10 @@ import { DirectoryListing } from "~/components"
 
 Cloudflare supports two types of packet captures: full and sample. Full packet captures is the default behavior.
 
+:::note
+The maximum packet capture runtime has increased from five minutes to 24 hours for sample and full packet captures.
```
deadlypants1973: This reads more like a changelog than docs. Is there a better place to put this? Or can we just say it is 24 hours rather than note the increase?

Maddy-Cloudflare: I'll change it to "24 hours" and also inform the PM. Thank you!

PR #22938, opened 2025-06-05
[Docs Site] Add Stream content collection

KianNH:
### Summary

Add Stream content collection


src/content/stream/example-video.yaml
```diff
@@ -0,0 +1,11 @@
...
+title: Example Video
+description: A Super cool video
+products:
+  - workers
+  - workers-ai
+chapters:
+  00:01: Chapter 1
+  00:02: Chapter 2
+thumbnail:
+  url: https://example.com/foo.png
```
kodster28: Should we have a separate field for `transcript_location` too?

I don't love string matching b/c that's less grok-able + more easy errors.

kodster28 (approved):

src/content/stream/example-video/index.yaml
```diff
@@ -0,0 +1,11 @@
+id: streamid
```
kodster28, line 1: Should we add `slug` too?

Since we'll be rendering these as URLs?

PR #22948, opened 2025-06-06
[D1] Improving D1 error doc page.

Oxyjun:
### Summary

1. Removing (largely) duplicated text at the beginning of the page.
2. Extending the error list table to include a "Recommended action" column.
3. Adding more D1 errors into the table.

### Documentation checklist


src/content/docs/d1/observability/debug-d1.mdx
```diff
@@ -60,18 +33,42 @@ The code above throws the following error message:
...
+| `Internal error while starting up D1 DB storage caused object to be reset.` | The Durable Object containing the D1 database is failing to start.                                                                                                        | Retry the operation.                                                                                                                              |
+| `Network connection lost.`                                                  | A network error.                                                                                                                                                          | Retry the operation.                                                                                                                              |
```
joshthoward: This operation cannot safely be retried unless the customer's query is idempotent.

src/content/docs/d1/observability/debug-d1.mdx
```diff
@@ -60,18 +33,42 @@ The code above throws the following error message:
...
+| `Can't read from request stream because client disconnected.`               | A query request was made (e.g. uploading a SQL query), but the connection was closed during the query was fully executed.                                                 | App action: Retry the operation, and ensure the connection stays open.                                                                            |
+| `D1 DB storage operation exceeded timeout which caused object to be reset.` | A query is trying to write a large amount of information (e.g. GBs), and is taking too long.                                                                              | App action: Retry the operation, or split the query into smaller shards.                                                                          |
```
joshthoward: The guidance here should be the same as `D1 DB is overloaded. Requests queued for too long.` immediately retrying could cause further problems for the service.

src/content/docs/d1/observability/debug-d1.mdx
```diff
@@ -60,18 +33,42 @@ The code above throws the following error message:
...
-| `D1_COLUMN_NOTFOUND` | Column not found.                                                                                                                                                |
-| `D1_DUMP_ERROR`      | Database dump error.                                                                                                                                             |
-| `D1_EXEC_ERROR`      | Exec error in line x: y error.                                                                                                                                   |
+D1 returns the following error constants, in addition to the extended (detailed) error message:
 
+| Error message                                                               | Description                                                                                                                                                      | Recommended action                                                                                                                                |
```
joshthoward, line 46: Nit: We could broadly categorize these errors into failed, disconnect, and overload errors following the KJ convention, but I'm not going to insist on that.

joshthoward (approved):
I am going to go ahead and approve, but my two comments other than the nit need to be addressed.

src/content/docs/d1/observability/debug-d1.mdx
Oxyjun: Addressing in new commit

src/content/docs/d1/observability/debug-d1.mdx
Oxyjun: What's the recommended action in this case? CC @lambrospetrou

Oxyjun: Added a warning note at the top of the box to only retry if the query is idempotent, as this guidance needs to be given whenever the recommended action is to retry.

src/content/docs/d1/observability/debug-d1.mdx
Oxyjun: This seems like it might require a code change too. I'll leave the docs as-is for now, but happy to update if we decide to change the logical categorizations

src/content/docs/d1/observability/debug-d1.mdx
```diff
@@ -54,6 +54,17 @@ The following table lists specific instances of `D1_ERROR`.
 
 <Example title="List of D1_ERRORs">
 
+:::caution[Retry operations]
+While some D1 errors can be resolved by retrying the operation, retrying is only safe if your query is idempotent (produces the same result when executed multiple times).
+
+Before retrying any failed operation:
+- Verify your query is idempotent
+- Avoid retrying non-idempotent operations
+- Consider implementing application-level safeguards
```
lambrospetrou: I wouldn't say avoid retrying, but they should make sure their application can make sure they retry only what's needed.
Maybe doing a query to see the current state of the DB and then retrying for example. Or not retrying.

I would merge these two items, so that we only have 2 bullets. The first is making sure the query is idempotent (e.g. readonly, or something like `CREATE TABLE IF NOT EXISTS`), and the second bullet prompts the users to make sure the operation can be retried by doing application-level checks.

PR #22953, opened 2025-06-06
MY_DURABLE_OBJECT => MY_DURABLE_OBJECT_NAMESPACE

irvinebroque:
It is confusing to say that the binding name is `MY_DURABLE_OBJECT` — singular — and then call methods on it in order to....get a single Durable Objects instance.

I don't think this creates the correct mental model, makes it seem like `MY_DURABLE_OBJECT` is a single durable object, instead of what it is — a namepaces of Durable Objects — plural.

Something I am missing as to why we don't use this language today?


Oxyjun (approved):

irvinebroque (commented):
@vy-ton you good with this? any concerns?

PR #22416, opened 2025-05-14
Create terraform-example for bulk redirect

keefetangcf:
### Summary

Adding a new page to demonstrate an example terraform config file for bulk redirect

### Documentation checklist

- [x] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.

PR #23064, opened 2025-06-17
[Terraform] Add link

pedrosousa:
### Summary

Follow-up of #22416.


src/content/docs/terraform/additional-configurations/link-single-redirects.mdx
pedrosousa, line 1: Note: We don't need a redirect for this renamed file because it's an external link.

Maddy-Cloudflare (approved):

PR #23099, opened 2025-06-18
[Bots] Web Bot Auth docs

Oxyjun:
### Summary

We're now introducing "Web Bot Auth" (WBA), which is a more secure authentication method for verifying bots. Previously, verifying bots was only possible through two flavours of IP validation, `public IP list`, and `reverse DNS`.

Web Bot Auth may become the new IETF standard in the near future, and paves the way for better bot detection across the Internet.

This PR restructures and adds information for WBA. Specifically:

1. Creates a new chapter `Verified bots requirements`, which explains what's involved for a bot to be verified.
2. Cuts down the existing `Verified bots policy` chapter to only talk about the policy
3. Creates a new chapter `Verification methods`, which :
    - Guides users through verifying their bot via WBA
    - Guides users through verifying their bot via IP validation (text taken from existing `Policy` chapter)
    - Talks about generic user agents (text taken from existing `Policy` chapter)
    - Points users to other useful resources under "Additional resources" header.

**Things that we need to improve:**

1. [x] Link to blog
2. [x] Possible restructure such that WBA is its own page (and how this affects rest of IA)
3. ~~Consider absorbing `Categories` chapter into `Verification methods` (or somewhere else)~~ Decided against after discussion.
4. [x] Consider defining the three main required headers for WBA into its own subheadings (such that they can be hyperlinked). This relates to point (2).
5. [x] Work out where the FAQs should go. Possibly a new `Reference` > `FAQs` > "WBA FAQs"?

### Documentation checklist


src/content/docs/bots/concepts/bot/verified-bots/verification.mdx
```diff
@@ -0,0 +1,213 @@
...
+ ;alg="ed25519"
+ ;expires=1735693200
+ ;nonce="e8N7S2MFd/qrd6T2R3tdfAuuANngKI7LFtKYI/vowzk4lAZYadIX6wW25MwG7DCT9RUKAJ0qVkU0mEeLElW1qg=="
+ ;tag="web-bot-auth"
+Signature: sig2=:jdq0SqOwHdyHr9+r5jw3iYZH6aNGKijYp/EstF4RQTQdi5N5YYKrD+mCT1HA1nZDsi6nJKuHxUi/5Syp3rLWBA==:
+```
+
+## IP validation
+
+There are two type of IP validation: public IP list and reverse DNS.
```
gcloudflare: I think the previous description of IP validation `A set of validation methods and requirements to gather set IP ranges for a verified service.` is better because it explains what the two types are trying to do.

src/content/docs/bots/concepts/bot/verified-bots/verification.mdx
```diff
@@ -0,0 +1,213 @@
...
+- `got`
+- `nginx-ssl early hints`
+- `node`
+- `node-fetch`
+- `okhttp`
+- `python-requests`
+- `uTorrent`
+
+
+## Additional resources
```
gcloudflare: I wonder if this would be better to add in the Web Bot Auth Signing Requests section?

src/content/docs/bots/concepts/bot/verified-bots/verification.mdx
Oxyjun: Could you help me understand in a bit more detail what the IP validation does? I'm curious to know the higher-level logic.

Is the logic that if a bot comes from IPs that we can independently verify, it shows that it's not suspicious?

gcloudflare: The IP validation method aims to identify all the IP addresses that a verified bot may use to send requests. There are two ways to do this:
1. IP list provided by the bot owner:
The bot owner can host a public list of IP ranges, like [Googlebot’s list](https://developers.google.com/static/search/apis/ipranges/googlebot.json). We fetch and use this list directly for validation.
2. Domain-based reverse DNS validation:
Alternatively, the bot owner can provide a domain (or set of domains) that their bot requests originate from. In this case, we collect the IPs observed in requests with that bot's user agent and perform reverse DNS lookups. If the reverse DNS of an IP resolves to the provided domain, we consider it valid and store it.

Oxyjun: Brilliant, thank you so much!

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,179 @@
...
+3. Go to the **Verified Bots** tab.
+4. For **Verification Method**: select **Request Signature**.
+5. For **Validation Instructions**: enter the URL of your key directory. You can additionally supply User Agents values (and their match patterns) that will be sent by your bot.
+6. Select **Submit**.
+</Steps>
+
+Cloudflare accepts all valid Ed25519 keys found in your key directory. In the event a key already exists in Cloudflare's registered database, Cloudflare will work with you to supply a new key, or rotate your existing key.
+
+:::note[Estimated review time]
+The estimated review time is approximately 1 week.
```
patriciasantaana:
```suggestion
The estimated review time is approximately one week.
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,179 @@
...
+1. Host a key directory at a well known message signatures directory. The key directory should serve a JSON Web Key Set (JWKS) including the public key derived from your signing key.
+
+   An example directory would be:
+   ```txt
+   /.well-known/http-message-signatures-directory/
+   ```
+2. Serve the web page over HTTPS (not HTTP).
+3. Sign your HTTP response using the HTTP message signature specification by attaching one signature per key in your key directory. This ensures no one else can mirror your directory and attempt to register on your behalf. Your response must include the following headers:
+   - `Signature`: TBD
+   - `Signature-Input`: TBD
```
patriciasantaana: will `TBD` be added later on or is a placeholder more suitable for this?

src/content/docs/bots/reference/FAQs.mdx
patriciasantaana, line 1: A [FAQ](https://699adacb.preview.developers.cloudflare.com/bots/troubleshooting/frequently-asked-questions/) already exists here. We could merge the two pages but separate the overall Bots FAQ from Web Bot Auth FAQ with a line break between sections.

Maybe we can even move it out of Troubleshooting/Reference and into its own top-level nav item, since it doesn't only cover troubleshooting now. What do you think?

src/content/docs/bots/reference/FAQs.mdx
```diff
@@ -0,0 +1,55 @@
...
+sidebar:
+  order: 99
+
+---
+
+import { AvailableNotifications } from "~/components"
+
+## Web Bot Auth FAQs
+
+Also refer to [Web Both Auth](/bots/concepts/bot/verified-bots/web-bot-auth/).
```
patriciasantaana: Just makes it a bit clearer.
```suggestion
You may also refer to [Web Both Auth](/bots/concepts/bot/verified-bots/web-bot-auth/) documentation.
```

src/content/docs/bots/reference/FAQs.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+### What key algorithms does Cloudflare support?
+
+Cloudflare does not support key algorithms other than ed25519 .
```
patriciasantaana:
```suggestion
Cloudflare does not support key algorithms other than Ed25519.
```

src/content/docs/bots/reference/FAQs.mdx
```diff
@@ -0,0 +1,55 @@
...
+- `bs` (for HTTP header fields)
+- `key` (for HTTP header fields)
+- `req` (for HTTP header fields or derived components)
+- `name` (for `@query-param` support - this requires `@query-param` support)
+
+### Should I supply a `nonce` parameter in `Signature-Input`?
+
+The `nonce` parameter allows you to supply a `nonce` to prevent attackers from replaying past messages against a server.
+
+While Cloudflare recommends including it, currently there is no `nonce` validation, nor does Cloudflare guard against replay attacks using a database of seen `nonces`.
```
patriciasantaana:
```suggestion
While Cloudflare recommends including it, there is currently no `nonce` validation, nor does Cloudflare guard against replay attacks using a database of seen `nonces`.
```

src/content/docs/bots/reference/FAQs.mdx
```diff
@@ -0,0 +1,55 @@
+---
+pcx_content_type: reference
+title: FAQs
+sidebar:
+  order: 99
+
+---
+
+import { AvailableNotifications } from "~/components"
```
patriciasantaana: Is this component needed?
```suggestion
import { AvailableNotifications } from "~/components"
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: Ah yeah I'll have to add something in here

Edit: I've added in the missing info

src/content/docs/bots/reference/FAQs.mdx
Oxyjun: Nope! Thanks for catching

Oxyjun:
```suggestion

```

src/content/docs/bots/reference/FAQs.mdx
Oxyjun: Ah sorry, I couldn't find it! I'll:

- Make it a top-level item
- Separate out the two FAQs by theme.

Thanks!

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: Edit: I've added in the missing info.

patriciasantaana (approved):

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+</Steps>
+
+By following these steps, you have generated a private key and a public key, then converted the public key to a JWK.
+
+## 2. Host a key directory
+
+You need to host a key directory which creates a way for Cloudflare to authenticate your bot's requests.
+
+<Steps>
+1. Host a key directory at a well known message signatures directory. The key directory should serve a JSON Web Key Set (JWKS) including the public key derived from your signing key.
```
AkshatM: > well known

This term here refers to being under `/.well-known`, and only certain paths are allowed, I believe. 

The example below suggests you could put it under any other name e.g. `'/.well-known/http-message-signatures-directory/foo` or `/.well-known/foo`.

That's incorrect - users can _only_ host their signature directory on `/.well-known/http-message-signatures-directory`. Our tooling will flag anything else as invalid. 

I think we should simply say `host it on /.well-known/http-message-signatures-directory`

AkshatM: Also note we require `Content-Type: application/http-message-signatures-directory+json` today in the response. We might be open to others in the future, but this is mandatory right now.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+Choose a set of components to sign.
+
+A component is either an HTTP header, or any [derived components](https://www.rfc-editor.org/rfc/rfc9421#name-derived-components) in the HTTP Message Signatures specification. Cloudflare recommends the following:
+  - Choose at least the `@authority` derived component, which represents the domain you are sending requests to. For example, a request to `https://example.com` will be interpreted to have an `@authority` of `example.com`.
+  - Use components that only contain ASCII values. HTTP Message Signature specification disallows non-ASCII characters, which will result in failure to validate your bot's requests.
+
+  :::note[Use components with only ASCII values]
+  Cloudflare currently does not support `bs` or `sf` parameter designed to serialize non-ASCII values into ASCII equivalents.
+  :::
+  - Add a `Content-Digest` header if you wish to sign your [message content](https://www.rfc-editor.org/rfc/rfc9421#name-message-content), then specify `Content-Digest` as a component to sign.
```
AkshatM: We should note we don't actually validate their `Content-Digest` header :cry: 

What this means is: anyone can give us a random `Content-Digest` header and sign it. We don't actually check the `Content-Digest` represents the hash of the message body - we only check if the signature _over_ that hash  was valid. Anyone can slap a `Content-Digest` header on.

What this means is that there's no guarantee a `Content-Digest` came from the message it was signed on, and that makes it a security concern.

I think we should recommend people only use this option if there's no risk of a message being altered on the way to us - like if the message was proxied unencrypted to us.

Or we don't talk about `Content-Digest` at all. It's not something we have first class support for anyway. 

I'll let you decide @Oxyjun , but I don't feel comfortable not calling out the caveats of our support if people want to do this.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+#### `Signature` header
+
+Construct a [`Signature` header](https://www.rfc-editor.org/rfc/rfc9421#name-the-signature-http-field) over your chosen components.
+
+#### `Signature-Agent` header
+
+Construct a [`Signature-Agent` header](https://www.ietf.org/archive/id/draft-meunier-http-message-signatures-directory-00.html#name-header-field-definition) that points to your key directory. Note that Cloudflare will fail to verify a message if:
+  - The message includes a `Signature-Agent` header that is not an `https://`.
+  - The message includes a valid URI but do not enclose it in double quotes.
```
AkshatM: Typo: does, not do

src/content/docs/bots/frequently-asked-questions.mdx
```diff
@@ -171,3 +171,59 @@ This is a known issue the Bots team is working to resolve in the near future. In
...
+
+Cloudflare does not support key algorithms other than Ed25519.
+
+---
+
+### What `web-bot-auth` features from the spec are not supported?
+
+The following derived components are not supported, and we will fail to verify a message if they are included:
+
+- `@query-params`: Cloudflare recommends signing the whole query instead of an individual parameter.
```
AkshatM: suggestion: signing the whole query _using the `@query` component_

src/content/docs/bots/frequently-asked-questions.mdx
```diff
@@ -171,3 +171,59 @@ This is a known issue the Bots team is working to resolve in the near future. In
...
+
+While Cloudflare recommends including it, there is currently no `nonce` validation, nor does Cloudflare guard against replay attacks using a database of seen `nonces`.
+
+Instead, Cloudflare recommends short `expires` as a protection against replay attacks. A minute is often sufficient.
+
+---
+
+### How do I know my JSON Web Key set directory will be accepted?
+
+Cloudflare uses [`http-signature-directory` tool](https://crates.io/crates/http-signature-directory) to validate your directory. Please your this works before submitting a verification request.
```
AkshatM: typo + suggestion: Please _ensure_ this works against your directory before _registering with us_. 

(submitting a verification request is ambiguous - does it refer to registration or to sending us a signed request?)

src/content/docs/bots/frequently-asked-questions.mdx
```diff
@@ -171,3 +171,59 @@ This is a known issue the Bots team is working to resolve in the near future. In
...
+
+---
+
+### My message is failing validation. What could be the cause?
+
+- Ensure you have a [`Signature-Agent` header](/bots/concepts/bot/verified-bots/web-bot-auth/#signature-agent-header), and that its value in double-quotes.
```
AkshatM: typo: value _is_ in

AkshatM (changes requested):

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+3. Go to the **Verified Bots** tab.
+4. For **Verification Method**: select **Request Signature**.
+5. For **Validation Instructions**: enter the URL of your key directory. You can additionally supply User Agents values (and their match patterns) that will be sent by your bot.
+6. Select **Submit**.
+</Steps>
+
+Cloudflare accepts all valid Ed25519 keys found in your key directory. In the event a key already exists in Cloudflare's registered database, Cloudflare will work with you to supply a new key, or rotate your existing key.
+
+:::note[Estimated review time]
+The estimated review time is approximately one week.
```
AkshatM, line 114: We should let them know how to track their verification process. Unfortunately, only way to do so is to ask support on the status and have them escalate to us. Should we mention this?

src/content/docs/bots/frequently-asked-questions.mdx
```diff
@@ -171,3 +171,59 @@ This is a known issue the Bots team is working to resolve in the near future. In
...
 
 Note that you need to replace `<API_TOKEN>` with your own [API token](/fundamentals/api/get-started/create-token/).
+
+---
+
+## Web Bot Auth
+
+### What key algorithms does Cloudflare support?
+
+Cloudflare does not support key algorithms other than Ed25519.
```
thibmeu:
```suggestion
Cloudflare supports Ed25519 key algorithm.
```

be in the affirmative, not negative

src/content/docs/bots/frequently-asked-questions.mdx
```diff
@@ -171,3 +171,59 @@ This is a known issue the Bots team is working to resolve in the near future. In
...
+
+---
+
+### What `web-bot-auth` features from the spec are not supported?
```
thibmeu:
```suggestion
### What `web-bot-auth` features from the IETF draft are not supported?
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+   openssl pkey -in private-key.pem -pubout -out public-key.pem
+   ```
+3. Convert the public key to JSON Web Key (JWK) using a tool of your choice. This example uses [`jwker`](https://github.com/jphastings/jwker) command line application.
+   ```sh
+   go install github.com/jphastings/jwker/cmd/jwker@latest
+   jwker public-key.pem public-key.jwk
+   ```
+</Steps>
+
+By following these steps, you have generated a private key and a public key, then converted the public key to a JWK.
```
thibmeu, line 41: we could also point out to JavaScript key generation using WebCrypto API

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey

this would be directly in the right JWK format

most of the existing JWK libraries or provider should be able to do that as well https://jwt.io/libraries

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+title: Web Bot Auth
+sidebar:
+    order: 6
+    label: Web Bot Auth
+
+---
+
+import { GlossaryTooltip, Steps } from "~/components"
+
+Web Bot Auth is an authentication method that leverages cryptographic signatures in HTTP messages to verify that a request comes from an automated bot.
```
thibmeu, line 12:
```suggestion
Web Bot Auth is an authentication method that leverages cryptographic signatures in HTTP messages to verify that a request comes from an automated bot.

It relies on two active IETF drafts: a [directory draft](https://datatracker.ietf.org/doc/html/draft-meunier-http-message-signatures-directory) allowing the crawler to share their public keys, and a [protocol draft](https://datatracker.ietf.org/doc/html/draft-meunier-web-bot-auth-architecture) defining how these keys should be used to attach crawler's identity to HTTP requests.

This documentation goes over specific integration within Cloudflare.
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+   go install github.com/jphastings/jwker/cmd/jwker@latest
+   jwker public-key.pem public-key.jwk
+   ```
+</Steps>
+
+By following these steps, you have generated a private key and a public key, then converted the public key to a JWK.
+
+## 2. Host a key directory
+
+You need to host a key directory which creates a way for Cloudflare to authenticate your bot's requests.
```
thibmeu: changing the sentence to have bots as the actor rather than cloudflare.

In addition, clearly reference the IETF draft, given this is where the format is defined.

```suggestion
You need to host a key directory which creates a way for your bot to authenticate its requests to Cloudflare.
This directory should follow the definition from the active IETF draft [draft-meunier-http-message-signatures-directory-01](https://datatracker.ietf.org/doc/html/draft-meunier-http-message-signatures-directory-01).
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+   - `Signature`: Construct a [`Signature` header](https://www.rfc-editor.org/rfc/rfc9421#name-the-signature-http-field) over your chosen components.
+   - `Signature-Input`: Construct a [`Signature-Input` header](https://www.rfc-editor.org/rfc/rfc9421#name-the-signature-input-http-fi) over your chosen components. The header must meet the following requirements.
+     | Required component parameter | Requirement                                                                                                                 |
+     | ---------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+     | `tag`                        | This should be equal to `http-message-signatures-directory`.                                                                |
+     | `alg`                        | This should be equal to `ed25519`.                                                                                          |
```
thibmeu: @AkshatM i don't think this should be required.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+<Steps>
+1. Host a key directory at a well known message signatures directory. The key directory should serve a JSON Web Key Set (JWKS) including the public key derived from your signing key.
+
+   An example directory would be:
+   ```txt
+   /.well-known/http-message-signatures-directory/
+   ```
+2. Serve the web page over HTTPS (not HTTP).
+3. [Calculate the base64 URL-encoded JWK thumbprint](https://www.rfc-editor.org/rfc/rfc8037.html#appendix-A.3) associated with your Ed25519 public key.
+4. Sign your HTTP response using the HTTP message signature specification by attaching one signature per key in your key directory. This ensures no one else can mirror your directory and attempt to register on your behalf. Your response must include the following headers:
```
thibmeu, line 58: I think we could refer to the security consideration of the directory draft, but open to not

https://datatracker.ietf.org/doc/html/draft-meunier-http-message-signatures-directory-01#name-binding-keys-to-the-directo

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+       "kty": "OKP",
+       "crv": "Ed25519",
+       "x": "JrQLj5P_89iXES9-vFgrIy29clF9CC_oPPsw3c5D0bs", // Base64 URL-encoded public key, with no padding
+     }]
+   }
+   ```
+</Steps>
+
+:::note
+This URL serves a standard JSON Web Key Set. Besides `x`, `crv`, and `kty`, you can include other standard JSON Web Key parameters, and you may publish non-Ed25519 keys as well. Multiple Ed25519 keys are acceptable as well.
```
thibmeu:
```suggestion
This URL serves a standard JSON Web Key Set. Besides `x`, `crv`, and `kty`, you can include other standard JSON Web Key parameters, and you may publish non-Ed25519 keys as well. Multiple Ed25519 keys are supported. Only those for which you provide a signature in the above format are going to be used.
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+Cloudflare will ignore all other key types and key parameters except those containing `kty`, `crv`, and `x` formatted above. Do not include information that would leak your private key, such as the `d` parameter.
+:::
+
+You can use the Cloudflare-developed [`http-signature-directory` CLI tool](https://crates.io/crates/http-signature-directory) to assist you in validating your directory.
```
thibmeu, line 96: nits: it would be nice to have a website to do this. maybe https://github.com/cloudflare/web-bot-auth/issues/37

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+:::note[Estimated review time]
+The estimated review time is approximately one week.
+
+After successful verification, you will be able to send verified requests.
+:::
+
+## 4. (After verification) Sign your requests
+
+After your bot has been successfully verified, you need to sign your bot's requests.
```
thibmeu: slightly updating the wording.

clearly reference the revision of the IETF draft we support. this is going to be valuable in the future as the draft evolves

```suggestion
After your bot has been successfully verified, your bot is ready to sign its requests. The signature protocol is defined in [draft-meunier-web-bot-auth-architecture-02](https://datatracker.ietf.org/doc/html/draft-meunier-web-bot-auth-architecture-02)

```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+| Required component parameter | Requirement                                                                                                                                                                                                                                               |
+| ---------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `tag`                        | This should be equal to `web-bot-auth`.                                                                                                                                                                                                                   |
+| `alg`                        | This should be equal to `ed25519`.                                                                                                                                                                                                                        |
```
thibmeu: the draft suggest that using `alg` should be avoided. not prohibited, but I don't think we should have it in our docs @AkshatM 

```suggestion
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+  - Use components that only contain ASCII values. HTTP Message Signature specification disallows non-ASCII characters, which will result in failure to validate your bot's requests.
+
+  :::note[Use components with only ASCII values]
+  Cloudflare currently does not support `bs` or `sf` parameter designed to serialize non-ASCII values into ASCII equivalents.
+  :::
+  - Add a `Content-Digest` header if you wish to sign your [message content](https://www.rfc-editor.org/rfc/rfc9421#name-message-content), then specify `Content-Digest` as a component to sign.
+
+### 4.2. Calculate the JWK thumbprint
+
+[Calculate the base64 URL-encoded JWK thumbprint](https://www.rfc-editor.org/rfc/rfc8037.html#appendix-A.3) associated with your Ed25519 public key registered with Cloudflare.
```
thibmeu:
```suggestion
[Calculate the base64 URL-encoded JWK thumbprint](https://www.rfc-editor.org/rfc/rfc8037.html#appendix-A.3) from the public key you registered with Cloudflare.
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
```diff
@@ -0,0 +1,187 @@
...
+| `created`                    | This should be equal to a `Unix` timestamp associated with when the message was sent by your application.                                                                                                                                                 |
+| `expires`                    | This should be equal to a `Unix` timestamp associated with when Cloudflare should no longer attempt to verify the message. A short `expires` reduces the likelihood of replay attacks, and Cloudflare recommends choosing suitable short-lived intervals. |
+
+#### `Signature` header
+
+Construct a [`Signature` header](https://www.rfc-editor.org/rfc/rfc9421#name-the-signature-http-field) over your chosen components.
+
+#### `Signature-Agent` header
+
+Construct a [`Signature-Agent` header](https://www.ietf.org/archive/id/draft-meunier-http-message-signatures-directory-00.html#name-header-field-definition) that points to your key directory. Note that Cloudflare will fail to verify a message if:
```
thibmeu:
```suggestion
Construct a [`Signature-Agent` header](https://www.ietf.org/archive/id/draft-meunier-http-message-signatures-directory-01.html#name-header-field-definition) that points to your key directory. Note that Cloudflare will fail to verify a message if:
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
thibmeu:
```suggestion
  - The message includes a valid URI but does not enclose it in double quotes. This is due to Signature-Agent being a structured field.
```

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: It will need to be changed after LDW. Today, we require this in the implementation, and things will not verify otherwise. I'll raise a ticket.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: It will need to be changed after LDW. Today, we require this in the implementation, and things will not verify otherwise, so it needs to be in the docs right now. I'll raise a ticket.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: actually, we can probably change this sooner - it's only required in `https://crates.io/crates/http-signature-directory`. @Oxyjun yes, please, remove this in this section, it is not needed.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: This requires some pretty intense changes to the whole `keyring` concept in `web-bot-auth` crate, hence the slowness on my end eradicating the need for `alg`.

AkshatM: I updated web-bot-auth crate to not need `alg` anymore in v0.3.0. 

*However, this documentation should not be changed* - we still need to update upstream dependencies to use `web-bot-auth 3.0` and that's unlikely to happen until after LDW. Until then, customers will be forced to send `alg`.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: I have updated `http-signature-directory` to not require an `alg` parameter in v0.3.0. This documentation string is safe to delete.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: Added comment:

image

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: image
Added

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: I'll leave it out for now, as I feel too many external hyperlinks can detract users from coherently reading through a page. Happy to add it in later if we feel it's necessary though.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: To clarify, we're only removing the `alg` parameter from the `Signature-Input` header?

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: Thanks so much for the context! I'll:

- Remove it as an action step
- Add it as a warning note to specifically **not** use it if there's a risk of it being altered.

image

src/content/docs/bots/frequently-asked-questions.mdx
Oxyjun: Thanks for catching!

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: I'm tempted to leave out for now, unless there's an automatic `progress tracker` they can see?

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: We can add this as a future update to the docs once the feature is live

src/content/docs/bots/concepts/bot/verified-bots/overview.mdx
```diff
@@ -0,0 +1,37 @@
...
+## Verified bot requirement
+
+For a bot to be verified, it must meet the following requirements:
+
+1. The bot must follow [verified bots policy](/bots/concepts/bot/verified-bots/policy/).
+2. The bot must be verified using one of the following verification methods:
+   - [Web Bot Auth](/bots/concepts/bot/verified-bots/web-bot-auth/)
+   - [IP validation](/bots/concepts/bot/verified-bots/ip-validation/)
+
+Once Cloudflare verifies a bot, it will appear on the [Cloudflare Radar's list of verified bots](https://radar.cloudflare.com/verified-bots).
```
thibmeu:
```suggestion
Once Cloudflare verifies a bot, it should appear on the [Cloudflare Radar's list of verified bots](https://radar.cloudflare.com/verified-bots).
```

thibmeu (commented):
overall looks ok. I'll let @AkshatM confirm the `alg` integration detail

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: Can we please remove the `/` at the end of `/.well-known/http-message-signatures-directory/` - that's specifically what I meant, it's a file, not a directory despite the name.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: KLeave it out, should be fine.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: Yes.

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
AkshatM: @Oxyjun I've released a change that should fix this. We can safely remove the `alg` requirement here as well.

AkshatM (approved):
One or two final comments, but marking as appriove to unblock you

src/content/docs/bots/concepts/bot/verified-bots/web-bot-auth.mdx
Oxyjun: I'd misunderstood - fixed!

PR #23140, opened 2025-06-20
[Email Security] Add limitations to DLP Assist

Maddy-Cloudflare:


src/content/docs/cloudflare-one/email-security/outbound-dlp.mdx
```diff
@@ -83,4 +83,20 @@ To block, encrypt, or send approval, you can configure rules within Microsoft Pu
...
+
+Outbound DLP presents its limitations:
+
+- Outbound DLP only protects user-managed inboxes.
+- Outbound DLP offers the most consistent experience on Outlook Web App and Outlook desktop, due to limitations imposed by Microsoft.
+
+| Platform                             	| Status                                                   	|
+|--------------------------------------	|----------------------------------------------------------	|
+| Web client                           	| Stable                                                   	|
+| New Outlook Desktop client - Windows 	| Stable                                                   	|
```
caley-b: You may be able to lowercase "desktop" here and the following line instead: https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/d/desktop. Super tiny and doesn't inhibit the user's ability to get the info they need. Sharing in case you want to implement it.

src/content/docs/cloudflare-one/email-security/outbound-dlp.mdx
```diff
@@ -83,4 +83,20 @@ To block, encrypt, or send approval, you can configure rules within Microsoft Pu
...
+:::
+
+### Limitations
+
+Outbound DLP presents its limitations:
+
+- Outbound DLP only protects user-managed inboxes.
+- Outbound DLP offers the most consistent experience on Outlook Web App and Outlook desktop, due to limitations imposed by Microsoft.
+
+| Platform                             	| Status                                                   	|
```
caley-b, line 95: Tiny. Not blocking feedback. Just to be extra clear that these are Microsoft limitations, it may be helpful to say something like "Platform with Outlook" or "Outlook version and Platform" or something like that? Trying to thread the needle between the different versions of Outlook (from legacy vs. current to mobile, web, and desktop) with each platform.

caley-b (approved):
Tiny suggestions only.

src/content/docs/cloudflare-one/email-security/outbound-dlp.mdx
Maddy-Cloudflare: Thanks for your feedback :) 

These limitations are for Microsoft and Apple (Apple has an operating system called macOS for desktops, and iOS for phones).

PR #23190, opened 2025-06-24
[Docs Site] Sync filters to URL params in catalog components

KianNH:
### Summary

Sync filters to URL params in catalog components


kodster28 (approved):
Thank you, @KianNH !

PR #23198, opened 2025-06-24
[Docs Site] Fetch all Linux platforms for WARP releases

KianNH:
### Summary

Fetch all Linux platforms for WARP releases


kodster28 (approved):
There's some other stuff that has to happen here -- namely updating to pull names from the API + updating the size dynamically based on the selected package -- but none of it is blocking.

PR #23203, opened 2025-06-24
[Cache] Adds clarification about `Cloudflare-Cdn-Cache-Control`.

angelampcosta:
### Summary

Add clarification about `Cloudflare-Cdn-Cache-Control` on OCC docs.

### Documentation checklist

### DIFF
--- a/src/content/docs/cache/concepts/cache-control.mdx
+++ b/src/content/docs/cache/concepts/cache-control.mdx
@@ -197,7 +197,7 @@ Certain scenarios also affect Origin Cache Control behavior when it is enabled o
 </table>
 
 :::note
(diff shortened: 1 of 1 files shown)


src/content/docs/cache/concepts/cache-control.mdx
```diff
@@ -196,6 +196,10 @@ Certain scenarios also affect Origin Cache Control behavior when it is enabled o
   </tbody>
 </table>
 
+:::note
+When the `Cloudflare-Cdn-Cache-Control` header is set, OCC is effectively turned **on** (regardless of whether OCC is enabled or disabled). As a result, we apply our Authorization header logic (per [RFC 7234, Section 3.2](https://tools.ietf.org/html/rfc7234#section-3.2)) to allow only the `s-maxage`, `must-revalidate`, or `public` directives. If any other directive is present, we do not cache the asset and instead return `BYPASS`.
```
angelampcosta:
```suggestion
When the `Cloudflare-Cdn-Cache-Control` header is set, OCC is turned **on** (regardless of whether OCC is enabled or disabled). As a result, we apply our Authorization header logic (per [RFC 7234, Section 3.2](https://tools.ietf.org/html/rfc7234#section-3.2)) to allow only the `s-maxage`, `must-revalidate`, or `public` directives. If any other directive is present, we do not cache the asset and instead return `BYPASS`.
```

patriciasantaana (approved):

PR #23218, opened 2025-06-25
Document how to run User Worker on only specific paths

nicolaic:
### Summary

Adds a new section to Worker script routing documentation about how to configure routing such that a User Worker is only invoked on specific paths.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/static-assets/routing/worker-script.mdx
+++ b/src/content/docs/workers/static-assets/routing/worker-script.mdx
@@ -127,3 +127,44 @@ export default class extends WorkerEntrypoint<Env> {
 }
 ```
 </TypeScriptExample>
+
+### Run Worker only on selective paths
+
+You can configure a User Worker to only be invoked on specific paths, by using an array of positive route patterns, followed by a negative wildcard pattern. This configuration can be very helpful when hosting a static website with a Worker as a simple backend service:
+
+For more information you can take a look at the [routing reference diagram](/workers/static-assets/routing/static-site-generation/#reference).
+
+<WranglerConfig>
+
+```json
+{
+	"name": "my-worker",
+	"compatibility_date": "$today",
+	"main": "./worker/index.ts",
+	"assets": {
+		"directory": "./dist/",
+		"not_found_handling": "single-page-application",
+		"binding": "ASSETS",
+		"run_worker_first": ["/api/*", "!**"]
+	}
+}
+```
+
+</WranglerConfig>
+
+<TypeScriptExample filename="./worker/index.ts">
+
+```ts
+import { WorkerEntrypoint } from "cloudflare:workers";
+
+export default class extends WorkerEntrypoint<Env> {
+	async fetch(request: Request) {
+		// This Worker script is only called for requests on paths starting with `/api/`
+		return Response.json({
+			success: true,
+			message: 'Hello from API endpoint',
+		});
+	}
+}
+```
+</TypeScriptExample>


src/content/docs/workers/static-assets/routing/worker-script.mdx
```diff
@@ -127,3 +127,44 @@ export default class extends WorkerEntrypoint<Env> {
...
+```json
+{
+	"name": "my-worker",
+	"compatibility_date": "$today",
+	"main": "./worker/index.ts",
+	"assets": {
+		"directory": "./dist/",
+		"not_found_handling": "single-page-application",
+		"binding": "ASSETS",
+		"run_worker_first": ["/api/*", "!**"]
```
GregBrimble, line 148: I'll need to double-check, but I had thought that:

(a) rules need to start with either `/` or `!/`,
(b) exclude rules were evaluated first (so this exclude would mean no path ever matches `run_worker_first`), and
(c) splats are greedy (so a double-splat in the exclude like this is meaningless)

@matthewdavidrodgers , can you confirm?

matthewdavidrodgers: yes to all three points; this is more effectively written as 
`"run_worker_first": ["/api/*"],`

src/content/docs/workers/static-assets/routing/worker-script.mdx
```diff
@@ -127,3 +127,44 @@ export default class extends WorkerEntrypoint<Env> {
 }
 ```
 </TypeScriptExample>
+
+### Run Worker only on selective paths
+
+You can configure a User Worker to only be invoked on specific paths, by using an array of positive route patterns, followed by a negative wildcard pattern. This configuration can be very helpful when hosting a static website with a Worker as a simple backend service:
```
matthewdavidrodgers, line 133:
```suggestion
You can configure a User Worker to only be invoked on specific paths by using an array of route patterns. This configuration can be very helpful when hosting a static website with a Worker as a simple backend service:
```

src/content/docs/workers/static-assets/routing/worker-script.mdx
matthewdavidrodgers: to illustrate a situation where a negative path is necessary, it must also match a broader positive rule to exempt those paths from being directed to the worker.
e.g. `"run_worker_first": ["/api/*", "!/api/docs/*"]`

matthewdavidrodgers (changes requested):

nicolaic:
@matthewdavidrodgers As far as I can tell, that unfortunately does not achieve the desired behaviour of routing to the User Worker only on specific routes and no other routes, that was originally described in the documentation. Looking at the routing reference diagram indicates to me, the original configuration should, in theory, achieve this behaviour.

I haven't used Cloudflare Pages before, but I now see that I was mistaken, in thinking that you cloud achieve this with `excludes` field in `_routes.json`, which I assume in the intention that this routing behaviour is attempting to mimic?


I'm not doubtful that it is possible to achieve the configuration described, and perhaps my time could be spent better, either try update existing documentation to make this more clear, or maybe find an existing issue or create a new one regarding the desire for this behaviour.

I've tried to figure this out previously, and now again with the update `run_worker_first` config options, but in the end had to settle for having two separate Workers for both frontend and backend.

GregBrimble:
If you want to just have single positive matches, that should just be `run_worker_first = ["/foo", "/bar"]`!

matthewdavidrodgers:
Not sure I understand what you are unable to achieve; the default is to serve assets without hitting the worker, so adding a positive path to `run_worker_first` exempts those paths from asset-first behavior while retaining the default for all other paths. If you don't find that to be the case when trying it out, please open an issue in workers-sdk

GregBrimble:
Will close this in the meantime. Thanks! :)

PR #23219, opened 2025-06-25
fix: syntax - missing parentheses

pReya:
### Summary

Containers docs "Get started" page was missing a closing parentheses, so the enclosed link did not work.


kodster28:
Dupe of #23231

PR #23226, opened 2025-06-25
[Docs] Update External CI/CD guide

mhdcodes:
### Summary

I have added an extra step to install pnpm, as the action fails to install wrangler without.


kodster28:
Can you provide more details about the error messages you're seeing?

From my testing, I see some warnings around specific wrangler installations without the pnpm... but that doesn't prevent Wrangler from being installed.

PR #23233, opened 2025-06-25
Update DNS record instructions for browser-based RDP

asamborski:
Update DNS record instructions to be more descriptive on how to handle a AAAA record and a CNAME record. Reorder instructions to try to call better attention to DNS requirements (put application creation after DNS setup).


ranbel:
changes published in #23238

PR #23238, opened 2025-06-25
[ZT] RDP DNS record edits

ranbel:


src/content/docs/cloudflare-one/connections/connect-networks/use-cases/rdp/rdp-browser.mdx
```diff
@@ -34,7 +34,47 @@ Browser-based RDP can be used in conjunction with [routing over WARP](/cloudflar
...
+To make your RDP targets (that is, your Windows machines) available through the browser, you will need a [Cloudflare DNS record](/dns/manage-dns-records/how-to/create-dns-records/) for the domain and subdomain that users will connect to.  This domain will be used to access any targets that are available to users through your Access application (see Step 4).
+
+For example, if want users to connect to targets on `rdp.example.com`, [create a DNS record](/dns/manage-dns-records/how-to/create-dns-records/#create-dns-records) for `rdp.example.com`. You can create either an `A`, `AAAA`, or `CNAME` record:
+
+<Details header="A record" open={false}>
+The following DNS record points your public subdomain (`rdp`) to the default IPv4 address range:
+
+- **Type**: _A_
+- **Name**: `rdp`
+- **IPv4 address**: `0.0.0.0`
```
asamborski: This address won't work. For example, we could use 240.0.0.0 instead (this is part of the class E for future use IPv4 addresses)

src/content/docs/cloudflare-one/connections/connect-networks/use-cases/rdp/rdp-browser.mdx
```diff
@@ -34,7 +34,47 @@ Browser-based RDP can be used in conjunction with [routing over WARP](/cloudflar
 
 <Render file="access/add-target" params={{ protocol: "rdp" }}/>
 
-## 3. Create an Access application
+## 3. Create a DNS record
+
+To make your RDP targets (that is, your Windows machines) available through the browser, you will need a [Cloudflare DNS record](/dns/manage-dns-records/how-to/create-dns-records/) for the domain and subdomain that users will connect to.  This domain will be used to access any targets that are available to users through your Access application (see Step 4).
```
deadlypants1973, line 39:
```suggestion
To make your RDP targets (that is, your Windows machines) available through the browser, you will need a [Cloudflare DNS record](/dns/manage-dns-records/how-to/create-dns-records/) for the domain and subdomain that users will connect to. This domain will be used to access any targets that are available to users through your Access application (refer to step 4).
```

deadlypants1973: I believe "targets" is ZTIA language meaning "machine being accessed by users"? We discussed the same about SSH. In my understanding, targets is not an industry term but rather something CF came up with to describe how to use us. If I am correct in my understanding, we should get clear on this definition and add it to the CF1 glossary to avoid confusion

cc @asamborski

src/content/docs/cloudflare-one/connections/connect-networks/use-cases/rdp/rdp-browser.mdx
```diff
@@ -34,7 +34,47 @@ Browser-based RDP can be used in conjunction with [routing over WARP](/cloudflar
...
+- **Type**: _CNAME_
+- **Name**: `rdp`
+- **Target**: `www.rdp.example.com`
+- **Proxy status**: On
+
+The CNAME **Target** field is unrelated to the RDP targets configured in Step 2.
+
+</Details>
+
+The DNS record does not need to point to an active destination IP address or hostname; the record just needs to be valid. Cloudflare's RDP proxy will handle the routing to the correct RDP target.
```
deadlypants1973:
```suggestion
The DNS record does not need to point to an active destination IP address or hostname; the DNS record just needs to be valid. Cloudflare's RDP proxy will handle the routing to the correct RDP target.
```

deadlypants1973 (approved):

PR #23243, opened 2025-06-25
[Gateway] Connection establishment

maxvp:
PCX-17962


src/content/partials/cloudflare-one/gateway/order-of-enforcement.mdx
```diff
@@ -135,6 +135,27 @@ flowchart TB
     egress1-- "Egress with dedicated IP" -->internet
 ```
 
+## Connection establishment
+
+When a user connects to a server with Gateway, Gateway first establishes a TCP connection with the destination server on the port the user requested. If the connection is successful, Gateway will apply policies. If Gateway policies allow the connection, Gateway will connect the user to the destination server. If Gateway policies block the connection, Gateway will end the connection and will not send any data between the user and the destination server. If the TCP connection to the destination server is unsuccessful, Gateway will not run any policies nor accept further TCP connections from the user to the server.
```
Encore-Encore: @jgpaiva what can we say about the first packet? We don't send any user info so it would be helpful to be as specific as possible to call that out.

jgpaiva: > nor accept further TCP connections from the user to the server.

I'd like @Encore-Encore's opinion here, but this reads a bit strange to me. Reading this, I'd interpret that gateway is keeping state that the upstream is broken and immediately rejecting new connections when they arrive. But that's not true, gateway tries to connect again (this time the upstream might already be responding).

jgpaiva: Maybe something around "since TCP traffic is proxied by Cloudflare, the connection that Gateway establishes with the upstream is independent from the connection the eyeball establishes with Gateway. This means it has new source IP and port, and no details from the original eyeball TCP handshake are included in this TCP handshake with the origin"?

src/content/partials/cloudflare-one/gateway/order-of-enforcement.mdx
```diff
@@ -135,6 +135,29 @@ flowchart TB
     egress1-- "Egress with dedicated IP" -->internet
 ```
 
+## Connection establishment
+
+When a user connects to a server with Gateway, Gateway first establishes a TCP connection with the destination server on the port the user requested. If the connection is successful, Gateway will apply policies. If Gateway policies allow the connection, Gateway will connect the user to the destination server. If Gateway policies block the connection, Gateway will end the connection and will not send any data between the user and the destination server. If the TCP connection to the destination server is unsuccessful, Gateway will not run any policies and retry TCP connections from the user to the server.
```
Encore-Encore: Can we add this as the second sentence here? I want to make sure its the first thing admins see when reading this section. 

Because TCP traffic is proxied by Cloudflare, the connection Gateway establishes with the origin is independent from the connection users establish with Gateway. This means Gateway assigns a new source IP and port to the user's connection and no details from the user's TCP handshake are included in the TCP handshake with the origin server.

patriciasantaana (approved):

PR #23303, opened 2025-06-28
fix: use the right `type` when using `degit`

yeskunall:
### Summary

`exec` executes a shell command within the scope of the project, but in this case we’re creating a brand new project using `degit`, so the right [`type`](https://starlight-package-managers.vercel.app/usage/#dlx) is `dlx`.

### Documentation checklist

---

💖


yeskunall:
@GregBrimble @irvinebroque @mikenomitch @korinne @WalshyDev @kodster28

Apologies, but I’m just doing as the comment above suggests. I believe this is a small fix, so can one of you folks review it? Thanks very much!

GregBrimble (approved):

PR #23402, opened 2025-07-02
[ZT] Max SSH session duration

ranbel:
PCX-17835

### DIFF
--- a/src/content/docs/cloudflare-one/faq/troubleshooting.mdx
+++ b/src/content/docs/cloudflare-one/faq/troubleshooting.mdx
@@ -377,7 +377,7 @@ Cloudflare Access uses a [`CF_Session` cookie](/cloudflare-one/identity/authoriz
 
 ## Long-lived SSH sessions frequently disconnect.
 
-All SSH connections proxied through Cloudflare Gateway, including traffic to [Access for Infrastructure](/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access/) targets, have a maximum guaranteed duration of 10 hours. It is possible for connections to last longer than 10 hours. However, if a connection is active at the time of a Gateway release, Gateway will terminate the connection 10 hours later. Releases are not scheduled and can occur multiple times a week.
+All connections proxied through Cloudflare Gateway, including traffic to [Access for Infrastructure](/cloudflare-one/connections/connect-networks/use-cases/ssh/ssh-infrastructure-access/) SSH targets, have a maximum guaranteed duration of 10 hours. It is possible for connections to last longer than 10 hours. However, if a connection is active at the time of a Gateway release, Gateway will terminate the connection 10 hours later. Releases are not scheduled and can occur multiple times a week.
 
 To prevent long-lived SSH connections from breaking unexpectedly, we recommend terminating sessions on a predefined schedule. For example, you could set an 8-hour idle timeout so that inactive sessions automatically disconnect during off hours. To configure an idle timeout, add the `ChannelTimeout` option to either the SSH server (`/etc/ssh/sshd_config`) or client configuration file (`~/.ssh/config`):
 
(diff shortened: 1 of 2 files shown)


ranbel:
@asamborski

src/content/docs/cloudflare-one/faq/troubleshooting.mdx
```diff
@@ -374,3 +374,18 @@ Get-ChildItem -Path "HKLM:\SOFTWARE\Microsoft\NET Framework Setup\NDP" -Recurse
...
+To ensure a consistent user experience, we recommend configuring your environment to automatically terminate sessions over 10 hours. For example, you could require users to sign in once a day so that connections do not cut off unexpectedly during the work day. One option is to add [`ControlPersist`](https://man.openbsd.org/ssh_config#ControlPersist) to your personal (`~/.ssh/config`) or global (`/etc/ssh/ssh_config`) SSH client configuration file:
+
+```txt
+Host *
+    ControlMaster auto
+    ControlPath ~/.ssh/control_sockets/%r@%h:%p
+    ControlPersist 10h
+```
+
+`ControlPersist` will close inactive SSH connections on the client machine after the specified time period.
```
lgmugnier: This is slightly incorrect, so I'm going to be a annoyingly pedantic here because SSH makes this hard. (also yikes sorry the below turned out so long)

So, the premise is that the longest duration we _guarantee_ is 10 hours. Connections can last longer than that but we don't have a regular release schedule and when a release goes out, we support existing connections for 10 hours before closing them. 

The issue that customers complain about is that a release can happen and then 10 hours later the connection is broken (and it's in the middle of the work day when people really need reliable and consistent behavior). So, the challenge is figuring out how to make sure any live connections die during off-hours so that even if a user is unlucky enough to make their first connection of the day one millisecond before a new release goes live, they don't experience any inconvenience. 

The `ControlPersist` timer starts as soon as the user's first SSH command ends. It keeps a background connection open so that any new SSH commands use the existing connection. However, and key to our situation, it doesn't close SSH commands that are running when that timer expires. A user might start a long-lived connection at the end of their day and when they interact with it the next day, it'll break unexpectedly. So, it doesn't really solve the issue in all cases.

We could instead (or in addition to) suggest using idle timeouts so that connections close automatically during off-hours. If you set a connection to close after 8 hours without any traffic,  it should make sure that any unused connections die during off-hours. You can do this via the SSH server or client using the `ChannelTimeout` config option (available in OpenSSH >9.2). Implementing it on the client side allows your users to choose a time that works for them but implementing it at the server prevents your users from having to worry about it.
- server-side documentation: https://man.openbsd.org/sshd_config#ChannelTimeout
- client-side documentation: https://man.openbsd.org/ssh_config#ChannelTimeout

src/content/docs/cloudflare-one/faq/troubleshooting.mdx
```diff
@@ -374,3 +374,18 @@ Get-ChildItem -Path "HKLM:\SOFTWARE\Microsoft\NET Framework Setup\NDP" -Recurse
 ## I get an `Invalid session. Please try logging in again.` error from Access when trying to log in to the Cloudflare dashboard via SSO.
 
 Cloudflare Access uses a [`CF_Session` cookie](/cloudflare-one/identity/authorization-cookie/#access-cookies) to validate that the same browser both initiated and completed your sign-in. The `Invalid session` error means Access was unable to validate this cookie. Ensure that there is no software or firewall on your device or network that may be interfering with requests to Access.
+
+## Long-lived SSH sessions frequently disconnect.
```
Michael9127, line 378: This 10 hour guarantee applies to more than just SSH. If we mention that it applies to all Gateway traffic here, shouldnt we also call it out in a section thats not specific to SSH?

ranbel: where else would users run into the 10-hour limit, besides SSH?

Michael9127: All traffic that passes through Gateway. So outbound public Gateway traffic, warp-to-tunnel, Access Private Apps, Access Infra Apps, Biso, etc

ranbel: Given that Ankur hasn't seen any complaints from the Gateway side (from the gchat thread), I'm inclined to keep it specific to SSH until we hear otherwise.

deadlypants1973 (approved):

PR #23478, opened 2025-07-07
[DO] Adding that ReadableStream cancellation propagates from Worker to DO

Oxyjun:
### Documentation checklist

### DIFF
--- a/src/content/docs/durable-objects/examples/readable-stream.mdx
+++ b/src/content/docs/durable-objects/examples/readable-stream.mdx
@@ -1,0 +1,99 @@
+---
+type: example
+summary: Stream ReadableStream from Durable Objects.
+pcx_content_type: example
+title: Use ReadableStream with Durable Object and Workers
+sidebar:
+  order: 3
+description: Stream ReadableStream from Durable Objects.
+---
+
+import { GlossaryTooltip, WranglerConfig } from "~/components";
+
+This example demonstrates:
+
+- A Worker receives a request, and forwards it to a Durable Object `my-id`.
+- The Durable Object streams an incrementing number every second, until it receives `AbortSignal`.
+- The Worker reads and logs the values from the stream.
+- The Worker then cancels the stream after 5 values.
+
+```ts
+import { DurableObject } from 'cloudflare:workers';
+
+async function* dataSource(signal: AbortSignal) {
+    let counter = 0;
+    while (!signal.aborted) {
+        yield counter++;
+        await new Promise((resolve) => setTimeout(resolve, 1_000));
+    }
+
+    console.log('Data source cancelled');
+}
+
+export class MyDurableObject extends DurableObject<Env> {
+    async fetch(request: Request): Promise<Response> {
+        const abortController = new AbortController();
+
+        const stream = new ReadableStream({
+            start(controller) {
+                if (request.signal.aborted) {
+                    controller.close();
+                    return;
+                }
+
+                (async () => {
+                    for await (const value of dataSource(AbortSignal.any([request.signal, abortController.signal]))) {
+                        controller.enqueue(new TextEncoder().encode(String(value)));
+                    }
+                })();
+            },
+            cancel() {
+                console.log('Stream cancelled');
+                abortController.abort();
+            },
+        });
+
+        const headers = new Headers({
+            'Content-Type': 'application/octet-stream',
+        });
+
+        return new Response(stream, { headers });
+    }
+}
+
+export default {
+    async fetch(request, env, ctx): Promise<Response> {
+        const id: DurableObjectId = env.MY_DURABLE_OBJECT.idFromName('my-id');
+        const stub = env.MY_DURABLE_OBJECT.get(id);
+        const response = await stub.fetch(request);
+        if (!response.ok || !response.body) {
+            return new Response('Invalid response', { status: 500 });
+        }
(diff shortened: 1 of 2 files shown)


src/content/docs/durable-objects/examples/readable-stream.mdx
```diff
@@ -0,0 +1,99 @@
...
+import { GlossaryTooltip, WranglerConfig } from "~/components";
+
+This example demonstrates:
+
+- A Worker receives a request, and forwards it to a Durable Object `my-id`.
+- The Durable Object streams an incrementing number every second, until it receives `AbortSignal`.
+- The Worker reads and logs the values from the stream.
+- The Worker then cancels the stream after 5 values.
+
+```ts
```
joshthoward, line 21: We should have both JavaScript and TypeScript examples for all Durable Objects examples. I think we favor TypeScript currently, but Gemini could convert these for us. Even better, we could probably have `tsc` compile them for us. This does not need to be a blocker though.

src/content/docs/durable-objects/examples/readable-stream.mdx
```diff
@@ -0,0 +1,99 @@
...
+import { DurableObject } from 'cloudflare:workers';
+
+async function* dataSource(signal: AbortSignal) {
```
joshthoward, line 25: Add comment, e.g. `// Send incremented counter value every second`

src/content/docs/durable-objects/examples/readable-stream.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+        return new Response(stream, { headers });
+    }
+}
+
+export default {
+    async fetch(request, env, ctx): Promise<Response> {
+        const id: DurableObjectId = env.MY_DURABLE_OBJECT.idFromName('my-id');
+        const stub = env.MY_DURABLE_OBJECT.get(id);
+        const response = await stub.fetch(request);
```
joshthoward: I believe that this should be `const response = await stub.fetch(request, {signal});`, right @TimoWilhelm?

src/content/docs/durable-objects/examples/readable-stream.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+        const stream = new ReadableStream({
+            start(controller) {
+                if (request.signal.aborted) {
+                    controller.close();
+                    return;
+                }
+
+                (async () => {
+                    for await (const value of dataSource(AbortSignal.any([request.signal, abortController.signal]))) {
```
joshthoward: We could probably simplify this example to allow the client to only cancel the stream via `abortController.signal` and not `request.signal`. The example here is more complete though. I could go either way.

src/content/docs/durable-objects/examples/readable-stream.mdx
```diff
@@ -0,0 +1,99 @@
...
+            }
+            const { value, done } = await reader.read();
+            console.log(`Got value ${new TextDecoder().decode(value)}`);
+            if (done) {
+                break;
+            }
+            i++;
+        }
+
+        return new Response('Done', { status: 200 });
```
joshthoward: I would prefer it if the client got the streamed data rather than a dummy response after the stream is complete.

src/content/docs/durable-objects/examples/readable-stream.mdx
Oxyjun: We have a component for this that I'd forgotten to use - added it in now.

TimoWilhelm (commented):
This should return the values and simplify the cancellation a bit

```ts
import { DurableObject } from 'cloudflare:workers';

async function* dataSource(signal: AbortSignal) {
    let counter = 0;
    while (!signal.aborted) {
        yield counter++;
        await new Promise((resolve) => setTimeout(resolve, 1_000));
    }

    console.log('Data source cancelled');
}

export class MyDurableObject extends DurableObject<Env> {
    async fetch(request: Request): Promise<Response> {
        const abortController = new AbortController();

        const stream = new ReadableStream({
            async start(controller) {
                if (request.signal.aborted) {
                    controller.close();
                    abortController.abort();
                    return;
                }

                for await (const value of dataSource(abortController.signal)) {
                    controller.enqueue(new TextEncoder().encode(String(value)));
                }
            },
            cancel() {
                console.log('Stream cancelled');
                abortController.abort();
            },
        });

        const headers = new Headers({
            'Content-Type': 'application/octet-stream',
        });

        return new Response(stream, { headers });
    }
}

export default {
        const id: DurableObjectId = env.MY_DURABLE_OBJECT.idFromName('my-id');
        const stub = env.MY_DURABLE_OBJECT.get(id);
        const response = await stub.fetch(request, { ...request });
        if (!response.ok || !response.body) {
            return new Response('Invalid response', { status: 500 });
        }

        const reader = response.body.pipeThrough(new TextDecoderStream()).getReader();

        let data = [] as string[];
        let i = 0;
        while (true) {
            // Cancel the stream after 5 messages
            if (i > 5) {
                reader.cancel();
                break;
            }
            const { value, done } = await reader.read();

            if (value) {
                console.log(`Got value ${value}`);
                data = [...data, value];
            }

            if (done) {
                break;
            }
            i++;
        }

        return Response.json(data);
    },
```

angelampcosta (approved):

PR #23481, opened 2025-07-07
Add node:http and node:https modules

anonrig:
### Summary

- Adds node:http and node:https modules

Required by https://github.com/cloudflare/workerd/pull/4456

### DIFF
--- a/src/content/docs/workers/runtime-apis/nodejs/http.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/http.mdx
@@ -1,0 +1,121 @@
+---
+pcx_content_type: configuration
+title: http
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## Agent
+
+An Agent manages HTTP connection reuse by maintaining request queues per host/port. In the
+workers environment, however, such low-level management of the network connection, ports,
+etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the
+implementation of `Agent` in Workers is a stub implementation that does not support connection
+pooling or keep-alive.
+
+```js
+import { Agent } from 'node:http';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'http:');
+```
+
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
+
+```js
+import { get } from 'node:http';
+import { strictEqual, ok } from 'node:assert';
+
+get('http://docs.cloudflare.com/robots.txt', (res) => {
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+	let data = '';
+	res.setEncoding('utf8');
+	res.on('data', (chunk) => {
+		data += chunk;
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
+```
+
+## request
+
+The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
+
+```js
+import { request } from 'node:http';
+import { strictEqual } from 'node:assert';
+
+const req = request({
+	method: 'GET',
+	protocol: 'http:',
+	hostname: 'docs.cloudflare.com',
+	path: '/'
+}, (res) => {
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+
+	let data = '';
+	res.setEncoding('utf8');
+	res.on('data', (chunk) => {
+		data += chunk;
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
+req.end();
+
+const req2 = request(new URL('http://docs.cloudflare.com'),{
+	method: 'GET',
+}, (res) => {
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+});
+
+req2.end();
+```
+
+The following options passed to the `request` method are not supported due to the differences in the Cloudflare Workers and the implementation of the node:http module:
+- `maxHeaderSize`
+- `insecureHTTPParser`
+- `createConnection`
+- `lookup`
+- `socketPath`
+
+## OutgoingMessage
+
+The [`OutgoingMessage`](https://nodejs.org/docs/latest/api/http.html#class-httpoutgoingmessage) class represents an HTTP response that is sent to the client. It provides methods for writing response headers and body, as well as for ending the response. `OutgoingMessage` extends from the [`Writable` stream class](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable).
+
+```js
+import { OutgoingMessage } from 'node:http';
+
+const res = new OutgoingMessage();
+res.writeHead(200, { 'Content-Type': 'text/plain' });
+res.write('Hello, World!');
+res.end();
+```
+## IncomingMessage
+
+The `IncomingMessage` class represents an HTTP request that is received from the client. It provides methods for reading request headers and body, as well as for ending the request. `IncomingMessage` extends from the `Readable` stream class.
+
+```js
+import { get, IncomingMessage } from 'node:http';
+import { ok, strictEqual } from 'node:assert';
+
+get('http://docs.cloudflare.com', (res) => {
+	strictEqual(res.statusCode, 301);
+	ok(res instanceof IncomingMessage);
+});
+```
--- a/src/content/docs/workers/runtime-apis/nodejs/http.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/http.mdx
@@ -25,6 +25,8 @@ strictEqual(agent.protocol, 'http:');
 
 ## get
 
+An implementation of the Node.js [`http.get`](https://nodejs.org/docs/latest/api/http.html#httpgetoptions-callback) method.
+
 The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
 
 ```js
--- a/src/content/docs/workers/runtime-apis/nodejs/http.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/http.mdx
@@ -9,7 +9,9 @@ import { Render } from "~/components"
 
 ## Agent
 
-An Agent manages HTTP connection reuse by maintaining request queues per host/port. In the
+An implementation of the Node.js [`http.Agent'](https://nodejs.org/docs/latest/api/http.html#class-httpagent) class.
+
+An [Agent](https://nodejs.org/docs/latest/api/http.html#class-httpagent) manages HTTP connection reuse by maintaining request queues per host/port. In the
 workers environment, however, such low-level management of the network connection, ports,
 etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the
 implementation of `Agent` in Workers is a stub implementation that does not support connection
@@ -42,14 +44,14 @@ get('http://docs.cloudflare.com/robots.txt', (res) => {
 		data += chunk;
 	});
 	res.on('end', () => {
-		// Returns the response body as a string.
-		console.log(data);
 		ok(data.includes('301 Moved Permanently'));
 	});
 });
 ```
 
 ## request
+
+An implementation of the Node.js [`http.request'](https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback) method.
 
 The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
 
@@ -63,7 +65,7 @@ const req = request({
 	hostname: 'docs.cloudflare.com',
 	path: '/'
 }, (res) => {
-	// http://docs.cloudflare.com redirects to https url.
+	// requests to http://docs.cloudflare.com get redirected to their https counterpart.
 	strictEqual(res.statusCode, 301);
 
 	let data = '';
@@ -72,21 +74,21 @@ const req = request({
 		data += chunk;
 	});
 	res.on('end', () => {
-		// Returns the response body as a string.
-		console.log(data);
 		ok(data.includes('301 Moved Permanently'));
 	});
 });
 req.end();
+```
 
-const req2 = request(new URL('http://docs.cloudflare.com'),{
+```js
+const req = request(new URL('http://docs.cloudflare.com'),{
 	method: 'GET',
 }, (res) => {
-	// http://docs.cloudflare.com redirects to https url.
+	// requests to http://docs.cloudflare.com get redirected to their https counterpart.
 	strictEqual(res.statusCode, 301);
 });
 
-req2.end();
+req.end();
 ```
 
 The following options passed to the `request` method are not supported due to the differences in the Cloudflare Workers and the implementation of the `node:http` module:
--- a/src/content/docs/workers/runtime-apis/nodejs/http.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/http.mdx
@@ -87,7 +87,7 @@ const req2 = request(new URL('http://docs.cloudflare.com'),{
 req2.end();
 ```
 
-The following options passed to the `request` method are not supported due to the differences in the Cloudflare Workers and the implementation of the node:http module:
+The following options passed to the `request` method are not supported due to the differences in the Cloudflare Workers and the implementation of the `node:http` module:
 - `maxHeaderSize`
 - `insecureHTTPParser`
 - `createConnection`
--- a/src/content/docs/workers/runtime-apis/nodejs/http.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/http.mdx
@@ -32,7 +32,7 @@ import { get } from 'node:http';
 import { strictEqual, ok } from 'node:assert';
 
 get('http://docs.cloudflare.com/robots.txt', (res) => {
-	// http://docs.cloudflare.com redirects to https url.
+	// requests to http://docs.cloudflare.com get redirected to their https counterpart.
 	strictEqual(res.statusCode, 301);
 	let data = '';
 	res.setEncoding('utf8');
--- a/src/content/docs/workers/runtime-apis/nodejs/https.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/https.mdx
@@ -25,7 +25,7 @@ strictEqual(agent.protocol, 'https:');
 
 ## get
 
-The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
+The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
 
 `get` methods behaves the same as `node:http` get method.
 
--- a/src/content/docs/workers/runtime-apis/nodejs/https.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/https.mdx
@@ -1,0 +1,82 @@
+---
+pcx_content_type: configuration
+title: https
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## Agent
+
+An Agent manages HTTPS connection reuse by maintaining request queues per host/port. In the
+workers environment, however, such low-level management of the network connection, ports,
+etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the
+implementation of `Agent` in Workers is a stub implementation that does not support connection
+pooling or keep-alive.
+
+```js
+import { Agent } from 'node:https';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'https:');
+```
+
+## get
+
+The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
+
+`get` methods behaves the same as `node:http` get method.
+
+```js
+import { get } from 'node:https';
+import { strictEqual, ok } from 'node:assert';
+
+get('https://developers.cloudflare.com/robots.txt', (res) => {
+	strictEqual(res.statusCode, 200);
+	let data = '';
+	res.setEncoding('utf8');
+	res.on('data', (chunk) => {
+		data += chunk;
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('User-agent'));
+	});
+});
+```
+
+## request
+
+The [`request`](https://nodejs.org/docs/latest/api/https.html#httpsrequesturl-options-callback) method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
+
+Request method accepts all options from `http.request` with some differences in default values:
+- `protocol`: default `https:`
+- `port`: default `443`
+- `agent`: default `https.globalAgent`
+
+The following additional options are not supported: ca, cert, ciphers, clientCertEngine (deprecated), crl, dhparam, ecdhCurve, honorCipherOrder, key, passphrase, pfx, rejectUnauthorized, secureOptions, secureProtocol, servername, sessionIdContext, highWaterMark.
+
+```js
+import { request } from 'node:https';
+import { strictEqual, ok } from 'node:assert';
+
+const req = request('https://developers.cloudflare.com/robots.txt', {
+	method: 'GET',
+}, (res) => {
+	strictEqual(res.statusCode, 200);
+	let data = '';
+	res.setEncoding('utf8');
+	res.on('data', (chunk) => {
+		data += chunk;
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('User-agent'));
+	});
+});
+req.end();
+```
--- a/src/content/docs/workers/runtime-apis/nodejs/https.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/https.mdx
@@ -3,7 +3,7 @@ pcx_content_type: configuration
 title: https
 ---
 
-import { Render } from "~/components"
+import { Render } from "~/components";
 
 <Render file="nodejs-compat-howto" />
 
@@ -18,11 +18,11 @@ implementation of `Agent` in Workers is a stub implementation that does not supp
 pooling or keep-alive.
 
 ```js
-import { Agent } from 'node:https';
-import { strictEqual } from 'node:assert';
+import { Agent } from "node:https";
+import { strictEqual } from "node:assert";
 
 const agent = new Agent();
-strictEqual(agent.protocol, 'https:');
+strictEqual(agent.protocol, "https:");
 ```
 
 ## get
@@ -32,20 +32,25 @@ An implementation of the Node.js [`https.get'](https://nodejs.org/docs/latest/ap
 The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
 
 ```js
-import { get } from 'node:https';
-import { strictEqual, ok } from 'node:assert';
+import { get } from "node:https";
 
-get('https://developers.cloudflare.com/robots.txt', (res) => {
-	strictEqual(res.statusCode, 200);
-	let data = '';
-	res.setEncoding('utf8');
-	res.on('data', (chunk) => {
-		data += chunk;
-	});
-	res.on('end', () => {
-		ok(data.includes('User-agent'));
-	});
-});
+export default {
+	async fetch() {
+		const { promise, resolve, reject } = Promise.withResolvers();
+		get("http://example.com", (res) => {
+			let data = "";
+			res.setEncoding("utf8");
+			res.on("data", (chunk) => {
+				data += chunk;
+			});
+			res.on("error", reject);
+			res.on("end", () => {
+				resolve(new Response(data));
+			});
+		}).on("error", reject);
+		return promise;
+	},
+};
 ```
 
 ## request
@@ -55,27 +60,32 @@ An implementation of the Node.js [`https.request'](https://nodejs.org/docs/lates
 The [`request`](https://nodejs.org/docs/latest/api/https.html#httpsrequestoptions-callback) method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://developers.cloudflare.com/workers/runtime-apis/streams/writablestream/) for sending request data.
 
 Request method accepts all options from `http.request` with some differences in default values:
+
 - `protocol`: default `https:`
 - `port`: default `443`
 - `agent`: default `https.globalAgent`
 
 ```js
-import { request } from 'node:https';
-import { strictEqual, ok } from 'node:assert';
+import { request } from "node:https";
+import { strictEqual, ok } from "node:assert";
 
-const req = request('https://developers.cloudflare.com/robots.txt', {
-	method: 'GET',
-}, (res) => {
-	strictEqual(res.statusCode, 200);
-	let data = '';
-	res.setEncoding('utf8');
-	res.on('data', (chunk) => {
-		data += chunk;
-	});
-	res.on('end', () => {
-		ok(data.includes('User-agent'));
-	});
-});
+const req = request(
+	"https://developers.cloudflare.com/robots.txt",
+	{
+		method: "GET",
+	},
+	(res) => {
+		strictEqual(res.statusCode, 200);
+		let data = "";
+		res.setEncoding("utf8");
+		res.on("data", (chunk) => {
+			data += chunk;
+		});
+		res.on("end", () => {
+			ok(data.includes("User-agent"));
+		});
+	},
+);
 req.end();
 ```
 
--- a/src/content/docs/workers/runtime-apis/nodejs/https.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/https.mdx
@@ -9,7 +9,9 @@ import { Render } from "~/components"
 
 ## Agent
 
-An Agent manages HTTPS connection reuse by maintaining request queues per host/port. In the
+An implementation of the Node.js [`https.Agent'](https://nodejs.org/docs/latest/api/https.html#class-httpsagent) class.
+
+An [Agent](https://nodejs.org/docs/latest/api/https.html#class-httpsagent) manages HTTPS connection reuse by maintaining request queues per host/port. In the
 workers environment, however, such low-level management of the network connection, ports,
 etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the
 implementation of `Agent` in Workers is a stub implementation that does not support connection
@@ -25,9 +27,9 @@ strictEqual(agent.protocol, 'https:');
 
 ## get
 
-The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
+An implementation of the Node.js [`https.get'](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method.
 
-`get` methods behaves the same as `node:http` get method.
+The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
 
 ```js
 import { get } from 'node:https';
@@ -41,8 +43,6 @@ get('https://developers.cloudflare.com/robots.txt', (res) => {
 		data += chunk;
 	});
 	res.on('end', () => {
-		// Returns the response body as a string.
-		console.log(data);
 		ok(data.includes('User-agent'));
 	});
 });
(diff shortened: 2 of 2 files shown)


vicb:
Thanks @anonrig Could you please add details about `Agent`s?

anonrig:
> Thanks @anonrig Could you please add details about `Agent`s?

@vicb Agent acts as a stub. It's not intended for usage, and not yet fully implemented.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+title: http
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## ClientRequest
+
+The `ClientRequest` class represents an outgoing HTTP request. It implements a writable stream interface, allowing you to send request data and provides events for monitoring the request lifecycle.
```
jasnell: `writable stream interface` could probably be a link to the Node.js docs so it's clearer that we mean it's a `Node.js` `stream.Writable` and not a `WritableStream`.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
+
+```js
+import { get } from 'node:http';
+
+get(new URL('https://docs.cloudflare.com/robots.txt'), (res) => {
+	assert.strictEqual(res.statusCode, 200);
+});
```
jasnell, line 49: I would show this example with the url passed as a string also, since that is by far more common in the ecosystem.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+ws.on('message', (data) => {
+	console.log('Received message:', data);
+});
+
+ws.on('close', () => {
+	console.log('WebSocket connection closed');
+});
+```
```
jasnell: It should be documented whether this is exactly the same `WebSocket` class used elsewhere in our API or is this specifically modeled after Node.js' implementation? There are distinct differences so we should be clear.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+## OutgoingMessage
+
+The `OutgoingMessage` class represents an HTTP response that is sent to the client. It provides methods for writing response headers and body, as well as for ending the response.
```
jasnell: Worth mentioning that this extends from `stream.Writable`

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+The `ClientRequest` class represents an outgoing HTTP request. It implements a writable stream interface, allowing you to send request data and provides events for monitoring the request lifecycle.
+
+```js
+import { ClientRequest } from 'node:http';
+
+const request = new ClientRequest({
+	method: 'GET',
+	protocol: 'http:',
+	hostname: 'example.com'
+});
```
jasnell: In Node.js' docs, we do not document the `ClientRequest` constructor and I'm not sure there's much reason to actually do so. If we are going to here, however, we should document all of the available `options`.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+request.on('response', (response) => {
+	assert.strictEqual(response.statusCode, 200);
+});
+
+request.end();
+``
+
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
```
jasnell, line 32: This should include coverage of the `options` that are supported/not-supported. Both here and with `request` below.

src/content/notifications/index.yaml
```diff
@@ -625,4 +625,4 @@ entries:
     availability: Purchase of Cloudflare Network Interconnect (CNI).
     associatedProducts: Network Interconnect
     nextSteps: No action is needed.
-    otherFilters: None.
\ No newline at end of file
+    otherFilters: None.
```
jasnell: These appear to be unrelated to the actual change?

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
+
+```js
+import { get } from 'node:https';
+
+get(new URL('https://docs.cloudflare.com/robots.txt'), (res) => {
+	assert.strictEqual(res.statusCode, 200);
+});
+```
```
jasnell, line 26: Feels a bit odd to duplicate the docs on these. If we keep these here then the support/unsupported `options` should be documented.

jasnell:
> ... It's not intended for usage, and not yet fully implemented.

That should be included in the documentation.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,87 @@
...
+The `OutgoingMessage` class represents an HTTP response that is sent to the client. It provides methods for writing response headers and body, as well as for ending the response.
+
+```js
+import { OutgoingMessage } from 'node:http';
+
+const res = new OutgoingMessage();
+res.writeHead(200, { 'Content-Type': 'text/plain' });
+res.write('Hello, World!');
+res.end();
+```
```
vicb, line 112: Add `IncomingMessage`
ref: https://github.com/cloudflare/workerd/pull/4535

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,33 @@
...
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
```
vicb:
```suggestion
The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,33 @@
...
+import { get } from 'node:https';
+
+get(new URL('https://docs.cloudflare.com/robots.txt'), (res) => {
+	assert.strictEqual(res.statusCode, 200);
+});
+```
+
+## request
+
+The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a writable stream for sending request data.
```
vicb:
```suggestion
The `request` method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a writable stream for sending request data.
```

vicb:
Could you please add docs for the compat flag in this PR? (in src/content/compatibility-flags)
Also ref the compat flag from the module docs
Thanks!

ref: https://github.com/cloudflare/workerd/pull/4456

src/content/notifications/index.yaml
anonrig: This is in a different commit

package.json
danlapid, line 1: Are the changes here strictly necessary? why? let's revert if not

src/content/apps/index.yaml
danlapid, line 1: revert change please

src/content/notifications/index.yaml
danlapid, line 1: Revert change please

src/content/glossary/d1.yaml
danlapid, line 1: Revert change please

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,117 @@
...
+title: http
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## Agent
+
+An Agent manages HTTP connection reuse by maintaining request queues per host/port. The implementation is a stub implementation that does not support connection pooling or keep-alive.
```
jasnell: I've been wanting to start improving our docs around node.js compat to provide additional context and this is a good place to start doing so... for something like Agent, where we know why we aren't really going to support it, we should take a bit more time to explain why.

```suggestion
An Agent manages HTTP connection reuse by maintaining request queues per host/port. In the
workers environment, however, such low-level management of the network connection, ports,
etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the
implementation of `Agent` in Workers is a stub implementation that does not support connection
pooling or keep-alive.
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,117 @@
...
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
+
+```js
+import { get } from 'node:http';
+import { strictEqual } from 'node:assert/strict';
+
+get('http://docs.cloudflare.com/robots.txt', (res) => {
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+});
```
jasnell, line 49: Can be left for later but a more complete example that shows the `res` payload being consumed would be helpful. Each of these sections could probably also benefit from including links to the relevant parts of the node.js docs

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,117 @@
...
+import { Agent } from 'node:http';
+import { strictEqual } from 'node:assert/strict';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'http:');
+```
+
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
```
jasnell, line 32: Given that `get` is documented before `request` it's a bit confusing here to call it a convenience method without a forward link to `request` or a specific callout that it's a convenience method **for request**.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,117 @@
...
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+});
+
+request(new URL('http://docs.cloudflare.com'),{
+	method: 'GET',
+}, (res) => {
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+});
```
jasnell, line 89: Likewise here, an example showing the `resp` being fully consumed would be helpful.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,117 @@
...
+```
+
+The following options passed to the `request` method are not supported:
+- `maxHeaderSize`
+- `insecureHTTPParser`
+- `createConnection`
+- `lookup`
+- `socketPath`
```
jasnell, line 99: It would be helpful to briefly explain why they are not supported.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+## WebSocket
+
+The `WebSocket` class provides a convenient way to establish a WebSocket connection to a server. It simplifies the process of creating and managing WebSocket connections.
+
+This API is exactly modeled after Node.js' [`WebSocket`](https://nodejs.org/docs/latest/api/globals.html#class-websocket) class.
```
jasnell: Because there real differences between this `WebSocket` and the `globalThis.WebSocket`, I think the docs here should go into a bit more detail on what the differences are and why someone would/should use this one instead of `globalThis.WebSocket`... otherwise we're inviting confusion

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,120 @@
...
+ws.on('message', (data) => {
+	console.log('Received message:', data);
+});
+
+ws.on('close', () => {
+	console.log('WebSocket connection closed');
+});
+```
+
+## OutgoingMessage
```
jasnell, line 101: Linking to the relevant parts of the node.js docs in these sections would be helpful.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
+
+`get` methods behaves the same as `node:http` get method.
+
+## request
+
+The `request` method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
+
+`request` methods behaves the same as `node:http` request method.
```
jasnell: The `https` versions of these docs feel very incomplete. There should be examples, some discussion of what options are supported (especially given that there are a range of options specific to `https` requests that aren't used in `http` that we know we don't (and won't) support.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'https:');
+```
+
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
+
+`get` methods behaves the same as `node:http` get method.
```
jasnell: This is `https.get` but you're referring to `node:http` `get`. We also know there are behavioral differences between the two since we don't support the full range of TLS options, events, etc. So I think this comment is rather misleading and incomplete.

jasnell (commented):
A few nits and suggestions.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: I was referring to the `request` options (not the request method options). I'm not sure how to make it clear. Maybe `configuring method options` is better wording?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: You mean one by one or in general?

danlapid:
Can we have a general comment on both pages saying that in terms of node:http and node:https we believe the client side usecases should be covered by the current implementation and that the server side implementation is Work In Progress.

Also, Do those pages actually cover everything we support? Any other API in node:http/s that we support but isn't in those pages?

No need to have a full on example for everything but if someone needs some API and they search our docs for it, will they find it or possibly not?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: In general

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,76 @@
...
+	method: 'GET',
+}, (res) => {
+	strictEqual(res.statusCode, 200);
+	let data = '';
+	res.on('data', (chunk) => {
+		data += chunk.toString();
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
```
jasnell: `data` would be expected to be a `Buffer` here unless `res.setEncoding('utf8')` is called.

jasnell: oh, actually, nevermind lol... I misread the example lol

anonrig: `chunk.toString();` will convert it into utf8 since the default value for buffer toString is utf8

jasnell: but if you did set `res.setEncoding('ut8')` then you wouldn't need the `toString()` and it would correctly handle the case of split multi-byte sequences. As it is here there's likely bugs if the chunks are not complete utf8

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: I removed this for now.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
anonrig: Yeah, this example code works, but no harm in calling setEncoding for sure.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,76 @@
...
+import { strictEqual, ok } from 'node:assert/strict';
+
+const req = request('https://developers.cloudflare.com/robots.txt', {
+	method: 'GET',
+}, (res) => {
+	strictEqual(res.statusCode, 200);
+	let data = '';
+	res.on('data', (chunk) => {
+		data += chunk.toString();
+	});
```
jasnell:
```suggestion
  res.setEncoding('utf8');
	res.on('data', (chunk) => {
		data += chunk;
	});
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,76 @@
...
+```js
+import { get } from 'node:https';
+import { strictEqual, ok } from 'node:assert/strict';
+
+get('https://developers.cloudflare.com/robots.txt', (res) => {
+	strictEqual(res.statusCode, 200);
+	let data = '';
+	res.on('data', (chunk) => {
+		data += chunk.toString();
+	});
```
jasnell:
```suggestion
  res.setEncoding('utf8');
	res.on('data', (chunk) => {
		data += chunk;
	});
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
jasnell: It works only because the returned result does not happen to end up spliting a multi-byte utf8 sequence on the boundary between `data` event calls.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,78 @@
...
+title: https
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## Agent
+
+An Agent manages HTTPS connection reuse by maintaining request queues per host/port. The implementation is a stub implementation that does not support connection pooling or keep-alive.
```
jasnell: The same change made to the `Agent` description in `http.mdx` would be relevant here.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+```js
+import { Agent } from 'node:https';
+import { strictEqual } from 'node:assert/strict';
```
jasnell: Our impl of `node:assert` is always strict mode.

```suggestion
import { strictEqual } from 'node:assert';
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+res.write('Hello, World!');
+res.end();
+```
+## IncomingMessage
+
+The `IncomingMessage` class represents an HTTP request that is received from the client. It provides methods for reading request headers and body, as well as for ending the request. `IncomingMessage` extends from the `Readable` stream class.
+
+```js
+import { get, IncomingMessage } from 'node:http';
+import { ok, strictEqual } from 'node:assert/strict';
```
jasnell:
```suggestion
import { ok, strictEqual } from 'node:assert';
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the
+implementation of `Agent` in Workers is a stub implementation that does not support connection
+pooling or keep-alive.
+
+```js
+import { Agent } from 'node:http';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'http:');
```
jasnell, line 25: Nit: The example would likely be more useful if it showed where `agent` is actually used as opposed to just asserting a property on it.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+	res.setEncoding('utf8');
+	res.on('data', (chunk) => {
+		data += chunk;
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
```
jasnell, line 49: Just thinking that it might be useful to show a more complete example that shows how to use the api within the context of a worker request. Specifically, something like...

```js
export default {
  async fetch(req) {
    const { promise, resolve, reject } = Promise.withResolvers();
    const req = get('...', (res) => {
     res.on('error', reject);
     res.on('data', ...);
     res.on('end', () => {
       resolve(....);
     });
    });
    req.on('error', reject);
    return promise;
  }
}
```

etc.

the reason is that it's going to be a bit tricky for users to correctly adapt the node.js callback/event-emitter pattern used by this API to the promise-based API used in fetch, and I foresee a number of bugs that users will accidentally introduce into their applications because they fail to correctly wire things up. 

This can be tackled in an iteration of this PR tho.

jasnell: Oh, it should probably also be highlighted that these, like fetch, can only be done within a request, will be canceled when the request completes, and count towards the concurrent request limits.... all things that should generally be obvious but are still well worth calling out.

jasnell (approved):
Few remaining nits but also happy to make improvements incrementally

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+```
+
+## request
+
+The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
```
mikenomitch, line 56: Should we link to our WritableStream docs here instead? - https://developers.cloudflare.com/workers/runtime-apis/streams/writablestream/

mikenomitch: Ditto for https

src/content/docs/workers/runtime-apis/nodejs/http.mdx
dario-piotrowicz: Sorry but I don't find this wording clear at all 😕 

Could the second sentence just be removed? what is it trying to convey?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
+
+```js
+import { get } from 'node:http';
+import { strictEqual, ok } from 'node:assert';
+
+get('http://docs.cloudflare.com/robots.txt', (res) => {
+	// http://docs.cloudflare.com redirects to https url.
```
dario-piotrowicz:
```suggestion
	// requests to http://docs.cloudflare.com get redirected to their https counterpart.
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
+import { Agent } from 'node:https';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'https:');
+```
+
+## get
+
+The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
```
mikenomitch:
```suggestion
The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
```

Nitpick

src/content/docs/workers/runtime-apis/nodejs/http.mdx
dario-piotrowicz: or something like that, it took me a sec to get the meaning of the comment 😅

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+	strictEqual(res.statusCode, 301);
+	let data = '';
+	res.setEncoding('utf8');
+	res.on('data', (chunk) => {
+		data += chunk;
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
```
dario-piotrowicz: is the `console.log` intentional? 🤔

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+import { request } from 'node:http';
+import { strictEqual } from 'node:assert';
+
+const req = request({
+	method: 'GET',
+	protocol: 'http:',
+	hostname: 'docs.cloudflare.com',
+	path: '/'
+}, (res) => {
+	// http://docs.cloudflare.com redirects to https url.
```
dario-piotrowicz: as before I think that maybe this could be clarified a bit

src/content/docs/workers/runtime-apis/nodejs/https.mdx
mikenomitch: Ditto on `http`

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
+req.end();
+
+const req2 = request(new URL('http://docs.cloudflare.com'),{
+	method: 'GET',
+}, (res) => {
+	// http://docs.cloudflare.com redirects to https url.
```
dario-piotrowicz: as before 🙂

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
+req.end();
+
+const req2 = request(new URL('http://docs.cloudflare.com'),{
```
dario-piotrowicz: instead of having `req2` I would split these two in two different js blocks and always use `req`

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+	method: 'GET',
+}, (res) => {
+	// http://docs.cloudflare.com redirects to https url.
+	strictEqual(res.statusCode, 301);
+});
+
+req2.end();
+```
+
+The following options passed to the `request` method are not supported due to the differences in the Cloudflare Workers and the implementation of the node:http module:
```
dario-piotrowicz:
```suggestion
The following options passed to the `request` method are not supported due to the differences in the Cloudflare Workers and the implementation of the `node:http` module:
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
dario-piotrowicz: Also maybe it'd be better to say that the `get` here behaves the same at the _our_ http `get` (the one you are documenting in the other file in this PR) and provide a link to it?

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
+## request
+
+The [`request`](https://nodejs.org/docs/latest/api/https.html#httpsrequesturl-options-callback) method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
+
+Request method accepts all options from `http.request` with some differences in default values:
+- `protocol`: default `https:`
+- `port`: default `443`
+- `agent`: default `https.globalAgent`
+
+The following additional options are not supported: ca, cert, ciphers, clientCertEngine (deprecated), crl, dhparam, ecdhCurve, honorCipherOrder, key, passphrase, pfx, rejectUnauthorized, secureOptions, secureProtocol, servername, sessionIdContext, highWaterMark.
```
mikenomitch: Nit: I'd move this below the code example and then list this in bullet points

src/content/docs/workers/runtime-apis/nodejs/http.mdx
dario-piotrowicz, line 56: should `request` link to https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback?

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
```
dario-piotrowicz:
```suggestion
The following additional options are not supported: `ca`, `cert`, `ciphers`, `clientCertEngine` (deprecated), `crl`, `dhparam`, `ecdhCurve`, `honorCipherOrder`, `key`, `passphrase`, `pfx`, `rejectUnauthorized`, `secureOptions`, `secureProtocol`, `servername`, `sessionIdContext`, `highWaterMark`.
```

dario-piotrowicz (approved):
Looks good to me, I just left a few small comments 🙂

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: Hmm, looks like the example wasn't change but the comment was marked resolved? This example, as is, is not very useful, so I'd recommend removing it or showing an example that actually passes the non-op agent into a request. It would also be good here to have a link to the relevant Node.js docs

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+implementation of `Agent` in Workers is a stub implementation that does not support connection
+pooling or keep-alive.
+
+```js
+import { Agent } from 'node:http';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'http:');
+```
```
jasnell, line 26: I had a comment here about the example that was marked resolved but the example wasn't actually changed. This example, as is, is not very useful, so I'd recommend removing it or showing an example that actually passes the non-op agent into a request. It would also be good here to have a link to the relevant Node.js docs

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: Yeah, good call. Those docs need to be improved tho too, but linking to our docs makes the most sense here.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+	});
+	res.on('end', () => {
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
+```
+
+## request
```
jasnell, line 52: I had a comment somewhere that appears to have been marked resolved about how these should include links to the relevant node.js docs for these methods. I still think we should include those links.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'https:');
+```
+
+## get
+
+The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
+
+`get` methods behaves the same as `node:http` get method.
```
jasnell: My previous comment here was marked resolved but I'm not seeing a difference here. The way this is worded is confusing. It's also not entirely correct, is it? `https.get(...)` for instance should require `https://...` URLs while `http.get(...)` should forbid them. Also, it's not clear if this is saying that it behaves the same as Node.js `node:http` get or our `node:http` get. As is this is fairly confusing as worded

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: Your previous comment started with a nit.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: You mean the header should be a link to node.js docs? Can you be specific?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: Yep, and it's ok not to address it in this PR, but the comment being resolved and generally disappearing makes it less likely to be addressed even in a follow up. I'm as guilty of that as anyone in these doc changes. I just want to make sure some of these things don't get lost and forgotten.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+import { Agent } from 'node:http';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'http:');
+```
+
+## get
+
+The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
```
jasnell, line 32:
```suggestion
An implementation of the Node.js [`http.get`](https://nodejs.org/docs/latest/api/http.html#httpgetoptions-callback) method.

The `get` method performs a GET request to the specified URL and invokes the callback with the response. It's a convenience method that simplifies making HTTP GET requests without manually configuring request options.
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('301 Moved Permanently'));
+	});
+});
+```
+
+## request
+
+The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
```
jasnell, line 56:
```suggestion
An implementation of the Node.js [`http.request'](https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback) method.

The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -0,0 +1,121 @@
...
+title: http
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## Agent
+
+An Agent manages HTTP connection reuse by maintaining request queues per host/port. In the
```
jasnell:
```suggestion
An [Agent](https://nodejs.org/docs/latest/api/http.html#class-httpagent) manages HTTP connection reuse by maintaining request queues per host/port. In the
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
anonrig: I forgot to remove this apparently. I'll remove this line directly.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
+title: https
+---
+
+import { Render } from "~/components"
+
+<Render file="nodejs-compat-howto" />
+
+## Agent
+
+An Agent manages HTTPS connection reuse by maintaining request queues per host/port. In the
```
jasnell:
```suggestion
An [Agent](https://nodejs.org/docs/latest/api/https.html#class-httpsagent) manages HTTPS connection reuse by maintaining request queues per host/port. In the
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
+import { Agent } from 'node:https';
+import { strictEqual } from 'node:assert';
+
+const agent = new Agent();
+strictEqual(agent.protocol, 'https:');
+```
+
+## get
+
+The [`get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
```
jasnell:
```suggestion
An implementation of the Node.js [`https.get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method.

The `get` method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -0,0 +1,82 @@
...
+		// Returns the response body as a string.
+		console.log(data);
+		ok(data.includes('User-agent'));
+	});
+});
+```
+
+## request
+
+The [`request`](https://nodejs.org/docs/latest/api/https.html#httpsrequesturl-options-callback) method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
```
jasnell:
```suggestion
An implementation of the Node.js [`https.request`](https://nodejs.org/docs/latest/api/https.html#httpsrequestoptions-callback) method.

The `request` method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: Yes, we can address it in a future pull-request. Specifically when I add the server-side methods, I'll revisit this.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: I don't think this is necessary since it already links the request (in the newest commit)

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: Sounds good to me

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: Updated the pull-request to use a similar code.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: I'll resolve this now. I'd like to land this.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: @jasnell is the current state sufficient?

jasnell (approved):
We can make further improvements in follow up

PR #23558, opened 2025-07-10
[Durable Objects] Explain that ctx.id.name is undefined within Durable Object constructor

lukevalenta:


src/content/docs/durable-objects/api/id.mdx
```diff
@@ -57,7 +57,7 @@ console.assert(!id1.equals(id2), "Different unique ids should never be equal.");
 
 ### `name`
 
-`name` is an optional property of a `DurableObjectId`, which returns the name that was used to create the `DurableObjectId` via [`DurableObjectNamespace::idFromName`](/durable-objects/api/namespace/#idfromname). This value is undefined if the `DurableObjectId` was constructed using [`DurableObjectNamespace::newUniqueId`](/durable-objects/api/namespace/#newuniqueid).
+`name` is an optional property of a `DurableObjectId`, which returns the name that was used to create the `DurableObjectId` via [`DurableObjectNamespace::idFromName`](/durable-objects/api/namespace/#idfromname). This value is undefined if the `DurableObjectId` was constructed using [`DurableObjectNamespace::newUniqueId`](/durable-objects/api/namespace/#newuniqueid). This value is also undefined within the `ctx.id` passed into the Durable Object constructor (see [GitHub issue](https://github.com/cloudflare/workerd/issues/2240) for details).
```
Oxyjun:
```suggestion
`name` is an optional property of a `DurableObjectId`, which returns the name that was used to create the `DurableObjectId` via [`DurableObjectNamespace::idFromName`](/durable-objects/api/namespace/#idfromname). This value is undefined if the `DurableObjectId` was constructed using [`DurableObjectNamespace::newUniqueId`](/durable-objects/api/namespace/#newuniqueid). This value is also undefined within the `ctx.id` passed into the Durable Object constructor (refer to [GitHub issue](https://github.com/cloudflare/workerd/issues/2240) for discussion).
```

Can we capture the information directly in the docs, instead of linking to an open issue? I feel that would be most useful to the users.

Happy to work together to capture this info

lukevalenta: Happy to try to add the information in-line, but some benefits of linking to the open issue are:

1) If the issue is ever resolved but no-one updates these docs, users have a way to discover that by clicking the link (and hopefully then submit a docs PR). The Workers team has indicated that they'd _like_ to solve this issue, but just don't have a great solution yet. I personally found the discussion very useful for _why_ this seemingly simple feature is not supported.
2) There are several workarounds discussed in the GitHub issue that might be helpful to users. For example, one can pass in the Durable Object name as a request parameter, or for cases where all possible DO names are known, within the DO constructor one can iterate over the possible values and generate ObjectIds until one matches the current DO. Discussing these workarounds here in the docs (especially with code samples) might get quite verbose.

What do you think?

lukevalenta: Just bumping this PR (cc @Oxyjun)

Oxyjun: Thanks for the bump! And understood re: link. As a nit, I'd like to refer to this as a "discussion" rather than "details" - will edit and get it merged. Thank you!

Oxyjun (approved):

PR #22141, opened 2025-05-01
Fix agent code example

SamyPesse:
### Summary

Fixes a small error in the code example for agents.

PR #23647, opened 2025-07-14
Fix calling-agents.mdx code example

ertembiyik:
### Summary

getState was called globally, which is not right and may confuse readers. Fixed it by calling that method on namedAgent instance.


Oxyjun:
Related to https://github.com/cloudflare/cloudflare-docs/pull/22141

threepointone:
we have a full rewrite landing that's taken care of this

PR #23661, opened 2025-07-15
clarify Simultaneous open connections for workers

vicb:
### Summary

It could be surprising for users to get a "Response closed due to connection limit" exception while the docs says that requests beside the max 6 connections are queue.

This PR adds some details and the exception message.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/platform/limits.mdx
+++ b/src/content/docs/workers/platform/limits.mdx
@@ -321,7 +321,7 @@ To reduce the upload size of a Worker, consider some of the following strategies
 
 - Removing unnecessary dependencies and packages
 - Storing configuration files, static assets, and binary data using [Workers KV](/kv/), [R2](/r2/), [D1](/d1/), or [Workers Static Assets](/workers/static-assets/) instead of bundling them within your Worker code.
-- Splitng functionality across multiple Workers and connecting them using [Service bindings](/workers/runtime-apis/bindings/service-bindings/).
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/platform/limits.mdx
```diff
@@ -233,6 +233,30 @@ Once an invocation has six connections open, it can still attempt to open additi
 
 - These attempts are put in a pending queue — the connections will not be initiated until one of the currently open connections has closed.
 - Earlier connections can delay later ones, if a Worker tries to make many simultaneous subrequests, its later subrequests may appear to take longer to start.
+- Earlier connections that are stalled<sup>1</sup> might get closed with a `Response closed due to connection limit` exception.
+
+<sup>1</sup>A connections is considered stalled when it is not not being actively read from or written to, for example:
```
harrishancock: Later on in this file, a few paragraphs down, we have:

```
If the system detects that a Worker is deadlocked on open connections — for example, if the Worker has pending connection attempts but has no in-progress reads or writes on the connections that it already has open — then the least-recently-used open connection will be canceled to unblock the Worker.
```

This is describing the same thing. I think both descriptions are useful, as it hammers the point home. Perhaps we can modify the blurb below -- that is, the one I quoted -- to include a callback to this definition of stalled?

harrishancock (approved):

jasnell (approved):

vicb:
Thanks James and Harris 🙏

penalosa (approved):

PR #23705, opened 2025-07-15
[ZT] update Access wildcard example

ranbel:
PCX-18359

### DIFF
--- a/src/content/docs/cloudflare-one/policies/access/app-paths.mdx
+++ b/src/content/docs/cloudflare-one/policies/access/app-paths.mdx
@@ -22,11 +22,11 @@ When you create an application for a specific subdomain or path, you can use ast
 
 #### Match all subdomains of an apex domain
 
-Using a wildcard in the **Subdomain** field does not cover the apex domain.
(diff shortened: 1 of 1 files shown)


kennyj42 (approved):

kennyj42:
only thought is that maybe we should say that "if you want to cover multiple subdomain levels, you can use multiple wildcards"?

jroyal (approved):

src/content/docs/cloudflare-one/policies/access/app-paths.mdx
```diff
@@ -22,11 +22,11 @@ When you create an application for a specific subdomain or path, you can use ast
 
 #### Match all subdomains of an apex domain
 
-Using a wildcard in the **Subdomain** field does not cover the apex domain.
+A wildcard in the **Subdomain** field only matches that specific subdomain level. It does not cover the apex domain or multiple levels of the subdomain.
```
ranbel:
```suggestion
A wildcard in the **Subdomain** field only matches that specific subdomain level. It does not cover the apex domain or multiple levels of the subdomain. If you want to cover multiple subdomain levels, you can use multiple wildcards.
```

kodster28 (approved):

PR #23716, opened 2025-07-16
[doc] Python worker doc dict fix

sjsadowski:
### Summary

Python worker examples used dot notation for json objects converted to dicts, which doesn't work out of the box. In order to keep the examples simple, changed dot notation to key notation. Also removed string concatenation using plus signs and string.format() to both be f-strings as the current releases of pyodide all target support for python > 3.6

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/languages/python/examples.mdx
+++ b/src/content/docs/workers/languages/python/examples.mdx
@@ -24,7 +24,8 @@ async def on_fetch(request, env):
     params = parse_qs(url.query)
 
     if "name" in params:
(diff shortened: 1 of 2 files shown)


sjsadowski:
closing this PR, may reopen - I have now been using python workers in two different setups and I'm getting inconsistent results with the dot notation; on one machine running under wrangler dev (await request.json)["key"] works and on another it does not, but the documented (await request.json()).key works. Will open a new request once this is resolved with the correct updates.

PR #23757, opened 2025-07-17
[doc] Python Worker handler/dict fixes

sjsadowski:
### Summary
Confirmed with @dom96 that the examples should use the `@handler` wrapper. 

Python worker examples used dot notation for json objects converted to dicts, which doesn't work out of the box. In order to keep the examples simple, changed dot notation to key notation. Also removed string concatenation using plus signs and string.format() to both be f-strings as the current releases of pyodide all target support for python > 3.6

Updated the queue example to remove the unused import, and alter the example to be more pythonic with json.dumps() 

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/languages/python/examples.mdx
+++ b/src/content/docs/workers/languages/python/examples.mdx
@@ -24,7 +24,8 @@ async def on_fetch(request, env):
     params = parse_qs(url.query)
 
     if "name" in params:
-        greeting = "Hello there, {name}".format(name=params["name"][0])
+				name = params["name"][0]
+        greeting = f"Hello there, {name}"
         return Response(greeting)
 
 
@@ -40,8 +41,8 @@ async def on_fetch(request, env):
 from workers import Response
 
 async def on_fetch(request):
-    name = (await request.json()).name
-    return Response("Hello, {name}".format(name=name))
+    name = (await request.json())["name"]
+    return Response(f"Hello, {name}")
 ```
 
 ## Emit logs from your Python Worker
--- a/src/content/docs/workers/languages/python/examples.mdx
+++ b/src/content/docs/workers/languages/python/examples.mdx
@@ -14,9 +14,10 @@ In addition to those examples, consider the following ones that illustrate Pytho
 ## Parse an incoming request URL
 
 ```python
-from workers import Response
+from workers import handler, Response
 from urllib.parse import urlparse, parse_qs
 
+@handler
 async def on_fetch(request, env):
     # Parse the incoming request URL
     url = urlparse(request.url)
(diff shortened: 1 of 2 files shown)


src/content/docs/workers/languages/python/examples.mdx
```diff
@@ -93,7 +94,7 @@ async def on_fetch(request, env):
     # We can also pass plain text strings
     await env.QUEUE.send("hello", contentType="text")
     # Send a JSON payload
-    await env.QUEUE.send(to_js({"hello": "world"}))
+    await env.QUEUE.send(json.dumps({"hello": "world"}))
```
dom96, line 97: Have you tested that this works? I think we need the `to_js` here.

dom96 (commented):
Thank you for putting this together. Just one thing that I think needs to be changed, the rest looks good.

sjsadowski:
@dom96 I think it should actually be ok; looking at the docs for the worker binding here: https://developers.cloudflare.com/queues/configuration/javascript-apis/#queue 
> Sends a message to the Queue. The body can be any type supported by the 
> [structured clone algorithm ↗](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types), as long as its size is less than 128 KB.

and the api docs here for the push method: https://developers.cloudflare.com/api/python/resources/queues/subresources/messages/methods/push/

I think it's okay - as long as it's structured data that can be cloned (and a string is in that category) we're okay. The larger issue might be if you have a js app that is reading from the queue and expecting a js object instead of a json string.

_Philosophically_ I'd say that for interop, you'd never want a real javascript object in your queue because you won't have a guarantee that whatever is reading the queue can interpret a javascript object.

_Pragmatically_ in this example nothing's actually reading from the queue

Finally, the comment on the line above (95 before merge/96 after) says we're sending a json object to the queue. From what I can tell, to_js makes that actually incorrect, because to_js actually has us pushing/sending an actual javascript object. 

I'm okay with removing the updates, but if I do, we should probably update the comment on line 95 to be correct. 

Let me know your preference and I'll make the necessary adjustments if needed.

dom96:
@sjsadowski You're totally right. Thanks for looking into this in so much detail!

dom96 (approved):

src/content/docs/workers/languages/python/examples.mdx
```diff
@@ -37,22 +39,24 @@ async def on_fetch(request, env):
 ## Parse JSON from the incoming request
 
 ```python
-from workers import Response
+from workers import handle, Response
```
mikenomitch:
```suggestion
from workers import handler, Response
```

mikenomitch (approved):

mikenomitch:
@sjsadowski really great additions - thanks so much for the PR

PR #23761, opened 2025-07-17
[Chore] Rationalize GHA as part of 'how we docs' exercise

kodster28:
### Summary

As part of writing the `content reviews` section of [How we docs](https://developers.cloudflare.com/style-guide/how-we-docs/), I got confused myself parsing through our GHA.

This consolidates dupe actions and clarifies the file naming structure.

### Risk

Small to no risk for this change. If this breaks, we lose labeling and assignment for a bit, which can be fixed easily and doesn't impact the team too drastically.


.github/workflows/issue-label-assign.yml
```diff
@@ -0,0 +1,21 @@
+name: Issue - Label and assign
+
+# **What it does**: Labels Issues with associated products. Updates assignees to match our CODEOWNERS file.
```
kodster28, line 3: Re-creates `issue.yml`, but with clearer naming

.github/workflows/pr-label-assign.yml
```diff
@@ -36,3 +39,16 @@ jobs:
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     continue-on-error: true
+
+  assign:
```
kodster28, line 43: This pulls in from the `pr.yml`. These are triggered on the same event, so we should have them all running in the same action.

KianNH (approved):

PR #23773, opened 2025-07-17
Start to making some incremental improvements to node compat docs

jasnell:
### Summary

Start making some incremental improvements to the Workers nodejs_compat docs.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/runtime-apis/nodejs/EventEmitter.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/EventEmitter.mdx
@@ -8,20 +8,22 @@ import { Render } from "~/components"
 
 <Render file="nodejs-compat-howto" />
 
-An `EventEmitter` is an object that emits named events that cause listeners to be called.
+An [`EventEmitter`](https://nodejs.org/docs/latest/api/events.html#class-eventemitter)
+is an object that emits named events that cause listeners to be called.
 
 ```js
 import { EventEmitter } from 'node:events';
 
 const emitter = new EventEmitter();
 emitter.on('hello', (...args) => {
-  console.log(...args);
+  console.log(...args); // 1 2 3
 });
 
 emitter.emit('hello', 1, 2, 3);
 ```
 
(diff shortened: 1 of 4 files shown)


irvinebroque (approved):

src/content/docs/workers/runtime-apis/nodejs/EventEmitter.mdx
```diff
@@ -33,4 +35,12 @@ emitter.on('error', (err) => {
 });
 ```
 
+Like Node.js, when an `'error'` event is emitted on an `EventEmitter` and there
+is no listener for it, the error will be immediately thrown. However, in Node.js
+it is possible to create add a handler on the `process` object for the
```
jasnell:
```suggestion
it is possible to add a handler on the `process` object for the
```

src/content/docs/workers/runtime-apis/nodejs/buffer.mdx
```diff
@@ -35,4 +35,13 @@ const writer = writable.getWriter();
...
 ```
 
+One key difference between the Workers implementation of `Buffer` and the Node.js
+implementation is that some methods of creating a `Buffer` in Node.js will allocate
+those from a global memory pool as a performance optimization. The Workers implementation
+does not use a memory pool and all `Buffer` instances are allocated independently.
+
+Further, in Node.js it is possible to allocate a `Buffer` with uninitialized memory
+using the `Buffer.allocUnsafe()` method. This is not supported in Workers and `Buffer`
+instances are always zero-filled when allocated.
```
mikenomitch: Layman's description of zero-filled would probably help

mikenomitch (approved):

src/content/docs/workers/runtime-apis/nodejs/buffer.mdx
jasnell:
```suggestion
instances are always initialized so that the `Buffer` is always filled
with null bytes (`0x00`) when allocated.
```

PR #23787, opened 2025-07-18
[Logs] Update firewall_events.md

luishcardoso:
Hello,

### Summary

Ray ID as per its docs https://developers.cloudflare.com/fundamentals/reference/cloudflare-ray-id/

Thanks!


soheiokamoto:
This change needs to be done on the field template on the internal repo.

Also, you can update the field description, but not the field name (with a whitespace).

PR #23815, opened 2025-07-19
Containers docs: Uses importable env and removes manualStart

mikenomitch:
### Summary

This PR uses importable env to set environment variable at the class level. This is needed for most use cases.

It also removes the secret store example. While Secret Store can be used, most users are setting secrets at the class level, which happens outside of Workers I/O context. This means that you can't get a secret from the store. Rather than explain this, I've just removed it, and will circle back once a deeper integration is made.

Also, removes `manualStart` as this is no longer used and clarifies some wording around when containers start.

### Documentation checklist

### DIFF
--- a/src/content/docs/containers/faq.mdx
+++ b/src/content/docs/containers/faq.mdx
@@ -135,10 +135,12 @@ to define secrets for your Workers.
 Then you can pass these secrets to your Container using the `envVars` property:
 
 ```javascript
+import { env } from "cloudflare:workers";
+
 class MyContainer extends Container {
 	defaultPort = 5000;
 	envVars = {
-		MY_SECRET: this.env.MY_SECRET,
+		MY_SECRET: env.MY_SECRET,
 	};
 }
 ```
@@ -146,9 +148,11 @@ class MyContainer extends Container {
 Or when starting a Container instance on a Durable Object:
 
 ```javascript
+import { env } from "cloudflare:workers";
+
 this.ctx.container.start({
 	env: {
-		MY_SECRET: this.env.MY_SECRET,
+		MY_SECRET: env.MY_SECRET,
 	},
 });
 ```
--- a/src/content/docs/containers/examples/cron.mdx
+++ b/src/content/docs/containers/examples/cron.mdx
@@ -61,7 +61,6 @@ import { Container, getContainer } from "@cloudflare/containers";
 
 export class CronContainer extends Container {
 	sleepAfter = "5m";
-	manualStart = true;
 }
 
 export default {
--- a/src/content/docs/containers/examples/env-vars-and-secrets.mdx
+++ b/src/content/docs/containers/examples/env-vars-and-secrets.mdx
@@ -13,47 +13,22 @@ import { WranglerConfig, PackageManagers } from "~/components";
 Environment variables can be passed into a Container using the `envVars` field
 in the `Container` class, or by setting manually when the Container starts.
 
-Secrets can be passed into a Container by using [Worker Secrets](/workers/configuration/secrets/)
-or the [Secret Store](/secrets-store/integrations/workers/), then passing them into the Container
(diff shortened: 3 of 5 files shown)


anthonywu:
> This means that you can't get a secret from the store

ok so this means a config like this would not have worked as of July 17 2025

```
[[env.production.secrets_store_secrets]]
binding = "MY_SECRET"
store_id = "mystoreid"
secret_name = "some-secret-readable-name"
```

My issue was that `env.MY_SECRET` binding comes through (as configured in toml and .ts), but then is just some `[object Fetcher]` when it arrives in the .ts worker and `console.log`s as literally `[object Fetcher]`)

src/content/docs/containers/faq.mdx
```diff
@@ -135,20 +135,24 @@ to define secrets for your Workers.
 Then you can pass these secrets to your Container using the `envVars` property:
 
 ```javascript
+import { env } from "cloudflare:workers";
+
 class MyContainer extends Container {
 	defaultPort = 5000;
 	envVars = {
-		MY_SECRET: this.env.MY_SECRET,
+		MY_SECRET: env.MY_SECRET,
```
anthonywu, line 143: I think this needs stay `this.env` as `this` refers to `Container` and `this.env` is inherited `Container.env`

it is `env.MY_SECRET` pattern that's throwing this on the dev tail:

```
✘ [ERROR] Uncaught ReferenceError: env is not defined
```

When I put this back to `this.env` in the Container envVars, it works again.

anthonywu:
The FAQ says:

> You can use [Worker Secrets](/workers/configuration/secrets/) or the [Secrets Store](/secrets-store/integrations/workers/) to define secrets for your Workers.

If the Secrets Store implementation is not working today, it should not be mentioned at all, or labeled as coming later.

image

^ this is what I've been trying to get to work, want to know if this is essentially a burned strategy

I'm at the stage where I'm nuking my worker on the GUI to start over.

anthonywu:
In env-vars-and-secrets, this doesn't make sense to me:

image

I'm confused why the config mentions the secret, but sets it to an obviously fake value.

```
name = "my-container-worker"

[vars]
ACCOUNT_API_KEY = "abcdef12345"
```

Shouldn't `abcdef12345` not be typed in at all, because the value comes from the Worker secrets?

`CONTAINER_SECRET_KEY` literal string is mentioned on the page, I don't see any other refs to it, what is this value - is this some magical string that is special, or is this some value that is just being shown as an example? In any case, it's a dangling reference?

Later:

```
  envVars = {
    ACCOUNT_NAME: env.ACCOUNT_NAME,
    ACCOUNT_API_KEY: env.ACCOUNT_API_KEY,
  };
```

all I see so far is you typed in `ACCOUNT_API_KEY = "abcdef12345"` litearlly in the config, I still don't make the connection how this value is read from a place that's not literally in the file.

anthonywu:
Everywhere resources are being added, but can be inherited into an environment, I think it's worth mentioning that setting `--env envname` is SIGNIFICANT.

anthonywu:
I'll circle back on Monday. Have a great weekend.

I have to state this though: Kubernetes would have been easier.

anthonywu:
I forked the uv maintainer's cloudflare container example repo and made some alterations:
 https://github.com/anthonywu/uv-cloudflare-containers-example/tree/demo-env-port-changes

The few changes I made:

1. use a diff port than 8080, prove we can deviate from this
2. use a 0.0.0.0 binding interface, deviate from 10.0.0.1
3. sets a secret via "wrangler secret put" and also via ".dev.vars" for "wrangler dev" mode
4. set a plaintext config

I think this template (after a few more bells and whistles) should be offered on the "Get Started" -> "Select a template" menu.

anthonywu:
I'm ok letting this issue close out

PR #23821, opened 2025-07-20
Update rules-of-workflows.mdx

loveJesus:
✝️Hallelujah I think the good instance unique id is fixed

### Summary
I think the goodInstance was the same as the bad instance and intends a new id per request which this should help alleviate.


Oxyjun:
Thanks for this PR! This has been addressed via https://github.com/cloudflare/cloudflare-docs/pull/19143.

PR #23842, opened 2025-07-21
[Docs Site] Create videos pages

KianNH:
### Summary

Create videos pages


kodster28 (approved):
Since these are all hidden from public view, think this is fine to merge in for now (though removed the `example video` content just in case someone stumbles on it).

PR #23879, opened 2025-07-22
[ZT] update mTLS instructions for Access

ranbel:
PCX-11280
PCX-9420


src/content/docs/cloudflare-one/identity/devices/access-integrations/mutual-tls-authentication.mdx
```diff
@@ -18,84 +17,226 @@ With a root certificate authority (CA) in place, Access only allows requests fro
...
 
-mTLS is checked on a per host basis. Access sets a flag for when a client certificate was presented and successfully completed mTLS authentication. However, to actually enforce mTLS, you need an Access policy in place, and Access policies are both host and path specific. If you want to enforce mTLS on a specific path, you need to make sure your Access policies are configured accordingly.
-
 :::
 
-To enforce mTLS authentication from [Zero Trust](https://one.dash.cloudflare.com):
+## Add mTLS to your Access application
 
-1. Go to **Access** > **Service auth** > **Mutual TLS**.
+mTLS is checked on a per host basis. Access sets a flag for when a client certificate was presented and successfully completed mTLS authentication. However, to actually enforce mTLS, you need to secure the specific hostname and path with an Access application and policy.
```
kennyj42: I think this statement is confusing. The second and third sentence feel like they contradict.

Is it something like:
"mTLS is checked on a per host basis. Regardless of policy, Access sets a flag when a client certificate was presented and the mutual TLS was successful. If you want to enforce a specific mTLS condition, this requires an Access application and policy."

kennyj42 (approved):

src/content/docs/cloudflare-one/identity/devices/access-integrations/mutual-tls-authentication.mdx
kennyj42: @jroyal do you know what we would've meant by "sets a flag" back in the day? Is that a header or something like that?

jroyal: It's set in the internal context for the request. Workers exposes this now, but Access effectively looks at some of these values when you set the rule in a policy.
 
https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties:~:text=GCM%2DSHA256%22.-,tlsClientAuth,-Object%20%7C%20null

ranbel: What are the customer-facing implications of this flag? 

My current understanding of the note: Uploading an Access mTLS cert for `test.example.com` automatically turns on mTLS checks for `test.example.com`. Access app is only required if you want to check for a specific common name or add other rules like IdP, device posture, etc...
@kennyj42 does that sound right?

jroyal: Uploading a cert only makes Cloudflare do the MTLS dance and tell us/workers about what happened. You can still visit that hostname all day long and nothing will stop you. We don't actually enforce that you presented a valid certificate unless you add a rule to your access policy.

src/content/docs/cloudflare-one/identity/devices/access-integrations/mutual-tls-authentication.mdx
```diff
@@ -18,84 +17,229 @@ With a root certificate authority (CA) in place, Access only allows requests fro
...
+	 openssl genrsa -out client.key 2048
+	 ```
+
+2. Create a CSR for the client certificate:
+
+	```sh
+	openssl req -new -key client.key -out client.csr
+	```
+
+	You will be prompted to fill in some optional fields. For testing purposes, you can set **Common Name** to something like `John Doe`.
```
Maddy-Cloudflare, line 218:
```suggestion
You will be prompted to fill in some optional fields. For testing purposes, you can set **Common Name** to something like `John Doe`.
```

Maddy-Cloudflare (approved):

PR #23930, opened 2025-07-24
[Vectorize] Vectorize improves insert and upsert reliability and performance

sejoker:
Changelog update

### Summary

Vectorize has been updated to improve the latency of vector inserts and upserts by up to 20% for p95, and to eliminate the risk of API rate limiting from these operations


sejoker:
Closing PR as performance claims are not confirmed.

PR #23947, opened 2025-07-24
Add backfill changelog entry for http publish for queues

jonesphillip:
Add backfill changelog entry for http publish for queues


irvinebroque (commented):
Can you publish to a Queue using https://github.com/cloudflare/cloudflare-typescript ?

Think so: https://github.com/cloudflare/cloudflare-typescript/blob/main/src/resources/queues/messages.ts#L92

...but we don't mention here, so not sure anyone would know they can use the SDK or that it exists

aninibread (approved):

irvinebroque (approved):

PR #23974, opened 2025-07-25
Update errors.mdx

dinasaur404:
Add new error code for incorrect content-type


WalshyDev (approved):

irvinebroque (commented):
what do i do about it if i get this error?

PR #23975, opened 2025-07-25
[Docs Site] Limit /changelog/ to 5 entries per product

KianNH:
### Summary

Limit /changelog/ to 5 entries per product


kodster28 (commented):
We need a way to get folks from filtered `/changelog` to the historical entries.

Probably should adjust the behavior of the `Load more` button to only show when there's less than 7 entries and change the destination of that link?

kodster28 (approved):

PR #24003, opened 2025-07-28
Mention remote bindings do not work with RPC yet

emily-shen:
DEVX-2023


dario-piotrowicz (commented):
I feel like the wording might be a bit off 🤔 

Service bindings with RPC are supported, but the RPC communication isn't.

Your wording does make it sound to me like adding RPC to a service binding will make that binding unusable in remote mode, which is not the case. 🤔

dario-piotrowicz (approved):
Looks good to me 🙂

PR #24012, opened 2025-07-28
[Fundamentals] Create copy-settings

ngayerie:
PCX-18561


dcpena:
Closing for now because we've discussed in the ticket that this content should be in a note on an existing page instead of a brand new page.

PR #24031, opened 2025-07-28
set http and https modules as partially supported

anonrig:
These modules have documentation but we forgot to set it as "partially supported"


mikenomitch (approved):

irvinebroque (commented):
👍

PR #24048, opened 2025-07-29
Update pricing.mdx

kathayl:
better explain concurrency calc

### Documentation checklist

### DIFF
--- a/src/content/docs/browser-rendering/platform/pricing.mdx
+++ b/src/content/docs/browser-rendering/platform/pricing.mdx
@@ -4,10 +4,11 @@ title: Pricing
 sidebar:
   order: 31
 ---
+import { Details } from "~/components"
 
 Browser Rendering is currently available at no cost up to the [limits](/browser-rendering/platform/limits/) specified, with billing beginning on August 20, 2025.
 
-There are two ways to use Browser Rendering. Depending on the method you use, here’s how billing will work:
+There are two ways to use Browser Rendering. Depending on the method you use, here is how billing works:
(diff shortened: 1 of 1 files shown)


src/content/docs/browser-rendering/platform/pricing.mdx
```diff
@@ -18,6 +18,12 @@ There are two ways to use Browser Rendering. Depending on the method you use, he
 
 You can monitor your Browser Rendering usage in the [Cloudflare dashboard](https://dash.cloudflare.com). Go to **Compute (Workers)** > **Browser Rendering**.
 
+:::note
```
ToriLindsay: I suggest reframing this as a more generic informational paragraph (or even [a dropdown ](https://developers.cloudflare.com/style-guide/components/details/)- which I can help you create) defining the concurrent browser calculation. I think notes can sometimes get overused and make the content look very noisy. Also, if I were scanning the page, looking for more info on how concurrent browsers are calculated, I think I would expect to find a definition that explains the calculation, and not think to read the two-paragraph note.
I would also suggest focussing more on what _is_ the calculation rather than what is _not_ the calculation. And I thought "penalizing" sounded a bit too intense.

Something like this...

**How is the number of concurrent browsers calculated?**
Cloudflare calculates concurrent browsers as the monthly average of your daily peak usage. This approach reflects your typical traffic and ensures you are not disproportionately charged for brief, exceptional spikes in browser concurrency.

ToriLindsay (approved):

src/content/docs/browser-rendering/platform/pricing.mdx
ToriLindsay: Update after chatting with Kathy separately

PR #24116, opened 2025-08-01
Mention Vite's server.headers in Static Assets page

jamesopstad:
### Summary

Small update to Vite plugin docs.

### Documentation checklist


irvinebroque (commented):
use case for this would be if someone needs to replicate some zone-level configuration while in local dev? like a rule or snippet?

should we say that?

PR #24154, opened 2025-08-04
[CF1] identity-based selectors

deadlypants1973:
### Summary

PCX-18417

### Documentation checklist


src/content/docs/cloudflare-one/policies/access/index.mdx
```diff
@@ -133,7 +133,18 @@ To require only one country and one email ending:
 
 When you add a rule to your policy, you will be asked to specify the criteria/attributes you want users to meet. These attributes are available for all Access application types, including [SaaS](/cloudflare-one/applications/configure-apps/saas-apps/), [self-hosted](/cloudflare-one/applications/configure-apps/self-hosted-public-app/), and [non-HTTP](/cloudflare-one/applications/non-http/) applications.
 
-Identity-based attributes are only checked when a user authenticates to Access. Non-identity attributes are polled continuously, meaning they are-evaluated with each new HTTP request for changes during the [user session](/cloudflare-one/identity/users/session-management/). If you have configured [SCIM provisioning](/cloudflare-one/identity/users/scim/), you can force a user to re-attest all attributes with Access whenever you revoke the user in the IdP or update their IdP group membership.
+Identity-based attributes are only checked when a user authenticates to Access. The following selectors are identity-based:
```
jroyal: I'd maybe add a comment about this being an example list of identity based rules. Just because we occasionally add rules and I can almost guarantee we are going to forget to come update this list.

jroyal: Does this list add anything that is not answered in the table below?

deadlypants1973: @jroyal it is implied that if you read 'checked at login' in the table you can deduce that these items are identity-based selectors **but** we do not explicitly call out our identity based selectors. also 'Everyone' is also checked at login but I do not think that is an identity-based selector?

I can:
1. delete the list 
2. leave the the PR as just deleting the 'IP' bit above

if you think we are good as is. But I would change the word 'attributes' to 'selectors' explicity because we use 'selectors' language in docs and dash and only say attributes here somewhat randomly

jroyal: Everyone is a weird one because its basically just a `true`. It always passes. To your point I guess this is clearer, but we will just need to make sure we update it if it changes.

jroyal (approved):

Maddy-Cloudflare (approved):

PR #24268, opened 2025-08-08
add node:http and node:https documentation

anonrig:
### Summary

Adds the documentation for node http and https modules.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/runtime-apis/nodejs/http.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/http.mdx
@@ -3,9 +3,37 @@ pcx_content_type: configuration
 title: http
 ---
 
-import { Render } from "~/components"
+import { Render } from "~/components";
 
 <Render file="nodejs-compat-howto" />
+
+## Compatibility flags
+
+### Client-side methods
+
+To use the HTTP client-side methods (`http.get`, `http.request`, etc.), you must enable the [`enable_nodejs_http_modules`](/workers/configuration/compatibility-flags/) compatibility flag in addition to the [`nodejs_compat`](/workers/runtime-apis/nodejs/) flag.
+
+This flag is automatically enabled for Workers using a [compatibility date](/workers/configuration/compatibility-dates/) of `2025-08-15` or later when `nodejs_compat` is enabled. For Workers using an earlier compatibility date, you can manually enable it by adding the flag to your `wrangler.toml`:
+
+```toml
+compatibility_flags = ["nodejs_compat", "enable_nodejs_http_modules"]
+```
+
+### Server-side methods
+
+To use the HTTP server-side methods (`http.createServer`, `http.Server`, `http.ServerResponse`), you must enable the `enable_nodejs_http_server_modules` compatibility flag in addition to the [`nodejs_compat`](/workers/runtime-apis/nodejs/) flag.
+
+This flag is automatically enabled for Workers using a [compatibility date](/workers/configuration/compatibility-dates/) of `2025-09-01` or later when `nodejs_compat` is enabled. For Workers using an earlier compatibility date, you can manually enable it by adding the flag to your `wrangler.toml`:
+
+```toml
+compatibility_flags = ["nodejs_compat", "enable_nodejs_http_server_modules"]
+```
+
+To use both client-side and server-side methods, enable both flags:
+
+```toml
+compatibility_flags = ["nodejs_compat", "enable_nodejs_http_modules", "enable_nodejs_http_server_modules"]
+```
 
 ## get
 
@@ -18,25 +46,25 @@ fetch or similar handler. Outside of such a handler, attempts to use `get` will 
 an error.
 
 ```js
-import { get } from 'node:http';
+import { get } from "node:http";
 
 export default {
 	async fetch() {
 		const { promise, resolve, reject } = Promise.withResolvers();
-		get('http://example.org', (res) => {
-			let data = '';
-			res.setEncoding('utf8');
-			res.on('data', (chunk) => {
+		get("http://example.org", (res) => {
+			let data = "";
+			res.setEncoding("utf8");
+			res.on("data", (chunk) => {
 				data += chunk;
 			});
-			res.on('end', () => {
+			res.on("end", () => {
 				resolve(new Response(data));
 			});
-			res.on('error', reject);
-		}).on('error', reject);
+			res.on("error", reject);
+		}).on("error", reject);
 		return promise;
-  }
-}
+	},
+};
 ```
 
 The implementation of `get` in Workers is a wrapper around the global
@@ -58,31 +86,35 @@ fetch or similar handler. Outside of such a handler, attempts to use `request` w
 an error.
 
 ```js
-import { get } from 'node:http';
+import { get } from "node:http";
 
 export default {
 	async fetch() {
 		const { promise, resolve, reject } = Promise.withResolvers();
-		get({
-			method: 'GET',
-			protocol: 'http:',
-			hostname: 'example.org',
-			path: '/'
-		}, (res) => {
-			let data = '';
-			res.setEncoding('utf8');
-			res.on('data', (chunk) => {
-				data += chunk;
-			});
-			res.on('end', () => {
-				resolve(new Response(data));
-			});
-			res.on('error', reject);
-		}).on('error', reject)
-		  .end();
+		get(
+			{
+				method: "GET",
+				protocol: "http:",
+				hostname: "example.org",
+				path: "/",
+			},
+			(res) => {
+				let data = "";
+				res.setEncoding("utf8");
+				res.on("data", (chunk) => {
+					data += chunk;
+				});
+				res.on("end", () => {
+					resolve(new Response(data));
+				});
+				res.on("error", reject);
+			},
+		)
+			.on("error", reject)
+			.end();
 		return promise;
-  }
-}
+	},
+};
 ```
 
 The following options passed to the `request` (and `get`) method are not supported due to the differences required by Coudflare Workers implementation of `node:http` as a wrapper around the global `fetch` API:
@@ -97,26 +129,17 @@ The following options passed to the `request` (and `get`) method are not support
 
 The [`OutgoingMessage`](https://nodejs.org/docs/latest/api/http.html#class-httpoutgoingmessage) class represents an HTTP response that is sent to the client. It provides methods for writing response headers and body, as well as for ending the response. `OutgoingMessage` extends from the Node.js [`stream.Writable` stream class](https://developers.cloudflare.com/workers/runtime-apis/nodejs/streams/).
 
-```js
-import { OutgoingMessage } from 'node:http';
+The `OutgoingMessage` class is a base class for outgoing HTTP messages (both requests and responses). It provides methods for writing headers and body data, as well as for ending the message. `OutgoingMessage` extends from the [`Writable` stream class](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable).
 
-export default {
-	async fetch() {
-		// ...
-		const res = new OutgoingMessage();
-		res.writeHead(200, { 'Content-Type': 'text/plain' });
-		res.write('Hello, World!');
-		res.end();
-		// ...
-	}
-}
-```
+Both `ClientRequest` and `ServerResponse` both extend from and inherit from `OutgoingMessage`.
 
 ## IncomingMessage
 
 The `IncomingMessage` class represents an HTTP request that is received from the client. It provides methods for reading request headers and body, as well as for ending the request. `IncomingMessage` extends from the `Readable` stream class.
 
-```js
+The `IncomingMessage` class represents an HTTP message (request or response). It provides methods for reading headers and body data. `IncomingMessage` extends from the `Readable` stream class.
+
+````js
 import { get, IncomingMessage } from 'node:http';
 import { ok, strictEqual } from 'node:assert';
 
@@ -129,7 +152,37 @@ export default {
 		// ...
 	}
 }
-```
+
+The Workers implementation includes a `cloudflare` property on `IncomingMessage` objects:
+
+```js
+import { createServer } from "node:http";
+import { httpServerHandler } from "cloudflare:node";
+
+const server = createServer((req, res) => {
+	console.log(req.cloudflare.cf.country);
+	console.log(req.cloudflare.cf.ray);
+	res.write("Hello, World!");
+	res.end();
+});
+
+server.listen(8080);
+
+export default httpServerHandler({ port: 8080 });
+````
+
+The `cloudflare.cf` property contains [Cloudflare-specific request properties](/workers/runtime-apis/request/#incomingrequestcfproperties).
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Trailer headers are not supported
+- The `socket` attribute **does not extend from `net.Socket`** and only contains the following properties: `encrypted`, `remoteFamily`, `remoteAddress`, `remotePort`, `localAddress`, `localPort`, and `destroy()` method.
+- The following `socket` attributes behave differently than their Node.js counterparts:
+  - `remoteAddress` will return `127.0.0.1` when ran locally
+  - `remotePort` will return a random port number between 2^15 and 2^16
+  - `localAddress` will return the value of request's `host` header if exists. Otherwise, it will return `127.0.0.1`
+  - `localPort` will return the port number assigned to the server instance
+  - `req.socket.destroy()` falls through to `req.destroy()`
 
 ## Agent
 
@@ -138,37 +191,161 @@ A partial implementation of the Node.js [`http.Agent'](https://nodejs.org/docs/l
 An `Agent` manages HTTP connection reuse by maintaining request queues per host/port. In the workers environment, however, such low-level management of the network connection, ports, etc, is not relevant because it is handled by the Cloudflare infrastructure instead. Accordingly, the implementation of `Agent` in Workers is a stub implementation that does not support connection pooling or keep-alive.
 
 ```js
-import { get, Agent } from 'node:http';
-import { strictEqual } from 'node:assert';
+import { Agent } from "node:http";
+import { strictEqual } from "node:assert";
+
+const agent = new Agent();
+strictEqual(agent.protocol, "http:");
+```
+
+## createServer
+
+An implementation of the Node.js [`http.createServer`](https://nodejs.org/docs/latest/api/http.html#httpcreateserveroptions-requestlistener) method.
+
+The `createServer` method creates an HTTP server instance that can handle incoming requests.
+
+```js
+import { createServer } from "node:http";
+import { httpServerHandler } from "cloudflare:node";
+
+const server = createServer((req, res) => {
+	res.writeHead(200, { "Content-Type": "text/plain" });
+	res.end("Hello from Node.js HTTP server!");
+});
+
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+## Node.js integration
+
+### httpServerHandler
+
+The `httpServerHandler` function integrates Node.js HTTP servers with the Cloudflare Workers request model. It supports two API patterns:
+
+```js
+import http from "node:http";
+import { httpServerHandler } from "cloudflare:node";
+
+const server = http.createServer((req, res) => {
+	res.end("hello world");
+});
+
+// Pass server directly (simplified) - automatically calls listen() if needed
+export default httpServerHandler(server);
+
+// Or use port-based routing for multiple servers
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+The handler automatically routes incoming Worker requests to your Node.js server. When using port-based routing, the port number acts as a routing key to determine which server handles requests, allowing multiple servers to coexist in the same Worker.
+
+### handleAsNodeRequest
+
+For more direct control over request routing, you can use the `handleAsNodeRequest` function from `cloudflare:node`. This function directly routes a Worker request to a Node.js server running on a specific port:
+
+```js
+import { createServer } from "node:http";
+import { handleAsNodeRequest } from "cloudflare:node";
+
+const server = createServer((req, res) => {
+	res.writeHead(200, { "Content-Type": "text/plain" });
+	res.end("Hello from Node.js HTTP server!");
+});
+
+server.listen(8080);
 
 export default {
-	async fetch() {
-
-    const agent = new Agent();
-    get({
-			hostname: 'example.org',
-			port: 80,
-			path: '/',
-			agent, // possible but not all that useful.
-		}, (res) => {
-			// ...
-		});
-
-		return new Response('ok');
-	}
-}
+	fetch(request) {
+		return handleAsNodeRequest(8080, request);
+	},
+};
 ```
+
+This approach gives you full control over the fetch handler while still leveraging Node.js HTTP servers for request processing.
+
+:::note
+Failing to call `close()` on an HTTP server may result in the server persisting until the worker is destroyed. In most cases, this is not an issue since servers typically live for the lifetime of the worker. However, if you need to create multiple servers during a worker's lifetime or want explicit lifecycle control (such as in test scenarios), call `close()` when you're done with the server, or use [explicit resource management](https://v8.dev/features/explicit-resource-management).
+:::
+
+## Server
+
+An implementation of the Node.js [`http.Server`](https://nodejs.org/docs/latest/api/http.html#class-httpserver) class.
+
+The `Server` class represents an HTTP server and provides methods for handling incoming requests. It extends the Node.js `EventEmitter` class and can be used to create custom server implementations.
+
+When using `httpServerHandler`, the port number specified in `server.listen()` acts as a routing key rather than an actual network port. The handler uses this port to determine which HTTP server instance should handle incoming requests, allowing multiple servers to coexist within the same Worker by using different port numbers for identification. Using a port value of `0` (or `null` or `undefined`) will result in a random port number being assigned.
+
+```js
+import { Server } from "node:http";
+import { httpServerHandler } from "cloudflare:node";
+
+const server = new Server((req, res) => {
+	res.writeHead(200, { "Content-Type": "application/json" });
+	res.end(JSON.stringify({ message: "Hello from HTTP Server!" }));
+});
+
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Connection management methods such as `closeAllConnections()` and `closeIdleConnections()` are not implemented
+- Only `listen()` variants with a port number or no parameters are supported: `listen()`, `listen(0, callback)`, `listen(callback)`, etc. For reference, see the [Node.js documentation](https://nodejs.org/docs/latest/api/net.html#serverlisten).
+- The following server options are not supported: `maxHeaderSize`, `insecureHTTPParser`, `keepAliveTimeout`, `connectionsCheckingInterval`
+
+## ServerResponse
+
+An implementation of the Node.js [`http.ServerResponse`](https://nodejs.org/docs/latest/api/http.html#class-httpserverresponse) class.
+
+The `ServerResponse` class represents the server-side response object that is passed to request handlers. It provides methods for writing response headers and body data, and extends the Node.js `Writable` stream class.
+
+```js
+import { createServer, ServerResponse } from "node:http";
+import { httpServerHandler } from "cloudflare:node";
+import { ok } from "node:assert";
+
+const server = createServer((req, res) => {
+	ok(res instanceof ServerResponse);
+
+	// Set multiple headers at once
+	res.writeHead(200, {
+		"Content-Type": "application/json",
+		"X-Custom-Header": "Workers-HTTP",
+	});
+
+	// Stream response data
+	res.write('{"data": [');
+	res.write('{"id": 1, "name": "Item 1"},');
+	res.write('{"id": 2, "name": "Item 2"}');
+	res.write("]}");
+
+	// End the response
+	res.end();
+});
+
+export default httpServerHandler(server);
+```
+
+The following methods and features are not supported in the Workers implementation:
+
+- `assignSocket()` and `detachSocket()` methods are not available
+- Trailer headers are not supported
+- `writeContinue()` and `writeEarlyHints()` methods are not available
+- 1xx responses in general are not supported
 
 ## Other differences between Node.js and Workers implementation of `node:http`
 
 Because the Workers implementation of `node:http` is a wrapper around the global `fetch` API, there are some differences in behavior and limitations compared to a standard Node.js environment:
 
-* `Connection` headers are not used. Workers will manage connections automatically.
-* `Content-Length` headers will be handled the same way as in the `fetch` API. If a body is provided, the header will be set automatically and manually set values will be ignored.
-* `Expect: 100-continue` headers are not supported.
-* Trailing headers are not supported.
-* The `'continue'` event is not supported.
-* The `'information'` event is not supported.
-* The `'socket'` event is not supported.
-* The `'upgrade'` event is not supported.
-* Gaining direct access to the underlying `socket` is not supported.
+- `Connection` headers are not used. Workers will manage connections automatically.
+- `Content-Length` headers will be handled the same way as in the `fetch` API. If a body is provided, the header will be set automatically and manually set values will be ignored.
+- `Expect: 100-continue` headers are not supported.
+- Trailing headers are not supported.
+- The `'continue'` event is not supported.
+- The `'information'` event is not supported.
+- The `'socket'` event is not supported.
+- The `'upgrade'` event is not supported.
+- Gaining direct access to the underlying `socket` is not supported.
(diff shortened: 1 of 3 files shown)


jasnell:
We should probably mark this draft or hold off on merging until the server side stuff is actually rolled out without the experimental flag.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -51,13 +75,13 @@ get('http://docs.cloudflare.com/robots.txt', (res) => {
...
 
-An implementation of the Node.js [`http.request'](https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback) method.
+An implementation of the Node.js [`http.request`](https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback) method.
 
 The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
 
 ```js
 import { request } from 'node:http';
-import { strictEqual } from 'node:assert';
+import { strictEqual, ok } from 'node:assert';
```
jasnell: these examples are, unfortunately, not directly runnable in workers since they show the requests at the top level scope (where we can't perform an i/o). This can be addressed in a follow-up but ideally these examples would show the use of `request` and `get` as subrequests in a request to show how they would actually be used. Probably also a good idea to show the proper use of await or `waitUntil` to show how to properly wait on them.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+	console.log(res.cloudflare.cf.ray);
+});
+```
+
+The `cloudflare.cf` property contains [Cloudflare-specific request properties](/workers/runtime-apis/request/#incomingrequestcfproperties).
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Trailer headers are not supported
+- The `socket` attribute only contains the following properties: `encrypted`, `remoteFamily`, `remoteAddress`, `remotePort`, `localAddress`, `localPort`, and `destroy()` method
```
jasnell: It's likely worth expanding this a bit to indicate that the `socket` is not actually a `net.Socket` or a `Duplex`, with a short explanation about why. Can be done in a follow up.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: I agree that would be a good improvement, but it's not really related to this PR which just adds the missing modules and methods.

Mind if we track this separately? It's a bigger change that probably deserves its own discussion about how we want to handle the examples and async patterns.

Happy to work on it after this gets merged if you want.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: We actually do mention this - see the bullet point about `socket` not extending `net.Socket`. 

Got any specific wording changes in mind?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+const server = createServer((req, res) => {
+	res.writeHead(200, { 'Content-Type': 'text/plain' });
+	res.end('Hello from Node.js HTTP server!');
+});
+
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+The `httpServerHandler` function integrates Node.js HTTP servers with the Cloudflare Workers request model. When a request arrives at your Worker, the handler automatically routes it to your Node.js server running on the specified port. This bridge allows you to use familiar Node.js server patterns while benefiting from the Workers runtime environment, including automatic scaling, edge deployment, and integration with other Cloudflare services.
```
jasnell: I'm not sure about this way of explaining it. `httpServerHandler` returns a workers handler object that implements the `fetch` export, the fetch export uses the port as a key to determine which app to forward to. The way this is worded it gives the impression that the `node:http` server is actually running on an actual port. Just comes across a bit misleading.

anonrig: Good point. The current wording makes it sound like there's an actual server running on a real port, when it's really using the port as a routing key.

How would you suggest rewording this to be more accurate? Something along the lines of the handler using the port to determine which app to route to?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+
+:::note
+Failing to call `close()` on an HTTP server may result in the server being leaked. To prevent this, call `close()` when you're done with the server, or use explicit resource management:
```
jasnell: "being leaked" here is going to be scary for folks. Most of the time they are likely to only be creating a single server instance per worker and that it lives for the lifetime of the worker. It's only "leaked" in the sense that calling close never unregisters the server but when the isolate/worker is destroyed so is the server. The only time calling close explicitly is going to be necessary is when a worker wants these, for whatever reason, to be more transient in nature (such as in the test cases we have in workerd that create a new server for each individual test)

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+	res.writeHead(200, { 'Content-Type': 'application/json' });
+	res.end(JSON.stringify({ message: 'Hello from HTTP Server!' }));
+});
+```
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Connection management methods such as `closeAllConnections()` and `closeIdleConnections()` are not implemented
+- Only `listen()` variants with a port number or no parameters are supported: `listen()`, `listen(0, callback)`, `listen(callback)`, etc.
+- The following server options are not supported: `maxHeaderSize`, `insecureHTTPParser`, `keepAliveTimeout`, `connectionsCheckingInterval`
```
jasnell, line 297: Can be done in a separate PR but a discussion here about what a `port` actually means in our case would likely be helpful. There's bound to be some confusion around this.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+
+The `Server` class represents an HTTP server and provides methods for handling incoming requests. It extends the Node.js `EventEmitter` class and can be used to create custom server implementations.
+
+```js
+import { Server } from 'node:http';
+
+const server = new Server((req, res) => {
+	res.writeHead(200, { 'Content-Type': 'application/json' });
+	res.end(JSON.stringify({ message: 'Hello from HTTP Server!' }));
+});
```
jasnell, line 287: The example should likely be more complete, showing the call to listen and exporting the handler.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: Though from a technical standpoint, not calling close() does mean the server instance persists indefinitely, which is the definition of a resource leak.

Maybe we could clarify the context better - something like "may result in the server persisting until the worker is destroyed" and then explain when explicit cleanup is actually needed?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+	res.write(']}');
+
+	// End the response
+	res.end();
+});
+```
+
+The following methods and features are not supported in the Workers implementation:
+
+- `assignSocket()` and `detachSocket()` methods are not available
```
jasnell, line 334: It should also be pointed out in here that `writeContinue`, `writeEarlyHints` and 1xx responses in general are not supported

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+
+	// Stream response data
+	res.write('{"data": [');
+	res.write('{"id": 1, "name": "Item 1"},');
+	res.write('{"id": 2, "name": "Item 2"}');
+	res.write(']}');
+
+	// End the response
+	res.end();
+});
```
jasnell, line 327: Likewise here, the example should likely be complete, showing the listen() and exporting the handler.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: > Maybe we could clarify the context better - something like "may result in the server persisting until the worker is destroyed" and then explain when explicit cleanup is actually needed?

Yes, this.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: My preference would be to improve the examples before it lands but I'll leave it to you.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: I would just make it a bit more visible. As is I missed the other bullet point entirely

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -51,13 +75,13 @@ get('http://docs.cloudflare.com/robots.txt', (res) => {
 
 ## request
 
-An implementation of the Node.js [`http.request'](https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback) method.
+An implementation of the Node.js [`http.request`](https://nodejs.org/docs/latest/api/http.html#httprequesturl-options-callback) method.
 
 The `request` method creates an HTTP request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
 
 ```js
 import { request } from 'node:http';
```
jasnell: Better to replace this example with a more fully functional one that shows the request used in the correct way (i.e. as a subrequest within a request handler where there is an i/o context, with proper error propagation, a promise that gets resolved for the response, etc).

```js
import { request } from 'node:http';

export default {
  fetch() {
    const { promise, resolve, reject } = Promise.withResolvers();

    request({
	    method: 'GET',
	    protocol: 'http:',
	    hostname: 'example.com',
	    path: '/'
    }, (res) => {
	    let data = '';
	    res.setEncoding('utf8');
	    res.on('data', (chunk) => {
		    data += chunk;
	    });
	    res.once('end', () => {
        resolve(new Response(data));
	    });
      res.once('error', reject);
    }).once('error', reject)
      .end();

    return promise;
  }
}
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -91,7 +118,7 @@ const req = request(new URL('http://docs.cloudflare.com'),{
 req.end();
 ```
```
jasnell, line 118: I would just remove this second example as it does not really add much value. You can just use text to explain that the input can be an object, URL string, or URL object.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -110,9 +139,12 @@ res.writeHead(200, { 'Content-Type': 'text/plain' });
 res.write('Hello, World!');
 res.end();
 ```
```
jasnell: Given that it would be exceedingly unlikely that the typical use would ever need to create an `OutgoingMessage` themselves, this example is not all that useful. Instead, I would point out that the `ClientRequest` and `ServerResponse` both extend from and inherit from `OutgoingMessage` and remove the example

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
 	ok(res instanceof IncomingMessage);
 });
```
jasnell:
```js
import { get, IncomingMessage } from 'node:http';
import { ok, strictEqual } from 'node:assert';

export default {
  fetch() {
    const { promise, resolve } = Promise.withResolvers();
    get('http://docs.cloudflare.com', (res) => {
	    strictEqual(res.statusCode, 301);
	    ok(res instanceof IncomingMessage);
      resolve(new Response('ok');
    });
    return promise;
  }
}
```

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+The Workers implementation includes a `cloudflare` property on `IncomingMessage` objects:
+
+```js
+import { get } from 'node:http';
+
+get('http://example.com', (res) => {
+	// Access Cloudflare-specific request properties
+	console.log(res.cloudflare.cf.country);
+	console.log(res.cloudflare.cf.ray);
+});
```
jasnell, line 167: the `cloudflare.cf` is a server-side property on `Request` objects. It would not be included on fetch response, so this example is not correct.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+:::note
+Failing to call `close()` on an HTTP server may result in the server persisting until the worker is destroyed. In most cases, this is not an issue since servers typically live for the lifetime of the worker. However, if you need to create multiple servers during a worker's lifetime or want explicit lifecycle control (such as in test scenarios), call `close()` when you're done with the server, or use explicit resource management:
+
+```js
+import { createServer } from 'node:http';
+
+await using server = createServer((req, res) => {
+	res.end('Hello World');
+});
+// Server will be automatically closed when it goes out of scope
```
jasnell: In this example, the server would never actually be retained because it's not listening. Also since the `await using` is at the top level scope, even if it was listening, it would end up being immediately closed when the top-level evaluation scope exits before any requests could be received and processed, so this example is buggy and incorrect as is.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+```
+:::
+
+## Server
+
+An implementation of the Node.js [`http.Server`](https://nodejs.org/docs/latest/api/http.html#class-httpserver) class.
+
+The `Server` class represents an HTTP server and provides methods for handling incoming requests. It extends the Node.js `EventEmitter` class and can be used to create custom server implementations.
+
+When using `httpServerHandler`, the port number specified in `server.listen()` acts as a routing key rather than an actual network port. The handler uses this port to determine which HTTP server instance should handle incoming requests, allowing multiple servers to coexist within the same Worker by using different port numbers for identification.
```
jasnell: It's also worth pointing out that passing a port value of `0` (or `null` or `undefined`) will result in a random port number being assigned.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+	console.log(res.cloudflare.cf.ray);
+});
+```
+
+The `cloudflare.cf` property contains [Cloudflare-specific request properties](/workers/runtime-apis/request/#incomingrequestcfproperties).
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Trailer headers are not supported
+- The `socket` attribute **does not extend from `net.Socket`** and only contains the following properties: `encrypted`, `remoteFamily`, `remoteAddress`, `remotePort`, `localAddress`, `localPort`, and `destroy()` method
```
jasnell: Should likely explain that calling `msg.socket.destroy()` just falls through to `msg.destroy()`. Otherwise, it would probably be good to document these a bit more (e.g. explaining the expected `remotePort` range, what `remoteAddress` represents, the fact that `localAddress` is generally non-op, etc and possibly provide links to the node.js equivalent documentation for these. Can be done in a separate PR tho.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+});
+
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Connection management methods such as `closeAllConnections()` and `closeIdleConnections()` are not implemented
+- Only `listen()` variants with a port number or no parameters are supported: `listen()`, `listen(0, callback)`, `listen(callback)`, etc.
```
jasnell: General users may not be too familiar with the fact that Node.js' API allows listening on a domain socket, etc so this may be a bit confusing. A link here to the relevant node.js docs would help.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+});
+
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+The `httpServerHandler` function integrates Node.js HTTP servers with the Cloudflare Workers request model. When a request arrives at your Worker, the handler automatically routes it to your Node.js server running on the specified port. This bridge allows you to use familiar Node.js server patterns while benefiting from the Workers runtime environment, including automatic scaling, edge deployment, and integration with other Cloudflare services.
+
+:::note
+Failing to call `close()` on an HTTP server may result in the server persisting until the worker is destroyed. In most cases, this is not an issue since servers typically live for the lifetime of the worker. However, if you need to create multiple servers during a worker's lifetime or want explicit lifecycle control (such as in test scenarios), call `close()` when you're done with the server, or use explicit resource management:
```
jasnell: Users likely aren't going to be familiar with explicit resource management since it is so new. A link here to something that explains it would be most helpful.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+The following differences exist between the Workers implementation and Node.js:
+
+- Trailer headers are not supported
+- The `socket` attribute **does not extend from `net.Socket`** and only contains the following properties: `encrypted`, `remoteFamily`, `remoteAddress`, `remotePort`, `localAddress`, `localPort`, and `destroy()` method
+
+## createServer
+
+An implementation of the Node.js [`http.createServer`](https://nodejs.org/docs/latest/api/http.html#httpcreateserveroptions-requestlistener) method.
+
+The `createServer` method creates an HTTP server instance that can handle incoming requests. It's a convenience function that creates a new `Server` instance and optionally sets up a request listener callback.
```
jasnell:
```suggestion
The `createServer` method creates an HTTP server instance that can handle incoming requests.
```

It's not really a "convenience" method. It's the way users are supposed to create servers. The fact that `new Server(...)` is possible is a historical artifact of the original node.js implementation of the API. I would just drop the second sentence here.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -90,3 +114,62 @@ req.end();
...
+
+:::note
+Failing to call `close()` on an HTTPS server may result in the server being leaked. To prevent this, call `close()` when you're done with the server, or use explicit resource management:
+
+```js
+import { createServer } from 'node:https';
+
+await using server = createServer((req, res) => {
+	res.end('Hello World');
+});
```
jasnell, line 153: As in the http.mdx case, this example is incorrect. The server is not listening (so it would never be retained) and the use of `await using` at the top level scope would be even if it were listening, it would be closed before any requests could be processed.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -90,3 +114,62 @@ req.end();
...
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
+```
+
+The following differences exist between the Workers implementation and Node.js:
+
+- Connection management methods such as `closeAllConnections()` and `closeIdleConnections()` are not implemented
+- Only `listen()` variants with a port number or no parameters are supported: `listen()`, `listen(0, callback)`, `listen(callback)`, etc.
+- The following server options are not supported: `maxHeaderSize`, `insecureHTTPParser`, `keepAliveTimeout`, `connectionsCheckingInterval`
+- TLS/SSL-specific options such as `ca`, `cert`, `key`, `pfx`, `rejectUnauthorized`, `secureProtocol` are not supported in the Workers environment
```
jasnell: It would be most helpful if these all included links to the relevant parts of the node.js docs. Can be added in a separate PR tho.

src/content/docs/workers/runtime-apis/nodejs/index.mdx
```diff
@@ -47,9 +47,9 @@ The runtime APIs from Node.js listed below as "🟢 supported" are currently nat
 | Events                                                                               | 🟢 supported                              |
 | File system                                                                          | ⚪ coming soon                            |
 | Globals                                                                              | 🟢 supported                              |
-| HTTP                                                                                 | 🟡 partially supported                    |
+| HTTP                                                                                 | 🟢 supported                              |
 | HTTP/2                                                                               | ⚪ not yet supported                      |
-| HTTPS                                                                                | 🟡 partially supported                    |
+| HTTPS                                                                                | 🟢 supported                              |
```
jasnell, line 52: I actually think it might be good to keep this as "partially supported" since there's a fair amount of the https options that aren't implemented or are non-ops that we currently have no plans to support. (If the gaps are just temporary, then that's different)

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -90,3 +114,62 @@ req.end();
...
+});
+// Server will be automatically closed when it goes out of scope
+```
+:::
+
+## Server
+
+An implementation of the Node.js [`https.Server`](https://nodejs.org/docs/latest/api/https.html#class-httpsserver) class.
+
+The `Server` class represents an HTTPS server and provides methods for handling incoming secure requests. It extends the Node.js `EventEmitter` class and can be used to create custom secure server implementations.
```
jasnell: Suggest something like this instead:

```
In Node.js, the `https.Server` class represents an HTTPS server and provides
methods for handling incoming secure requests. In Workers, handling of secure
requests is provided by the Cloudflare infrastructure so there really is not
much difference between using `https.Server` or `http.Server`. The workers
runtime provides an implementation for completeness but most workers should
probably just use `http.Server`.
```

likely with a link to the `http.Server` doc.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -55,11 +79,11 @@ export default {
...
 ## request
 
-An implementation of the Node.js [`https.request'](https://nodejs.org/docs/latest/api/https.html#httpsrequestoptions-callback) method.
+An implementation of the Node.js [`https.request`](https://nodejs.org/docs/latest/api/https.html#httpsrequestoptions-callback) method.
 
-The [`request`](https://nodejs.org/docs/latest/api/https.html#httpsrequestoptions-callback) method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://developers.cloudflare.com/workers/runtime-apis/streams/writablestream/) for sending request data.
+The `request` method creates an HTTPS request with customizable options like method, headers, and body. It provides full control over the request configuration and returns a [writable stream](https://nodejs.org/docs/latest/api/stream.html#class-streamwritable) for sending request data.
 
-Request method accepts all options from `http.request` with some differences in default values:
+The request method accepts all options from `http.request` with some differences in default values:
```
jasnell: Making `http.request` a link over to the http.mdx file would be helpful.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -123,3 +155,123 @@ get('http://docs.cloudflare.com', (res) => {
...
+import { createServer } from 'node:http';
+import { httpServerHandler } from 'cloudflare:node';
+
+const server = createServer((req, res) => {
+	res.writeHead(200, { 'Content-Type': 'text/plain' });
+	res.end('Hello from Node.js HTTP server!');
+});
+
+server.listen(8080);
+export default httpServerHandler({ port: 8080 });
```
jasnell, line 217: The signature and behavior of the `httpServerHandler(...)` API needs to be documented in here as well.

src/content/docs/workers/runtime-apis/nodejs/index.mdx
danlapid: I would say exactly the opposite.
If we believe this is what full support looks like, we should say supported.
If anyone is missing something they can create a ticket.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: Note that a couple of the details of how `httpServerHandler(...)` works have been updated here: https://github.com/cloudflare/workerd/pull/4735

src/content/docs/workers/runtime-apis/nodejs/https.mdx
anonrig: Let's add them as a follow up.

src/content/docs/workers/runtime-apis/nodejs/index.mdx
anonrig: I agree that we should mark this as fully supported.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -3,10 +3,52 @@ pcx_content_type: configuration
...
+compatibility_flags = ["nodejs_compat", "enable_nodejs_http_server_modules"]
+```
+
+To use both client-side and server-side methods, enable both flags:
+
+```toml
+compatibility_flags = ["nodejs_compat", "enable_nodejs_http_modules", "enable_nodejs_http_server_modules"]
+```
+
+## Agent
```
jasnell: I had landed a change last week I think that moved this to the end of the docs since it is non-functional. Rather than moving it back to the front I prefer it at the end. There's nothing users are able to do with this right now so it doesn't make sense for it to be at the top of the doc

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -18,25 +60,29 @@ fetch or similar handler. Outside of such a handler, attempts to use `get` will
...
+import { strictEqual, ok } from "node:assert";
 
 export default {
 	async fetch() {
 		const { promise, resolve, reject } = Promise.withResolvers();
-		get('http://example.org', (res) => {
-			let data = '';
-			res.setEncoding('utf8');
-			res.on('data', (chunk) => {
+		get("http://docs.cloudflare.com/robots.txt", (res) => {
```
jasnell: this example really doesn't need to be changed again, and I prefer the simpler example without the use of the asserts.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -58,34 +104,42 @@ fetch or similar handler. Outside of such a handler, attempts to use `request` w
...
-			res.on('data', (chunk) => {
-				data += chunk;
-			});
-			res.on('end', () => {
-				resolve(new Response(data));
-			});
-			res.on('error', reject);
-		}).on('error', reject)
-		  .end();
+		const req = request(
```
jasnell: Likewise, this example really doesn't need to be updated again, and I prefer the simpler version without the asserts.

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -95,80 +149,201 @@ The following options passed to the `request` (and `get`) method are not support
...
+		const { promise, resolve } = Promise.withResolvers();
+		get("http://docs.cloudflare.com", (res) => {
+			strictEqual(res.statusCode, 301);
 			ok(res instanceof IncomingMessage);
+			resolve(new Response("ok"));
 		});
-		// ...
-	}
-}
+		return promise;
```
jasnell: Let's include the error handling here too (e.g. reject and error events)

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -95,80 +149,201 @@ The following options passed to the `request` (and `get`) method are not support
...
+const server = createServer((req, res) => {
+	console.log(req.cloudflare.cf.country);
+	console.log(req.cloudflare.cf.ray);
+	res.write("Hello, World!");
+	res.end();
+});
+
+server.listen(8080);
+
+export default httpServerHandler({ port: 8080 });
```
jasnell, line 171:
```suggestion
export default httpServerHandler(server);
```

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -7,26 +7,64 @@ import { Render } from "~/components";
...
 
-An implementation of the Node.js [`https.get'](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method.
+An implementation of the Node.js [`https.get`](https://nodejs.org/docs/latest/api/https.html#httpsgetoptions-callback) method.
 
 The `get` method performs a GET request to the specified URL and invokes the callback with the response. This is a convenience method that simplifies making HTTPS GET requests without manually configuring request options.
 
-Because `get` is a wrapper around `fetch(...)`, it may be used only within an exported
-fetch or similar handler. Outside of such a handler, attempts to use `get` will throw
-an error.
-
```
jasnell, line 19: https://github.com/cloudflare/cloudflare-docs/pull/23808 made a number of changes to these docs and was merged last week. It looks like you're simply undoing a good portion those changes here. How come?

src/content/docs/workers/runtime-apis/nodejs/http.mdx
jasnell: Also, not sure why the example needed to be changed here tho. I prefer the simpler version without the asserts.

jasnell (approved):

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -49,83 +87,106 @@ when the handler returns.
...
-* `Content-Length` headers will be handled the same way as in the `fetch` API. If a body is provided, the header will be set automatically and manually set values will be ignored.
-* `Expect: 100-continue` headers are not supported.
-* Trailing headers are not supported.
-* The `'continue'` event is not supported.
-* The `'information'` event is not supported.
-* The `'socket'` event is not supported.
-* The `'upgrade'` event is not supported.
-* Gaining direct access to the underlying `socket` is not supported.
-* Configuring TLS-specific options like `ca`, `cert`, `key`, `rejectUnauthorized`, etc, is not supported.
+- Connection management methods such as `closeAllConnections()` and `closeIdleConnections()` are not implemented
```
irvinebroque: Can we say _why_? (i.e. because you don't have to do xyz on Workers)

src/content/docs/workers/runtime-apis/nodejs/https.mdx
```diff
@@ -49,83 +87,106 @@ when the handler returns.
...
+- Only `listen()` variants with a port number or no parameters are supported: `listen()`, `listen(0, callback)`, `listen(callback)`, etc.
+- The following server options are not supported: `maxHeaderSize`, `insecureHTTPParser`, `keepAliveTimeout`, `connectionsCheckingInterval`
+- TLS/SSL-specific options such as `ca`, `cert`, `key`, `pfx`, `rejectUnauthorized`, `secureProtocol` are not supported in the Workers environment
```
irvinebroque:
```suggestion
- TLS/SSL-specific options such as `ca`, `cert`, `key`, `pfx`, `rejectUnauthorized`, `secureProtocol` are not supported in the Workers environment. If you need to use mTLS, use the [mTLS binding](/workers/runtime-apis/bindings/mtls/).
```

irvinebroque (approved):
✅ so you have it, few comments, @mikenomitch

src/content/docs/workers/runtime-apis/nodejs/http.mdx
anonrig: It didn't needed to be changed but because there was a conflict, it was hard to see what was there at the first place, and what wasn't. I'll take a look at it once again.

src/content/docs/workers/runtime-apis/nodejs/https.mdx
anonrig: Because of a conflict. Reverted the changes.

mikenomitch:
Some weird rendering here: https://yagiz-add-nodejs-http-https-modules.preview.developers.cloudflare.com/workers/runtime-apis/nodejs/http/#incomingmessage

src/content/docs/workers/runtime-apis/nodejs/http.mdx
```diff
@@ -129,7 +152,37 @@ export default {
 		// ...
 	}
 }
-```
+
```
mikenomitch, line 155: Missing ```

PR #24279, opened 2025-08-08
Clarify environments

irvinebroque:
Using capitalizeed "Cloudflare Environments" is a bit confusing since it's not a named product


anonrig (approved):

petebacondarwin (approved):
"Environments" is so overloaded. There are at least three different types: Cloudflare environments (these), Process environments (e.g. `process.env`), Vite environments (aka `modes`). In the Vite docs we call the first [Cloudflare Environments](https://developers.cloudflare.com/workers/vite-plugin/reference/cloudflare-environments/).

But happy to remove that here.

kodster28 (approved):

PR #24314, opened 2025-08-11
[DNS] Update index.mdx

ngayerie:
Following case 01688018


RebeccaTamachiro:
Thanks, Nic! Moving to PCX-18776

PR #24429, opened 2025-08-14
Remove calls to sql.end() in Hyperdrive examples

a-robinson:
As discussed, the workers runtime already handles closing outgoing connections as soon as a worker invocation finishes. And we have reason to believe that sql.end() can cause hangs due to a bug in postgres.js (https://github.com/porsager/postgres/issues/1097), so it's better to just not include it.

I guess I didn't have to remove it from our node-postgres examples, but it isn't needed there either, so why clutter our examples with extra code?

@thomasgauvin @ReppCodes


ReppCodes (commented):
Thanks, Alex!

Oxyjun (approved):

PR #24446, opened 2025-08-15
Updating content on outstanding balances

simon-says:
### Summary

Renaming the doc to align with the dashboard error message wording, adding more detail on the error strings that appear in the dashboard relating to this problem.

### DIFF
--- a/src/content/docs/billing/pay-invoices-overdue-balances.mdx
+++ b/src/content/docs/billing/pay-invoices-overdue-balances.mdx
@@ -1,36 +1,46 @@
 ---
-title: Pay invoices and overdue balances
+title: Pay an outstanding balance
 pcx_content_type: overview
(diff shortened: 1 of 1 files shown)


dcpena:
Hey @simon-says, let's go ahead and just change the title in the frontmatter of the file to "Pay an outstanding balance" instead of renaming the entire file and adding a redirect. Changing the file would have minimal impact on SEO and not quite worth the redirect hassle. 😄

simon-says:
> Hey @simon-says, let's go ahead and just change the title in the frontmatter of the file to "Pay an outstanding balance" instead of renaming the entire file and adding a redirect. Changing the file would have minimal impact on SEO and not quite worth the redirect hassle. 😄

OK - done :)

dcpena (approved):

PR #24453, opened 2025-08-15
[ResourcesBySelector] Error if no resources found

kodster28:
### Summary

fix root cause of #24451


src/components/ResourcesBySelector.astro
```diff
@@ -52,6 +52,12 @@ const resources: Array<CollectionEntry<"docs"> | CollectionEntry<"stream">> = [
 	);
 });
 
+if (resources.length < 1) {
```
KianNH:
```suggestion
if (resources.length === 0) {
```

KianNH (approved):

Oxyjun (approved):

PR #24548, opened 2025-08-20
[WAF, Rules] Remove single-use partials

pedrosousa:


src/content/docs/fundamentals/manage-domains/redirect-domain.mdx
```diff
@@ -21,4 +20,31 @@ If you have an alias domain that only forwards traffic to another domain (that i
 
 3. Use [Redirect rules](/rules/url-forwarding/) to forward traffic from your alias domain to your other domain.
 
-   <Render file="url-forwarding/different-hostname-example" product="rules" />
+   This example single redirect will redirect all requests for `smallshop.example.com` to a different hostname using HTTPS, keeping the original path and query string.
```
Maddy-Cloudflare:
```suggestion
This example will redirect all requests for `smallshop.example.com` to a different hostname using HTTPS, keeping the original path and query string.
```

Maddy-Cloudflare: The sentence: 

`This example single redirect will redirect`

Is not clear to me, so I changed it. But let me know if the meaning stays the same.

Maddy-Cloudflare (approved):

PR #24682, opened 2025-08-26
Adjust awkward spacing in sql string literal

chris-martin:
I offer this change as an excuse to inquire: Is there a reason this example is written this way, to prefer starting one SQL query on the same line as the previous?


Oxyjun (approved):

Oxyjun:
@chris-martin not particularly - we try to ensure code blocks are formatted consistently, but can end in differences due to different plugins / settings, etc (given that our docs are open-source, and we're a small team). Thanks for catching these!

PR #24684, opened 2025-08-26
[Docs Site] Add WranglerCommands component

KianNH:
### Summary

Add WranglerCommands component


WalshyDev (approved):

kodster28 (approved):
The only tweak here is that I'd rename the component (and then the style guide page) to `WranglerCommandsByNamespace` or something like that.

Realize that's a touch long, but it'd help prevent typos / confusion with the other one (which is only an `s` away).

PR #24731, opened 2025-08-27
Fix AutoRAG sub-folder filter

abegehr:
I tried the filter with `gt: "user/1//"`, however that didn't yield any files from "user/1/". When I exclude the second "/", it works as expected. I think the issue is that for files directly in the folder, the `gt` with trailing "/" doesn't catch, since the folder is exactly "user/1/", not "user/1/…".

Sample `response.data` from AutoRAG:

```ts
  [
    {
      file_id: 'cb9b94f5-cc24-410f-97b0-8a1d7eb76613',
      filename: 'user/1/summer_clothing_line.csv',
      score: 0.642221,
      attributes: {
          timestamp: 1756299780000,
          folder: "users/1/",
          filename: "summer_clothing_line.csv"
      },
      content: […]
    }
  ]
```

### Summary

Fix the ["Starts with" filter for folders](https://developers.cloudflare.com/autorag/configuration/metadata/#starts-with-filter-for-folders) section for files directly in the folder.

### Documentation checklist

### DIFF
--- a/src/content/docs/autorag/configuration/metadata.mdx
+++ b/src/content/docs/autorag/configuration/metadata.mdx
@@ -113,7 +113,7 @@ filters: {
     type: "and",
     filters: [
       {
-        type: "gt",
+        type: "gte",
         key: "folder",
         value: "customer-a/",
       },
@@ -129,7 +129,7 @@ filters: {
 This filter identifies paths starting with `customer-a/` by using:
 
 - The `and` condition to combine the effects of the `gt` and `lte` conditions.
-- The `gt` condition to include paths greater than the folder.
+- The `gte` condition to include paths greater than and including the folder itself.
 - The `lte` condition to include paths less than and including the lower case `z` ASCII character.
 
 Together, these conditions effectively select paths that begin with the provided path value.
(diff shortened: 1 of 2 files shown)


abegehr:
In the sample from the docs, with the following folder structure:

Screenshot 2025-08-27 at 3 42 51 PM

and using the mentioned filters:

```ts
filters: {
    type: "and",
    filters: [
      {
        type: "gt",
        key: "folder",
        value: "customer-a//",
      },
      {
        type: "lte",
        key: "folder",
        value: "customer-a/z",
      },
    ],
  },
```

it wouldn't include the `profile.md` file.

To include the `profile.md` file, `gt` needs to be "customer-a/" (single "/", not double):

```ts
filters: {
    type: "and",
    filters: [
      {
        type: "gt",
        key: "folder",
        value: "customer-a/",
      },
      {
        type: "lte",
        key: "folder",
        value: "customer-a/z",
      },
    ],
  },
```

abegehr:
UPDATE: It needs to be `gte` actually, to include the folder itself also.

PR #24763, opened 2025-08-28
Prices should be per input for audio input models

johnspurlock-skymethod:


mchenco:
thanks for catching this!

mchenco (approved):

PR #24787, opened 2025-08-29
D1 tutorial: Clarify that bindings step is automated by CLI

hoodmane:
### Summary

If you used the CLI in step 2, it says 
```
✔ Would you like Wrangler to add it on your behalf? › 
```
and if you say yes (the default) it automatically updates the config. So then I get to step 3 and there is nothing to do. I think this helps clarify.


hoodmane:
Actually, this is misleading because the autogen'd binding is created with name `prod_d1_tutorial` but the tutorial assumes it will be called `DB`.

PR #24790, opened 2025-08-29
Fix uses of env and ctx in Python examples

hoodmane:
### Summary

Since #24392, most of the Python examples use env and ctx incorrectly. This fixes it.

I also replaced all tabs with spaces in signing-requests.mdx.

### DIFF
--- a/src/content/docs/workers/examples/signing-requests.mdx
+++ b/src/content/docs/workers/examples/signing-requests.mdx
@@ -388,74 +388,74 @@ EXPIRY = 60
 
 class Default(WorkerEntrypoint):
     async def fetch(self, request):
-		# Get the secret key
-		secret_key_data = encoder.encode(self.env.SECRET_DATA if hasattr(self.env, "SECRET_DATA") else "my secret symmetric key")
+        # Get the secret key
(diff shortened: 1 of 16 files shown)


src/content/docs/workers/examples/signing-requests.mdx
```diff
@@ -61,16 +61,16 @@ export default {
 		// Refer to https://developers.cloudflare.com/workers/configuration/secrets/
 		const secretKeyData = encoder.encode(
 			env.SECRET_DATA ?? "my secret symmetric key",
-		);
+        );
```
dom96: some incorrect formatting here?

dom96 (approved):

src/content/docs/workers/examples/signing-requests.mdx
hoodmane: There were tabs instead of spaces. I converted all tabs in the file to 4 spaces.

dom96: I think there is still some tabs in there then, because the code isn't aligned properly when viewed on GitHub

Oxyjun (approved):

PR #24903, opened 2025-09-03
Update remote bindings docs in regards to RPC now being supported

dario-piotrowicz:
### Summary

RPC is now supported for remote bindings (https://github.com/cloudflare/workers-sdk/pull/10249), so here I am removing the item from the unsupported bindings list 🚀


penalosa (approved):
🎉

PR #24960, opened 2025-09-05
[DNS] Update index.mdx

ngayerie:
Following support case 01729746

Should we clarify that Internal DNS is not the correct solution to get local DNS records to resolve externally?


RebeccaTamachiro:
Closing as per PCX-19155 and #25038

PR #24984, opened 2025-09-08
[Docs Site] Create WorkersTypesInterface component

KianNH:
### Summary

Create WorkersTypesInterface component


kodster28 (changes requested):
Love this as a demo PR, but we don't want to merge this officially until we'd have a source of truth to pull from.

Otherwise, we're setting ourselves up for some pain here.

PR #24993, opened 2025-09-08
[Workers] Nav changes - batch 1

ToriLindsay:
Adding DashButton component for new nav


deadlypants1973 (approved):
Are we saying "xyz page" or just "go to XYZ"?

PR #25044, opened 2025-09-09
Default code block title -- PCX-16857

kodster28:
### Summary

Add default language titles to more code blocks per PCX-16857

### DIFF
--- a/src/plugins/expressive-code/default-titles.js
+++ b/src/plugins/expressive-code/default-titles.js
@@ -10,26 +10,17 @@ export default () => {
 						context.codeBlock.props.title ??= "PowerShell";
 						break;
 					}
-					case "javascript": {
-						context.codeBlock.props.title ??= "JavaScript";
-						break;
-					}
+					case "javascript":
 					case "js": {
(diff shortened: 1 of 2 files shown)


src/plugins/expressive-code/default-titles.js
```diff
@@ -10,6 +10,30 @@ export default () => {
...
+						break;
+					}
+					case "typescript": {
+						context.codeBlock.props.title ??= "TypeScript";
+						break;
+					}
+					case "ts": {
+						context.codeBlock.props.title ??= "TypeScript";
+						break;
+					}
```
KianNH:
```suggestion
					case "javascript":
					case "js": {
						context.codeBlock.props.title ??= "JavaScript";
						break;
					}
					case "py":
					case "python": {
						context.codeBlock.props.title ??= "Python";
						break;
					}
					case "typescript":
					case "ts": {
						context.codeBlock.props.title ??= "TypeScript";
						break;
					}
```

KianNH (approved):

PR #25088, opened 2025-09-11
[Rules] Update how-to.mdx

luishcardoso:
Hello team, 

### Summary

To avoid mentioning "Go to Trace" twice in the same menu.

Thanks!


pedrosousa:
Currently that's the way it's implemented across all docs, even though it's somewhat repeated (but it's consistent. 
If we want to change it, we'll do it in more than just a single place.

PR #25150, opened 2025-09-15
Update node.js compat supported matrix

jasnell:
Builds on the nodejs-fs PR, which should land first
https://github.com/cloudflare/cloudflare-docs/pull/24082/files

### DIFF
--- a/src/content/docs/workers/runtime-apis/nodejs/index.mdx
+++ b/src/content/docs/workers/runtime-apis/nodejs/index.mdx
@@ -11,7 +11,10 @@ When you write a Worker, you may need to import packages from [npm](https://www.
 
 Cloudflare Workers provides a subset of Node.js APIs in two forms:
 
-1. As built-in APIs provided by the Workers Runtime
+1. As built-in APIs provided by the Workers Runtime. Most of these APIs are
+   full implementations of the corresponding Node.js APIs, while a few are
+	 partially supported or non-functional stubs intended for the APIs to be
+	 available for import only but not for actual use.
 2. As polyfill shim implementations that [Wrangler](/workers/wrangler/) adds to your Worker's code, allowing it to import the module, but calling API methods will throw errors.
 
 ## Get Started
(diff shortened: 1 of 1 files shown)


src/content/changelog/workers/2025-08-15-nodejs-fs.mdx
```diff
@@ -0,0 +1,60 @@
...
+    const writer = writable.getWriter();
+		await writer.write('Hello, World!');
+		await writer.close();
+
+		return new Response('File written successfully!');
+	}
+}
+```
+
+As there are still some parts of the Web File System API tht are not fully standardized, there may be some differences between the Workers implementation and the implementations in browsers.
```
vicb, line 60:
```suggestion
As there are still some parts of the Web File System API that are not fully standardized, there may be some differences between the Workers implementation and the implementations in browsers.
```

src/content/docs/workers/runtime-apis/nodejs/fs.mdx
```diff
@@ -0,0 +1,140 @@
...
+	data written to it. Reading from `/dev/random` is only permitted when within
+	the context of a request.
+* `/dev/full`: A device that always returns EOF on reads and discards all data
+	written to it.
+* `/dev/zero`: A device that provides an infinite stream of zero bytes on reads
+	and discards all data written to it.
+
+All operations on the VFS are synchronous. You can use the synchronous,
+asynchronous callback, or promise-based APIs provided by the `node:fs` module
+but all operations will be performed synchronously.
```
vicb, line 105: I am wondering if this could be confusing for users. I think what should matter most to them is that they can use the sync or async versions of the the API.

I see the fact that the underlying impl is synchronous more as an implementation detail.

But maybe I miss something?

vicb (approved):
LGTM with a couple inline comment.

Should the doc details what the cwd is set to?

src/content/changelog/workers/2025-08-15-nodejs-fs.mdx
jasnell: This is not the pr for the fs details... it just happens to include that PR since this PR builds on that one. Review comments for the fs pr should be moved to https://github.com/cloudflare/cloudflare-docs/pull/24082

vicb (approved):

PR #25234, opened 2025-09-17
[Changelog] Fix multi-product entries bug per PCX-19256

kodster28:
### Summary

Tweak to behavior introduced in #23975


KianNH (approved):

EHESPO:
@kodster28 Thanks for the fix and the follow-up to #23975. The changelog behavior looks clearer with the multi-product entry issue addressed in #25234. Reviewed from the EHESPO side — this looks good.

PR #25285, opened 2025-09-19
Changelog for workers-rs panic recovery

guybedford:
### Summary

Provides a technical changelog for the new workers-rs panic recovery functionality.

It's quite nice to go into the details here, but we could cut back on the content too if this is the wrong place for it.

### Documentation checklist

### DIFF
--- a/src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
+++ b/src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
@@ -1,6 +1,6 @@
 ---
 title: Panic Recovery for Rust Workers
-description: Making workers-rs deployments more reliable with panic recovery
+description: Deployments on workers-rs are now more reliable with improved panic recovery
 date: 2025-09-19
 ---
 
@@ -8,17 +8,13 @@ import { WranglerConfig, Aside } from "~/components";
 
 A long-standing problem we've had with [workers-rs](https://github.com/cloudflare/workers-rs) is that its Rust panics are non-recoverable, and can put the worker into an invalid state.
 
-Working on the [wasm-bindgen](https://github.com/wasm-bindgen/wasm-bindgen) internals, we've been able to implement a solution for panic recovery to ensure more reliable deployments.
+We've been able to implement a solution for panic recovery to ensure more reliable deployments. This new automatic panic recovery feature is enabled for all new workers-rs deployments, as of version 0.6.5, with no further configuration required.
 
 ## Fixing Rust Panics with Wasm Bindgen
 
-We use [wasm-bindgen](https://github.com/wasm-bindgen/wasm-bindgen) to build and deploy Rust workers, but Wasm Bindgen is not designed to support recoverable panics.
+Rust Workers are built with Wasm Bindgen, which treats panics as non-recoverable - the entire Wasm application is considered to be in an invalid state, and any further function calls could result in overflows or memory exceptions.
 
-When a panic happens in Wasm Bindgen, the entire Wasm application is considered to be in an invalid state, and any further function calls could result in overflows or memory exceptions.
-
-But Wasm is a virtual machine - if we can reset all the Wasm internal state back to its initial state then we can continue to take new requests.
-
-It's possible to add a panic handler in Rust to detect immediately when a panic happens:
+We now attach a default panic handler in Rust:
 
 ```rust
 std::panic::set_hook(Box::new(move |panic_info| {
@@ -26,7 +22,7 @@ std::panic::set_hook(Box::new(move |panic_info| {
 }));
 ```
 
-With some extra wiring we can then export a registration function to JavaScript to allow a custom panic hook to be attached:
+Which is registered by default in the JS initialization:
 
 ```js
 import { setPanicHook } from "./index.js";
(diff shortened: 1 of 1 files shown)


irvinebroque (commented):
This is awesome and think it is good to get into this kind of detail here

danlapid (approved):
LOVE IT

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
```diff
@@ -0,0 +1,69 @@
+---
+title: Panic Recovery for Rust Workers
+description: Making workers-rs deployments more reliable with panic recovery
```
mikenomitch: nit: I think tense on this is slightly different from other changelogs - IDK if this matters or not but suggesting for consistency

```suggestion
description: Deployments on workers-rs are now more reliable with panic recovery
```

mikenomitch:
```suggestion
description: Deployments on workers-rs are now more reliable with improved panic recovery
```

Nitpick: Just making the tense of this a bit more consistent w the others.

mikenomitch (changes requested):
@guybedford this feels more like part of a blog post or docs than a Changelog.

Instead of going into the details (at least initially), I'd make it much more to the point:
- Previously, when in workers-rs when a Rust panic occured, X would happen. This was bad because the Worker would then do Y.
- Now you can add the new `setPanicHook` functionality to handle these errors.

I thin we can and should go a little into the details after this, but I'd probably try to trim it down a bit from where its at.

Also, I see setPanicHook being imported from index.js. It's not clear to me where it actually comes from and how to implement. Is there a way to add some more color there?

EDIT: Brendan and Dan seemed more into including the details here than I was. 😄 - I'm okay keeping what we have as long as they don't interfere with immediately defining the problem and how we now solve it. So I'd push the wasm-bindgen bits down below, and explain "Wasm is a virtual machine" later too, and then I think its good.

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
```diff
@@ -0,0 +1,69 @@
...
+```rust
+std::panic::set_hook(Box::new(move |panic_info| {
+  hook_impl(panic_info);
+}));
+```
+
+With some extra wiring we can then export a registration function to JavaScript to allow a custom panic hook to be attached:
+
+```js
+import { setPanicHook } from "./index.js";
```
mikenomitch, line 30: Where is this actually defined? - Not sure as a user where I go to do this.

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Fixing Rust Panics with Wasm Bindgen
+
+We use [wasm-bindgen](https://github.com/wasm-bindgen/wasm-bindgen) to build and deploy Rust workers, but Wasm Bindgen is not designed to support recoverable panics.
+
+When a panic happens in Wasm Bindgen, the entire Wasm application is considered to be in an invalid state, and any further function calls could result in overflows or memory exceptions.
+
+But Wasm is a virtual machine - if we can reset all the Wasm internal state back to its initial state then we can continue to take new requests.
+
+It's possible to add a panic handler in Rust to detect immediately when a panic happens:
```
mikenomitch: I would be more direct here.

`You can now add panic handlers in Rust, like so:`

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
mikenomitch: I think the "With some extra wiring" is doing some work behind the scenes here we need to make explicit.

If its too long to include, we should add a docs link here.

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
guybedford: Added.

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
guybedford: I've now clarified that this is on by default.

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
guybedford: Let me know if the new clarifications help here.

src/content/changelog/workers/2025-09-19-workers-rs-panic-recovery.mdx
```diff
@@ -0,0 +1,61 @@
...
+```
+
+When a panic occurs, we then reset the Wasm state to reinitialize the Wasm application back to as it was when the application first started.
+
+## Resetting VM State in Wasm Bindgen
+
+We worked upstream on the Wasm Bindgen project to implement a new [`--experimental-reset-state-function` compilation option](https://github.com/wasm-bindgen/wasm-bindgen/pull/4644) which outputs a new `__wbg_reset_state` function.
+
+This function clears all internal state related to the Wasm VM, and also ensures object references are uniquely associated with the Wasm instance identity. Wasm bindgen exports stateless JS wrapper functions which call into Wasm. Updating their internal
+Wasm instance binding to the new instance allows exposing the new Wasm instance without having to rebind the exported functions.
```
guybedford:
```suggestion
This function clears all internal state related to the Wasm VM, and updates all function bindings in place to reference the new WebAssembly instance.
```

mikenomitch (approved):

PR #25394, opened 2025-09-24
[SSL] Correcting AOP doc

ngayerie:
PCX-19386

In the zone level doc:
remove the step #4 as the PUT is not needed if you enable on the UI

In the hostname level doc:
remove the step #3 as you don't need to enable AOP for the zone if you enable it per hostname with the API


RebeccaTamachiro:
Checking with Product

RebeccaTamachiro (approved):

PR #25403, opened 2025-09-24
[ZT] clarify Windows pre-login behavior

ranbel:
PCX-16003


ranbel:
@aw-cf

aw-cf:
@cf-rhett

cf-rhett:
This looks correct to me -- a bit confusing but that's more our fault than anything. The differing behavior with multi-user enabled vs disabled is non-intuitive and makes documentation hard. Might be worth pushing for some sort of reconciliation between the behavior

maxvp (approved):

PR #25426, opened 2025-09-25
remove sql.end/client.end()

thomasgauvin:


Oxyjun (approved):

irvinebroque (approved):
should we _tell_ people not to do this explicitly?

thomasgauvin:
> should we _tell_ people not to do this explicitly?

The memory of the `fetch` handler context is purged on every invocation, so Workers takes care of closing I/O context and therefore db connections (in the Worker memory). 

It is ok to run ctx.waitUntil(sql.end()) which will close the connection between the Worker isolate and Hyperdrive, but we have had some folks report that this was causing issues, partly because of [driver bugs](https://github.com/porsager/postgres/issues/1097) that we've reported. To avoid this noise, we are not documenting it (and it's also simpler, as effective).

We are going to be writing a new doc section on how connections between Worker, Hyperdrive and the origin database work to clarify some of these things.

PR #25439, opened 2025-09-26
[ZT] Deploy WARP on headless Linux

ranbel:
PCX-4105

### DIFF
--- a/src/content/docs/cloudflare-one/tutorials/warp-on-headless-linux.mdx
+++ b/src/content/docs/cloudflare-one/tutorials/warp-on-headless-linux.mdx
@@ -113,4 +113,4 @@ To install WARP using the example script:
 	sudo ./install_warp.sh
 	```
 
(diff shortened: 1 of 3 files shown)


src/content/docs/cloudflare-one/tutorials/warp-on-headless-linux.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+		```bash
+		#!/bin/bash
+		set -e
+
+		# Download and install the WARP client
+		function warp() {
+				curl -fsSL https://pkg.cloudflareclient.com/pubkey.gpg | sudo gpg --yes --dearmor --output /usr/share/keyrings/cloudflare-warp-archive-keyring.gpg
+				echo "deb [signed-by=/usr/share/keyrings/cloudflare-warp-archive-keyring.gpg] https://pkg.cloudflareclient.com/ $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/cloudflare-client.list
+				sudo apt-get update --assume-yes
```
cf-rhett, line 62: This will handle ubuntu & debian but won't handle RHEL and other distros that don't use apt-get but yum or whatever

patriciasantaana (approved):

PR #25556, opened 2025-10-01
remove text stating we have no support for heic

dochne:
### Summary

While we will continue to not support outputting HEIC, we now support the input format of HEIC which makes this sentence somewhat misleading.


ToriLindsay (approved):

ToriLindsay:
cc: @deanna-cf

PR #25605, opened 2025-10-03
update Wrangler env-vars to include WRANGLER_LOG_SANITIZE and WRANGLER_SEND_ERROR_REPORTS

petebacondarwin:
### Summary

Add docs about the new WRANGLER_SEND_ERROR_REPORTS env-var for Wrangler.

This was only really added to help non-interactive tooling (such as AIs) avoid having to deal with unwanted confirmation dialogs.

### Documentation checklist

---

<a href="https://app.devin.ai/review/cloudflare/cloudflare-docs/pull/25605" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/content/docs/workers/wrangler/system-environment-variables.mdx
+++ b/src/content/docs/workers/wrangler/system-environment-variables.mdx
@@ -61,8 +61,11 @@ Wrangler supports the following environment variables:
 - `NODE_ENV` <Type text="string" /> <MetaInfo text="optional" />
   - Sets the value of `process.env.NODE_ENV` in your Worker code. Defaults to `"development"` for `wrangler dev` and `"production"` for `wrangler deploy` and `wrangler versions upload`. Refer to [Bundling](/workers/wrangler/bundling/#node_env) for more information.
 
-- `WRANGLER_SEND_METRICS` <Type text="string" /> <MetaInfo text="optional" />
+- `WRANGLER_SEND_METRICS` <Type text="boolean" /> <MetaInfo text="optional" />
   - Options for this are `true` and `false`. Defaults to `true`. Controls whether Wrangler can send anonymous usage data to Cloudflare for this project. You can learn more about this in our [data policy](https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/telemetry.md).
+
+- `WRANGLER_SEND_ERROR_REPORTS` <Type text="boolean" /> <MetaInfo text="optional" />
+  - Options for this are `true` and `false`. Defaults to `undefined`. Controls whether Wrangler can send non-user error reports to Cloudflare for this project. If `undefined`, Wrangler will ask the user whether to send an error report each time there is a non-user error.
 
 - `CLOUDFLARE_HYPERDRIVE_LOCAL_CONNECTION_STRING_<BINDING_NAME>`<Type text="string" /> <MetaInfo text="optional" />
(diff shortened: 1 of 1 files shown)


irvinebroque:
> to help non-interactive tooling (such as AIs) avoid having to deal with unwanted confirmation dialogs

So we are saying that AI tools are going to set this env var themselves? What's an example?

petebacondarwin:
> > to help non-interactive tooling (such as AIs) avoid having to deal with unwanted confirmation dialogs
> 
> So we are saying that AI tools are going to set this env var themselves? What's an example?

The motivation came from this issue - https://github.com/cloudflare/workers-sdk/issues/10719

petebacondarwin:
This is not stale (from my side) - Just waiting on a review.

MattieTK:
I think it would be sensible to let https://github.com/cloudflare/cloudflare-docs/pull/28364 merge first so there's no conflicts

src/content/docs/workers/wrangler/system-environment-variables.mdx
```diff
@@ -76,6 +79,9 @@ Wrangler supports the following environment variables:
 - `WRANGLER_LOG_PATH` <Type text="string" /> <MetaInfo text="optional" />
   - A file or directory path where Wrangler will write debug logs. If the path ends in `.log`, Wrangler will consider this the path to a file where all logs will be written. Otherwise, Wrangler will treat the path as a directory where it will write one or more log files using a timestamp for the filenames.
 
+- `WRANGLER_LOG_SANITIZE` <Type text="boolean" /> <MetaInfo text="optional" />
+  - Options for this are `true` and `false`. Defaults to `true`. Controls whether Wrangler will sanitize any sensitive information from logs written to the console or to a log file. Sensitive information includes API tokens, email addresses, account IDs, and more.
```
MattieTK, line 83: Is there somewhere we document what 'sensitive information' is? If not maybe that's something we should do in the docs and then link to. Thoughts?

src/content/docs/workers/wrangler/system-environment-variables.mdx
```diff
@@ -61,9 +61,12 @@ Wrangler supports the following environment variables:
 - `NODE_ENV` <Type text="string" /> <MetaInfo text="optional" />
   - Sets the value of `process.env.NODE_ENV` in your Worker code. Defaults to `"development"` for `wrangler dev` and `"production"` for `wrangler deploy` and `wrangler versions upload`. Refer to [Bundling](/workers/wrangler/bundling/#node_env) for more information.
 
-- `WRANGLER_SEND_METRICS` <Type text="string" /> <MetaInfo text="optional" />
+- `WRANGLER_SEND_METRICS` <Type text="boolean" /> <MetaInfo text="optional" />
```
MattieTK, line 64: For my own learning: I'm guessing this is safe to change because there's no real difference between a string true/false and a boolean in how these are retrieved?

MattieTK (approved):

src/content/docs/workers/wrangler/system-environment-variables.mdx
petebacondarwin: Here is the code that processes boolean env vars: 

https://github.com/cloudflare/workers-sdk/blob/e0d854be649fa53fee2c95f0a79400e6fd92d08f/packages/workers-utils/src/environment-variables/factory.ts#L184-L196

You can see that it is definitely different to just a "string".

In terms of this docs change, this is just aligning the docs with what actually happens in the code.

src/content/docs/workers/wrangler/system-environment-variables.mdx
petebacondarwin: I can't find any docs around this term...

PR #25615, opened 2025-10-03
[Docs site] Flag unslashed source urls in redirects

kodster28:
### Summary

Fix common issue where we have unslashed redirect sources that lead to confusion (b/c we always end slash the URLs of our site).

### DIFF
--- a/bin/validate-redirects.ts
+++ b/bin/validate-redirects.ts
@@ -6,6 +6,7 @@ async function main() {
 	let numInfiniteRedirects = 0;
 	let numUrlsWithFragment = 0;
 	let numDuplicateRedirects = 0;
+	let numNonSlashedRedirects = 0;
 
 	const redirectSourceUrls: string[] = [];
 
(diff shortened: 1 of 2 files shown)


bin/validate-redirects.ts
```diff
@@ -6,6 +6,7 @@ async function main() {
 	let numInfiniteRedirects = 0;
 	let numUrlsWithFragment = 0;
 	let numDuplicateRedirects = 0;
+	let numNonSlashedRedirects = 0;
```
kodster28, line 9:
```suggestion
	let numNonSlashedRedirects = 0;
	
	const validEndings = ["/", "*", ".xml", ".md", ".json", ".html"];
```

bin/validate-redirects.ts
```diff
@@ -24,6 +25,11 @@ async function main() {
 			numUrlsWithFragment++;
 		}
 
+		if (!from.endsWith("/") && !from.endsWith("*") && !from.endsWith(".xml")) {
```
kodster28:
```suggestion

		if (!validEndings.some(ending => from.endsWith(ending))) {
```

haleycode (approved):

PR #25650, opened 2025-10-06
workers: Request.clone does not return a Promise

MattIPv4:
### Summary

Incorrect type in docs for Request.clone result.


vicb (approved):
Thanks!

PR #25719, opened 2025-10-08
[ZT] cURL cert instructions

maxvp:
PCX-18818

### DIFF
--- a/src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/manual-deployment.mdx
+++ b/src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/manual-deployment.mdx
@@ -482,6 +482,26 @@ To configure Git to trust a Cloudflare certificate, run the following command:
 git config --global http.sslcainfo [PATH_TO_CLOUDFLARE_CERT]
 ```
 
+### cURL
+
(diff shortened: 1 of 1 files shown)


src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/manual-deployment.mdx
```diff
@@ -482,6 +482,26 @@ To configure Git to trust a Cloudflare certificate, run the following command:
 git config --global http.sslcainfo [PATH_TO_CLOUDFLARE_CERT]
 ```
 
+### cURL
+
+By default, cURL will use your system's native certificate store. To force cURL to use your default certificate, add the the `--ca-native` flag. For example:
```
ranbel: is the "native store" the same thing as the operating system store mentioned below? Do you need to use the --ca-native flag AND add the cert to macOS/Windows/Linux?

ranbel (approved):

src/content/docs/cloudflare-one/connections/connect-devices/user-side-certificates/manual-deployment.mdx
maxvp: Yes, it's the same thing, but different operating systems call it different things (certificate store, keychain, etc.). I'll update the language to be more clear.

PR #25753, opened 2025-10-10
Update images-workers.mdx

DavidJKTofan:
PCX-9970

Missing `/` in the path


pedrosousa:
Each path segment in `pathname` will always have a `/` prefix, including the first one, so this change would add two slashes:
https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

DavidJKTofan:
That makes sense now. Thank you for clarifying! We can close this then.

PR #25769, opened 2025-10-10
Add checks for POLICY_AUD

jillesme:
### Summary

I followed https://developers.cloudflare.com/cloudflare-one/identity/authorization-cookie/validating-json/#cloudflare-workers-example without setting the AUD environment variable on accident. That ends up omitting the audience verification entirely. It only verifies if the audience value is set to a truthy value. Even an empty string gets skipped (thanks JavaScript) 

I skipped to Go example, because instead of reading from an environment variable, the audience is hard-coded in. 

There is also https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/changelog/workers/2025-10-03-one-click-access-for-workers.mdx -- which has the same example. I did **not** update that yet because I don't know how we feel about retroactively updating changelogs. Happy to update that too. 

### Screenshots (optional)

CleanShot 2025-10-10 at 13 24 27@2x

### Documentation checklist

### DIFF
--- a/src/content/changelog/workers/2025-10-03-one-click-access-for-workers.mdx
+++ b/src/content/changelog/workers/2025-10-03-one-click-access-for-workers.mdx
@@ -34,6 +34,14 @@ import { jwtVerify, createRemoteJWKSet } from "jose";
 
 export default {
 	async fetch(request, env, ctx) {
+		// Verify the POLICY_AUD environment variable is set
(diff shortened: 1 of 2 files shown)


jroyal:
lgtm

kodster28 (approved):

jroyal (approved):

PR #25781, opened 2025-10-12
Add docs about workflow re-execution after hibernation

kevinsimper:
I was curious how local state was preserved across hibernations, so I tested it with a simple example using Math.random(). Turns out the entire workflow function re-executes from the start after hibernation, only step returns are preserved. It could have been that the state of the isolate was saved.                                                         

This isn't immediately obvious from the existing "Do not rely on state outside of a step" section, so I've added a new section that explicitly explains: 
 - The workflow function re-executes after hibernation 
 - Variables outside steps get new values (especially problematic with Math.random()) 
 - Loop counters need to be in step returns or they'll reset

I also added a loop example which there is none of in the examples right now 😄

### DIFF
--- a/src/content/docs/workflows/build/rules-of-workflows.mdx
+++ b/src/content/docs/workflows/build/rules-of-workflows.mdx
@@ -205,6 +205,80 @@ export class MyWorkflow extends WorkflowEntrypoint {
 ```
 </TypeScriptExample>
 
+### Workflow code re-executes after hibernation
+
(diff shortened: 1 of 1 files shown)


timfish:
Great addition. We had to work this out ourselves!

kevinsimper:
Still relevant :)

kevinsimper:
Still relevant :) @celso what do you think about this? 😄

kevinsimper:
Hi @elithrar , would you have time to take a look? Any feedback of what you think I wrote would be really cool 😄

kevinsimper:
A bit weird

kevinsimper:
Annoying with the stale bot

PR #25793, opened 2025-10-13
[Resources] Sort by optional date reviewed

kodster28:
### Summary

Small tweak

### DIFF
--- a/src/components/ResourcesBySelector.tsx
+++ b/src/components/ResourcesBySelector.tsx
@@ -156,7 +156,7 @@ export default function ResourcesBySelector({
 
 			return true;
 		})
-		.sort((a, b) => (b?.data?.reviewed || 100) - (a?.data?.reviewed || 100));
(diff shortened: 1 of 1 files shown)


src/components/ResourcesBySelector.tsx
```diff
@@ -72,83 +72,91 @@ export default function ResourcesBySelector({
...
+					}
 				}
 			}
-		}
 
-		return true;
-	});
+			return true;
+		})
+		.sort((a, b) => (b?.data?.reviewed || 100) - (a?.data?.reviewed || 100));
```
pedrosousa:
```suggestion
		.sort((a, b) => Number(b?.data?.reviewed ?? 100) - Number(a?.data?.reviewed ?? 100));
```
Similar to the cast done in https://github.com/cloudflare/cloudflare-docs/blob/production/src/components/ListTutorials.astro?plain=1#L36 .

Also, I believe it makes more sense to use the `??` operator here, to address null/undefined only and not any "falsy" value: https://javascript.info/nullish-coalescing-operator#comparison-with

pedrosousa (approved):

PR #25794, opened 2025-10-13
Python workers: Replace remaining examples with on_fetch with WorkerEntrypoint class

hoodmane:
Updating Python workers examples


dom96 (approved):

mikenomitch (approved):

src/content/docs/workflows/python/bindings.mdx
```diff
@@ -71,14 +71,17 @@ Create (trigger) a new instance of a given Workflow.
...
 
-async def on_fetch(request, env, ctx):
-    event = {"foo": "bar"}
-    options = to_js({"params": event}, dict_converter=Object.fromEntries)
-    await env.MY_WORKFLOW.create(options)
-    return Response.json({"status": "success"})
+class Default(WorkerEntrypoint):
+    async def fetch(self, request):
+        event = {"foo": "bar"}
+        options = to_js({"params": event}, dict_converter=Object.fromEntries)
```
Caio-Nogueira, line 82: Hmm this is correct at the moment. But since we recently started wrapping the binding in the workers module, we can do this internally (with a compat flag)?

dom96: yeah, it would be good to do that, we plan to do this for all bindings eventually

Caio-Nogueira:
We probably need to bump the compatibility date on the docs, right? We should have it set according to the `python_no_global_handlers` compat flag.

src/content/changelog/workflows/2025-08-22-workflows-python-beta.mdx
```diff
@@ -36,9 +36,10 @@ class PythonWorkflowStarter(WorkflowEntrypoint):
 
         await my_second_step()
 
-async def on_fetch(request, env):
-    await env.MY_WORKFLOW.create()
-    return Response("Hello Workflow creation!")
+class Default(WorkerEntrypoint):
```
Caio-Nogueira, line 39: We need to import `WorkerEntrypoint` as well

celso (approved):

PR #25807, opened 2025-10-13
Use pywrangler in the python overview

hoodmane:


src/content/docs/workers/languages/python/index.mdx
```diff
@@ -48,16 +50,23 @@ class Default(WorkerEntrypoint):
...
 
-To run a Python Worker locally, you use [Wrangler](/workers/wrangler/), the CLI for Cloudflare Workers:
+To run a Python Worker locally, you use pywrangler, the CLI for Python Workers.
+To set it up, first you need to set up your development environment:
+```
+uv init
+uv add --dev workers-py
+```
+This will create a `pyproject.toml` file with `workers-py` as a development
+dependency. You can then run `pywrangler` with:
```
dom96: instead of this I think we should instruct folks to use `pywrangler init` after running `uv tool install workers-py`

dom96: without doing this the deployment will fail because there will be no wrangler config file

dom96 (approved):

mikenomitch (approved):

PR #25823, opened 2025-10-14
[Rules] Update disable-all-brotli.mdx

luishcardoso:
Hello team, 

### Summary

Can we link here our relevant docs to Compression Rules?

https://developers.cloudflare.com/rules/compression-rules/

Thanks!


pedrosousa:
I don't think we need the link here. The example is already located in the Compression Rules documentation section, so it's a matter of going up one level.
Also, we currently have a bug where Markdown formatting isn't rendered properly in the [Examples](https://developers.cloudflare.com/rules/compression-rules/examples/) list page.

PR #25831, opened 2025-10-14
Refine Zone Deletion Conditions

tpaulus:
### Summary

This change is based on feedback from both customers and the DNS Team based on the internal logic used to determine when a zone is transitioned from Moved to Deleted.


tpaulus:
cc @svnrb

RebeccaTamachiro (approved):

tpaulus (commented):
LGTM. Thank you!

PR #25902, opened 2025-10-17
[Rules] Update index.mdx

luishcardoso:
Hello team,

### Summary

Can we link it to our webpage on Cloudflare plans?

https://www.cloudflare.com/en-gb/plans/

Thanks!


pedrosousa:
I don't think we should link to the plans page here.
- If we were pointing to a specific plan, this could make sense. With the feature being available to everyone, it's less relevant and not the best place to promote our paid plans or differentiate them.
- We use the expression "available on all Cloudflare plans" often, and in other occurrences we don't have any link.

PR #116, opened 2020-10-20
Add serve command to package.jsons

PeshekDotDev:

PR #25929, opened 2025-10-19
Removes sandbox start times

mikenomitch:
### Summary

We are reporting incorrect numbers for sandbox cold start times. Current start times are closer to ~3 seconds I believe.

Removing the sandbox performance until we can measure and report correct times.


mikenomitch:
Closes [#116](https://github.com/cloudflare/sandbox-sdk/issues/116)

whoiskatrin (approved):

PR #25959, opened 2025-10-21
Fix link locations for `wrangler versions upload` and `wrangler versions deploy` commands

dario-piotrowicz:
### Summary

Currently there is a bunch of links for `wrangler versions upload` and `wrangler versions deploy` that point to the wrong fragment of in the wrangler/commands documentation. In this PR I am fixing the wrong fragments so that the links point to the correct location instead.


emily-shen (commented):
if you search /workers/wrangler/commands/#upload in the docs there are another couple of broken links, could you also fix those while you're at it? thanks (i think i broke this when i updated the command to use the autogenerated component :'))

dario-piotrowicz:
> if you search /workers/wrangler/commands/#upload in the docs there are another couple of broken links, could you also fix those while you're at it? thanks (i think i broke this when i updated the command to use the autogenerated component :'))

done 😄

maxvp (approved):

PR #25962, opened 2025-10-21
Add changelog for experimental resource provisioning

penalosa:


src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,33 @@
...
+	},
+	"d1_databases": {
+		"binding": "MY_D1",
+	},
+}
+```
+
+When using this feature, the `id` field for KV and D1 and the `bucket_name` for R2 are optional in your wrangler configuration.
+
+When developing locally, you can interact with local versions of any KV, D1, R2 resources configured in your Wrangler configuration file, without having to include an id to existing resource
```
emily-shen:
```suggestion
When developing locally, you can interact with local versions of any KV, D1, R2 resources configured in your Wrangler configuration file, without having to include an id to existing resource.
```

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+Once ready to deploy, run `npx wrangler deploy`. Cloudflare will automatically create the resources that do not exist and bind them to your worker. Future deployments will use the same created resources.
```
emily-shen: we can also bind to existing resources right?

emily-shen (commented):
should we briefly mention the implication for template authors too?

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
MattieTK: Bit nitpicky. Do we want to be clear here here that 'binding' here means writing the created resources to the configuration file. Could be an addition to the final sentence eg.

> As part of this process, these bindings will be written to your wrangler configuration file. Future deployments of your worker will use these bindings.

penalosa: Not non-interactively, which is what's on by default

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,33 @@
...
+title: Automatic resource provisioning
+description: Wrangler can now automatically provision resources when deploying your Worker
+products:
+  - workers
+date: 2025-10-24
+---
+
+import { Render, TypeScriptExample, WranglerConfig } from "~/components";
+
+If your Wrangler configuration file includes a KV namespace, D1 database, or R2 bucket that does not yet exist on your account, you can still develop locally and deploy your application seamlessly, without having to run additional commands to create resources.
```
lrapoport-cf: maybe include a sentence that contrasts the before to the now, something like below (but could probably be made smoother):

```suggestion
Previously, if you wanted to develop or deploy a worker with attached resources, you'd have to first manually create the desired resources. Now, if your Wrangler configuration file includes a KV namespace, D1 database, or R2 bucket that does not yet exist on your account, you can still develop locally and deploy your application seamlessly, but without having to run additional commands first.
```

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,33 @@
+---
+title: Automatic resource provisioning
```
lrapoport-cf: should we narrow to "Automatic resource provisioning for KV, D1, and R2" or maybe put the specific resources in the description?

MattieTK:
> should we briefly mention the implication for template authors too?

Emily makes a good point here; we have templates also that would benefit from updates as a result of this change which we could follow up with in another PR. eg. https://github.com/cloudflare/templates/tree/main/chanfana-openapi-template which would no longer require the `database_id` or the instructions to change it.

Happy to do this before/after we merge.

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,33 @@
...
+		"binding": "MY_D1",
+	},
+}
+```
+
+When using this feature, the `id` field for KV and D1 and the `bucket_name` for R2 are optional in your wrangler configuration.
+
+When developing locally, you can interact with local versions of any KV, D1, R2 resources configured in your Wrangler configuration file, without having to include an id to existing resource.
+
+Once ready to deploy, run `npx wrangler deploy`. Cloudflare will automatically create the resources that do not exist and bind them to your worker. Future deployments will use the same created resources.
```
lrapoport-cf: i think the language you included in the changeset for https://github.com/cloudflare/workers-sdk/pull/11030 (slightly tweaked) would actually work really well here instead:

```suggestion
To use this feature, update to wrangler@<insert_version_number> and add bindings to your config file _without_ resource IDs e.g.:

```jsonc
{
    "kv_namespaces": [{ "binding": "MY_KV" }],
    "d1_databases": [{ "binding": "MY_DB" }],
    "r2_buckets": [{ "binding": "MY_R2" }]
}
```

`wrangler dev` will then automatically create these resources for you locally, and on your next run of `wrangler deploy`, Wrangler will call the Cloudflare API to create the requested resources and link them to your Worker. 

Though resource IDs will get automatically written back to your `.jsonc` after resource creation, resources will stay linked across future deploys even without adding the resource IDs to the config file. This is especially useful for shared templates, which now no longer need to include account-specific resource IDs when adding a binding.
```

File: src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx (hunk 2 above)
Comment (penalosa) [reply src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx]: (this feature is staying "experimental" for now in case we want to change this default)

File: src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+title: Automatic resource provisioning for KV, R2, and D1
+description: Wrangler can now automatically provision KV, R2, and D1 resources when deploying your Worker
+products:
+  - workers
+date: 2025-10-24
+---
+
+import { Render, TypeScriptExample, WranglerConfig } from "~/components";
+
+Previously, if you wanted to develop or deploy a worker with attached resources, you'd have to first manually create the desired resources. Now, if your Wrangler configuration file includes a KV namespace, D1 database, or R2 bucket that does not yet exist on your account, you can develop locally and deploy your application seamlessly, without having to run additional commands to create resources.
```
lrapoport-cf: nit suggestion to try to reduce the number of repetitions of "resource":

```suggestion
Previously, if you wanted to develop or deploy a worker with attached resources, you'd have to first manually create the desired resources. Now, if your Wrangler configuration file includes a KV namespace, D1 database, or R2 bucket that does not yet exist on your account, you can develop locally and deploy your application seamlessly, without having to run additional commands.
```

or 

```suggestion
Previously, if you wanted to develop or deploy a worker with attached resources, you'd have to first manually create them. Now, if your Wrangler configuration file includes a KV namespace, D1 database, or R2 bucket that does not yet exist on your account, you can develop locally and deploy your application seamlessly, without having to run additional resource creation commands.
```

?

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+This is still an experimental feature, but we're turning it on by default to make it easier for people to test it and try it out. You can disable the feature using the `--no-x-provision` flag. It currently works for KV, R2, and D1 bindings.
```
lrapoport-cf:
```suggestion
Automatic provisioning is still experimental, but we're turning it on by default to make it easier to try out and test. It currently works for KV, R2, and D1 bindings. You can disable the feature using the `--no-x-provision` flag. 
```

lrapoport-cf (approved):

emily-shen (approved):

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
```diff
@@ -0,0 +1,27 @@
...
+products:
+  - workers
+date: 2025-10-24
+---
+
+import { Render, TypeScriptExample, WranglerConfig } from "~/components";
+
+Previously, if you wanted to develop or deploy a worker with attached resources, you'd have to first manually create the desired resources. Now, if your Wrangler configuration file includes a KV namespace, D1 database, or R2 bucket that does not yet exist on your account, you can develop locally and deploy your application seamlessly, without having to run additional commands.
+
+Automatic provisioning is still experimental, but we're turning it on by default to make it easier to try out and test. It currently works for KV, R2, and D1 bindings. You can disable the feature using the `--no-x-provision` flag.
```
irvinebroque: This sentence reads odd - it's experimental but also default?

irvinebroque: I think I intuit the logic here

It can be default because it doesn't actually change anything if you have a configuration already

As a customer i still have questions though

- if it's default behavior how can it also be experimental? Sounds like it is just default?
- if I mess up and omit a KV namespace ID, and then deploy, can I inadvertently have a bad deployment?

I'm guessing we have answers here but feels like there's something left unsaid maybe?

Oxyjun (approved):
Approving from PCX to ensure there's no holdup.

src/content/changelog/workers/2025-10-24-automatic-resource-provisioning.mdx
penalosa: > if it's default behavior how can it also be experimental?

The idea is that it's on by default (so anyone can use it without having to add experimental flags), but that it's still experimental to give us a bit of an out to be able to make breaking changes. In particular, I'm trying to leave space open for changing whether or not this is interactive by default. That all said, it's a stable feature, so maybe that's being a bit overcautious.

> if I mess up and omit a KV namespace ID, and then deploy, can I inadvertently have a bad deployment?

No, that's what this feature supports. As part of the deploy process a real namespace would be provisioned, and so you'd end up with a deployment that worked exactly as it did during local dev.

petebacondarwin: Perhaps we should change the wording to say that this feature is in open beta (which is really what we are doing here), and point out that this means we may make breaking changes to it?

penalosa: Changed the wording a bit—how does that look now?

MattieTK: Looks good to me, very much in favour of the beta wording.

irvinebroque (approved):

PR #26127, opened 2025-10-28
remove outdated docs for workers free-tier burst and workers anti-abuse rate limits

xortive:
These limits are no longer active and don't manifest in the same way, continuing to document them will confuse customers.


rita3ko (approved):
🎉

PR #26181, opened 2025-10-30
[Docs builds] Move node options to package.json for viz + standardize

kodster28:
Follow up of other work, we should really specify these in package.json and then remove the node options from the individual actions.

This doesn't _quite_ remove from the individual actions that matter, but will do once we confirm that this behavior works as expected in production (namely, by running [Anchor Link audit](https://github.com/cloudflare/cloudflare-docs/actions/runs/18946052941) against this branch -- TBD).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
 	"type": "module",
 	"scripts": {
 		"astro": "npx astro",
-		"build": "npx astro build",
+		"build": "NODE_OPTIONS='--max-old-space-size=6192' npx astro build",
 		"typegen:worker": "npx wrangler types ./worker/worker-configuration.d.ts",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
 	"type": "module",
 	"scripts": {
 		"astro": "npx astro",
-		"build": "npx astro build",
+		"build": "NODE_OPTIONS='--max-old-space-size=6192' npx astro build",
```
kodster28:
```suggestion
		"build": "export NODE_OPTIONS='--max-old-space-size=6192' || set NODE_OPTIONS='--max-old-space-size=6192'&& npx astro build",
```

pedrosousa (approved):

PR #26253, opened 2025-11-03
PCX-19909 [Cloudflare Tunnel] Update links to GPG key

nikitacano:
### Summary

We have rotated our public GPG key used for package signing. The "Install and run a connector" instructions are now outdated and will cause installation failures for users on Debian and RedHat. We must update the curl command and the signed-by path to use the new key URL and filename.


ranbel:
We'll also need to update the installation commands in the dash.

ranbel (approved):

ranbel:
relates to #26299

PR #26265, opened 2025-11-04
Add details on how to disable remote binding in `wrangler dev`, `getPlatformProxy` and the vite plugin

dario-piotrowicz:


irvinebroque (commented):
what's the equivalent for the vite plugin?

dario-piotrowicz:
> what's the equivalent for the vite plugin?

It's `remoteBindings: false`:
https://github.com/cloudflare/cloudflare-docs/pull/26265/files#diff-dbb9011c1e9bc85e74d3d256bb136428570def0d8332bd6046b3856c880fd4f0R76-R78

Let me update the PR's title 👍

src/content/docs/workers/wrangler/commands.mdx
```diff
@@ -246,6 +246,8 @@ None of the options for this command are required. Many of these options can be
   - Show the interactive dev session.
 - `--alias` `Array<string>`
   - Specify modules to alias using [module aliasing](/workers/wrangler/configuration/#module-aliasing).
+- `--local` <Type text="boolean" /> <MetaInfo text="(default: false) optional" />
```
petebacondarwin, line 132: We also have `--remote` arg above. How do these two interact?

src/content/docs/workers/wrangler/commands.mdx
```diff
@@ -246,6 +246,8 @@ None of the options for this command are required. Many of these options can be
...
+  - Develop against all local resources, this includes both the worker's logic and [remote bindings](/workers/development-testing/#remote-bindings) which will be disabled (behaving exactly as if there were configured with `remote: false`).
```
petebacondarwin: It is not 100% clear what "develop against local resources" means, despite the rest of the sentence.

Perhaps something like:


```suggestion
  - Run in local mode. In this mode:
    - the Worker code is running locally on your machine
    - all [remote bindings](/workers/development-testing/#remote-bindings) are disabled, which behaves exactly as if there were configured with `remote: false`.
```

dario-piotrowicz: sounds good 🙂

src/content/docs/workers/wrangler/commands.mdx
dario-piotrowicz: when `--remote` is used we run the worker remotely so `--local` doesn't have any effect there, shall we add something here to the extent of: ``This flag ha no effect when `--remote` is specified``?

src/content/docs/workers/vite-plugin/reference/api.mdx
```diff
@@ -73,6 +73,11 @@ It accepts an optional `PluginConfig` parameter.
   You can inspect the `dist` directory and then run `wrangler deploy -c dist/<auxiliary-worker>/wrangler.json` for each.
   :::
 
+- `remoteBindings` <Type text='boolean' /> <MetaInfo text='optional' />
+
+  Whether or not [remote bindings](/workers/development-testing/#remote-bindings) should be enabled, they are by default.
```
ascorbic:
```suggestion
  Whether or not [remote bindings](/workers/development-testing/#remote-bindings) should be enabled. Defaults to `true`.
```

src/content/docs/workers/wrangler/commands.mdx
petebacondarwin: What about `--remote=false`?

dario-piotrowicz: What about something like ``This flag ha no effect when running in remote mode (e.g. `--remote=true`)`` ?

dario-piotrowicz: Or maybe ``This flag ha no effect when `--remote` or `--remote=true` is provided)`` ?

petebacondarwin (approved):

PR #26271, opened 2025-11-04
 Set the sitemap modified on date to the last mod date in git

WalshyDev:
### Summary
As discussed in Chat, the modified on date for the sitemap is always set to the build time. This means that every entry is always seen as "updated" even though it hasn't been.
This is causing some issues in AI Search and is generally just not correct behaviour.

This updates it to using the Git modified on date and falling back to today if not possible

### Documentation checklist

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -11,6 +11,8 @@ import react from "@astrojs/react";
 
 import { readdir } from "fs/promises";
 import { fileURLToPath } from "url";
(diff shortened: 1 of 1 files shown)


WalshyDev:
This seems to work well, the only ones falling back to current time is the changelogs which makes sense. We can probably support those too but I think this is in a much better place now and happy with it as is :)

kodster28 (approved):

PR #26274, opened 2025-11-04
Cefan/linux warp release format difference

cdrubin:
### Summary

The release notes for Linux clients were rendering differently to the release notes of the Mac and Windows clients.

### Resolution

YAML multiline strings can be represented as [block scalars](https://yaml-multiline.info/) with different "block chomping" behaviour. The Mac and Windows release files were using: `|-` whereas the latest Linux release file was using `>-` which is so-called "folded style". This replaces newlines with spaces in the source Markdown which resulted in Markdown -> HTML not seeing the newlines needed to signal headers.

### DIFF
--- a/src/components/changelog/Header.astro
+++ b/src/components/changelog/Header.astro
@@ -32,10 +32,8 @@ const products = await uniqueProducts(notes);
 ---
 
 <div
-	id="changelogHeader"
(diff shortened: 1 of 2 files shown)


cdrubin:
Review again before requesting merge...

PR #26283, opened 2025-11-04
[Gateway] CONNECT proxy/PAC file identity

Maddy-Cloudflare:


src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/configure-pac-files.mdx
```diff
@@ -61,6 +61,19 @@ Your Cloudflare proxy server domain is of the form:
 https://<SUBDOMAIN>.proxy.cloudflare-gateway.com
 ```
 
+Once you create an endpoint, the dashboard will display an option to add additional endpoints:
```
dfialho: This sentence seems to imply an endpoint needs to exist for the authorization endpoint option to be available. That is not the case. The user can now create two types of proxy endpoints: IP and Authorization. Consider adding this as part of the normal flow of creating a proxy endpoint instead of something you can do later after creating the first one. The current steps there might be confusing because they don't mention this step of picking the type of endpoint.

I'd include a note saying that once you pick a type you cannot go back.
Consider also including something to help the user decide between one option and the other

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/configure-pac-files.mdx
```diff
@@ -114,6 +127,19 @@ https://<SUBDOMAIN>.proxy.cloudflare-gateway.com
 
 </Tabs>
 
+### Edit an existing proxy endpoint
+
+To edit an existing proxy endpoint:
+
+1. In [Cloudflare One](https://one.dash.cloudflare.com/), go to **Networks** > **Resolvers & Proxies** > **Proxy endpoints**.
+2. Select **Proxy endpoints**, then select the three dots.
+3. Select **Configure**, and choose the information you want to edit:
```
dfialho: The information here only applies to authorization endpoints. IP endpoints allow you to edit the name but also configure the source IPs that are allowed

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/limitations.mdx
```diff
@@ -0,0 +1,16 @@
...
+title: Limitations
+sidebar:
+  order: 4
+---
+
+## Traffic limitations
+
+The agentless HTTP proxy does not support [identity-based policies](/cloudflare-one/traffic-policies/identity-selectors/) or mTLS authentication.
+
+To enforce HTTP policies for UDP traffic, you must turn on the [Gateway proxy for UDP](/cloudflare-one/traffic-policies/http-policies/http3/#enable-http3-inspection).
```
dfialho: This is still true. Additionally, there's a new limitation for authorization endpoints, they don't support anything that is not HTTP/HTTPS. That means no other TCP or UDP protocol is supported, including HTTP3.

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/limitations.mdx
```diff
@@ -0,0 +1,16 @@
+---
+pcx_content_type: how-to
+title: Limitations
+sidebar:
+  order: 4
+---
+
+## Traffic limitations
+
+The agentless HTTP proxy does not support [identity-based policies](/cloudflare-one/traffic-policies/identity-selectors/) or mTLS authentication.
```
dfialho: This is now only a limitation for IP endpoints. The whole point of authorization endpoints is so that we have identity available for policies.

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/configure-pac-files.mdx
```diff
@@ -164,6 +190,29 @@ function FindProxyForURL(url, host) {
 - Use a proper text editor such as VS Code to avoid added characters.
   :::
 
+### Add additional PAC files
```
dfialho: Why the term additional here? The expected flow is that a user would create an endpoint and then go and create a PAC file. Before customers would have to host their own PAC files, with this feature we're providing the hosting for them.

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/configure-pac-files.mdx
```diff
@@ -164,6 +190,29 @@ function FindProxyForURL(url, host) {
...
+
+1. Go to **Networks** > **Resolves & Proxies** > **Proxy endpoints**.
+2. Select **Add PAC files**.
+3. Here, you can add **PAC file details** and **Setup instructions**. 
+    In **PAC files details:**
+        - Enter the **Basic Information**.
+        - Enter the **PAC file configuration** > Select **Browse PAC file configuration templates** and choose a pre-configured template to customize. The only available outputs are Okta and Azure. Once you select the template, the **PAC file JavaScript** is going to be populated with a template.
+    In **Setup instructions:**
+        - Choose a browser and follow the instructions.
+4. Select **Create**.
```
dfialho: There's an important field that set up, the "slug". That will define how the URL for the pac file will look like. Do you think we should call this out here? This field is also optional - a random slug is generated if not provided. Additionally, this field is not editable because that would break any devices already setup to use the PAC file.

Once the PAC file is created the user can see a URL for the file. That is the URL they should use to setup the devices. Should we mention this here, similar to what we do for proxy endpoints on line 58?

Maddy-Cloudflare: @dfialho we can call it out.

Would you be able to give me the generic output URL?

dfialho: Yes
retail: `https://pac.cloudflare-gateway.com/<account-tag>/<slug>`
fed: `https://pac.fed.cloudflare-gateway.com/<account-tag>/<slug>`

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/limitations.mdx
jgpaiva: I think originally this was wrong. Turning on gateway proxy for UDP does nothing for PAC files because they only support TCP traffic.

I think we should remove this paragraph, or have something simple at the end like
> Both HTTP proxy endpoints only support proxying and filtering TCP traffic.

src/content/docs/cloudflare-one/team-and-resources/devices/agentless/pac-files/configure-pac-files.mdx
```diff
@@ -43,24 +45,74 @@ All devices you add to the proxy endpoint will be able to access your Cloudflare
...
-3. Give your endpoint any name.
+3. Select between **Add an authorization endpoint** or **Add a source IP endpoint**.
+
+:::note
+Once you choose a type of proxy endpoint, you cannot revert this decision.
+:::
+
+## Authorization endpoint
+
+Authorization endpoints support being associated with [Cloudflare Access](/cloudflare-one/access-controls/policies/) to provide full authorization capabilities. Authorization endpoint allows you yo configure authorization methods (for example, with Google), before traffic is sent for processing.
```
jgpaiva: I'm not sure we should mention Google explicitly here. Maybe we can say something like "allow using the full power of access policies to authorize traffic" and link there? https://developers.cloudflare.com/reference-architecture/design-guides/designing-ztna-access-policies/

We should probably also call out the other main advantage of Authorization endpoints (vs source IP endpoints): they support associating a user identity with the connection, these *do* support identity-based policies (the limitation for source IP endpoints is called out on "traffic limitations").

maxvp (approved):

PR #26291, opened 2025-11-04
[DNS] Nameserver assignment consistency

RebeccaTamachiro:
### Summary

PCX-13571


src/content/docs/dns/nameservers/nameserver-options.mdx
```diff
@@ -14,10 +14,18 @@ Refer to the sections below to learn about different nameserver options.
...
+The [default assignment method](/dns/zone-setups/reference/nameserver-assignment/) is to use standard nameservers and favor consistent nameserver names across all zones within an account. Nonetheless, in case there are conflicts — for example, if someone else has already added the same zone to a different account — you may get different nameserver names.
 
 To have control over what nameservers are assigned for different zones within an account, you can use [account custom nameservers](/dns/nameservers/custom-nameservers/account-custom-nameservers/).
 
+### Nameserver consistency
+
+The level of consistency you can expect when adding new zones depends on the nameserver type configured.
+
+For [standard nameservers](/dns/nameservers/#standard-nameservers), since a conflict can be caused by anyone adding the same zone to any other Cloudflare account, the likelihood of your new zone being assigned different nameserver names than your previously existing ones is way higher.
```
hannes-cf: We could add a note that this can be prevented by placing a zone hold on your zone, if it's active already: https://developers.cloudflare.com/fundamentals/account/account-security/zone-holds/

src/content/docs/dns/nameservers/nameserver-options.mdx
```diff
@@ -14,10 +14,18 @@ Refer to the sections below to learn about different nameserver options.
...
+
+For [account custom nameservers](/dns/nameservers/custom-nameservers/account-custom-nameservers/), [tenant custom nameservers](/dns/nameservers/custom-nameservers/tenant-custom-nameservers/), and [Foundation DNS advanced nameservers](/dns/foundation-dns/advanced-nameservers/#nameservers-hosting-and-assignment), the only conflict would be between a parent and a child zone, which makes consistent assignment across new zones more likely.
```
hannes-cf: With Foundation DNS advanced nameservers, I could have the same conflict if someone else signs up the same zone name in their account, BUT since advanced nameservers is an ENT-only feature the likelihood is significantly smaller.

For account custom nameserver, your statement is correct, since I cannot have two zones with the same name twice in the same account, meaning the only conflict indeed could be between parent and child (or deeper relation).

For tenant custom nameserver, however, there could be a conflict between two zones with the same name in two different accounts, if both accounts belong to the same tenants. But again, the likelihood is very small that this would happen.

src/content/docs/dns/nameservers/nameserver-options.mdx
```diff
@@ -12,11 +12,25 @@ Refer to the sections below to learn about different nameserver options.
...
-To have control over what nameservers are assigned for different zones within an account, you can use [account custom nameservers](/dns/nameservers/custom-nameservers/account-custom-nameservers/).
+:::caution
+To prevent domain hijacking, you can no longer preset Cloudflare nameservers at your registrar before creating the respective zone in Cloudflare. If you preset your nameservers and then add the domain, your domain will be assigned a new pair of nameservers.
+:::
+
+These nameserver assignments cannot be changed. However, depending on your subscription, you may have different options for better nameserver consistency.
+
+### Nameserver consistency
+
+The level of consistency you can expect when adding new zones depends on the nameserver type configured.
```
pedrosousa:
```suggestion
The level of consistency you can expect when adding new zones depends on the configured nameserver type.
```
Nit: I believe this is more readable, and hopefully doesn't change the intended meaning.

pedrosousa (approved):

PR #26372, opened 2025-11-07
adding-instance-restarts-changelog-entry

mia303:
### Summary

Adding changelog for instance.restart() support in Cloudflare Workflows


src/content/changelog/workflows/2025-11-10-instance-restarts.mdx
```diff
@@ -0,0 +1,21 @@
+---
+title: Instance restarts now supported in Workflows
+description: You can now restart an instance of your workflow without losing its concurrency spot
```
rita3ko: capitalize Workflow?

src/content/changelog/workflows/2025-11-10-instance-restarts.mdx
```diff
@@ -0,0 +1,21 @@
...
+---
+
+You can now restart a [Cloudflare Workflow](/workflows/) instance via the [API](/api/resources/workflows/methods/list/), [Wrangler](/workers/wrangler/commands/#workflows), or [Workers](/workflows/build/workers-api/):
+
+```typescript
+let instance = await env.MY_WORKFLOW.get("abc-123");
+await instance.restart(); // Returns Promise<void>
+```
+
+You may want to [restart an instance](/workflows/build/trigger-workflows/#restart-a-workflow) if the Workflow has been updated, the parameters have changed, or the Workflow has reached its step limit. Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
```
rita3ko: Would be good to add a sentence here of why this is important

src/content/changelog/workflows/2025-11-10-instance-restarts.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+If the instance is already running, the restart will not lose its concurrency slot (limit of 10,000 concurrent instances). However, the restarted instance will count towards the creation rate limit (100 instances per second).
```
rita3ko: this is a bit confusing to parse. i would clarify concurrency vs. creation here a bit more. i.e. does this help if i have 10,001 instances? or is this useful in general?
what does creation mean here?

rita3ko (changes requested):

Maddy-Cloudflare (approved):

src/content/changelog/workflows/2025-11-12-instance-restarts.mdx
```diff
@@ -0,0 +1,23 @@
...
+```typescript
+let instance = await env.MY_WORKFLOW.get("abc-123");
+await instance.restart(); // Returns Promise<void>
+```
+
+You may want to [restart an instance](/workflows/build/trigger-workflows/#restart-a-workflow) if the Workflow has been updated, the parameters have changed, or the Workflow has reached its step limit. This allows you to make changes without waiting until the instance is complete or preserve a long-running instance through restart. Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
+
+If the instance is already running, the restart will not lose its concurrency slot (limit of 10,000 concurrent instances). For example, say you have 10,000 instances running and 2,000 instances queued. If you restart a runnning instance, it will maintain its spot in the 10,000 rather than slotting into the back of the queue.
+
+However, note that the restarted instance will count towards the instance creation rate limit (100 Workflow instances per second).
```
rita3ko: I would drop this line. I think it adds more questions than it answers (assuming this is documented)

src/content/changelog/workflows/2025-11-12-instance-restarts.mdx
```diff
@@ -0,0 +1,23 @@
...
+You can now restart a [Cloudflare Workflow](/workflows/) instance via the [API](/api/resources/workflows/methods/list/), [Wrangler](/workers/wrangler/commands/#workflows), or [Workers](/workflows/build/workers-api/):
+
+```typescript
+let instance = await env.MY_WORKFLOW.get("abc-123");
+await instance.restart(); // Returns Promise<void>
+```
+
+You may want to [restart an instance](/workflows/build/trigger-workflows/#restart-a-workflow) if the Workflow has been updated, the parameters have changed, or the Workflow has reached its step limit. This allows you to make changes without waiting until the instance is complete or preserve a long-running instance through restart. Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
+
+If the instance is already running, the restart will not lose its concurrency slot (limit of 10,000 concurrent instances). For example, say you have 10,000 instances running and 2,000 instances queued. If you restart a runnning instance, it will maintain its spot in the 10,000 rather than slotting into the back of the queue.
```
rita3ko: I like the example. You can be more concrete — and say if the instance was in the 300th slot, it would go back in there as opposed to the 2001st.

PR #26412, opened 2025-11-10
[Starlight] Auto-set theme from param

kodster28:
### Summary

Diff implementation of #26409

### DIFF
--- a/src/components/overrides/ThemeProvider.astro
+++ b/src/components/overrides/ThemeProvider.astro
@@ -1,0 +1,14 @@
+---
+import Default from "@astrojs/starlight/components/ThemeProvider.astro";
+---
+
+<script is:inline>
+	const params = new URL(location.href).searchParams;
+	const theme = params.get("preferred-color-scheme");
+	if (theme === "dark" || theme === "light") {
+		document.documentElement.dataset.theme = theme;
+		localStorage.setItem("starlight-theme", theme);
+	}
+</script>
+
+<Default><slot /></Default>
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -190,6 +190,7 @@ export default defineConfig({
 				Sidebar: "./src/components/overrides/Sidebar.astro",
 				PageTitle: "./src/components/overrides/PageTitle.astro",
 				TableOfContents: "./src/components/overrides/TableOfContents.astro",
+				ThemeProvider: "./src/components/overrides/ThemeProvider.astro",
 			},
 			sidebar,
 			customCss,


src/components/overrides/ThemeProvider.astro
```diff
@@ -0,0 +1,14 @@
+---
+import Default from "@astrojs/starlight/components/ThemeProvider.astro";
+---
+
+<script is:inline>
+	const params = new URL(location.href).searchParams;
+	const theme = params.get("preferred-color-scheme");
+	if (theme === "dark" || theme === "light") {
+		document.documentElement.dataset.theme = theme;
```
kodster28, line 9: This line doesn't appear to do anything based on testing. Keeping it for the moment, but probably superceded by the local storage setting line in 10.

Ideally, I'd prefer to keep this behavior (9 over 10) -- and just load this time in this theme -- instead of setting the local variable to avoid someone getting docs theming set unexpectedly

src/components/overrides/ThemeProvider.astro
```diff
@@ -0,0 +1,14 @@
...
+		localStorage.setItem("starlight-theme", theme);
+	}
+</script>
+
+<Default><slot /></Default>
```
colbywhite, line 14: How does this play with [Starlight's version](https://github.com/withastro/starlight/blob/0b98ddbf43c07b6b4aaf25502bedc586d31f1ce4/packages/starlight/components/ThemeProvider.astro)? Which one gets loaded first?

You might want to not use the `Default` and combine the existing logic to give you something along the lines of

```js
const urlTheme = new URL(location.href).searchParams.get("preferred-color-scheme");
const storedTheme =
			typeof localStorage !== 'undefined' && localStorage.getItem('starlight-theme');
const theme =
			urlTheme || storedTheme ||
			(window.matchMedia('(prefers-color-scheme: light)').matches ? 'light' : 'dark');
// do the same logic as Starlight's from here on out
```

I'm guessing the standard way to override Starlight comps won't allow that? And thus we need to _replace_ it instead of _adding_ to it?

colbywhite: Also, I think the following use case could happen:
- The URL is `theme=light`, which loads the light theme.
- User clicks the doc's theme picker and chooses the dark theme.
- Now the URL has light but the doc is dark. 

Not sure we care about that though. It in theory could break copy/pasting of links depending on how Astro handle relative links when there's a query param. (I assume it ignores it.) If this case matters, then something in [ThemeSelect](https://starlight.astro.build/reference/overrides/#themeselect) likely needs to be overridden as well.

m0tts:
Updating the query param works great on initial load, when the docs are loaded in an iframe. However, if you start navigating the docs in the iframe, and then switch themes, the iframe's src is reloaded and you lose your navigation.

Clipboard-20251110-231548-186


With the url concerns here as well https://github.com/cloudflare/cloudflare-docs/pull/26412#discussion_r2512028955, is there any appetite for reading from either reading from a shared cookie and/or include a [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) listener.

I don't love the postMessage, only because it would be functionality that's only used when the docs are loaded in an iframe.

I wouldn't mind trying to clean up the [original PR ](https://github.com/cloudflare/cloudflare-docs/pull/26409/files)a bit, so the changes are a bit tidier, and not as complex. I will admit i hit "enter" a bit prematurely on that PR, as i was trying to open it in Draft mode, originally

src/components/overrides/ThemeProvider.astro
kodster28: > How does this play with [Starlight's version](https://github.com/withastro/starlight/blob/0b98ddbf43c07b6b4aaf25502bedc586d31f1ce4/packages/starlight/components/ThemeProvider.astro)? Which one gets loaded first?
> 
> You might want to not use the `Default` and combine the existing logic to give you something along the lines of
> 
> ```js
> const urlTheme = new URL(location.href).searchParams.get("preferred-color-scheme");
> const storedTheme =
> 			typeof localStorage !== 'undefined' && localStorage.getItem('starlight-theme');
> const theme =
> 			urlTheme || storedTheme ||
> 			(window.matchMedia('(prefers-color-scheme: light)').matches ? 'light' : 'dark');
> // do the same logic as Starlight's from here on out
> ```
> 
> I'm guessing the standard way to override Starlight comps won't allow that? And thus we need to _replace_ it instead of _adding_ to it?

I _think_ so, as far as I understand Starlight's overrides.

You can append logic to the component via the Default pattern, but once you start touching the internals you have to rewrite a lot.

kodster28:
> Updating the query param works great on initial load, when the docs are loaded in an iframe. However, if you start navigating the docs in the iframe, and then switch themes, the iframe's src is reloaded and you lose your navigation.
> 
> Clipboard-20251110-231548-186 [ Clipboard-20251110-231548-186 ](https://private-user-images.githubusercontent.com/2081238/512431346-9913c0d5-718b-45a7-ab07-3668a741842d.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjI4Nzk2OTEsIm5iZiI6MTc2Mjg3OTM5MSwicGF0aCI6Ii8yMDgxMjM4LzUxMjQzMTM0Ni05OTEzYzBkNS03MThiLTQ1YTctYWIwNy0zNjY4YTc0MTg0MmQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTExMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTExMTFUMTY0MzExWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZTAyYzlhOGJiOGM0YTgxZTNlMjJkMmNhNmRjZWY5MDMzZjU5OTQ4NTUyMjE1M2VkNzM1ZGJkOTFjN2ViNTVkYiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.UVh7_5IlhDkl3kJS1JkU9amtMJEkPeNPNlL9PlGSKKA) [ ](https://private-user-images.githubusercontent.com/2081238/512431346-9913c0d5-718b-45a7-ab07-3668a741842d.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjI4Nzk2OTEsIm5iZiI6MTc2Mjg3OTM5MSwicGF0aCI6Ii8yMDgxMjM4LzUxMjQzMTM0Ni05OTEzYzBkNS03MThiLTQ1YTctYWIwNy0zNjY4YTc0MTg0MmQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTExMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTExMTFUMTY0MzExWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZTAyYzlhOGJiOGM0YTgxZTNlMjJkMmNhNmRjZWY5MDMzZjU5OTQ4NTUyMjE1M2VkNzM1ZGJkOTFjN2ViNTVkYiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.UVh7_5IlhDkl3kJS1JkU9amtMJEkPeNPNlL9PlGSKKA)
> 
> With the url concerns here as well [#26412 (comment)](https://github.com/cloudflare/cloudflare-docs/pull/26412#discussion_r2512028955), is there any appetite for reading from either reading from a shared cookie and/or include a [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) listener.
> 
> I don't love the postMessage, only because it would be functionality that's only used when the docs are loaded in an iframe.
> 
> I wouldn't mind trying to clean up the [original PR ](https://github.com/cloudflare/cloudflare-docs/pull/26409/files)a bit, so the changes are a bit tidier, and not as complex. I will admit i hit "enter" a bit prematurely on that PR, as i was trying to open it in Draft mode, originally

I agree that that could be a problem... but do we have any data about how prevalent it is? Like, imagine you could look at that via analytics.

From my initial PoV, I'd guess that's very much an edge case and not something we need to focus on right now. Also, I don't love introducing that level of complexity unless we really have to.

m0tts:
We have alignment that we don't need to worry about this edge case, so feel free to merge in these query param changes whenever!

m0tts:
We were able to release the changes on the dashboard today. Is there an eta on when the changes here to the docs site to support `?preferred-color-scheme` may be released?

kodster28:
> We were able to release the changes on the dashboard today. Is there an eta on when the changes here to the docs site to support `?preferred-color-scheme` may be released?

Should be this week. Will block off some time to focus on it tomorrow / Wednesday.

You can also follow CED-176 internally for updates.

PR #26418, opened 2025-11-10
[Publish GHA] Update copying logic to populate recent llms-full.txt files

### LINKED ISSUE #25917: LLM Text doesnt generate
### Example URL(s)

https://developers.cloudflare.com/sandbox/llms-full.txt

### Actual Behavior

`https://developers.cloudflare.com/sandbox/llms-full.txt` is showing up empty.

### Expected Behavior

Content Showing up

### Additional information

_No response_

kodster28:
**Context**
Ever since #23688, we actually haven't had updated `/$product/llms-full.txt` files getting uploaded into R2. But they're present locally... just not correctly getting uploaded via Rclone. 

This has meant that -- as we add new products (or change the name of existing ones) -- we see behavior like #25917.

This also means that the current `llms-full.txt` files aren't up to date (most have a last modified date of July 15th).

**Fix**

Based on some AI suggestions, tweaked the script.

I tested this behavior locally by creating a new R2 bucket in a testing account (`PCX Team > test-dev-docs-vendored-markdown`) and then manually tried uploading via rclone.

Seemed to have the correct result when I uploaded and then downloaded a sample.
[ai-search_llms-full.txt](https://github.com/user-attachments/files/23463300/ai-search_llms-full.txt)

image

closes #25917

### DIFF
--- a/.github/workflows/publish-production.yml
+++ b/.github/workflows/publish-production.yml
@@ -51,12 +51,10 @@ jobs:
           cd ..
 
           cd distllms
-          for file in $(find . -type f); do
-            rclone copy \
-              --config ../bin/rclone.conf \
-              $file \
-              devdocs:vendored-markdown
-          done
+          rclone copy \
+            --config ../bin/rclone.conf \
+            . \
+            devdocs:vendored-markdown
           cd ..
       - name: Upload vendored Markdown files to ZT DevDocs bucket
         env:
--- a/.github/workflows/publish-production.yml
+++ b/.github/workflows/publish-production.yml
@@ -51,7 +51,7 @@ jobs:
           cd ..
 
           cd distllms
-          rclone copy \
+          rclone copy --max-depth 2 \
             --config ../bin/rclone.conf \
             . \
             devdocs:vendored-markdown


.github/workflows/publish-production.yml
```diff
@@ -51,12 +51,10 @@ jobs:
           cd ..
 
           cd distllms
-          for file in $(find . -type f); do
-            rclone copy \
-              --config ../bin/rclone.conf \
-              $file \
-              devdocs:vendored-markdown
-          done
+          rclone copy \
```
cdrubin: This is also a more sensible use of `copy` - for single files as in the previous commit `copyto` is recommended: https://rclone.org/commands/rclone_copy/

.github/workflows/publish-production.yml
```diff
@@ -51,12 +51,10 @@ jobs:
...
+            --config ../bin/rclone.conf \
+            . \
```
cdrubin, line 56: It would probably be sensible to use `--max-depth 0` to avoid recursion into a dir if one is found in `.`

kodster28: We want to update the files inside the directory if they're present, so in testing seemed like `--max-depth 2` was required to see the expected behavior

.github/workflows/publish-production.yml
kodster28:
```suggestion
          rclone copy --max-depth 2 \
```

cdrubin (approved):

PR #26479, opened 2025-11-13
Adds drafts docs for default environments

korinne:
## Summary

This PR introduces documentation for Workers' two default environments:
  - Production: For serving live traffic
  - Preview: For testing PR changes before they go live

This is all exploratory and still in development. 

#### Configuration

  - New envVars structure in Wrangler config with explicit targetEnv
  arrays
  - Environment variables can have types: text, secret, or binding (KV,
   D1, R2, etc.)
  - No inheritance - everything is explicitly configured per environment

####  Local Development

  - Pulls from Preview environment by default via wrangler env pull
  - Uses Miniflare to simulate resources locally
  - @ directive syntax for connecting to remote resources when needed

#### Guides

  Includes step-by-step guides for:
  - CLI (Wrangler)
  - Dashboard
  - API
  - TypeScript SDK
  - Terraform


src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+	InlineBadge,
+	CardGrid,
+	Card,
+	Type,
+	TypeScriptExample,
+	Tabs,
+	TabItem,
+} from "~/components";
+
+An **environment** is a collection of settings that determines how your Worker runs - such as which resources it connects to, what [environment variables](/workers/configuration/environment-variables/) it uses, and what domains it's available on.
```
irvinebroque:
```suggestion
An **environment** is a collection of settings that determines how your Worker runs. These settings include which resources it can connect to via [bindings](/workers/runtime-apis/bindings/), what [environment variables](/workers/configuration/environment-variables/) and [secrets](/workers/configuration/secrets/) it uses, and what domains it's available on.
```

Calling out secrets and bindings directly helpful I think? Since you use the languge below.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+Every Worker has two default **environments** that map to specific phases in your workflow:
```
irvinebroque:
```suggestion
Every Worker has two environments:
```

If you caveat with default, kind of implies that there can be more than production and preview, rather than just saying it directly? Does it need the "default" here?

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+You can also easily set up your **[local development environment](#local-development)** by pulling down your **Preview** settings with `wrangler env pull` to develop and test changes on your local machine.
+
+## Configuration
+
+When you write a Worker, you need to tell Cloudflare two things:
+
+1. **What your Worker is** - Its name, which code file to run, and how it should behave
+2. **What your Worker connects to** - Which databases, API keys, and resources to use
+
+The second part is different for **Production** (your live site) and **Preview** (testing changes). You configure these separately so your tests don't accidentally mess with real user data.
```
irvinebroque, line 42:
```suggestion
The second part is different for **Production** (your live site) and **Preview** (testing changes). You configure these separately so that when you test out preview deployments, you don't accidentally mess with real user data.
```

I can see a case where it's actual programmatic integration tests that run against the preview deployment, but is the common base case just trying out a change on a preview url?

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+	TabItem,
+} from "~/components";
+
+An **environment** is a collection of settings that determines how your Worker runs - such as which resources it connects to, what [environment variables](/workers/configuration/environment-variables/) it uses, and what domains it's available on.
+
+Every Worker has two default **environments** that map to specific phases in your workflow:
+
+- **[Production](#production)**: For serving your live application to real users on your [`*.workers.dev` subdomain](/workers/configuration/routing/workers-dev/), [custom domain](/workers/configuration/routing/custom-domains/), or [route](/workers/configuration/routing/routes/).
+
+- **[Preview](#preview)**: For testing and sharing Pull Request changes before they go live. Each Pull Request deploys its own unique **PR Preview** using the **Preview** environment settings.
```
irvinebroque:
```suggestion
- **[Preview](#preview)**: For testing and sharing changes before they go live to your users. Each pull request deploys its own unique [**PR Preview**](workers/configuration/environments/#preview) using the **Preview** environment settings.
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+```
+</WranglerConfig>
+
+Notice how `production` and `preview` have different values for `database_id`? Your Worker runs the same code in both environments, but connects to different databases.
+
+The settings outside of `production` and `preview` - like `name` and `main` - describe your Worker itself. These stay the same everywhere because you want to test the exact same Worker that your users will see. For the complete list of these settings, refer to [Global settings](#global-settings).
+
+<Details header="View complete configuration example">
+
+Here's a comprehensive example showing all available configuration options:
```
irvinebroque, line 81: I like this pattern for having a very complete example, but default collapsed, especially for a coding agent referring back to docs for context

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+  // ============================================================================
+  // PRODUCTION ENVIRONMENT
+  // ============================================================================
+  "production": {
+    // Where your production Worker is accessible
+    "domains": ["app.example.com"],
+
+    // Environment variables
+    "env": {
+      // Plain text variable - separated by "value"
```
irvinebroque: what does "separated" mean here? I see it's the language used across all of these, but don't quite follow. Like I get how this works, but I don't quite follow the meaning you are trying to convey by the word "separated"

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+- The `abc123preview` [KV namespace](/kv/) (not the **Production** namespace)
+- The `staging-db-id-456` test database (not the **Production** database)
+- The environment variable `ENVIRONMENT` set to `"preview"`
+
+### Domains
+
+Preview domains work differently than Production. Instead of picking specific domains, Cloudflare automatically generates a unique URL for each pull request so you can test multiple changes at the same time without them interfering with each other.
+
+By default, each PR Preview gets a URL like `pr-123-my-worker.my-account.workers.dev` where `123` is the pull request number. If you want to customize this pattern, you can set a `domain_template`:
```
irvinebroque: I get the idea here, looking back at [preview aliases](https://developers.cloudflare.com/workers/configuration/previews/#aliased-preview-urls). But think we are combining a few distinct things.

If I think about mechanically what is happening:

- a build runs
- it runs `wrangler deploy --preview`
- as part of running ^ it needs to provide the alias / branchname, the equivalent of: `wrangler deploy --preview --alias pr-123-my-worker`
- having to specify that in the CLI is painful, so we should be able to set a default behavior. If I run `wrangler deploy --preview`, by default it — to your point, it should pick up the worker name from `wrangler.jsonc`, and the branchname from git, and use these

So — I guess we _could_ let someone customize the URL structure in configuration like this, but if you need to do this — there can be ways to do it with the CLI? Having a hard time coming up for the _need_ for this to be here in configuration?

Then, the second part here is, "I want to use a custom domain for my preview URLs" which if I read between the lines, we are saying this is how you do that? (I can't quite tell, its implicit) 

If custom domains for preview URLs is the thing that remains here to solve for — make it part of [that setting](https://developers.cloudflare.com/workers/configuration/previews/#from-the-wrangler-configuration-file)? Which today I don't _think_ is covered in these docs?

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+}
+
+// Route only specific URL paths to your Worker
+{
+  "name": "my-worker",
+  "production": {
+    "domains": [
+      "workers.dev",
+      "app.example.com",
+      { "name": "www.example.com", "path": "/api/*" }
```
irvinebroque: It's helpful to see the examples here spelled out, a few thoughts now that I grok what you're saying:

- it feels awkward to have some items in here as strings and others as objects
- Routes and Custom Domains have very specific behavior and meaning (ex: same zone subrequests). The closer these are like this, the more it kind of implies that a route is just a custom domain that runs only on some paths. Having them all jammed together under `routes` in [current state](https://developers.cloudflare.com/workers/wrangler/configuration/#routes) isn't all that clear either.
- It's harder to look at my configuration and find my routes when they're mixed in like this
- If I want to disable `workers.dev`, and then audit this, feel like I can trust that it's explicitly disabled — in this mode I can only tell by looking for the _absence_ of `workers.dev`
- I'm not quite sure how `workers.dev` being _enabled_ could be the default behavior here. Feels important to keep, otherwise you deploy and omit this and ask "why isn't it working?"
- For a route, I don't quite follow how you can cover the full behavior of routes only using `path`. You _need_ `pattern` here and that's the language used in the docs. Ex: `*example.com/*` is a valid route. But `path` only covers the path after the hostname, not subdomain.
- You're going to need the `zone_id`, because it's possible to have subdomains for the same domain as separate zones, and trying to get clever about inferring the zone ID based on what is provided in `pattern` ends up hitting edge cases. And as a customer I want to make sure this is very specific and deterministic. And for the most part if I'm just here to go build something new — I'm probably using custom domains, not routes anyways.

Might suggest something like:

```json
  "production": {
    "custom_domains": ["my-domain.com"],
    "routes": [
      { "pattern": "subdomain.example.com/*",  "zone_id": "<YOUR_ZONE_ID>" }
     ],
    "workers_dev": false
  }
```

That way, in the base case, it's still simple:

```json
  "production": {
    "custom_domains": ["my-domain.com"]
  }
```

...but things still work as expected, and aren't ambiguous

@dinasaur404

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+      }
+    }
+  }
+}
+```
+</WranglerConfig>
+
+</Details>
+
+### Types
```
irvinebroque:
```suggestion
### Configuring the env object
```

I think fundamentally this section answers the question "how do I configure what is available on `env`?"

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+An environment variable can have one of the following types:
```
irvinebroque:
```suggestion
The [`env` object](/workers/runtime-apis/bindings/) provides:
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+- **Plain text** (`type: "text"`) - String values
```
irvinebroque:
```suggestion
- [**Environment variables**](/workers/configuration/environment-variables/) — as text strings or JSON values
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+- **Secret** (`type: "secret"`) - Sensitive data like API keys (managed via [`wrangler secret put`](/workers/configuration/secrets/), not in the configuration file)
```
irvinebroque:
```suggestion
- [**Secrets**](/workers/configuration/secrets/) - Sensitive data like API keys (managed via [`wrangler secret put`](/workers/configuration/secrets/), not in the configuration file)
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+- **Resource bindings** (`type: "kv_namespace"`, `type: "d1"`, `type: "r2_bucket"`, etc.) - Live objects that let you interact with Cloudflare resources
```
irvinebroque:
```suggestion
- [**Bindings**](/workers/runtime-apis/bindings/) APIs that let you interact with Cloudflare resources, without the need for API tokens or keys
```

Not all bindings are to a specific "resource" and we don't use this language consistently.

In context, "live objects" works but out of context of the blog post it's not obvious what it means, since it's not an established and well understood term.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
```
irvinebroque, line 332:
```suggestion
You configure your Worker's `env` object by declaring keys on `env`, and specifying their type and value, as shown below:

```

Think you need something that points this out

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+        "type": "d1",
+        "database_id": "staging-db-id-456"
+      }
+    }
+  }
+}
+```
+</WranglerConfig>
+
+### Accessing environment variables in your Worker
```
irvinebroque:
```suggestion
```

Suggest merging

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+All environment variables you configure are accessible in your Worker code via the `env` object. Access them using `env.<VARIABLE_NAME>`:
```
irvinebroque:
```suggestion
A Worker with the above configuration can access `env` as shown below:
```

I get the intent behind calling them environment variables, but not quite convinced yet it is intuitive or well understood to frame this way

For example, `process.env` exposes [environment variables, secrets, and version metadata](https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv), but not bindings.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+    // Access D1 database binding
+    const results = await env.DB.prepare("SELECT * FROM users").all();
+
+    return new Response("Hello!");
+  }
+}
+```
+</TypeScriptExample>
+
+The `env` parameter is automatically passed to your Worker's `fetch` handler and contains all the environment variables specific to the environment (Production or Preview) your Worker is running in.
```
irvinebroque:
```suggestion
The `env` parameter, available from your Worker's `fetch` handler or from [`import { env } from "cloudflare:workers"`](/workers/runtime-apis/bindings/#importing-env-as-a-global), contains all the environment variables specific to the environment (Production or Preview) your Worker is running in.
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+### Secrets
+
+[Secrets](/workers/configuration/secrets/) contain sensitive data like API keys, passwords, and tokens. You cannot define secrets in your [Wrangler configuration file](/workers/wrangler/configuration/) - you must set them using the CLI or dashboard.
+
+Secrets are **not shared** between environments. You must set them separately for each environment ([Production](#production), [Preview](#preview)) to keep them secure.
```
irvinebroque:
```suggestion
Secrets are **not shared** between environments. You must set them separately for each environment ([Production](#production), [Preview](#preview)).
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+#### Set secrets for your Production environment
+
+**CLI**
+
+<PackageManagers type="exec" pkg="wrangler" args="secret put API_SECRET" />
```
irvinebroque:
```suggestion
<PackageManagers type="exec" pkg="wrangler" args="secret put MY_API_KEY" />
```

suggest slightly clearer so that no room for thinking that `API_SECRET` is a _type_ of secret

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+      { "name": "www.example.com", "path": "/api/*" }
+    ]
+  }
+}
+```
+</WranglerConfig>
+
+## Preview
+
+Your **Preview** environment lets you test changes in isolated **PR Previews** before they reach **Production**. For example, if your **Production** environment uses your real customer [D1 database](/d1/), you can configure **Preview** to use a test database instead - keeping your **Production** data safe while you test changes.
```
irvinebroque: I get the reasons to hesitate calling them preview deployments because of overlap with the deployments concept, and trying to be more explicit about them being previews of a pull reques, but so often just called "preview deployments", feels so well understood. And the pull request bit is covered below.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+Cloudflare creates a **PR Preview** with its own unique URL using your **Preview** settings when you:
+
+- Push a commit to a branch that's not your **Production** branch when you've connected your [Git repository](/workers/ci-cd/builds/git-integration/). Cloudflare automatically posts a comment on your PR with a link to your **PR Preview**.
+- Run [`wrangler deploy --preview`](/workers/wrangler/commands/#deploy) using the Wrangler CLI.
```
irvinebroque: See other comment about behavior of `wrangler deploy --preview` and default values for name and branch it might use. I think that part needs to have some flavor of docs in here so that it's clear how it works (knowing that the docs on this page will eventually need to end up in other/many places in the docs in addition to this page)

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+```
+</WranglerConfig>
+
+The template can use `{pr}` for the pull request number, `{branch}` for the branch name, `{commit}` for the commit hash, and `{worker}` for the Worker name. So a pull request #123 from the `new-feature` branch would become `123-new-feature.preview.example.com`.
+
+### Override settings for a specific PR Preview
+
+If a specific PR needs to use different resources, [environment variables](/workers/configuration/environment-variables/), or [secrets](/workers/configuration/secrets/) than what's configured for the **Preview** environment, you can override them for that specific **PR Preview**.
+
+For example, one PR might be debugging a complex issue and needs `DEBUG_MODE=true` with verbose logging enabled, while other PRs use normal logging levels.
```
irvinebroque:
```suggestion
For example, you can use this to set the environment variable `DEBUG_MODE=true` for **only** your preview deployment in order to enable verbose logging, without affecting preview deployments from other pull requests, which will still use normal logging levels.
```

Another good one here might be enabling tracing for 100% of requests to your preview deployment. Could include a few different ones here, think this can help people understand how to use this.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+**To set overrides for a specific PR Preview:**
```
irvinebroque: This is _one_ way to do this — but is it the primary way? And if it's the only way we show, then do people know that there are other ways to do it?

Ex: all I need to do is change the value in `wrangler.jsonc` — and then push up a branch — the preview deployment for my branch will use that value, but this won't impact preview deployments for other branches.

That seems like the primary way to use this, and a lot easier than going into the dash?

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+2. Go to **Deployments** > find your **PR Preview**
+3. Click **Settings** > **Variables & secrets** > toggle the **Override** button
+4. Add or modify the values you want to override for this specific PR
+5. Click **Save** - this redeploys the **PR Preview** with your overrides
+
+All future commits to that branch will automatically use these overrides. When the branch is deleted, the overrides are automatically cleaned up.
+
+## Local Development
+
+**Local Development** is where you develop and test changes on your local machine. Unlike **[Production](#production)** and **[Preview](#preview)** which run on Cloudflare's network, local development uses [Miniflare](https://miniflare.dev/) to simulate the Workers runtime on your computer.
```
irvinebroque:
```suggestion
**Local Development** is where you develop and test changes on your local machine. When you run your Worker in local development, using either [`wrangler dev`](/workers/wrangler/commands/#dev) or `vite dev` with the [Cloudflare Vite Plugin](/workers/vite-plugin/), your code runs in the open-source Workers Runtime, called [workerd](https://github.com/cloudflare/workerd)
```

We're _actually_ running the Workers runtime in local dev, not simulating it

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+Local development is not a separate stored environment. Instead, you pull environment variables from Preview or Production.
```
irvinebroque:
```suggestion
```

I get what this is saying, but think clearest if you just say how it works than "separate stored environment" (what's a "stored environment", etc.) - and think the sentence below conveys same thing clearly

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+You can set up your local development environment by pulling down environment settings from **Preview** (default) or **Production**.
```
irvinebroque:
```suggestion
You can set up your local development environment by pulling down settings from your **Preview environment**:
```

(think you convey the production bit below

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+By default, `wrangler env pull` pulls from your **Preview** environment:
+
+<PackageManagers type="exec" pkg="wrangler" args="env pull" />
+
+This creates a `.env.local` file in your project root that contains:
```
irvinebroque:
```suggestion
This creates a `.env.local` file in your project's root directory that contains:
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+# MY_BUCKET=@r2_bucket:preview-bucket
+```
+
+You can also pull from **Production**:
+
+<PackageManagers type="exec" pkg="wrangler" args="env pull --prod" />
+
+### Local vs Remote bindings
+
+By default, all bindings in your `.env.local` file are commented out. This means [Miniflare](https://miniflare.dev/) will create local simulations of your resources (like [KV namespaces](/kv/) or [D1 databases](/d1/)). This lets you develop offline without costs or needing to connect to deployed resources.
```
irvinebroque: Customers shouldn't need to think about Miniflare — what they interact with for local dev is either `wrangler dev` or `vite dev` with the Cloudflare Vite Plugin.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+You can pull from **Production** with `wrangler env pull --prod` if you need to connect to production resources during local development.
+
+To connect a binding to its remote resource in **Preview** or **Production**, uncomment the line in your `.env.local` file:
```
irvinebroque:
```suggestion
To make a binding run remotely in local dev, uncomment the line in your `.env.local` file:
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+# Environment variables from Preview environment
+API_KEY=your-preview-api-key
+ENVIRONMENT=preview
+
+# Bindings (commented out by default - simulated locally)
+# Uncomment to connect to the remote resource
+# MY_KV=@kv_namespace:abc123preview
+# DB=@d1:staging-db-id-456
+# MY_BUCKET=@r2_bucket:preview-bucket
+```
```
irvinebroque, line 617: I think somewhere, maybe here, maybe below, this section would benefit from a short bit that shows a use case for this

Ex — that each person can have their own remote bindings to different resources, with that configuration _not_ committed to source code.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+You can set up your local development environment by pulling down environment settings from **Preview** (default) or **Production**.
+
+By default, `wrangler env pull` pulls from your **Preview** environment:
+
+<PackageManagers type="exec" pkg="wrangler" args="env pull" />
+
+This creates a `.env.local` file in your project root that contains:
+
+```ini title=".env.local"
+# Environment variables from Preview environment
```
irvinebroque: Can we be explicit and separate out the secrets from the environment variables in this? So that people don't mix them up?

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+- `name` - Your Worker's name
+- `compatibility_date` - Which API behaviors your code uses
+- `compatibility_flags` - Specific feature flags
+- `main` - Your code's entry point file
+- `migrations` - Durable Object migration definitions
+
+## Legacy
+
+If you have an existing Wrangler configuration, you can still use default environments without migrating to the new `env` structure.
```
irvinebroque:
```suggestion
If you have an existing Wrangler configuration, you can still use production and preview environments without migrating your Worker's `wrangler.jsonc` file to the new `env` structure.
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+        "database_name": "preview-db",
+        "database_id": "staging-db-id-456"
+      }
+    ]
+  }
+}
+```
+</WranglerConfig>
+
+This format is valid and will work with default environments. You can keep your existing configuration structure and simply add the `preview` block to define different values for **Preview**.
```
irvinebroque, line 1480: In this example, are there inheritable keys? Asking since it's not obvious what the behavior is, when production is implicit like this

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+      "ENVIRONMENT": "dev-preview"
+    },
+    "kv_namespaces": [
+      {
+        "binding": "MY_KV",
+        "id": "dev-preview-kv-id"
+      }
+    ]
+  },
+  "env": {
```
irvinebroque, line 1518: It is worthwhile, but it is going to be confusing that Wrangler Environments uses `env` and so does the `env` object in environments...

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+  }
+}
+```
+</WranglerConfig>
+
+This format is valid and will work with default environments. You can keep your existing configuration structure and simply add the `preview` block to define different values for **Preview**.
+
+### Using with Wrangler Environments
+
+If you're using [Wrangler Environments](/workers/wrangler/environments/), each environment creates a separate Worker with the name `<worker-name>-<environment-name>`. You can add a `preview` block within each Wrangler environment to give each Worker its own **Production** and **Preview** configurations with separate **PR Previews**. For example:
```
irvinebroque:
```suggestion
When you use [Wrangler Environments](/workers/wrangler/environments/) and run `wrangler deploy --env staging`, a separate Worker with the name `<worker-name>-<environment-name>` is deployed. Each Wrangler Environment describes an entirely separate Worker. Wrangler Environments are best for deploying an entirely separate staging service, or deploying the same application to multiple distinct production environments.

It is possible to use Wrangler Environments and Workers Environments together. You can add a `preview` block within each Wrangler environment to give each Worker its own **Production** and **Preview** configurations with separate **PR Previews**. For example:
```

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+        ]
+      }
+    }
+  }
+}
+```
+</WranglerConfig>
+
+In this example:
+- **`my-worker`** (top-level) has Production deployment and PR Previews using `dev-preview-kv-id`
```
irvinebroque:
```suggestion
- **`my-worker`** (top-level) has Production deployment and PR Previews that use the KV namespace `dev-preview-kv-id`
```

(and below)

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+
+1. **We don't pass bindings** when creating the preview version - they automatically come from the `preview_settings` template we configured earlier. The API applies those settings (`ENVIRONMENT`, `MY_KV`) without us specifying them again.
+
+2. **`{ params: { deploy: true } }`** - This creates the version and deploys it in one step, like adding `?deploy=true` to the API URL.
+
+This keeps your test data separate from production while making it easy to deploy previews - you set up the template once, then every preview automatically uses those settings! Secrets are managed separately using the Wrangler CLI as shown in the [Set secrets](#set-secrets-1) section above.
+
+</TabItem>
+
+<TabItem label="Terraform">
```
irvinebroque, line 1271: nice one with the TF example

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1576 @@
...
+An **environment** is a collection of settings that determines how your Worker runs - such as which resources it connects to, what [environment variables](/workers/configuration/environment-variables/) it uses, and what domains it's available on.
+
+Every Worker has two default **environments** that map to specific phases in your workflow:
+
+- **[Production](#production)**: For serving your live application to real users on your [`*.workers.dev` subdomain](/workers/configuration/routing/workers-dev/), [custom domain](/workers/configuration/routing/custom-domains/), or [route](/workers/configuration/routing/routes/).
+
+- **[Preview](#preview)**: For testing and sharing Pull Request changes before they go live. Each Pull Request deploys its own unique **PR Preview** using the **Preview** environment settings.
+
+You can also easily set up your **[local development environment](#local-development)** by pulling down your **Preview** settings with `wrangler env pull` to develop and test changes on your local machine.
+
```
irvinebroque, line 34: (non-blocking)

There is some kind of awesome diagram of the lifecycle of:

1. Pull down configuration and develop locally
2. Create a pull request
3. Get a preview URL back that uses your preview environment, test out changes
4. Merge a pull request
5. That gets deployed to production, using your production environment

That really shows the end-to-end and might help cement the right mental model for people

irvinebroque (commented):
Looking great!

src/content/docs/workers/configuration/environments.mdx
korinne: I think the dash needs to be the primary way, because if you change some value in your `preview` config and then merge your change in -- you'll merge that override into that shared `preview` setting for everyone in the future to use

src/content/docs/workers/configuration/environments.mdx
korinne: I get what you're saying here and will fix -- but something we aimed to do with remote bindings was to make clear that Miniflare was responsible for the local simulations, and wrangler and vite were clients to Miniflare. Made it easier to explain how Wrangler and Vite supported things.

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1585 @@
...
+      }
+    }
+  }
+}
+```
+</WranglerConfig>
+
+### Domains
+
+You can configure which domains your Worker is available on by setting the `domains` field. If you don't include a `domains` field, your existing domain configuration stays unchanged.
```
irvinebroque, line 481:
```suggestion
You can configure which domains your Worker is available on by setting the `domains` field. If you omit the `domains` field from your configuration and then deploy, domains that you have previously added to your Worker will remain connected — they will not be removed.
```

I understood what this meant given context but to be most specific

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1585 @@
...
+
+<WranglerConfig>
+```jsonc title="wrangler.jsonc"
+// Use the free workers.dev subdomain
+{
+  "name": "my-worker",
+  "production": {
+    "domains": ["my-worker.workers.dev"]
```
irvinebroque, line 489: Does this being a string value close doors to doing something like:

```
"domains": [ { "domain": "my-worker.workers.dev", "access_enabled": true }]
```

Or other settings? Even if ultimately something like:

```js
const domain1 = { ... }

const worker = {
  domains: [domain1],
  ...
}
```

may ultimately be most ideal future state

Think this is simplest but wondering about what it would look like to give someone the ability to enable something like [this](https://developers.cloudflare.com/changelog/2025-10-03-one-click-access-for-workers/) from configuration

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1585 @@
...
+```
+</WranglerConfig>
+
+### Domains
+
+You can configure which domains your Worker is available on by setting the `domains` field. If you don't include a `domains` field, your existing domain configuration stays unchanged.
+
+<WranglerConfig>
+```jsonc title="wrangler.jsonc"
+// Use the free workers.dev subdomain
```
irvinebroque, line 485: Let's say we ship this new `domains` format.

Then, I go and run `npm create cloudflare` or do something else to spin up a new Worker. And I get a configuration file like this:

```jsonc
{
        "$schema": "node_modules/wrangler/config-schema.json",
        "account_id": "<ID>",
        "name": "worker-name",
        "main": "src/index.ts",
        "compatibility_date": "2025-10-01",
        "observability": {
                "enabled": true
        }
}
```

I do not have anything set currently for `workers_dev = true`. I also do not have anything for `domains`. And I am deploying something _new_ — not an existing Worker.

What are we saying should happen? Is workers.dev enabled or disabled for my Worker? I think kind of _has_ to be enabled otherwise it's a breaking change?

But then I go and add:

```jsonc
"domains": ["my-custom-domain.com"]
```

Does this _disable_ workers.dev that had previously been enabled, because workers.dev is not set?

src/content/docs/workers/configuration/environments.mdx
```diff
@@ -0,0 +1,1585 @@
...
+    "domains": ["app.example.com"]
+  }
+}
+
+// Use both workers.dev and a custom domain
+{
+  "name": "my-worker",
+  "production": {
+    "domains": [
+      "my-worker.workers.dev",
```
irvinebroque, line 506: I think will be clearest to reason about this feature using current `workers.dev` behavior — because no matter what have to support that anyways.

`worker-name.account-name.workers.dev`

Which is a lot for the _user_ to have to add in here? When it is information that we already know — which is why `workers_dev = true` is possible?

I could see this format working in _future state_ beyond workers.dev - that is not coupled to either account name or worker name - that we have talked about — but still harder to see for `workers.dev` in particular as it stands today, where URL is coupled to both account name and worker name.

PR #26489, opened 2025-11-13
fix the display of wrangler positional args

vicb:
### Screenshots (optional)

Before

image

After

image

### DIFF
--- a/src/components/WranglerCommand.astro
+++ b/src/components/WranglerCommand.astro
@@ -28,6 +28,8 @@ const { globalFlags } = commands;
 const positionals = definition.positionalArgs
 	?.map((p) => `[${p.toUpperCase()}]`)
 	.join(" ");
+
+const positionalSet = new Set(definition.positionalArgs);
 ---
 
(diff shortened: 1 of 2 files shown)


src/components/WranglerCommand.astro
```diff
@@ -45,7 +47,12 @@ const positionals = definition.positionalArgs
 		Object.entries(definition.args)
 			.filter(([_, value]) => !value.hidden)
 			.map(([key, value]) => {
-				return <WranglerArg key={key} definition={value} />;
+				return (
+					<WranglerArg
+						key={key}
+						definition={{ ...value, positional: positionalSet.has(key) }}
```
ascorbic, line 53: Rather than smooshing everything into this untyped object, maybe it would be better to have specific props

vicb: Agreed, what about we clean that up in a follow-up PR?

vicb: _I initially added `positional` as a separate prop..._

Oxyjun (approved):

kodster28 (approved):

PR #26518, opened 2025-11-13
Remove obsolete `metroCode` from request properties

johnspurlock-skymethod:


pedrosousa:
The field description has already been updated. The field is not being populated, but this may change in the future.

PR #26536, opened 2025-11-14
[Workflows] Adding warning about nested steps & example for instance restarts

mia303:
### Summary

Adding warning about nested steps & code sample for instance restarts

### DIFF
--- a/src/content/docs/workflows/build/trigger-workflows.mdx
+++ b/src/content/docs/workflows/build/trigger-workflows.mdx
@@ -166,7 +166,7 @@ The following example shows how you can use the `restart` method to maintain a d
 
 - Chunks of a document are stored in, and retrieved from, R2 with an identifiable object [prefix](/r2/api/workers/workers-api-reference/#r2listoptions)
 - Each [R2 object key](/r2/api/workers/workers-api-reference/#r2object-definition) is used as the `instanceId` for processing that chunk
-- The workflow retrieves 500 chunks at a time, processes the chunks, and updates their prefix
+- The workflow retrieves 100 chunks at a time, processes the chunks, and updates their prefix
 - Then, if there are remaining objects with the original prefix, the Workflow instance restarts
 
 ```TypeScriptExample
--- a/src/content/docs/workflows/build/trigger-workflows.mdx
+++ b/src/content/docs/workflows/build/trigger-workflows.mdx
@@ -5,7 +5,6 @@ tags:
   - Bindings
 sidebar:
   order: 3
-
 ---
 
 import { WranglerConfig } from "~/components";
@@ -14,7 +13,7 @@ You can trigger Workflows both programmatically and via the Workflows APIs, incl
 
 1. With [Workers](/workers) via HTTP requests in a `fetch` handler, or bindings from a `queue` or `scheduled` handler
 2. Using the [Workflows REST API](/api/resources/workflows/methods/list/)
-2. Via the [wrangler CLI](/workers/wrangler/commands/#workflows) in your terminal
+3. Via the [wrangler CLI](/workers/wrangler/commands/#workflows) in your terminal
 
 ## Workers API (Bindings)
 
@@ -22,10 +21,10 @@ You can interact with Workflows programmatically from any Worker script by creat
 
 You can interact with a Workflow:
 
-* Directly over HTTP via the [`fetch`](/workers/runtime-apis/handlers/fetch/) handler
-* From a [Queue consumer](/queues/configuration/javascript-apis/#consumer) inside a `queue` handler
-* From a [Cron Trigger](/workers/configuration/cron-triggers/) inside a `scheduled` handler
-* Within a [Durable Object](/durable-objects/)
+- Directly over HTTP via the [`fetch`](/workers/runtime-apis/handlers/fetch/) handler
+- From a [Queue consumer](/queues/configuration/javascript-apis/#consumer) inside a `queue` handler
+- From a [Cron Trigger](/workers/configuration/cron-triggers/) inside a `scheduled` handler
+- Within a [Durable Object](/durable-objects/)
 
 :::note
 
@@ -59,9 +58,9 @@ The `binding = "MY_WORKFLOW"` line defines the JavaScript variable that our Work
 
 The following example shows how you can manage Workflows from within a Worker, including:
 
-* Retrieving the status of an existing Workflow instance by its ID
-* Creating (triggering) a new Workflow instance
-* Returning the status of a given instance ID
+- Retrieving the status of an existing Workflow instance by its ID
+- Creating (triggering) a new Workflow instance
+- Returning the status of a given instance ID
 
 ```ts title="src/index.ts"
 interface Env {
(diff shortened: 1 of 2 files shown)


maxvp (approved):

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -162,6 +162,86 @@ await instance.restart() // Returns Promise<void>
 
 Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
 
+The following example shows how you can use the `restart` method to maintain a document processing pipeline:
```
avenceslau, line 168: I can explain further on Monday. There are some caveats  with having workflows that restart themselves, and we should explain them here.

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -162,6 +162,86 @@ await instance.restart() // Returns Promise<void>
...
+
+- Chunks of a document are stored in, and retrieved from, R2 with an identifiable object [prefix](/r2/api/workers/workers-api-reference/#r2listoptions)
+- Each [R2 object key](/r2/api/workers/workers-api-reference/#r2object-definition) is used as the `instanceId` for processing that chunk
+- The workflow retrieves 500 chunks at a time, processes the chunks, and updates their prefix
```
pombosilva:
```suggestion
- The workflow retrieves 100 chunks at a time, processes the chunks, and updates their prefix
```

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -162,6 +162,86 @@ await instance.restart() // Returns Promise<void>
...
+- Then, if there are remaining objects with the original prefix, the Workflow instance restarts
+
+```TypeScriptExample
```
pombosilva: This should be a tag <TypeScriptExample filename="index.ts">

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -162,6 +162,86 @@ await instance.restart() // Returns Promise<void>
...
+export class MyWorkflow extends WorkflowEntrypoint<Env, Params> {
+  async run(event: WorkflowEvent<Params>, step: WorkflowStep) {
+    // Step 1: List first 500 files from R2 with prefix "chunk/"
```
pombosilva:
```suggestion
    // Step 1: List first 100 files from R2 with prefix "chunk/"
```

src/content/docs/workflows/build/trigger-workflows.mdx
avenceslau: Restarting an instance from within itself, will keep that instance running, without going through the queued state. This can lead to concurrency slot deprivation if users are not careful.

We should at least warn that this is something that can happen.

avenceslau: Restarts delete all state except input params. Maybe we should explain that users should keep all relevant state outside of the instance itself. (That's the reason why the example uses R2)

With restarting/recursive workflows, progress should be tracked outside of workflow. This is a bit of an anti-pattern, but as it enables some different workloads I think we should still keep the example.

avenceslau: I have cleaned up the example a bit, let me know what you think.

```ts
export class MyWorkflow extends WorkflowEntrypoint<Env, Params> {
  async run(event: WorkflowEvent<Params>, step: WorkflowStep) {
    const listResult = await step.do("list files from R2", async () => {
      // Restarts delete all state with the exception of search params.
     // So we will need a cursor to keep track of progress
      const maybeCursor = await this.env.CHUNKS_BUCKET.get(
        `cursor/${event.instanceId}`,
      );

      const cursor = maybeCursor
        ? (await maybeCursor.json<{ cursor: string }>()).cursor
        : undefined;

      // Get the first 100 chunks to process
      const listed = await this.env.CHUNKS_BUCKET.list({
        prefix: "chunked/",
        limit: 100,
        cursor,
      });

      if (listed.truncated) {
        const cursor = listed.cursor;
        await this.env.CHUNKS_BUCKET.put(
          `cursor/${event.instanceId}`,
          JSON.stringify({ cursor }),
          {
            httpMetadata: {
              contentType: "application/json",
            },
          },
        );
      }

      return {
        keys: listed.objects.map((obj) => obj.key),
        truncated: listed.truncated,
      };
    });

    for (const key of listResult.keys) {
      await step.do(`process file: ${key}`, async () => {
        return await doSomething(key, this.env.CHUNKS_BUCKET, event.instanceId);
      });
    }

    await step.do("restart workflow", async () => {
      if (listResult.truncated) {
        // There are still chunks to process, restarting the workflow to continue processing
        const instance = await this.env.CHUNK_PROCESSOR.get(event.instanceId);
        await instance.restart();
      } else {
        // If the workflow has ended we reset the cursor, so that we can preserve determinism
        // if/when we re-run this workflow we will still process all the files
        await this.env.CHUNKS_BUCKET.delete(`cursor/${event.instanceId}`);

        return "was not truncated, done";
      }
    });
  }
}
```

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -160,7 +160,80 @@ let instance = await env.MY_WORKFLOW.get("abc-123")
 await instance.restart() // Returns Promise<void>
 ```
 
-Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
+Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time. However, the restarted instance will maintain the concurrency slot of the previous instance.
```
avenceslau, line 163: I would say something like the restarted instance will maintain any concurrency slot it might have.

avenceslau: If the instance is not running it won't keep a concurrency slot. And for restart currently we don't create a new instance. We clean up the state an run from the beginning

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -160,7 +160,80 @@ let instance = await env.MY_WORKFLOW.get("abc-123")
 await instance.restart() // Returns Promise<void>
 ```
 
-Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
+Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time. However, the restarted instance will maintain the concurrency slot of the previous instance. 
+
+:::caution[Exercise caution when using the `restart()` method to avoid concurrency slot deprivation. If you call the `restart()` method recursively without a conditional, that instance will run indefinitely, effectively eliminating a concurrency slot from your queue.]
```
avenceslau, line 165: "concurrency slot from your queue", don't know if we want to call it queue

src/content/docs/workflows/build/trigger-workflows.mdx
```diff
@@ -160,7 +160,80 @@ let instance = await env.MY_WORKFLOW.get("abc-123")
 await instance.restart() // Returns Promise<void>
 ```
 
-Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time.
+Restarting an instance will immediately cancel any in-progress steps, erase any intermediate state, and treat the Workflow as if it was run for the first time. However, the restarted instance will maintain the concurrency slot of the previous instance. 
+
+:::caution[Exercise caution when using the `restart()` method to avoid concurrency slot deprivation. If you call the `restart()` method recursively without a conditional, that instance will run indefinitely, effectively eliminating a concurrency slot from your queue.] 
+
+
+The following example shows how you can use the `restart` method to maintain a document processing pipeline:
```
pombosilva, line 168: I feel like we should enforce concurrency here with something like: "... to maintain a document processing pipeline **holding a concurrency slot**:"

avenceslau:
We should, given that we are discouraging the usage of nested steps, make sure to change the promise.race example, races should be done like so. U

This avoids nesting, and ensures that the result is always the same.

```ts
await step.do("race step", async () => {
   return await Promise.race([asyncCallOne(), asyncCall2()])
})
```

PR #26540, opened 2025-11-14
Mchen/experimental

mchenco:
### Summary

This is a big PR to help Workers AI users get better input/output code examples for each model.

### Screenshots (optional)

Screenshot 2025-11-14 at 2 10 04 PM

## File Tree

  src/util/workers-ai/
  ├── schema-formats.ts      # Core detection system & schema matchers
  ├── input-formats.ts       # Input format registry
  ├── output-formats.ts      # Output format registry
  ├── templates.ts           # JSON example templates
  └── format-labels.ts       # Format labeling logic

##  Quick Reference

  | Task                | File to Edit                                         |
  |---------------------|------------------------------------------------------|
  | Add new model type  | input-formats.ts, output-formats.ts, templates.ts    |
  | Fix JSON examples   | templates.ts                                         |
  | Format not detected | Check priority in input-formats.ts/output-formats.ts |
  | Need new matcher    | Add to schema-formats.ts                             |

### How to Add a New Model Type

  1. Add templates (templates.ts):
  export const myModelInput = () => `{ "text": "Example" }`;
  export const myModelOutput = () => `{ "result": "Output" }`;

  2. Add input matcher (input-formats.ts):
  {
    id: 'my-model',
    label: 'My Format',
    priority: 1,  // Lower = checked first
    matches: (schema, ctx) => ctx.taskName === 'My Task',
    generateExample: () => templates.myModelInput(),
  }

  3. Add output matcher (output-formats.ts):
  {
    id: 'my-model-output',
    label: 'JSON',
    priority: 1,
    matches: (schema, ctx) => ctx.taskName === 'My Task',
    generateExample: () => templates.myModelOutput(),
  }


mchenco:
no need to review it - just experimenting. pushing up to get a preview branch and feedback

src/pages/workers-ai/models/[name].astro
```diff
@@ -230,7 +230,9 @@ const starlightPageProps = {
 		model.name === "@cf/deepgram/flux" && (
 			<Aside>
 				<p>
-					For the month of October 2025, Deepgram's Flux model will be free to use on Workers AI. Official pricing will be announced soon and charged after the promotional pricing period ends on October 31, 2025.
+					For the month of October 2025, Deepgram's Flux model will be free to
```
Comment (craigsdennis) [inline src/pages/workers-ai/models/[name].astro:233]: Remove?

src/util/workers-ai/templates.ts
```diff
@@ -0,0 +1,470 @@
+// Binary templates (reused across types)
```
craigsdennis, line 1: This file seems like a bad idea to me. There is already a file based solution. Do you want all future edits in here in a single file? If you are going to move to this I suggest moving to the GitHub component solution.

src/util/workers-ai/output-formats.ts
```diff
@@ -0,0 +1,273 @@
+import type { FormatMatcher, Schema, FormatContext } from "./schema-formats";
+import { schemaMatchers } from "./schema-formats";
+import * as templates from "./templates";
+
+export const OUTPUT_FORMATS: FormatMatcher[] = [
```
craigsdennis, line 5: Is Workers AI team maintaining this going forward?

src/util/workers-ai/output-formats.ts
```diff
@@ -0,0 +1,273 @@
...
+	},
+
+	// Translation Output
+	{
+		id: "translation-output",
+		label: "JSON",
+		priority: 1,
+		matches: (schema: Schema, ctx: FormatContext) =>
+			ctx.taskName === "Translation" &&
+			schemaMatchers.hasProperty("properties.translated_text")(schema),
```
craigsdennis, line 234: What's this?

src/util/workers-ai/output-formats.ts
craigsdennis: I really think these should come from types, but I'm not gonna stop if you want multiple sources of truth. 

Raising that this will be EXTREMELY frustrating if the TypeScript type red squiggles.

@thatsKevinJain

src/util/workers-ai/schema-formats.ts
```diff
@@ -0,0 +1,161 @@
+// Type definitions for format matching system
+export interface Schema {
```
craigsdennis, line 2: This is a dupe of work afaict right?

src/util/workers-ai/format-labels.ts
```diff
@@ -0,0 +1,85 @@
...
+		const label = option.title
+			.replace(/\s+Response$/i, "")
+			.replace(/\bOuput\b/g, "Output")
+			.replace(/^[A-Z0-9\s]+\s+(Input|Output)\s+/i, "");
+
+		if (label.includes(" ")) return label;
+
+		// Handle underscore-separated titles
+		const parts = label.split("_");
+		const keywords = [
```
craigsdennis, line 28: Note that this will need to be updated as features get added

src/pages/workers-ai/models/[name].astro
```diff
@@ -289,20 +292,75 @@ const starlightPageProps = {
...
+					{model.task.name === "Text-to-Image" ? "image" : "audio"}
+					{model.schema.output.contentType &&
+					typeof model.schema.output.contentType === "string"
+						? ` in ${model.schema.output.contentType.split("/")[1].toUpperCase()} format`
+						: " (check the model's output schema for format details)"}
+					.
 				</p>
+				<div class="mb-4">
+					<Code
+						code={`const response = await env.AI.run("${model.name}", { prompt: "..." });
```
Comment (craigsdennis) [inline src/pages/workers-ai/models/[name].astro:330]: Seems like this should go in the templates, but maybe I'm misunderstanding

craigsdennis (commented):
First off, this looks great!

This is definitely AI generated and feels that way. 

How does editing it feel?

I do think most of this could be achieved in a set place and also help the TypeScript types. I do feel like this is moving very far from something I can read/maintain, and I would like to make sure that is known. Were this to land, I do not feel comfortable making updates to it, and I imagine most folks would feel that way too. 

If this feels like a motion we want to get into, that works. You just will lose editors.

Part of me thinks you should build these examples externally in a script and include the output as files instead of doing all the looping. Would be easier to maintain and test and keep things separated. But I'm not on that roadmap, so I don't know what that would look like.

I would suggest @celso takes a quick look in here too.

Also of note Replicate has this sorted a bit with a UI, and not sure how this will go together in the future.

PR #26581, opened 2025-11-17
[CF1] Audit redirects

maxvp:
PCX-19765


public/__redirects
```diff
@@ -2438,6 +2330,7 @@
 /cloudflare-one/api-terraform/access-api-examples/* /api/resources/zero_trust/subresources/access/subresources/policies/ 301
 
 # Email Security new revamp (dynamics)
+/cloudflare-one/email-security/detection-settings/* /cloudflare-one/email-security/settings/detection-settings/:splat 301
 /cloudflare-one/email-security/detection-settings/detection-settings/* /cloudflare-one/email-security/settings/detection-settings/:splat 301
 /cloudflare-one/email-security/detection-settings/impersonation-registry/* /cloudflare-one/email-security/settings/detection-settings/impersonation-registry/:splat 301
```
pedrosousa: Since the order of redirects matters (if there are multiple redirects for the same source path, the top-most redirect is applied), I'd start with the most specific redirect and end with the most generic one.
If we don't do this, redirects 2 and 3 below would be ignored.

Also, the 3rd redirect under "Email Security new revamp" seems to be contained in the first, so we can remove it.

```suggestion
/cloudflare-one/email-security/detection-settings/detection-settings/* /cloudflare-one/email-security/settings/detection-settings/:splat 301
/cloudflare-one/email-security/detection-settings/* /cloudflare-one/email-security/settings/detection-settings/:splat 301
```

pedrosousa (approved):

PR #26097, opened 2025-10-27
pubsub: remove docs

elithrar:
Remove the Pub/Sub docs: private beta was shut down Aug 20th 2025.

Issue #4985, opened 2022-06-30
Flutter example codes to generate tokens (android & ios)


yuwono95:
### Which Cloudflare product(s) does this pertain to?

Stream

### Subject Matter

Flutter example codes (android & ios)

### Content Location

https://developers.cloudflare.com/stream/viewing-videos/securing-your-stream/#step-2-generate-tokens-using-the-key

### Additional information

I've been days trying to code this "generate tokens"  using Flutter, but no luck. If you guys can make this, it would be awesome. Thank you.

yuwono95:
Got it, might be helpful for others.

First, install "crypto_keys" package from https://pub.dev/packages/crypto_keys

```
  import 'package:crypto_keys/crypto_keys.dart';

    .....

  // Global variables
  const keyID = '$KEYID';
  const videoID = '$VIDEOID';
  const keyJWK = {PRIVATE-KEY-IN-JWK-FORMAT};
  // expiresTimeInS is the expired time in second of the video
  const expiresTimeInS = 3600;

  String streamSignedUrl () {
      var strJWK = new String.fromCharCodes(base64.decode(keyJWK));
      var jsonJWK = json.decode(strJWK);

      var now = (DateTime.now().millisecondsSinceEpoch / 1000).round();
      var exp = now + expiresTimeInS;
      var header = {"alg": "RS256", "kid": keyID};
      var payload = {"sub": videoID, "kid": keyID, "exp": exp, "nbf": now};

      var encodedHeader = objectToBase64Url(header);
      var encodedPayload = objectToBase64Url(payload);
      var data = "$encodedHeader.$encodedPayload";

      var keyPair = new KeyPair.fromJwk(jsonJWK);
      var privateKey = keyPair.privateKey!;
      var signer = privateKey.createSigner(algorithms.signing.rsa.sha256);
      var signatureBytes = signer.sign(data.codeUnits);
      var encodedSignature = arrayBufferToBase64Url(signatureBytes.data);

      return "$data.$encodedSignature";
  }

  String arrayBufferToBase64Url(buffer) {
      return base64UrlEncode(buffer).replaceAll("+", "-").replaceAll("/", "_").replaceAll("=", "");
  }

  String objectToBase64Url(payload) {
    return arrayBufferToBase64Url(utf8.encode(json.encode(payload)));
  }
```

Regards.

Issue #1423, opened 2021-05-25
Document `caches.open`


kristianfreeman:
Per https://github.com/cloudflare/cloudflare-docs/pull/1395, `caches.default` is _not_ the only available cache to work from in the Cache API. We should document it in our [Cache runtime API page](https://developers.cloudflare.com/workers/runtime-apis/cache).

PR #26606, opened 2025-11-18
feat: publish html files as artifact

colbywhite:
This artifact will be used by later by external processes, i.e. a separate reverse build. This is the first step for removing the reverse build from the build time.

### DIFF
--- a/.github/workflows/publish-production.yml
+++ b/.github/workflows/publish-production.yml
@@ -29,6 +29,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ENABLE_LAST_MOD_IN_SITEMAP: true
+      - uses: actions/upload-artifact@v4
+        with:
+          name: site-html
+          path: dist
       - run: npx wrangler deploy
         name: Deploy to Cloudflare Workers
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-production.yml
```diff
@@ -29,6 +29,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ENABLE_LAST_MOD_IN_SITEMAP: true
+      - uses: actions/upload-artifact@v4
```
kodster28, line 32: Should we move to after `deploy to CF workers` just in case there might be errors on this step?

colbywhite: The utopia would be to restructure this workflow to have several jobs (particularly all the various uploads) that run in parallel based on dependencies. For now, I'll add a `continue-on-error` so it doesn't fail the build regardless of where it happens. 

When we remove the reverse build altogether, I'll restructure this workflow to embrace parallelism more.

kodster28 (approved):

.github/workflows/publish-production.yml
bitcoink188-cpu: #526 
#982 
#4985 
#26097 
#25744 
#1 
#34 
#23 
#6 
#69 
#20393 
#1423 
#32583 
#15 
#155 
#125 
#4609 
#26097 
#1 
#5 
#32583 
#1483 
8

PR #26619, opened 2025-11-19
[Workflows] Clarify NonRetryableError information

pombosilva:
### Summary

Clarify NonRetryableError docs regarding it being applied inside of a step.

### DIFF
--- a/src/content/docs/workflows/build/workers-api.mdx
+++ b/src/content/docs/workflows/build/workers-api.mdx
@@ -156,7 +156,7 @@ Refer to the [documentation on sleeping and retrying](/workflows/build/sleeping-
 
 {/* prettier-ignore */}
 - <code>throw new NonRetryableError(message: <Type text="string" />, name <Type text="string" /> <MetaInfo text="optional" />)</code>: <Type text="NonRetryableError" />
-  - When thrown inside [`step.do()`](/workflows/build/workers-api/#step), this error stops step retries, propagating the failure to the top level (the [run](/workflows/build/workers-api/#run) function) and causing the Workflow instance to fail.
+  - When thrown inside [`step.do()`](/workflows/build/workers-api/#step), this error stops step retries, propagating the error to the top level (the [run](/workflows/build/workers-api/#run) function). Any error not handled at this top level will cause the Workflow instance to fail.
   - Refer to the [documentation on sleeping and retrying](/workflows/build/sleeping-and-retrying/) to learn more about how Workflows steps are retried.
 
 ## Call Workflows from Workers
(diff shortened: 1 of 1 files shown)


src/content/docs/workflows/build/workers-api.mdx
```diff
@@ -156,8 +156,8 @@ Refer to the [documentation on sleeping and retrying](/workflows/build/sleeping-
 
 {/* prettier-ignore */}
 - <code>throw new NonRetryableError(message: <Type text="string" />, name <Type text="string" /> <MetaInfo text="optional" />)</code>: <Type text="NonRetryableError" />
-  - Throws an error that forces the current Workflow instance to fail and not be retried.
-  - Refer to the [documentation on sleeping and retrying](/workflows/build/sleeping-and-retrying/) to learn more about how Workflows are retried.
+  - When thrown inside [`step.do()`](/workflows/build/workers-api/#step), this error stops step retries, propagating the failure to the top level (the [run](/workflows/build/workers-api/#run) function) and causing the Workflow instance to fail.
```
LuisDuarte1: the instance might not fail if the user catches the `NonRetryableError` by `try {} catch {}`-ing the `step.do` (also applies to any other error, for that matter)

LuisDuarte1 (changes requested):

src/content/docs/workflows/build/workers-api.mdx
pombosilva: Yeah, will add something on that

celso (approved):

src/content/docs/workflows/build/workers-api.mdx
```diff
@@ -156,8 +156,8 @@ Refer to the [documentation on sleeping and retrying](/workflows/build/sleeping-
 
 {/* prettier-ignore */}
 - <code>throw new NonRetryableError(message: <Type text="string" />, name <Type text="string" /> <MetaInfo text="optional" />)</code>: <Type text="NonRetryableError" />
-  - Throws an error that forces the current Workflow instance to fail and not be retried.
-  - Refer to the [documentation on sleeping and retrying](/workflows/build/sleeping-and-retrying/) to learn more about how Workflows are retried.
+  - When thrown inside [`step.do()`](/workflows/build/workers-api/#step), this error stops step retries, propagating the error to the top level (the [run](/workflows/build/workers-api/#run) function). Any error not handled at this top level will cause the Workflow instance to fail.
```
LuisDuarte1, line 159:
```suggestion
  - When thrown inside [`step.do()`](/workflows/build/workers-api/#step), this error stops further retries, propagating the error to the top level (the [run](/workflows/build/workers-api/#run) function). Any error not handled at this top level will cause the Workflow instance to fail.
```

LuisDuarte1 (approved):

PR #26703, opened 2025-11-23
[Fundamentals] Add PageGym to third party page speed tests

razcoj:
Hi,

I would like to propose a **non-lighthouse** page speed testing tool for the list of third-party tools. It's free, and it's one of the most advanced and accurate out there (calibrated servers, dedicated resources for each test, etc).

[Example report](https://pagegym.com/speed/test/developers-cloudflare-com/ff7p6zz9nb)

Should be extremely effective at showing the difference that Cloudflare makes when it is enabled, versus when it is not.


razcoj:
Just wanted to add that for desktop users it can also provide this type of before and after comparison:

https://pagegym.com/compare/fe2q8pol8e/qzom7ap0xh

razcoj:
@dcpena please let me know if there's anything else you want to know about the tool.

dcpena (approved):

PR #26710, opened 2025-11-24
Optimise sitemap creation

ascorbic:
### Summary

The sitemap generator uses git to get the last modified time. Currently this execs git for every path, which is very slow – currently accounting for half of the build time. This PR optimises this by instead getting all of the last modified times with a single git log command, and caching them in a `Map`. This reduces the time on my local machine from ~5 minutes to 400 ms, and in production from around 9 mins to ~800ms, cutting build time approximately in half.

### DIFF
--- a/.github/workflows/publish-production.yml
+++ b/.github/workflows/publish-production.yml
@@ -28,7 +28,6 @@ jobs:
         name: Build
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          ENABLE_LAST_MOD_IN_SITEMAP: true
       - uses: actions/upload-artifact@v4
         continue-on-error: true
         with:
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -61,23 +61,82 @@ const customCss = await autogenStyles();
 
 const RUN_LINK_CHECK =
 	process.env.RUN_LINK_CHECK?.toLowerCase() === "true" || false;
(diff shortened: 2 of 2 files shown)


.github/workflows/publish-production.yml
```diff
@@ -28,7 +28,6 @@ jobs:
         name: Build
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          ENABLE_LAST_MOD_IN_SITEMAP: true
```
ascorbic, line 31: Not needed anymore, as it's fast

kodster28: We should wait to remove this until we've confirmed that across a couple builds. Making assumptions about what does / doesn't affect the build time is sorta how we got into the sitemap generation (and it's build time regression) situation.

kodster28:
cc: @colbywhite

.github/workflows/publish-production.yml
ascorbic: Sure. I did benchmark a few scenarios locally though, and removing it here means it runs in this PR, which it wouldn't otherwise.

colbywhite (approved):
@ascorbic , Thanks for the input. I'm currently working on this exact logic but was implementing it slightly differently to account for edge cases. Let's merge this for now and I'll refactor it later today.

PR #26711, opened 2025-11-24
Cancel in-progress production builds

ascorbic:
### Summary

We already put PRs in concurrency groups so that new pushes cancel in-progress builds, but we don't do that for production. This PR changes it so that new merges to production cancel any in-progress builds. We only care about the most recent commit, so it's wasteful to keep running builds that will be out of date very soon.

### DIFF
--- a/.github/workflows/publish-production.yml
+++ b/.github/workflows/publish-production.yml
@@ -4,6 +4,10 @@ on:
     branches:
       - production
 
+concurrency:
+  group: publish-production
+  cancel-in-progress: true
+
 jobs:
   publish:
     name: Production


threepointone (approved):
love it

kodster28:
We don't want this. If you cancel in progress builds, you can get into an endless queuing situation where you might then wait longer for something to publish. Seen this happen quite a lot in other contexts (search indexing, for example) and it can be maddening.

In progress builds also have no effect on the speed / queues for other builds.

Happy to provide more context or discuss more if I'm missing something here. Just let me know!

PR #26713, opened 2025-11-24
Update azure-vpn-gateway.mdx

kbitr:
I've clarified the instruction to explicitly state that both tunnel interface addresses are required for Active/Active configuration:

Untitled-2.md#126
4. If using an Active/Active configuration, also add the Interface Address of the second Magic IPsec Tunnel from the Cloudflare Dashboard in CIDR notation (for example, `10.252.3.56/32`) under **Address Space(s)**. Both tunnel interface addresses must be configured in the Local Network Gateway Address Space to ensure both tunnels remain healthy. Key improvements:

Added "also" to emphasize this is in addition to the first tunnel Changed example IP to 10.252.3.56/32 (different from the first) to clarify it's a separate address Added explicit statement that both tunnel interface addresses must be configured to ensure both tunnels remain healthy This eliminates the ambiguity and prevents users from incorrectly configuring only the second tunnel's interface address.

and fixed a typo

- [ x] The change adheres to the [documentation style guide](https://developers.cloudflare.com/style-guide/).

### DIFF
--- a/src/content/partials/networking-services/magic-wan/third-party/azure-vpn-gateway.mdx
+++ b/src/content/partials/networking-services/magic-wan/third-party/azure-vpn-gateway.mdx
@@ -70,7 +70,7 @@ The time it takes for Azure to fully provision the Virtual Network Gateway depen
    10. **Replay protection**: **Enable**.
 3. If you are using the Active/Active configuration, select **Add IPsec tunnel** and repeat step 2 to create the second Magic WAN IPsec tunnel. Use the same **Cloudflare endpoint** as for the first tunnel.
 4. Select **Add Tunnels** when you are finished.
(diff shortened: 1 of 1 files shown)


kbitr:
Note:

I changed the example to use `10.252.3.56/32` for the second tunnel configuration value. While adapting the example makes sense, I believe `.56` will not be a valid value for the reasons below.

The CIDR for the first tunnel example is `10.252.3.54/31`, where:

  * `10.252.3.54/31` is the Cloudflare Peer IP.
  * `10.252.3.55/31` is the Azure Peer IP (which needs to be configured in Cloudflare).

Therefore, if we use a second tunnel subnet, it should probably be `10.252.3.56/31`, where:

  * `10.252.3.56/31` is the Cloudflare Peer IP.
  * `10.252.3.57/31` is the Azure Peer IP (which needs to be configured in Cloudflare and mentioned in the documentation).

Alternatively, perhaps it would be better to introduce distinct values for both tunnels (we previously used `10.252.1.54/31` and `10.25.2.54/31` for the two tunnels) and use those consistently throughout the document.

marciocloudflare (approved):

PR #26763, opened 2025-11-26
[Workers] Update logpush.mdx

luishcardoso:
Hello team, 

### Summary

Per CF dashboard, the actual menu in question is **Analytics & logs > Logpush**

Screenshot 2025-11-26 at 12 46 57

Thanks.


irvinebroque (approved):

kodster28:
We want to keep following the pattern of utilizing DashButtons + a bit more generic instructions. This doesn't seem to impair the current experience... from data, we can see that the majority of folks here just click on the button (which is a direct deep link to the page in question)

image

We're working with the Dash team to actually be able to construct the route tree programmatically... but until that happens we should keep a bit more generic to make this maintainable.

PR #26766, opened 2025-11-26
[Localization] Spanish PoC for fundamentals

kodster28:
### Summary

CED-7, adding routing for a localized version of our `/fundamentals/` docs in Spanish.

This adjusts the underlying Workers routing, which we have to do because workers evaluate before other DNS records (meaning other methods don't work as expected).

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -128,6 +128,16 @@ export default defineConfig({
 			},
 			sidebar,
 			customCss,
+			locales: {
+				root: {
+					label: "English",
+					lang: "en",
+				},
+				"es-la": {
+					label: "Español",
+					lang: "es",
+				},
+			},
 			pagination: false,
 			plugins: [
 				...(RUN_LINK_CHECK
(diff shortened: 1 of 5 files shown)


astro.config.ts
```diff
@@ -128,6 +128,16 @@ export default defineConfig({
...
+			locales: {
+				root: {
+					label: "English",
+					lang: "en",
+				},
+				"es-la": {
+					label: "Español",
+					lang: "es",
+				},
+			},
```
colbywhite, line 140: Are we sure `es-la` is a valid [locale](http://www.lingoes.net/en/translator/langcode.htm)? `es-mx` maybe? Just `es`?

src/components/overrides/Sidebar.astro
```diff
@@ -208,29 +234,29 @@ const [product, module] = Astro.url.pathname.split("/").filter(Boolean);
...
 
 		// Clear search on Escape key
-		searchInput.addEventListener('keydown', (e) => {
-			if (e.key === 'Escape') {
-				searchInput.value = '';
-				filterSidebarItems('');
+		searchInput.addEventListener("keydown", (e) => {
+			if (e.key === "Escape") {
+				searchInput.value = "";
+				filterSidebarItems("");
```
colbywhite, line 254: - Let's get your IDE to follow the prettier rules.
- Let's run the prettier rules as part of the build.

colbywhite (commented):
We talked in person. Sounds like there's another way to test this out with out using Astro's router. Publishing the comments I had unpublished for posterity's sake.

PR #26802, opened 2025-11-28
Adds Python examples in durable-objects/api/ files.

dom96:
### Summary

Adds Python examples alongside JS/TS ones in durable-objects/api


src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -115,8 +119,47 @@ export class AlarmExample extends DurableObject {
...
 
+</TabItem>
+
+<TabItem label="Python" icon="seti:python">
+
+```python
+from workers import DurableObject, WorkerEntrypoint
+
+class Default(WorkerEntrypoint):
+	async def fetch(self, request):
```
ryanking13: How about changing tab to spaces to make it consistent with other example PRs you added? It seems like all the examples in this PR uses tab for some reason.

src/content/docs/durable-objects/api/alarms.mdx
```diff
@@ -115,8 +119,47 @@ export class AlarmExample extends DurableObject {
...
+class AlarmExample(DurableObject):
+	def __init__(self, ctx, env):
+		super().__init__(ctx, env)
+		self.storage = ctx.storage
+
+	async def fetch(self, request):
+		# If there is no alarm currently set, set one for 10 seconds from now
+		current_alarm = await self.storage.getAlarm()
+		if current_alarm is None:
+			self.storage.setAlarm(Date.now() + 10 * SECONDS)
```
ryanking13: import `Date` from js or use `time.time()` instead?

src/content/docs/durable-objects/api/container.mdx
```diff
@@ -173,6 +307,32 @@ class MyContainer extends DurableObject {
...
+
+		def on_container_exit():
+			print("Container exited")
+
+		# the "err" value can be customized by the destroy() method
+		async def on_container_error(err):
+			print(f"Container errored: {err}")
+
+		self.ctx.container.start()
+		self.ctx.container.monitor().then(on_container_exit).catch(on_container_error)
```
ryanking13: Maybe we need create_proxy here too?

src/content/docs/durable-objects/api/container.mdx
```diff
@@ -42,24 +44,63 @@ export class MyDurableObject extends DurableObject {
 ```
 </TypeScriptExample>
 
+<TabItem label="Python" icon="seti:python">
+
+```python
+from workers import DurableObject
+
+class MyDurableObject(DurableObject):
```
mikenomitch: Part of me just wants to remove the containers examples for now.

We mostly don't want people using Containers via DOs without the "Container" helper class. It's at least a pretty "advanced" thing to do. - I'm not sure this example leads people to a great experience.

That said... I guess if you want to us it, we can let them. Hrm... I think this points to us probably wanting a Python Container class that acts like the JS one but also that might be a paint to maintain.

mikenomitch (approved):
Approved to unblock, but see @ryanking13's comments.

I also left a Container comment, but I'm of two minds on that TBH.

src/content/docs/durable-objects/api/alarms.mdx
dom96: I think it's because the JS examples are using tabs too, but I'll change it (I'm a spaces fan myself)

src/content/docs/durable-objects/api/container.mdx
dom96: Alright, that's fine. I'll just remove them.

PR #26836, opened 2025-12-01
[RealtimeKit] Move peer-lifecycle page under Core SDK

swapnilmadavi:
### Summary

This PR:
- moves the Peer-Lifecycle page under the Core SDK docs
- also, fixes a heading typo in the SDK Selection page


swapnilmadavi:
Instead of moving the page we've decided to make the content top-level, conceptual that applies to both the SDKs.
CC: @ishita1805 @ravindra-cloudflare

PR #26840, opened 2025-12-01
[draft] adding build image policies for Workers Builds and Pages

yomna-shousha:
initial pass at adding build image policies for Workers Builds and Cloudflare Pages


src/content/docs/workers/ci-cd/builds/build-image.mdx
```diff
@@ -88,3 +88,22 @@ Workers Builds are run in the following environment:
...
+
+## Build Image Policy
+
+### Default Version Updates
+
+Default versions of languages and tools are updated regularly to their latest stable releases.
+
+- **Minor version updates**: Default versions are updated to the latest minor version without notice (e.g., Node.js 22.15.0 → 22.16.0). If you need a specific minor version, [override the default version](/workers/ci-cd/builds/build-image/#overriding-default-versions).
+
+- **Major version updates**: When a language runtime or tool approaches its end-of-life (EOL) date, e.g. Node.js, we update to the latest supported version. We will provide at least 3 months advance notice via
```
irvinebroque:
```suggestion
- **Major version updates**: When a language runtime or tool approaches its end-of-life (EOL) date, e.g. Node.js, we update the build image to the latest supported version. We will provide at least 3 months advance notice via
```

We update to the _latest_ supported version or to the _oldest_ supported version?

Ex — we are saying Node.js 24 or 26?

Screenshot 2025-12-01 at 4 57 18 PM

src/content/docs/workers/ci-cd/builds/build-image.mdx
```diff
@@ -88,3 +88,22 @@ Workers Builds are run in the following environment:
...
+  - [Cloudflare Changelog](https://developers.cloudflare.com/changelog/)
+  - Dashboard notification to affected projects upon announcement
```
irvinebroque: "affected" is ambiguous here - i could take that to mean

- cloudflare will notify me if my code relies on an API that will break with the transition
- cloudflare will notify me if they know I am relying on that dependency
- cloudflare will notify me if I am currently pinned to an older version

src/content/docs/workers/ci-cd/builds/build-image.mdx
yomna-shousha: clarified in this most recent commit -- we'd be moving to the next version not the latest one

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
```diff
@@ -0,0 +1,17 @@
...
+description: Workers Builds and Cloudflare Pages now have documented build image policies outlining deprecation timelines and version update procedures.
+products:
+  - workers
+date: 2025-12-01
+---
+
+We've published build image policies for [Workers Builds](/workers/ci-cd/builds/build-image/#build-image-policy) and [Cloudflare Pages](/pages/configuration/build-image/#build-image-policy) that establish:
+
+- **Minor version updates**: Default versions are updated to the latest available minor version without notice 
+- **Major version updates**: When language runtimes or tools reach end-of-life, we update to the next stable LTS version with at least 3 months advance notice via Changelog and dashboard notifications
```
Cherry: It would be great if this was also an email notification I think. It's not uncommon for folks to not touch projects for >3 months, and changelogs are just supplementary to user comms.

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
```diff
@@ -0,0 +1,17 @@
...
+title: Build image policies for Workers Builds and Cloudflare Pages
+description: Workers Builds and Cloudflare Pages now have documented build image policies outlining deprecation timelines and version update procedures.
+products:
+  - workers
+date: 2025-12-01
+---
+
+We've published build image policies for [Workers Builds](/workers/ci-cd/builds/build-image/#build-image-policy) and [Cloudflare Pages](/pages/configuration/build-image/#build-image-policy) that establish:
+
+- **Minor version updates**: Default versions are updated to the latest available minor version without notice
```
Cherry: How will this work for anything that is zero-versioned, like Hugo? The current version is `extended_0.147.7`, and I believe breaking changes are very common in minor versions for them at this stage.

Cherry: How will this work for projects that don't follow semver, like Bun? Bun is quite popular, but doesn't really follow semver - we saw breaking changes from 1.1 -> 1.2, 1.2 -> 1.3, etc.

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
```diff
@@ -0,0 +1,17 @@
...
+title: Build image policies for Workers Builds and Cloudflare Pages
+description: Workers Builds and Cloudflare Pages now have documented build image policies outlining deprecation timelines and version update procedures.
+products:
+  - workers
+date: 2025-12-11
+---
+
+We've published build image policies for [Workers Builds](/workers/ci-cd/builds/build-image/#build-image-policy) and [Cloudflare Pages](/pages/configuration/build-image/#build-image-policy) that establish:
+
+- **Minor version updates**: Default versions are updated to the latest available minor version without notice. For tools that don't follow semantic versioning (e.g., Bun, Hugo), minor updates that may contain breaking changes will receive 3 months notice via Changelog, dashboard notifications, and email
```
jahands: Should soften to `may be` here?

```suggestion
- **Minor version updates**: Default versions may be updated to the latest available minor version without notice. For tools that don't follow semantic versioning (e.g., Bun, Hugo), minor updates that may contain breaking changes will receive 3 months notice via Changelog, dashboard notifications, and email
```

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
```diff
@@ -0,0 +1,17 @@
...
+- **Major version updates**: When language runtimes or tools reach end-of-life, we update to the next stable LTS version with 3 months notice via Changelog, dashboard notifications, and email
```
jahands: Maybe this should say "before they reach EOL"?

We may update tools well before they reach EOL so that we aren't having to make many updates throughout the year

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
irvinebroque: Default versions of what?

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
irvinebroque: +1

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
```diff
@@ -0,0 +1,17 @@
...
+- **Build image version deprecation** (Pages only): Build image versions are supported for at least 2 years and receive 6 months notice before removal via Changelog, dashboard notifications, and email
```
irvinebroque: When a build image version is no longer supported, customers will be moved to the newer build image version

Would be good to say this, otherwise it is undefined what happens when a build image is "removed"

But if I get it that I'll just be moved onto the latest, I get that - in the base case, lot of things keep working, it's not like we are going to just turn off your builds, etc.

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+To prepare for updates and avoid unexpected build failures:
+- Monitor the [Cloudflare Changelog](https://developers.cloudflare.com/changelog/), the dashboard, and your email for announcements
+- [Override default versions](/workers/ci-cd/builds/build-image/#overriding-default-versions) if you need to maintain specific versions for your applications
```
irvinebroque: Version of what? (especially because directly above you are talking about _build image versions_ but also about versions of software packages that are provided as part of the build image

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
irvinebroque: One thing that would be good to clarify

We are saying that for _both_ Workers and Pages, we make minor bumps all the time, and we make major bumps with advance notice and before EOL.

So — for _both_ Workers and Pages — there is no reason from here forward to _deprecate_ a build image version. Because even for Pages — so long as we are keeping dependencies up-to-date, there's not a need to deprecate a build image. Deps just get updated — [`v3`](https://developers.cloudflare.com/pages/configuration/build-image/) will just continue to get updated dependencies moving forward

i.e. — _before_ our approach to this was to have distinct build images

...and _now_ our approach to this is just — update dependencies based on policy described above — which obviates the need for distinct build images

The only case where the build image version deprecation is relevant is for deprecating the _existing_ pages v1 and v2 build images. Once that happens — both Pages and Workers have the same policy

accurate? if so there is a bit of dot connecting maybe to do here

irvinebroque: i.e. — if the above is accurate, then is this actually just simple as saying

If you are on the Pages v1 build image, you will be moved automatcially to v3 on <date in advance>
If you are on the Pages v2 build image, you will be moved automatcially to v3 on <date in advance>
If you are on the Pages v3 build image — dependencies will be updated based on the policy described above

src/content/changelog/workers/2025-12-01-build-image-policies-dev-plat.mdx
irvinebroque: i.e. Github says "preinstalled software" 

https://docs.github.com/en/actions/concepts/runners/github-hosted-runners#preinstalled-software-for-github-owned-images

which I immediately get what that means

src/content/docs/pages/configuration/build-image.mdx
```diff
@@ -105,3 +105,36 @@ The following features are not currently supported when using the v3 build syste
...
 <PagesBuildEnvironment />
+
+
+## Build Image Policy
+
+### Build Image Version Migration
+
+If you are currently using the v1 or v2 build image, your project will be automatically moved to v3:
+
+- **v1 build image**: if you are on the Pages v1 build image, you will be moved automatcially to v3 on [date]
```
irvinebroque: what date? :)

irvinebroque (approved):
See comment re: date placeholders but otherwise 👍

PR #26865, opened 2025-12-02
semgrep github workflow updated to show warnings

cdrubin:
### Summary

There is currently a semgrep workflow running that check against the rules defined in `.semgrep` directory of this repo. This PR aims to add warning messages as annotations to the workflow output inside the PR.

All changes are to the semgrep workflow in `.github` and use **_jq_** to reformat the `--json` based output of semgrep into the format required by github for surfacing annotations.


cdrubin:
Is it perhaps worth also checking for [semgrep skip] as well as [skip semgrep] since I expect this could be a easily swapped?

kodster28 (approved):

.github/workflows/semgrep.yml
```diff
@@ -35,10 +44,12 @@ jobs:
       # scans using rules inside the .semgrep/ folder and fails on error
       # include [skip semgrep] in top-most commit message to skip scan
       - name: Semgrep Repo Rules (Custom rules found in .semgrep/)
```
kodster28: Can you make this name more descriptive of what a contributor should know?

Maybe something like `Style guide check` or something like that? (naming is hard, so I'll leave it there)

src/content/docs/hyperdrive/get-started.mdx
```diff
@@ -21,7 +21,7 @@ This guide will instruct you through:
 
 :::note
 
-Hyperdrive currently works with PostgreSQL, MySQL and many compatible databases. This includes CockroachDB and Materialize (which are PostgreSQL-compatible), and PlanetScale.
+Hyperdrive 2026 currently works with PostgreSQL, MySQL and many compatible databases. This includes CockroachDB and Materialize (which are PostgreSQL-compatible), and PlanetScale.
```
kodster28: One edge case here now that this is staring me in the face, is there any way to exclude content contained within triple code fences?
(maybe `backticks` too, but don't think that's quite as necessary)

And/or test with something like a WranglerConfig file that's setting a compatibility date?

https://developers.cloudflare.com/style-guide/components/wrangler-config/
image

cdrubin: Great suggestions! Because semgrep does not support mdx files for more context-aware parsing all of our style rules must use the generic parser which may be able to wrangled a little more to give us this.

colbywhite: A separate lint from something like [markdownlint](https://github.com/markdownlint/markdownlint) could be used. Likely a separate ticket.

tools/semgrep-repo-rules
```diff
@@ -17,8 +18,9 @@ if [ -s tools/relevant_changed_files.txt ]; then
     semgrep scan \
       --config .semgrep --metrics=off \
       --include "*.mdx" --include "*.mdx" \
-      --force-color \
+      --error \
       $list_of_files
+  echo "return code: $?"
```
colbywhite: It looks like nothing calls this script, correct? Looks like the GitHub Action runs the `semgrep` commands directly. 

If something does call this, this echo would effectively swallow the exit code which might be needed by the caller.

colbywhite (approved):
Instead of making the commit message trigger an ignore, it might make sense to follow the linter practice of using an ignore (i.e. `// @ts-ignore rule-name`). The commit message would get it ignored for one and only run. (And also potentially clutter up the git history but I think it's already pretty cluttered.) A comment will allow it to be ignored in subsequent runs.

But semgrep might not support that easily so it might be a whole can of worms that we ignore for now.

tools/semgrep-repo-rules
cdrubin: Good point. It is meant to be a locally run approximation of the CI run, I might as well let the exit code leave intact!

PR #26979, opened 2025-12-07
Update Workflows overview page with use cases, revised code example and features

jonesphillip:
Update Workflows overview page with use cases, revised code example and features, animated workflow diagram.

Screenshot:
workflows-overview-docs

### DIFF
--- a/src/components/AnimatedWorkflowDiagram.astro
+++ b/src/components/AnimatedWorkflowDiagram.astro
@@ -1,0 +1,357 @@
+---
+interface Props {
+	steps: {
+		id: string;
+		label: string;
+		description?: string;
+		duration: string;
+	}[];
+	autoPlay?: boolean;
+	loop?: boolean;
+}
+
+const { steps, autoPlay = true, loop = true } = Astro.props;
+---
+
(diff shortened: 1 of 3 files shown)


mia303 (approved):

kodster28 (changes requested):
This is neat, but I'm a bit confused about the underlying purpose.

As constructed, don't see how this animated diagram conveys significantly more value than a static diagram or [mermaid diagram](https://developers.cloudflare.com/style-guide/documentation-content-strategy/component-attributes/diagrams/#mermaid-diagrams).

For example -- not a designer, but think it illustrates what I'm talking about

Image

☝️ would be significantly easier to maintain , as well as better suited to AI / LLM consumption.

Maybe there's some context here I'm not entirely understanding... but without a clear value proposition I don't think this is a direction we should go right now.

kodster28 (approved):
Chatted through this a bit more internally and sounds like there's intrinsic value in the animation above and beyond what a static diagram is bringing.

Fine trying this out as an experiment... though logging an internal ticket to see if there are better patterns we can set up for animation-y interactions in the future.

PR #27076, opened 2025-12-11
PCX-20163 Onoma: document expected behaviour for partial wildcard matching in subdomains

nikitacano:
### Summary

Wildcards are supported in private hostname configurations, but they must represent a full DNS label.


ranbel:
updating this in #27097 (to avoid merge conflicts)

PR #27105, opened 2025-12-13
Theme Updates

vaibhavshn:
### Summary

Initial PR to get feedback and iterate on more in the future.

- Making the default starlight theme a bit more cleaner.
- Cleaned the cards UI on homepage which are also used elsewhere
- Moved the feedback and contribute buttons to the bottom and deemphasized them a bit
  image

Tried to customize the default starlight theme by only applying styles wherever necessary so it doesn't break other styles, but if you still find it breaking anywhere lmk!

"Cleaner" design might be subjective, so looking forward to your feedback!

Also some questions:

- Is the black background too dark? Needs some more contrast?

### More to do

There's more to do here I think which I can pick up once I have some initial feedback:

- Update orange shade in dark mode to be a bit more lighter
- Some dropdowns in docs pages need to be updated too
- Syntax highlighting colors to be improved to go better with new theme

### Screenshots (optional)

| A docs page | Home page |
|--------|--------|
| image | image |

### DIFF
--- a/src/components/Description.astro
+++ b/src/components/Description.astro
@@ -1,3 +1,3 @@
-<div class="mb-4 text-xl leading-6" style="color: var(--sl-color-white);">
+<div class="mb-4 text-xl leading-7" style="color: var(--sl-color-white);">
 	<slot />
 </div>
--- a/src/components/CopyPageButton.tsx
+++ b/src/components/CopyPageButton.tsx
@@ -18,8 +18,6 @@ import {
 	PiChatCircleLight,
 	PiLinkLight,
 	PiPlugsConnectedLight,
-	PiTagChevron,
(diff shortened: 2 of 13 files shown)


src/components/Description.astro
```diff
@@ -1,3 +1,3 @@
-<div class="mb-4 text-xl leading-6" style="color: var(--sl-color-white);">
+<div class="mb-4 text-xl leading-7" style="color: var(--sl-color-white);">
```
vaibhavshn, line 1: Description looked a bit crammed, so relaxed it a bit

This is how it looked before. You can see the after one in the description.

image

kodster28:
We have a new hire coming onboard soon who can help evaluate this.

PR #27129, opened 2025-12-15
workers: streaming JSON example

elithrar:
This PR adds a streaming JSON example to the Workers docs and makes the Streams API docs a little clearer.


elithrar:
cc @irvinebroque since this comes up all the time.

irvinebroque (approved):

PR #27153, opened 2025-12-16
feat: add promise pipelining explanation to rules of durable objects doc

deloreyj:
### Summary

Adds an additional Durable Object best practice -- a get method that fetches a stub and pipelines any required initialization RPC. Also adds a caveat to the `Always await RPC calls` for promise pipelining

### Screenshots (optional)

image

### DIFF
--- a/src/content/docs/durable-objects/best-practices/rules-of-durable-objects.mdx
+++ b/src/content/docs/durable-objects/best-practices/rules-of-durable-objects.mdx
@@ -970,9 +970,66 @@ export default {
 ```
 </TypeScriptExample>
 
+### Use promise pipelining to guarantee initialization
+
+You can use [promise pipelining](/workers/runtime-apis/rpc/#promise-pipelining) to make multiple calls to a Durable Object in a single round trip by omitting `await` on intermediate RPC calls. This is useful for ensuring initialization runs before subsequent methods without the latency cost of separate requests. A common pattern is to create a helper function that retrieves a stub and pipelines an `init()` call. The Durable Object processes the RPC calls in order, guaranteeing initialization completes first.
+
+<TypeScriptExample filename="index.ts">
+```ts
+import { DurableObject } from "cloudflare:workers";
+
+export interface Env {
+	CHAT_ROOM: DurableObjectNamespace<ChatRoom>;
+}
+
+export class ChatRoom extends DurableObject<Env> {
+	private roomId: string | null = null;
+
+    // Static helper that pipelines initialization
+    static getChatRoom(env: Env, roomId: string) {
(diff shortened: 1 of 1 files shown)


src/content/docs/durable-objects/best-practices/rules-of-durable-objects.mdx
```diff
@@ -970,10 +970,70 @@ export default {
...
+	return stub;
+}
+
+export default {
+	async fetch(request: Request, env: Env): Promise<Response> {
+		const roomId = "lobby";
+
+    	// init() is pipelined — only one round trip for both calls
+    	const room = getChatRoom(env, roomId);
+    	await room.sendMessage("user-123", "Hello!");
```
elithrar, line 1018: Do you want to call out that the await is necessary here?

Important otherwise a user just has dangling promises.

src/content/docs/durable-objects/best-practices/rules-of-durable-objects.mdx
```diff
@@ -970,10 +970,70 @@ export default {
 ```
 </TypeScriptExample>
 
+### Use promise pipelining to guarantee initialization
+
+A common pattern is to create a helper function that retrieves a Durable Object stub and pipelines an initialization call. This ensures the `init()` method runs before any subsequent RPC methods, without requiring an extra round trip.
```
elithrar: Reword this - start with the “why” - “You can make multiple calls to a Durable Object in a single round-trip by …”

src/content/docs/durable-objects/best-practices/rules-of-durable-objects.mdx
```diff
@@ -970,10 +970,70 @@ export default {
...
+    	this.roomId = roomId;
+    }
+
+    async sendMessage(userId: string, content: string) {
+    	// Chat room logic
+    }
+
+}
+
+// Helper function that pipelines initialization
```
elithrar: I would simplify this by adding it to the DO class. The standalone helper adds extra cognitive load to the example.

src/content/docs/durable-objects/best-practices/rules-of-durable-objects.mdx
```diff
@@ -970,10 +970,70 @@ export default {
...
+
+    	return new Response("Message sent");
+    },
+
+};
+
+````
+</TypeScriptExample>
+
+By omitting `await` on the `init()` call, both `init()` and `sendMessage()` are sent together. The Durable Object processes them in order, guaranteeing initialization completes first. This helper can be a standalone utility function or a static method on the Durable Object class.
```
elithrar: re: the helper - we say it here but don’t have to show it

PR #27234, opened 2025-12-18
Add changelog for Workers for Platforms dashboard improvements

dinasaur404:
## Summary
- Adds changelog entry for new Workers for Platforms dashboard features including namespace creation, dispatch worker templates, and tag management


MattieTK (approved):

korinne (approved):
Non-blocking, but would suggest adding an introduction about what Workers for Platforms is, what's possible with it -> lead into this new change.

PR #27250, opened 2025-12-19
[AI GATEWAY] Generate code examples dynamically

palashgo:
### Summary

AI Gateway can be configured in a bunch of ways:

- With different providers (OpenAI / Anthropic / etc.)
- Different client SDKs (OpenAI JS, Vercel AI SDK, etc.)
- How upstream vendor auth is handled (Unified Billing, key passed in, key stored in Cloudflare)
- Whether you're on Workers or not (use binding or not)
- Whether Gateway Auth is on or off

Right now it's pretty hard for folks to take the small sample snippets we have and adapt them to their exact combo. This PR tries a madlib style, natural language form that generates the right code snippet based on what you pick.

(Since `Code` component is server-rendered, the generator is put together with a bit of a hack to wire everything up.)

### DIFF
--- a/src/components/ai-gateway/code-example-selector.tsx
+++ b/src/components/ai-gateway/code-example-selector.tsx
@@ -1,0 +1,504 @@
+import { useEffect, useState } from "react";
+import ReactSelect from "react-select";
+
+const STORAGE_KEY = "ai-gateway-code-selector";
+const AIG_EVENT = "ai-gateway-selector-change";
+
+export type Provider =
+	| "openai"
+	| "anthropic"
+	| "google"
+	| "grok"
+	| "dynamic"
+	| "workers-ai";
+export type KeyType = "byok" | "in-request" | "unified";
+export type ClientType = "openai-js" | "curl" | "aisdk";
+export type APIType = "native" | "unified";
+
+export interface Config {
+	provider: Provider;
+	keyType: KeyType;
+	clientType: ClientType;
+	apiType: APIType;
+}
+
+/**
+ * Shared hook to read and update the currently selected platform/framework.
+ *
+ * - Persists selection in localStorage under STORAGE_KEY.
+ * - Broadcasts changes via the `ai-gateway-selector-change` custom event.
+ * - Listens to the same event to stay in sync across multiple components.
+ */
+export function useAIGConfig() {
+	const [config, setConfig] = useState<Config>({
+		provider: "openai",
+		keyType: "byok",
+		clientType: "openai-js",
+		apiType: "unified",
+	});
+
+	// Helper: broadcast selection changes so other listeners can sync.
+	function notifySelectionChange(_config: Partial<Config>) {
+		if (typeof window === "undefined") return;
+		try {
+			window.dispatchEvent(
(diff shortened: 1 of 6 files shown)


kodster28 (changes requested):
Few things here:
- we want to keep `.tsx` and `.astro` files in `src/components/`
- Also, testing it out on a few of the samples and it doesn't seem to work correctly.
- Generally -- and this is part of our ongoing code orange work -- we want to avoid react in the components that are used in our docs. Assuming that these can be built out more natively in either Astro-native language or in more vanilla JS.

palashgo:
> Few things here:
> - we want to keep `.tsx` and `.astro` files in `src/components/`
> - Also, testing it out on a few of the samples and it doesn't seem to work correctly.
> - Generally -- and this is part of our ongoing code orange work -- we want to avoid react in the components that are used in our docs. Assuming that these can be built out more natively in either Astro-native language or in more vanilla JS.

I can change all the code to vanilla js and move it in src/components but can you tell me what's not working, these are the two examples I changed and every combination seemed to be working

https://palash-aig-examples-exp.preview.developers.cloudflare.com/ai-gateway/get-started/#example

https://palash-aig-examples-exp.preview.developers.cloudflare.com/ai-gateway/integrations/vercel-ai-sdk/#examples

kodster28:
> https://palash-aig-examples-exp.preview.developers.cloudflare.com/ai-gateway/usage/chat-completion/

Unified API > Curl is just blank

image

kodster28 (commented):
Thanks for fixing that specific bug!

Have two flavors of comment here.

## Logistical

These components are still in the wrong location. Should be moved to `src/components/ai-gateway/`.

## Technical

This fits into really common pattern I see from folks who are used to developing within a dashboard / application... namely that I don't think it's paying enough attention to the **robots** interacting with our content.

In a frontend web property, you have two broad consumers: **people** and **bots**.

I think this work is a net positive for people, which is great! Think the experiential part of this does help people more accurately self-select and find relevant code examples.

It's a negative for bots though, for two reasons:
1. The source of truth doesn't exist in HTML / MD. When you have something that's purely dynamic -- and never rendered statically -- LLM crawlers can't ever get the source of truth from the HTML / MD content.
2. Anything dynamic can affect page load time. Any bot / crawler gets affected by page load times... and Google specifically penalizes you if that page load time gets longer. I don't think this is directly applicable to this component... but it's a broader consideration for anything "dynamic".

With that in mind (especially #1), I'd recommend a different approach here. I'd guess that you could do something similar re: the filtering experience, but:
1) Move the code blocks to use the [Dynamic Render](https://developers.cloudflare.com/style-guide/components/render/#properties-to-render-content-conditionally) pattern that we already support. You need those files fully rendered at build time to make it consumable for bots.
2) Change the behavior of the filtering to use CSS to show / hide individual code blocks as appropriate. This means that those rendered files will then be visible to bots (which is often missed with pure JS implementations).

Let me know if there are follow-up questions or clarifications.

kodster28 (approved):
this still needs to be moved into `src/components` for the build to pass, btw.

palashgo:
@kodster28 addressed all your comments, thank you for your help

PR #27353, opened 2025-12-29
Add missing redirects for Workers for Platforms platform/ paths

dinasaur404:
## Summary

- Adds redirects for old `/platform/` paths that are now under `/reference/`:
  - `/platform/pricing/` → `/reference/pricing/`
  - `/platform/limits/` → `/reference/limits/`
  - `/platform/worker-isolation/` → `/reference/worker-isolation/`

These pages are showing up in SEO results but returning 404s.


irvinebroque (approved):
@kodster28 FYI I don't think I realized that we didn't fail CI on missing redirects anymore — but instead post a comment. Could see why it makes sense to do, just maybe worth a post in chat? (I could have missed this too)

kristianfreeman (approved):

PR #27396, opened 2026-01-04
Changelog for automated Queue provisioning

irvinebroque:
### Summary

refs https://github.com/cloudflare/workers-sdk/pull/11788

Don't merge until ^ lands and is released.

- [ ] TODO: Spell out the version of Wrangler that you must use in order to use this feature

### DIFF
--- a/src/content/changelog/queues/src2026-01-03-automatic-queue-provisioning.mdx
+++ b/src/content/changelog/queues/src2026-01-03-automatic-queue-provisioning.mdx
@@ -1,0 +1,40 @@
+---
+title: Automatic queue provisioning
+description: Wrangler now creates queues automatically when you deploy Workers with queue bindings.
+products:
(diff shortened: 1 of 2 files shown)


MattieTK:
Thanks @irvinebroque , can you also update the docs page to include queues in the list as a resource that we support for auto provisioning.

Also worth nothing that we count this as a beta feature currently as we are evaluating whether we want to make it more interactive for users outside of a CI context. For queues that interactivity might be even more useful as we could also offer to create a dead letter queue for discovery reasons(not sure if your PR does this yet, I will take a look) even if the user hasn't specified one.

Thank you for the contribution in any case!

irvinebroque:
https://github.com/cloudflare/cloudflare-docs/pull/27396/commits/9ec7074981d0d00109e546126fab4a1d717debcc

PR #27496, opened 2026-01-08
Enhance dynamic routing instructions with Custom Metadata

Vortexmind:
Added a note about using Custom Metadata in conditionals - as it is not immediately obvious


roerohan (approved):
LGTM

PR #27504, opened 2026-01-08
refactor: replace product labeling action with standard action

colbywhite:
Replaces the custom labeling logic with the off-the-shelve `actions/labeler`. This moves the logic from obtuse code to explicit configuration.

An example of this running can be found on #27449's resulting [job](https://github.com/cloudflare/cloudflare-docs/actions/runs/20763301252/job/59623450308).


.github/labeler.yml
```diff
@@ -0,0 +1,384 @@
...
+          - src/content/docs/workers-ai/**
+product:workers-vpc:
+  - changed-files:
+      - any-glob-to-any-file:
+          - src/content/docs/workers-vpc/**
+product:workflows:
+  - changed-files:
+      - any-glob-to-any-file:
+          - src/content/docs/workflows/**
+product:zaraz:
```
kodster28, line 381: Is there any way to auto-gen this? We add new folders pretty consistenly.

kodster28 (approved):

.github/labeler.yml
colbywhite: Add new folders or add new _products_? Right now there's a one-to-one relationship there. As I was doing this I was curious if that was supposed to be one-to-one or if that was just a convienent way to do it in code. Any more context on how we're using these labels? That might give context for what to do here.

kodster28: New products, which *almost* always correspond to new folders.
(and I think we'd be fine with 1x1)

Generally, for two purposes:
1. helping see what I (as a writer) might need to pay attention to.
2. reporting / trendlines over time. Which products get a lot of updates vs fewer.

colbywhite: Let's chat about this at some point. I think those purposes are achieved without a label (and may or may not be better ways to do them). I'll merge it but keep the ticket open until we discuss.

PR #27568, opened 2026-01-12
Update Explicit Resource Management documentation link

dimitropoulos:
I noticed this language is now out of date.  I updated the link to Explicit Resource Management in the documentation to reflect its inclusion in V8 and Node.js 24.0.0 and above.


MattieTK:
/bonk can you verify this with sourcing?

PR #27599, opened 2026-01-13
DRAFT: RFC for DO tracing spans

vy-ton:
Draft PR captures proposal for all DO related spans and attributes we should have for public traces.

- Assume tracing context propagation exists
- Ideally docs could be generated from code for public spans/attributes

### DIFF
--- a/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
+++ b/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
@@ -442,12 +442,24 @@ All RPC requests within a [RPC session](/workers/runtime-apis/rpc/lifecycle/) us
 
 #### `durable_object_subrequest`
 
-- `cloudflare.durable_object.queue_time_us`
+- `cloudflare.durable_object.queue_time_ms`
 - `cloudflare.durable_object.startup_time_ms`
 - `cloudflare.durable_object.constructor_invoked`
 - `cloudflare.durable_object.constructor_time_ms`
 - `cloudflare.durable_object.cpu_time_ms`
 - `cloudflare.durable_object.wall_time_ms`
+// Added as attributes for now while we discusse span vs attribute
+- `cloudflare.durable_object.input_gate_lock_hold_ms` is the time holding the gate
+- `cloudflare.durable_object.input_gate_lock_wait_ms` is the time waiting for gate to be available
+- `cloudflare.durable_object.output_gate_lock_hold_ms`
+- `cloudflare.durable_object.output_gate_lock_wait_ms`
+
+#### For handlers[/workers/observability/traces/spans-and-attributes/#handlers] invoked on a Durable Object such as RPC or fetch(), these attributes exist:
+
+- `cloudflare.durable_object.objectId`
+- `cloudflare.durable_object.objectName`
+
+#### `durable_object_blockConcurrencyWhile`
 
 ---
 
@@ -459,6 +471,7 @@ The SQLite Storage API allows you to modify the SQLite database embedded within 
 
 - `cloudflare.durable_object.response.rows_read`
 - `cloudflare.durable_object.response.rows_written`
+- `cloudflare.durable_object.response.bytes_read`
 - `cloudflare.durable_object.response.bytes_written`
 - `cloudflare.durable_object.response.sql_duration_ms`
 
--- a/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
+++ b/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
@@ -440,14 +440,12 @@ All RPC requests within a [RPC session](/workers/runtime-apis/rpc/lifecycle/) us
 
 ### [Durable Object API](/durable-objects/)
 
-When a Durable Object is invoked, the Worker invocation span will include:
-
-- `cloudflare.durable_object.queue_time_us`
-
 #### `durable_object_subrequest`
 
-- `cloudflare.durable_object.startup_duration_ms`
+- `cloudflare.durable_object.queue_time_us`
+- `cloudflare.durable_object.startup_time_ms`
 - `cloudflare.durable_object.constructor_invoked`
+- `cloudflare.durable_object.constructor_time_ms`
 - `cloudflare.durable_object.cpu_time_ms`
 - `cloudflare.durable_object.wall_time_ms`
 
--- a/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
+++ b/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
@@ -438,7 +438,16 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
 
 ### [Durable Object API](/durable-objects/)
 
+When a Durable Object is invoked, the Worker invocation span will include:
+
+- `cloudflare.durable_object.queue_time_us`
+
 #### `durable_object_subrequest`
+
+- `cloudflare.durable_object.startup_duration_ms`
+- `cloudflare.durable_object.constructor_invoked`
+- `cloudflare.durable_object.cpu_time_ms`
+- `cloudflare.durable_object.wall_time_ms`
 
 ---
 
@@ -446,14 +455,19 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
 
 The SQL API allow you to modify the SQLite database embedded within a Durable Object.
 
+#### Attributes available on all Durable Object Storage SQL API spans
+
+- `cloudflare.durable_object.response.rows_read`
+- `cloudflare.durable_object.response.rows_written`
+- `cloudflare.durable_object.response.bytes_written`
+- `cloudflare.durable_object.response.sql_duration_ms`
+
 #### [`durable_object_storage_exec`](/durable-objects/api/sqlite-storage-api/#exec)
 
 - `db.system.name`
 - `db.operation.name`
 - `db.query.text`
 - `cloudflare.durable_object.query.bindings`
-- `cloudflare.durable_object.response.rows_read`
-- `cloudflare.durable_object.response.rows_written`
 
 #### [`durable_object_storage_getDatabaseSize`](/durable-objects/api/sqlite-storage-api/#databasesize)
 
--- a/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
+++ b/src/content/docs/workers/observability/traces/spans-and-attributes.mdx
@@ -451,9 +451,9 @@ All RPC requests within a [RPC session](/workers/runtime-apis/rpc/lifecycle/) us
 
 ---
 
-### [Durable Object Storage SQL API](/durable-objects/api/sqlite-storage-api)
+### [Durable Object SQLite Storage API](/durable-objects/api/sqlite-storage-api)
 
-The SQL API allow you to modify the SQLite database embedded within a Durable Object.
+The SQLite Storage API allows you to modify the SQLite database embedded within each Durable Object via SQL or key-value methods.
 
 #### Attributes available on all Durable Object Storage SQL API spans
 
@@ -501,9 +501,9 @@ The SQL API allow you to modify the SQLite database embedded within a Durable Ob
 
 ---
 
-### [Durable Object Storage KV API](/durable-objects/api/legacy-kv-storage-api)
+### [Durable Object KV Storage API](/durable-objects/api/legacy-kv-storage-api)
 
-The legacy KV-backed API allows you to modify embedded storage within a Durable Object.
+The legacy KV-backed Storage API allows you to modify embedded storage within a Durable Object.
 
 #### [`durable_object_storage_get`](/durable-objects/api/legacy-kv-storage-api/#do-kv-async-get)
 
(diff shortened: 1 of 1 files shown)


src/content/docs/workers/observability/traces/spans-and-attributes.mdx
```diff
@@ -438,22 +440,36 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
 
 ### [Durable Object API](/durable-objects/)
 
+When a Durable Object is invoked, the Worker invocation span will include:
+
+- `cloudflare.durable_object.queue_time_us`
```
joshthoward: I think this needs to be part of durable_object_subrequest since each subrequest has its own queue time.

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
```diff
@@ -438,22 +440,36 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
...
+
 #### `durable_object_subrequest`
 
+- `cloudflare.durable_object.startup_duration_ms`
```
joshthoward: Nit: Would call this startup_time_ms for consistency

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
```diff
@@ -438,22 +440,36 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
...
 ### [Durable Object Storage SQL API](/durable-objects/api/sqlite-storage-api)
 
 The SQL API allow you to modify the SQLite database embedded within a Durable Object.
 
+#### Attributes available on all Durable Object Storage SQL API spans
+
+- `cloudflare.durable_object.response.rows_read`
+- `cloudflare.durable_object.response.rows_written`
+- `cloudflare.durable_object.response.bytes_written`
+- `cloudflare.durable_object.response.sql_duration_ms`
```
joshthoward, line 476: We need to understand whether this will always be 0 or not because it is a synchronous operation. If it's always 0, then I do not think we should include it.

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
```diff
@@ -438,22 +440,36 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
...
 ### [Durable Object API](/durable-objects/)
 
+When a Durable Object is invoked, the Worker invocation span will include:
+
+- `cloudflare.durable_object.queue_time_us`
+
 #### `durable_object_subrequest`
 
+- `cloudflare.durable_object.startup_duration_ms`
+- `cloudflare.durable_object.constructor_invoked`
```
joshthoward, line 447: Maybe it's also worth adding constructor_time_ms? This would depend on the time resolution that we can get for potentially synchronous operations. Some constructors might make outbound requests and this would end up being very useful.

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
```diff
@@ -438,22 +440,36 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
 
 ### [Durable Object API](/durable-objects/)
```
vy-ton, line 441: @vy-ton to add `blockConcurrencyWhile` span and TODO:`output_gate_lock_wait`

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
vy-ton: Do you want both? or does `constructor_tims_ms=0` indicate the constructor did not run

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
vy-ton: cc @jmorrell-cloudflare

shrima-cf:
How should I read this PR? Is the goal just to list all the spans available? Will there be documentation explaining what the span means and expected duration for it? (similar to what we have in https://gitlab.cfdata.org/cloudflare/ew/edgeworker/-/blob/master/src/edgeworker/scheduling/jaeger-spans.c%2B%2B ?

vy-ton:
> How should I read this PR? Is the goal just to list all the spans available? Will there be documentation explaining what the span means and expected duration for it? (similar to what we have in https://gitlab.cfdata.org/cloudflare/ew/edgeworker/-/blob/master/src/edgeworker/scheduling/jaeger-spans.c%2B%2B ?

@shrima-cf We will certainly add explanation as part of releasing. I'd love to have all spans/attributes generated from code somehow to avoid syncing issues.

Right now, read this PR as capturing all the DO related spans and attributes we want to add

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
vy-ton: for all DO spans:
- objectId

justin-mp:
I find it odd that we're adding timing attributes to spans because spans themselves are supposed to represent the time it took something to run.

In particular, I think things that are in the programming model that the programmer has control over should be spans.  In particular,  `constructor_time_ms`, `sql_duration_ms`, and `output_gate_lock_held` should really be spans (which would then have different names).  Generally when you're doing performance engineering, you need to go down to the primitives, and these are the primitives we give users.  I could argue that queue time and the like could also be spans.  Then a user knows it's blocked because of queuing, which might be due to load.

Additionally, things like CPU time are OK as an attribute because that's an orthogonal dimension than you get out of a span, but having wall time as an attribute makes no sense as that's exactly what the span's duration measures.

jmorrell-cloudflare:
> I find it odd that we're adding timing attributes to spans because spans themselves are supposed to represent the time it took something to run.
> having wall time as an attribute makes no sense as that's exactly what the span's duration measures.

@justin-mp I pretty strongly disagree with those assertions. [Creating many spans for every possible timing is one of the most common tracing anti-patterns IMO](https://jeremymorrell.dev/blog/a-practitioners-guide-to-wide-events/#versions-of-important-things#timings:~:text=Wrapping%20absolutely%20everything%20in%20its%20own%20span%20is%20the%20most%20common%20failure%20mode%20I%20see%20when%20engineers%20first%20get%20access%20to%20tracing%20tools.%20You%20have%20to%20design%20the%20structure%20of%20your%20data%20for%20the%20way%20you%20want%20to%20query%20it.). It makes querying across many operations far harder than it needs to be, and complicates the waterfall visualization unnecessarily. You need to design your data for how you want to query it.

> Generally when you're doing performance engineering, you need to go down to the primitives, and these are the primitives we give users.

Attributes are also a core part of the span model, and there is no reason you can't put timing information there. You should think of spans as capturing all of the information about a specific operation. For performance engineering profiling is usually the better tool.

I address this attributes-vs-child-spans tradeoff in my guest chapter in [Observability Engineering](https://www.oreilly.com/library/view/observability-engineering/9781492076438/):

Let's say in your system you've just shipped a new subsystem that prioritizes payload parsing for enterprise users, and you want to see the impact of that change on tail latencies across all of the regions where you have systems deployed. If all of those attributes are present on the wide event, then this is straightforward:

```
SELECT
  P99(payload_parse.duration_ms)
WHERE
  main = true AND
  service.name = "api-service"
GROUP BY
  user.type,
  cloud.region
  ```
  
However this may seem like we are duplicating data available in the child spans. Surely we can accomplish the same thing by wrapping the payload parsing method in its own span?
This is a valid approach, but now if we want to query that data alongside any other data that we've captured in our wide event, querying has become much more complicated, forcing the use of JOINs: 

```
SELECT
  P99(parse_span.duration_ms)
FROM spans AS main_span
JOIN spans AS parse_span ON main_span.trace_id = parse_span.trace_id
WHERE
  main_span.main = true AND
  main_span.service.name = "api-service" AND
  parse_span.name = "payload-parse"
GROUP BY main_span.user.type, main_span.cloud.region
```

Mature observability tooling is capable of running these types of queries at the cost of additional processing, however we should keep in mind how we will be using this data. We want to prioritize quick, iterative exploration, often while responding to active incidents, and we want any engineer on our team to be able to easily navigate our observability tooling.
Viewed through this lens, identifying a few important timings and adding them to the wide event is well worth the slight data duplication.

File: src/content/docs/workers/observability/traces/spans-and-attributes.mdx [hunk 6]
```diff
@@ -440,20 +442,32 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
...
+### [Durable Object SQLite Storage API](/durable-objects/api/sqlite-storage-api)
+
+The SQLite Storage API allows you to modify the SQLite database embedded within each Durable Object via SQL or key-value methods.
+
+#### Attributes available on all Durable Object Storage SQL API spans
 
-The SQL API allow you to modify the SQLite database embedded within a Durable Object.
+- `cloudflare.durable_object.response.rows_read`
+- `cloudflare.durable_object.response.rows_written`
+- `cloudflare.durable_object.response.bytes_written`
```
lambrospetrou, line 475: Are these added after the returned cursor is iterated, so after the operation function returns, or before the cursor is iterated hence it will be zero for rows_read?

Similarly, could we also have "bytes_read"?

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
shrima-cf: There are four spans: input_gate_lock_hold, input_gate_lock_wait, output_gate_lock_hold, output_gate_lock_wait - we might have to add all of these

vy-ton: @shrima-cf So I understand, `*_gate_lock_hold` is when we trigger the gate and `*_gate_lock_wait` spans until the gate is released, right?

shrima-cf: Its the other way round, *_gate_lock_wait is triggered when we wait for the lock to be available for us. *_gate_lock_hold is when we are holding the lock.

shrima-cf:
@vy-ton In addition to spans, there are also tags that hold useful information, Alex added a couple for the input/output gate spans - https://github.com/cloudflare/workerd/pull/5827
Are you planning on adding these to the documentation as well?

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
vy-ton: > Are these added after the returned cursor is iterated, so after the operation function returns, or before the cursor is iterated hence it will be zero for rows_read?

For usefulness, I would expect after the cursor is iterated. But we need to confirm.

Added `bytes_read`

vy-ton:
> @vy-ton In addition to spans, there are also tags that hold useful information, Alex added a couple for the input/output gate spans - [cloudflare/workerd#5827](https://github.com/cloudflare/workerd/pull/5827) Are you planning on adding these to the documentation as well?

I would not expose the span attributes in https://github.com/cloudflare/workerd/pull/5827 - those seem pretty internal-only to me.

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
```diff
@@ -440,20 +442,45 @@ Cloudflare Workers provides automatic tracing instrumentation **out of the box**
...
+- `cloudflare.durable_object.constructor_time_ms`
+- `cloudflare.durable_object.cpu_time_ms`
+- `cloudflare.durable_object.wall_time_ms`
+// Added as attributes for now while we discusse span vs attribute
+- `cloudflare.durable_object.input_gate_lock_hold_ms` is the time holding the gate
+- `cloudflare.durable_object.input_gate_lock_wait_ms` is the time waiting for gate to be available
+- `cloudflare.durable_object.output_gate_lock_hold_ms`
+- `cloudflare.durable_object.output_gate_lock_wait_ms`
+
+#### For handlers[/workers/observability/traces/spans-and-attributes/#handlers] invoked on a Durable Object such as RPC or fetch(), these attributes exist:
```
vy-ton, line 457: @lambrospetrou you mentioned having primary/replica context, what other attributes would you expect here?

src/content/docs/workers/observability/traces/spans-and-attributes.mdx
vy-ton: From Alex's comment [here](https://chat.google.com/room/AAAAVrF5vCc/pe34ly4s28E/S6Y6AnSTDZI?cls=10), we probably need an attribute to indicate whether the cursor is fully consumed.

danieltroger:
@TimoWilhelm Asked me (CF user) to take a look and give some feedback. Here's what I think

I think it looks good, added observability is great. The most useful information in this proposal would probably be

- `cloudflare.durable_object.queue_time_ms`
- `cloudflare.durable_object.cpu_time_ms`
- `cloudflare.durable_object.wall_time_ms`

As for if something is missing:
You can see here how most traces for our DO look like:

Screenshot 2026-02-03 at 09 53 34

(Not incredibly useful yet)



### What I think would be most helpful to add in terms of DO observability would be:
0. RPC payload size receive and send after a connection is closed (to see how close one is to the 32MiB RPC limit)
2. Stats from the DO when the request was made and ended (ideally a graph during the whole connection):
3. 1. Current memory usage (we're constantly fighting OOM errors, "Durable Object's isolate exceeded its memory limit and was reset")
4. 2. Concurrent open RPC connections to the DO (gives an indication of load, helps debug overloaded DO)
5. A way of actually getting to the relevant DO trace from the trace of the worker that called the DO Screenshot 2026-02-03 at 09 57 38, right now there's no way to follow the code flow here in the observability
6. 1. Also above one should see the function name invoked on the DO instead of just `durable_object_subrequest`

PR #24875, opened 2025-09-02
Minor Fixes for China Network Available Products

esheacf:
### Summary

During an Austin onsite session, SE team informed me that having Durable Objects on the dev docs was a bit misleading for our available products page. Additionally, it would be best to caveat that Turnstile is not available in Mainland China even though Bot Management is generally available in Mainland China.

### Documentation checklist

- [x] The [documentation style guide](https://developers.cloudflare.com/style-guide/) has been adhered to.

PR #27618, opened 2026-01-14
[China-Network] Update available-products.mdx

luishcardoso:
Hello team, 

### Summary

Workers in China Network now can talk to DOs in other colos.

Per https://wiki.cfdata.org/display/PM/Cloudflare+features+in+China+Network?focusedCommentId=1324662160#comment-1324662160

https://jira.cfdata.org/browse/STOR-1547

Can we publicly add this info here?

Thanks.


pedrosousa:
We previously had information about Durable Objects on the page you're updating, but it was later removed in #24875 because it was considered misleading info (there's just minimal support for DOs, not full support).

CC @esheacf

esheacf:
I currently do not have high confidence that durable objects works reliably as this is not something my team has personally tested yet. However, I imagine it could be more reliable with Global Acceleration. Do we have a customer case study where this has worked successfully? It'd be great if my team can be pulled into the conversation if this is the case.

If not, it may be better not to add Durable Object in the meantime until we have had a chance to test it stringently or at least have an internal discussion over it. We already had a similar incident where Pages appeared to be functional in theory, but customers were not able to use it reliably.

pedrosousa:
Not adding Durable Objects for now.

PR #27633, opened 2026-01-14
[Docs] Update dash routes (account level)

pedrosousa:
### Summary

Updates dashboard routes for account-level navigation.

### DIFF
--- a/src/content/dash-routes/core.json
+++ b/src/content/dash-routes/core.json
@@ -39,6 +39,10 @@
 		"parent": ["Analytics & logs"]
 	},
 	{
+		"name": "Domains",
+		"deeplink": "/?to=/:account/domains/overview"
+	},
+	{
 		"name": "Workers & Pages",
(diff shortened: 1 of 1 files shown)


src/content/dash-routes/core.json
```diff
@@ -255,7 +269,7 @@
 	},
 	{
 		"name": "Zero Trust",
-		"deeplink": "https://one.dash.cloudflare.com"
+		"deeplink": "/?to=/:account/one"
```
kodster28: This one technically doesn't work unless you directly navigate to it. I'll flag as a bug to that team

kodster28 (approved):

src/content/dash-routes/core.json
pedrosousa:
```suggestion
		"deeplink": "https://one.dash.cloudflare.com"
```

PR #27654, opened 2026-01-15
[Docs] Update prettier extension

pedrosousa:
### Summary

From the extension [README](https://github.com/prettier/prettier-vscode/blob/main/README.md):

> 🚨🚨🚨 **Extension Migration**: This extension is being moved from `esbenp.prettier-vscode` to [`prettier.prettier-vscode`](https://marketplace.visualstudio.com/items?itemName=prettier.prettier-vscode). Version 12+ is only published to the new for now as it is a major change. Once it is stable, we will publish v12 to both extensions and deprecate the `esbenp.prettier-vscode` extension. **Version 12.x is currently not stable, use with caution and report bugs.**

Waiting for 12.x to become stable before merging.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -6,7 +6,7 @@
 		"unifiedjs.vscode-mdx",
 		"bradlc.vscode-tailwindcss",
 		"redhat.vscode-yaml",
-		"esbenp.prettier-vscode",
+		"prettier.prettier-vscode",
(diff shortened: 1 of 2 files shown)


pedrosousa:
After [some discussion](https://github.com/prettier/prettier-vscode/issues/3872), the extension namespace (and publisher) will remain the same — `esbenp` — so we won't need to update our list of recommended extensions.

PR #27666, opened 2026-01-15
[Docs] Update dash routes (account level)

pedrosousa:
### Summary

Updates dashboard routes at the account level and fixes `DashButton` occurrences that were using routes that are no longer available. Also fixes some dash links (without a button) so that we can replace those links with a button at any time.

Removes all routes from `core-manually-defined.json`, since they're now defined in `core.json`.

### Documentation checklist


src/content/docs/cloudflare-one/traffic-policies/packet-filtering/create-rate-limiting-policies.mdx
```diff
@@ -22,28 +22,28 @@ For Advanced Magic Firewall customers, rate limiting (beta) is available by requ
 
 To add a policy:
 
-1. In the Cloudflare dashboard, go to the [Firewall Policies](https://dash.cloudflare.com/?to=/:account/network-security/magic_firewall/rate_limiting) page.
+1. In the Cloudflare dashboard, go to the [Firewall Policies](https://dash.cloudflare.com/?to=/:account/network-security/magic_firewall) page.
```
pedrosousa, line 25: Preparing to add back DashButtons (use a route that's available in `core.json`).

marciocloudflare: I think Magic Firewall is still in the Cloudflare One dashboard - unline Magic WAN stuff that was removed last minute. So I think these instructions should be on the CF1 dash, not the Core dash

marciocloudflare (approved):

PR #27692, opened 2026-01-16
fix(): ensure all files are written before shutdown

blyedev:
### Summary

This flag on tigrisfs ensures all files to object storage are written before shutdown.


blyedev:
As a follow up, turns out this didn't actually fix my issue, the write I have just ran successfully by coincidence. Turns out the proper fix is to run a flush via python that waits synchronously until all r2 writes complete before finishing. I added this to my entry point at the end. Question is if the readonly example should include it or if I should write a separate mention for flushing writes before container completion.

```
    echo "Flushing directory (fsync)..."\n\
    python3 -c "import os, sys; os.fsync(os.open(sys.argv[1], os.O_RDONLY))" /mnt/r2\n\
```

PR #27714, opened 2026-01-18
A first example of using product availability data from middlecache

cdrubin:
- Custom loader file added, middleacache remote-to-local retrieval loader added
 - Basic ProductAvailability component added
 - Single file content collection config and schema pattern placed in content/ directory but not as a directory since the content itself is downloaded
 - Pattern for single line content collection definitions


src/util/custom-loaders.ts
```diff
@@ -0,0 +1,47 @@
...
+				const response = await fetch(MIDDLECACHE_URL + pathOnMiddlecache);
+				const content = await response.text();
+
+				fs.writeFileSync(destination, content);
+				context.logger.debug(`Download of ${pathOnMiddlecache} completed.`);
+			}
+
+			const fileLoader = file(destination, options);
+			// re-use all the functionality of the built-in file loader
+			return await fileLoader.load(context);
```
colbywhite, line 57: What's the reasoning for this pattern? If you use an object loader, there's no need for a tmp directory. Astro does all of that for you via its data store. The Astro file loader is meant to be used for files checked into git, i.e. all the other files in `src/content`. 

The comment indicates you want to reuse the file loader's functionality, but I don't think it's meant to be re-used like this. The file loader is a wrapper on top of the API used by object loaders (i.e. LoaderContext). That API is the root primitive that is designed to be built on top of. 

Any benefit to this pattern instead of the object loader pattern?

src/util/custom-loaders.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as path from "path";
+
+/**
+ * middlecache loader expects a middlecache path, the name of content collection to be built with the
+ *
+ * @param path - data file path in the middlecache R2 bucket (required)
+ *
+ */
+
+function middlecacheLoader(pathOnMiddlecache: string, options = {}): Loader {
```
colbywhite: Let's type these options. As written, you want to type them the same as what gets passed into the file loader, or a subset of it. But I don't think building on top of the file loader is the best pattern. So more likely it looks something like this.

```ts
type MiddlecacheLoaderOptions = {
  path: string;
  url?: string
}

function middlecacheLoader({
   path, 
   url = "https://middlecache.ced.cloudflare.com/"} : MiddlecacheLoaderOptions
)
```

Allowing the URL to be provided could aid debugging.

src/util/custom-loaders.ts
```diff
@@ -0,0 +1,47 @@
...
+			}
+
+			const fileLoader = file(destination, options);
+			// re-use all the functionality of the built-in file loader
+			return await fileLoader.load(context);
+		},
+	};
+}
+
+export { middlecacheLoader as middlecache };
```
colbywhite: The original name of the function is more precise. This isn't a function that represents all of Middlecache, but a loader function that loads from Middlecache. 

Regardless, of the name I would just export the function directly.

```ts
export function middlecacheLoader(...)
```

src/content/collection.product-maturity.ts
colbywhite, line 1: The files might be easier to grok if the filename was `src/content/collection/product-maturity.ts`. This would put all the collection-related stuff in a single directory that's collapsible.

src/components/ProductMaturity.astro
```diff
@@ -0,0 +1,16 @@
...
+const props = z.object({
+	product: z.string(),
+});
+
+const input = props.parse(Astro.props);
+
+const entry = await getEntry("product-maturity", input.product);
+---
+
+{entry && entry.data}
```
colbywhite: There's no UI around it? Just the string? I assumed there was a badge or some other UI pattern we were using to display. Interesting.

I would push for our components to be named with their UI usage in mind. Since this one has very minimal UI around it, maybe `ProductMaturityString` or `ProductMaturityText`. I suppose there's likely some value to wrapping it in a `<span>`, which would make it `ProductMaturitySpan`. (When I incorrectly assumed these were going to be badges, I was envisioning `ProductMaturityBadge`.)

Either way, we might want to do some error checking here. Throwing an error would break the build, but that's likely what we want. We want a writer to know right away that they typo'd a product name that we don't have maturity for. 

Note: This isn't error handling around retrieving the data from Middlecache. This is an error with the usage of the component. i.e. `<ProductMaturityString product="qwerty" />`

src/content/docs/rules/trace-request/index.mdx
```diff
@@ -5,17 +5,19 @@ sidebar:
   order: 14
   badge:
     text: Beta
+    // TODO: set the badge text in the frontmatter with <ProductMaturity product="cloudflare-trace" /> as well
```
colbywhite: Oh. I didn't realize we were doing this.

It'll be tough to set the value itself "in the frontmatter". Instead we likely want this frontmatter to indicate the product for the badge.

```yml
sidebar:
    badge:
        product: cloudflare-trace
```

And then the component that renders the sidebar (I think it's `src/components/overrides/Sidebar.astro`) can pass that into the `ProductMaturity*` component.

Let's make that a separate ticket. The sidebar isn't a standard sidebar and I'm curious if we want to clean that up as part of this or keep kicking the can.

src/content/collection.product-maturity.ts
```diff
@@ -0,0 +1,23 @@
...
+		parser: (fileContent: string) => {
+			const data = JSON.parse(fileContent);
+			const lookup: Record<string, string | null> = {};
+
+			for (const item of data) {
+				lookup[item.name] = item.maturity;
+			}
+
+			return lookup;
+		},
```
colbywhite: Using the object loader will reduce this second JSON parsing that's done here.

src/util/custom-loaders.ts
```diff
@@ -0,0 +1,47 @@
...
+import { fileURLToPath } from "node:url";
+import fs from "fs";
+import * as path from "path";
+
+/**
+ * middlecache loader expects a middlecache path, the name of content collection to be built with the
+ *
+ * @param path - data file path in the middlecache R2 bucket (required)
+ *
+ */
```
colbywhite, line 15: It looks the comment just didn't get finished. Complete sentence is ideal. 

And an example for the path would also help. Or make it clear that it includes the `v1`.

src/components/ProductMaturityText.astro
```diff
@@ -0,0 +1,16 @@
+---
+import { getEntry } from "astro:content";
+
+import { z } from "astro/zod";
+
+type Props = z.infer<typeof props>;
+const props = z.object({
+	product: z.string(),
+});
```
colbywhite: The types auto-generated from the collection can make this more specific. I think it ends up being something along the lines of `CollectionEntry<"product-maturity">`. And thus you likely don't need zod at all here.

```ts
type Props = { product: CollectionEntry<"product-maturity"> }
```

Or something close to that.

colbywhite: And now that I think about it, if you type it to only allow for valid products, then maybe that's enough error handling to make sure writers know when they've passed in a bad product name. That of course assumes that the writers are actively reacting to the red squiggles in their editors, which I don't know is true.

src/content/collection.product-maturity.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from "astro/zod";
+
+import { middlecache } from "../util/custom-loaders";
+
+const schema = z.string().nullable();
+
+const collectionConfig = {
```
colbywhite: I'd just name if `productMaturityConfig` and export it as is so you don't have to rename it later.

```ts
export const productMaturityConfig = ...
```

colbywhite:
FYI: The format check fails for `custom-loaders.ts` file. A `npm run format:core:fix` can fix that. I'd also set your IDE up to run prettier frequently to catch that for you sooner.

src/util/custom-loaders.ts
cdrubin: The middlecache custom loader adds retrieval of remote files to the local filesystem. The Astro `file` loader is for local files (however they get there) so it makes sense to re-use that functionality. A good rule of thumb is to use lower level APIs when needed and are not required for this. One reason for using the .tmp directory and only downloading files that are not present in it already is because there are already use-cases for reusing the same file from middlecache for different content collections. Another great reason to download to the filesystem is for easier debugging and development.

src/util/custom-loaders.ts
cdrubin: This file may have other custom loaders in it. I wanted to "follow" the Astro convention of doing something like this: `import { file } from "astro/loaders";` (not having Loader in the name and potentially having more than one custom loader exported from that custom-loaders file.

src/content/collection.product-maturity.ts
cdrubin: I worked along those lines for a while but eventually decided on this convention mainly because the subdirectories of that `/content` directory all contain content, with the name of the directory being the kind of content. Using the filename like this marks this as a collection immediately - dynamic source of content - and the idea is that there will be a number of `collection.` files present for each of these content sources _not_ created from the static files in a dir. I'm open to a different prefix, collection seemed the most general but specific enough for this use.

src/components/ProductMaturity.astro
cdrubin: Great point about renaming the component. There may be more interesting components required later but the basic and first use case is just a text provider. I've added throwing a more useful Error. Is there a nice way to get Astro to return/show the page that was being generated to improve the error usefulness?

src/content/docs/rules/trace-request/index.mdx
cdrubin: I have the skeleton of a remark plugin in the works to cover this use-case. Seems like the correct way to insert additional frontmatter items and we have a plan for marking the pages on which that automatic badge addition should occur.

src/content/collection.product-maturity.ts
cdrubin: Using this simple filter is desirable - no need to store the whole object that is present in the original JSON. I could write the filter in fewer lines but I like how easy it is to read this, The small amount of custom code is a benefit of re-using the file loader.

src/components/ProductMaturity.astro
cdrubin: Found it: `${Astro.url.pathname}` and added!

src/components/ProductMaturityText.astro
cdrubin: Both are probably good. Thank you for the great suggestion. Indeed a content collection super power we should be leaning on!

src/content/collection.product-maturity.ts
cdrubin: Astro itself does not seem to follow this "specific name" convention. I hear you that it may be simpler in many cases but I'm happy to follow the convention. Also allows others to copy this file and make a new content collection without as much distraction with name changing.

src/util/custom-loaders.ts
cdrubin: After digging more today I see that starlight uses more specific naming and happy to copy that too! Updated as you recommended.

src/components/ProductMaturityText.astro
cdrubin: Implemented as you suggested. For the full power we should talk more because we may not be the right project to enjoy this full power because to get the squigglies I believe we would need to require writers to access values from the content collection directly and pass that value to the component which can then do this checking.

kodster28 (approved):
PoC approved

PR #27756, opened 2026-01-20
Add no-response workflow for pull requests

haleycode:
This workflow automatically manages pull requests that lack author responses by closing them after a specified period.


.github/workflows/no-response-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+# to work properly.
+
+# This action helps us flag pull requests for more information.
+# To trigger, add the `more-information-needed` label to a pull request.
+# The label will be removed if the author comments or makes updates.
+# If the author does not respond for a specific number of days, the pull request will be automatically closed.
+# However, if they do respond, the pull request will be re-opened automatically and the label will be removed.
+
+on:
+  pull_request_review:
```
colbywhite, line 13: The [usage](https://github.com/lee-dohm/no-response?tab=readme-ov-file#use) implies that this particular action might only work on issue events. Looking at the underlying [query](https://github.com/lee-dohm/no-response/blob/main/src/no-response.ts#L132), it might only get issues. 

It's confusing because a pull request sometimes is an issue and sometimes not. 🤔

colbywhite: Yea. It [looks like](https://github.com/cloudflare/cloudflare-docs/issues?q=is%3Aissue%20is%3Aopen) that's only going to close issues. Which can still be a good thing. But I suspect you were looking for PRs to close. 

Edit: And testing these sorts of actions is always tricky.

colbywhite:
Ah. What you're probably looking for is [actions/stale](https://github.com/marketplace/actions/close-stale-issues). The actions/* ones are official GitHub ones so we'd want to lean on those. They tend to be better, which I think is the case here since the official one does both issues and PRs.

haleycode:
Thanks @colbywhite! Given that info I'll close this out and try again.

PR #27867, opened 2026-01-26
sandbox: update Dockerfile docs to reflect SDK v0.7.0 changes

whoiskatrin:
## Summary

Updates the Sandbox Dockerfile reference documentation to align with the latest SDK changes (v0.7.0).

- Update version references from 0.3.3 to 0.7.0
- Document new image variants (default, python, opencode)
- Add standalone binary approach for arbitrary base images
- Update startup script guidance to use new `/sandbox` entrypoint
- Clarify that Python is only available in the `-python` variant

## Context

The current docs reference SDK version 0.3.3, but the SDK is now at v0.7.0 with significant changes:
- Multiple image variants instead of a single image
- New standalone binary for adding sandbox to any Dockerfile
- New entrypoint approach replacing the old `exec bun /container-server/dist/index.js`

Reference: https://github.com/cloudflare/sandbox-sdk

### DIFF
--- a/src/content/docs/sandbox/configuration/dockerfile.mdx
+++ b/src/content/docs/sandbox/configuration/dockerfile.mdx
@@ -7,37 +7,63 @@ sidebar:
 
 Customize the sandbox container image with your own packages, tools, and configurations by extending the base runtime image.
 
-## Base image
+## Base images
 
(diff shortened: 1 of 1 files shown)


ghostwriternr (approved):
💯

PR #27882, opened 2026-01-26
Change compatibility date to 2026-01-25

irvinebroque:
Updated the compatibility date for Node.js APIs in the documentation. Without this, confusing to humans and LLMs what compat date to use and you end up using an unnecessarily old date.


irvinebroque:
closing in favor of https://github.com/cloudflare/cloudflare-docs/pull/27887

PR #27887, opened 2026-01-26
Use dynamic date for nodejs_compat compatibility_date examples

MattieTK:
## Summary

- Updates `compatibility_date` examples to use `$today` instead of hardcoded dates
- Adds comments clarifying the minimum required date (2024-09-23) for Node.js compatibility
- Changes apply to the Node.js compatibility docs and reusable partials for Workers and Hyperdrive

The `WranglerConfig` component automatically replaces `$today` with the current date at build time, ensuring readers always see the latest recommended date.

## Context

We have noticed users creating projects with outdated compatibility dates, which we suspect is due to AI tools reading and reproducing the hardcoded dates from our documentation. Using dynamic dates should help ensure AI-assisted code generation uses current dates.

### DIFF
--- a/src/content/docs/hyperdrive/configuration/rotate-credentials.mdx
+++ b/src/content/docs/hyperdrive/configuration/rotate-credentials.mdx
@@ -30,7 +30,8 @@ The command above will output the ID of your Hyperdrive. Set this ID in the [Wra
 ```toml
 # required for database drivers to function
 compatibility_flags = [ "nodejs_compat" ]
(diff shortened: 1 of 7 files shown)


irvinebroque (approved):
nice one!!

irvinebroque:
👏

MattieTK:
I've got a bunch more matches to go over here so feel free to merge this but this PR will either increase in size or have followups

PR #27908, opened 2026-01-27
Add granular applications list component

cdrubin:
- Add granular applications list component
- improve error handling of product availability component and add parentheses handling, 
- Remove partial for granular controls list
- add better use of zod and typescript in middlecache-based content collections


src/components/ProductAvailabilityText.astro
```diff
@@ -4,17 +4,30 @@ import type { CollectionEntry } from "astro:content";
 
 interface Props {
 	product: CollectionEntry<"product-availability">["id"];
+	parentheses?: string;
```
colbywhite, line 7: Why not a boolean?

src/content/collections/granular-control-applications.ts
```diff
@@ -0,0 +1,52 @@
...
+	loader: middlecacheLoader("v1/application-controls/applications.json", {
+		parser: (fileContent: string) => {
+			const data = JSON.parse(fileContent);
+			const lookup: Record<
+				string,
+				{
+					display_id: string;
+					applications: [{ name: string; display_name: string }];
+				}
+			> = {};
```
colbywhite:
```ts
type GranularControlApplication = z.infer(granularControlApplicationsSchema)
```

By defining the type elsewhere, you can reuse it here.


```ts
const lookup: Record<string, GranularControlApplication>
```

src/content/docs/rules/trace-request/index.mdx
```diff
@@ -15,7 +15,7 @@ import { DirectoryListing, Plan, ProductAvailabilityText } from "~/components";
 
 <Plan type="all" />
 
-Cloudflare Trace (<ProductAvailabilityText product="cloudflare-trace" />) follows an HTTP/S request through Cloudflare's reverse proxy to your origin. Use this tool to understand how different Cloudflare configurations interact with an HTTP/S request for one of your hostnames. If the hostname you are testing is not [proxied by Cloudflare](/dns/proxy-status/), Cloudflare Trace will still return all the configurations that Cloudflare would have applied to the request.
+Cloudflare Trace <ProductAvailabilityText product="cloudflare-trace" parentheses="true" /> follows an HTTP/S request through Cloudflare's reverse proxy to your origin. Use this tool to understand how different Cloudflare configurations interact with an HTTP/S request for one of your hostnames. If the hostname you are testing is not [proxied by Cloudflare](/dns/proxy-status/), Cloudflare Trace will still return all the configurations that Cloudflare would have applied to the request.
```
colbywhite, line 18: Why move the parentheses as a prop? This now requires users of the comp to write 18 chars as opposed to 2.

src/util/custom-loaders.ts
```diff
@@ -18,10 +18,7 @@ type MiddlecacheOptions = {
 	url?: string;
 	parser?: (
 		text: string,
-	) =>
-		| Record<string, Record<string, unknown>>
-		| Array<Record<string, unknown>>
-		| Record<string, string | null>;
+	) => Record<string, Record<string, unknown>> | Array<Record<string, unknown>>;
```
colbywhite: Instead of making your own type, you can reuse the astro type. Unfortunately they don't explicitly export it so you'll need a bit of TS utilities.

```ts
type FileOptions = Parameters<typeof file>[1];
type MiddlecacheOptions = FileOptions & {url?: string}
```

src/content/collections/granular-control-applications.ts
```diff
@@ -0,0 +1,52 @@
...
+				const display_id = item.category
+					.split("-")
+					.map((w: string) => w[0].toUpperCase() + w.substring(1).toLowerCase())
+					.join(" ");
+
+				lookup[item.category] = {
+					display_id: display_id,
+					applications: item.applications,
+				};
+			}
```
colbywhite, line 44: What you're looking for here is a `reduce`

```ts
const lookup = data.reduce(
    (acc, item) => {
        // do stuff
        return {...acc, newEntry}
    }, 
    {} as Record<string, GranularControlApplication>
)
```

src/content/collections/granular-control-applications.ts
```diff
@@ -0,0 +1,52 @@
...
+});
+
+//const granularControlApplicationsSchema = z.array(categorySchema);
+
+const granularControlApplicationsCollectionConfig: CollectionConfig<
+	typeof granularControlApplicationsSchema
+> = {
+	loader: middlecacheLoader("v1/application-controls/applications.json", {
+		parser: (fileContent: string) => {
+			const data = JSON.parse(fileContent);
```
colbywhite, line 25: Can you confirm that this fileContent string will match the zod schema in this unconventional loader pattern? Will you need to use zod to validate this string? I think Astro will apply the schema against the the `lookup` value you return, but are you guaranteed that this `data` is a valid of array of X? 

(I say X because I think it wouldn't match the schema you gave Astro but some other schema. And now that I type that out I'm guessing that Astro is _not_ validating this string since you would need to provide _two_ schemas, right?)

src/content/collections/granular-control-applications.ts
```diff
@@ -0,0 +1,52 @@
...
+				{
+					display_id: string;
+					applications: [{ name: string; display_name: string }];
+				}
+			> = {};
+
+			for (const item of data) {
+				const display_id = item.category
+					.split("-")
+					.map((w: string) => w[0].toUpperCase() + w.substring(1).toLowerCase())
```
colbywhite: You had to manually specify string here? It wasn't able to infer that?

colbywhite: Oh, I think that's because you need a second schema. So it makes sense that at this point in the flow it doesn't know that category is a string. 

Some of that is semantic "make TS happy" but the reason TS isn't happy is because you have a potential bug here. If the thing you write to the file isn't in the right shape, then this code will fail. There's no check that category exists and is a string. There's also no check that `item.applications` exists, but I think when Astro checks the resulting `lookup` that would effectively get checked.

(There's also no check that `w` has a length > 2, but TS wouldn't save you there.)

My original recommendation of using a standard pattern for object loader still applies. But if we're too far down this path, I think you'll want more validation.

colbywhite (approved):

src/components/ProductAvailabilityText.astro
cdrubin: I also liked the idea of using a boolean, but the properties in HTML must be specified as strings in double quotes so it did not seem that helpful to introduce different types.

src/content/docs/rules/trace-request/index.mdx
cdrubin: To give the writers control of how to display the component and especially to handle cases - within the component - of when nothing should be displayed (like the value is GA) even though parentheses have been asked for.

src/util/custom-loaders.ts
cdrubin: Thank you, that is a great solution. I will try it! I wanted to import it :) Interesting that there is a way to do so anyway!

src/content/collections/granular-control-applications.ts
cdrubin: I have recently started to use less of the map and reduce functions than previously. I don't find them as easy to read as the explicit actions in a loop and for the work being done here readability is a stronger consideration I think.

src/content/collections/granular-control-applications.ts
cdrubin: Astro is validating the data in fileContent against schema provided and, if there are issues throws a `[InvalidContentEntryDataError] product-availability → XXX data does not match collection schema` for that array item

PR #27913, opened 2026-01-27
[WARP] Add 2026.1.89.1 beta release notes

toksdotdev:
Release note for Warp Client Beta (2026.1.89.1)

### DIFF
--- a/bin/fetch-warp-releases.js
+++ b/bin/fetch-warp-releases.js
@@ -21,6 +21,7 @@ const linesToRemove = [
 	"For Zero Trust documentation please see: <https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp>",
 	"For related Consumer documentation please see: https://developers.cloudflare.com/warp-client/",
 	"For Consumer documentation please see: <https://developers.cloudflare.com/warp-client/>",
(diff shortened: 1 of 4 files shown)


ranbel (approved):

bin/fetch-warp-releases.js
```diff
@@ -21,6 +21,7 @@ const linesToRemove = [
 	"For Zero Trust documentation please see: <https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp>",
 	"For related Consumer documentation please see: https://developers.cloudflare.com/warp-client/",
 	"For Consumer documentation please see: <https://developers.cloudflare.com/warp-client/>",
+	"For Consumer documentation please see: <https://developers.cloudflare.com/warp-client>",
```
kodster28, line 24: Why is this here? Is there a reason you can't fix this upstream?

kodster28 (changes requested):

bin/fetch-warp-releases.js
toksdotdev: this is written by hand, and depending on whoever authored the release, they might prefer to end with a `/` suffix or not. given this is trivial, i don't see reason not to include this here.

kodster28 (approved):

PR #27933, opened 2026-01-28
PCX: AI Review of Basic Auth code example

Oxyjun:
## Summary

- Fix potential panics in Rust example by adding bounds checking for malformed authorization headers and credentials
- Update Hono example to use only `PASSWORD` binding (hardcode username as `"admin"`) for consistency with JS/TS/Rust examples
- Add `/logout` route to Hono example for consistency with other examples


elithrar:
/bonk fix the latent issues here. Rust code should be correct.

PR #27934, opened 2026-01-28
Add D1 and React Router example

Oxyjun:
## Summary

- Add new example for querying D1 from a React Router v7 application
- Replaces the deprecated Remix pattern with the recommended React Router approach
- Includes setup instructions, route example, and type safety configuration

This is the React Router equivalent of the existing D1 and Remix example, following the recommendation that new projects use React Router instead of Remix.


elithrar:
/bonk review Jun's code. is it correct? ensure it is valid TypeScript. is it beginner friendly? 

Remove the "CloudflareEnviroment" and show how to use `wrangler types` to generate types.

PR #27995, opened 2026-01-30
[Stream] Webooks local testing guidance

### LINKED ISSUE #27546: Document how to test Cloudflare Stream webhooks

- https://developers.cloudflare.com/stream/manage-video-library/using-webhooks/

1. Include a _full_ example POST data, not a truncated one, of the kind of response Cloudflare would submit to a developer's registered Webhook. The example given is incomplete, and there is no explanation as to what the various properties are. For example, what is the `uid` property in the data sent to the Webhook?
2. Document how Cloudflare recommends the development and testing of a Cloudflare Stream webhook. I assumed that by using ngrok, and registering my ngrok URL to Cloudflare, that I would be able to see the incoming Webhook posts in my development environment. But despite registering my ngrok URL, and the registration being successful, I never received any Webhook calls to that URL. Additionally, my ngrok account showed no POST attempts to my ngrok domain from Cloudflare. My test posts from my own command line to my ngrok domain _were_ received so the ngrok domain and setup were working.

How does Cloudflare suggest a developer write and develop and test a Cloudflare Stream Webhook? 

### Additional information

I see that Cloudflare Tunnel might work, but it requires a developer to have DNS control over the Cloudflare account. Most developers might only have an API token given to them to work with Cloudflare stream with no control over DNS.  I do not have DNS control so I can't use Cloudflare Tunnel.

ToriLindsay:
Fixes #27546 

Updates the "Limitations" section to address local development. Adds a warning that third-party tunnels (like ngrok) are often blocked and recommends Cloudflare Tunnel as the supported alternative.


elithrar:
/bonk review this PR.

also create a new `Testing Webhooks locally` example under https://developers.cloudflare.com/stream/examples/ (and cross-link from https://developers.cloudflare.com/stream/manage-video-library/using-webhooks/ - that shows:

1. how to create a simple webhook server as a Cloudflare Worker
2. how to run Cloudflare Tunnel locally to expose that for testing
3. ensure that you update the webhook URL after publishing to production

Follow the structure of other Stream examples, make sure all code is valid TypeScript, and then re-review your work.

tsmith512:
We should note that this is for VOD (on-demand) videos only. Live webhooks are configured differently, which is documented.

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+  order: 11
+description: Test Cloudflare Stream webhook notifications locally using a Cloudflare Worker and Cloudflare Tunnel.
+---
+
+Cloudflare Stream cannot send [webhook notifications](/stream/manage-video-library/using-webhooks/) to `localhost` or local IP addresses. To test webhooks during local development, you need a publicly accessible URL that forwards requests to your local machine.
+
+This example shows how to:
+
+1. Create a simple Cloudflare Worker that receives and logs Stream webhook events.
+2. Use [Cloudflare Tunnel](/cloudflare-one/networks/connectors/cloudflare-tunnel/do-more-with-tunnels/trycloudflare/) to expose your local Worker dev server to the internet.
```
marciocloudflare:
```suggestion
2. Use [Cloudflare Tunnel](/cloudflare-one/networks/connectors/cloudflare-tunnel/do-more-with-tunnels/trycloudflare/) to expose your local Worker dev server to the Internet.
```

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+https://example-words-here.trycloudflare.com
+```
+
+Copy this URL. It changes every time you restart the tunnel.
+
+## 4. Register the tunnel URL as your webhook endpoint
+
+Use the Stream API to set the tunnel URL as your webhook notification URL:
+
+```bash
```
marciocloudflare: I'd suggest you use the <APIRequest> component here instead of the raw curl command. This also has the advantage of listing any required tokens

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+When you are done testing locally, deploy the Worker and update the webhook URL to your production endpoint:
+
+```sh
+npx wrangler deploy
+```
+
+Then update the webhook subscription to point to your deployed Worker URL:
+
+```bash
```
marciocloudflare: same comment here about APIRequest

marciocloudflare (approved):

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+- [Node.js](https://nodejs.org/) (v18 or later)
+- The [Wrangler CLI](/workers/wrangler/install-and-update/) installed (`npm install -g wrangler`)
+
+## 1. Create the webhook Worker
+
+Create a new Worker project and replace the generated `src/index.ts` with the following code. This Worker receives webhook `POST` requests, [verifies the signature](/stream/manage-video-library/using-webhooks/#verify-webhook-authenticity), and logs the payload.
+
+```ts title="src/index.ts"
+export interface Env {
+  // Set this secret with: npx wrangler secret put WEBHOOK_SECRET
```
tsmith512: If this is a demo about how to develop locally, this has to be stored in a local file https://developers.cloudflare.com/workers/configuration/secrets/#local-development-with-secrets

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+  if (!time || !receivedSig) {
+    return { valid: false, body };
+  }
+
+  // Build the source string: "<time>.<body>"
+  const sourceString = `${time}.${body}`;
+  const encoder = new TextEncoder();
+
+  const key = await crypto.subtle.importKey(
```
tsmith512: Before I had the secret stored in a local file, this line threw the following error:
`DataError: Imported HMAC key length (0) must be a non-zero value up to 7 bits less than, and no greater than, the bit length of the raw key data (0).`

src/content/docs/stream/examples/test-webhooks-locally.mdx
ToriLindsay: /bonk use the component here instead of the raw curl command. This also has the advantage of listing any required tokens

src/content/docs/stream/examples/test-webhooks-locally.mdx
ToriLindsay: /bonk use the component here instead of the raw curl command. This also has the advantage of listing any required tokens

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+    if (request.method !== "POST") {
+      return new Response("Method not allowed", { status: 405 });
+    }
+
+    const { valid, body } = await verifyWebhookSignature(
+      request,
+      env.WEBHOOK_SECRET,
+    );
+
+    if (!valid) {
```
tsmith512: Since this is a demo that webhooks are verified, we should probably say so

src/content/docs/stream/examples/test-webhooks-locally.mdx
```diff
@@ -0,0 +1,201 @@
...
+```bash
+curl -X PUT \
+  --header "Authorization: Bearer <API_TOKEN>" \
+  https://api.cloudflare.com/client/v4/accounts/<ACCOUNT_ID>/stream/webhook \
+  --data '{"notificationUrl":"https://your-worker.your-subdomain.workers.dev"}'
+```
+
+:::caution
+
+Remember to update the webhook URL after deploying to production. Stream only supports one webhook subscription per account, and a stale `trycloudflare.com` URL will silently fail once the tunnel is stopped.
```
tsmith512: So this is a thing that was tricky in testing --- when you update the webhook URL, we roll the secret, so the order of operations is tricky because the script is described first, then setting the url requires updating it. Also there's on mention of how to start the tunnel. That's not a Stream issue, but if this tutorial is intended to cover that, it should be mentioned

tsmith512 (changes requested):
This is a good start but following this tutorial was a little weird. The key thing I wanted to confirm is that the code to actually validate the webhook signature is correct --- that's really the only thing related to Stream here. And it does! But end-to-end as a tutorial, a couple key points are missing and some of the steps feel out of order. Let's call out clearly how to start the tunnel, then set up the script to use it next.

tsmith512:
/bonk can you address the feedback I left on the tutorial?

tsmith512 (approved):

PR #28023, opened 2026-02-02
Update caution note on sensitive information sharing

zegevlier:
### Summary

The caution note on the "Contacting Cloudflare Support" page warns against sharing sensitive information with Cloudflare. This warning implies that one should never share things like credit card information or API keys with Cloudflare, when it is meant to say that these should not be shared with Support.

### Documentation checklist


zegevlier:
Still seems relevant to me, please un-stale

WalshyDev (approved):

PR #28211, opened 2026-02-09
Add concurrency settings to bonk workflow

elithrar:
Implement concurrency settings for the bonk job to manage pull request executions.

Subsequent invocations on the PR will be queued instead of running and conflicting when they attempt to update the PR branch.

### DIFF
--- a/.github/workflows/bonk.yml
+++ b/.github/workflows/bonk.yml
@@ -10,7 +10,7 @@ jobs:
   bonk:
     if: github.event.sender.type != 'Bot'
     concurrency:
-      group: pr-${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bonk.yml
```diff
@@ -9,6 +9,9 @@ on:
 jobs:
   bonk:
     if: github.event.sender.type != 'Bot'
+    concurrency:
+      group: pr-${{ github.event.pull_request.number }}
```
kodster28:
```suggestion
      group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

kodster28: We use this pattern in the other actions we have.

kodster28 (approved):

PR #28295, opened 2026-02-12
[Workers] add best practices guide

### LINKED ISSUE #28291: new: Workers Best Practices guide
### guidance

We want a single page “Workers Best Practices” guide that is similar to the following:

- https://developers.cloudflare.com/durable-objects/best-practices/rules-of-durable-objects/
- https://developers.cloudflare.com/workflows/build/rules-of-workflows/
- informed by other “best practices” and “well architected” content beyond Cloudflare

Look at the issue & PR history of this repo as well as cloudflare/workers-sdk - deeply understand the most common issues, errors and configuration challenges users run into and how this best practices guide can help users.

It should use the code example format of the Workflows & Durable Objects guides in order to not only tell but SHOW users how to do things the right way. We should have opinions.

### requirements

The doc should ensure it covers:

* using the latest compatibility_date
* enabling nodejs_compat
* enabling logging and tracing and using sampling rates to control volume and manage costs
* using waitUntil to do work after a request, or tail workers for logging
* using streaming for unknown or large request and/or response bodies (both)
* using `wrangler secret` for secrets and not storing in source
* using `wrangler types` to generate types based on your bindings
* using bindings for Cloudflare native services (R2, KV, Workflows) instead of REST APIs or SDKs
* using workflows and queues for async and long running tasks that can be done in the background

There’s probably another 10-15 here based on the issue & PR research I’ve asked for. Order the page so that the most common and obvious are at the top.

Find the right place to put this. Use the code-review skill to both review your changes as well as inspiration for best practices as it itself is a source of best practices.

elithrar:
Adds a new "Workers Best Practices" page covering 21 rules across 7 sections, with code examples for each. Closes #28291.

The guide is informed by common issues from cloudflare/workers-sdk and cloudflare/cloudflare-docs over the past year, the code-review skill's common-patterns reference, and the existing Durable Objects and Workflows best practices guides.

**Covers the 9 explicitly requested topics** — compatibility dates, nodejs_compat, observability/tracing with sampling, waitUntil and Tail Workers, streaming, wrangler secret, wrangler types, bindings over REST APIs, Queues/Workflows for background work — **plus 12 more from issue/PR research:**

- Use wrangler.jsonc for non-secret config
- Set cache headers explicitly
- Service bindings for Worker-to-Worker RPC
- Do not store request-scoped state in global scope
- Keep startup fast (lazy init)
- Always await or waitUntil Promises
- Handle nullable binding returns
- Prefer Workers-compatible libraries
- Web Crypto for secure token generation + timingSafeEqual
- Explicit error handling over passThroughOnException
- Test with @cloudflare/vitest-pool-workers (with nodejs_compat gotcha)

Format mirrors "Rules of Durable Objects" and "Rules of Workflows" — each rule has concise prose followed by a TypeScriptExample or WranglerConfig showing the correct pattern (and anti-pattern where useful).

All code examples validated against latest `@cloudflare/workers-types` and wrangler config schema via the code-review skill. 0 critical/high/medium issues.

New files:
- `src/content/docs/workers/best-practices/index.mdx`
- `src/content/docs/workers/best-practices/workers-best-practices.mdx`


src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,943 @@
...
+		// console.log("Got a request to " + url.pathname);
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+For more information, refer to [Workers Logs](/workers/observability/logs/workers-logs/) and [Traces](/workers/observability/traces/).
+
+### Use Tail Workers for durable logging
```
irvinebroque: we should remove this and we shouldn't recommend Tail Workers — instead Workers Observability

https://developers.cloudflare.com/workers/observability/

@nevikashah

irvinebroque: I think if we leave this in we end up telling the agent that using Tail Workers is best practice, when in the vast majority of cases it's really not what you want

There is kind of a branch here that the agent can choose between the two, and I think better outcome to just always push towards first-class observability, versus rolling its own with Tail Workers

We should bring up Tail Workers in context of Workers for Platforms

irvinebroque: /bonk please take my feedback above, and use it to revise this file

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,943 @@
...
+
+`console.log` is best-effort. If a Worker crashes or is evicted, in-flight logs can be lost. For logs that must not be dropped (audit trails, billing events, compliance), use a [Tail Worker](/workers/observability/logs/tail-workers/). Tail Workers receive log events asynchronously and can forward them to any destination.
```
irvinebroque: This guidance doesn't really make a ton of sense, because to get something into a Tail Worker you have to....`console.log` in the first place, and so that's subject to crashes and evictions in the first place

The Workers Observability platform is a Tail Worker anyways, and so it's just always a simpler option, and no difference in the guarantees you can get

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,943 @@
...
+}
+```
+
+</TypeScriptExample>
+
+For more information, refer to [Workers errors](/workers/observability/errors/#cannot-perform-io-on-behalf-of-a-different-request).
+
+### Keep startup fast
+
+Top-level code runs on every cold start and counts against [CPU time limits](/workers/platform/limits/). Defer expensive setup to the first request using lazy initialization.
```
irvinebroque: I think this is an anti-pattern that is increasingly less and less helpful, particularly more that we do around cold starts

And the vast majority of apps, whatever client is getting loaded here is something used across many/most requests, not just one off for a narrow set of cases

If anything this kind of just makes performance feel non-deterministic...am skeptical of including it.

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,943 @@
...
+		return Response.json(result);
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+### Always await or waitUntil your Promises
+
+A `Promise` that is not `await`ed, `return`ed, or passed to `ctx.waitUntil()` is a floating promise. Floating promises cause silent bugs: dropped results, swallowed errors, and unfinished work. The Workers runtime may terminate your isolate before a floating promise completes.
```
irvinebroque: https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-floating-promises.html
https://typescript-eslint.io/rules/no-floating-promises/

irvinebroque: Maybe should explicitly recommend the agent enable this rule? Possibly too aggressive but think can help people a lot

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,943 @@
...
+		return new Response("OK");
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+## Security
+
+### Use Web Crypto for secure token generation
```
irvinebroque, line 747: Do we want people using WebCrypto or Node.js Crypto?

Will this guide the agent to think that it MUST use WebCrypto and can't use Node.js Crypto?

I think we should be explicit that Node.js crypto is available

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,943 @@
...
+	return crypto.subtle.timingSafeEqual(providedHash, expectedHash);
+
+	// 🔴 Bad: direct string comparison leaks timing information
+	// return provided === expected;
+}
+```
+
+</TypeScriptExample>
+
+### Do not use passThroughOnException as error handling
```
irvinebroque, line 805: /bonk please add this to the reference docs on the passThroughOnException()` API, if we do not already call this out. If you are able to, do this in a separate pull request

src/content/docs/workers/best-practices/workers-best-practices.mdx
elithrar: SGTM and good call.

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,898 @@
...
+}
+```
+
+</WranglerConfig>
+
+For more information, refer to [Compatibility dates](/workers/configuration/compatibility-dates/).
+
+### Enable nodejs_compat
+
+The [`nodejs_compat`](/workers/runtime-apis/nodejs/) compatibility flag gives your Worker access to Node.js built-in modules like `node:crypto`, `node:buffer`, `node:stream`, and others. Many libraries depend on these modules, and enabling this flag avoids cryptic import errors at runtime. Requires `compatibility_date` of `2024-09-23` or later.
```
irvinebroque:
```suggestion
The [`nodejs_compat`](/workers/runtime-apis/nodejs/) compatibility flag gives your Worker access to Node.js built-in modules like `node:crypto`, `node:buffer`, `node:stream`, and others. Many libraries depend on these modules, and enabling this flag avoids cryptic import errors at runtime.
```

I am tempted to omit this because this specific date being in our docs is the source of models setting this as the date when creating new Workers, and don't want to inadvertently reinforce this date by including this particular value

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,898 @@
...
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+For more information, refer to [wrangler types](/workers/wrangler/commands/#types).
+
+### Store secrets with wrangler secret, not in source
+
+Secrets (API keys, tokens, database credentials) must never appear in your Wrangler configuration or source code. Use [`wrangler secret put`](/workers/configuration/secrets/) to store them securely, and access them through `env` at runtime. Use a `.dev.vars` file for local development. It is automatically excluded from version control.
```
irvinebroque: I think it is time for us to recommend `.env` and stop with the `.dev.vars` thing

https://developers.cloudflare.com/workers/configuration/environment-variables/

It works, we just aren't explicit about it as default

irvinebroque: > It is automatically excluded from version control.

this isn't a true statement, it is only excluded so long as it is gitignored. This statement is presuming that 100% of projects start from our own templates, where it is gitignored.

src/content/docs/workers/best-practices/workers-best-practices.mdx
elithrar: /bonk address this:

* Remove the startup section. It's confusing.
* floating promises: suggest that users enable no-floating-promises rules in eslint or oxlint (link to https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-floating-promises.html + https://typescript-eslint.io/rules/no-floating-promises/) and show a one-liner for each as added context 
* the crypto - add a line that Node.js `crypto` is also supported and link to https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/
* recommend `.env` and NOT `.dev.vars` - make this clearer: https://developers.cloudflare.com/workers/configuration/environment-variables/

src/content/docs/workers/best-practices/workers-best-practices.mdx
elithrar: addressing + agree

elithrar:
/bonk address this feedback:

- Remove the startup section. It's confusing.
- floating promises: suggest that users enable no-floating-promises rules in eslint or oxlint (link to https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-floating-promises.html + https://typescript-eslint.io/rules/no-floating-promises/) and show a one-liner for each as added context
- the crypto - add a line that Node.js crypto is also supported and link to https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/
- recommend .env and NOT .dev.vars - make this clearer: https://developers.cloudflare.com/workers/configuration/environment-variables/

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,898 @@
...
+		});
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+For more information, refer to [Streams](/workers/runtime-apis/streams/).
+
+### Use waitUntil for work after the response
```
irvinebroque, line 267: /bonk I think this may need more to call out how `waitUntil` is needed in order to do more complex things with streams — ex when you return the stream but need to continue doing work to pump content through the stream, then close it out. Please make some suggestions here on this thread but DO NOT attempt to update this PR yet

irvinebroque:
/bonk please review the content in this pull request, extract out the topics and guidance, and then perform a thorough audit of the existing developer docs, to identify specific topics that are covered here, but are not addressed within our existing developer docs. Do not yet attempt to address these, just leave a comment here on this pull request that details your findings, and provides a prioritized list of recommendations for content additions and modifications to the existing developer docs

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,882 @@
...
+
+</TypeScriptExample>
+
+For more information, refer to [Context](/workers/runtime-apis/context/).
+
+## Architecture
+
+### Use bindings for Cloudflare services, not REST APIs
+
+Cloudflare services like R2, KV, D1, Queues, and Workflows are available as [bindings](/workers/runtime-apis/bindings/). Bindings are direct, in-process references that require no network hop, no authentication, and no extra latency. Using the REST API from within a Worker wastes time and adds unnecessary complexity.
```
irvinebroque:
```suggestion
Some Cloudflare services like R2, KV, D1, Queues, and Workflows are available as [bindings](/workers/runtime-apis/bindings/). Bindings are direct, in-process references that require no network hop, no authentication, and no extra latency. Using the REST API from within a Worker wastes time and adds unnecessary complexity.
```

Think needs qualifier to not mislead into thinking everything available via bindings

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,882 @@
...
+		});
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+### Use Queues and Workflows for async and background work
+
+Long-running, retriable, or non-urgent tasks should not block a request. Use [Queues](/queues/) to decouple producers from consumers with guaranteed delivery and automatic retries. Use [Workflows](/workflows/) for multi-step, durable processes that survive failures.
```
irvinebroque: /bonk this doesn't provide much guidance on WHEN to use a Workflow rather than a Queue, or visa-versa. Please research the existing Cloudflare Docs to understand if we already provide such guidance. Use what you find to recommend changes we should make to this section. Do not yet make edits to this PR.

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,882 @@
...
+		return Response.json({ userId: auth.userId });
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+### Use Hyperdrive for external database connections
+
+Connecting directly to a remote database (PostgreSQL, MySQL) from a Worker adds hundreds of milliseconds of latency per request due to TCP handshake, TLS negotiation, and connection setup on every invocation. [Hyperdrive](/hyperdrive/) eliminates this by maintaining a regional connection pool close to your database and caching query results where possible.
```
irvinebroque: /bonk this should be more direct — you should always use Hyperdrive when connecting to a remote PostgreSQL or MySQL database

And the first line is confusing and makes it sound like anytime you connect from a Worker to a database, it's slow. That's not true. it just needs to be explicit about - always use Hyperdrive

src/content/docs/workers/best-practices/workers-best-practices.mdx
```diff
@@ -0,0 +1,882 @@
...
+
+</TypeScriptExample>
+
+For more information, refer to [Durable Objects WebSocket best practices](/durable-objects/best-practices/websockets/).
+
+### Use Workers Static Assets for new projects
+
+[Workers Static Assets](/workers/static-assets/) is the recommended way to deploy static sites, single-page applications, and full-stack apps on Cloudflare. If you are starting a new project, use Workers instead of Pages. Pages continues to work, but new features and optimizations are focused on Workers.
+
+For a purely static site, point `assets.directory` at your build output. No Worker script is needed. For a full-stack app, add a `main` entry point and an `ASSETS` binding to serve static files alongside your API.
```
irvinebroque, line 554: /bonk please evalulate if you recommend that this section also include guidance around `run_worker_first`. https://developers.cloudflare.com/workers/static-assets/binding/#run_worker_first

src/content/docs/workers/best-practices/workers-best-practices.mdx
elithrar: /bonk address the comment here

elithrar:
/bonk do a final review. use the code-review skill to make sure our examples are valid. catch any outstanding comments.

elithrar:
Reviewed the current state against all PR comments:

- All 10 actionable review comments are incorporated: Tail Workers removed in favor of Workers Observability, `.env` over `.dev.vars`, no hardcoded compatibility date, Node.js crypto mentioned, `no-floating-promises` lint rules added, Hyperdrive rewritten to lead with "always use", bindings qualified, Queues vs Workflows guidance added, startup section removed, `passThroughOnException` caveat present.
- Fixed the concat example (lines 228-263) — added `ctx: ExecutionContext` to the signature and wrapped the pipeline promise in `ctx.waitUntil(pipeline)` so the Worker stays alive until all chunks are piped through. Without this, the runtime could terminate the Worker after returning the response but before the stream is fully written.
- The broader docs audit bonk posted (`.env` vs `.dev.vars` inconsistency across 8+ tutorials, 32 pages with hand-written `Env`, `passThroughOnException` messaging conflict in errors.mdx) is out of scope for this PR but worth tracking as follow-ups.

irvinebroque (approved):

PR #28304, opened 2026-02-12
Fix Bonk getting stuck on tool approval in CI

elithrar:
Bonk keeps getting stuck in CI waiting for tool approval when the code-review skill tells it to `npm pack` packages. Example: https://github.com/cloudflare/cloudflare-docs/actions/runs/21962906948/job/63444925338

- Add `opencode.jsonc` with `permission: "allow"` and `continue_loop_on_deny: true` so the agent does not block on approval prompts in GitHub Actions
- Remove `npm pack` retrieval steps from the code-review skill — point it at `node_modules/` instead, which avoids the bash commands that trigger the stuck loop
- Add `timeout-minutes: 30` to the Bonk workflow so stuck runs get killed instead of burning 6 hours

### DIFF
--- a/.agents/skills/code-review/SKILL.md
+++ b/.agents/skills/code-review/SKILL.md
@@ -5,19 +5,17 @@ description: Reviews Workers and Cloudflare Developer Platform code for type cor
 
 Your knowledge of Cloudflare Workers APIs, types, and wrangler configuration may be outdated. **Prefer retrieval over pre-training** for any Workers code review task.
 
-## Retrieval Sources
+## Reference Sources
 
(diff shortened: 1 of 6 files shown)


irvinebroque (approved):

kodster28 (approved):
confirmed based on:
- https://opencode.ai/docs/permissions#configuration
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idtimeout-minutes
- https://github.com/anomalyco/opencode/blob/4f51c0912d76698325862e8fcd7d484b7b9a61fe/packages/sdk/openapi.json#L10021

PR #28356, opened 2026-02-16
[Workers/Frameworks] Fix: Section header consistency

markdembo:
Fix inconsistent styling + numbering

Screenshot 2026-02-16 at 07 28 46


MattieTK (approved):
thanks mark!

PR #28361, opened 2026-02-16
[MWAN] Fix missed references from product rename (pass 2)

marciocloudflare:
## Summary

Follow-up to #28358 — fixes ~70 remaining references missed in the Magic WAN → Cloudflare WAN rename.


src/content/partials/networking-services/account-id-api-key.mdx
pedrosousa, line 2: This file should declare `showSiteId` as an optional parameter.

pedrosousa (approved):

PR #28369, opened 2026-02-16
Add ids to all directory entry YAML files and helper script

cdrubin:
To help push the conversation reasonably forward, adding an id entry to every directory entry file within src/content/directory. This id should be unique across these files.

In tools/directory-entry-ids is a helper script for adding ids to files that do not have them and for doing a very basic check that the existing ids are unique.


tools/directory-entry-ids
```diff
@@ -0,0 +1,107 @@
+#! /bin/env node
```
pedrosousa, line 1:
```suggestion
#!/bin/env node
```
This usually doesn't have a space.

pedrosousa (approved):

PR #28372, opened 2026-02-16
Add Cache section header and link to Cache API in Images bindings docs

olafghanizadeh:
### Summary

Adds header and a URL to the Cache API page, that makes it easier to find information on caching when using Image Transformations via a Worker Binding.

### DIFF
--- a/src/content/docs/images/transform-images/bindings.mdx
+++ b/src/content/docs/images/transform-images/bindings.mdx
@@ -101,7 +101,9 @@ return response;
 
 - Outputs information about the image, such as `format`, `fileSize`, `width`, and `height`.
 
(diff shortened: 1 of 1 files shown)


ToriLindsay (approved):

ToriLindsay:
Hi @olafghanizadeh - Agreed with you on giving it distinct formatting, but updated it to be a note rather than a section, since it's more like a callout, and approved. But let me know if you disagree.

olafghanizadeh:
@ToriLindsay Sounds good to me! I was looking for information on caching when implementing image transformations in a Worker and felt like I just stumbled across it when reading about .info().

PR #28373, opened 2026-02-16
Add helper script and include id in directory entry yaml files

cdrubin:
To help push the conversation reasonably forward, adding an id entry to every directory entry file within src/content/directory. This id should be unique across these files.

In tools/directory-entry-ids is a helper script for adding ids to files that do not have them and for doing a very basic check that the existing ids are unique.


tools/directory-entry-ids
```diff
@@ -0,0 +1,107 @@
...
+			);
+		}
+	} while (existingIds.has(id));
+
+	return id;
+}
+
+// Process all .yaml files in the directory
+const dir = join(__dirname, "../src/content/directory");
+const files = readdirSync(dir).filter((f) => f.endsWith(".yaml"));
```
pedrosousa, line 38: We have several `.yml` files (instead of `.yaml`) in the `directory` folder... maybe we can rename them now for consistency?

pedrosousa (approved):

PR #28479, opened 2026-02-20
Update build image docs

yomna-shousha:
Draft


elithrar:
/bonk review this and suggest any fixes for @yomna-shousha to address.

yomna-shousha:
Working on a release to auto-update version releases so this draft hopefully should not be needed.

PR #28484, opened 2026-02-20
[Changelog] Only render descriptions

kodster28:
### Summary

The changelog as it is just doesn't work, size-wise (it's loading too many entries that -- at times -- crashes the browser by consuming too much memory).

This is an alternative approach that still supports filtering.
image


mvvmm (approved):

mvvmm:
Reduces DOM size to 0.45 MB and DOM elements to 3666.

PR #28489, opened 2026-02-20
doc: custom hostname PATCH requires type+method when changing CA

dmmulroy:
## Summary

- Documents that `ssl.type` and `ssl.method` are **required** when using PATCH to change `ssl.certificate_authority` on an existing custom hostname — omitting them silently fails or returns error 1434
- Adds examples and valid values across three pages: common API calls, certificate issuance, and troubleshooting

### `common-api-calls.mdx`
- Updated "Edit custom hostname" table row to link to new CA switching section
- Added `### Change certificate authority` subsection with valid values, method info, and two collapsible curl examples (switch to Google, reset to default)

### `issue-certificates.mdx`
- Added `## Change certificate authority` section between existing "Certificate authorities" and "Certificate details" sections
- Includes PATCH examples and a note on valid `type`/`method` values

### `troubleshooting.mdx`
- Added `## Certificate authority change not taking effect via API` troubleshooting entry with example and valid CA values

## Context

The PATCH `/zones/{zone_id}/custom_hostnames/{id}` endpoint requires `ssl.type` ("dv") and `ssl.method` ("http"/"txt") when changing `ssl.certificate_authority`, but this wasn't documented. The API reference marks `type` and `method` as optional, which is misleading for this use case. Customers (notably beehiiv) hit error 1434 (`ErrValidationTypeMethodRequired`) when sending only `certificate_authority`.

### DIFF
--- a/src/content/docs/cloudflare-for-platforms/cloudflare-for-saas/reference/troubleshooting.mdx
+++ b/src/content/docs/cloudflare-for-platforms/cloudflare-for-saas/reference/troubleshooting.mdx
@@ -124,6 +124,33 @@ Consider the following solutions:
 - Use the [Edit Custom Hostname](/api/resources/custom_hostnames/methods/edit/) endpoint to set the `certificate_authority` parameter to `google`: this sets Google Trust Services as the CA for your custom hostnames. In your API call, make sure to also include `method` and `type` in the `ssl` object.
 - If you are using a custom certificate for your custom hostname, refer to the [custom certificates troubleshooting](/ssl/edge-certificates/custom-certificates/troubleshooting/#lets-encrypt-chain-update).
 
+## Certificate authority change not taking effect via API
+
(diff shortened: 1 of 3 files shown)


mitch292:
I think if we can just document at a high level that there are essentially two flavors of custom hostname SSL objects. "Managed" and "Custom". Managed means Cloudflare handles the lifecycle of the certificate (order / validation / renewal). Custom means a customer comes with a private key and certificate already.

When it's a managed custom hostname, two fields are always required in the PATCH 
- ValidationType (json "type")
- ValidationMethod (json "method")

When it's custom, the requirements on the ssl object differ in the PATCH
- CustomCert (json "custom_cert") is always required
- One of CustomKey (json "custom_key") or CustomCSRID (json "custom_csr_id") is required

mitch292 (approved):
LGTM - Thank you!

elithrar:
/bonk review and suggest any fixes for style or code issues

PR #28541, opened 2026-02-23
[Workers] Split cohort-based deployments and version overrides into standalone pages

irvinebroque:
## Summary

Restructures the gradual deployments documentation by extracting two large sections into their own dedicated pages, and refocuses the parent page on core concepts.

### New pages created

- **`cohort-based-deployments.mdx`** — standalone page for cohort-based deployments (previously an H2 section inside `gradual-deployments.mdx`). Contains all the original content: how cohort routing works, creating a cohort-based deployment via API, setting the cohort header, Transform Rules example, authenticated cohort assignment with TypeScript example, routing behavior details, naming rules, deployment limits, service bindings caveat, Durable Objects caveat, and error handling.

- **`version-overrides.mdx`** — standalone page for version overrides (previously an H2 section inside `gradual-deployments.mdx`). Contains all the original content: header format, dictionary structured header spec, when overrides apply, verification note, and the 0%-deployment testing example. Adds an explicit mention of Logpush as an observability option alongside the version metadata binding.

### Restructured `gradual-deployments.mdx`

**Before:** A 500+ line page that mixed core gradual deployment concepts with the full cohort-based deployments reference, the full version overrides reference, and a step-by-step tutorial — all at the same level.

**After:** The page is reorganized into a clearer flow:
1. **Intro** — what gradual deployments are, with bullet points that now cross-link to the new cohort-based deployments and version overrides pages.
2. **How it works** — new short section explaining the percentage-based routing model and linking to version affinity.
3. **Version affinity** — unchanged.
4. **Gradual deployments with static assets** — moved up from its previous position (was buried after the tutorial).
5. **Gradual deployments for Durable Objects** — unchanged.
6. **Observability** — unchanged.
7. **Limits** — unchanged.
8. **Get started** — the Wrangler + Dashboard tutorial moved to the bottom of the page so conceptual content comes first. Internal anchor links updated to match new heading positions.

### Link updates

- **`previews.mdx`**: version overrides link updated from `gradual-deployments/#version-overrides` → `version-overrides/`.
- **`workers.yaml`** (release notes): same version overrides link update.
- **`gradual-deployments.mdx`**: internal `#version-overrides` reference changed to cross-link to the new page. Self-referential full-path anchors simplified to relative `#anchors`.
- Removed unused `TypeScriptExample` import from `gradual-deployments.mdx` (the TypeScript example moved to the cohort page).
- Fixed minor typo: "in to access" → "to access" in the runtime binding section.

### DIFF
--- a/src/content/docs/workers/configuration/previews.mdx
+++ b/src/content/docs/workers/configuration/previews.mdx
@@ -27,7 +27,7 @@ Preview URLs can be:
 - Used for collaboration between teams to test code changes in a live environment and verify updates.
 - Used to test new API endpoints, validate data formats, and ensure backward compatibility with existing services.
 
-When testing zone level performance or security features for a version, we recommend using [version overrides](/workers/configuration/versions-and-deployments/gradual-deployments/#version-overrides) so that your zone's performance and security settings apply.
+When testing zone level performance or security features for a version, we recommend using [version overrides](/workers/configuration/versions-and-deployments/version-overrides/) so that your zone's performance and security settings apply.
 
 :::note
 Preview URLs are only available for Worker versions uploaded after 2024-09-25.
--- a/src/content/docs/workers/configuration/versions-and-deployments/cohort-based-deployments.mdx
+++ b/src/content/docs/workers/configuration/versions-and-deployments/cohort-based-deployments.mdx
@@ -1,0 +1,226 @@
+---
+pcx_content_type: configuration
+title: Cohort-based deployments
+description: Route requests to different Worker version configurations based on named user segments using cohort-based deployments.
+---
+
+import { Example, TypeScriptExample } from "~/components";
+
+Cohort-based deployments extend [gradual deployments](/workers/configuration/versions-and-deployments/gradual-deployments/) by routing requests to different version configurations based on named user segments. Instead of applying one percentage split to all traffic, you define independent version percentages for each cohort.
+
+Use cohort-based deployments to:
+
+- Test changes with internal users or beta testers before external customers.
+- Roll out to free-tier customers before enterprise customers to limit blast radius.
+- Hold back specific user segments during incident response.
+
+## How cohort-based deployments work
+
+A standard gradual deployment has a single `versions` array that applies to every request. A cohort-based deployment adds a `cohorts` array alongside that default. Each cohort defines its own name and version configuration.
+
+When a request arrives, the system decides which version configuration to use in two steps:
+
+**Step 1 — Select the version configuration.**
+
+1. If the request has a `Cloudflare-Workers-Version-Overrides` header that matches a version ID present anywhere in the deployment (default or any cohort), that version is used immediately. No further routing applies. For more information, refer to [Version overrides](/workers/configuration/versions-and-deployments/version-overrides/).
+2. If the request has a `Cloudflare-Workers-Cohort` header that matches a defined cohort name, that cohort's `versions` array replaces the default for the next step.
+3. Otherwise, the default `versions` array is used.
+
+**Step 2 — Select a version within the chosen configuration.**
+
+1. If the request has a `Cloudflare-Workers-Version-Key` header, deterministic hashing selects a version from the configuration chosen in Step 1.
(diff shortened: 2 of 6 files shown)


irvinebroque:
/bonk please review to ensure that we are not removing content we shouldn't be from the docs that already existed in src/content/docs/workers/configuration/versions-and-deployments/gradual-deployments.mdx

irvinebroque:
/bonk the gradual deployments docs should link to the cohort-based deployments docs at the bottom as a next step / next thing to read. In each of the three docs pages, it is important that the reader is able to discover the other one, ex: discover version overrides from the gradual deployments docs, discover gradual deployments from the cohort-based deployments docs, etc.

Please also include documentation on the following pattern for using cohort-based deployments:

Gateway Entrypoint Pattern
This pattern uses your Worker's current entry point to determine cohort, then calls itself on a separate entrypoint via a [loopback binding](https://developers.cloudflare.com/workers/runtime-apis/context/#exports) with the resolved cohort:


```ts
import { WorkerEntrypoint} from "cloudflare:workers";

// Default entrypoint - resolves cohort and forwards to Service entrypoint
export default {

    // Step 1: Determine cohort based on your application logic
    const accountId = request.headers.get("cf-account-id");

    const cohort = accountId
      ? resolveCohortFromAccountId(request, env, accountId)
      : resolveCohortFromUrl(request, env);

    // Step 2: Call yourself via loopback binding with the resolved cohort
    return ctx.exports.Service({ version: { cohort } }).handle(request);
  }
};

function resolveCohortFromAccountId(request, env, accountId) {  

  const plan = lookupAccountPlan(accountId, env);
  
  if (isEmployee(request)) return "employee";
  if (plan === "free") return "free";
  if (plan === "enterprise") return "ent";
  return "paid";
}

function resolveCohortFromUrl(request, env) {
  
  // URL pattern example: https://<ACCOUNT_ID>.r2.cloudflarestorage.com/bucket/key
  const url = new URL(request.url);
  const hostname = url.hostname;
  const accountMatch = hostname.match(/^([a-f0-9]+)\.r2\.cloudflarestorage\.com$/i);

  // Default when account cannot be determined
  if (!accountMatch) {
    return "ent"; 
  }

  const accountId = accountMatch[1];
  const plan = lookupAccountPlan(accountId, env);
  
  if (plan === "free") return "free";
  if (plan === "some-special-entitlement") return "ent";
  return "paid";
}

export class Service extends WorkerEntrypoint {
	async handle(request){
		//The rest of your code goes here
   }
}
```

Note: Be sure to enable the `enable_ctx_exports`  compatibility flag in wrangler. (https://developers.cloudflare.com/workers/configuration/compatibility-flags/#enable-ctxexports)

Accessing Cohort in Your Code
Although some Observability support will be included (see Observability section below), you can use ctx.version to access cohort and version selection metadata for any other observability needs you may have.


```ts
export default {
    const { cohort, override, key, metadata } = ctx.version;
    
    // cohort: the cohort used for version selection (if any)
    // override: the version specified by an override header (if any)
    // key: the version specified by an affinity key header (if any)  
    // metadata: { id, tag, timestamp } of selected version
    
    console.log(`Serving request with cohort: ${cohort}`);
    // ...
  }
};
```

irvinebroque:
/bonk the gradual deployments docs should be first in this list, then rollbacks, then cohort based deployments

Screenshot 2026-02-23 at 4 28 26 PM

Right now the gateway entrypoint pattern is treated as a later example, rather than a mainline part of the docs.

I want you to refactor the cohort-based deployment docs to show that there are effectively two ways of doing cohort-based deployments - one where in the entrypoint, you lookup the cohort, and another where the cohort is passed in externally, via the header. And these should be treated equal weight in the docs. They are just two different approaches.

irvinebroque:
/bonk

Based on these docs:

https://bib-cohorts.preview.developers.cloudflare.com/workers/configuration/versions-and-deployments/cohort-based-deployments/#accessing-cohort-in-your-code

I think that these docs need to be updated as well:

https://developers.cloudflare.com/workers/runtime-apis/bindings/version-metadata/

PR #28547, opened 2026-02-24
feat: Replace mermaid.js with beautiful-mermaid for build-time diagram rendering

irvinebroque:
## Summary

Replaces the client-side mermaid.js rendering pipeline with build-time SVG rendering using [beautiful-mermaid](https://github.com/lukilabs/beautiful-mermaid) ([live demo](https://agents.craft.do/mermaid)). Diagrams are now rendered to inline SVG at build time by the rehype plugin, eliminating FOUC (flash of unstyled content) and reducing client-side JavaScript.

## Before / After

Screenshot 2026-02-23 at 9 39 40 PM
Screenshot 2026-02-23 at 9 39 26 PM

### Rehype plugin (`src/plugins/rehype/mermaid.ts`) — complete rewrite

The plugin now calls `renderMermaidSVG()` from `beautiful-mermaid` at build time for all `` ```mermaid `` code fences (~161 diagrams). Key design decisions:

- **CSS custom properties for theme switching**: Instead of baking in color values, the SVG references `var(--mermaid-bg)`, `var(--mermaid-fg)`, etc. Since the SVG is inline in the page HTML, CSS variable resolution happens in the browser via the cascade. Theme switching works by changing CSS custom properties on `.mermaid-container` — no JavaScript re-rendering needed.

- **Proper HAST parsing**: Uses `hast-util-from-html` to parse the SVG string into proper HAST nodes (the `type: "raw"` approach is unreliable in Astro's MDX pipeline).

- **SVG post-processing**: Strips the Google Fonts `@import` (unnecessary since the page already loads Inter), injects `<title>` and `<desc>` elements from `accTitle`/`accDescr` for accessibility, and adds `role="img"` with `aria-labelledby`.

- **Graceful error handling**: If a diagram fails to render (invalid syntax, unsupported features), the plugin logs a warning and preserves the raw diagram text in a visible error state — the build never breaks.

- **Original text preservation**: Stores the original mermaid source as a `data-diagram` attribute on the container for export pipelines (e.g., LLM markdown export).

- **Diagram normalization**: Strips trailing semicolons from diagram headers (e.g., `flowchart TD;` → `flowchart TD`) for compatibility with 4 legacy diagrams.

### Client script (`src/scripts/mermaid.ts`) — rewritten as slim fallback

The client script now only handles ~9 legacy `<pre class="mermaid">` blocks that use JSX template expressions (which are evaluated at component render time, after the rehype plugin has already run). Key improvements:

- **Dynamic import**: Uses `import("beautiful-mermaid")` so the library is only fetched on the few pages that actually have legacy blocks, not on every page.
- **Same library**: Uses `beautiful-mermaid` (not the old `mermaid` package) for consistent rendering between build-time and client-side paths.
- **Theme switching**: Observes `data-theme` changes and re-renders with the correct Cloudflare-branded color palette.

### CSS (`src/styles/mermaid.css`) — updated for dual rendering

- Added `--mermaid-*` CSS custom properties with Cloudflare-branded light/dark variants
- Added `.mermaid-diagram` styles for build-time rendered SVGs
- Added `.mermaid-error` styles for failed diagram fallbacks
- Preserved all annotation footer styles
- Kept `pre.mermaid` styles for the client-side fallback path

### Dependencies

- **Added**: `beautiful-mermaid` (MIT — build-time only for most pages)
- **Added**: `hast-util-from-html` (for parsing SVG strings into proper HAST nodes)
- **Removed**: `mermaid` v11.12.0

### Tests (`src/plugins/rehype/index.node.test.ts`)

Updated from 2 basic tests to 7 comprehensive tests covering:
- Successful SVG rendering from code fences
- `accTitle` extraction into annotation footer and SVG `<title>`
- No annotation footer without `accTitle`
- Graceful error fallback for invalid syntax
- CSS variable references in SVG output
- Original diagram text stored as data attribute
- Existing "ignores non-mermaid code" test preserved

## Diagram audit

Audited all 170 mermaid diagrams across the docs:

| Type | Count | beautiful-mermaid support |
|------|-------|--------------------------|
| `flowchart` / `graph` | 151 | ✅ |
| `sequenceDiagram` | 17 | ✅ |
| `stateDiagram-v2` | 1 | ✅ |
| `@{ shape: }` v11 syntax | 4 | ✅ (tested) |
| Trailing semicolons (`TD;`) | 4 | ✅ (normalized) |

**Zero unsupported diagram types found.** No pie, gantt, journey, mindmap, or other unsupported types exist in the docs.

## Performance impact

| Metric | Before | After |
|--------|--------|-------|
| Client JS (every page) | mermaid loaded globally via Head.astro | 0 (no JS for build-time diagrams) |
| Client JS (legacy pages) | same | beautiful-mermaid via dynamic import (~9 pages only) |
| FOUC | Yes (`visibility: hidden` hack) | No (SVG in initial HTML) |
| Theme switch | Full re-render via MutationObserver | CSS cascade (no JS) |
| Build time | Minimal (no diagram processing) | +~2-5s (renders 161 SVGs) |

## How theme switching works

```
┌─────────────────────────────────────────────────┐
│ .mermaid-container                              │
│   --mermaid-bg: #ffffff  (light)                │
│   --mermaid-fg: #1d1d1d  (light)               │
│   --mermaid-accent: #f6821f                     │
│                                                 │
│   ┌─────────────────────────────────────────┐   │
│   │ <svg style="--bg:var(--mermaid-bg);     │   │
│   │              --fg:var(--mermaid-fg)">    │   │
│   │   <style>                               │   │
│   │     --_text: var(--fg);                  │   │
│   │     --_node-fill: color-mix(in srgb,    │   │
│   │       var(--fg) 3%, var(--bg));          │   │
│   │   </style>                              │   │
│   │   ... diagram elements ...              │   │
│   └─────────────────────────────────────────┘   │
│                                                 │
│   :root[data-theme="dark"] .mermaid-container   │
│     --mermaid-bg: #1d1d1d  (dark)              │
│     --mermaid-fg: #f2f2f2  (dark)              │
└─────────────────────────────────────────────────┘
```

When Starlight toggles `data-theme`, the CSS custom properties update via the cascade, and `color-mix()` in the SVG's `<style>` block automatically derives all element colors — no JavaScript involved.

## Validation

- [x] `npm run check` — 0 errors, 0 warnings
- [x] `npm run lint` — clean
- [x] `npm run format:core:check` — clean
- [x] `npm run test` — 36/36 tests pass (Worker test failure is pre-existing, unrelated)
- [ ] `npm run build` — needs full local build verification

## Known limitations

1. **Legacy `<pre class="mermaid">` blocks**: The ~9 blocks using JSX template expressions (`${props.productName}`) cannot be rendered at build time. They use the client-side fallback. A follow-up PR could create a `<Mermaid>` Astro component to handle these server-side.

2. **Client-side theme switching for legacy blocks**: Unlike build-time diagrams (which use CSS variables), client-side rendered diagrams bake in color values and require re-rendering on theme change. This is functionally equivalent to the old behavior.

## Files changed

| File | Change |
|------|--------|
| `package.json` | +beautiful-mermaid, +hast-util-from-html, -mermaid |
| `src/plugins/rehype/mermaid.ts` | Complete rewrite: build-time SVG rendering |
| `src/scripts/mermaid.ts` | Rewritten as slim fallback with dynamic import |
| `src/styles/mermaid.css` | Added theme variables, new selectors, error styles |
| `src/plugins/rehype/index.node.test.ts` | 2 → 7 tests, covering SVG rendering |
| `src/scripts/MERMAID.md` | Updated architecture docs |


kodster28 (changes requested):
Just putting a preemptive, `we need to talk about this approach` comment here.

Anything -- and really _anything_ -- that has the potential to increase build times deserves a lot of scrutiny.

Need to have some stats / clear confirmation here as to the "time cost" of this optimization before we'd want to move any further.

PR #28553, opened 2026-02-24
Image Transformation Nits: Make Aliases more visible + Remove unwanted JS annotations on URL blocks

mglewis:
### Summary

Some small changes based on internal user feedback

### Screenshots (optional)

Old: 
image

New:
image

### DIFF
--- a/src/content/partials/images/anim.mdx
+++ b/src/content/partials/images/anim.mdx
@@ -7,7 +7,7 @@ Whether to preserve animation frames from input files. Default is `true`. Settin
 
 <Tabs>
   <TabItem label="URL format">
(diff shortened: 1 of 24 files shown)


vicb (approved):
LGTM.

Thanks for the quick update @mglewis.

Would it make sense to make the title "height / h" or "height (h)" so that the alias is also shown on the right section

Image

mglewis:
> Would it make sense to make the title "height / h" or "height (h)" so that the alias is also shown on the right section

In this case I'd have a preference for keeping the titles clean

Oxyjun (approved):

PR #28555, opened 2026-02-24
[Bonk] Bonk should only run if explicitly asked

kodster28:
We want to avoid triggering a bunch of 1 min builds to get to [this outcome](https://github.com/cloudflare/cloudflare-docs/actions/runs/22355389298/job/64693468794?pr=28527#step:5:62).

cc: @elithrar / @irvinebroque for viz

### DIFF
--- a/.github/workflows/bonk.yml
+++ b/.github/workflows/bonk.yml
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   bonk:
-    if: github.event.sender.type != 'Bot' && contains(github.event.comment.body, '/bonk')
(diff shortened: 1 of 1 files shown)


.github/workflows/bonk.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   bonk:
-    if: github.event.sender.type != 'Bot'
+    if: github.event.sender.type != 'Bot' && contains(github.event.comment.body, '/bonk')
```
elithrar:
```suggestion
    if: github.event.sender.type != 'Bot' && (contains(github.event.comment.body, '/bonk') || contains(github.event.comment.body, '@ask-bonk'))
```

mvvmm:
/bonk does this change do what it intends?

colbywhite (approved):

PR #28602, opened 2026-02-25
Fix docs regarding custom Docker setup for sandboxes

### LINKED ISSUE #371: Docs probably incorrect about path for custom Dockerfile setup

Muhammad-Bin-Ali:
### Summary

Fixes https://github.com/cloudflare/sandbox-sdk/issues/371

The "Custom startup scripts" section recommends `ENTRYPOINT ["/sandbox"]` when extending the base sandbox image. But `/sandbox` does not exist in the base images — the binary is at `/container-server/sandbox`. This causes containers to fail on startup with `exec /sandbox: no such file or directory`.

The fix: don't override `ENTRYPOINT` at all when extending the base image. It already sets `ENTRYPOINT ["/container-server/sandbox"]`. Just use `CMD`.

#### Changes
- Removed `ENTRYPOINT ["/sandbox"]` from the "Custom startup scripts" Dockerfile example
- Added a paragraph explaining the base image already sets the correct ENTRYPOINT
- Updated the legacy startup scripts note to recommend `CMD` instead of `ENTRYPOINT ["/sandbox"]`

#### Not changed

The "Using arbitrary base images" section correctly uses `ENTRYPOINT ["/sandbox"]` — in that pattern the binary is explicitly copied to `/sandbox` via `COPY --from=... /container-server/sandbox /sandbox`.

---
*Generated with [OpenCode](https://opencode.ai)*

### DIFF
--- a/src/content/docs/sandbox/configuration/dockerfile.mdx
+++ b/src/content/docs/sandbox/configuration/dockerfile.mdx
@@ -137,9 +137,10 @@ COPY my-app.js /workspace/my-app.js
 COPY startup.sh /workspace/startup.sh
 RUN chmod +x /workspace/startup.sh
 
-ENTRYPOINT ["/sandbox"]
 CMD ["/workspace/startup.sh"]
 ```
+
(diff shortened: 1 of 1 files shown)


elithrar:
/bonk review this and validate the dockerfile commands

Oxyjun (approved):

PR #28648, opened 2026-02-27
Add Warp SSL error documentation

scuffi:
### Summary

Adds information & resolutions for SSL errors during Docker builds with wrangler when using Warp/VPN. We advise to either disable Warp/VPN during build time, or add a small script to your Dockerfile to use the host CA certificate.

### DIFF
--- a/src/content/docs/containers/local-dev.mdx
+++ b/src/content/docs/containers/local-dev.mdx
@@ -83,13 +83,17 @@ If you are running Cloudflare WARP or a VPN that performs TLS inspection, HTTPS 
 To resolve this, you can either:
 
 - Disable WARP or your VPN while running `wrangler dev` or `wrangler deploy`, then re-enable it afterwards.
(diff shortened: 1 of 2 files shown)


ghostwriternr (approved):

src/content/docs/containers/local-dev.mdx
```diff
@@ -75,3 +75,27 @@ This retry logic should be handled for you if you are using the [containers pack
...
+  RUN if [ -n "$SSL_CERT_FILE" ]; then \
+      cp "$SSL_CERT_FILE" /usr/local/share/ca-certificates/Custom_CA.crt && \
+      update-ca-certificates; \
+      fi
+  ```
+
+  :::note
+  The above Dockerfile snippet is an example. Depending on your base image, the commands to install certificates may differ (for example, Alpine uses `apk add ca-certificates` and a different certificate path).
+
+  This snippet will store the certificate into the image. This is only recommended for local development, do not share your certificate or any images that contain it.
```
ghostwriternr:
```suggestion
  This snippet will store the certificate into the image. Depending on whether your production environment needs the certificate, you may choose to do this only during development or use it in production too.
```

scuffi: less opinionated - makes sense 👍

ghostwriternr (approved):

th0m (approved):

PR #28657, opened 2026-02-27
Add Kimi K2.5 via AI Gateway and lilbonk workflow

elithrar:
## Summary

Adds Kimi K2.5 (Workers AI) as a model option via Cloudflare AI Gateway, and a new `/lilbonk` workflow that uses it.

- Add `workers-ai/@cf/moonshotai/kimi-k2.5` to `opencode.jsonc` with model capabilities (256K context, reasoning, vision, tool calling)
- Add `lilbonk.yml` workflow triggered by `/lilbonk` comments, using the same AI Gateway secrets as bonk
- Both bonk and lilbonk share the `Bonk` workflow name so they use the same concurrency group per-PR

cc @mchenco

### DIFF
--- a/.github/workflows/lilbonk.yml
+++ b/.github/workflows/lilbonk.yml
@@ -9,7 +9,7 @@ on:
     types: [submitted]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
+  group: bonk-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
   cancel-in-progress: false
 
 jobs:
--- a/.github/workflows/lilbonk.yml
+++ b/.github/workflows/lilbonk.yml
@@ -1,4 +1,4 @@
-name: Bonk
+name: Lil Bonk
 
 on:
   issue_comment:
--- a/.github/workflows/lilbonk.yml
+++ b/.github/workflows/lilbonk.yml
@@ -9,6 +9,7 @@ on:
     types: [submitted]
 
 concurrency:
+  # intentionally hardcoded so that all Bonk workflows coexist peacefully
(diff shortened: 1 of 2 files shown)


mchenco:
/bonk set parallel tool calls to true in the jsonc

.github/workflows/lilbonk.yml
```diff
@@ -0,0 +1,50 @@
+name: Bonk
```
kodster28: Would be easier for auditing / diagnosing issues if we have this with a unique name

```suggestion
name: Lil Bonk
```

kodster28 (approved):

.github/workflows/lilbonk.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  pull_request_review:
+    types: [submitted]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
```
elithrar:
```suggestion
  group: bonk-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
```

.github/workflows/lilbonk.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  pull_request_review:
+    types: [submitted]
+
+concurrency:
+  group: bonk-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
```
elithrar, line 13:
```suggestion
  # intentionally hardcoded so that all Bonk workflows coexist peacefully
  group: bonk-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
```

PR #28658, opened 2026-02-27
Get cloudflare skills from middlecache

mvvmm:
### Summary

Removes the committed copy of Cloudflare Skills from the repo and replaces it with a fetch-at-build-time approach sourcing from middlecache.

- `bin/fetch-skills.ts` — fetches `skills.tar.gz` from middlecache and extracts into `skills/`. Skips if `skills/` already exists; `--force` overrides. `--soft` mode (used by predev) warns on failure instead of blocking.
- `package.json` — adds `prebuild` and `predev` hooks to run the script automatically
- `skills/` — removed from git, added to `.gitignore`

In dev, fetch runs once and is skipped on subsequent restarts; network failures warn but don't block. In CI, `skills/` is always fetched fresh since it isn't committed — fetch failure fails the build.

The well known skills can be verified here: https://11705275.preview.developers.cloudflare.com/.well-known/skills/index.json


mvvmm:
/bonk please review this PR for any issues

elithrar:
this is great @mvvmm - thank you

cdrubin (changes requested):
Could you make use of this function perhaps: https://github.com/cloudflare/cloudflare-docs/blob/production/src/util/custom-loaders.ts#L20 for grabbing the file from middlecache? 

Also, is there a way to re-use some of the content collection pattern for doing any of this? Or not really useful because all the changes are done before the tarball is generated?

mvvmm:
> Could you make use of this function perhaps: https://github.com/cloudflare/cloudflare-docs/blob/production/src/util/custom-loaders.ts#L20 for grabbing the file from middlecache?

No for a couple of reasons I think. This function saves to `.tmp` but we need these in `/skills/`, and the need for the tarball streaming/extraction. This is also "pre-astro" in the build/dev lifecycle.
 
> Also, is there a way to re-use some of the content collection pattern for doing any of this? Or not really useful because all the changes are done before the tarball is generated?

I think the `astro-skills` is already using the content-collection pattern? See https://github.com/cloudflare/cloudflare-docs/blob/production/src/content.config.ts#L128, but let me know if I misunderstand something here.

mvvmm:
Used `downloadToDotTempIfNotPresent`.

mvvmm:
/bonk mind reviewing this one one more time after my new changes?

kodster28 (approved):

cdrubin (approved):

PR #28717, opened 2026-03-02
fix sql insert syntax

updbqn:
### Summary

"foo@bar.com" is an identifier not a string.

```
→  insert into example_users values ("foo@bar.com");
Error: failed to execute SQL:
SQLite input error: no such column: foo@bar.com (at offset 34)
```

updated to single quotes.

Replaces #28698 which I closed accidentally :sweat_smile:

### DIFF
--- a/src/content/docs/workers/databases/third-party-integrations/turso.mdx
+++ b/src/content/docs/workers/databases/third-party-integrations/turso.mdx
@@ -5,58 +5,39 @@ title: Turso
 
 import { Render, PackageManagers } from "~/components";
 
-[Turso](https://turso.tech/) is an edge-hosted, distributed database based on [libSQL](https://libsql.org/), an open-source fork of SQLite. Turso was designed to minimize query latency for applications where queries comes from anywhere in the world.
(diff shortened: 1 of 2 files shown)


MattieTK:
/bonk update this to also point to the latest instructions for tursodb which involve the CLI instructions here https://github.com/tursodatabase/turso (brew install turso) and the binary `tursodb` not `turso`

updbqn:
> /bonk update this to also point to the latest instructions for tursodb which involve the CLI instructions here https://github.com/tursodatabase/turso (brew install turso) and the binary `tursodb` not `turso`

Verify that this is actually what you want. Turso db and Turso cloud are not the same thing. I don't think the rest of the guide makes much sense in the context of Turso db.

MattieTK:
> > /bonk update this to also point to the latest instructions for tursodb which involve the CLI instructions here [tursodatabase/turso](https://github.com/tursodatabase/turso) (brew install turso) and the binary `tursodb` not `turso`
> 
> Verify that this is actually what you want. Turso db and Turso cloud are not the same thing. I don't think the rest of the guide makes much sense in the context of Turso db.

Understood! Definitely want this to focus on the cloud option, sorry. Fixing this up now to just correct our install options.

*Edit*: I'll just make these changes separately rather than bundling them in with this which is a nice small change!

MattieTK (approved):

PR #28786, opened 2026-03-05
[Sandbox] Add proxyToSandbox() to Ports API reference

whoiskatrin:
## Summary

- Adds a dedicated `proxyToSandbox()` reference entry to the Ports API page (`/sandbox/api/ports/`) under a new `## Module functions` section
- Documents the function signature, parameters, return value, behavior, and includes a `TypeScriptExample` code example
- Adds a note clarifying it is a module-level import (not a `Sandbox` instance method)
- Extends the Related resources section with links to the Expose Services and WebSocket Connections guides

## Context

`proxyToSandbox()` is used in nearly every Sandbox Worker fetch handler but previously had no API reference entry — it only appeared in passing in how-to guides and concept pages. This PR makes it a first-class documented API.

### DIFF
--- a/src/content/docs/sandbox/api/ports.mdx
+++ b/src/content/docs/sandbox/api/ports.mdx
@@ -13,6 +13,50 @@ Preview URLs require a custom domain with wildcard DNS routing in production. Se
 
 Expose services running in your sandbox via public preview URLs. See [Preview URLs concept](/sandbox/concepts/preview-urls/) for details.
 
+## Module functions
+
+### `proxyToSandbox()`
+
+Route incoming HTTP and WebSocket requests to the correct sandboxed container. Call this at the top of your Worker's `fetch` handler, before any application logic, so that preview URL requests are intercepted and forwarded automatically.
+
+```ts
+const response = await proxyToSandbox(request: Request, env: Env): Promise<Response | null>
+```
+
+**Parameters**:
+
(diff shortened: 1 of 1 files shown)


whoiskatrin:
/bonk review and fix the minor issues if needed, make sure the wording is clean and clear

src/content/docs/sandbox/api/ports.mdx
```diff
@@ -13,6 +13,50 @@ Preview URLs require a custom domain with wildcard DNS routing in production. Se
...
 Expose services running in your sandbox via public preview URLs. See [Preview URLs concept](/sandbox/concepts/preview-urls/) for details.
 
+## Module functions
+
+### `proxyToSandbox()`
+
+Route incoming HTTP and WebSocket requests to the correct sandbox container. Call this at the top of your Worker's `fetch` handler, before any application logic, so that it intercepts and forwards preview URL requests automatically.
+
+```ts
+const response = await proxyToSandbox(request: Request, env: Env): Promise<Response | null>
```
aron-cf: Should this be just the interface? The assignment is a bit weird because this code will not run...

```suggestion
proxyToSandbox(request: Request, env: Env): Promise<Response | null>
```

aron-cf (approved):
Looks good to me! 👍

PR #28806, opened 2026-03-06
Adds docs for Worker Previews 

korinne:
### Summary

Draft documentation for Worker Previews 

### To dos

- [ ] Needs actual API docs 
- [ ] Add Terraform resources 
- [ ] Audit / clean up old Preview URLs docs 
- [ ] Add details about `wrangler secret` put for previews
- [ ] Add a nice diagram


src/content/docs/workers/configuration/previews.mdx
```diff
@@ -1,175 +1,14 @@
...
-If you enable or disable Preview URLs in the Cloudflare dashboard, but do not update your Worker's Wrangler file accordingly, the Preview URLs status will change the next time you deploy your Worker with Wrangler.
-:::
-
-## Limitations
-
-- Preview URLs are not generated for Workers that implement a [Durable Object](/durable-objects/).
-- Preview URLs are not currently generated for [Workers for Platforms](/cloudflare-for-platforms/workers-for-platforms/) [user Workers](/cloudflare-for-platforms/workers-for-platforms/how-workers-for-platforms-works/#user-workers). This is a temporary limitation, we are working to remove it.
-- You cannot currently configure Preview URLs to run on a subdomain other than [`workers.dev`](/workers/configuration/routing/workers-dev/).
-- You cannot view logs for Preview URLs today, this includes Workers Logs, Wrangler tail and Logpush.
+This page has moved. For the latest documentation on Previews, refer to [Previews](/workers/previews/).
```
1000hz, line 14: this page is no longer reachable because of the redirect, no? should we just delete this file?

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+---
+
+## Set up Previews
+
+Previews use their own [bindings](/workers/runtime-apis/bindings/), [environment variables](/workers/configuration/environment-variables/), and [secrets](/workers/configuration/secrets/) — separate from your production Worker settings. Before creating your first Preview, you need to configure what resources and environment variables Previews should use. These are called **[Preview Defaults](/workers/previews/api/#preview-defaults)** — a template of starting values that every new Preview inherits automatically.
+
+### From the dashboard
+
+Go to your Worker > **Settings**, and you will see a banner for setting up your default Preview settings. This walks you through the bindings, environment variables, and [observability](#observability-and-logging) settings you want every Preview to start with. This is the fastest way to get your whole team set up — anyone who creates a Preview (manually or by pushing a branch) gets the same configuration without needing to specify it themselves.
```
1000hz: This banner will only appear if the Worker already has production bindings (and will likely be removed or retooled before we launch). Brand new Workers won't require any unique setup flows for Previews; configuring a Worker with Previews settings is just like configuring a Worker with Production settings. In other words, we should just point users to configuring their **Previews** settings from the Settings and Bindings tabs.

src/content/docs/workers/previews/api.mdx
1000hz, line 1: I'm pretty skeptical of the value of this page. The important reference information will already be captured by our actual API reference docs (and OpenAPI schema), and the high-level feature concept explanations don't belong in a page focused on interacting with the API. Topics like how settings provided at deployment time interact with Previews settings probably shouldn't be presented through the narrow lens of the API.

We don't document other APIs in this way. Why Previews? If it's to aid in agent discoverability, I think API reference docs (or just relying on wrangler commands instead of making API requests directly) will suffice.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+	Render,
+	WranglerConfig,
+	Details,
+	PackageManagers,
+	Tabs,
+	TabItem,
+	InlineBadge,
+} from "~/components";
+
+A **Preview** is an isolated copy of your Worker that runs on its own URL, so you can test changes before deploying to production. A Worker can have many Previews running at the same time, each isolated from each other. Each Preview runs with its own [bindings](/workers/runtime-apis/bindings/), [environment variables](/workers/configuration/environment-variables/), and [secrets](/workers/configuration/secrets/), so your preview traffic never touches your production data.
```
1000hz:
```suggestion
A **Preview** is a separate instance of your Worker that runs on its own URL, so you can test changes before deploying to production. Previews run with their own set of [bindings](/workers/runtime-apis/bindings/), [environment variables](/workers/configuration/environment-variables/), and [secrets](/workers/configuration/secrets/), so your preview traffic never touches your production data.
```

1000hz: > A Worker can have many Previews running at the same time, **each isolated from each other**

This is only true insofar as two Previews don't share the same binding configuration. This isn't a claim we can reliably make.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+	args={"my-app -- --framework=astro --git --no-deploy -y"}
+/>
+
+2. Create a feature branch:
+
+```bash
+git checkout -b new-feature
+```
+
+3. Deploy a Preview:
```
1000hz: We should probably include `git push` as an option here (assuming the user has connected this Worker to their git repo in the interim...)

I wonder if a "quick start" focused on Previews is actually valuable. Previews are a means to and end of supporting the overall SDLC of your Workers application, not the destination itself.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+<PackageManagers type="exec" pkg="wrangler" args="preview" />
+
+Wrangler builds your code locally and deploys it as a Preview. It prints a **[Preview URL](#preview-urls-and-deployment-urls)** and a **[Deployment URL](#preview-urls-and-deployment-urls)** to your terminal.
+
+---
+
+## Set up Previews
+
+Previews use their own [bindings](/workers/runtime-apis/bindings/), [environment variables](/workers/configuration/environment-variables/), and [secrets](/workers/configuration/secrets/) — separate from your production Worker settings. Before creating your first Preview, you need to configure what resources and environment variables Previews should use. These are called **[Preview Defaults](/workers/previews/api/#preview-defaults)** — a template of starting values that every new Preview inherits automatically.
```
1000hz: > Before creating your first Preview, you need to configure what resources and environment variables Previews should use.

Maybe, maybe not. This is tantamount to the statement: "Before creating your first Production deployment, you need to configure what resources and environment variables your Worker should use." which is not necessarily true!

I think we should instead frame this idea as Previews having a distinct set of settings from your Worker's Production deployments, and must be configured _separately_. This is slightly different from what's written in that it doesn't assume a Worker (and thus its Previews) require configuration; whether it does or not is application-dependent.

And maybe this ends up being self-evident enough to not warrant a whole `## Set up Previews` section at all. We could keep just the `### From code` bits and move it to a wrangler-focused section.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+vars = { ENVIRONMENT = "preview" }
+
+[[previews.kv_namespaces]]
+binding = "MY_KV"
+id = "preview-kv-id"
+```
+
+</WranglerConfig>
+
+When you configure settings in both the dashboard and the Wrangler configuration file, values from your Wrangler file take precedence for that deployment. Settings configured in the dashboard still apply to anything not overridden — including Previews created by teammates or by automated builds triggered from Git pushes.
```
1000hz:
```suggestion
These settings are scoped to deployments created with `wrangler preview`. When you configure settings in both the dashboard and the Wrangler configuration file, values from your Wrangler file take precedence. To apply these settings to all future Previews, run `wrangler preview settings update` to persist them to your Worker's Previews settings.
```

1000hz: > Settings configured in the dashboard still apply to anything not overridden — including Previews created by teammates or by automated builds triggered from Git pushes.

We have to be careful here because the automated builds currently rely on calling `wrangler preview` to handle preview deployments (at least for the time being), and so settings configured declaratively in `previews` will most likely still apply during automated builds.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+### Automate with Workers Builds
+
+Connect a GitHub or GitLab repository, and Cloudflare deploys a Preview for you every time you push to a branch.
+
+1. Go to **Workers & Pages** > your Worker > **Settings** > **Builds**.
+2. Connect your repository.
+3. Set your tracked branch (usually `main`).
+
+When you push to `main`, Cloudflare builds and deploys your Worker to production using your top-level Wrangler configuration — your production bindings, environment variables, and secrets. When you push to any other branch, Cloudflare deploys a Preview using your Preview settings instead.
```
1000hz:
```suggestion
When you push to `main`, Cloudflare builds and deploys your Worker to production using your top-level Wrangler configuration — your production bindings, environment variables, and secrets. When you push to any other branch, Cloudflare deploys a Preview using your _Previews_ settings instead.
```

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+The first push to a branch creates a new Preview. Each subsequent push creates a new **[Preview Deployment](/workers/previews/api/#preview-deployment)** within that Preview — an immutable snapshot of your code at that point. This means you can always go back and inspect or compare any previous deployment, even after pushing new changes.
```
1000hz:
```suggestion
Every branch gets its own Preview, and each push to a branch creates a new deployment — an immutable snapshot of your code at that point. This means you can always go back and inspect or compare any previous deployment, even after pushing new changes.
```

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+If a Preview with that name does not exist yet, Wrangler creates it. If it already exists, Wrangler creates a new deployment within it. Each time you run `wrangler preview`, Wrangler prints the Preview URL and Deployment URL to your terminal.
+
+Use this for ad-hoc testing, external CI/CD pipelines, or when you are not using a Git-based workflow.
+
+---
+
+## Preview URLs and Deployment URLs
+
+When you deploy a Preview, Cloudflare generates two URLs. For the full URL patterns across `workers.dev`, `cloudflare.app`, and custom domains, refer to [URL structure](/workers/previews/api/#url-structure).
```
1000hz: > For the full URL patterns across `workers.dev`, `cloudflare.app`, and custom domains, refer to [URL structure](/workers/previews/api/#url-structure).

The URL _patterns_ aren't particularly relevant.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+  https://f498jo-my-worker.my-account.workers.dev
+  ```
+
+Workers Builds posts both URLs as a comment on your pull request. `wrangler preview` prints both to your terminal.
+
+---
+
+## Enable or disable Preview URLs
+
+You control whether Preview URLs are active through the `previews_enabled` setting on your Worker's subdomain configuration. By default, `previews_enabled` follows the state of your `workers.dev` subdomain — if you disable your `workers.dev` route, Preview URLs are also disabled unless you explicitly enable them.
```
1000hz: This also affects Deployment URLs.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+You can serve Previews from your own **custom domain** instead of (or in addition to) `workers.dev`. When you enable Previews on a custom domain, Cloudflare creates a wildcard DNS record and adds a wildcard SAN to the domain's certificate to cover all Preview subdomains. For more details, refer to [Custom domains](/workers/previews/api/#custom-domains).
+
+| URL type   | Example                             |
+| ---------- | ----------------------------------- |
+| Production | `app.example.com`                   |
+| Preview    | `my-feature.app.example.com`        |
+| Deployment | `f498jo-my-feature.app.example.com` |
+
+To enable Previews on a custom domain via the API, set `previews_enabled` to `true`:
```
1000hz: Feels out-of-place to show how to do this via API.

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+</WranglerConfig>
+
+You can also disable production routing on a custom domain while keeping Previews active by setting `production_enabled` to `false` — this serves only Preview traffic on that domain.
+
+---
+
+## Secure access to Previews
+
+To restrict who can access your Preview URLs, you can enable [Cloudflare Access](/cloudflare-one/access-controls/policies/) with a single click in your Worker's settings. This protects both Preview URLs and Deployment URLs.
```
1000hz:
```suggestion
To restrict who can access your Previews, you can enable [Cloudflare Access](/cloudflare-one/access-controls/policies/) with a single click in your Worker's settings. This protects both Preview URLs and Deployment URLs.
```

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+1. Go to **Workers & Pages** > your Worker > **Settings** > **Domains & Routes**.
+2. For Preview URLs, select **Enable Cloudflare Access**.
```
1000hz:
```suggestion
2. Select **Enable Cloudflare Access** within the actions menu of **Preview URLs**.
```

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+3. Optionally, select **Manage Cloudflare Access** to configure which users, groups, or email addresses are authorized.
+
+For more information on configuring access policies, refer to [Access policies](/cloudflare-one/access-controls/policies/).
+
+---
+
+## Observability and logging
+
+Each Preview runs its own observability and logging settings, independent of your production Worker. This means you can:
```
1000hz:
```suggestion
Each Preview has its own observability and logging settings, independent of your production Worker. This means you can:
```

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+- Set full sampling (`head_sampling_rate: 1`) on a Preview to capture every request during testing, without increasing log volume on your production Worker.
+- Enable [Logpush](/workers/observability/logpush/) to send a Preview's logs to your own storage for debugging.
+- Attach [Tail Workers](/workers/observability/logs/tail-workers/) to a Preview to pipe its logs into a custom analysis pipeline.
+
+Configure these in the dashboard under **Workers & Pages** > your Worker > **Settings** > **Preview Defaults**, and Cloudflare applies them to every Preview created for that Worker. For details on how observability settings are structured, refer to the [Preview object](/workers/previews/api/#preview-object).
```
1000hz:
```suggestion
Configure observability settings for all Previews in the dashboard under **Workers & Pages** > your Worker > **Settings** > **Observability**. You can also configure these settings for an individual Preview under **Workers & Pages** > your Worker > **Previews** > your Preview > **Settings**.
```

src/content/docs/workers/previews/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+<PackageManagers type="exec" pkg="wrangler" args="preview" />
+
+By default, Wrangler names the Preview after your current Git branch. You can also provide a name explicitly:
+
+<PackageManagers type="exec" pkg="wrangler" args="preview --name my-feature" />
+
+If a Preview with that name does not exist yet, Wrangler creates it. If it already exists, Wrangler creates a new deployment within it. Each time you run `wrangler preview`, Wrangler prints the Preview URL and Deployment URL to your terminal.
+
+Use this for ad-hoc testing, external CI/CD pipelines, or when you are not using a Git-based workflow.
```
1000hz:
```suggestion
Run `wrangler preview` from your project directory to build your code locally and deploy it as a Preview. Use this for ad-hoc testing, external CI/CD pipelines, or when you are not using a Git-based workflow.

<PackageManagers type="exec" pkg="wrangler" args="preview" />

By default, Wrangler names the Preview after your current Git branch. You can also provide a name explicitly:

<PackageManagers type="exec" pkg="wrangler" args="preview --name my-feature" />
```

PR #28828, opened 2026-03-06
[DO] Update rules-of-durable-objects.mdx

luishcardoso:
Hello team,

### Summary

For further info about Try/Catch statements.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

Thanks.


Oxyjun:
Hi @luishcardoso , thanks for this PR! In this particular instance, I don't think we'd need an external link to explain try/catch, as I would assume most readers of this page will be familiar with the term.

I'll discuss internally and update the page if we feel it's needed!

Thank you all the same for opening the PR 😄

PR #28840, opened 2026-03-06
Code Quality fix within the Sidebar

ahaywood:
#### Problem
The sidebar search initialization function was being called twice on page load when `document.readyState !== "loading"`:

1. Once in the conditional else block (line 294)
2. Once unconditionally immediately after (line 298)

This caused duplicate event listeners to be attached for:

- Search input events
- Keyboard shortcuts (Escape and "/" focus)
- Global search link clicks

#### Solution
Removed the unconditional `initSidebarSearch()` call at line 298. The conditional block already handles both scenarios correctly:

- If DOM is still loading → waits for DOMContentLoaded event
- If DOM is already loaded → initializes immediately

#### Impact
- Prevents duplicate event listener registration
- Eliminates potential for multiple tracking events per user action
- Reduces unnecessary function calls on page load

### DIFF
--- a/src/components/overrides/Sidebar.astro
+++ b/src/components/overrides/Sidebar.astro
@@ -293,9 +293,6 @@ const [product, module] = Astro.url.pathname.split("/").filter(Boolean);
 	} else {
 		initSidebarSearch();
 	}
-
-	// Re-initialize on navigation (for SPA-like behavior)
-	initSidebarSearch();
 </script>
 
 <style is:global>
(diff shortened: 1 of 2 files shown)


kodster28:
Ah, think this one might also need to be rebased

mvvmm:
/bonk does this PR still fix the issue it outlines in it's description and achieved in this commit: eacd049eeeb103f767dd1641e0448b5ac4accbab after the recent rebase changes and conflicts resolutions?

kodster28 (approved):

PR #28870, opened 2026-03-09
Match sidenav ui state after search navigation

mvvmm:
### Summary

When using the search functionality in the sidenav to navigate between pages, the state of sidenav wouldn't update to match to the newly navigated page. Specifically, the sidenav folders would still be open to the previous page, not the new page.

The fix here involves clearing starlight's persisted `sl-sidebar-state` session storage on search navigation to stop starlight from overriding it with the most recently persisted state (the _filtered_ state during the search).

#### Before
https://github.com/user-attachments/assets/e80f443c-a2ee-40a3-b971-8aae185d4bfd

#### After
https://github.com/user-attachments/assets/1256e4b7-54b7-4b05-ad2b-d76a762f7f80

### DIFF
--- a/src/components/overrides/Sidebar.astro
+++ b/src/components/overrides/Sidebar.astro
@@ -262,6 +262,22 @@ const [product, module] = Astro.url.pathname.split("/").filter(Boolean);
 			}
 		});
 
+		// On link click during search, clear persisted open[] so the next page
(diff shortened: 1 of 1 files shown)


mvvmm:
/bonk Please review this PR for any regressions and code quality issues

kodster28 (approved):

PR #28883, opened 2026-03-10
[Cloudflare One] DNS Policies ELI5

Oxyjun:
Improves clarity across the 4 pages in `/cloudflare-one/traffic-policies/dns-policies/` by adding inline definitions, spelling out acronyms, and surfacing non-obvious behavior. Generated via ELI5 analysis of all sections.

**`index.mdx`**
- Add a plain-language opening paragraph explaining what DNS policies are and why they matter
- Link to Gateway overview on first mention
- Define the three evaluation phases (Before/During/After DNS resolution) that appear in every selector table but were never explained
- Rewrite the Allow action description to clarify its purpose as an exception mechanism, not a redundant pass-through
- Fix the grammatically broken Override action description ("respond to all DNS queries for a given domain to another destination" → "replace the real DNS answer with a destination you specify")
- Add inline DNSSEC definition before the validation toggle
- Annotate `A`/`AAAA` record types with (IPv4)/(IPv6) in the block page section
- Spell out EDNS, DoH, DoT on first use
- Clarify the authoritative nameserver selector (remove redundant "IP address of the IP address")
- Rewrite fallback DNS limitation to explain the bypass mechanism

**`common-policies.mdx`**
- Spell out EAR, OFAC, ITAR acronyms on first use
- Add EDNS definition in the dynamic categories section
- Define TLDs inline ("the last segment of a domain name, such as `.com` or `.ru`")
- Replace "egress to the origin server" with "when Gateway connects to the destination server"
- Annotate Force IPv4/IPv6 with the record types being blocked

**`test-dns-filtering.mdx`**
- Expand prerequisites to distinguish WARP vs DNS-only deployments
- Define `dig` and `nslookup` tools for readers unfamiliar with them
- Correct "answer section" to "header line" for where `REFUSED`/`NOERROR` appear in `dig` output
- Add inline definitions for `REFUSED` and `NOERROR` DNS response codes
- Define EDNS client subnet before the EDNS test section
- Spell out DoH on first use

**`timed-policies.mdx`**
- Add use-case examples to the opening paragraph (work-hours blocking, maintenance windows)
- Add a `:::caution` callout for the non-obvious duration timer behavior (absolute end time, not a pausable countdown)
- Add a `:::note` callout for VPN/proxy timezone inference pitfall
- Fix "we will fall back" → "it will fall back" (consistent third-person voice)


src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
```diff
@@ -83,15 +83,15 @@ API value: `allow`
 
 </Details>
 
-Policies with Allow actions allow DNS queries to reach destinations you specify within the Selector and Value fields. For example, the following configuration allows DNS queries to reach domains we categorize as belonging to the Education content category:
+Policies with Allow actions explicitly permit DNS queries to resolve, even if a lower-precedence policy would block them. Use Allow to create exceptions to broader Block policies. For example, the following configuration allows DNS queries to reach domains categorized as belonging to the Education content category:
```
Oxyjun: Need factcheck here.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
```diff
@@ -285,7 +285,13 @@ This setup ensures users will be blocked from accessing offensive sites using DN
...
 ## Selectors
 
-Gateway matches DNS queries against the following selectors, or criteria:
+Gateway matches DNS queries against the following selectors, or criteria.
+
+Each selector is evaluated during a specific phase of the DNS resolution process:
+
+- **Before DNS resolution** — Gateway inspects properties of the incoming query (for example, the domain name or source IP) before looking up the answer.
+- **During DNS resolution** — Gateway inspects information discovered while resolving the query (for example, the authoritative nameserver IP).
+- **After DNS resolution** — Gateway inspects the DNS answer (for example, the resolved IP or CNAME record) after resolution completes. These selectors cannot be used with the Override action, because the override must be applied before the answer is returned.
```
Oxyjun: Need factcheck.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
```diff
@@ -297,7 +303,7 @@ Gateway matches DNS queries against the following selectors, or criteria:
 
 ### Authoritative Nameserver IP
 
-Use this selector to match against the IP address of the authoritative nameserver IP address.
+Use this selector to match against the IP address of the authoritative nameserver — the DNS server that holds the definitive records for a domain.
```
Oxyjun: This should be removed, adds clutter.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/test-dns-filtering.mdx
```diff
@@ -116,9 +114,9 @@ Once you have configured your Gateway policy to block the category, the test dom
 
 ## Test EDNS configuration
 
-If you [enabled EDNS client subnet](/cloudflare-one/networks/resolvers-and-proxies/dns/locations/) for your DNS location, you can validate EDNS as follows:
+EDNS client subnet is a DNS extension that sends a portion of the user's IP address (a `/24` subnet) to the upstream DNS resolver, allowing it to return geographically optimal answers. If you [enabled EDNS client subnet](/cloudflare-one/networks/resolvers-and-proxies/dns/locations/) for your DNS location, you can validate it as follows:
```
Oxyjun: Need to factcheck.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
Oxyjun: **Factcheck result: corrected.**

The original claim — "even if a lower-precedence policy would block them" — was oversimplified. The Cloudflare docs ([Order of enforcement](/cloudflare-one/traffic-policies/order-of-enforcement/#dns-policies)) confirm a **first-match principle**, but also state that evaluation phase (before/during/after DNS resolution) takes precedence over numerical policy order. A pre-resolution Block at precedence 2 can override a post-resolution Allow at precedence 1.

**Fix applied:** Replaced the unqualified precedence claim with a reference to the first-match principle and linked to the order of enforcement page. The rewrite no longer implies precedence is the only factor.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
Oxyjun: **Factcheck result: partially incorrect, corrected.**

The three evaluation phases (Before/During/After DNS resolution) are correct — they match the exact labels used in every selector table on this page and in the [Order of enforcement](/cloudflare-one/traffic-policies/order-of-enforcement/#dns-policies) docs.

However, the Override restriction was too narrow. I wrote "These selectors cannot be used with the Override action" only on the "After DNS resolution" bullet, but the Override action's own selector list note says: **"The Override action cannot be used with selectors evaluated during or after DNS resolution."** Both "during" and "after" are excluded.

**Fix applied:** Moved the Override restriction out of the "After" bullet into its own sentence that correctly covers both phases, and added a cross-link to the order of enforcement page.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
Oxyjun: **Reverted.** The inline definition has been removed. The selector description now matches the original text.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/test-dns-filtering.mdx
Oxyjun: **Factcheck result: two issues found, both corrected.**

1. **"/24" is Cloudflare-specific, not a protocol requirement.** RFC 7871 defines a configurable `SOURCE PREFIX-LENGTH` field — the prefix can be any length. The [Gateway locations docs](https://developers.cloudflare.com/cloudflare-one/networks/resolvers-and-proxies/dns/locations/) confirm: "Cloudflare enables EDNS in a privacy preserving way by not sending the user's exact IP address but rather the first `/24` range." The `/24` is Cloudflare's implementation choice.

2. **"upstream DNS resolver" is wrong.** Per RFC 7871, ECS is sent from the recursive resolver to **authoritative DNS nameservers**, not to a generic "upstream resolver."

**Fix applied:** Changed "upstream DNS resolver" → "authoritative DNS nameservers" and clarified that the `/24` prefix is Cloudflare's specific behavior for privacy.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/common-policies.mdx
```diff
@@ -10,7 +10,7 @@ head:
 
 import { Render, Tabs, TabItem, APIRequest } from "~/components";
 
-The following policies are commonly used to secure DNS traffic.
+The following policies are commonly used to secure DNS traffic. Each example includes both Dashboard and API instructions that you can adapt for your organization.
```
marciocloudflare: Any particular reason you capitalized Dashboard?

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
```diff
@@ -353,7 +363,7 @@ Use this selector to filter DNS responses by their `TXT` records.
 | ---------------------- | ------------------------------------------- | -------------------- |
 | DNS TXT Response Value | `any(dns.response.txt[*] in {"your_text"})` | After DNS resolution |
 
-### DoH Subdomain
+### DoH Subdomain (DNS over HTTPS)
```
marciocloudflare, line 366: Any pages pointed over to the old header? (double checking potential future broken links)

marciocloudflare (approved):

src/content/docs/cloudflare-one/traffic-policies/dns-policies/common-policies.mdx
Oxyjun: Nope - missed that, thanks!

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
Oxyjun: Good shout - thankfully none for this particular heading.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/index.mdx
```diff
@@ -83,15 +85,15 @@ API value: `allow`
 
 </Details>
 
-Policies with Allow actions allow DNS queries to reach destinations you specify within the Selector and Value fields. For example, the following configuration allows DNS queries to reach domains we categorize as belonging to the Education content category:
+Policies with Allow actions explicitly permit DNS queries to resolve. Because Gateway uses a [first-match principle](/cloudflare-one/traffic-policies/order-of-enforcement/#order-of-precedence), an Allow policy stops evaluation for matching queries, preventing lower-precedence policies from blocking them. For example, the following configuration allows DNS queries to reach domains categorized as belonging to the Education content category:
```
alexamavrogianis: Minor thing, but "an Allow policy stops evaluation" may be slightly misleading. It only stops evaluation _if it is higher in precedence_ than other policies. It is not related to the action itself.

Maybe: "Gateway uses a first-match principle. This means that if an Allow policy for a matching query is set with a higher precedence than a block policy, the query will be allowed to resolve."

src/content/docs/cloudflare-one/traffic-policies/dns-policies/common-policies.mdx
```diff
@@ -10,7 +10,7 @@ head:
 
 import { Render, Tabs, TabItem, APIRequest } from "~/components";
 
-The following policies are commonly used to secure DNS traffic.
+The following policies are commonly used to secure DNS traffic. Each example includes both dashboard and API instructions that you can adapt for your organization.
```
alexamavrogianis: Can we add something to the intro that specifically calls out Cloudflare/Gateway to help with AI retrieval?

src/content/docs/cloudflare-one/traffic-policies/dns-policies/test-dns-filtering.mdx
```diff
@@ -13,19 +13,19 @@ This section covers how to validate your Gateway DNS configuration.
 
 ## Prerequisites
```
alexamavrogianis, line 14: The skill didn't make any edits to the intro sentence "This section covers how to validate your Gateway DNS configuration." but I think it's worth adding some more context/why it is helpful to test.

src/content/docs/cloudflare-one/traffic-policies/dns-policies/test-dns-filtering.mdx
```diff
@@ -13,19 +13,19 @@ This section covers how to validate your Gateway DNS configuration.
...
 
-Before you start, make sure you are connected to a network that is associated with the [DNS location](/cloudflare-one/networks/resolvers-and-proxies/dns/locations/) where the policy is applied.
+Before you start, make sure your device is sending DNS queries through a [DNS location](/cloudflare-one/networks/resolvers-and-proxies/dns/locations/) where the policy is active. If your device uses WARP, DNS queries route through Gateway automatically. If you are using a DNS-only deployment (without WARP), verify that your network's DNS resolver points to your Gateway DNS location's IP address.
```
alexamavrogianis: This can be framed more as an either/or, goal is to make sure the device is sending DNS queries to Gateway. It can be through a DNS location or through WARP.

PR #28886, opened 2026-03-10
[Workers] Split up wrangler page into subpages

kodster28:
### Summary

Move Wrangler commands into separate subpages where possible (around products) to break up the overall page. The current version is actually so large that it's crashing folks' browsers and rendering the info inaccessible.

This also aligns with [best practices for agents](https://dacharycarey.com/2026/02/18/agent-friendly-docs/), who have much more limited context windows than LLM crawlers.


kodster28:
> This PR requires additional review attention because it affects the following areas:
> 
> ### Redirects
> This PR changes current filenames or deletes current files. Make sure you have [redirects](https://developers.cloudflare.com/pages/configuration/redirects/) set up to cover the following paths:
> 
> * [x]  `/workers/wrangler/commands/`

This is incorrect b/c our check isn't smart enough to understand page -> folder conversions.

Oxyjun (approved):

src/content/docs/workers/wrangler/commands/index.mdx
```diff
@@ -0,0 +1,49 @@
...
+  order: 3
+---
+
+import { DirectoryListing, PackageManagers, Render } from "~/components";
+
+[Wrangler](/workers/wrangler/) offers a number of commands to manage your Cloudflare Workers.
+
+## Commands
+
+<DirectoryListing />
```
emily-shen, line 18: Might be a huge pain to update links sorry, but what do you think about including the full list of commands here that used to be at the top of the page? Since 'general commands' is a bit vague, it might help with finding stuff

kodster28: I think we can add something to the backlog to evaluate a programmatic fix here.

That list was always a mix of incomplete / inaccurate (and a pain to update), so think we can come back to evaluating it from a different approach after this gets merged in.

src/content/docs/workers/wrangler/commands/general.mdx
emily-shen, line 1: the cert and mtls cert commands being in here is a bit random, they're their own bindings and/or used with hyperdrive so not sure where exactly to put them...

emily-shen, line 1: What about splitting out 'Worker commands' (dev, deploy, secrets, versions) from 'system' commands like auth stuff?

emily-shen (commented):
thanks for jumping on this so quickly!

src/content/docs/workers/wrangler/commands/general.mdx
kodster28: Ah, gotcha. I think we could even do that as a new page too.

kodster28: Think this should be a fast follow.

Not familiar enough with those commands to make that distinction myself.

PR #28910, opened 2026-03-11
[Cloudflare One] Network policies ELI5

Oxyjun:
Improves readability and reduces jargon across all four network policies pages (`/cloudflare-one/traffic-policies/network-policies/`), based on an ELI5 clarity analysis.

- **Rewrite introduction** (`index.mdx`): Replace dense paragraph with concrete language — "Network policies control TCP and UDP traffic" with examples (SSH, RDP, database connections). Separates identity-based capability into its own paragraph. Fixes misleading "traffic leaving your endpoints" phrasing.
- **Add caution for Network Override silent failure** (`index.mdx`): The override destination rewrite happens even when the target IP is unreachable, with no log entry. This operational surprise is now a `:::caution` aside instead of buried prose.
- **Define SNI** (`index.mdx`): Replace "The host whose SNI header Gateway will filter traffic against" with a plain explanation of what SNI is and how Gateway uses it.
- **Add CIDR example** (`index.mdx`): Expand the CIDR notation note with a concrete example (`10.0.0.0/8` matches `10.0.0.0` to `10.255.255.255`).
- **Clarify Allow action** (`index.mdx`): Explain that Allow is meaningful in a default-block configuration where it defines exceptions.
- **Explain policy evaluation order** (`common-policies.mdx`): Add note that policies are evaluated top-to-bottom, first match wins. The two-policy patterns on this page depend on this but never stated it.
- **Explain `filters: ["l4"]`** (`common-policies.mdx`): Add note in the API tab explaining that `l4` means this is a network (Layer 4) policy.
- **Surface TLS pass-through security behavior** (`protocol-detection.mdx`): Non-HTTP protocols inside TLS bypass network policy filtering. Previously a dense paragraph — now a `:::caution` aside with a descriptive title.
- **Strengthen SSH logging deprecation** (`ssh-logging.mdx`): Upgrade from `:::note` to `:::caution` with explicit "Legacy feature" title and "deprecated" language.
- **Fix outdated UI navigation path** (`ssh-logging.mdx`): Remove "Firewall policies" from nav path — current UI uses "Network" directly.

### DIFF
--- a/src/content/docs/cloudflare-one/traffic-policies/network-policies/ssh-logging.mdx
+++ b/src/content/docs/cloudflare-one/traffic-policies/network-policies/ssh-logging.mdx
@@ -61,9 +61,9 @@ cat /etc/ssh/sshd_config
 
 ## 7. Create an Audit SSH policy
 
-1. In [Cloudflare One](https://one.dash.cloudflare.com), go to **Traffic policies** > **Network**.
+1. In [Cloudflare One](https://one.dash.cloudflare.com), go to **Traffic policies** > **Firewall policies**.
 
-2. In the **Network** tab, create a new network policy.
+2. In the **Network** tab, select **Add a network policy**.
 
 3. Name the policy and specify the [Destination IP](/cloudflare-one/traffic-policies/network-policies/#destination-ip) for your origin server.
 
--- a/src/content/docs/cloudflare-one/traffic-policies/network-policies/ssh-logging.mdx
+++ b/src/content/docs/cloudflare-one/traffic-policies/network-policies/ssh-logging.mdx
@@ -10,8 +10,8 @@ sidebar:
 
 import { Render } from "~/components";
 
-:::note
-Not recommended for new deployments. We recommend using [Access for Infrastructure](/cloudflare-one/networks/connectors/cloudflare-tunnel/use-cases/ssh/ssh-infrastructure-access/) to manage SSH sessions and log SSH commands.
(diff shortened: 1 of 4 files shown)


src/content/docs/cloudflare-one/traffic-policies/network-policies/ssh-logging.mdx
```diff
@@ -61,7 +61,7 @@ cat /etc/ssh/sshd_config
 
 ## 7. Create an Audit SSH policy
 
-1. In [Cloudflare One](https://one.dash.cloudflare.com), go to **Traffic policies** > **Network** > **Firewall policies**.
+1. In [Cloudflare One](https://one.dash.cloudflare.com), go to **Traffic policies** > **Network**.
```
Oxyjun: What is the actual path for the dashboard?

marciocloudflare (approved):

src/content/docs/cloudflare-one/traffic-policies/network-policies/ssh-logging.mdx
Oxyjun: Good catch. The correct path is **Traffic policies** > **Firewall policies**, then selecting the **Network** tab — matching the canonical partial in `src/content/partials/cloudflare-one/gateway/get-started/create-network-policy.mdx`.

The original had `**Traffic policies** > **Network** > **Firewall policies**` which was also wrong. Fixed in f6bb5ea.

PR #28921, opened 2026-03-11
Use Markdown for Agents for all docs markdown generation

mvvmm:
## Summary

This PR replaces the custom rehype-based HTML-to-Markdown pipeline with [Cloudflare's Markdown for Agents (MFA)](https://developers.cloudflare.com/fundamentals/reference/markdown-for-agents/) network-layer feature, which converts any HTML page to Markdown when requested with an `Accept: text/markdown` header.

The `/path/index.md` URL pattern is preserved for backwards compatibility. Requests to `/path/index.md` are handled by a Cloudflare Transform Rule that rewrites the URL to `/path/` and a Request Header Transform Rule that adds `Accept: text/markdown` — triggering MFA to return Markdown instead of HTML.

### What changed

**Removed the custom Markdown pipeline:**

- Deleted `src/util/markdown.ts` — the custom `htmlToMarkdown()` function and rehype pipeline
- Deleted `src/content/docs/style-guide/fixtures/` — test fixture pages used by the old pipeline
- Removed the `index.md` handler from `worker/index.ts` and `src/middleware/index.ts`
- Removed associated Worker tests

**Updated documentation:**

- `src/content/docs/style-guide/how-we-docs/ai-consumability.mdx` — updated the "Turning our components into Markdownable HTML" section to describe MFA instead of the rehype plugin
- `src/content/partials/style-guide/llms-txt.mdx` — updated to reflect MFA as the underlying mechanism, with `Accept: text/markdown` as the primary recommendation

**Updated `llms.txt` generation:**

- Replaced `src/pages/[product]/llms.txt.ts` with `src/pages/[...product]/llms.txt.ts` — a catch-all route that derives each product's llms.txt URL from `entry.data.entry.url` instead of the YAML filename (`entry.id`). This fixes broken and mismatched llms.txt URLs for products whose canonical URL path differs from their entry ID (e.g. `cloudflare-for-saas` → `/cloudflare-for-platforms/cloudflare-for-saas/llms.txt`).
- Root `llms.txt` now links to each product using its canonical URL (`entry.data.entry.url`) rather than `/${entry.id}/`, fixing the same class of broken links.
- Root `llms.txt` no longer lists sub-products that are fully covered by their parent's llms.txt (e.g. Cache Rules, Logpush, Access), eliminating ~58 duplicate entries and preventing the same pages from appearing twice when enumerating from root. Sub-product llms.txt files still exist at their canonical URLs.
- Root `llms.txt` adds a new **## Other** section listing ungrouped directory entries (e.g. `style-guide`, `security`, `use-cases`) that have docs pages but no product group.
- Per-product llms.txt files group pages into sections using `sidebar.order` from each section's `index.mdx` frontmatter, matching the visible sidebar structure. Products without sidebar ordering fall back to a flat list.
- Page links in per-product llms.txt files now point directly to `index.md` URLs (e.g. `/workers/get-started/guide/index.md`) so agents and parsers receive Markdown immediately without needing to set headers or rewrite URLs.
- Sidebar links to `llms.txt` and `llms-full.txt` in the Developer Platform section are updated to use canonical URL paths.
- The per-product blockquote is simplified — since links already point to Markdown, the two-method explanation is removed and replaced with a single note that links are Markdown, plus the `llms-full.txt` pointer.


mvvmm:
/bonk please review this PR for any regressions or issues

kodster28 (approved):

mvvmm:
/bonk please review this PR for any regressions or issues otherwise

PR #28941, opened 2026-03-12
Add Sandbox request proxying guide

### LINKED ISSUE #368: Document proxy pattern for authenticating clients inside sandbox
Currently available as an example at [`examples/authentication`](https://github.com/cloudflare/sandbox-sdk/tree/main/examples/authentication)

scuffi:
### Summary

We already have an example of how we would go about proxying requests from Sandbox -> Worker to protect credentials [here](https://github.com/cloudflare/sandbox-sdk/tree/main/examples/authentication). We're adding a specific guide for this functionality to make it more discoverable & help explain the concept & usage.

### Documentation checklist

Closes: https://github.com/cloudflare/sandbox-sdk/issues/368

### DIFF
--- a/src/content/docs/sandbox/concepts/security.mdx
+++ b/src/content/docs/sandbox/concepts/security.mdx
@@ -135,6 +135,20 @@ try {
 }
 ```
 
+## Proxying external API requests
+
+Passing credentials directly to a sandbox — via environment variables or files — means the sandbox process holds a live credential that any code running inside it can read. A Worker proxy removes that exposure by keeping credentials exclusively in the Worker and giving the sandbox a short-lived JWT instead.
+
+The flow works as follows:
+
+```
+Sandbox (short-lived JWT) → Worker proxy (validates JWT, injects real credentials) → External API
+```
+
+The sandbox never sees the real credential. If the JWT is compromised, it expires after a short window and cannot be reused.
+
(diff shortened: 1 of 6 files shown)


aron-cf (approved):
Docs look great from my perspective. 

I have two pieces of feedback on the implementation:

1. Having to copy files out of an example in the SDK kind of sucks. Can we publish these files to a subdirectory of the `@cloudflare/sandbox` sdk? 

        import { createProxy } from "@cloudflare/sandbox/proxy";

 2. Is there any way to do this without exposing the proxy to the public internet? It's not a huge deal if we're going to replace this with better networking controls very soon but it would be nice if the proxy was internal.

scuffi:
> Docs look great from my perspective.
> 
> I have two pieces of feedback on the implementation:
> 
> 1. Having to copy files out of an example in the SDK kind of sucks. Can we publish these files to a subdirectory of the `@cloudflare/sandbox` sdk?
>    ```
>     import { createProxy } from "@cloudflare/sandbox/proxy";
>    ```
> 2. Is there any way to do this without exposing the proxy to the public internet? It's not a huge deal if we're going to replace this with better networking controls very soon but it would be nice if the proxy was internal.

I agree, this would be nicer as first class support but as we're going to have better control of this soon I don't think it's worth the time/effort as it'll have a short shelf life until it is replaced. These docs just outline the temporary solution to show it is possible, and hopefully will be replaced with nicer, first class support documentation soon

PR #28947, opened 2026-03-12
[Cloudflare One] Traffic Policies Root-Level ELI5

Oxyjun:
Applies ELI5 (Explain Like I'm 5) clarity analysis to all 11 root-level pages in `src/content/docs/cloudflare-one/traffic-policies/`. The goal is to reduce jargon barriers and missing context for IT admins and security engineers who are new to Cloudflare-specific concepts.

- **index.mdx** — Expand "GRE" acronym, state egress default behavior, link WARP client and Gateway logs
- **identity-selectors.mdx** — Add value framing intro, reorder stale-identity explanation (base-behavior-first), simplify selector descriptions, add SAML Attribute inline definition, add Generic OIDC workaround note
- **resolver-policies.mdx** — Add problem statement intro, explain IP overlap with virtual networks link, split LDF paragraph, expand "Magic tunnel", add `:::caution` for QUIC/UDP requirement
- **proxy.mdx** — Add purpose clause to intro, link "Traffic and DNS mode" to WARP modes, fix "systcl" → "sysctl" typo, link TLS decryption
- **order-of-enforcement.mdx** — Add framing sentence explaining why enforcement order matters, link Cloudflare Gateway
- **enable-ids.mdx** — Replace duplicated intro with concise summary + links, fix grammar "an Cloudflare" → "a Cloudflare", standardize Logpush links, fix `${managed_ruleset_id}` → `<MANAGED_RULESET_ID>` placeholder, fix phase name inconsistency, add Rulesets Engine preamble, restructure disable instructions
- **application-app-types.mdx** — Restructure hostnames vs support hostnames with explicit Block/Allow behavioral rules, add real-world Google Drive example, add TLS decryption problem lead-in, add SIP/XMPP functional descriptions
- **troubleshoot-gateway.mdx** — Add Split Tunnel and Logpush cross-links, distinguish "Do Not Scan" from "Do Not Inspect", fix "still block" → "still blocks" grammar, add resolver policy cross-links, explain why resolver policies preferred over Local Domain Fallback
- **domain-categories.mdx** — Make security risks vs threats distinction explicit (suspicious vs confirmed malicious), add DNS/HTTP policy and Gateway logs cross-links, expand CIPA acronym, state default filtering behavior, note "Security Risks" name collision, link DGA and DNS tunneling definitions
- **managed-service-providers.mdx** — Replace "group or individual account level" with parent/child terminology, add "When used with Gateway" qualifier, add tiered-vs-siloed comparison preview, explain `allow_child_bypass` default behavior, fix "messaing" → "messaging" typo
- **global-policies.mdx** — Define what global policies are and why they exist, add `allow` and `resolve` action definitions, add Network and HTTP inspection policy intro paragraphs with action definitions (`bypass`, `allow`, `block`, `noisolate`, `noscan`)


src/content/docs/cloudflare-one/traffic-policies/application-app-types.mdx
```diff
@@ -58,23 +58,25 @@ Gateway sorts applications into the following app type groups:
...
 | Video Streaming & Editing                      | Applications used for streaming and editing video                                                                           |
 | [Do Not Inspect](#do-not-inspect-applications) | Applications incompatible with the TLS certificate required by the [Gateway proxy](/cloudflare-one/traffic-policies/proxy/) |
 
 ## Application hostnames
 
-Applications categorized by Cloudflare may independently rely on a number of different internal and external resources to provide functionality. To enable effective behavior of Allow and Block Gateway policies, Cloudflare One separates application definitions into [hostnames](#hostnames) and [support hostnames](#support-hostnames).
+An application like Google Drive uses its own domains (for example, `drive.google.com`) and shared resources (for example, `accounts.google.com` for login). Gateway separates these into [hostnames](#hostnames) and [support hostnames](#support-hostnames) so that blocking one application does not accidentally break another.
```
Oxyjun: While I agree that this is clearer, do we want to keep using real URLs? Or would an illustrative example be better?

I actually think using real examples make this clearer than illustrative, but worried about maintenance + not sure how I feel about consistently using Google as an example. Maybe we should use CF sites?

src/content/docs/cloudflare-one/traffic-policies/domain-categories.mdx
```diff
@@ -31,7 +31,7 @@ Subdomains that have not been assigned a category will inherit the category of t
...
 | Compromised Domain            | Sites where a legitimate domain has been compromised or taken over and had malicious content planted or injected.                                                                                                     |
 | Cryptomining                  | Sites that mine cryptocurrency by taking over the user's computing resources.                                                                                                                                         |
-| DGA Domains                   | Domains detected as generated by algorithms seen in malware.                                                                                                                                                          |
+| DGA Domains                   | Domains generated programmatically by Domain Generation Algorithms (DGA) associated with malware. These algorithmically created domain names change frequently, making them harder to block individually.             |
```
Oxyjun, line 34: This has been verified by adversarial review step. Can we double check its accuracy?

src/content/docs/cloudflare-one/traffic-policies/global-policies.mdx
```diff
@@ -5,7 +5,7 @@ sidebar:
   order: 8
 ---
 
-Cloudflare Zero Trust applies a set of global policies to all accounts.
+Cloudflare Zero Trust applies a set of global policies to all accounts. These policies prevent you from accidentally blocking Cloudflare services that Zero Trust depends on, such as the dashboard, API, and client registration.
```
Oxyjun, line 8: Is this correct?

src/content/docs/cloudflare-one/traffic-policies/identity-selectors.mdx
```diff
@@ -9,13 +9,19 @@ import { Render } from "~/components";
 
 With Cloudflare One, you can create Secure Web Gateway policies that filter outbound traffic down to the user identity level. To do that, you can build DNS, HTTP or Network policies using a set of [identity-based selectors](#identity-based-selectors). These selectors require you to deploy the Cloudflare One Client in [Traffic and DNS mode](/cloudflare-one/team-and-resources/devices/cloudflare-one-client/configure/modes/).
 
+For example, you can create different security rules for different teams — block social media for contractors but allow it for marketing.
```
Oxyjun, line 12: Interesting that this example was added purely from scratch, without rewording anything.

src/content/docs/cloudflare-one/traffic-policies/proxy.mdx
```diff
@@ -7,13 +7,13 @@ sidebar:
...
 
 The Gateway proxy is required for filtering HTTP and network traffic via the Cloudflare One Client in Traffic and DNS mode. To proxy HTTP traffic without deploying the Cloudflare One Client, you can configure [PAC files](/cloudflare-one/networks/resolvers-and-proxies/proxy-endpoints/) on your devices.
 
 ## Proxy algorithm
 
-Gateway uses the [Happy Eyeballs algorithm](https://datatracker.ietf.org/doc/html/rfc6555) to proxy traffic in the following order:
+Gateway uses the [Happy Eyeballs algorithm](https://datatracker.ietf.org/doc/html/rfc6555), which tries IPv4 and IPv6 connections with a staggered fallback and uses whichever address family responds first, to proxy traffic in the following order:
```
Oxyjun, line 16: This passed vetting from adversarial reviewer stage. Need to double check.

src/content/docs/cloudflare-one/traffic-policies/application-app-types.mdx
caley-b: I do think it's much clearer. However, I'm not too worried about Google changing domains too often. It's a real-world example of the type of application and resource that will help it click more readily for customers. I'm also not opposed to using something like drive.example.com and accounts.example.com to illustrate the point.

src/content/docs/cloudflare-one/traffic-policies/identity-selectors.mdx
caley-b: That is interesting! Makes me wonder what it would say if we asked why (after making it clear that we're not mad at it or providing feedback, etc.).

src/content/docs/cloudflare-one/traffic-policies/index.mdx
```diff
@@ -33,15 +33,15 @@ Use DNS policies to block malicious domains, restrict content categories, or pre
...
+**[Network policies](/cloudflare-one/traffic-policies/network-policies/)** inspect individual TCP, UDP, and Generic Routing Encapsulation (GRE) packets. They can match on IP addresses, ports, protocols, and the server name sent at the start of an encrypted connection (Server Name Indication, or SNI).
 
 Use network policies to block access to specific ports or non-HTTP services such as SSH and RDP.
 
 </Details>
 
 <Details header="HTTP policies">
 
-**[HTTP policies](/cloudflare-one/traffic-policies/http-policies/)** inspect the full content of web requests — including URLs, headers, and uploaded or downloaded files. Gateway decrypts HTTPS traffic so it can examine what DNS and network policies cannot see.
+**[HTTP policies](/cloudflare-one/traffic-policies/http-policies/)** inspect the full content of web requests — including URLs, headers, and uploaded or downloaded files. Gateway decrypts HTTPS traffic so it can examine what DNS and network policies cannot see. This requires installing a [Cloudflare root certificate](/cloudflare-one/team-and-resources/devices/user-side-certificates/) on user devices.
```
caley-b, line 44: Have you noticed many instances where ELI5 proactively calls out requirements to clarify what needs to be done, where maybe we haven't in the past?

caley-b (approved):

src/content/docs/cloudflare-one/traffic-policies/application-app-types.mdx
```diff
@@ -13,7 +13,7 @@ import {
 	Render,
 } from "~/components";
 
-Gateway allows you to create DNS, Network, and HTTP policies based on applications and app types. You can select individual applications or groups of app types to filter specific traffic on your network.
+Gateway allows you to create DNS, Network, and HTTP policies based on applications and app types. Instead of writing separate rules for every domain an application uses, you can select the application by name. You can select individual applications or groups of app types to filter specific traffic on your network.
```
alexamavrogianis: A few pre-existing-to-these-changes-nits and general thoughts

- I'm not sure why we abbreviate 'app type' but otherwise spell out application. Feels like it should be spelled out in both instances. Second, "groups of app types" (and similarly in the table description for App 
- Types that says "app type groups" feels a little repetitive. The type is essentially a group.
- I think it would be helpful to add a super short sentence up front that a single application may have more than one domain. Right now it reads as if we assume users would know that blocking primevideo.com may not fully block Amazon Prime. We touch on this later in the doc when speaking to hostnames and support hostnames, so don't need anything to in depth. Should probably use "hostname" instead of "domain" for consistency here, too.

src/content/docs/cloudflare-one/traffic-policies/application-app-types.mdx
alexamavrogianis: Agree on the general point about the domains being helpful. However, the part about "so that blocking one application does not accidentally break another" isn't maybe a little misleading. Blocking support hostnames may actually have unexpected behavior for other apps so we only use them in Allow policies. The benefit is really about controlling behavior for this specific application.

src/content/docs/cloudflare-one/traffic-policies/domain-categories.mdx
alexamavrogianis: I believe it is accurate.

src/content/docs/cloudflare-one/traffic-policies/global-policies.mdx
alexamavrogianis: Yes

src/content/docs/cloudflare-one/traffic-policies/index.mdx
alexamavrogianis: It is a good addition by ELI5!

src/content/docs/cloudflare-one/traffic-policies/proxy.mdx
alexamavrogianis: Believe it is correct

PR #28975, opened 2026-03-13
Only build once in PR workflows: Step 1

mvvmm:
## Summary

**_Note:_** _We're adding back the compiles job, to meet the repo requirements of having a compiles job. So currently, this still builds twice. We will change the repo rules after merging this PR to only require build and validate. Then we will add another PR to remove the compiles job._

Previously, opening a PR triggered two separate full site builds: one in `ci.yml` (for checks) and one in `publish-preview.yml` (for preview deployment). This PR consolidates them into a single build.

## What changed

**Deleted `publish-preview.yml`** — the separate workflow that built the site on every push to non-production branches is gone.

**Restructured `ci.yml`** into four focused jobs:

- **`build`** — runs the full site build, uploads `dist/` as a GitHub Actions artifact. Handles link-check failures gracefully: if the build produced output but exited non-zero (link validation failure), the job still passes and sets an output flag so downstream jobs can report it without blocking the preview deploy.
- **`validate`** — downloads the artifact and runs ESLint, format checks, redirect validation, and tests. No rebuild needed.
- **`publish-preview`** — downloads the artifact and deploys to Cloudflare Workers preview. Replaces the build+deploy that `publish-preview.yml` used to do. The computed `BRANCH_SLUG` is passed as a step output to the `post-preview-url-comment` script, so branch slug logic is no longer duplicated between shell and TypeScript.
- **`notify`** — posts or removes the CI failure comment on the PR.

**Updated bin scripts:**

- `bin/post-pr-ci-failure-comment/index.ts` — looks for `Build` and `Validate` jobs instead of the old `Compiles` job name.
- `bin/post-preview-url-comment/index.ts` — reads `BRANCH_SLUG` from the environment (set by the deploy step) instead of computing it independently. Supports both `pull_request` and `push` event contexts for the PR number lookup.
- `bin/post-preview-url-comment/util.ts` — removed `branchToSubdomain` since the slug is now computed once in the shell deploy step and passed through.
- `bin/post-preview-url-comment/index.node.test.ts` — removed `branchToSubdomain` tests.

## Testing

_**Note:**_ _Ignore the `Compiles` job in these screenshots, it would need to be removed_

#### With invalid link, but otherwise successful build
Screenshot 2026-03-13 at 11 35 34 AM
Screenshot 2026-03-13 at 11 35 29 AM

#### With actual failing build
Screenshot 2026-03-13 at 3 25 00 PM
Screenshot 2026-03-13 at 3 43 27 PM
Screenshot 2026-03-13 at 3 43 37 PM


src/content/docs/r2/get-started/index.mdx
```diff
@@ -13,6 +13,7 @@ Cloudflare R2 Storage allows developers to store large amounts of unstructured d
 ## Before you begin
 
 You need a Cloudflare account with an R2 subscription. If you do not have one:
+
```
kodster28:
```suggestion
```

kodster28 (approved):

PR #28978, opened 2026-03-13
[Network Flow] ELI5 work #3

marciocloudflare:
### Summary

PCX-21054

### DIFF
--- a/src/content/partials/networking-services/mnm/faqs.mdx
+++ b/src/content/partials/networking-services/mnm/faqs.mdx
@@ -2,7 +2,7 @@
 {}
 ---
 
-Find answers to the most commonly asked questions. If you cannot find the answer you are looking for, refer to the [community page](https://community.cloudflare.com/) to explore more resources.
+If you cannot find your answer here, refer to the [community page](https://community.cloudflare.com/) for more resources.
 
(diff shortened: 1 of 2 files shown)


src/content/partials/networking-services/mnm/faqs.mdx
```diff
@@ -46,14 +49,16 @@ For example, you create a Network Flow rule with the following parameters:
 "duration": "1m0s"
 ```
 
-With this rule, your network needs to receive a throughput greater than 50,000,000 bits per second (50 Gigabits per second or Gbps) for 60 seconds. If both of these conditions are met, then Network Flow will send you an alert.
+With this rule, your network needs to receive a throughput greater than 50,000,000 bits per second (50 Megabits per second or Mbps) for 60 seconds. If both of these conditions are met, then Network Flow will send you an alert.
```
caley-b, line 52: Just flagging this one since the measurement changed. I think it makes sense to me, but flagging to verify  since I'm not an SME.

caley-b (approved):
Only one thing to check before pushing.

src/content/partials/networking-services/mnm/faqs.mdx
marciocloudflare: I think we're fine because this is just an example of how to setup the data, we're not saying this is the correct value

PR #28995, opened 2026-03-14
[Support Feedback] Improve certificate pinning, cipher suite, and Universal SSL docs (SSL/TLS)

dmmulroy:
## Summary

Improves SSL/TLS documentation for common support topics, based on an audit of active support macros.

### Changes

- **Certificate pinning**: Clarify that pinning to Cloudflare edge certificates is not supported and document alternatives (pin to origin, use Authenticated Origin Pulls)
- **Cipher suite customization**: Document prerequisites and plan requirements for custom cipher suites
- **Universal SSL subdomain limitations**: Clarify coverage depth and SAN limits for Universal SSL certificates

### Context

These changes are driven by recurring support cases. If you want to see the underlying support data (macro frequency, case volume by topic), reach out to @dmmulroy internally.

### DIFF
--- a/src/content/docs/ssl/edge-certificates/additional-options/cipher-suites/customize-cipher-suites/index.mdx
+++ b/src/content/docs/ssl/edge-certificates/additional-options/cipher-suites/customize-cipher-suites/index.mdx
@@ -13,13 +13,23 @@ import { Render, TabItem, Tabs, DirectoryListing } from "~/components";
 
 With an [Advanced Certificate Manager](/ssl/edge-certificates/advanced-certificate-manager/) subscription, you can restrict connections between Cloudflare and clients — such as your visitor's browser — to specific [cipher suites](/ssl/edge-certificates/additional-options/cipher-suites/).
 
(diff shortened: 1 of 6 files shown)


elithrar:
/bonk review this and suggest any fixes

RebeccaTamachiro:
Thanks for the PR, @dmmulroy 👍 I think the majority of it looks good but flagging to @ngayerie for an extra set of eyes.

src/content/docs/ssl/edge-certificates/additional-options/cipher-suites/customize-cipher-suites/index.mdx
RebeccaTamachiro, line 1: I'm just not really convinced about these changes to `/customize-cipher-suites/index.mdx`: the reason is that having CF4SaaS is not really a prerequisite for this guide in particular, and we actually point SaaS providers wanting to customize ciphers for their custom hostnames away from this page. So for this, let's maybe:
1- keep the previous version, with the shorter intro and separate note for SaaS providers 
2- add notes about the ACM prereq to the specific how-to pages nested under this, instead

What do you think?

RebeccaTamachiro: Applied the changes. Checked in with Nic via chat.

RebeccaTamachiro (approved):

PR #29087, opened 2026-03-18
[SSL, CF1] Update GPG key name

RebeccaTamachiro:
### Summary

PCX-21326
https://pkg.cloudflare.com/index.html accounts for the new key (cloudflare-main.gpg) but the change had not been reflected on the dev docs


ranbel:
@nikitacano, fyi the tunnel installation command in the dashboard still uses the `-public-v2` key.

ranbel (approved):

PR #29113, opened 2026-03-19
GitHub Actions workflows for automated branch cleanup

mvvmm:
### Summary
The number of published branches we have is causing checkout to be a bit slow, which gets ultra exemplified when github workers are being slow: https://github.com/cloudflare/cloudflare-docs/actions/runs/23264658721/job/67645493487

#### Workflows:
- Add `delete-closed-pr-branches.yml` to delete head branches immediately when PRs close
- Add `delete-stale-branches.yml` to run daily cleanup of closed PR branches and unassociated branches older than 30 days
- Both workflows respect protected branches and skip fork branches
- Support dry-run mode via workflow_dispatch for safe testing
- Limit to 100 deletions per run to avoid rate limits


mvvmm:
/bonk do you see any consequences to these github action workflows?

kodster28 (commented):
What do we get here that we don't get from https://github.com/marketplace/actions/stale-branches?

kodster28 (approved):

PR #29114, opened 2026-03-19
chore(tanstack-start): Update main path in WranglerConfig to src/server.ts to match example handler

Friss:
### Summary

The example handler is at `src/server.ts` but the config was pointing to `app/server.ts` which was confusing and wrong. It should point to the `src` file. 

### Documentation checklist


MattieTK:
Thanks for noticing and fixing this!

MattieTK (approved):

PR #29129, opened 2026-03-19
[Chore] Optimize info call in WranglerNamespace

kodster28:
### Summary

We ignore the exported `commands` in WranglerNamespace and instead re-call Wrangler to grab those commands.

Might save a bit of duration on build times.
- https://github.com/cloudflare/cloudflare-docs/blob/production/src/util/wrangler.ts#L3
- https://github.com/cloudflare/cloudflare-docs/blob/production/src/components/WranglerCommand.astro#L8

### DIFF
--- a/src/components/WranglerNamespace.astro
+++ b/src/components/WranglerNamespace.astro
@@ -1,6 +1,6 @@
 ---
 import { z } from "astro/zod";
-import { experimental_getWranglerCommands } from "wrangler";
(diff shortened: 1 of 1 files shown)


kodster28:
/bonk can you find any issues with this PR?

kodster28:
/bigbonk do you agree with the review already left on this PR?

haleycode (approved):

PR #29147, opened 2026-03-19
Rerun failed publish-production jobs

mvvmm:
# Summary

Adds automatic retry logic to the production publish workflow to reduce noise from transient deployment failures.

- **`publish-production.yml`**: Adds a `publish-retry` job that runs automatically when the `publish` job fails. The Google Chat alert has been moved from the initial job to the retry job, so alerts only fire if both attempts fail.

In both cases: first failure is silent, second failure sends the alert.


mvvmm:
Not sure how I feel about having a second action the duplicates the first one...

PR #29163, opened 2026-03-20
Add L4 stats docs for Rulesets and Workers

jannes:
### Summary

Adds documentation for new Rulesets fields and their corresponding Workers properties.

### DIFF
--- a/src/content/docs/workers/runtime-apis/request.mdx
+++ b/src/content/docs/workers/runtime-apis/request.mdx
@@ -227,6 +227,14 @@ All plans have access to:
 
   * If Cloudflare replaces the value of the `Accept-Encoding` header, the original value is stored in the `clientAcceptEncoding` property, for example, `"gzip, deflate, br"`.
 
+* `clientQuicRtt` number | undefined
+
+  * The smoothed round-trip time (RTT) between Cloudflare and the client for QUIC connections, in milliseconds. Only present when the client connected over QUIC (HTTP/3). For example, `42`.
+
+* `clientTcpRtt` number | undefined
+
+  * The smoothed round-trip time (RTT) between the client and Cloudflare for TCP connections, in milliseconds. Only present when the client connected over TCP (HTTP/1 and HTTP/2). For example, `22`.
+
 * `colo` string
 
   * The three-letter [`IATA`](https://en.wikipedia.org/wiki/IATA_airport_code) airport code of the data center that the request hit, for example, `"DFW"`.
@@ -234,6 +242,12 @@ All plans have access to:
 * `country` string | null
 
   * Country of the incoming request. The two-letter country code in the request. This is the same value as that provided in the `CF-IPCountry` header, for example, `"US"`.
+
+* `edgeL4` Object | undefined
+
+  * Layer 4 transport statistics for the connection between the client and Cloudflare. Contains the following property:
+    * `deliveryRate` number - The most recent data delivery rate estimate for the connection, in bytes per second. For example, `123456`.
+
 
 * `isEUCountry` string | null
 
--- a/src/content/fields/index.yaml
+++ b/src/content/fields/index.yaml
@@ -898,13 +898,47 @@ entries:
     example_value: |-
       true
 
+  - name: cf.edge.l4.delivery_rate
+    data_type: Integer
+    categories: [Request]
+    keywords:
+      [request, l4, delivery rate, bandwidth, network, performance, transport]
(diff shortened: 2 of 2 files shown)


src/content/fields/index.yaml
```diff
@@ -898,6 +898,32 @@ entries:
...
+      true
+
+  - name: cf.edge.l4.delivery_rate
+    data_type: Integer
+    categories: [Request]
+    keywords:
+      [request, l4, delivery rate, bandwidth, network, performance, transport]
+    summary: The most recent data delivery rate estimate for the client connection, in bytes per second.
+    description: |-
+      Reports the transport layer delivery rate estimate measured by Cloudflare for the connection between the client and Cloudflare, in bytes per second.
```
pedrosousa: The final description in the field reference page will also include the summary as the first paragraph, so we may want to rephrase this description so that doesn't repeat what's in the summary.

src/content/changelog/rules/2026-03-20-l4-transport-telemetry-fields.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+Three new fields are now available in the Ruleset Engine that expose layer 4 transport telemetry from the connection between the client and Cloudflare. These fields let you build rules that respond to connection quality and protocol type.
+
+---
+
+### New fields
+
+| Field                             | Type    | Description                                                                                                                                   |
+| --------------------------------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| `cf.edge.client_quic`             | Boolean | Indicates whether the request was made over QUIC (HTTP/3).                                                                                    |
```
marmeladema: this one is planned but does not exist yet

jannes: removed

src/content/fields/index.yaml
jannes: thanks, i removed the redundant first sentence in description.

src/content/fields/index.yaml
```diff
@@ -898,14 +898,45 @@ entries:
...
       true
 
+  - name: cf.edge.l4.delivery_rate
+    data_type: Integer
+    categories: [Request]
+    keywords:
+      [request, l4, delivery rate, bandwidth, network, performance, transport]
+    summary: The most recent data delivery rate estimate for the client connection, in bytes per second.
+    description: |-
+      This metric reflects the rate at which data is being successfully delivered over the connection. It can be used to build rules that respond to connection quality, such as serving lower-quality content to clients on slow connections.
```
LPardue: Do we really need examples? There's nothing special about this field compared to others we might add in future so I'd prefer deleting examples, otherwise we'd have to add it to everything and bloat docs

```suggestion
      This metric reflects the rate at which data is being successfully delivered over the connection.
```

src/content/docs/workers/runtime-apis/request.mdx
```diff
@@ -227,6 +227,14 @@ All plans have access to:
 
   * If Cloudflare replaces the value of the `Accept-Encoding` header, the original value is stored in the `clientAcceptEncoding` property, for example, `"gzip, deflate, br"`.
 
+* `clientQuicRtt` number | undefined
+
+  * The smoothed round-trip time (RTT) between the client and Cloudflare for QUIC connections, in milliseconds. Only present when the client connected over QUIC (HTTP/3). For example, `42`.
```
LPardue: note that the RTT is measured at the server, so its maybe better to state it as 

```suggestion
  * The smoothed round-trip time (RTT) between Cloudflare and the client for QUIC connections, in milliseconds. Only present when the client connected over QUIC (HTTP/3). For example, `42`.
```

src/content/fields/index.yaml
```diff
@@ -898,14 +898,45 @@ entries:
...
+
   - name: cf.timings.client_tcp_rtt_msec
     data_type: Number
     categories: [Request]
     keywords: [request, timing, tcp, rtt, performance, latency]
     summary: The smoothed TCP round-trip time (RTT) from client to Cloudflare in milliseconds.
+    description: |-
+      This field is only populated for TCP (HTTP/1, HTTP/2) connections. For QUIC connections, the value is `0`.
     example_value: |-
       20
```
LPardue, line 925: nit: you're missing an example_block, not sure if that's intended

jannes: added

LPardue (approved):

marmeladema (approved):

pedrosousa (approved):

PR #29169, opened 2026-03-20
Prevent CLRF line endings with pre-commit hook

mvvmm:
### Summary

- Pre-commit hook that forces LF line endings
- CI check stolen from #27186


kodster28 (approved):

mvvmm:
/bonk do you see any issues with this?

PR #29194, opened 2026-03-20
[Chore] tech cleanup

kodster28:
With the launch of [Cloudflare Agent](https://developers.cloudflare.com/cloudflare-agent/), think we're finally to the point where we should remove `/workers/ai/` from our repo.

Similar to `/support/ai/` (which we removed about a month ago), `/workers/ai/` is no longer the best place to get AI-supported advice about Cloudflare. Instead, we should be directing folks to the Cloudflare Agent experience, which offers better advice as well as knowing more about you contextually.


public/__redirects
```diff
@@ -2025,6 +2025,7 @@
 /workers/tutorials/store-data-with-fauna/ https://fauna.com/blog/the-future-of-fauna 301
 /workers/tutorials/live-cursors-with-nextjs-rpc-do/ /resources/ 301
 /workers/tutorials/automated-analytics-reporting/ /resources/ 301
+/workers/ai/ /workers/ 301
```
kodster28:
```suggestion
/workers/ai/ /cloudflare-agent/ 301
```

ahaywood (approved):

PR #29247, opened 2026-03-23
Disable auto-formatter in OpenCode.

kentonv:
OpenCode tries to "helpfully" run prettier after every change an agent makes.

But somehow it does it in a way that actually breaks formatting.

Here is where OpenCode actually broke the formatting of a page: https://github.com/cloudflare/cloudflare-docs/pull/28343#discussion_r2976024973

That change got merged and the formatting on that page remains broken today, a month later. Nobody noticed.

Running `prettier` by hand does not cause this. It's only when OpenCode runs it, for some reason.

This change disables the feature entirely because it is causing much more harm than good.

### DIFF
--- a/opencode.jsonc
+++ b/opencode.jsonc
@@ -1,6 +1,15 @@
 {
 	"$schema": "https://opencode.ai/config.json",
 	"permission": "allow",
+
+	// OpenCode tries to auto-run prettier whenever the agent makes an edit, but it seems to run it
+	// in the wrong mode, formatting mdx files as if they were jsx files, causing breaking changes
+	// to the formatting, like adding newlines inside `<code>` elements that actually get rendered
+	// as line breaks. This has caused actual broken formatting in production because people don't
+	// notice it happening. Let's disable it. When you run prettier manually (not through OpenCode),
+	// it doesn't cause problems.
+	"formatter": false,
+
 	"experimental": {
 		"continue_loop_on_deny": true,
 	},


pedrosousa:
> // When you run prettier manually (not through OpenCode),
> // it doesn't cause problems.

Actually, I haven't found any different behavior between what AI agents do and what VS Code/Windsurf+prettier extension does. In the [functions reference](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/ruleset-engine/rules-language/functions.mdx?plain=1#L49) we've used `prettier-ignore` for this kind of blocks starting with `<code>` to prevent `prettier` from touching them (it would always handle this type of content like it would handle JSX).

Are you able to correctly format MDX files such as `src/content/docs/workers/runtime-apis/bindings/worker-loader.mdx` locally using `prettier`?

kentonv:
@pedrosousa I linked to an example where OpenCode broke the formatting of a `<code>` tag. I find that OpenCode re-introduces this corruption any time it edits any part of the file.

When I ran `prettier` directly on the command line, it did not seem to cause the same problem.

I don't have the mentioned vscode extension installed so no idea what happens there.

kentonv:
Can someone approve this please?

While working on #29639 I had to specifically fix all the `<code>` blocks after the auto-formatter messed them up. Poor Claude went into a loop of trying to fix them itself, only to have the formatter automatically break them again each time it tried to fix them.

mvvmm:
I can confirm that prettier is breaking this via windsurf, via the cli, and via opencode. So it's not specific to opencode.  I'll look today at how we might be able to fix this

https://github.com/user-attachments/assets/dafa8a50-d660-4726-b576-935848bacd58

mvvmm:
okay so tldr

This is quite the rabbit hole. Basically mdx injects `<p>` tags into the output html when there are newlines in between jsx elements: https://github.com/mdx-js/mdx/issues/1798. This is intended behavior. Our problem obviously is that prettier formats jsx in a way where it forces elements onto new lines in these cases. 

We can't just disable formatting completely, as that causes an even bigger headache. There may be a solution where we don't allow html elements like `<code>` that break when `<p>` tags are injected into them. There may be a solution where we build custom mdx components to wrap inline elements like this, and don't break. There may be a solution where we just find a better formatter for `.mdx` files. Either way, it requires some more thought. I'll add it to our backlog.

For now, the best solution I can offer you in this case is to just use `{/* prettier-ignore */}`

kentonv:
Hi @mvvmm,

There are many instances of `<code>` and other JSX in this repo, and they are being corrupted whenever someone auto-formats the respective files.

If you feel that adding `prettier-ignore` comments is the right way to address them, then please go ahead and do that. If you don't have time to implement that or some other solution right now, please accept my change to stop the bleeding.

mvvmm:
Closing for the reasons mentioned above. Fix is here: #29770

@kentonv let me know if this misses any of the breakages you're aware of.

PR #29248, opened 2026-03-23
[ZT] WARP Client --> Cloudflare One Client in changelog

ranbel:
Render [changelog ](https://developers.cloudflare.com/cloudflare-one/changelog/cloudflare-one-client/)title and prefix based on client version #. Old entries remain "WARP client". Newer entries (starting with the next GA) will use "Cloudflare One Client".

Example from test environment:
image

image


kodster28:
/bigbonk review this PR. Is there a simpler way to implement this behavior?

kodster28 (approved):

PR #29255, opened 2026-03-23
Documents all Container class methods on own page

mikenomitch:
### Summary

Adding all Container class methods - we were relying on GitHub docs and should have in our own.


mikenomitch:
/bonk review and suggest (only) changes in-line

mikenomitch:
/bonk make the following changes

- The documentation structure should more closely resemble /sandbox/api/commands in that the code example's should include example usage, and the Parameters and Return Values should be documented similarly (not in a code block)
- Make it clear that onActivityExpired needs to call await this.stop() or await this.destroy() or it will not sleep the container. Add an example of usage that does this.
- Add a code example showing setting all of the class properties at the top
- In general, add example usage when it makes sense. Probably to most methods.

th0m (approved):

korinne (approved):
LGTM

kristianfreeman (approved):

PR #29263, opened 2026-03-24
chore: bump npm version

MohamedH1998:
### Summary

Npm 11 was released with Node 24, makes sense to update them together

### Documentation checklist


mvvmm (approved):

package.json
```diff
@@ -149,8 +149,8 @@
 		"wrangler": "4.75.0"
 	},
 	"engines": {
-		"npm": "10.x",
+		"npm": "11.x",
```
colbywhite: If you're going to pin it in `engines`, I would think it would need to be the same pin that's in `packageManager` i.e. `11.12.0`. Otherwise you could have `11.3.0` and satisfy one constraint, but not the other.

MohamedH1998: Updated @colbywhite

colbywhite (changes requested):
After some thought on this, I think the root problem is that we want people to use `npm`, not `yarn`/`pnpm`.  But the only way I knew how to do that was to pin to a specific npm version. But I don't think we care _that_ much about the version of npm you're using; just use npm. 

With that in mind, here's what I propose.

- Pin the node version in engines.
    - Pin it to what, though? I don't think we actually care that it's 24.14.0 vs 24.13.1. (I'm not even sure if we care about 24 vs 22. But I can see there being a reason for that.) So, either ">=24" or "^24" is the pin that better matches what we actually care about. 
    - I'll cede to you on which pin it should be. Just pin to a major in some fashion.
    - The `engine-strict=true` in `.npmrc` would enforce that.
- Be looser on the npm version in engines. Or maybe don't pin it at all.
    - I dont think we really care about the version of npm. Just that it is npm. So not pinning at all seems reasonable. But since npm@11 is what is bundled with node@24, it also seems reasonable to pin both to at least a major version. So if we pin, it should likely just be the major version. So either ">=X" or "^X" where X is the major version bundled with the pinned node major version.
    - I'll cede to you on whether or not this needs a pin. If it is pinned, it should just be at the major level
    - If we pin to anything, the `engine-strict=true` in `.npmrc` would enforce that.
- Do not use packageManager to pin npm
    - This strategy requires pinning the full version when we don't really care. This puts us in the boat of keeping up with very specific version numbers that likely don't matter for this project.
    - I only used this strategy the first time because I didn't fully realize there were other options.
- Use [only-allow](https://github.com/pnpm/only-allow#readme) to enforce npm (but not the specific version of npm).
    - This is likely what I should've done from the get go but didn't know that this existed in such a mainstream way (i.e. it's an official tool owned by pnpm). 
- Sync up `.node-version` and `.nvmrc` with whatever pin is in engines.
    - `.node-version` helps volta users; `.nvmrc` helps nvm users
    - These are just helpers for the most part. Nothing is going to enforce them per se. Just best practice
- Use node version pins to power CIs
    - Instead of hardcoding the `node-version`, point [actions/setup-node](https://github.com/actions/setup-node/tree/v6?tab=readme-ov-file#usage) to a `node-version-file`. That way we don't have different specifications in different files. 
    - Note: CIs is plural. This needs to be done in multiple files
- Remove the cache logic around `npm ci` in CIs
    - We currently don't run `npm ci` if there's a cache hit. We should run it no matter what. But when there's a cache hit, the command will just run faster. This decouples the command from any potential bugs in our cache logic.
    - Note: This is separate from the original intent of this PR but is a related convo that makes sense to solve as part of this PR.
    - Note: CIs is plural. This needs to be done in multiple files

colbywhite:
After more reading on the [diff between](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) `engines` and `devEngines`, we should be doing all of this in `devEngines`.  This means we could get rid of the `engine-strict=true` in `.npmrc`. 

So I'd amend the proposal to use `devEngines` to pin major node version and (optionally) the major npm version. And then remove the bullet point about using `only-allow` altogether since devEngines gets the job done without a dependency. 


Note: `devEngines` didn't exist in npm@9. I think I knew it got added but this was first time I actually read the docs on it. I probably have a lot of projects that I should move to `devEngines` 😬 .

PR #29272, opened 2026-03-24
[Cloudflare One] ELI5 on DNS Policies

Oxyjun:
Improves readability and accuracy across all four pages in the DNS policies folder (`src/content/docs/cloudflare-one/traffic-policies/dns-policies/`), based on an ELI5 accessibility review.

- **Add "why" framing to `index.mdx`** — New opening paragraph explains what DNS policies do and why they matter before diving into how Gateway processes queries. Clarifies that 1.1.1.1 is Cloudflare's public DNS resolver.
- **Define "evaluation phase" in `index.mdx`** — Every selector table references this concept but it was never defined. Added a three-bullet explanation before the Selectors section covering Before/During/After DNS resolution, including why Override cannot use post-resolution selectors.
- **Clarify A/AAAA record behavior in `index.mdx`** — Adds inline context that `A` records are IPv4 and `AAAA` are IPv6, and explains that `0.0.0.0`/`::` are non-routable.
- **Define EDNS inline in `index.mdx` and `common-policies.mdx`** — Expands "EDNS" to "Extension Mechanisms for DNS" on first use and adds a sentence explaining what it does and who sends the data.
- **Fix regex bug in `common-policies.mdx`** — The phishing policy regex `.sso.*` was missing the leading `*` wildcard. Changed to `.*sso.*` in both the Dashboard table and the API traffic expression to match domains containing "sso" anywhere in the name.
- **Improve `test-dns-filtering.mdx` prerequisites** — Replaces vague "connected to a network" language with specific guidance for WARP vs. DNS-only deployments. Adds context for `dig`/`nslookup` tools and explains what `REFUSED` and block page IPs mean in `dig` output.
- **Define EDNS client subnet in `test-dns-filtering.mdx`** — Adds one-sentence definition of ECS before the validation steps.
- **Add use-case framing to `timed-policies.mdx`** — Opening paragraph now includes example scenarios (work-hours restrictions, temporary access, weekday/weekend differences).
- **Fix first-person voice in `timed-policies.mdx`** — Changes "we will fall back" to "it falls back" per Cloudflare docs style. Adds a note about IP geolocation inaccuracy on VPNs.
- **Clarify duration behavior in `timed-policies.mdx`** — Rephrases "retain their duration" to make clear the expiration clock does not pause when a policy is temporarily turned off.

### DIFF
--- a/src/content/docs/cloudflare-one/traffic-policies/dns-policies/common-policies.mdx
+++ b/src/content/docs/cloudflare-one/traffic-policies/dns-policies/common-policies.mdx
@@ -69,7 +69,7 @@ The categories included in this policy are not always a security threat, but blo
 
 ## Block a dynamic list of categories
 
-You can add a list of category IDs to the [EDNS header](https://datatracker.ietf.org/doc/html/rfc6891) of a request sent to Gateway as a JSON object using OPT code `65050`. For example:
+You can add a list of category IDs to the [EDNS (Extension Mechanisms for DNS)](https://datatracker.ietf.org/doc/html/rfc6891) header of a request sent to Gateway as a JSON object using OPT code `65050`. A client application or custom DNS resolver embeds these IDs in the DNS query so Gateway can filter by categories not known at policy creation time. For example:
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


Oxyjun:
Closing as a different PR has merged ELI5 changes into DNS policies.

PR #29297, opened 2026-03-24
[Chore] Update cache logic key

kodster28:


mvvmm:
/bonk is there any reason why this cache key would cause issues for us, or to cause us to use an outdated node_modules

.github/workflows/anchor-link-audit.yml
mvvmm, line 23: I noticed this setup is slightly different than the others. It doesn't include the npm cache (which just caches the individual package requests themselves - helpful on our cache misses)

kodster28: Updated!

mvvmm (approved):

PR #29302, opened 2026-03-24
Replace R2 llms-full.txt with fetch from middlecache (build/request time hybrid)

mvvmm:
## Summary

- Adds `bin/fetch-llms-full.ts` that fetches the llms-full.txt tarball from middlecache at build time and extracts per-product files into `public/`
- Per-product llms-full.txt files are served as static assets via the `ASSETS` binding
- Root `/llms-full.txt` (~40 MB) exceeds the Workers 25 MiB per-asset limit, so it is excluded from extraction and read from the `middlecache` R2 bucket at request time via direct binding (same account, no HTTP overhead)
- Replaces the `VENDORED_MARKDOWN` R2 binding with a `MIDDLECACHE` R2 binding pointing to the `middlecache` bucket

## Approach

**Build-time static assets + R2 read for root**: The `prebuild` hook downloads `llms-full.tar.gz` (~15MB) from middlecache and extracts only per-product files into `public/{product}/llms-full.txt` (96 products). Astro copies these into `dist/` and they are served as static assets.

The root `llms-full.txt` (~40 MB) exceeds the Workers 25 MiB per-asset size limit, so it is excluded from extraction via `tar --exclude` and the worker reads it from the `middlecache` R2 bucket at request time via direct binding.

The fetch script follows the same pattern as `bin/fetch-skills.ts`:
- Downloads to `.tmp/middlecache/` with caching (skips if already present)
- Supports `--soft` flag for `predev` (warns on failure instead of blocking)
- Supports `--force` flag to re-fetch
- Skips extraction if per-product files already exist (idempotent)

This is one of two approaches being tested — compare with #29301 which reads all files from the middlecache R2 bucket at request time.

## Tradeoffs vs request-time approach

| | Build-time (this PR) | Request-time (#29301) |
|---|---|---|
| Per-product latency | Served from edge cache/static assets — fast | R2 read at request time |
| Root llms-full.txt | R2 read (same account, direct binding) | R2 read (same account, direct binding) |
| Deploy size | +~40MB in static assets | No change |
| Freshness | Updated on each deploy | Always current |
| Runtime dependency | Middlecache R2 (root file only) | Middlecache R2 (all files) |

## Changes

- **New**: `bin/fetch-llms-full.ts` — Build-time fetch script (excludes root llms-full.txt from extraction)
- `package.json` — Added fetch-llms-full to `prebuild` and `predev` hooks
- `worker/index.ts` — Replaced `VENDORED_MARKDOWN` R2 read with targeted `MIDDLECACHE` R2 read for root `/llms-full.txt` only (per-product files served as static assets)
- `wrangler.toml` — Replaced `VENDORED_MARKDOWN` → `vendored-markdown` binding with `MIDDLECACHE` → `middlecache` binding
- `worker/worker-configuration.d.ts` — Replaced `VENDORED_MARKDOWN: R2Bucket` with `MIDDLECACHE: R2Bucket` in `Env` interface
- `.gitignore` — Added pattern for extracted per-product llms-full.txt files in `public/`


mvvmm:
Closing in favor of: #29301

I don't think its worth it at this time to bring any of this into the build. The main goal here was to make llms-full rely on markdown for agents, rather than our own custom conversion, which this work achieves. I don't want this work to also be responsible for making decisions around build/request time consumption of llms-full and which one we prefer (although, I'd probably lean towards request time anyways).

PR #29331, opened 2026-03-25
[Chore] Remove custom handling of unslashed requests

kodster28:


kodster28:
/bigbonk review this PR and provide any suggestions as to how we can make it better or more understandable. Also, any potential issues here?

kodster28:
Ugh, so this doesn't work b/c our redirects _still_ need to be evaluated before the Astro site loads and then does the behavior

PR #29334, opened 2026-03-25
feat: custom package manager component

MohamedH1998:
### Summary

Main reasoning behind this change was that existing `starlight-package-managers ` third-party dependency was a significant limitation — its design was fixed and could not be customised, which was a blocker for the visual redesign of one of the most visible and frequently used components across the docs site. This PR replaces it with a fully custom implementation.

- update PackageManagers component to use a self-contained astro component - we weren't able to change much of the design on the old component that we were using, so it made sense to just build or own version - this component 
- create a new generic card component (LayerCard) based on the kumo component

### Documentation checklist

**Before**
Screenshot 2026-03-25 at 18 20 46

**After**

Screenshot 2026-03-25 at 18 18 18

**Screen recording**

https://github.com/user-attachments/assets/effa00ac-67fc-4522-9ed7-b720e0891f5a

### DIFF
--- a/src/components/LayerCard.astro
+++ b/src/components/LayerCard.astro
@@ -1,0 +1,42 @@
+---
+interface Props {
+	header?: string;
+	href?: string;
+}
+
+const { header = "Next Steps", href } = Astro.props;
+---
+
+<div
+	class="flex w-full flex-col overflow-hidden rounded-lg bg-cl1-gray-9 text-base ring-1 ring-cl1-gray-8 dark:bg-cl1-gray-0 dark:ring-cl1-gray-1"
+>
+	<div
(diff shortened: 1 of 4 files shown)


ahaywood (approved):
This looks great! 👏 

I thought I could "get you" on checking what happens when multiple package manager components are on the page, but when a selection is made on one, it updates across all of them. _And_ it sticks on page refresh. Well done! 

The only thing I thought of (and maybe this is a separate ticket). Do we need a "Copied" confirmation message when the user clicks on the copy code button? Is the check enough of an indicator? 

Image

MohamedH1998:
> This looks great! 👏
> 
> I thought I could "get you" on checking what happens when multiple package manager components are on the page, but when a selection is made on one, it updates across all of them. _And_ it sticks on page refresh. Well done!
> 
> The only thing I thought of (and maybe this is a separate ticket). Do we need a "Copied" confirmation message when the user clicks on the copy code button? Is the check enough of an indicator?
> 
> Image

Lets gooo

It should be enough, this is the convention we're using in the dashboard

PR #29346, opened 2026-03-25
[Style Guide] Add solution guide content type

codyanthony850:
Adds `solution-guide` as a new content type to the Cloudflare developer documentation style guide, and updates related content type disambiguation resources.

Solution guides address a gap that no existing content type fills: customers who arrive with a goal or problem rather than a product name have no guided path through Cloudflare's documentation. How-tos and tutorials assume the user has already chosen a product. Solution guides start from the customer's goal, identify which products apply, and walk them through configuring those products together. This is a content type that major competitors (AWS, Azure, Google Cloud, Stripe) have established as a distinct category, and one that Cloudflare's own customer research confirms is needed — see the [Solution Guide Project Plan](https://wiki.cfdata.org/spaces/Prod/pages/1374954395/Solution+Guide+Project+Plan) for background and supporting data.

**Changes:**

- Adds `solution-guide.mdx` — new content type page covering purpose, when to use, tone, `pcx_content_type` value, callout pattern, structure, and template. Establishes that solution guides commit to a recommended path with alternatives surfaced via callouts and links, distinguishing them from descriptive product how-tos.
- Adds `content-type-disambiguation.mdx` partial — replaces `howto-vs-tutorial.mdx` with an updated note covering all three procedural types (how-to, tutorial, solution guide), using entry-point framing as the primary differentiator: how-to and tutorial assume the user has already chosen a product; solution guides are for users who arrive with a goal or problem and have not yet chosen a product.
- Deletes `howto-vs-tutorial.mdx` — superseded by the broader partial above.
- Updates `index.mdx` — adds solution guide to the content types table. Also removes a broken `Support content` row — the corresponding style guide page (`support-content.mdx`) was deleted in PR #20801 but the index table was never updated, leaving a dead link that was causing the link validator build check to fail. No pages in the current codebase use `pcx_content_type: support-content`.
- Updates `select-content-type.mdx` — merges previously separate disambiguation sections into one unified table covering all four procedural types (how-to, tutorial, 3rd-party integration guide, solution guide) with entry-point framing applied consistently.
- Updates `tutorial.mdx` — adds entry-point clarification to guidelines; updates Render reference to new partial; adds link to selection page.
- Updates `how-to.mdx` — updates Render reference to new partial; adds link to selection page.
- Updates `AGENTS.md` and `.opencode/agent/docs.md` — adds `solution-guide` to the valid `pcx_content_type` values lists in both AI agent instruction files so agents authoring or editing docs recognize the new content type as valid.

Related: PCX-21242


pedrosousa:
We could also update AI guidance in the following files so that the knew content type is considered valid:

- [`AGENTS.md`](https://github.com/cloudflare/cloudflare-docs/blob/production/AGENTS.md?plain=1#L86)
- [`.opencode/agent/docs.md`](https://github.com/cloudflare/cloudflare-docs/blob/production/.opencode/agent/docs.md?plain=1#L32)

codyanthony850:
Good call, thanks Pedro! Added `solution-guide` to the valid `pcx_content_type` lists in both `AGENTS.md` and `.opencode/agent/docs.md` in f5f6306.

dcpena (approved):

src/content/docs/style-guide/documentation-content-strategy/content-types/solution-guide.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+import { Render } from "~/components";
+
+## Purpose
+
+A solution guide is for users who arrive with a goal or problem rather than a product name. Some users know they have a problem and need to solve it. Others are just getting started and want to know what they should configure. In both cases, the guide identifies which Cloudflare products apply and walks the user through configuring them together.
+
+A solution guide answers the question: "I want to do X — what do I need, why does each piece matter, and how do I set it up?"
+
+Solution guides live at `/use-cases/{guide-slug}/` on developers.cloudflare.com.
```
dcpena: This line is a little bit confusing. Is this more of a guideline on where a user would find these guides in different product areas of the docs? If yes, I'd consider adding that info and then removing the "on developers.cloudflare.com" bit.

codyanthony850: Intent here is to explain where solution guides are deployed to for content creators/users to understand where to find and how architecture differs relative to product guides. Solution guides are centrally housed at developers.cloudflare.com/use-cases/solution-guide-slug, unlike product guides which are organized within their specific product vertical. This placements aid SEO discovery (because users haven't chosen a product vertical yet) and aligns with standard industry practice for this content type - I'm hoping that by explicitly stating this in the style guide, writer will understand the difference and not deviate from this practice.

codyanthony850:
@dcpena I updated the line to include the rationale. It now explains that guides live centrally outside any product vertical to support discovery for users who haven't chosen a product yet, which is the key distinction from product guides.

PR #29353, opened 2026-03-26
[DNS] Add enforce DNS only

RebeccaTamachiro:
### Summary

PCX-21243

### DIFF
--- a/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
+++ b/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
@@ -66,20 +66,24 @@ Due to DNS caching by recursive resolvers, the transition back to proxied behavi
 
 ## Other Cloudflare products
 
+Refer to the sections below in case you use other Cloudflare products that rely on DNS records.
+
 ### Included
 
-Enforce DNS-only affects proxied [Load Balancing](/load-balancing/) records.
+Enforce DNS-only affects the following records:
+
+- Proxied [Load Balancing](/load-balancing/): An **LB** record is visible on the DNS records table but proxy status is managed through the [Load Balancing configurations](/load-balancing/how-to/create-load-balancer/).
+- [Workers](/workers/) custom domains: records added to the DNS records table when you set up Workers [custom domains](/workers/configuration/routing/custom-domains/).
 
 ### Exceptions
 
 Enforce DNS-only does not affect the following records:
 
-- [Spectrum](/spectrum/) applications: CNAME records pointing to Spectrum endpoints are already DNS-only.
-- [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) records on SaaS provider zones: The default origin where [custom hostnames](/cloudflare-for-platforms/cloudflare-for-saas/domain-support/) are routed must be proxied.
-- [Workers](/workers/) custom domains: Proxied A records pointing to `100::`.
-- [R2](/r2/) custom domains: Proxied CNAME records pointing to `public.r2.dev`.
-- [Web3 gateways](/web3/): Ethereum and IPFS DNSLink gateways (CNAMEs to `ethereum.cloudflare.com` or `ipfs.cloudflare.com`).
-- [Cloudflare Tunnel](/tunnel/) records: These records are required for tunnel routing and setting them to DNS-only would break tunnel connectivity.
+- [Spectrum](/spectrum/) applications: DNS records managed by the Spectrum application.
+- [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) fallback origin: The proxied DNS record you designate as the [fallback origin](/cloudflare-for-platforms/cloudflare-for-saas/start/getting-started/#1-create-fallback-origin) for custom hostnames.
+- [R2](/r2/) custom domains: read-only proxied records added to the DNS records table when you set up [R2 custom domains](/r2/buckets/public-buckets/#connect-a-bucket-to-a-custom-domain).
+- [Web3 gateways](/web3/): CNAMEs to `ethereum.cloudflare.com`, `ipfs.cloudflare.com`, or `universal-gateway.ipfs.cloudflare.com`. These are read-only proxied records managed by the [gateway configuration](/web3/reference/gateway-dns-records/).
+- [Cloudflare Tunnel](/cloudflare-one/connections/connect-networks/): CNAME records pointing to a tunnel subdomain (**Target** ending in `.cfargotunnel.com`).
 
 ## Check current status
 
--- a/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
+++ b/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
@@ -72,18 +72,22 @@ Refer to the sections below in case you use other Cloudflare products that rely 
 
 Enforce DNS-only affects the following records:
 
-- Proxied [Load Balancing](/load-balancing/): An **LB** record is visible on the DNS records table but proxy status is managed through the [Load Balancing configurations](/load-balancing/load-balancers/create-load-balancer/).
-- [Workers](/workers/) custom domains: records added to the DNS records table when you set up Workers [custom domains](/workers/configuration/routing/custom-domains/).
+- [Load Balancing](/load-balancing/): proxied LB records visible on the DNS records table but managed through the [Load Balancing configurations](/load-balancing/load-balancers/create-load-balancer/).
+- Proxied DNS records that match a [Worker route](/workers/configuration/routing/routes/).
 
 ### Exceptions
 
 Enforce DNS-only does not affect the following records:
 
-- [Spectrum](/spectrum/) applications: DNS records managed by the Spectrum application.
 - [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) fallback origin: The proxied DNS record you designate as the [fallback origin](/cloudflare-for-platforms/cloudflare-for-saas/start/getting-started/#1-create-fallback-origin) for custom hostnames.
 - [R2](/r2/) custom domains: Read-only proxied records added to the DNS records table when you set up [R2 custom domains](/r2/buckets/public-buckets/#connect-a-bucket-to-a-custom-domain).
-- [Web3 gateways](/web3/): Read-only proxied records managed by the [Web3 gateway configuration](/web3/reference/gateway-dns-records/)
+- [Spectrum](/spectrum/) applications: DNS records managed by the Spectrum application.
 - [Tunnel](/tunnel/): CNAME records pointing to a tunnel subdomain. Refer to [Tunnel routing](/tunnel/routing/#create-a-dns-record) or [Cloudflare One](/cloudflare-one/networks/connectors/cloudflare-tunnel/routing-to-tunnel/dns/) for details.
+- [Web3 gateways](/web3/): Read-only proxied records managed by the [Web3 gateway configuration](/web3/reference/gateway-dns-records/).
+- [Workers](/workers/) custom domains: Read-only proxied records added to the DNS records table when you set up Workers [custom domains](/workers/configuration/routing/custom-domains/).
+   :::note[Custom domain or route match]
+	 Proxied records that match a Worker [route](/workers/configuration/routing/routes/) are regular DNS records and will be [affected](#included) by the enforce DNS-only setting.
+   :::
 
 ## Check current status
 
--- a/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
+++ b/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
@@ -76,7 +76,6 @@ Enforce DNS-only does not affect the following records:
 
 - [Spectrum](/spectrum/) applications: CNAME records pointing to Spectrum endpoints are already DNS-only.
 - [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) records on SaaS provider zones: The default origin where [custom hostnames](/cloudflare-for-platforms/cloudflare-for-saas/domain-support/) are routed must be proxied.
-- Cloudflare for SaaS records on customer zones: Setting CNAME records on a customer zone to DNS-only would not help if the SaaS provider's zone remains proxied.
 - [Workers](/workers/) custom domains: Proxied A records pointing to `100::`.
 - [R2](/r2/) custom domains: Proxied CNAME records pointing to `public.r2.dev`.
 - [Web3 gateways](/web3/): Ethereum and IPFS DNSLink gateways (CNAMEs to `ethereum.cloudflare.com` or `ipfs.cloudflare.com`).
--- a/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
+++ b/src/content/docs/dns/proxy-status/enforce-dns-only.mdx
@@ -8,13 +8,12 @@ sidebar:
 
 import { APIRequest } from "~/components";
 
-The enforce DNS-only setting is an account-level break-glass mechanism that allows you to bypass Cloudflare's reverse proxy for all zones in your account in a single action. When enabled, Cloudflare responds to DNS queries with your origin IP addresses instead of Cloudflare anycast IP addresses, effectively setting all [proxied DNS records](/dns/proxy-status/) to DNS-only without modifying the records themselves.
+The enforce DNS-only setting is an account-level break-glass mechanism that allows you to bypass Cloudflare's reverse proxy for all zones in your account in a single action. When enabled, Cloudflare responds to DNS queries with your origin IP addresses instead of Cloudflare's anycast IP addresses, effectively setting all [proxied DNS records](/dns/proxy-status/) to DNS-only without modifying the records themselves.
 
 This setting is intended for emergency situations only, such as during an outage when you need to quickly route traffic directly to your origins.
 
 :::caution
-
-Enabling this setting immediately exposes your origin IP addresses and removes all Cloudflare protections — including DDoS mitigation, WAF, caching, and all other proxy-based features — for every zone in your account. Use with extreme caution and only after proper preparation.
+Enabling this setting exposes your origin IP addresses and removes all Cloudflare protections — including DDoS mitigation, WAF, caching, and all other proxy-based features — for every zone in your account. Use with extreme caution and only after proper [preparation](#preparation).
 
 :::
 
@@ -24,7 +23,11 @@ Enabling this setting immediately exposes your origin IP addresses and removes a
 - Non-destructive: Does not modify your DNS records. Disabling the setting restores normal proxy behavior.
 - API-only: Available through the API only, not in the Cloudflare dashboard.
 
-## Before you begin
+:::note[Auto TTL for proxied records]
+Due to DNS caching by recursive resolvers, the transitions from proxied to DNS-only and back may not be instantaneous. Since all proxied records have a TTL of **Auto**, this value (5 minutes by default) determines how long resolvers may continue to serve Cloudflare's anycast IPs or your origin IP addresses.
+:::
+
+## Preparation
 
 Before relying on enforce DNS-only as part of your incident response plan, you should:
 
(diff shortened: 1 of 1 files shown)


src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+	method="PATCH"
+	json={{
+		enforce_dns_only: false,
+	}}
+/>
+
+After you disable the setting, Cloudflare resumes responding to DNS queries with anycast IP addresses for proxied records and all proxy-based features are restored.
+
+:::note
+Due to DNS caching by recursive resolvers, the transition back to proxied behavior may not be instantaneous. The TTL of your proxied records (default 300 seconds) determines how long resolvers may continue to serve the origin IP addresses after you disable the setting.
```
RebeccaTamachiro:
```suggestion
Due to DNS caching by recursive resolvers, the transition back to proxied behavior may not be instantaneous. The TTL of each record determines how long resolvers may continue to serve the origin IP addresses after you disable the setting.
```

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+Enforce DNS-only affects proxied [Load Balancing](/load-balancing/) records.
+
+### Exceptions
+
+Enforce DNS-only does not affect the following records:
+
+- [Spectrum](/spectrum/) applications: CNAME records pointing to Spectrum endpoints are already DNS-only.
+- [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) records on SaaS provider zones: The default origin where [custom hostnames](/cloudflare-for-platforms/cloudflare-for-saas/domain-support/) are routed must be proxied.
+- Cloudflare for SaaS records on customer zones: Setting CNAME records on a customer zone to DNS-only would not help if the SaaS provider's zone remains proxied.
+- [Workers](/workers/) custom domains: Proxied A records pointing to `100::`.
```
hannes-cf: We should rephrase this. While it's technically true, that workers custom domain records are just normal AAAA (not A) records pointing to `100::`, that fact is not relevant to the user. It's also not exposed in the UI.

Instead we should rather say that these records are managed by the Workers custom domains settings in the workers dashboard section.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+### Included
+
+Enforce DNS-only affects proxied [Load Balancing](/load-balancing/) records.
+
+### Exceptions
+
+Enforce DNS-only does not affect the following records:
+
+- [Spectrum](/spectrum/) applications: CNAME records pointing to Spectrum endpoints are already DNS-only.
```
hannes-cf: "CNAME records pointing to Spectrum endpoints (...)" sounds like user created CNAME records that point to a Spectrum app.

Let's rephrase and not mention what Spectrum records actually are under the hood. Rather we should say something that these records are managed by the Spectrum application.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+- [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) records on SaaS provider zones: The default origin where [custom hostnames](/cloudflare-for-platforms/cloudflare-for-saas/domain-support/) are routed must be proxied.
```
hannes-cf: Let's rephrase to specifically reference the "fallback origin" record. This is a conventional A, AAAA, CNAME record that must be proxied and will be locked once it is configured as the "fallback origin" in the Cloudflare for SaaS settings.

This record is exempted from the enforce dns-only logic, because it might be used as target by all custom hostnames and thus needs to be proxied in order to provide the functionality of Cloudflare for SaaS.

We probably don't have to list all these details here (although I wouldn't mind if we did).

Maybe also reference this section instead?

https://developers.cloudflare.com/cloudflare-for-platforms/cloudflare-for-saas/start/getting-started/#1-create-fallback-origin

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+- Cloudflare for SaaS records on customer zones: Setting CNAME records on a customer zone to DNS-only would not help if the SaaS provider's zone remains proxied.
```
hannes-cf: Need to delete this line as records on eyeball zones pointing to a SaaS provider's zone are *not* exempt from the enforce dns-only logic.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+- [Workers](/workers/) custom domains: Proxied A records pointing to `100::`.
+- [R2](/r2/) custom domains: Proxied CNAME records pointing to `public.r2.dev`.
```
hannes-cf: Rephrase to something like: Records managed by R2 which can only be managed in the R2 settings and is otherwise read-only to the user. The fact that this is a proxied CNAME pointing to public.r2.dev is irrelevant.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+- [Web3 gateways](/web3/): Ethereum and IPFS DNSLink gateways (CNAMEs to `ethereum.cloudflare.com` or `ipfs.cloudflare.com`).
```
hannes-cf: Same comment as for R2

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,94 @@
...
+- [Cloudflare Tunnel](/tunnel/) records: These records are required for tunnel routing and setting them to DNS-only would break tunnel connectivity.
```
hannes-cf: While these records are not read-only and could be modified by the user in the DNS section, as long as they're pointing to something ending in cfargotunnel.com (or the fed version of this) they will be exempted from enforce dns-only.

cc @matilde-cf

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
hannes-cf: We probably should also add that conventional proxied DNS records (not managed by Workers custom domains) that a Worker is running on are NOT exempted from the enforce dns-only logic.

RebeccaTamachiro: `workers custom domain records are just normal AAAA (not A) records pointing to 100::` is this surfaced at all? From the Workers custom domain UI I seem to only be pointed to the DNS records table.

Then, on the records table, I find what I think is the same you're referring to by `conventional proxied DNS records that a Worker is running on` --> Added a note on this [here](https://github.com/cloudflare/cloudflare-docs/pull/29353/changes#diff-609a1e18d49389183b671e8dfddabc7b2928dcb1d90e474d2a586ec5cf0d31d5R76)

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
RebeccaTamachiro: I was in doubt about this one bc, although I really can't see `public.r2.dev` anywhere, with the Web3 gateway targets, I do see them - so mentioning it could be a good visual reference?

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
hannes-cf: Actually the transitions in both directions are impacted by the TTL. Since we store a TTL of 1 (meaning auto, currently set to 300 seconds) for proxied records, that TTL would also apply once the enforce dns-only setting is enabled. So then for the transition back from unproxied -> proxied, resolvers observe the same TTL (of 300 seconds).

The change you suggested "The TTL of each record (...)" sounds good to me, we just need to change the first sentence to reflect what I mentioned above.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,97 @@
...
+## Other Cloudflare products
+
+Refer to the sections below in case you use other Cloudflare products that rely on DNS records.
+
+### Included
+
+Enforce DNS-only affects the following records:
+
+- Proxied [Load Balancing](/load-balancing/): An **LB** record is visible on the DNS records table but proxy status is managed through the [Load Balancing configurations](/load-balancing/load-balancers/create-load-balancer/).
+- [Workers](/workers/) custom domains: records added to the DNS records table when you set up Workers [custom domains](/workers/configuration/routing/custom-domains/).
```
hannes-cf: Wait, there might have been a misunderstanding. These records (meaning the ones managed by the Workers custom domain feature) are exempted from the enforce dns-only setting.

What is included in the setting are manually created proxied records where a worker runs on. And for these we should explicitly call out that once the enforce dns-only setting is applied, the Workers functionality will stop worker as requests aren't proxied anymore.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+### Exceptions
+
+Enforce DNS-only does not affect the following records:
+
+- [Spectrum](/spectrum/) applications: DNS records managed by the Spectrum application.
+- [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) fallback origin: The proxied DNS record you designate as the [fallback origin](/cloudflare-for-platforms/cloudflare-for-saas/start/getting-started/#1-create-fallback-origin) for custom hostnames.
+- [R2](/r2/) custom domains: read-only proxied records added to the DNS records table when you set up [R2 custom domains](/r2/buckets/public-buckets/#connect-a-bucket-to-a-custom-domain).
+- [Web3 gateways](/web3/): CNAMEs to `ethereum.cloudflare.com`, `ipfs.cloudflare.com`, or `universal-gateway.ipfs.cloudflare.com`. These are read-only proxied records managed by the [gateway configuration](/web3/reference/gateway-dns-records/).
+- [Cloudflare Tunnel](/cloudflare-one/networks/connectors/cloudflare-tunnel/routing-to-tunnel/dns/): CNAME records pointing to a tunnel subdomain (**Target** ending in `.cfargotunnel.com`).
```
hannes-cf: Let's rephrase: Records created by the Cloudflare Tunnel service, displayed as type: Tunnel in the records table in the dashboard.

While the user sees the actual target once they edit the records, the value is irrelevant. I will actually raise this with the Tunnel team and nudge them to flag these records as read-only so we can display them the same way as R2 records.

matildeopbravo: I think this is a bit tricky because the records are not always created by the Cloudflare Tunnel Service. If the customers creates a tunnel through the API, they need to create the CNAME themselves, not the service. So they will definitely be aware of the CNAME target at that point. Ideally they will change this flow to be more like R2 but, as it is now, I think we should either mention the tunnel subdomain (as Rebecca did) or something like "records pointing to Tunnels" but not sure if that's a bit hard to understand.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
matildeopbravo: I think "proxied records that match a Worker route" might be a good way to describe these records that are affected (not exempt)?

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
matildeopbravo: Yeah, you're right, R2 completely omits the target from the docs but not Web3. But I still think we can avoid mentioning the target because it might change in the future and then the docs could become outdated. And since the records are marked as read-only, should be fairly easy to spot them? What do you think about "Records managed by Web3 gateway"?

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
RebeccaTamachiro: Ooh, I see! Because all proxied records have the same TTL, which cannot be overridden - I was forgetting that. Also, enforcing DNS-only will only change the proxy status, leaving the TTL as-is. Let me clarify that.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
RebeccaTamachiro: Oh, I see where the problem was! I was looking at a Pages project in my testing, not a Workers one. So I meant: From the Pages custom domain UI I seem to only be pointed to the DNS records table. I can confirm the Workers custom domain one appears as a read-only record

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,100 @@
...
+Enforce DNS-only affects the following records:
+
+- [Load Balancing](/load-balancing/): proxied LB records visible on the DNS records table but managed through the [Load Balancing configurations](/load-balancing/load-balancers/create-load-balancer/).
+- Proxied DNS records that match a [Worker route](/workers/configuration/routing/routes/).
+
+### Excluded
+
+Enforce DNS-only does not affect the following records:
+
+- [Cloudflare for SaaS](/cloudflare-for-platforms/cloudflare-for-saas/) fallback origin: The proxied DNS record you designate as the [fallback origin](/cloudflare-for-platforms/cloudflare-for-saas/start/getting-started/#1-create-fallback-origin) for custom hostnames.
```
hannes-cf:
```suggestion
```

hannes-cf: We decided to not exclude this.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,100 @@
...
+pcx_content_type: troubleshooting
+title: Enforce DNS-only
+sidebar:
+  order: 3
+  label: Enforce DNS-only
+---
+
+import { APIRequest } from "~/components";
+
+The enforce DNS-only setting is an account-level break-glass mechanism that allows you to bypass Cloudflare's reverse proxy for all zones in your account in a single action. When enabled, Cloudflare responds to DNS queries with your origin IP addresses instead of Cloudflare's anycast IP addresses, effectively setting all [proxied DNS records](/dns/proxy-status/) to DNS-only without modifying the records themselves.
```
hannes-cf:
```suggestion
The enforce DNS-only setting is an account-level break-glass mechanism that allows you to bypass Cloudflare's reverse proxy for all zones in your account in a single action. When enabled, Cloudflare responds to DNS queries with the underlying record content — origin IP addresses for proxied `A` and `AAAA` records, and CNAME targets for proxied `CNAME` records — instead of Cloudflare's anycast IP addresses, effectively setting all [proxied DNS records](/dns/proxy-status/) to DNS-only without modifying the records themselves.
```

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,100 @@
...
+:::note[Auto TTL for proxied records]
+Due to DNS caching by recursive resolvers, the transitions from proxied to DNS-only and back may not be instantaneous. Since all proxied records have a TTL of **Auto**, this value (5 minutes by default) determines how long resolvers may continue to serve Cloudflare's anycast IPs or your origin IP addresses.
+:::
+
+## Preparation
+
+Before relying on enforce DNS-only as part of your incident response plan, you should:
+
+- Verify origin server capacity: Without Cloudflare proxying, your origin servers handle all traffic directly, including traffic that Cloudflare would normally cache or filter. Ensure your infrastructure can sustain this load.
+- Review exposed IP addresses: When enforce DNS-only is active, all origin IPs configured in proxied records become publicly visible through DNS queries. If your origins rely on IP obscurity for security, plan accordingly.
```
hannes-cf:
```suggestion
- Review exposed record content: When enforce DNS-only is active, all origin IPs configured in proxied `A` and `AAAA` records, as well as the targets of proxied `CNAME` records, become publicly visible through DNS queries. If your origins rely on IP obscurity for security, plan accordingly.
- ```

File: src/content/docs/dns/proxy-status/enforce-dns-only.mdx [hunk 14]
```diff
@@ -0,0 +1,100 @@
...
+
+<APIRequest
+	path="/accounts/{account_id}/dns_settings"
+	method="PATCH"
+	json={{
+		enforce_dns_only: true,
+	}}
+/>
+
+Once enabled, Cloudflare responds to DNS queries for all proxied records with your configured origin IP addresses instead of Cloudflare's anycast IPs.
```
hannes-cf:
```suggestion
Once enabled, Cloudflare responds to DNS queries for all proxied records with the underlying record content — your configured origin IP addresses for `A` and `AAAA` records, and the configured CNAME target for `CNAME` records — instead of Cloudflare's anycast IPs.
```

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
RebeccaTamachiro: Should I move this item to the `#included` section above then?

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
hannes-cf: Sorry for the late response, just noticed your response now. No, this isn't surfaced. The extra note looks good!

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
hannes-cf: Yes, that doesn't hurt to call out explicitly there.

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,100 @@
...
+import { APIRequest } from "~/components";
+
+The enforce DNS-only setting is an account-level break-glass mechanism that allows you to bypass Cloudflare's reverse proxy for all zones in your account in a single action. When enabled, Cloudflare responds to DNS queries with the underlying record content — origin IP addresses for proxied `A` and `AAAA` records, and CNAME targets for proxied `CNAME` records — instead of Cloudflare's anycast IP addresses, effectively setting all [proxied DNS records](/dns/proxy-status/) to DNS-only without modifying the records themselves.
+
+This setting is intended for emergency situations only, such as during an outage when you need to quickly route traffic directly to your origins.
+
+:::caution
+Enabling this setting exposes your origin IP addresses and removes all Cloudflare protections — including DDoS mitigation, WAF, caching, and all other proxy-based features — for every zone in your account. Use with extreme caution and only after proper [preparation](#preparation).
+
+:::
```
marciocloudflare:
```suggestion
:::
```

src/content/docs/dns/proxy-status/enforce-dns-only.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+## Key characteristics
+
+- Account-level: Affects all zones in the account simultaneously.
+- Non-destructive: Does not modify your DNS records. Disabling the setting restores normal proxy behavior.
+- API-only: Available through the API only, not in the Cloudflare dashboard.
+
+:::note[Auto TTL for proxied records]
+Due to DNS caching by recursive resolvers, the transitions from proxied to DNS-only and back may not be instantaneous. Since all proxied records have a TTL of **Auto**, this value (5 minutes by default) determines how long resolvers may continue to serve Cloudflare's anycast IPs or your origin IP addresses.
```
marciocloudflare:
```suggestion
Due to DNS caching by recursive resolvers, the transitions from proxied to DNS-only and back may not be instantaneous. Since all proxied records have a TTL of **Auto**, this value (five minutes by default) determines how long resolvers may continue to serve Cloudflare's anycast IPs or your origin IP addresses.
```

marciocloudflare (approved):

PR #29358, opened 2026-03-26
feat: collapsible sidebar

MohamedH1998:
### Summary

This change adds functionality to make the sidebar collapsible, this is a prereq for the landing page designs changes I have coming.

I had initially explored different ways of collapsing the sidebar that I decided against:
- Adding the trigger in the header: it would cause the logo in the top left to be offset
- Adding the trigger in the footer of the sidebar: this would mean that the sidebar would still be required to be visible on the page, which we don't want

This the least invasive, yet clean implementation I came up with, open to suggestions

### Screenshots (optional)

**Before**
Screenshot 2026-03-26 at 12 19 52

**After**

Screenshot 2026-03-26 at 12 19 21

**Screen recording**

https://github.com/user-attachments/assets/8faba279-0487-4a92-ad0a-92b57ddfb31f

### DIFF
--- a/src/components/overrides/Head.astro
+++ b/src/components/overrides/Head.astro
@@ -270,7 +270,8 @@ metaTags.map((attrs) => {
 <script src="src/scripts/mermaid.ts"></script>
 <script src="src/scripts/analytics.ts"></script>
 
-{/* Sidebar collapse: apply persisted state before first paint to prevent flash */}
+{/* Sidebar collapse: apply persisted state before first paint to prevent flash.
+    Key must stay in sync with STORAGE_KEY in Header.astro. */}
 <script is:inline>
 (function () {
 	try {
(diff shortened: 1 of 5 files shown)


ahaywood:
Nit: Icon alignment with the Search "/" and Chevrons
CleanShot 2026-03-26 at 11 15 16@2x

On tablet (`~800px`) the border divider ends abruptly. Does it make sense to have this extend across the page horizontally or leave a vertical border?
CleanShot 2026-03-26 at 11 16 04@2x

Also on tablet, when the subnav is collapsed it leaves awkward spacing on the left. Seems like part of the reason someone would want to collapse the sidebar is to have more room for reading. Does it makes sense to shift the layout? Or, if the content doesn't extend 100%, at least center the main content area as a distraction free view?
CleanShot 2026-03-26 at 11 17 03@2x

PR #29418, opened 2026-03-27
Updates to WfP configuration docs

dinasaur404:
Minor updates to Workers for Platforms configuration section.

### DIFF
--- a/src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
+++ b/src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
@@ -1,0 +1,178 @@
+---
+title: Get started
+pcx_content_type: how-to
+sidebar:
+  order: 2
+---
+
+import { TypeScriptExample, WranglerConfig } from "~/components";
+
+This guide walks through configuring binding groups and using them to inject per-customer resources at runtime. Binding groups work with both [Dynamic Workers](/dynamic-workers/) and the Workers for Platforms [dispatch namespace](/cloudflare-for-platforms/workers-for-platforms/configuration/dynamic-dispatch/).
+
+## Configure the bindings
+
+On your platform Worker (the Worker Loader or dispatch Worker), add binding groups for each resource type your customers need. Each binding group is a single binding that manages many customer-scoped resources.
+
+<WranglerConfig>
+
+```toml
+name = "my-platform"
+
+[[kv_namespaces]]
+binding = "CUSTOMER_KV"
+namespace_id = "platform-customer-stores"
+
+[[ai_search]]
+binding = "CUSTOMER_SEARCH"
+namespace = "platform-customer-indexes"
+
+[[r2_buckets]]
+binding = "CUSTOMER_STORAGE"
+bucket_name = "platform-customer-uploads"
+```
+
+</WranglerConfig>
+
+## Use with Dynamic Workers
+
+In the Worker Loader, resolve each customer's resources from the binding groups and pass them as bindings to the Dynamic Worker.
+
+<TypeScriptExample filename="src/loader.ts">
+
+```ts
+export default {
+	async fetch(request: Request, env: Env) {
+		const customerId = await authenticateRequest(request);
+
+		// Get this customer's resources from each binding group
+		const customerKV = env.CUSTOMER_KV.get(customerId);
+		const customerSearch = env.CUSTOMER_SEARCH.get(customerId);
+		const customerBucket = env.CUSTOMER_STORAGE.get(customerId);
+
+		// Load or reuse this customer's Dynamic Worker
+		const worker = env.LOADER.get(`customer-${customerId}`, async () => {
+			const code = await env.PLATFORM_CODE.get(`${customerId}/worker.js`);
+
+			return {
+				compatibilityDate: "2026-01-01",
+				mainModule: "index.js",
+				modules: { "index.js": code },
+
+				// Pass in ONLY this customer's resources.
+				// The Dynamic Worker sees these as its own bindings.
+				bindings: {
+					KV: customerKV,
+					SEARCH: customerSearch,
+					STORAGE: customerBucket,
+				},
+				globalOutbound: null,
+			};
+		});
+
+		return worker.getEntrypoint().fetch(request);
+	},
+};
+```
+
+</TypeScriptExample>
+
+The Dynamic Worker receives `env.KV`, `env.SEARCH`, and `env.STORAGE` as standard bindings. The customer's code does not need to know about binding groups.
+
+## Use with Workers for Platforms dispatcher
+
+Pass binding groups when you dispatch to a user Worker. The bindings are injected at dispatch time instead of deploy time.
+
+<TypeScriptExample filename="src/dispatcher.ts">
+
+```ts
+export default {
+	async fetch(request: Request, env: Env) {
+		const customerId = new URL(request.url).hostname.split(".")[0];
+
+		const userWorker = env.DISPATCHER.get(customerId, {
+			bindings: {
+				KV: env.CUSTOMER_KV.get(customerId),
+				DB: env.CUSTOMER_DBS.get(customerId),
+				SEARCH: env.CUSTOMER_SEARCH.get(customerId),
+				STORAGE: env.CUSTOMER_STORAGE.get(customerId),
+			},
+		});
+
+		return userWorker.fetch(request);
+	},
+};
+```
+
+</TypeScriptExample>
+
+## Understand resource auto-creation
+
+Resources within a binding group are created automatically on first use. You do not need a separate provisioning step.
+
+The flow works as follows:
+
+1. Your customer writes a normal Worker that calls `env.KV.put("key", "value")`.
+2. On the platform side, you pass in the binding group scoped to the customer ID.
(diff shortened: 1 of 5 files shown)


src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
```diff
@@ -0,0 +1,178 @@
...
+## Configure the bindings
+
+On your platform Worker (the Worker Loader or dispatch Worker), add binding groups for each resource type your customers need. Each binding group is a single binding that manages many customer-scoped resources.
+
+<WranglerConfig>
+
+```toml
+name = "my-platform"
+
+[[kv_namespaces]]
```
a-robinson, line 21: What's differentiating these configs as binding groups rather than just individual bindings? These look like the config for binding to just a single KV namespace / AI search namespace / R2 bucket.

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+## Understand resource auto-creation
+
+Resources within a binding group are created automatically on first use. You do not need a separate provisioning step.
+
+The flow works as follows:
+
+1. Your customer writes a normal Worker that calls `env.KV.put("key", "value")`.
+2. On the platform side, you pass in the binding group scoped to the customer ID.
+3. When the customer's Worker calls `env.KV.put()` for the first time, the runtime sees "this is a write to binding group `CUSTOMER_KV`, scoped to `acme-corp`" — and if a KV store for `acme-corp` does not exist yet, it creates one.
```
a-robinson, line 116: How important is it that this gets created when the binding is first used vs when `env.CUSTOMER_KV.get(customerId)` is called in the dispatching worker?

Is it bad if the dispatching worker's call is what actually triggers creation of the resource?

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
```diff
@@ -0,0 +1,178 @@
...
+4. Each resource is created as a real, first-class resource with its own ID and data.
```
a-robinson, line 117: Is there anything special about how these resources will show up in the control plane APIs for each of the different resource types? e.g. will they be given special names, or tags, or some other field or organization that makes their binding group clear when you, for example, [list KV namespaces](https://developers.cloudflare.com/api/resources/kv/subresources/namespaces/methods/list)?

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/supported-bindings.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+This page lists the bindings that binding groups support and how each resource is created.
+
+## Auto-created bindings
+
+These bindings create the underlying resource automatically on first use. You do not need a separate provisioning step.
+
+| Binding                                                     | What gets created | Creation trigger                                                 |
+| ----------------------------------------------------------- | ----------------- | ---------------------------------------------------------------- |
+| [KV](/kv/)                                                  | KV namespace      | First write                                                      |
```
a-robinson, line 16: Is there a reason these lists are so specific about what provisions a resource?

Like, is it really intended that doing a `get()` would n't cause a KV namespace / R2 bucket / AI search index to be created?

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/supported-bindings.mdx
```diff
@@ -0,0 +1,43 @@
...
+| [D1](/d1/)                                                  | D1 database       | First use (empty SQLite database, apply schema with `db.exec()`) |
+| [AI Search](/ai-search/)                                    | Search index      | First `upsert()`                                                 |
+| [Durable Objects](/durable-objects/)                        | DO namespace      | First `.get()` of an object within the namespace                 |
+| [Queues](/queues/)                                          | Queue             | First `send()`                                                   |
+| [Analytics Engine](/analytics/analytics-engine/)            | Dataset           | First write                                                      |
+| [Rate Limiting](/workers/runtime-apis/bindings/rate-limit/) | Rate limiter      | First `.limit()` call                                            |
+| [Workflows](/workflows/)                                    | Workflow instance | First `.create()` call                                           |
```
a-robinson, line 24: Workflows require a script to be defined in order to create an instance, and presumably different tenants will want their own scripts, won't they? Maybe the best option is to have the dispatcher create() the Workflow binding with a script for the tenant, something like `env.WORKFLOW.create(customerId, {script: customerScript})`.

IDK. Either way, this can probably better be sorted out elsewhere.

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/limits.mdx
```diff
@@ -0,0 +1,185 @@
...
+};
+```
+
+</TypeScriptExample>
+
+## Operations vs. storage limits
+
+Limits fall into two categories:
+
+**Operation limits** count actions over the billing cycle. They reset at the start of each monthly billing period.
```
a-robinson, line 147: I shouldn't be making writing nits on docs that will surely be updated before publishing, but it'd be better to call out that these are examples, since as is I read it as the full list of each (which obviously isn't true).

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/migrate.mdx
```diff
@@ -0,0 +1,91 @@
...
+```ts
+export default {
+	async fetch(request: Request, env: Env) {
+		const customerId = new URL(request.url).hostname.split(".")[0];
+
+		// Phase 1: Move KV to binding groups first.
+		// DB, SEARCH, STORAGE still come from deploy-time config.
+		const userWorker = env.DISPATCHER.get(customerId, {
+			bindings: {
+				KV: env.CUSTOMER_KV.get(customerId),
```
a-robinson, line 66: It seems like during these migration cases, many customers would much rather get back an error than have a new resource automatically created for them when calling `env.NS.get()` if the resource doesn't already exist.

Because if they get back an error they can fall back to using the existing deploy-time binding. But if they deploy this kind of code before the namespace has been adopted/migrated, then they'll pass in an empty namespace to their tenant, to whom it will look like data was lost. And if that tenant then writes data into the new namespace, we'll have a split-brain scenario on our hands that will have to be manually resolved (by somehow merging the data from the real namespace and the automatically created empty namespace).

a-robinson (commented):
There's plenty still to figure out, but it feels very much like the right direction, the right goal to aim for.

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+```ts
+// Creates a KV namespace named "acme-corp"
+// within the "platform-customer-stores" group
+env.CUSTOMER_KV.get("acme-corp");
+```
+
+## Create resources that require configuration
+
+Some resources cannot be auto-created from just a name — they need configuration that you provide. For these, create the resource explicitly before it is used.
```
vy-ton, line 151: Or if you want to pass additional config properties like D1 database location

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
vy-ton: Will need some differentiator in future state when `env` is a flat config:

```
  "env": {
    "DB": { "type": "d1", "id": "xxxx-xxxx-xxxx" },
    "CACHE": { "type": "kv", "id": "yyyy-yyyy-yyyy" },
    "UPLOADS": { "type": "r2", "name": "user-uploads" }
  }
```

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/binding-groups/get-started.mdx
```diff
@@ -0,0 +1,178 @@
...
+  order: 2
+---
+
+import { TypeScriptExample, WranglerConfig } from "~/components";
+
+This guide walks through configuring binding groups and using them to inject per-customer resources at runtime. Binding groups work with both [Dynamic Workers](/dynamic-workers/) and the Workers for Platforms [dispatch namespace](/cloudflare-for-platforms/workers-for-platforms/configuration/dynamic-dispatch/).
+
+## Configure the bindings
+
+On your platform Worker (the Worker Loader or dispatch Worker), add binding groups for each resource type your customers need. Each binding group is a single binding that manages many customer-scoped resources.
```
vy-ton, line 14: Folks will use binding groups in their normal Worker as well, ex D1 DBs as shards. Right now, I can't think of anything that would be different compared to Worker Loader or dispatch Worker so just worth keeping in mind.

marciocloudflare:
PCX triage note — PCX team pinged the assigned reviewer on 2026-04-20. Waiting for review.

Issue #22525, opened 2025-05-17
What is the official support policy for Bun?


Cherry:
### Existing documentation URL(s)

- https://developers.cloudflare.com/queues/tutorials/web-crawler-with-browser-rendering/
- https://developers.cloudflare.com/workers-ai/get-started/dashboard/
- https://developers.cloudflare.com/workers/frameworks/framework-guides/hono/
- etc.

### What changes are you suggesting?

https://github.com/cloudflare/cloudflare-docs/pull/19813 removed Bun from all C3 commands due to issues it's caused. For example:
- https://github.com/cloudflare/workers-sdk/pull/8889
- https://github.com/cloudflare/workers-sdk/issues/9067#issuecomment-2855954463
- etc.

However it seems these were added back in lots of places in https://github.com/cloudflare/cloudflare-docs/pull/22506. What's the official policy for Bun support? Is Bun actively supported in C3/wrangler now?

kodster28:
Heard + raised in an internal thread. Thanks for flagging, @Cherry.

eveeifyeve:
All of the issues should be fixed try using 1.3.

PR #29433, opened 2026-03-28
feat: add bun support to package manager component

MohamedH1998:
### Summary

Add bun support to package manager component

### Screenshots (optional)

Screenshot 2026-03-28 at 21 50 03

### Documentation checklist

### DIFF
--- a/src/util/package-managers.ts
+++ b/src/util/package-managers.ts
@@ -1,4 +1,4 @@
-export type Manager = "npm" | "yarn" | "pnpm";
+export type Manager = "npm" | "yarn" | "pnpm" | "bun";
 export type CommandType =
 	| "add"
 	| "create"
@@ -49,9 +49,19 @@ const commands: Record<
 		remove: "pnpm remove",
 		dev: "-D",
 	},
+	bun: {
+		add: "bun add",
+		create: "bun create",
+		dlx: "bunx",
+		exec: "bunx",
+		install: "bun install",
+		run: "bun run",
+		remove: "bun remove",
+		dev: "-d",
+	},
 };
 
-export const MANAGERS: Manager[] = ["npm", "yarn", "pnpm"];
+export const MANAGERS: Manager[] = ["npm", "yarn", "pnpm", "bun"];
 
 export function getCommand(
 	mgr: Manager,
--- a/src/util/package-managers.ts
+++ b/src/util/package-managers.ts
@@ -51,11 +51,7 @@ const commands: Record<
 	},
 	bun: {
 		add: "bun add",
-		create: "bun create",
-		dlx: "bunx",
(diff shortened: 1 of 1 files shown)


Cherry:
Is Bun actively supported in wrangler (etc.)? This was intentionally removed in the past as it's caused issues in wrangler, c3 and others: https://github.com/cloudflare/cloudflare-docs/issues/22525

Other ref: https://github.com/cloudflare/workers-sdk/pull/8889

threepointone:
for package management, it should work just fine there. for wrangler dev, we should be using node when the cli is started up, if not that's a bug.

threepointone:
tbh even if it fails with bun, we should file bugs there, we're friends and they should be able to fix it

Cherry:
Agreed bugs should be filed and fixed, but I believe this component is used by more than just package management today, like for actual deploy/dev commands, such as:

- https://feat-add-bun-to-pkm.preview.developers.cloudflare.com/workers/framework-guides/automatic-configuration/
- https://feat-add-bun-to-pkm.preview.developers.cloudflare.com/workers/development-testing/environment-variables/#basic-setup

etc.

which could/will have various issues with Bun and are very much not supported as per: https://x.com/MattieTK/status/2037810746868822458

- https://github.com/oven-sh/bun/issues/7703
- https://github.com/oven-sh/bun/issues/10464
- https://github.com/oven-sh/bun/issues/24229
- https://github.com/oven-sh/bun/issues/26674

Perhaps this should only show `bun` if the `type` isn't set to `create`, `exec`, `run`, etc. and any other non-package management commands? I'd argue it's still pretty confusing from a UX and education standpoint to show `bun` commands at all until it works across the board though personally.

threepointone (approved):

threepointone:
Passed them on to the team!

MattieTK:
@MohamedH1998 I agree it would be good to expand the component to not show the following commands in bun if that is the prop delivered

>Perhaps this should only show bun if the type isn't set to create, exec, run, etc.

Bun is a great package manager so I appreciate the value in doing this, but we don't want folks to run into problems using bun as their script runner.

MattIPv4:
> I'd argue it's still pretty confusing from a UX and education standpoint to show `bun` commands at all until it works across the board though personally.

Just want to ++ this, adding any `bun` commands to the docs, when it doesn't work with everything, seems like a pretty confusing message to be sending to new users who won't understand this nuance... they'll start using `bun` because some docs commands have it, and then run into commands that don't work...

threepointone:
it's now been scoped to only add/install, so I think it's ready to go?

threepointone (approved):

MattieTK (approved):

Oxyjun (approved):

PR #29521, opened 2026-04-01
[DLP] Improve clarity across DLP policies pages

Oxyjun:
Improves clarity, fixes bugs, and adds missing context across all three pages in `src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/` based on an ELI5 analysis. All new claims were adversarially reviewed against source documentation.

**Changes across all three pages:**

- **index.mdx (Scan HTTP traffic)**
  - Add `GlossaryTooltip` for DLP on first use and cross-link Secure Web Gateway
  - Clarify the profile-vs-policy distinction inline ("a DLP profile defines what to detect; a Gateway HTTP policy defines what action to take")
  - Add 1-sentence "why" to each prerequisite (HTTP filtering routes traffic through Gateway; TLS decryption allows reading encrypted content)
  - Expand the caution note to explain why a profile alone does not start scanning
  - Narrate the example policy's combined-condition logic so readers can parse the table
  - Fix typo in example table: "Identifer" → "Identifier"
  - Add cross-link for "Zero Trust organization"
  - Note that logs show match metadata by default, not matched content
  - Define "false positive" inline in the reporting section

- **common-policies.mdx (Common DLP policies)**
  - Replace undefined "in-line DLP policies" with clear terminology and add prerequisite reminder
  - Explain the Allow-as-logger mechanism ("all matches — including Allow — are recorded in your HTTP request logs")
  - Link _Financial Information_ to predefined profiles on first use
  - Add identity provider cross-link and clarify that non-matching users are unaffected
  - Explain certificate pinning consequence and expand Do Not Inspect trade-off to include HTTP logging and AV scanning (per `http-policies/index.mdx:233`)
  - Replace vague "noise" with "false positives"
  - Connect the List creation step to the `not in list` operator usage

- **logging-options.mdx (Logging options)**
  - Add comparison table of three logging approaches (payload / AI prompt / Logpush) with encryption model and plan availability
  - Define "payload" explicitly as the HTTP request body on first use
  - Explain why DLP uses customer-managed encryption (Cloudflare cannot read your data)
  - Describe the two keys (public for encryption, private for decryption) before the generation step
  - Fix broken anchor `#1-generate-a-key-pair` → `#generate-a-key-pair`
  - Fix broken anchor `#4-view-payload-logs` → `#view-payload-logs`
  - Fix broken anchor `#send-http-requests-to-logpush-destination` → `#send-dlp-forensic-copies-to-logpush-destination`
  - Fix link text "these instructions" → descriptive text per style guide
  - Fix typo: "confident threshold" → "confidence threshold"
  - Define "false positive" and "true positive" inline
  - Fix terminology: "enable payload logging" → "enable AI prompt content logging" in the AI section
  - Add inline definition of "encrypted at rest"
  - Clarify that redaction replaces characters with X while preserving format
  - Add inline gloss for Logpush and explain the DLP Profile selector
  - Clarify Logpush security model: data encrypted in transit only, stored per destination policies
  - Note that any Gateway HTTP policy (not just DLP) can send forensic copies
  - Add consequence of losing private key: previously logged data becomes permanently unreadable

### DIFF
--- a/src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/common-policies.mdx
+++ b/src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/common-policies.mdx
@@ -33,7 +33,7 @@ For more information on what file formats DLP can scan, refer to [Supported file
 
 ## Block uploads/downloads for specific users
 
-You can configure access on a per-user or group basis by adding [identity-based conditions](/cloudflare-one/traffic-policies/identity-selectors/) to your policies. These selectors match against user attributes from your configured [identity provider](/cloudflare-one/identity-providers/).
+You can configure access on a per-user or group basis by adding [identity-based conditions](/cloudflare-one/traffic-policies/identity-selectors/) to your policies. These selectors match against user attributes from your configured identity provider.
 
 The following example blocks only contractors from uploading/downloading Financial Information to file sharing apps. Users who are not in the _Contractors_ group are not affected by this policy.
 
--- a/src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/common-policies.mdx
+++ b/src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/common-policies.mdx
@@ -10,11 +10,15 @@ head:
 
 import { GlossaryTooltip, Render } from "~/components";
 
-The following in-line DLP policies are commonly used to secure data in uploaded and downloaded files.
+The following DLP policies are commonly used to secure sensitive data in uploaded and downloaded files. They are built as [Gateway HTTP policies](/cloudflare-one/traffic-policies/http-policies/) using the [DLP Profile](/cloudflare-one/traffic-policies/http-policies/#dlp-profile) selector.
+
+Before using these policies, complete the [prerequisites for scanning HTTP traffic](/cloudflare-one/data-loss-prevention/dlp-policies/#prerequisites).
 
 ## Log uploads/downloads
 
-The **Allow** action functions as an implicit logger, providing visibility into where your sensitive data is going without impacting the end user experience. The following example scans for your enabled Financial Information profile entries when users upload or download data to file sharing apps.
+When you want to monitor where sensitive data is going before enforcing blocks, use the **Allow** action. In a Gateway HTTP policy, all matches — including Allow — are recorded in your [HTTP request logs](/cloudflare-one/data-loss-prevention/dlp-policies/#4-view-dlp-logs). This gives you visibility into sensitive data transfers without disrupting users.
+
+The following example logs any upload or download that matches your enabled [Financial Information](/cloudflare-one/data-loss-prevention/dlp-profiles/predefined-profiles/#financial-information) DLP profile entries when users interact with file sharing applications.
 
 | Selector           | Operator | Value                   | Logic | Action |
 | ------------------ | -------- | ----------------------- | ----- | ------ |
@@ -29,7 +33,9 @@ For more information on what file formats DLP can scan, refer to [Supported file
 
 ## Block uploads/downloads for specific users
 
-You can configure access on a per-user or group basis by adding [identity-based conditions](/cloudflare-one/traffic-policies/identity-selectors/) to your policies. The following example blocks only contractors from uploading/downloading Financial Information to file sharing apps.
+You can configure access on a per-user or group basis by adding [identity-based conditions](/cloudflare-one/traffic-policies/identity-selectors/) to your policies. These selectors match against user attributes from your configured [identity provider](/cloudflare-one/identity-providers/).
+
+The following example blocks only contractors from uploading/downloading Financial Information to file sharing apps. Users who are not in the _Contractors_ group are not affected by this policy.
 
 | Selector           | Operator | Value                   | Logic | Action |
 | ------------------ | -------- | ----------------------- | ----- | ------ |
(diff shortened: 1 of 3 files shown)


src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/common-policies.mdx
```diff
@@ -50,15 +56,15 @@ Many Android applications (such as Google Drive) use <GlossaryTooltip term="cert
    | Application                  | in       | _Google Drive_       | And   | Do Not Inspect |
    | Passed Device Posture Checks | in       | _OS Version Android_ |       |                |
 
-Android users can now use the app, but the app traffic will bypass DLP scanning.
+Android users can now use the app, but the app traffic will bypass Gateway inspection entirely — including DLP scanning, HTTP logging, and antivirus scanning.
```
Oxyjun, line 59: Language got strengthened here. Factcheck? From Opencode:

The existing docs at http-policies/index.mdx:233 already say verbatim:
> "you will lose the ability to log or block HTTP requests, apply DLP policies, and perform AV scanning."
So the PR's claim that DLP, HTTP logging, and AV scanning are all bypassed is directly supported by the existing docs.

However, there is one subtlety: gateway-logs/index.mdx:202 reveals that Do Not Inspect traffic still generates a minimal log entry recording the SNI and the Do Not Inspect action. So "HTTP logging" is not completely eliminated — detailed HTTP logging is. The PR says "bypass Gateway inspection entirely — including … HTTP logging" which could mislead readers into thinking zero log evidence exists. A more precise phrasing would be "full HTTP request logging."

Verdict: Core claim is accurate. Minor precision issue on "HTTP logging."

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/index.mdx
```diff
@@ -35,21 +37,21 @@ DLP Profiles may be used alongside other Cloudflare One rules in a [Gateway HTTP
...
+   | Selector    | Operator | Value                                                     | Logic | Action |
+   | ----------- | -------- | --------------------------------------------------------- | ----- | ------ |
+   | DLP Profile | in       | _Social Security, Insurance, Tax, and Identifier Numbers_ | And   | Block  |
+   | HTTP Method | in       | _POST_                                                    | And   |        |
+   | Application | not in   | _Workday_                                                 |       |        |
 
 4. Select **Create policy**.
 
-DLP scanning is now turned on.
+DLP scanning is now turned on for HTTP traffic matching this policy.
```
Oxyjun, line 50: In general - does creating the policy automatically mean it turns on?

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/logging-options.mdx
```diff
@@ -27,16 +37,16 @@ To generate a public/private key pair in the command line, refer to [these instr
 4. Select **Save**.
 
 :::note
-The matching private key is required to view logs. If you lose your private key, you will need to [generate](#1-generate-a-key-pair) and [upload](#2-upload-the-public-key-to-cloudflare) a new public key. The payload of new requests will be encrypted with the new public key.
+The matching private key is required to view logs. If you lose your private key, you will need to [generate](#generate-a-key-pair) and [upload](#upload-the-public-key-to-cloudflare) a new public key. The payload of new requests will be encrypted with the new public key. Previously logged data encrypted with the old key will be permanently unreadable.
```
Oxyjun, line 40: Net new content, needs factcheck

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/logging-options.mdx
```diff
@@ -86,7 +96,7 @@ DLP can detect and log the prompt topic sent to an AI tool.
 
 ### Turn on AI prompt content logging for a DLP policy
 
-You can enable payload logging for any Allow or Block HTTP policy that uses the [_Application_](/cloudflare-one/traffic-policies/http-policies/#application) selector with a supported [Application Granular Controls](/cloudflare-one/traffic-policies/http-policies/#granular-controls) application.
+You can enable AI prompt content logging for any Allow or Block HTTP policy that uses the [_Application_](/cloudflare-one/traffic-policies/http-policies/#application) selector with a supported [Application Granular Controls](/cloudflare-one/traffic-policies/http-policies/#granular-controls) application. This means your policy must target a specific AI application (such as ChatGPT) that Gateway can inspect at a granular level.
```
Oxyjun, line 99: The block of text has been rewritten by ELI5 with an AI slant, possibly due to the H3 and the H2. Needs factcheck.

marciocloudflare (approved):

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/common-policies.mdx
alexamavrogianis: If we want to be as precise as the factchecker would like we can rephrase to "including DLP scanning, the ability to block or log HTTP requests, and antivirus scanning." 

Sentiment is correct overall.

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/index.mdx
alexamavrogianis: Yes, by default new policies are saved in an active state.

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/logging-options.mdx
alexamavrogianis: Technically true, but doesn't seem super valuable to add. It seems the intent is to better explain what granular controls are, which would be better served by saying something like "When you use applications with granular controls, you can control the actions users take once they are within the application. For example, you can block uploads or sharing within ChatGPT."

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/logging-options.mdx
alexamavrogianis: Previously logged data will still be stored with the old key.

src/content/docs/cloudflare-one/data-loss-prevention/dlp-policies/logging-options.mdx
Oxyjun: Thanks Alexa. For this particular paragraph I'm going to just revert to the original text.

PR #29568, opened 2026-04-02
fix: Use safe shell loop pattern for file iteration

mvvmm:
## Summary

Updates shell loops in the potential-redirects-or-partials workflow to use a more robust iteration pattern for processing filenames.

## Changes

Replaced `for` loops with `while read` pattern:
- `for path in $CHANGED_FILES` → `while IFS= read -r path; do ... done <<< "$CHANGED_FILES"`
- `for path in $PARTIAL_FILES` → `while IFS= read -r path; do ... done <<< "$PARTIAL_FILES"`
- Added empty line checks to skip blank lines

## Why

The new pattern handles filenames with spaces and special characters more reliably. It processes each line exactly as returned from the GitHub API and follows shell scripting best practices for iterating over multi-line variables.


mvvmm:
/bonk Do you see any change in behavior here that would matter?

kodster28 (approved):

PR #29586, opened 2026-04-03
[Chore] Remove sync from post-install

kodster28:
### Summary

This is already included in dev and build by default, so we don't need on install. Saves ~30 sec when we miss the node_modules cache.

Screenshot 2026-04-03 at 3 03 27 PM

https://docs.astro.build/en/reference/cli-reference/#astro-sync


mvvmm (approved):

mvvmm:
/bonk is there any issue with removing sync from the posinstall script like this?

PR #29639, opened 2026-04-07
[workers] Document Durable Object Facets.

kentonv:
For now, this is documented as a feature of Dynamic Workers, since that is the only use case enabled so far.

### DIFF
--- a/src/content/docs/dynamic-workers/usage/durable-object-facets.mdx
+++ b/src/content/docs/dynamic-workers/usage/durable-object-facets.mdx
@@ -1,0 +1,187 @@
+---
+title: Durable Object Facets
+description: Run dynamically-loaded code with isolated persistent storage.
+pcx_content_type: how-to
+sidebar:
+  order: 5
+---
+
+import {
+	TypeScriptExample,
+	WranglerConfig,
+	Type,
+	MetaInfo,
+} from "~/components";
+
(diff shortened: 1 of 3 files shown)


src/content/docs/dynamic-workers/usage/durable-object-facets.mdx
```diff
@@ -0,0 +1,187 @@
...
+import { DurableObject } from "cloudflare:workers";
+
+// In production, this code would come from an AI agent, a database,
+// or user input — not a static string.
+const AGENT_CODE = `
+  import { DurableObject } from "cloudflare:workers";
+
+  export class App extends DurableObject {
+    fetch(request) {
+      let counter = this.ctx.storage.kv.get("counter") || 0;
```
irvinebroque: I get that this works and uses the sqlite db — but to those not familiar with DO — we've made a point about sqlite above, and it seems like a code example that uses SQL expresses that point more clearly to agents and humans

irvinebroque (approved):
see https://github.com/cloudflare/cloudflare-docs/pull/29732/ for diagram — think might help to visualize?

docs lgtm — updated the code example to use sqlite api

dinasaur404 (approved):

Oxyjun (approved):

PR #29720, opened 2026-04-09
Deprecate and redesign account analytics

Techsurfer1981:
Hey folks,
This page is completely outdated and needs to be updated ASAP. Context: PCX-21526

Thanks!


rianvdm:
cc @jsells1

PR #29736, opened 2026-04-10
[DNS] Use ELI5 on zone-transfers and nameservers

RebeccaTamachiro:
PCX-21040

### DIFF
--- a/src/content/docs/dns/nameservers/index.mdx
+++ b/src/content/docs/dns/nameservers/index.mdx
@@ -6,9 +6,9 @@ sidebar:
   label: About
 ---
 
-import { Render, GlossaryTooltip } from "~/components";
+import { Render } from "~/components";
 
-<GlossaryTooltip term="nameserver">Nameservers</GlossaryTooltip> are DNS servers that answer queries about the domains they are responsible for. When a visitor types your domain into a browser, the [DNS resolution process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through several server types and eventually reaches the authoritative nameservers for the final answer.
+Nameservers are DNS servers that answer queries about the domains they are responsible for. When a visitor types your domain into a browser, the [DNS resolution process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through several server types and eventually reaches the authoritative nameservers for the final answer.
 
 In the context of Cloudflare DNS, nameservers refer to authoritative nameservers — the servers that hold the definitive DNS records for your domain and provide the final response in DNS resolution. When a nameserver is authoritative for `example.com`, DNS resolvers will consider responses from this nameserver when a user tries to access `example.com`.
 
--- a/src/content/docs/dns/nameservers/index.mdx
+++ b/src/content/docs/dns/nameservers/index.mdx
@@ -45,7 +45,7 @@ Each zone that uses advanced nameservers is assigned a set of three nameservers 
 
 With [custom nameservers](/dns/nameservers/custom-nameservers/), your nameservers are hosted on your own domain (or domains) and, in this sense, are not Cloudflare branded.
 
-You provide fully qualified domain names — complete domain names like `ns1.mydomain.com` — for your nameservers, and Cloudflare assigns one IPv4 and one IPv6 address to each.
+You provide fully qualified domain names — complete domain names like `ns1.example.com` — for your nameservers, and Cloudflare assigns one IPv4 and one IPv6 address to each.
 
 :::caution
 
--- a/src/content/docs/dns/nameservers/index.mdx
+++ b/src/content/docs/dns/nameservers/index.mdx
@@ -8,12 +8,7 @@ sidebar:
 
 import { Render, GlossaryTooltip } from "~/components";
 
-<GlossaryTooltip term="nameserver">Nameservers</GlossaryTooltip> are DNS servers
-that answer queries about the domains they are responsible for. When a visitor
-types your domain into a browser, the [DNS resolution
-process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through
-several server types and eventually reaches the authoritative nameservers for
-the final answer.
+<GlossaryTooltip term="nameserver">Nameservers</GlossaryTooltip> are DNS servers that answer queries about the domains they are responsible for. When a visitor types your domain into a browser, the [DNS resolution process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through several server types and eventually reaches the authoritative nameservers for the final answer.
 
 In the context of Cloudflare DNS, nameservers refer to authoritative nameservers — the servers that hold the definitive DNS records for your domain and provide the final response in DNS resolution. When a nameserver is authoritative for `example.com`, DNS resolvers will consider responses from this nameserver when a user tries to access `example.com`.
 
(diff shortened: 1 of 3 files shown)


src/content/docs/dns/nameservers/index.mdx
```diff
@@ -6,21 +6,27 @@ sidebar:
...
+<GlossaryTooltip term="nameserver">Nameservers</GlossaryTooltip> are DNS servers
+that answer queries about the domains they are responsible for. When a visitor
+types your domain into a browser, the [DNS resolution
+process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through
+several server types and eventually reaches the authoritative nameservers for
+the final answer.
 
-In the context of Cloudflare DNS, nameservers refer to authoritative nameservers, which are the last stop in the DNS query resolution. When a nameserver is authoritative for `example.com`, it means that DNS resolvers will consider responses from this nameserver when a user tries to access `example.com`.
+In the context of Cloudflare DNS, nameservers refer to authoritative nameservers — the servers that hold the definitive DNS records for your domain and provide the final response in DNS resolution. When a nameserver is authoritative for `example.com`, DNS resolvers will consider responses from this nameserver when a user tries to access `example.com`.
```
RebeccaTamachiro: I think the whole point of the first paragraph being more vague is that it would be complimented by the second one. Whereas now, I feel like they are repetitive - I agree line 18 is clearer now, but it's basically saying the same as lines 15 and 16 (which were weirdly added as separate lines)

RebeccaTamachiro: I also think I preferred starting with the more familiar/ordinary-life fact (a user typing an address on their browser) then starting with a definition that goes `Nameservers are DNS servers that answer queries about...`

src/content/docs/dns/zone-setups/zone-transfers/index.mdx
```diff
@@ -10,18 +10,23 @@ head:
...
+Zone transfers allow you to use multiple DNS providers for the same domain to increase availability and fault tolerance. If one provider has an outage, the other can still answer DNS queries, keeping your domain available.
 
-With AXFR, the entire zone will be transferred from the primary to the secondary provider, even if only one record changes. With IXFR, only the changes will be transferred. Cloudflare supports both protocols.
+With zone transfers, your providers synchronize DNS records between themselves using one of two protocols:
 
-With zone transfers, you have two configuration options:
+- **Authoritative zone transfer ([AXFR](https://datatracker.ietf.org/doc/html/rfc5936))**: Copies the entire zone from the primary to the secondary provider, even if only one record changes.
```
RebeccaTamachiro:
```suggestion
- **Authoritative zone transfer ([AXFR](https://www.rfc-editor.org/rfc/rfc5936.html))**: Copies the entire zone from the primary to the secondary provider, even if only one record changes.
```

src/content/docs/dns/zone-setups/zone-transfers/index.mdx
```diff
@@ -10,18 +10,23 @@ head:
...
+- **Incremental zone transfer ([IXFR](https://datatracker.ietf.org/doc/html/rfc1995))**: Transfers only the changes since the last transfer, rather than the entire zone.
```
RebeccaTamachiro:
```suggestion
- **Incremental zone transfer ([IXFR](https://www.rfc-editor.org/rfc/rfc1995.html))**: Transfers only the changes since the last transfer, rather than the entire zone.
```

src/content/docs/dns/nameservers/index.mdx
```diff
@@ -8,19 +8,20 @@ sidebar:
 
 import { Render } from "~/components";
 
-As explained in [How DNS works](https://www.cloudflare.com/learning/dns/what-is-dns/), from the moment a user types an address (`www.example.com`) into their web browser, the resolution of a DNS query takes place. Also, the process behind DNS resolution involves different computers (or servers).
+Nameservers are DNS servers that answer queries about the domains they are responsible for. When a visitor types your domain into a browser, the [DNS resolution process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through several server types and eventually reaches the authoritative nameservers for the final answer.
```
hannes-cf:
```suggestion
Nameservers are DNS servers that answer DNS queries about the domains they are authoritative for. When a visitor types your domain into a browser, the [DNS resolution process](https://www.cloudflare.com/learning/dns/what-is-dns/) passes through several server types and eventually reaches the authoritative nameservers for the final answer.
```

src/content/docs/dns/nameservers/index.mdx
```diff
@@ -8,19 +8,20 @@ sidebar:
...
+
 <Render file="ns-static-ips" product="dns" />
 :::
 
 ## Authoritative nameservers offering
 
 Within Cloudflare, and depending on your plan, you can choose between using Cloudflare-branded nameservers or setting up your own custom nameservers. The names for Cloudflare-branded nameservers are automatically assigned and cannot be changed.
 
-Regardless of the type you choose, for these nameservers to be authoritative for your domain, you need to [update your domain nameservers](/dns/nameservers/update-nameservers/). Updating your nameservers is required to activate your domain on Cloudflare and use most of our [application services](/fundamentals/concepts/how-cloudflare-works/).
+Regardless of the type you choose, for these nameservers to be authoritative for your domain, you need to [update your domain nameservers](/dns/nameservers/update-nameservers/). Updating your nameservers is required to activate your domain on Cloudflare and use most of Cloudflare's [application services](/fundamentals/concepts/how-cloudflare-works/), such as proxying, caching, and security features.
```
hannes-cf:
```suggestion
Regardless of the type you choose, for these nameservers to be authoritative for your domain, you need to [update your domain nameservers](/dns/nameservers/update-nameservers/), typically where you registered your domain. Updating your nameservers is required to activate your domain on Cloudflare and use most of Cloudflare's [application services](/fundamentals/concepts/how-cloudflare-works/), such as proxying, caching, and security features.
```

src/content/docs/dns/zone-setups/zone-transfers/index.mdx
```diff
@@ -10,18 +10,23 @@ head:
...
+
+You have two configuration options for zone transfers:
 
 - [Cloudflare as Primary](/dns/zone-setups/zone-transfers/cloudflare-as-primary/): Cloudflare is your primary DNS provider and performs outgoing zone transfers to your secondary DNS provider(s).
 - [Cloudflare as Secondary](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/): Cloudflare is your secondary DNS provider and initiates incoming zone transfers from your primary DNS provider.
 
 ## Peer DNS server
 
-Peer DNS servers can be used as primary and secondary external DNS servers. The same peer can be linked to multiple primary and secondary zones. Each peer can be associated with only one Transaction Signature (TSIG).
+A peer DNS server is the external DNS provider that participates in zone transfers with Cloudflare. The same peer can be linked to multiple primary and secondary zones. Each peer can be associated with only one Transaction Signature (TSIG) — an authentication mechanism that uses a shared secret to verify zone transfer messages between providers.
```
hannes-cf, line 29: "peer DNS server" is Cloudflare jargon, so maybe we should call that out?

RebeccaTamachiro: Since it's being defined explicitly, I think it's ok to keep it as-is. Also because, since `Peer DNS servers` is used in the UI, it's good to make it easier for users to find it in docs.

Oxyjun (approved):

PR #29782, opened 2026-04-12
Dane/support knowledge gaps - check with SRhea before approving

TownLake:
### Documentation checklist


nikitacano (approved):

src/content/docs/waf/troubleshooting/phase-interactions.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## Recommended troubleshooting workflow
+
+When a WAF decision looks wrong:
+
+1. Check for earlier request rewrites.
+2. Check for matching IP Access rules.
+3. Confirm whether the request was expected to stop in the custom rules phase or skip later phases.
+4. Review whether a managed rule still ran after a custom rule match.
+5. Use [Trace](/waf/tools/trace/) when available to confirm the actual phase-by-phase result.
```
pedrosousa:
```suggestion
5. Use [Trace](/rules/trace-request/) when available to confirm the actual phase-by-phase result.
```
Original URL does not exist.

src/content/docs/waf/reference/legacy/old-rate-limiting/upgrade.mdx
```diff
@@ -28,6 +28,8 @@ You must now use the [Rulesets API](/ruleset-engine/rulesets-api/) and the [`clo
 
 - **Advanced scope expressions:** The previous version of Rate Limiting allowed you to scope the rules based on a single path and method of the request. In the new version, you can write rules similar to [WAF custom rules](/waf/custom-rules/), combining multiple parameters of the HTTP request.
 
+- **Counter scope:** The previous version relied on a global counter model. The new version of rate limiting uses counters scoped per data center, with `cf.colo.id` always included as a characteristic. This means thresholds can behave differently for traffic distributed across multiple Cloudflare locations. For more information, refer to [How Cloudflare determines the request rate](/waf/rate-limiting-rules/request-rate/).
```
danielegm: this isn't correct, we never had global counter for RL, it has always been per colo (or even metal)
In the new system it's just explicit as we added cf.colo.id

src/content/docs/waf/troubleshooting/phase-interactions.mdx
```diff
@@ -0,0 +1,80 @@
...
+- A custom rule may appear to "not fire" for a specific IP even though the expression is correct.
+- Allowlisting a source too early can also prevent other WAF logic from running.
+
+### Resolution
+
+If a request is unexpectedly bypassing a custom rule, check for matching IP Access rules first.
+
+## Skip and Allow do not behave the same way
+
+In the custom rules phase, **Allow** and **Skip** solve different problems.
```
danielegm: This is incorrect. Allow was used in Firewall Rules (deprecated and almost all customers migrated over to the Custom RUles). Skip is the new allow but for custom rules

danielegm: IP Access Rules has allow though, and that is still valid

src/content/docs/waf/reference/legacy/old-rate-limiting/upgrade.mdx
pedrosousa: Removed the bullet about RL counter scope.

src/content/docs/dns/troubleshooting/dns-issues.mdx
```diff
@@ -57,6 +57,20 @@ In rare cases, the DNS resolver in the client requesting the URL might fail to r
...
+
+For new records, many resolvers use the zone's negative caching TTL rather than the new record's TTL. This negative cache duration is typically taken from the `MINIMUM` field in the zone's `SOA` record, as described in [RFC 2308](https://datatracker.ietf.org/doc/html/rfc2308).
+
+This means:
+
+- Lowering the TTL on the new record does not clear an existing negative cache entry.
+- Flushing your local DNS cache only affects your own device.
+- Different recursive resolvers may recover at different times depending on when they cached the negative answer.
+
+To confirm, query the zone's `SOA` record and check the negative caching value. Then wait for that interval to expire before testing again.
```
hannes-cf:
```suggestion
If you recently created a DNS record and resolvers still return `NXDOMAIN` (Non-Existent Domain) or no answer, it is likely because a negative response is currently stored in the resolver's cache.

When a resolver is queried for a hostname that has no DNS records yet, it caches the empty response so it does not have to ask the authoritative nameserver again immediately. This is known as negative caching.

For newly created records:

- The resolver might not have cached the new record yet. Instead, it is using a prior `NXDOMAIN` cache entry that says "this record does not exist," which was generated if the hostname was queried before you created the record.
- The duration of this negative cache is determined by the `MINIMUM` field in your zone's SOA record (per [RFC 2308](https://datatracker.ietf.org/doc/html/rfc2308)), not the TTL of the record you just created. Different resolvers may cache for varying durations.

This means:

- Lowering the TTL on your new record will not speed up resolution if a negative cache entry already exists; the resolver will only see your new TTL after the old negative entry expires.
- Flushing your local DNS cache only affects your specific device; the upstream recursive resolver (for example, your ISP or a public provider) still holds the negative result.
- Propagation appears uneven because different resolvers may have queried the name at different times, apply different negative cache TTLs, or have no negative cache entry at all.

The exact behavior differs per resolver, but to estimate how long you need to wait, query your zone's SOA record and look at the last value (the `MINIMUM` field). You must wait for that interval to pass since the last `NXDOMAIN` query before the new record will consistently resolve.

To verify the record resolves correctly, you can purge the cache for public resolvers and query the record. If this works, other resolvers will eventually start resolving as well:

- [Purge 1.1.1.1 cache](https://one.one.one.one/purge-cache/)
- [Purge 8.8.8.8 cache](https://dns.google/cache)
- [Query 8.8.8.8](https://dns.google/)
- [Query and refresh OpenDNS cache](https://cachecheck.opendns.com/)

#### Further debugging

To verify the record was correctly created, query Cloudflare's authoritative nameservers directly:

```sh
# Find the authoritative nameservers for your zone
dig @1.1.1.1 example.com NS +short
```

```sh
# Query the authoritative nameserver for your new record
dig @hera.ns.cloudflare.com mynewrecord.example.com A
```

Querying the authoritative nameserver directly bypasses resolver caching. If the record is returned, resolvers will eventually start returning it as well. If the record does not appear, verify the record exists in the Cloudflare dashboard and that the hostname matches exactly.
```

Comment (hannes-cf) [reply src/content/docs/dns/troubleshooting/dns-issues.mdx]: The original text was a bit confusing in my opinion. So here is a revised and more technically accurate suggestion, partly written by Gemini, partly with input from the team.

Comment (colinbendell) [reply src/content/docs/dns/troubleshooting/dns-issues.mdx]: can I suggest that we include a `dig` command that will reveal if it is being cached? Giving them lots of words doesn't help.

```
$ dig +noall +answer +authority     asdf.walmart.com 

walmart.com.		256	IN	SOA	..
```
In this example, the negative cache response will continue for 256s

cferike (commented):
SSL section looks good to me. The Origin CA expiry note is accurate, there is no notification type for Origin CA certificate expiration today.

For reference:
- SECENG-4047 (https://jira.cfdata.org/browse/SECENG-4047) has tracked this gap since 2017
- SHIP-9043 (https://jira.cfdata.org/browse/SHIP-9043) (Origin CA rebuild) includes expiration notifications as a planned feature

Good addition, customers hit 526 errors with no warning when these expire.

PR #29823, opened 2026-04-13
fix: link validation failures for fork PRs

mvvmm:
### Summary

The "Link validation" step in the `post-build` job was failing for fork PRs with "Resource not accessible by integration". This is a GitHub security restriction: fork PRs receive a read-only `GITHUB_TOKEN` that cannot write PR comments — this cannot be overridden by maintainer approval.

The original script mixed two concerns: failing the job on broken links, and posting a PR comment with the report. This PR separates them:

**`bin/check-link-validation/index.ts`** (new) — reads the report file from disk and calls `core.setFailed` if broken links exist. No GitHub API calls, no token required. Runs for all PRs including forks.

**`bin/post-link-validation-comment/index.ts`** (mostly unchanged) — posts/updates/deletes the PR comment.

**CI changes:**
- `post-build` retains "Link validation" using the new script — always runs, always fails on broken links, works for forks
- "Download link validation report" and "Post link validation comment" move into the `notify` job, both guarded with `if: ... full_name == github.repository` so they are skipped for forks. The comment step has `continue-on-error: true`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,9 +202,7 @@ jobs:
           path: .starlight-links-validator
 
       - name: Link validation
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: npx tsx bin/post-link-validation-comment/index.ts
+        run: npx tsx bin/check-link-validation/index.ts
 
   notify:
     name: Notify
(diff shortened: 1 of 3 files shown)


src/content/docs/workers/demos.mdx
```diff
@@ -6,6 +6,8 @@ sidebar:
 
 ---
 
+[This is a broken link](/workers/this-page-does-not-exist/)
```
ahaywood: Is this supposed to here?

```suggestion
```

ahaywood (commented):
1 issue needs fixing before merge (the broken link test artifact). The CI refactor itself looks correct — separating fail-on-broken-links from comment-posting is the right fix for the fork token restriction. Inline suggestion posted below.

ahaywood (approved):
Already talked to @mvvmm about a couple of packages that will need to be deleted before merging into production

PR #29872, opened 2026-04-14
[Style guide] Added info about cross-linking

dcpena:
Added guidance about how to best use cross-linking. Addresses PCX-21223.


caley-b (approved):
LGTM!

PR #29907, opened 2026-04-15
[API Shield, Bots, China Network] Adding front matter for App Sec products

dcpena:
Addresses PCX-21371.


caley-b (approved):
Thanks for adding the products metadata!

PR #29926, opened 2026-04-15
[DNS Firewall] Added product & description to frontmatter

dcpena:
Added product & description to frontmatter to optimize for AEO. PCX-21371.


caley-b (approved):
And another one!

PR #29962, opened 2026-04-16
[DNS] Clarify full-to-partial conversion steps

hannes-cf:
## Summary

- clarify that apex proxying only works when the apex points directly to Cloudflare Anycast IPs, and that this is only recommended with Static IPs or BYOIP
- make the conversion flow clearer by separating work done at the new authoritative DNS provider from cleanup in Cloudflare
- update the guide to explain the verification TXT record requirement and the active-zone assumption for this conversion flow

### DIFF
--- a/src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
+++ b/src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
@@ -22,7 +22,7 @@ This guide assumes your zone is already in an [active status](/dns/zone-setups/r
 
 Make sure you consider the following:
 
-- You can only proxy the zone apex (`example.com`) if your authoritative DNS provider points it directly to Cloudflare Anycast IPs. Cloudflare only recommends this if you use [Static IPs](/byoip/concepts/static-ips/) or [Bring Your Own IP (BYOIP)](/byoip/). Otherwise, you can only proxy subdomains.
+- A CNAME setup requires a CNAME record for each proxied hostname, but [CNAME records are not allowed on the zone apex](https://datatracker.ietf.org/doc/html/rfc1912#section-2.4) (`example.com`). You can only proxy the zone apex if your authoritative DNS provider supports [CNAME Flattening](https://blog.cloudflare.com/introducing-cname-flattening-rfc-compliant-cnames-at-a-domains-root/) (or an equivalent like ALIAS/ANAME records), or if you create A/AAAA records pointing the apex directly to Cloudflare [anycast IP addresses](/fundamentals/concepts/cloudflare-ip-addresses/). Cloudflare only recommends the A/AAAA approach if you use [Static IPs](/byoip/concepts/static-ips/) or [Bring Your Own IP (BYOIP)](/byoip/), because standard Cloudflare anycast IPs can change. Otherwise, you can only proxy subdomains.
 - On the dashboard, you will only be able to create A, AAAA, and CNAME records, which are the DNS record types that can be [proxied](/dns/proxy-status/).
 - You should plan for SSL/TLS certificates. If you are only using [Universal SSL](/ssl/edge-certificates/universal-ssl/) prior to converting your zone, a certificate will be provisioned for your subdomains only after each of the respective DNS records are proxied. If your domain is sensitive to downtime, instead of using Universal SSL, consider using an [advanced certificate](/ssl/edge-certificates/advanced-certificate-manager/) with [delegated DCV](/ssl/edge-certificates/changing-dcv-method/methods/delegated-dcv/#setup).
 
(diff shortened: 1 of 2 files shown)


src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
```diff
@@ -39,32 +45,51 @@ Make sure you consider the following:
 
    </Details>
 
-4. Remove any previously existing A, AAAA, or CNAME records referencing the hostnames you want to proxy through Cloudflare. For these hostnames, leave only the records pointing to `{your-hostname}.cdn.cloudflare.net`.
-
-5. Confirm you have the correct record for every subdomain that should be proxied through Cloudflare.
-
```
RebeccaTamachiro, line 45: Were these removed intentionally? IIRC, we have this instruction here (and it's intended to be done at the new primary DNS provider) so that customers don't get confused seeing A/AAAA records on a given hostname, when it is actually being proxied through Cloudflare due to a CNAME pointing to `{your-hostname}.cdn.cloudflare.net `

src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
```diff
@@ -39,32 +45,51 @@ Make sure you consider the following:
...
+
+    	A verification record for `example.com` might be:
+
+    	| Type | Name                            | Content             |
+    	| ---- | ------------------------------- | ------------------- |
+    	| TXT  | `cloudflare-verify.example.com` | 966215192-518620144 |
+
+    	</Details>
+
+    </TabItem>
```
RebeccaTamachiro: Need to come back to this. Right now, this isn't working because it got added inside the Dashboard tab, so the API steps are broken/ do not show the instructions to verify domain ownership.

Image

src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
```diff
@@ -6,28 +6,34 @@ sidebar:
...
+This guide assumes your zone is already in an [active status](/dns/zone-setups/reference/domain-status/#active).
+
+:::
+
 ## Before you begin
 
 Make sure you consider the following:
 
-- It will not be possible to use Cloudflare's reverse proxy on the zone apex (`example.com`), only on subdomains.
+- A CNAME setup requires a CNAME record for each proxied hostname, but [CNAME records are not allowed on the zone apex](https://datatracker.ietf.org/doc/html/rfc1912#section-2.4) (`example.com`). You can only proxy the zone apex if your authoritative DNS provider supports [CNAME Flattening](https://blog.cloudflare.com/introducing-cname-flattening-rfc-compliant-cnames-at-a-domains-root/) (or an equivalent like ALIAS/ANAME records), or if you create A/AAAA records pointing the apex directly to Cloudflare [anycast IP addresses](/fundamentals/concepts/cloudflare-ip-addresses/). Cloudflare only recommends the A/AAAA approach if you use [Static IPs](/byoip/concepts/static-ips/) or [Bring Your Own IP (BYOIP)](/byoip/), because standard Cloudflare anycast IPs can change. Otherwise, you can only proxy subdomains.
```
RebeccaTamachiro: Curious if this scenario where the new authoritative DNS provider supports CNAME Flattening is true/ would allow using this setup for the zone apex. Does it even make sense to want to do that?
But anyways, I was actually just hoping to add some more clarity around CF anycast IPs. Will suggest a consolidated version now.

RebeccaTamachiro:
Was working on this review but got into a weird state with git, so moved my suggestions to #30257

src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
hannes-cf: Yes. These got essentially moved to the new `Clean up DNS records on Cloudflare` section. At least that's how I understood 4. Which is to ensure you have matching pairs on new DNS provider + zone on Cloudflare like this:

```
NEW DNS Provider
www CNAME www.example.com.cdn.cloudflare.net

Cloudflare
www A     192.0.2.1 (origin IP)
```

I keep re-reading 4. and 5. and it really doesn't make any sense to me to keep it there. The prep new DNS provider section is complete without 4. and 5.

src/content/docs/dns/zone-setups/conversions/convert-full-to-partial.mdx
hannes-cf: Yes this version is true as is. For a CNAME setup we on the Cloudflare side don't care about the external DNS provider. We see two things:

1) DNS queries against `www.example.com.cdn.cloudflare.net` and then strip the suffix and try to find a zone that matches `www.example.com`. So the same would work for `example.com.cdn.cloudflare.net` if the external DNS provider supports CNAME flattening (or something similar) that allows the user to add that target to the zone apex there.

2) The HTTP request for hostnames like `www.example.com` or `example.com`. Once we see these, we don't care how DNS got resolve previously. So if I hardcode all of these hostnames to Cloudflare IPs, I can also achieve that HTTP requests will make it there. But hardcoding (or using A/AAAA records) is only recommended if IPs are static.

RebeccaTamachiro (approved):

PR #29992, opened 2026-04-16
[Argo Smart Routing & APO] Added product and description to frontmatter

dcpena:
Added product & description to frontmatter to optimize for AEO. PCX-21372.


caley-b (approved):
Let's do this.

PR #30016, opened 2026-04-17
Fix broken link to DigiCert AWS SSL installation guide

nanookclaw:
The link to DigiCert's AWS SSL installation guide had a typo ('certficate' instead of 'certificate'), causing a 404. This fixes the URL to the correct one.


RebeccaTamachiro:
Hey there 👋 Thanks for flagging. This and other links should already be fixed as per #29958

PR #30018, opened 2026-04-17
[Workers] Document Workers Caching

danlapid:
Add docs for Workers Caching — caching HTTP responses from a Worker so that Cloudflare serves them without executing the Worker.

New pages under src/content/docs/workers/cache/:

- configuration.mdx — enabling caching, Cache-Control semantics, TTL and stale-while-revalidate guidance, cf-cache-status, Cache-Tag
- cache-keys.mdx — what the cache key includes, multi-tenant safety with ctx.props, invalidating cache across deployments
- purge.mdx — ctx.cache.purge() with tags, pathPrefixes, and purgeEverything; versioned purging with the version metadata binding
- debugging.mdx — diagnosing why caching is not behaving as expected
- limitations.mdx — unsupported scenarios, limits, relationship to zone cache, the Cache API, and fetch() subrequests

index.mdx as a landing page with a quickstart showing wrangler deploy + curl observing HIT/MISS, plus a Durable Object wrapping pattern for caching DO responses.

workers/platform/pricing.mdx gets a worked example showing CPU savings with an 80% cache hit rate.

### Documentation checklist

### DIFF
--- a/src/content/docs/cache/interaction-cloudflare-products/workers.mdx
+++ b/src/content/docs/cache/interaction-cloudflare-products/workers.mdx
@@ -14,6 +14,10 @@ head:
 
 You can use [Workers](/workers/) to customize cache behavior on Cloudflare's network. Workers run as middleware in the request lifecycle — a single Worker handles both the request and response phases. When a request arrives, it hits the Worker before the cache is checked. The Worker can modify the incoming request (for example, rewrite the URL or add headers), then call `fetch()` to continue the request through the cache. When the response comes back — whether from cache or from the origin server — the Worker can also modify the response before it is sent to the visitor.
 
+:::note
+This page describes how Workers interact with a zone's Cloudflare Cache. You can can also enable **[Workers Cache](/workers/cache/)** for your Worker — a cache that sits in front of the Worker itself, so that Cloudflare returns a cached response without running the Worker. Use Workers Caching to cache the output of a Worker's logic directly, independent of any zone configuration.
+:::
+
 The diagram below illustrates a common interaction flow between Workers and Cache.
 
 ![Workers and cache flow example flow diagram.](~/assets/images/cache/workers-cache-flow.png)
--- a/src/content/docs/workers/cache/cache-keys.mdx
+++ b/src/content/docs/workers/cache/cache-keys.mdx
@@ -1,0 +1,254 @@
+---
+title: Cache keys
+pcx_content_type: concept
+description: How Workers Caching builds cache keys, with guidance for service bindings, multi-tenant Workers, and gradual deployments.
+sidebar:
+  order: 2
+---
+
+import { TypeScriptExample, WranglerConfig } from "~/components";
+
+Every cached response is stored under a **cache key**. When a request arrives, Cloudflare computes a cache key for it and looks it up — on a hit, the stored response is returned; on a miss, your Worker runs and its response is stored under that key for next time.
+
+Two requests that produce the same cache key share the same cached response. Two requests that produce different cache keys get independent cached entries.
+
+This page explains what Workers Caching puts into the cache key, why each component is there, and how to reason about it when designing your Worker.
+
+## What goes into the cache key
+
+Workers Caching keys responses by:
+
+- The **target entrypoint** — which specific [named entrypoint](/workers/runtime-apis/bindings/service-bindings/rpc/#named-entrypoints) of the Worker received the request. A `default` export and an exported class are different entrypoints and do not share a cache even if they produce identical responses.
+- The **path and query string** of the request URL. Query parameter order matters — `?a=1&b=2` and `?b=2&a=1` are different cache keys. Trailing slashes matter too.
+- The **Worker version**, by default. Each deployed version has its own cache, so a new deployment does not serve responses written by a previous version. You can turn this off with [`cache.cross_version_cache`](/workers/cache/configuration/#cross-version-caching) to share cached responses across versions. See [Invalidating cache across deployments](#invalidating-cache-across-deployments).
+- The invocation's [`ctx.props`](/workers/runtime-apis/bindings/service-bindings/rpc/#ctxprops), when the Worker is invoked through a service binding or RPC. See [Multi-tenant safety with `ctx.props`](#multi-tenant-safety-with-ctxprops).
+
+As anti-cache-poisoning measures, the key also includes:
+
+- The `x-http-method-override`, `x-http-method`, and `x-method-override` request headers.
+- The `x-forwarded-host`, `x-host`, `x-forwarded-scheme` (unless its value is `http` or `https`), `x-original-url`, `x-rewrite-url`, and `forwarded` request headers.
+- The value of the `Cloudflare-Workers-Version-Key` request header. This header is not set by Cloudflare automatically — it is only meaningful if a caller (for example, an upstream Worker or proxy) chooses to include it to explicitly partition the cache further. This is independent of the automatic per-version keying described above, which is controlled by [`cache.cross_version_cache`](/workers/cache/configuration/#cross-version-caching).
+
+These three bullets are not something you should normally need to reason about. Some frameworks interpret the method-override and URL-rewrite headers as overriding the effective method or URL of a request, which can lead to [cache poisoning](https://portswigger.net/research/practical-web-cache-poisoning) if two requests differ only in those headers but produce materially different responses. Including them in the cache key ensures a poisoned entry only affects requests that carry the same poisoned header.
+
+Requests that differ only in request headers that are not part of the cache key (for example, `User-Agent`, `Accept-Language`, `Cookie`, or `Authorization`) return the same cached response. This is usually what you want — you do not want every user agent string or language preference producing a separate cache entry. If you do need content negotiation, set [`Vary`](/workers/cache/configuration/#vary) on the response, or handle it inside your Worker and produce a canonical response per URL.
+
+Notably, the cache key does **not** include:
+
+- **The HTTP method.** `GET` and `HEAD` requests for the same URL share a single cache entry. A `HEAD` request can be served from a `GET` fill (Cloudflare returns the cached headers without the body). In the other direction, a `HEAD` request on a cold cache is converted to a `GET` internally so the full asset is fetched and stored — a subsequent `GET` then hits the entry that `HEAD` populated. (`POST`, `PUT`, `PATCH`, and `DELETE` are never cached at all, so the question does not arise for them.)
+- **The request's host.** The Worker's cache is keyed by path and query string, not the full URL. See [The cache belongs to the Worker, not to a domain](#the-cache-belongs-to-the-worker-not-to-a-domain).
+- **The request body.** Since only `GET` and `HEAD` are cacheable, this is rarely relevant — but worth noting if your Worker reads `request.body` on a cacheable method, the body does not partition the cache.
+
+At launch, you cannot inspect the exact cache key Cloudflare computed for a request. The primary signals you have for understanding cache behavior are the `Cf-Cache-Status` response header and per-invocation cache-hit information in the [Workers observability dashboard](/workers/observability/). See [Inspecting the cache key](#inspecting-the-cache-key).
+
+## The cache belongs to the Worker, not to a domain
+
+A Worker is a zoneless entity. It can be invoked through several different paths:
+
+- Directly on a `workers.dev` subdomain.
+- Through a [route](/workers/configuration/routing/routes/) on any zone you control.
+- Through a [custom domain](/workers/configuration/routing/custom-domains/) — and you can bind the same Worker to many custom domains.
+- Through a [service binding](/workers/runtime-apis/bindings/service-bindings/) from another Worker, with an arbitrary placeholder hostname in the URL.
+
+Workers Caching treats all of these as the same Worker and uses a single shared cache across them. The cache key does not include the host, so a request to `/api/users/42` hits the same cached entry whether it came in through `api.example.com`, `api.example.net`, a service binding, or a `workers.dev` URL.
+
+This is the behavior you almost always want. A Worker's responses are a function of its code and its inputs, not of which domain the request arrived through — so caching them once and serving that response back to every ingress path maximizes the cache hit rate without losing correctness.
+
+If you genuinely need different cached responses for the same path on different hostnames — for example, white-labeled tenants where `tenant-a.example.com/index` and `tenant-b.example.com/index` must produce different content — the cache key does not do this for you automatically. Instead, distinguish the tenants at your gateway Worker and pass the tenant identifier via `ctx.props`, which _is_ part of the cache key.
+
+## Invalidating cache across deployments
+
+By default, the currently invoked Worker version **is** part of the cache key. Each deployed version has its own cache, so:
+
+- A new deployment starts from a cold cache and never serves responses that a previous version wrote.
+- Cache-affecting changes apply immediately when the new version goes live — you do not have to purge anything to stop serving old content.
+- During a [gradual deployment](/workers/configuration/versions-and-deployments/gradual-deployments/), the old and new versions populate independent caches, so a slice of traffic on the new version never receives the old version's responses.
+
+This is the default because it is the simplest behavior to reason about. The trade-off is that **cache hit rate resets on every deployment** — the first requests to a new version are misses while its cache fills. This is the most common reason a Worker's cache hit rate drops right after a deploy.
+
+### Share the cache across versions
+
+If you deploy frequently and your responses rarely change between deployments, throwing away a warm cache on every deploy is wasteful. Set [`cache.cross_version_cache`](/workers/cache/configuration/#cross-version-caching) to `true` to drop the version from the cache key and share cached responses across versions. A response written by version A is then still served after version B is deployed, as long as its TTL has not expired.
+
+This maximizes cache hit rate at the expense of slower rollouts: because a deployment no longer invalidates the cache, a change that alters response content will not take effect for already-cached entries until they expire or you purge them. When you have `cross_version_cache` enabled and need a deployment to take effect immediately, use one of the two tools below.
+
+### Tag responses by version, purge the tag on rollback
+
+If you want fine-grained control, tag each cached response with the Worker version that produced it. Later, purging that version tag removes every entry that version wrote, without affecting cached responses from other versions.
+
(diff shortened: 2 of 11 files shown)


src/content/docs/workers/cache/cache-keys.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+Workers Caching treats all of these as the same Worker and uses a single shared cache across them. The cache key does not include the host, so a request to `/api/users/42` hits the same cached entry whether it came in through `api.example.com`, `api.example.net`, a service binding, or a `workers.dev` URL.
+
+This is the behavior you almost always want. A Worker's responses are a function of its code and its inputs, not of which domain the request arrived through — so caching them once and serving that response back to every ingress path maximizes the cache hit rate without losing correctness.
+
+If you genuinely need different cached responses for the same path on different hostnames — for example, white-labeled tenants where `tenant-a.example.com/index` and `tenant-b.example.com/index` must produce different content — the cache key does not do this for you automatically. Instead, distinguish the tenants at your gateway Worker and pass the tenant identifier via `ctx.props`, which _is_ part of the cache key.
+
+## Invalidating cache across deployments
```
zaidoon1, line 59: I believe this is changing/not true anymore by default?

src/content/docs/workers/cache/examples.mdx
```diff
@@ -0,0 +1,569 @@
...
+
+		if (request.method === "POST" && url.pathname === "/record") {
+			// Write path — bypass the cached entrypoint, hit the Durable
+			// Object directly, then invalidate the cached leaderboard
+			// so the next read returns fresh data.
+			const id = env.LEADERBOARD.idFromName("global");
+			const stub = env.LEADERBOARD.get(id);
+			const result = await stub.fetch(request);
+
+			await ctx.cache.purge({ tags: ["leaderboard"] });
```
zaidoon1: Can we double-check this pattern? The cache entry is owned by `CachedLeaderboard`, but the purge is called from `default`, and the config above disables caching on `default`. Since purges are scoped to the calling entrypoint, I do not think this invalidates `CachedLeaderboard` entries as written. I think this example needs to show the purge happening from the entrypoint that owns the cached response, or use a different invalidation pattern for gateway + cached-inner-entrypoint setups.

src/content/docs/workers/cache/purge.mdx
```diff
@@ -0,0 +1,369 @@
...
+} satisfies ExportedHandler;
+```
+
+</TypeScriptExample>
+
+The rest of this page uses `ctx.cache.purge(...)` in most examples because those examples already have `ctx` in scope. If you prefer the import form, substitute `cache.purge(...)` — nothing else changes.
+
+## Purge modes
+
+`purge()` accepts one or more of the following fields:
```
zaidoon1: I think this is a little too broad. `tags` and `pathPrefixes` can be combined, but `purgeEverything` should be exclusive. Maybe this should say that `purge()` accepts either `purgeEverything: true`, or one or both of `tags` and `pathPrefixes`.

src/content/docs/workers/cache/purge.mdx
```diff
@@ -0,0 +1,369 @@
...
+		});
+
+		return new Response("Purged", { status: 200 });
+	},
+} satisfies ExportedHandler;
+```
+
+</TypeScriptExample>
+
+Entries in `pathPrefixes` are **paths**, not full URLs. Do not include a scheme or host — `https://example.com/blog/` will not match. Use leading slashes to match from the root of the request path.
```
zaidoon1: I think "will not match" undersells what happens here. A full URL should be treated as invalid input, not as a path prefix that simply misses. Also, leading slashes are normalized, so `/images` and `images` should behave the same. Could we word this more like: path prefixes must not include a scheme, host, query string, or fragment; a leading slash is optional but recommended for clarity?

irvinebroque (approved):

src/content/docs/cache/interaction-cloudflare-products/workers.mdx
```diff
@@ -14,6 +14,10 @@ head:
 
 You can use [Workers](/workers/) to customize cache behavior on Cloudflare's network. Workers run as middleware in the request lifecycle — a single Worker handles both the request and response phases. When a request arrives, it hits the Worker before the cache is checked. The Worker can modify the incoming request (for example, rewrite the URL or add headers), then call `fetch()` to continue the request through the cache. When the response comes back — whether from cache or from the origin server — the Worker can also modify the response before it is sent to the visitor.
 
+:::note
+This page describes how Workers interact with a zone's Cloudflare Cache. Workers can also opt in to **[Workers Caching](/workers/cache/)** — a cache that sits in front of the Worker itself, so that Cloudflare returns a cached response without running the Worker. Use Workers Caching to cache the output of a Worker's logic directly, independent of any zone configuration.
```
irvinebroque:
```suggestion
This page describes how Workers interact with a zone's Cloudflare Cache. You can can also enable **[Workers Cache](/workers/cache/)** for your Worker — a cache that sits in front of the Worker itself, so that Cloudflare returns a cached response without running the Worker. Use Workers Caching to cache the output of a Worker's logic directly, independent of any zone configuration.
```

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+		return new Response("Purged", { status: 200 });
+	},
+} satisfies ExportedHandler;
+```
+
+</TypeScriptExample>
+
+You can also [import `cache` from `cloudflare:workers`](/workers/cache/purge/#two-ways-to-call-purge) and call `cache.purge({...})` when you do not have `ctx` in scope — for example, from a utility module. For all purge modes and patterns, refer to [Purging the cache](/workers/cache/purge/).
+
+## Billing
```
irvinebroque:
```suggestion
## Pricing
```

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+
+Workers Caching has no separate pricing. Requests to a Worker with caching enabled are billed at the standard [Workers request rate](/workers/platform/pricing/) — the same per-request rate as any other request to your Worker — whether the response comes from cache or from your Worker. There is no charge beyond the standard request rate. **CPU time is only billed when your Worker runs** — cache hits do not consume CPU time.
```
irvinebroque:
```suggestion
Workers Cache has no separate pricing. When you enable Workers Cache, all requests to your Worker are billed at the standard [Workers request rate](/workers/platform/pricing/) — the same per-request rate as any other request to your Worker — whether the response comes from cache or from your Worker. There is no charge beyond the standard request rate. **CPU time is only billed when your Worker runs** — cache hits do not consume CPU time.
```

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+  label: Overview
+---
+
+import {
+	DirectoryListing,
+	TypeScriptExample,
+	WranglerConfig,
+} from "~/components";
+
+Workers Caching lets Cloudflare return cached HTTP responses from your Worker without executing your Worker code. When an incoming request matches a cached response, Cloudflare serves the response directly from its edge cache — reducing latency and Workers CPU usage.
```
irvinebroque:
```suggestion
Workers Cache lets Cloudflare return cached HTTP responses from your Worker without executing your Worker code. When an incoming request matches a cached response, Cloudflare serves the response directly from its edge cache — reducing latency and Workers CPU usage.
```

Consistency with name of the page/section/feature

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+
+Caching works for any `fetch()` invocation of the Worker — eyeball requests (requests from browsers and API clients), requests sent through [service bindings](/workers/runtime-apis/bindings/service-bindings/), and loopback `fetch()` calls between entrypoints via [`ctx.exports`](/workers/runtime-apis/bindings/service-bindings/rpc/). You control caching with standard HTTP `Cache-Control` directives on your responses.
+
+## Your Worker's cache
+
+Workers Caching is **your Worker's cache**. It is owned by your Worker, operated by your Worker, and private to your Worker.
```
irvinebroque:
```suggestion
Workers Cache is **your Worker's cache**. It is owned by your Worker, operated by your Worker, and private to your Worker.
```

Re: "private to your Worker" when I read this I didn't immediately intuit what was meant by that

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+
+A Worker is a zoneless entity — a Worker can be bound to any number of zones, run on `workers.dev`, or be invoked entirely through service bindings without ever touching a zone. The cache follows the Worker, not a zone, so:
```
irvinebroque:
```suggestion
A Worker is a zoneless entity — a Worker can be bound to any number of [zones](/fundamentals/concepts/accounts-and-zones/#zones), run on `workers.dev`, or be invoked entirely through service bindings without ever touching a zone. The cache follows the Worker, not a zone, so:
```

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+
+- **No zone configuration for caching applies to Workers Caching.** [Cache Rules](/cache/how-to/cache-rules/), [Cache Response Rules](/cache/how-to/cache-response-rules/), Page Rules, cache level settings, the zone's default cached-file-extensions list, and every other zone-level cache control have no effect on a Worker's cache.
+- **Your Worker is in full control.** You set `Cache-Control` headers on your responses, and Cloudflare honors them per [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111). That is the entire configuration surface.
+- **The cache is shared across every way the Worker can be invoked.** A Worker bound to `api.example.com`, `api.example.net`, and invoked over a service binding serves the same cached responses to all three — the cache is keyed by the request path, entrypoint, `ctx.props`, and (by default) the Worker version, not by hostname. See [Cache keys](/workers/cache/cache-keys/).
```
irvinebroque, line 28: love this section

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+
+### The Worker is the configuration surface
```
irvinebroque, line 30:
```suggestion
### Control caching by writing code in your Worker
```

alt, can take or leave

src/content/docs/workers/cache/index.mdx
```diff
@@ -0,0 +1,432 @@
...
+cf-cache-status: HIT
+```
+
+The second request receives the cached response. The `timestamp` and `random` values in the body are identical between the two requests, even though the Worker generates fresh ones on every run — confirming that the second request did not execute your Worker.
+
+## What gets cached
+
+- HTTP invocations of the Worker's [`fetch`](/workers/runtime-apis/fetch/) handler are eligible for caching, including eyeball requests, service binding `fetch()` calls, and loopback `fetch()` calls via `ctx.exports`.
+- Only `GET` and `HEAD` requests are cached. Other methods always invoke your Worker. `GET` and `HEAD` for the same URL share a single cache entry — see [Cache keys](/workers/cache/cache-keys/#what-goes-into-the-cache-key).
+- **Only `fetch()` invocations go through the cache.** Custom RPC methods on a `WorkerEntrypoint` (for example `ctx.exports.Backend.getUser(id)`) bypass the cache entirely and always run the callee. To cache a piece of work, expose it as a `fetch` handler on its own entrypoint.
```
irvinebroque:
```suggestion
- **Only `fetch()` invocations go through the cache.** Custom [RPC methods](/workers/runtime-apis/rpc/) on a `WorkerEntrypoint` (for example `ctx.exports.Backend.getUser(id)`) bypass the cache entirely and always run the callee. To cache a piece of work, expose it as a `fetch` handler on its own entrypoint.
```

irvinebroque (approved):
going to keep making few suggested edits, can take or leave

src/content/docs/workers/cache/configuration.mdx
```diff
@@ -0,0 +1,440 @@
...
+title: Configuration
+pcx_content_type: configuration
+description: Enable and configure Workers Caching.
+sidebar:
+  order: 1
+---
+
+import { TypeScriptExample, WranglerConfig } from "~/components";
+
+Workers Caching is configured per Worker, in your Wrangler configuration file. When enabled, caching applies to every `fetch()` invocation — eyeball requests, service binding `fetch()` calls, and loopback `fetch()` calls between entrypoints via [`ctx.exports`](/workers/runtime-apis/bindings/service-bindings/rpc/) — unless you [disable it for a specific entrypoint](#per-entrypoint-caching). Custom RPC methods bypass the cache.
```
irvinebroque:
```suggestion
Workers Caching is configured per Worker, in your Wrangler configuration file. When enabled, caching applies to every `fetch()` invocation — eyeball requests, service binding `fetch()` calls, and loopback `fetch()` calls between entrypoints via [`ctx.exports`](/workers/runtime-apis/bindings/service-bindings/rpc/) — unless you [disable it for a specific entrypoint](#per-entrypoint-caching). Custom [RPC methods](/workers/runtime-apis/rpc/) bypass the cache.
```

src/content/docs/workers/cache/configuration.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+This is **your Worker's cache** — configured through your Worker's code and Wrangler file. Your Worker controls its cache entirely through:
+
+- The `cache.enabled` flag in your Wrangler configuration, which turns caching on or off. You can override it [per entrypoint](#per-entrypoint-caching) and control [cross-version behavior](#cross-version-caching).
+- The `Cache-Control` (and `cdn-cache-control`, `cloudflare-cdn-cache-control`) headers your Worker sets on its responses, per [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111).
+- The optional `Cache-Tag` response header for bulk purging, and [`ctx.cache.purge()`](/workers/cache/purge/) for programmatic invalidation.
+
+That is the entire configuration surface.
+
+Note: Wrangler must be version 4.69.0 or above. [Per-entrypoint caching](#per-entrypoint-caching) and [cross-version caching](#cross-version-caching) require Wrangler 4.107.0 or above.
```
irvinebroque: Suggest:

- use note component / syntax
- put this in the "Enable caching" section below

ala:

```
:::note
Requires Wrangler 4.107.0 or above.
:::
```

src/content/docs/workers/cache/configuration.mdx
```diff
@@ -0,0 +1,440 @@
...
+</WranglerConfig>
+
+Each entry is `{ "type": "worker", "cache": { "enabled": <boolean> } }`. A per-entrypoint `cache.enabled` overrides the top-level `cache.enabled` for that entrypoint; entrypoints you do not list inherit the top-level value. You can also enable caching for a single entrypoint without a top-level `cache` block by listing only that entrypoint.
+
+This lets you **opt specific entrypoints in and out** without changing your Worker code:
+
+- **Opt an entrypoint out** to keep it running on every request — the natural fit for a gateway or router entrypoint that authenticates, normalizes, or dispatches, and should never itself be served from cache. This is the recommended way to build the [gateway pattern](/workers/cache/examples/): disable caching on the gateway entrypoint and enable it on the inner entrypoint the gateway calls through `ctx.exports`.
+- **Opt an entrypoint in** to cache only the specific entrypoints that return reusable responses, leaving the rest of the Worker uncached.
+
+:::caution[Prefer disabling cache over returning `no-store` on a gateway]
```
irvinebroque:
```suggestion
:::caution[For lowest latency, disable cache for an entrypoint instead of returning `Cache-Control: no-store` for all responses]
```

alt, can take or leave

src/content/docs/workers/platform/pricing.mdx
```diff
@@ -41,6 +41,14 @@ WebSocket messages routed through a Worker do not count as requests.
 
 <sup>3</sup> Requests to static assets are free and unlimited.
 
+<sup>4</sup> When [Workers Caching](/workers/cache/) is enabled, requests served
+from the Worker's cache are billed at the same per-request rate as requests that
+invoke the Worker, and requests that are normally free — [static
```
irvinebroque:
```suggestion
invoke the Worker. This includes requests to [static
```

src/content/docs/workers/platform/pricing.mdx
```diff
@@ -41,6 +41,14 @@ WebSocket messages routed through a Worker do not count as requests.
...
+assets](/workers/static-assets/billing-and-limitations/) and [worker-to-worker
+invocations](/workers/platform/pricing/#service-bindings) — are also billed at
```
irvinebroque:
```suggestion
invocations](/workers/platform/pricing/#service-bindings).
```

src/content/docs/workers/platform/pricing.mdx
```diff
@@ -41,6 +41,14 @@ WebSocket messages routed through a Worker do not count as requests.
...
+the standard request rate. CPU time is only billed when the Worker runs (on a
```
irvinebroque:
```suggestion
CPU time is only billed when the Worker runs (on a
```

irvinebroque (approved):

southpolesteve (approved):

ascorbic (approved):

src/content/docs/workers/cache/purge.mdx
```diff
@@ -0,0 +1,371 @@
...
+} satisfies ExportedHandler;
+```
+
+</TypeScriptExample>
+
+On failure, each error in `errors` carries a numeric `code` and a human-readable `message` you can log or surface to the caller.
+
+## Rate limits
+
+`purge()` uses the same rate-limiting system as Cloudflare's zone purge API. For the rate limits that apply to your account's plan, refer to [Availability and limits](/cache/how-to/purge-cache/#availability-and-limits). When the purge is rate-limited, `success` is `false` and `errors` contains an entry describing the rejection.
```
irvinebroque, line 371: I was reading this and then realized wait — we make this point (good!) here in docs that this is your _Worker's_ cache

But then these purge limits are defined by your zone-level plan.

Non-blocking comment, and if we do the right things people shouldn't be purging the cache often in the first place, but if I'm a customer I'm not 100% sure what the limit is — I guess it's the limit from the zone that my Worker is running on? Or the limit from the highest zone level plan that we have?

danlapid: I'll respond in a DM

PR #30088, opened 2026-04-20
Add OCR documentation reference to DLP section

Vortexmind:
Added a reference to the OCR documentation for supported image formats.


elithrar (approved):
Approved after minor docs fix.

PR #30095, opened 2026-04-20
Adding page for using Workflows with Dynamic Workers

mia303:
### Summary

Adding page for using Workflows with Dynamic Workers & linking from Dynamic Workers docs

### Documentation checklist


src/content/docs/workflows/examples/dynamic-workers-and-workflows.mdx
```diff
@@ -0,0 +1,235 @@
...
+
+```txt
+Client
+  -> Dispatcher Worker
+  -> Dynamic Worker selected or loaded
+  -> Workflow instance created with routing metadata
+  -> Static Workflow entrypoint runs
+  -> Dispatcher reloads matching Dynamic Worker
+  -> Dynamic Worker handles run(event, step)
+  -> Workflows persists step state and output
```
LuisDuarte1: it's not very clear to me what certain steps do:
-  "Dynamic Worker selected or loaded" -> what does this mean exactly? validating the dynamic worker script?
- "Dispatcher reloads matching Dynamic Worker" -> maybe say something akin "create Dynamic Worker"

src/content/docs/workflows/examples/dynamic-workflows.mdx
LuisDuarte1, line 1: some things maybe worth doing  as suggestions:

- maybe point users to Dan's library: https://github.com/danlapid/dynamic-workflows
- combine R2 to pass "heavier" Dynamic Worker scripts/bundles and just pass the r2 object key as Instance payload


wdyt?

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+title: Dynamic Workflows
+description: Create and run Workflow logic dynamically while keeping Workflow steps durable.
+pcx_content_type: example
+sidebar:
+  order: 3
+---
+
+import { TypeScriptExample, WranglerConfig } from "~/components";
+
+Use this pattern when your workflow logic is not known at deploy time, but you still need Workflow durability. In this example, we implement a simple Dynamic Workflow which allows you to run arbitrary workflow code at runtime.
```
irvinebroque: The blog goes into examples, and helps you understand use cases for this. Valuable here as well

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+
+## Understand the model
+
+This setup has three parts:
+
+- **Loader Worker**: the front door. It receives HTTP requests, knows how to load Dynamic Worker code, and can create Workflow instances.
+- **Workflow class**: the durable orchestrator. It persists state, survives restarts, and executes steps, including `step.waitForEvent` or `step.sleep`, which may be longrunning.
```
irvinebroque:
```suggestion
- **Workflow class**: the durable orchestrator. It persists state, survives restarts, and executes steps, including `step.waitForEvent` or `step.sleep`, which may be long-running.
```

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+- **Dynamic Worker**: the user-authored code that defines what the workflow actually does. It is loaded on-demand by script ID.
```
irvinebroque: avoid use of "script"

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+
+![Dynamic Workers and Workflows architecture diagram](~/assets/images/workflows/dynamic-workflows.png)
```
irvinebroque: The diagram going in a circle confused me — it seems like what you are trying to show is that the arrow in (4) is coming from `waitForEvent` — and that then a user or client (not pictured) takes an action, which is handled by the Worker, the same dynamic worker code is invoked, and the Workflow is called

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+
+## Configure your Worker
+
+Your Worker needs a Worker Loader binding and a Workflow binding. The Worker Loader creates Dynamic Workers at runtime, and the Workflow binding points to the durable Workflow class that runs them.
```
irvinebroque:
```suggestion
The Loader Worker needs a [Worker Loader binding](/dynamic-workers/getting-started/#configure-worker-loader) and a Workflow binding. The Worker Loader creates Dynamic Workers at runtime, and the Workflow binding points to the durable Workflow class that runs them.
```

If you are going to call it "Loader Worker" above, should be consistent

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+		return new Response("Not Found", { status: 404 });
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+## Define the Workflow class
+
+You still deploy one normal Workflow class. Its job is to load the dynamic code for the current run and call it with the `WorkflowStep` object. This is what keeps the [Workflows API](/workflows/build/workers-api/), including `step.do()`, `step.sleep()`, and `step.waitForEvent()`, working as normal.
```
irvinebroque: "still" is comparing against using Workflows "normally", but we should assume the reader hasn't used Workflows before.

irvinebroque (commented):
/bonk please review these docs and suggest potential improvements

src/content/docs/workflows/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,190 @@
...
+name = "dynamic-workflow"
+binding = "WORKFLOW"
+class_name = "DynamicWorkflow"
+```
+
+</WranglerConfig>
+
+## Create your Worker entrypoint
+
+Your Worker entrypoint accepts a script, creates a Workflow instance with that script in `params`, and returns the new instance ID. The same entrypoint can also look up the status of an existing Workflow instance.
```
dinasaur404: make it clear that this is in the loader worker + explain why they need to do this and the the dynamic worker will call it

dinasaur404: should also have example of what they dynamic worker looks like

dinasaur404:
I think we should put these in the dynamic workers docs and link to it from workflows.

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+---
+
+import {
+	PackageManagers,
+	TypeScriptExample,
+	WranglerConfig,
+} from "~/components";
+
+You can run a Workflow inside a Dynamic Worker to get durable execution for code that is loaded at runtime. Each step in the Workflow survives failures, can sleep for hours or days, can wait for external events, and resumes exactly where it left off -- even if the isolate is recycled between steps.
+Because Dynamic Workers are created on-demand, you do not have to register each Workflow up front or manage them individually. Load the code when it is needed, and the Workflows engine handles persistence and retries behind the scenes. This works equally well for one-time executions as it does for long-running, multi-step processes.
```
patriciasantaana:
```suggestion
You can run a Workflow inside a Dynamic Worker to get durable execution for code that is loaded at runtime. Each step in the Workflow survives failures, can sleep for hours or days, can wait for external events, and resumes exactly where it left off — even if the isolate is recycled between steps.

Because Dynamic Workers are created on-demand, you do not have to register each Workflow up front or manage them individually. Load the code when it is needed, and the Workflows engine handles persistence and retries behind the scenes. This works equally well for one-time executions as it does for long-running, multi-step processes.
```

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+For example, you might be building:
+
+- A **SaaS platform** where each tenant defines their own automation -- onboarding sequences, approval chains, or billing retry logic -- and
+  you need each one to run durably without deploying a separate Workflow
+  per customer.
+- An **AI agent framework** where agents generate and execute multi-step plans at runtime, and each plan needs to survive restarts, sleep
+  between tool calls, and wait for human approval.
+- A **multi-tenant job system** where each customer submits their own processing logic -- data transforms, webhook chains, scheduled tasks -- and you want every step to persist progress and retry on failure without building your own orchestrator.
```
patriciasantaana:
```suggestion
- A SaaS platform where each tenant defines their own automation — onboarding sequences, approval chains, or billing retry logic — and you need each one to run durably without deploying a separate Workflow per customer.
- An AI agent framework where agents generate and execute multi-step plans at runtime, and each plan needs to survive restarts, sleep between tool calls, and wait for human approval.
- A multi-tenant job system where each customer submits their own processing logic — data transforms, webhook chains, scheduled tasks — and you want every step to persist progress and retry on failure without building your own orchestrator.
```

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+In this guide, you will use the `@cloudflare/dynamic-workflows` library to set up a Worker Loader, write a Dynamic Worker with durable steps, and trigger a Workflow instance.
+
+## Understand the model
+
+This setup has three parts:
+
+- **Worker Loader:** the main Worker you deploy. It receives requests, decides which Dynamic Worker to load, and creates Workflow instances. You write this code.
+- **Dynamic Worker:** the per-tenant code that defines what the Workflow actually does -- its steps, sleeps, and event waits. Each Dynamic Worker is loaded on-demand at runtime.
+- **DynamicWorkflow class:** a Workflow entrypoint created by the library. When the Workflows engine needs to execute a step, this class loads the correct Dynamic Worker for that instance and runs the step inside it.
```
patriciasantaana:
```suggestion
- **Worker Loader**: the main Worker you deploy. It receives requests, decides which Dynamic Worker to load, and creates Workflow instances. You write this code.
- **Dynamic Worker**: the per-tenant code that defines what the Workflow actually does — its steps, sleeps, and event waits. Each Dynamic Worker is loaded on-demand at runtime.
- **DynamicWorkflow class**: a Workflow entry point created by the library. When the Workflows engine needs to execute a step, this class loads the correct Dynamic Worker for that instance and runs the step inside it.
```

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+![Architecture](~/assets/images/dynamic-workers/dynamic-workflows.png)
+
+Here is how they work together:
+
+- The Worker Loader receives a request, loads the tenant's Dynamic Worker, and gives it a Workflow binding tagged with a tenant ID.
+- The Dynamic Worker calls env.WORKFLOWS.create() to start a new Workflow instance. The tenant ID is saved with the instance automatically.
+- The Workflows engine runs the steps defined in the Dynamic Worker -- step.do(), step.waitForEvent(), step.sleep(). Each step is durable: its result is persisted and will not re-run after it succeeds.
+- If the isolate is recycled between steps (for example, during a sleep or while waiting for an event), the engine reads the tenant ID back from the instance, reloads the same Dynamic Worker through the Worker Loader, and resumes where it left off.
```
patriciasantaana, line 45:
```suggestion
- The Worker Loader receives a request, loads the tenant's Dynamic Worker, and gives it a Workflow binding tagged with a tenant ID.
- The Dynamic Worker calls `env.WORKFLOWS.create()` to start a new Workflow instance. The tenant ID is saved with the instance automatically.
- The Workflows engine runs the steps defined in the Dynamic Worker — `step.do()`, `step.waitForEvent()`, `step.sleep()`. Each step is durable: its result is persisted and will not re-run after it succeeds.
- If the isolate is recycled between steps (for example, during a sleep or while waiting for an event), the engine reads the tenant ID back from the instance, reloads the same Dynamic Worker through the Worker Loader, and resumes where it left off.
```

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
ghost: Hola

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+The library provides two functions that handle the wiring between the Worker Loader and the Workflows engine, so you do not have to manually tag requests, parse payloads, or write your own WorkflowEntrypoint subclass.
```
patriciasantaana:
```suggestion
The library provides two functions that handle the wiring between the Worker Loader and the Workflows engine, so you do not have to manually tag requests, parse payloads, or write your own `WorkflowEntrypoint` subclass.
```

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+- `wrapWorkflowBinding`: creates a Workflow binding tagged with metadata (like { tenantId }) that you pass to a Dynamic Worker. The library attaches that metadata to every instance the Dynamic Worker creates, so the engine can trace each instance back to the right tenant.
```
patriciasantaana:
```suggestion
- `wrapWorkflowBinding`: creates a Workflow binding tagged with metadata (like `{ tenantId }`) that you pass to a Dynamic Worker. The library attaches that metadata to every instance the Dynamic Worker creates, so the engine can trace each instance back to the right tenant.
```

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+```
+
+</WranglerConfig>
+
+## Create the Worker Loader
+
+The Worker Loader is where you connect Dynamic Workers to the Workflows engine. In this file, you define:
+
+- **How to load a tenant's code**: a function that takes a tenant ID, fetches their code, and gives them a Workflow binding. The binding is created with wrapWorkflowBinding, which tags every Workflow instance with the tenant ID so the engine can route back to the right code later.
+- **How the engine resumes a Workflow**: using createDynamicWorkflowEntrypoint, you define a callback that the engine calls whenever it needs to run a step. The callback receives the tenant ID from the instance metadata and returns the tenant's Workflow class. This is what makes durable execution work across isolate restarts -- the engine knows how to reload the right code.
```
patriciasantaana:
```suggestion
- How to load a tenant's code: a function that takes a tenant ID, fetches their code, and gives them a Workflow binding. The binding is created with `wrapWorkflowBinding`, which tags every Workflow instance with the tenant ID so the engine can route back to the right code later.

- How the engine resumes a Workflow: using `createDynamicWorkflowEntrypoint`, you define a callback that the engine calls whenever it needs to run a step. The callback receives the tenant ID from the instance metadata and returns the tenant's Workflow class. This is what makes durable execution work across isolate restarts — the engine knows how to reload the right code.
```

patriciasantaana (approved):

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
```diff
@@ -0,0 +1,216 @@
...
+pcx_content_type: example
+sidebar:
+  order: 3
+---
+
+import {
+	PackageManagers,
+	TypeScriptExample,
+	WranglerConfig,
+} from "~/components";
```
patriciasantaana, line 14: Potential fix for breaking build.
```suggestion
import {
	PackageManagers,
	TypeScriptExample,
	WranglerConfig,
	tenantId
} from "~/components";
```

ghost: 👍

ghost: 2026-03-27T15:33:25.1991650Z Cleaning up orphan processes
2026-03-27T15:33:25.7949730Z ##[warning]Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-node@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

src/content/docs/dynamic-workers/examples/dynamic-workflows.mdx
ghost: 👍

jonesphillip (approved):

PR #30236, opened 2026-04-23
Expand browser close reasons with descriptions and retry guidance

kathayl:
## Summary

Expands the [Browser close reasons](https://developers.cloudflare.com/browser-run/reference/browser-close-reasons/) reference page from a bare list of reasons into a table with descriptions and actionable guidance for each close reason.

## Changes

- **Expanded close reasons table** — Each reason now includes a description explaining what happened and what the user can do about it
- **Added "Handling unexpected closures" section** — Code examples showing `try...catch` patterns and retry logic for resilient browser automation
- Reframed infrastructure-level events (session evicted) as normal managed-service behavior without exposing internal architecture details

## Why

A user asked why their session was being evicted mid-session and we had no explanation in the docs — just a row in a table that said "Browser session is evicted" with no context. This update helps users understand close reasons and write more resilient code.

### DIFF
--- a/src/content/docs/browser-run/reference/browser-close-reasons.mdx
+++ b/src/content/docs/browser-run/reference/browser-close-reasons.mdx
@@ -10,7 +10,7 @@ products:
 
 import { DashButton } from "~/components";
 
-A browser session may close for a variety of reasons, occasionally due to connection errors or errors in the headless browser instance. As a best practice, wrap `puppeteer.connect` or `puppeteer.launch` in a [`try...catch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) statement.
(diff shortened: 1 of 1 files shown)


src/content/docs/browser-run/reference/browser-close-reasons.mdx
```diff
@@ -22,10 +22,54 @@ To find the reason that a browser closed:
...
+    console.error("Browser session ended unexpectedly:", error.message);
+    // Retry or return an error response
+  } finally {
+    if (browser) {
+      try {
+        await browser.close();
+      } catch {
+        // Browser may already be closed
+      }
+    }
```
ruifigueira: I would replace this with:

```js
await browser?.close();
```

kathayl: Done, updated in 5a2c3a7. Thanks Rui!

patriciasantaana (approved):

PR #30254, opened 2026-04-23
ci: add Semgrep OSS scanning workflow

hrushikeshdeshpande:
## Summary

Adds Semgrep Community Edition (OSS) scanning to this repository as part of the App&ProdSec team's migration from Semgrep Pro to Semgrep CE.

## What it does

- Runs on every PR, on `push` to the main/master branch, and monthly on a staggered schedule.
- Uses `actions/cache@v5` so `pip install semgrep` only runs on cold cache (first run, version bump, or 7-day idle).
- Pinned to `semgrep==1.160.0` with `--config=auto` (default OSS ruleset).
- Runs on `ubuntu-slim` with `contents: read` token scope.

## For reviewers

- Findings are informational; the job does not block on findings.
- First PR after merge installs Semgrep; subsequent PRs skip that step.

See the internal App&ProdSec email for migration context, or ping us internally.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,75 +1,30 @@
+name: Semgrep OSS scan
 on:
+  pull_request: {}
+  push:
+    branches: [main, master]
   workflow_dispatch: {}
   schedule:
-    - cron: "0 4 * * *"
-
-name: Semgrep rules checking results
+    - cron: '0 0 20 * *'
+concurrency:
+  group: semgrep-${{ github.event_name }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 permissions:
   contents: read
-
 jobs:
   semgrep:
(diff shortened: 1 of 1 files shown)


.github/workflows/semgrep.yml
```diff
@@ -1,75 +1,30 @@
+name: Semgrep OSS scan
```
kodster28, line 1: Howdy @hrushikeshdeshpande, there seem to be a few issues with this PR that need to be fixed.

- Unpins our commit SHA for actions v6 -- which is a security practive -- as well as downgrading us to actions v5
- Doesn't seem to align with the strcuture of our repo (we use `production`, not main or master branches)
- Completely removes some of the previous logic built into this check.

Are those changes intentional? I'd guess not all of them. Please update and then let us know for re-review.

kodster28 (changes requested):

PR #30357, opened 2026-04-27
Stabilize catalog model encoding across re-runs

superhighfives:
Re-running \`bin/fetch-catalog-models.ts\` rewrites dozens of files with no real change — just an encoding flip between raw UTF-8 and \`\uXXXX\` escapes.

The catalog API returns non-ASCII characters inconsistently across providers. Some come through as raw UTF-8 (\`°\`, \`"\`, \`—\`), others as pre-escaped sequences (\`\u00b0\`, \`\u201c\`, \`\u2014\`). \`JSON.stringify\` preserves whatever form is in memory, so the on-disk content depends on which provider responded first, not on the actual data.

Concrete example from a current diff against \`production\`:

\`\`\`diff
- "...absolute zero (0 Kelvin or -273.15\u00b0C)..."
+ "...absolute zero (0 Kelvin or -273.15°C)..."
\`\`\`

- escape all non-ASCII chars to \`\uXXXX\` when writing
- add \`catalog:build\` script (\`pnpm catalog:build\`) for convenience

Verified 53 of 55 currently-modified catalog files round-trip through the new serializer with zero byte change vs HEAD. The 2 outliers have hand-formatted inlined \`tags\` arrays — pre-existing inconsistency unrelated to encoding.


superhighfives:
@ethulia @kodster28, this is model launch related, but should make these PR's less noisy in future. 👍

ethulia (approved):

kodster28 (approved):

PR #30441, opened 2026-04-29
feat: improve copy button UX with checkmark feedback and tippy tooltips

mvvmm:
### Summary

Standardizes tooltips across the app to just tippy.js and its default (but customized by us) theming

- Replaces the "Copied!" feedback tooltip on Expressive Code blocks with a checkmark icon swap (Phosphor Check Regular), which reverts after 2.5s
- Adds tippy.js tooltips to both the EC copy button ("Copy to clipboard") and the PackageManagers copy button, updating to "Copied!" on click
- Standardizes the copy icon to Phosphor Copy Regular across EC blocks and PackageManagers (previously EC used a stroke-style icon)
- Extracts tippy.js style overrides from an inline `<style is:global>` in `Head.astro` into `src/styles/tippy.css`
- Consolidates agent-setup components to use the shared `addTooltip()` util instead of calling tippy directly
- Migrates `AgentsToolkit` tooltips from a custom React portal implementation to tippy

### Screenshots (optional)
Screenshot 2026-04-29 at 12 20 21 PM
Screenshot 2026-04-29 at 12 20 25 PM
Screenshot 2026-04-29 at 1 16 37 PM
Screenshot 2026-04-29 at 1 16 42 PM
Screenshot 2026-04-29 at 1 16 52 PM
Screenshot 2026-04-29 at 1 16 57 PM

### DIFF
--- a/ec.config.mjs
+++ b/ec.config.mjs
@@ -33,6 +33,11 @@ export default defineEcConfig({
 		textMarkers: {
 			defaultLuminance: ["32%", "88%"],
 		},
+		frames: {
+			// Override default EC copy icon with Phosphor Copy (Regular) to match PackageManagers.
+			// Must be a CSS url() value as it is used directly as a mask-image.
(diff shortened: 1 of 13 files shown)


mvvmm:
Fixed in 45a017b — added an `initialized` guard that prevents double-init on page load, reset in `cleanup()` so subsequent Astro page navigations still re-init correctly.

ahaywood (approved):

PR #30505, opened 2026-05-01
Consume AI model catalog from middlecache

mvvmm:
Fetches AI model data from the `workers_ai_model_catalog` middlecache pipeline instead of committed JSON files.

## What changed

**Data sources**
- Replaces `catalog-models` + `workers-ai-models` content collections (committed JSON) with `ai-catalog` + `workers-ai-catalog` backed by `middlecacheLoader`
- Deletes ~154 committed model files; all model data now fetched from R2 at build time via `all-models-detail.json`
- Build fails loudly if middlecache is unavailable rather than silently dropping pages

**Model detail pages**
- Parameters section rendered at build time from `parameters.json` (pre-processed `SchemaRowData` per model, extracted from `models.tar.gz` by `bin/fetch-models.ts`)
- Raw schema files (`sync-input.json` etc.) served from R2 via worker proxy at request time — no longer static build outputs
- Dev middleware proxies `/ai/models/**` schema requests to middlecache

**Catalog index pages**
- `ModelCatalog` converted from a React island to Astro — cards are static HTML baked at build time, filter/sort/URL state managed by a vanilla JS script
- Filter dropdowns (searchable multi-select combobox, sort) remain as minimal React islands via `@base-ui/react`
- `ModelBadges` converted from React to Astro

**Infra**
- `workflow_dispatch` added to `publish-production.yml` so middlecache graduations can trigger a docs rebuild
- `bin/fetch-models.ts` pre-extracts `models.tar.gz` as a `predev`/`prebuild` hook; `--force` re-fetches and clears stale cached catalog files


ahaywood (approved):

MohamedH1998:
The cause of the build failure is the content-collection rename in `src/content.config.ts`, which both builds share:
```
catalog-models → ai-catalog
workers-ai-models → workers-ai-catalog
```
Production's model code was migrated to the new names, but Nimbus keeps its own copy of the models logic that still references the old ones, so its barrel fails to build and Rollup reports it as an unresolved import. The stale references are in `src/nimbus/util/models/model-resolver.ts`:
```
getCollection("catalog-models") / getCollection("workers-ai-models") (lines ~184–185, ~212–213)
CollectionEntry<"workers-ai-models"> / CollectionEntry<"catalog-models"> (lines ~23–24)
```
To get Nimbus green, `src/nimbus/util/models/ `needs the same migration to `ai-catalog / workers-ai-catalog` (and the middlecache-backed shape).

mvvmm:
Closing, will re-open if universal API is ever implemented

PR #30604, opened 2026-05-05
[Cache] ELI5 on cache security pages

Oxyjun:
### Summary

ELI5 pass on the 4 files in `/cache/cache-security/`. `index.mdx` is a navigation page with no prose — skipped. Three concept files were improved.

**Key changes:**

- **`avoid-web-poisoning.mdx`:** Replaced vague "client-side vulnerabilities are often exploited through HTTP headers" with a concrete explanation of how cache poisoning via reflected headers works. Fixed misleading "Never return HTTP headers to users in cached content" (headers are always returned) → "Do not include untrusted header values in your response body."
- **`cache-deception-armor.mdx`:** Removed unnecessary preamble ("Before learning about...you should first understand..."). Consolidated three paragraphs explaining the attack into two focused paragraphs. Replaced vague "flexible about what kinds of paths it can handle" with concrete "treats requests to non-existent paths as equivalent to a parent path." Added "by default" qualifier to the caching-by-extension claim. Removed first-person "we think."
- **`cors.mdx`:** Fixed inaccurate CORS description — the original said "CORS instructs the browser to determine" when CORS is actually a server-side header mechanism enforced by browsers. Fixed "block requests" → "block access to responses" (browsers block response access, not the request itself). Removed unsourced claim about Cloudflare's cache varying by Origin header.

All changes verified through adversarial fact-check review.

### Documentation checklist

### DIFF
--- a/src/content/docs/cache/cache-security/cache-deception-armor.mdx
+++ b/src/content/docs/cache/cache-security/cache-deception-armor.mdx
@@ -10,15 +10,11 @@ tags:
 
 import { DashButton } from "~/components";
 
-Before learning about Cache Deception Armor, you should first understand how Web Cache Deception attacks work.
-
 ## Web Cache Deception attacks
 
-Web Cache Deceptions attacks occur when an attacker tricks a user into opening a link in the format of `http://www.example.com/newsfeed/foo.jpg`, when `http://www.example.com/newsfeed` is the location of a dynamic script that returns different content for different users.
+A Web Cache Deception attack tricks a user into visiting a URL that looks like a static asset but actually serves dynamic, personalized content.
 
-This scenario becomes problematic when your website is configured to be flexible about what kinds of paths it can handle. To be more specific, when requests to a path that do not exist, such as `/x/y/z` are treated as equivalent to requests to a parent path that does exist `/x`.
-
-For example, an attacker could send a user a link to `http://www.example.com/newsfeed/foo.jpg` so that the user could be taken to their newsfeed. When the request passes through Cloudflare, the request would be cached because the path ends in `.jpg`. The attacker can then visit the same URL themselves, and their request will be served from Cloudflare's cache, exposing your user's sensitive content.
+This works when an origin treats requests to non-existent paths as equivalent to a parent path. For example, if your origin serves the same dynamic response for both `/newsfeed` and `/newsfeed/foo.jpg`, an attacker could send a user a link to `http://www.example.com/newsfeed/foo.jpg`. Because the path ends in `.jpg`, Cloudflare caches the response by default. The attacker then visits the same URL and receives the cached copy of the user's personalized content.
 
 ## Cache Deception Armor protects against attacks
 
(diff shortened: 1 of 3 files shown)


src/content/docs/cache/cache-security/cache-deception-armor.mdx
```diff
@@ -10,15 +10,11 @@ tags:
...
+A Web Cache Deception attack tricks a user into visiting a URL that looks like a static asset but actually serves dynamic, personalized content.
 
-For example, an attacker could send a user a link to `http://www.example.com/newsfeed/foo.jpg` so that the user could be taken to their newsfeed. When the request passes through Cloudflare, the request would be cached because the path ends in `.jpg`. The attacker can then visit the same URL themselves, and their request will be served from Cloudflare's cache, exposing your user's sensitive content.
+This works when an origin treats requests to non-existent paths as equivalent to a parent path. For example, if your origin serves the same dynamic response for both `/newsfeed` and `/newsfeed/foo.jpg`, an attacker could send a user a link to `http://www.example.com/newsfeed/foo.jpg`. Because the path ends in `.jpg`, Cloudflare caches the response by default. The attacker then visits the same URL and receives the cached copy of the user's personalized content.
```
Oxyjun: Need to check if the rewording is better than the previous one. I think the best might be a mix of both.

src/content/docs/cache/cache-security/cors.mdx
```diff
@@ -8,14 +8,14 @@ tags:
...
 
 Cloudflare supports CORS by:
 
 * Identifying cached assets based on the `Host` Header, `Origin` Header, URL path, and query. This allows different resources to use the same `Host` header but different `Origin` headers.
 * Passing `Access-Control-Allow-Origin` headers from the origin server to the browser.
 
-The `Access-Control-Allow-Origin` header allows servers to specify rules for sharing their resources with external domains. When a server receives a request to access a resource, it responds with a value for the `Access-Control-Allow-Origin` header. `Access-Control-Allow-Origin` headers are often applied to [cacheable content](/cache/concepts/default-cache-behavior/). A web server may respond with different `Access-Control` headers depending on the `Origin` header sent in the request.
+The `Access-Control-Allow-Origin` header lets a server specify which external origins are allowed to access its resources. A server may respond with different `Access-Control-Allow-Origin` values depending on the `Origin` header in the request. These headers are often present on [cacheable content](/cache/concepts/default-cache-behavior/).
```
Oxyjun: The phrase "rules for sharing" has been dropped. The new wording frames it as binary (allowed or disallowed). Needs a nuance check.

pedrosousa: I believe we still kept the correct meaning, even though we're not mentioning "sharing".

src/content/docs/cache/cache-security/cache-deception-armor.mdx
pedrosousa: Yes, this part seems relevant for setting the context:
> `http://www.example.com/newsfeed` is the location of a dynamic script that returns different content for different users.

pedrosousa (approved):

PR #30699, opened 2026-05-08
[Agents] Update agent-class.mdx

luishcardoso:
Hello team, 

### Summary

Can we link here the blog post for further info on Workers runtime?

https://blog.cloudflare.com/workerd-open-source-workers-runtime/

Thanks.

### DIFF
--- a/src/content/docs/agents/concepts/agent-class.mdx
+++ b/src/content/docs/agents/concepts/agent-class.mdx
@@ -28,7 +28,7 @@ Let's briefly consider which primitives are exposed by Durable Objects so we und
 constructor(ctx: DurableObjectState, env: Env) {}
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/docs/agents/concepts/agent-class.mdx
```diff
@@ -28,7 +28,7 @@ Let's briefly consider which primitives are exposed by Durable Objects so we und
 constructor(ctx: DurableObjectState, env: Env) {}
 ```
 
-The Workers runtime always calls the constructor to handle things internally. This means two things:
+The [Workers runtime](https://blog.cloudflare.com/workerd-open-source-workers-runtime/) always calls the constructor to handle things internally. This means two things:
```
kristianfreeman, line 31: maybe makes more sense to link here? https://developers.cloudflare.com/workers/reference/how-workers-works/

PR #30799, opened 2026-05-13
[ACM] Document multi-level subdomain support and limits

ngayerie:
## Summary

Documents the limits and capabilities for multi-level subdomains in Advanced Certificate Manager.

## Problem

Customers ask about subdomain depth limits when evaluating ACM for deep multi-level subdomains like `1.2.3.4.5.example.com`. The current documentation mentions "Cover more than one level of subdomain" but doesn't explain the actual constraints.

## Solution

Added a new "Multi-level subdomain support" section to the ACM page explaining:

- **No arbitrary depth limit** - subdomain depth is constrained by character limits, not level count
- **Domain name length limits** - 253 chars total, 63 chars per label (RFC 1035/5280)
- **Common Name workaround** - Cloudflare uses `sni.cloudflaressl.com` as CN when hostname exceeds 64 chars
- **Wildcard coverage** - wildcards only cover one level, with examples
- **50 hostnames per certificate** - practical limit for SANs
- **CA consistency** - same limits across Google Trust Services, Let's Encrypt, and SSL.com

## Ticket

- SPM-3540

### DIFF
--- a/src/content/docs/ssl/edge-certificates/advanced-certificate-manager/index.mdx
+++ b/src/content/docs/ssl/edge-certificates/advanced-certificate-manager/index.mdx
@@ -78,7 +78,7 @@ These limits are defined by internet standards ([RFC 1035](https://www.rfc-edito
 
 - **Total domain length**: The entire domain name cannot exceed 253 characters.
 - **Label length**: Each individual level (the text between dots) cannot exceed 63 characters.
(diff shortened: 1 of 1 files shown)


baubuchon-cf (approved):
Thanks for the fast turn on this, @ngayerie — really appreciate the same-day pickup.

I've reviewed the changes and all five items are resolved correctly: 

- SAN limit phrasing is now unambiguous (50 total, apex as one of those 50) — and you also caught the same issue in the existing "What the add-on includes" bullet, which is exactly right
 - cloudflare_branding opt-in via API is accurately described, with the API link
- "Consistency across certificate authorities" section now correctly distinguishes IETF-defined limits from Cloudflare-defined ones
- Total TLS cross-link and the standard www/api/staging example pattern are both in

LGTM from SSL/TLS.

ngayerie:
Hi @baubuchon-cf @elithrar 
Thanks!

PR #30806, opened 2026-05-13
[SaaS] Update reference to Static IP prefixes in Apex Proxying documentation

ngayerie:
From chat thread: let's remove the ambiguity, Apex Proxying requires Static IPs (and not just Dedicated IPs)


baubuchon-cf (approved):
LGTM — clarifying "IP prefixes" to "Static IP prefixes" with a link to /byoip/concepts/static-ips/ removes the ambiguity nicely. Static IPs is the correct product for Apex Proxying since it requires stable IPs for the customer A records.

Approving from the SSL/TLS product side.

ngayerie:
Hi @baubuchon-cf @elithrar 
Thanks!

PR #30841, opened 2026-05-15
[SSL] Update subdomain coverage details in documentation

ngayerie:
Clarify that advanced certificates can cover subdomains in full setup zones, as Universal SSL is enough on Partial zone. Raised in the SSL chat room

### DIFF
--- a/src/content/docs/ssl/edge-certificates/advanced-certificate-manager/index.mdx
+++ b/src/content/docs/ssl/edge-certificates/advanced-certificate-manager/index.mdx
@@ -26,7 +26,7 @@ Advanced Certificate Manager allows you to:
 
 - Order advanced certificates that can:
   - Include the zone apex and up to 50 hosts as covered hostnames.
-  - Cover more than one level of subdomain on [Full setup zones](/dns/zone-setups/full-setup/).
+  - Cover more than one level of subdomain.
   - Be issued by the certificate authority (CA) you choose.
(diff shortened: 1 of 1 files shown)


src/content/docs/ssl/edge-certificates/advanced-certificate-manager/index.mdx
```diff
@@ -26,7 +26,7 @@ Advanced Certificate Manager allows you to:
 
 - Order advanced certificates that can:
   - Include the zone apex and up to 50 hosts as covered hostnames.
-  - Cover more than one level of subdomain.
+  - Cover more than one level of subdomain on [Full setup zones](/dns/zone-setups/full-setup/).
```
lgarofalo: ACM also covers more than one level of subdomain in partial zones, maybe this isn't the right spot to make this comment?

ngayerie: How should we phrase it then?
The message we want to convey is that one doesn't need ACM on partial zone as USSL are provisioned for each proxied hostname, whatever the level.

baubuchon-cf (approved):
Thanks for the quick iteration, Nicolas — this reads well. 

Approving from the SSL/TLS product side.

ngayerie:
Hi @baubuchon-cf @elithrar 
Thanks!

PR #30847, opened 2026-05-15
[Access] Document granular target permissions for Infrastructure Access

kennyj42:
## Summary

- Adds new **Granular target permissions** section to the Infrastructure Access docs
- Documents that scoped roles can filter the targets list API to only show authorized targets

## Source MR

- hephaestus !340 (ZTIA-1043) — AAuthz support for GET /targets list endpoint


asamborski (approved):
TY for updating this one! Appreciate it!

PR #30889, opened 2026-05-18
[Chore] Remove standalone search page

kodster28:
### Summary

We have a standalone page at `/search/` that's a lot of tech debt for something this low traffic.


src/components/overrides/TableOfContents.astro
ahaywood, line 1: I think you could remove this altogether.

ahaywood (approved):

PR #30999, opened 2026-05-21
[Tunnel] Document IPv6 service address format

nikitacano:
Adds an **IPv6 service addresses** section to the shared `protocols-table.mdx` partial used by both `/tunnel/routing/` and `/cloudflare-one/networks/connectors/cloudflare-tunnel/routing-to-tunnel/protocols/`.

## What it documents

When the service value for a published application is an IPv6 literal, the address must be wrapped in square brackets (`[ ]`) per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2). The brackets are required so the `:` characters in the address are not confused with the port separator.

Example: `https://[2001:db8::1]:443`

## Why

The dashboard requires bracketed IPv6 service URLs but the docs did not mention IPv6 anywhere in the tunnel routing / protocols pages, leading users to think it might be a UI quirk. It is not — `cloudflared` parses the service value through Go's standard `net/url.Parse`, which follows RFC 3986 and requires the bracket form (verified in [`ingress/ingress.go`](https://github.com/cloudflare/cloudflared/blob/master/ingress/ingress.go) `validateIngress`).

## Scope

Single partial edit, +14 lines, no other files touched. The partial is rendered in both products so the docs change ships in `/tunnel/` and the CF1 Tunnel docs in one commit.


nikitacano:
/bonk search docs for other places we need to make this clear and update this PR

kennyj42 (approved):

PR #31001, opened 2026-05-21
Fix incorrect header names in mTLS Managed Transform documentation

trinsic-josh:
### Summary

The documentation for the "Add TLS client auth headers" Managed Transform Rule is incorrect. It is not possible to build a functional implementation against this documentation, because the header names are wrong.

Cloudflare does not send any header named `cf-cert-fingerprint-sha256` or `cf-cert-fingerprint-sha1`. Instead, the headers are named `cf-cert-sha256` and `cf-cert-sha1`.


elithrar (approved):

ngayerie:
Hi @mbullock1986 
Would it be possible for you to review/approve this one?

mbullock1986:
approved

mvvmm (approved):

mvvmm:
approving for @mbullock1986 so we can get this one in.

PR #31005, opened 2026-05-22
[Access] Expand DLP and Gateway routing docs for MCP server portals

kennyj42:
## What this PR does

Rewrites and significantly expands the Gateway routing section for MCP portals:

- **How Gateway routing works**: Explains the traffic flow (portal -> Gateway -> upstream server)
- **Background sync exclusion**: Documents that admin credential sync does NOT route through Gateway
- **Portal-level vs per-server scope**: Documents both the portal-level and per-server Gateway routing flags (the per-server flag was completely undocumented)
- **SSE transport limitation**: Documents that SSE is not supported through Gateway and the portal auto-falls back to Streamable HTTP
- **AI Prompt Protection clarification**: Expands the note explaining that AI prompt profiles do not work with portal traffic, and why
- **Limitations section**: Consolidates all known limitations in one place

## Why

The first live DLP customer (Aledade, ESCALATION-2034) hit multiple issues because these limitations were not documented. The team spent 60+ messages debugging in chat. Key findings:

1. AI Prompt Protection does not work with portals (only specific web client API paths)
2. The per-server SWG flag exists but was not documented
3. SSE transport silently fails through Gateway
4. Background sync bypasses Gateway (relevant for SHELP-834 egress IP use case)

### DIFF
--- a/src/content/docs/cloudflare-one/access-controls/ai-controls/mcp-portals.mdx
+++ b/src/content/docs/cloudflare-one/access-controls/ai-controls/mcp-portals.mdx
@@ -200,13 +200,13 @@ To turn off code mode for a portal:
 
 ## Route portal traffic through Gateway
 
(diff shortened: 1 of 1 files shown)


elithrar (approved):
Approved after minor docs fix.

PR #31012, opened 2026-05-22
[Cache] Clarify Always Online trigger condition — only activates on origin unreachability

ack-cf:
## Summary

Clarifies that Always Online only activates when Cloudflare cannot reach the origin server (resulting in Cloudflare-generated 520–527 status codes), not when the origin is reachable but returning 5xx errors.

### `src/content/docs/cache/how-to/always-online.mdx`
Makes explicit in the Visitor Experience section that:
- The 520–527 status codes are generated by Cloudflare, not the origin
- Always Online only activates on origin unreachability, not origin errors
- If the origin is reachable and returns 500/503 etc., Always Online will not trigger

### `src/content/docs/cache/troubleshooting/always-online.mdx`
Adds context to the existing limitation explaining why 404/500/503 errors don't trigger Always Online — because they indicate the origin is reachable and responding.


src/content/docs/cache/how-to/always-online.mdx
```diff
@@ -25,7 +25,9 @@ When the Internet Archive integration is enabled, Cloudflare tells the Internet
...
 
-When a visitor requests content for an offline website, Cloudflare returns an HTTP response status code in the range [520–527](/support/troubleshooting/http-status-codes/cloudflare-5xx-errors/error-520/), depending on the issue. These status codes indicate that the origin is unreachable.
+When a visitor requests content and Cloudflare is unable to reach your origin server, Cloudflare returns an HTTP response status code in the range [520–527](/support/troubleshooting/http-status-codes/cloudflare-5xx-errors/error-520/), depending on the issue. These status codes are generated by Cloudflare and indicate that the origin is unreachable. Always Online only activates when Cloudflare cannot connect to your origin — it does not activate when the origin is reachable but returning error responses.
+
+If your origin is reachable and returns a 5xx status code (such as 500 or 503), Always Online will not trigger because the origin is online. Always Online is designed to handle origin unreachability, not origin errors.
```
zaidoon1: let's say "such as 520" . 500/503 is not related to AO so it's not as strong as an example. Rest looks good to you.

ack-cf: added

MohamedH1998 (approved):

PR #31047, opened 2026-05-25
D1: add the column-limit and bound-parameter-limit errors to the D1_ERROR catalog

ajwcontreras:
### Summary

The "List of D1_ERRORs" table on the Debug D1 page does not include the two errors a developer hits when exceeding D1's documented per-table column limit and per-query bound-parameter limit (both 100, on the [Limits](/d1/platform/limits/) page). This adds those two rows so the errors are findable from the error catalog. Recommended actions link to the Limits page rather than inlining the numeric cap.

Verified against D1: a `CREATE TABLE` with 101 columns returns `too many columns on <table>: SQLITE_ERROR`; a prepared statement with 101 bound parameters returns `too many SQL variables at offset N: SQLITE_ERROR`; 100 columns succeeds.

### Documentation checklist


src/content/docs/d1/observability/debug-d1.mdx
```diff
@@ -80,6 +80,8 @@ Before retrying any failed operation:
...
+| `too many columns on <table>: SQLITE_ERROR`                                 | A `CREATE TABLE` statement specified more columns than D1's [per-table column limit](/d1/platform/limits/).                                                               | App action: Reduce the number of columns to within the [per-table column limit](/d1/platform/limits/), or split the schema across multiple related tables. |
+| `too many SQL variables at offset N: SQLITE_ERROR`                          | A prepared statement bound more parameters than D1's [per-query bound-parameter limit](/d1/platform/limits/).                                                             | App action: Reduce the bound parameters to within the [per-query bound-parameter limit](/d1/platform/limits/); for bulk inserts, split into multiple `db.batch()` statements. |
```
lambrospetrou, line 84: I am not sure we want to list all the possible SQLITE_ errors someone can get since those are many, and depend on the queries themselves. The limits page lists those constraints already, as you say in the description.

ajwcontreras: fair. thing is i was building some finance document ETL pipelines, way more columns than 100 for those doc types, agents kept complaining.

elithrar (approved):
Approved as a low-risk product-owner docs change.

PR #31048, opened 2026-05-25
Durable Objects: document the RangeError when a KV storage value exceeds 128 KiB

ajwcontreras:
### Summary

The KV-backed Durable Object `put()` documentation and the Limits page state the 128 KiB value-size cap but do not name the error thrown when it is exceeded. By contrast, the SQLite full-storage case is already documented with the `SQLITE_FULL` error and an example, so this adds the parallel for KV-backed storage.

Verified: a value whose serialized size exceeds 131072 bytes makes `put()` throw `RangeError: Values cannot be larger than 131072 bytes.` before the write is applied (values at and just under the cap succeed).

### Documentation checklist


iglesiasbrandon (approved):
Thanks for the PR! I verified the behavior in production.

PR #31066, opened 2026-05-26
Warn about using DO SQL cursor after yielding event loop

a-robinson:
### Summary

We previously have not warned about this, but it is possible to get inconsistent results if you yield the event loop in the middle of reading results back from a sql cursor. It's best to make this clear until we can better prevent users from doing it.

Discussed internally with @lambrospetrou and @justin-mp.

### Documentation checklist

### DIFF
--- a/src/content/docs/durable-objects/api/sqlite-storage-api.mdx
+++ b/src/content/docs/durable-objects/api/sqlite-storage-api.mdx
@@ -153,6 +153,29 @@ class MyDurableObject(DurableObject):
 
 A cursor (`SqlStorageCursor`) to iterate over query row results as objects. `SqlStorageCursor` is a JavaScript [Iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol), which supports iteration using `for (let row of cursor)`. `SqlStorageCursor` is also a JavaScript [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol), which supports iteration using `cursor.next()`.
 
+:::caution[Consume cursors synchronously]
+It is not safe to hold a reference to a `SqlStorageCursor` and continue reading from it after an `await`. When you `await` an asynchronous operation, other requests can interleave and modify the underlying database, which can silently invalidate the cursor. This does not produce an error — the cursor may appear to work but return unpredictable results.
+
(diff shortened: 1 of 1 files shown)


lambrospetrou (approved):

justin-mp (approved):

src/content/docs/durable-objects/api/sqlite-storage-api.mdx
```diff
@@ -153,6 +153,29 @@ class MyDurableObject(DurableObject):
...
+const rows = this.ctx.storage.sql.exec("SELECT * FROM users").toArray();
+const result = await fetch("https://example.com", { method: "POST", body: JSON.stringify(rows) });
+```
+
+```ts
+// 🔴 Unsafe: cursor is read after an await
+const cursor = this.ctx.storage.sql.exec("SELECT * FROM users");
+await fetch("https://example.com/notify");
+// Another request may have modified the database during the await above.
+// The cursor is now potentially invalid — it may appear to work but return wrong data.
```
irvinebroque, line 172: is it _invalid_ / _unsafe_ or is it that it is just stale data?

you say above "unpredictable" results — is it that they are unpredictable and inherently problematic, or just that they are potentially stale?

a-robinson: This was pretty vague, sorry. Before making this change I had an LLM try reproducing issues, and while it may not have exhaustively found all issues, it did confirm that the cursor could:

1. read data that was written after the sql statement was initially run
2. fail to read data that was deleted after the sql statement was initially run
3. observe changes from later statements prior to them committing, even if they get rolled back before committing (causing reads to reflect a state of the DB that was never committed)

Would you rather this warning be more specific?

irvinebroque: I was just curious since this seems like expected and possible to reason about behavior that I’m kind of used to from other contexts, versus language of it being “unsafe” piqued my attention about it maybe having more dramatic consequences

a-robinson: Yeah, if that's really all that's possible then it'll be good enough for plenty of users if they know what they're getting into, so we could document that as the extent of the impact.

There is still a level of risk in that we haven't specifically tested or tried to understand all the things that could go wrong. But based on very basic testing, sqlite does appear to have put effort into ensuring things don't totally blow up when you do this, so it's probably fine to soften the language here.

a-robinson: #31087

PR #31069, opened 2026-05-26
[Sandbox] Clarify user workspace guidance

whoiskatrin:
### Summary

Updates Sandbox SDK documentation to clarify how sandbox IDs, sessions, terminals, and mounted storage should be scoped in user-facing applications. The guidance now consistently presents sandboxes as user or workspace boundaries, while sessions remain workflow-level shell contexts inside a sandbox.

### Documentation checklist


ghostwriternr (approved):
this is super helpful, i've noticed coding agents very often don't get how and when to use unique identifiers as well as configuring the `containers` wrangler json object properly for the use-case. we should perhaps open a separate docs update later to emphasise on the latter.

PR #31116, opened 2026-05-28
[Magic Transit] Clarify that Cloudflare-originated traffic counts toward bandwidth

alpdot:
## Summary
Customers have reported unexpected bandwidth charges because it was not documented that traffic originating from Cloudflare services (for example, CDN) is routed through Magic Transit tunnels when the destination prefix is onboarded and advertised, even if a more specific route exists via their ISP.
## Changes
- **Bandwidth measurement** (`bandwidth-measurement.mdx`): Added section explaining that Cloudflare-originated traffic counts toward bandwidth, with a `/20` and `/24` example. Also added guidance on how to avoid this traffic pattern.
## Related
This addresses customer confusion where more specific ISP routes do not affect how Cloudflare delivers traffic that originates from our own network.


jeffh-cloudflare (approved):
Reviewed

PR #31125, opened 2026-05-29
[Cloudflare One] Document TLS inspection requirement for authorization proxy endpoints

grstnhbr:
## Summary

- Adds a new "TLS inspection required" limitation under authorization endpoint limitations, documenting that TLS decryption is always performed to read the auth cookie - even if TLS decryption is turned off at the account level.
- Clarifies the certificate installation step (3a) to explain that the Cloudflare certificate is mandatory for authorization endpoints to function, not just for HTTP policy enforcement.

Context: a customer discovered during testing that TLS decryption happens by default on authorization proxy endpoints despite being disabled at the account level. This is expected behavior - the auth cookie can only be read from decrypted requests - but it was not documented.


elithrar (approved):
Approved for docs auto-merge.

elithrar (approved):

PR #31202, opened 2026-06-03
[Sandbox] Add docs for Sandbox S3 credential proxy

scuffi:
### Summary

This PR adds documentation for the S3 mount credential proxy implemented in https://github.com/cloudflare/sandbox-sdk/pull/727

### Documentation checklist

### DIFF
--- a/src/content/docs/sandbox/api/storage.mdx
+++ b/src/content/docs/sandbox/api/storage.mdx
@@ -116,6 +116,7 @@ interface RemoteMountBucketOptions {
   endpoint: string;
   provider?: BucketProvider;
   credentials?: BucketCredentials;
+  credentialProxy?: boolean;
   readOnly?: boolean;
   s3fsOptions?: string[];
   prefix?: string;
@@ -152,6 +153,7 @@ type MountBucketOptions =
 
 - **Remote endpoint mount** - Set `endpoint` to mount any S3-compatible provider
   - Supports explicit `credentials` or environment variable auto-detection
+  - Set `credentialProxy: true` to keep credentials out of the container (egress interception)
   - Supports `provider`, `prefix`, `readOnly`, and `s3fsOptions`
 
 **Field details**:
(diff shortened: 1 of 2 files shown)


src/content/docs/sandbox/guides/mount-buckets.mdx
```diff
@@ -123,6 +123,38 @@ await sandbox.mountBucket('my-r2-bucket', '/data', {
 ```
 </TypeScriptExample>
 
+### Credential proxy
+
+When you mount with explicit credentials, s3fs writes those credentials to a password file on the container's disk. A compromised container process can read and exfiltrate the credentials, or use them to access storage outside the intended bucket scope.
```
aron-cf, line 128: I assume we default to `false` for backwards compatibility. I feel like this should be prefixed with a note saying this will become the default in the future. From what I understand from your write up there isn't a large enough performance penalty that would justify having credentials on disk right?

src/content/docs/sandbox/guides/mount-buckets.mdx
```diff
@@ -123,6 +123,38 @@ await sandbox.mountBucket('my-r2-bucket', '/data', {
...
+
+Set `credentialProxy: true` to keep credentials out of the container entirely. Instead of passing real credentials into the container, the Durable Object intercepts all outbound S3 requests at the network layer, re-signs them with the real credentials, and forwards them upstream. The container only ever holds dummy credentials that are useless outside the proxy.
+
+This works with AWS SigV4 signing for S3-compatible endpoints (including R2) and HMAC signing for Google Cloud Storage.
```
aron-cf: I would link to docs for AWS SigV4 here. I don't know what this is. Save me a google.

src/content/docs/sandbox/api/storage.mdx
```diff
@@ -172,6 +174,12 @@ type MountBucketOptions =
   - Contains `accessKeyId` and `secretAccessKey`
   - If not provided, uses environment variables
 
+- `credentialProxy` (remote endpoint mode only) - Route S3 requests through the Durable Object for signing
+  - When `true`, credentials are never written to the container's disk. The Durable Object intercepts and re-signs all outbound S3 requests at the network layer before forwarding them upstream.
+  - Supports AWS SigV4 signing for S3-compatible endpoints (including R2) and HMAC signing for Google Cloud Storage
+  - Requires `ContainerProxy` to be exported from your Worker entrypoint
+  - Default: `false`
```
aron-cf: Again, let's explain that this _should_ be set to true always. But current default is for backwards compat.

aron-cf (approved):
Looks great 😍 some minor comments inline around how we frame the default value.

src/content/docs/sandbox/guides/mount-buckets.mdx
scuffi: There may be a performance penalty for reads right now. I'll still update the docs that we're going to default this to `true` in future, assuming I can speed up the read overhead

PR #31221, opened 2026-06-03
[Chore] Update code fence values

kodster28:
add a few code fence values where none + replace unsupported values


mvvmm (approved):

mvvmm:
proto is supported. Need to update it in the style guide reference docs. ignore for now.

PR #31237, opened 2026-06-04
[DNS] Add FAQ entry about TXT record splitting at 255 characters

ngayerie:
### Summary

Adds a new FAQ entry to the DNS documentation explaining why TXT records appear split with a double quote in the middle of the value. This is standard DNS behavior per RFC 4408 and RFC 1035 — TXT record strings are limited to 255 characters, so longer values are automatically split into multiple quoted strings.

Addresses DEE-3574.

### Documentation checklist

### DIFF
--- a/src/content/docs/dns/faq.mdx
+++ b/src/content/docs/dns/faq.mdx
@@ -132,9 +132,9 @@ Refer to [Deprecating the DNS ANY meta-query type](https://blog.cloudflare.com/d
 
 ### Why does my TXT record show a double quote in the middle of the value?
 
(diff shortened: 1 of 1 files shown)


ngayerie:
Hi @hannes-cf , would it be possible for you to review/merge?

hannes-cf (approved):
Would love @svenr-cf 's eyes on this as well.

svenr-cf:
Looks pretty good. Just a few small points:

- `"first part""second part"` should be `"first part" "second part"` (space in the middle)
- Maybe: note that all DNS providers (necessarily) do this, even if they don't show it in UI/API
- Technically the statement about receiving applications concatenating together isn't guaranteed to be correct, though in practice that's what happens for all significant types of TXT records (DKIM, SPF, DMARC, etc.) nowadays. Maybe it's fine to leave it as-is. (To clarify: this is the defined behavior for DKIM, SPF, DMARC, etc., but any other application can use TXT records in its own way and then this would not necessarily be guaranteed/expected.)

ngayerie:
hey @hannes-cf @svenr-cf , is it better? can we merge?

svenr-cf:
I still only see yesterday's commit; maybe GitHub is having a moment...

ngayerie:
Sorry.. I hadn't noticed.. Updated now!

svenr-cf:
Beautiful, thanks! LGTM

svenr-cf (approved):

PR #31262, opened 2026-06-05
[Stream] Google Drive share links are error-prone

tsmith512:
### Summary

Agents are recommending Google Drive share links for Stream ingress, which frequently fail upstream because of, essentially, automated traffic prevention and rate limiting on Google's end. Added warning about it.

Will follow up with alternatives, but trying to steer solutions away from this path.


tsmith512:
/bonk the PR description is filled out. I explained exactly what I am changing, why, and next steps. I should be able to merge this.

tsmith512:
/bonk no it isn't, this PR is still blocked on your change request

kodster28 (approved):

PR #31292, opened 2026-06-08
[Python] Update Python workers intro docs to be up-to-date

ryanking13:
### Summary

Cleaning up Python workers documentation a bit before to prepare GA. I'll update the bindings doc after we land bindings updates to our sdk.

### Documentation checklist


src/content/docs/workers/languages/python/packages/index.mdx
```diff
@@ -22,7 +22,7 @@ To get started, create a pyproject.toml file with the following contents:
 name = "YourProjectName"
 version = "0.1.0"
 description = "Add your description here"
-requires-python = ">=3.12"
+requires-python = ">=3.13"
```
ryanking13, line 25: The version here doesn't actually matter, but I think it would be good to have a newer version.

src/content/docs/workers/languages/python/ffi.mdx
```diff
@@ -56,8 +56,6 @@ class Default(WorkerEntrypoint):
         return Response(bar) # returns "baz"
 ```
 
-Under the hood, `env` is actually a JavaScript object. When you call `.FOO`, you are accessing this property via a [`JsProxy`](https://pyodide.org/en/stable/usage/api/python-api/ffi.html#pyodide.ffi.JsProxy) — special proxy object that makes a JavaScript object behave like a Python object.
```
ryanking13, line 59: Removed mentioning JsProxy. Technically our SDK wraps JsProxy and I don't think users need to know about it.

src/content/docs/workers/languages/python/how-python-workers-work.mdx
```diff
@@ -75,6 +75,13 @@ Refer to the [blog post introducing Python Workers](https://blog.cloudflare.com/
...
+A new version of Python is released every year in August, and a new version of Pyodide is released six (6) months later.
+When this new version of Pyodide is published, we will add it to Workers by gating it behind a Compatibility Flag, which is only enabled after a specified Compatibility Date.
+This lets us continually provide updates, without risk of breaking changes, extending the commitment we’ve made for JavaScript to Python.
+
+Each Python release has a [five (5) year support window](https://devguide.python.org/versions/).
+Once this support window has passed for a given version of Python, security patches are no longer applied, making this version unsafe to rely on.
+Following the "A change to the Workers Runtime must never break an application that is live in production." policy,
```
dom96: Hmm, this long quote to describe the policy reads a bit weird. Could we shorten it? Maybe something like

> Following the Workers Runtime policy to never break an application that is live in production,

src/content/docs/workers/languages/python/how-python-workers-work.mdx
```diff
@@ -75,6 +75,13 @@ Refer to the [blog post introducing Python Workers](https://blog.cloudflare.com/
...
+existing Python Workers on versions outside the support window will continue to work.
+However, we do not recommend using Python versions outside the support window for new projects,
+and we will not provide patches for issues arising from using these versions.
```
dom96, line 87: Also add something like this here:

> We also cannot guarantee that these older Python versions won't suffer from degraded performance, including higher latency or CPU time usage.

dom96 (approved):
Thanks!

irvinebroque (approved):

Issue #31312, opened 2026-06-08
Availability table is not displayed correctly in documentation section for WAF


vozh-gp:
### Example URL(s)

https://developers.cloudflare.com/waf/#availability

### Actual Behavior

Table is displayed as plain text

### Expected Behavior

Table is displayed correctly

AntoineGagnon:
This is happening in multiple sections of the docs:

https://developers.cloudflare.com/workers/ci-cd/builds/build-image/

kodster28:
fixed by #31314

Issue #31310, opened 2026-06-08
Fix broken table in R2 docs


elithrar:
### Example URL(s)

https://developers.cloudflare.com/r2/platform/metrics-analytics/#storage

### Actual Behavior

The table does not render correctly. 

### Expected Behavior

I expect a markdown table to render correctly

elithrar:
/bonk please fix the broken markdown table in this docs page

rdimaio:
FYI this seems to be a docs-wide issue and not just with R2, e.g.

- https://developers.cloudflare.com/agents/harnesses/think/configuration/#session-integration
- https://developers.cloudflare.com/waf/rate-limiting-rules/parameters/

kodster28:
closed by #31314

PR #31314, opened 2026-06-08
fix: restore GFM table rendering broken by astro 6.4.0

mvvmm:
### Summary

Fixes broken table rendering introduced by the `astro@6.3.8` → `6.4.4` bump in #31244.

Astro 6.4.0 changed the `markdown.gfm` config schema from `.default(true)` to `.optional()` as part of the new `markdown.processor` API. When `gfm` is not explicitly set, it is now `undefined` instead of `true`. The `@astrojs/mdx` integration resolves `gfm` via a `??` chain that bottoms out at `undefined`, so `remark-gfm` is never added and GFM table syntax goes unprocessed — tables render as raw pipe-delimited text.

Fix is to explicitly set `gfm: true` in the `markdown` block of `astro.config.ts`, restoring the previously-implicit behavior.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -102,6 +102,7 @@ export default defineConfig({
 	site: "https://developers.cloudflare.com",
 	cacheDir: ".astro-cache",
 	markdown: {
+		gfm: true,
 		smartypants: false,
 		remarkPlugins: [remarkValidateImages],
 		rehypePlugins: [


colbywhite (approved):

MohamedH1998 (approved):

colbywhite:
Closes #31313.
Closes #31310.

colbywhite:
Closes #31311.

AntoineGagnon:
Closes #31312

PR #31346, opened 2026-06-09
Adds changelog to highlight the removal of Python built-in packages.

dom96:
### Summary

These have been removed and are no longer supported. They were only ever supported as part of a private beta.

### DIFF
--- a/src/content/changelog/workers/2026-06-10-python-requirements-txt-removal.mdx
+++ b/src/content/changelog/workers/2026-06-10-python-requirements-txt-removal.mdx
@@ -1,0 +1,14 @@
+---
+title: Python packages included using requirements.txt are no longer supported
+description: Support for `requirements.txt` has been removed and any running Workers using it will need to be updated to use the new package system.
+products:
+  - workers
+date: 2026-06-10
+---
+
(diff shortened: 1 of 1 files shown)


src/content/changelog/workers/2026-06-10-python-requirements-txt-removal.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Python packages included using requirements.txt are no longer supported
+description: Support for `requirements.txt` has been removed and any running Workers using it will need to be updated to use the new package system.
+products:
+  - workers
+date: 2026-06-10
+---
+
+As part of a closed private beta, Python Workers supported a limited set of packages as built-ins via `requirements.txt`.
```
danlapid, line 9: I don't know if this actually requires a changelog
This was never publicly released, it was only part of the private beta and all customers have already migrated away from it

vy-ton: I did not know it was private not public beta. Then the customer emails I sent should be sufficient

dom96: Alright, closing.

PR #31391, opened 2026-06-11
[Spectrum] Update Tunnel limitations to cover Private Load Balancing (LTM)

kai-cloudflare:
Adds details to the Spectrum limitations page explaining that while direct non-HTTP (TCP/UDP) routing from Spectrum to Cloudflare Tunnel is not supported directly, customers can configure a Cloudflare Load Balancer with Private Network Load Balancing (LTM) in between to successfully steer L4 traffic to private tunnel origins.


nikitacano:
/bonk call out that the proxy protocol doesn't work in this configuration

nikitacano (approved):

PR #31399, opened 2026-06-11
Python Workers docs/examples fixes.

dom96:
### Summary

Assortment of fixes to our docs based on agentic loop testing of Python Workers.

### DIFF
--- a/src/content/docs/workers/languages/python/examples.mdx
+++ b/src/content/docs/workers/languages/python/examples.mdx
@@ -72,8 +72,20 @@ from workers import WorkerEntrypoint, Response
 
 class Default(WorkerEntrypoint):
     async def fetch(self, request):
-        name = (await request.json()).name
+        body = await request.json()  # returns a native Python dict
+        name = body["name"]
         return Response("Hello, {name}".format(name=name))
+```
+
+## Return a JSON response
+
+```python
+from workers import WorkerEntrypoint, Response
+
+class Default(WorkerEntrypoint):
+    async def fetch(self, request):
+        data = {"greeting": "Hello, World!", "status": "ok"}
+        return Response.json(data)
 ```
 
 ## Read bundled asset files in your Worker
@@ -200,11 +212,13 @@ Refer to [Durable Objects documentation](/durable-objects/get-started/) for more
 ## Cron Trigger
 
 ```python
-from workers import WorkerEntrypoint, Response
+from workers import WorkerEntrypoint
 
 class Default(WorkerEntrypoint):
     async def scheduled(self, controller, env, ctx):
-  			print("cron processed")
+        # All four parameters (self, controller, env, ctx) are required —
+        # unlike fetch() which only takes (self, request).
+        print("cron processed")
 ```
 
 Refer to [Cron Triggers documentation](/workers/configuration/cron-triggers/) for more information.
--- a/src/content/docs/workers/languages/python/stdlib.mdx
+++ b/src/content/docs/workers/languages/python/stdlib.mdx
@@ -57,3 +57,9 @@ The following are present but cannot be imported due to a dependency on the term
 
 - pty
 - tty
+
+## In-memory filesystem
+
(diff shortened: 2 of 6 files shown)


src/content/docs/workers/languages/python/examples.mdx
```diff
@@ -72,10 +72,23 @@ from workers import WorkerEntrypoint, Response
...
 
+## Return a JSON response
+
+```python
+from workers import WorkerEntrypoint, Response
+
+class Default(WorkerEntrypoint):
+    async def fetch(self, request):
+        data = {"greeting": "Hello, World!", "status": "ok"}
+        # Use Response.json() to return JSON — do not pass a dict to Response() directly
```
ryanking13:
```suggestion
```

I guess this comment is unnecessary.

src/content/docs/workers/languages/python/examples.mdx
```diff
@@ -200,11 +213,13 @@ Refer to [Durable Objects documentation](/durable-objects/get-started/) for more
...
 
 ```python
-from workers import WorkerEntrypoint, Response
+from workers import WorkerEntrypoint
 
 class Default(WorkerEntrypoint):
     async def scheduled(self, controller, env, ctx):
-  			print("cron processed")
+        # All four parameters (self, controller, env, ctx) are required —
+        # unlike fetch() which only takes (self, request).
```
ryanking13, line 220:
```suggestion
```

This too?

src/content/docs/workers/languages/python/stdlib.mdx
```diff
@@ -57,3 +57,9 @@ The following are present but cannot be imported due to a dependency on the term
 
 - pty
 - tty
+
+## In-memory filesystem
+
+Python Workers have access to an ephemeral, in-memory filesystem. You can read and write files using standard Python file I/O (for example, `open()`, `pathlib.Path`), but all data is **lost when the Worker invocation ends**. The filesystem is not shared between invocations or instances.
```
ryanking13: I think this is not true. Files stored in the in-memory file system should be alive between invocations unless the isolate is ~~aborted~~ destroyed.

src/content/docs/workers/runtime-apis/handlers/scheduled.mdx
```diff
@@ -54,11 +54,13 @@ export default {
...
 
 ```python
-from workers import WorkerEntrypoint, Response, fetch
+from workers import WorkerEntrypoint
 
 class Default(WorkerEntrypoint):
     async def scheduled(self, controller, env, ctx):
-        await doSomeTaskOnASchedule()
+        # controller.cron contains the cron pattern that triggered this event
+        # controller.scheduledTime contains the scheduled time in ms since epoch
```
ryanking13, line 62:
```suggestion
```

This too? I guess AI agents are verbosely adding these kind of comment which I think is unnecessary to be included in the code example.

src/content/docs/workers/languages/python/stdlib.mdx
abstractedfox: I tested it to make sure and the local storage did persist between invocations

src/content/docs/workers/runtime-apis/handlers/scheduled.mdx
dom96: Yeah, they do. This is for docs though, so perhaps it is useful to mention it. That's the reason I left it in anyway.

src/content/docs/workers/languages/python/examples.mdx
dom96: I think this one is useful too, seems like a gotcha folks might run into

src/content/docs/workers/languages/python/stdlib.mdx
dom96: Yeah, missed this. Fixed!

ryanking13 (approved):

irvinebroque (approved):

PR #31426, opened 2026-06-12
More Python Workers examples in docs (email, rpc, handlers, streams)

dom96:
### Summary

Add more Python tabs across various docs pages. In particular email, rpc, handlers and streams.

### DIFF
--- a/src/content/docs/workers/runtime-apis/streams/index.mdx
+++ b/src/content/docs/workers/runtime-apis/streams/index.mdx
@@ -67,6 +67,20 @@ async function fetchAndStream(request) {
 }
 ```
 
+</TabItem> <TabItem label="Python" icon="seti:python">
+
+```python
+from workers import WorkerEntrypoint, Response, fetch
+
+class Default(WorkerEntrypoint):
+    async def fetch(self, request):
+        # Fetch from origin server.
+        response = await fetch(request)
+
+        # Stream the response body to the client.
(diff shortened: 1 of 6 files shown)


irvinebroque (approved):

src/content/docs/workers/runtime-apis/streams/index.mdx
```diff
@@ -125,6 +139,30 @@ async function fetchAndStream(request) {
...
+        enc = TextEncoder.new()
+
+        async def start(controller):
+            for i in range(5):
+                controller.enqueue(enc.encode(f"chunk {i}\n"))
+                await asyncio.sleep(0.1)
+            controller.close()
+
+        stream = ReadableStream.new(
+            to_js({"start": create_proxy(start)})
```
ryanking13, line 161: Probably we could create a wrapper for streams.

ryanking13 (approved):
Nice!

src/content/docs/workers/runtime-apis/streams/index.mdx
dom96: yeah, we should do so

dom96: basically I think we should get to a point where searching for `to_js` in our docs doesn't find anything (unless we are explicitly showing off `to_js` or FFI)

PR #31428, opened 2026-06-12
[Workers] Clarify observability pricing

WalshyDev:
## Summary

- Clarifies that Workers Logs pricing applies to persisted/queryable log events stored by Cloudflare.
- Clarifies that Workers Trace Events Logpush pricing applies to exported events.
- Documents that using both is additive where both overage rates apply.

## Validation

- `git diff --check origin/production...HEAD`
- Targeted pricing content assertions

Full docs validation was not run in this sandbox because it has Node 22 and no pnpm/corepack, while this repo requires Node 24.


irvinebroque:
/bonk update this PR to make @nevikashah codeowner of the workers observability partial that changed

PR #31437, opened 2026-06-12
harden fetch-ai-models against transient upstream failures

superhighfives:
### Summary

Makes `bin/fetch-ai-models.js` resilient to transient upstream failures.

The upstream API at `ai-cloudflare-com.pages.dev/api/models` fans out per-model schema fetches on cache miss and occasionally returns a 502 if any of those flake, which fails the whole sync (and any build that depends on it).

- Retries the request up to 4 times with exponential backoff + jitter
- Only retries on 429 and 5xx; other 4xx fail fast
- Adds a 30s per-request timeout via `AbortSignal.timeout`
- If all retries fail, logs and exits 0 so existing model JSON files stay in place rather than failing the build — the next successful run picks up any drift

### DIFF
--- a/bin/fetch-ai-models.js
+++ b/bin/fetch-ai-models.js
@@ -56,9 +56,7 @@ try {
 	// JSON files in place so docs continue to render the last known good
 	// catalog; the next successful run will sync new/removed models.
 	const message = err.message.endsWith(".") ? err.message : `${err.message}.`;
-	console.error(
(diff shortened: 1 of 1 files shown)


superhighfives:
@mvvmm, this should be good to go, and improves the script for fetching AI models.

mvvmm (approved):

PR #31443, opened 2026-06-14
[Durable Objects] Add apac-ne and apac-se location hints

samwang-cf:
## Summary

Adds two new finer-grained Asia-Pacific Durable Object location hints to the data location reference:

- `apac-ne` — Northeast Asia-Pacific
- `apac-se` — Southeast Asia-Pacific

Includes a note clarifying these are narrower sub-regions within `apac`, so developers can use the broad `apac` hint or narrow placement to the northeast or southeast to reduce latency for users concentrated there.

### DIFF
--- a/src/content/docs/durable-objects/reference/data-location.mdx
+++ b/src/content/docs/durable-objects/reference/data-location.mdx
@@ -118,7 +118,7 @@ North America instead.
 
 <sup>3</sup> `apac-ne` (Northeast Asia-Pacific) and `apac-se` (Southeast
 Asia-Pacific) are narrower sub-regions within `apac` (Asia-Pacific). Prefer
-`apac` for general Asia-Pacific placement; choose `apac-ne` or `apac-se` only
+`apac` for general Asia-Pacific placement. Choose `apac-ne` or `apac-se` only
 when your users are concentrated in the northeast or southeast of the region and
 you want to minimize latency for them.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/durable-objects/reference/data-location.mdx
```diff
@@ -102,10 +102,23 @@ Hints are a best effort and not a guarantee. Unlike with jurisdictions, Durable
 | weur      | Western Europe             |
 | eeur      | Eastern Europe             |
 | apac      | Asia-Pacific               |
+| apac-ne   | Northeast Asia-Pacific     |
+| apac-se   | Southeast Asia-Pacific     |
 | oc        | Oceania                    |
 | afr       | Africa <sup>2</sup>        |
 | me        | Middle East <sup>2</sup>   |
 
+:::note
```
iglesiasbrandon: a bit of a nit but should we add the information in this note right below as <sup>3</sup> instead? i dont want those footnotes to be drowned out by a note between the table and them?

samwang-cf: I like this idea better! Done.

iglesiasbrandon (approved):

samwang-cf:
Internal rollout for the new LocationHints are complete. We can merge now.

vy-ton:
@iglesiasbrandon let's combine with a changelog so api docs and changelog go together

PR #31521, opened 2026-06-17
 clarify required permissions

Techsurfer1981:
### Summary

Clarifies that Logpush API endpoints require specific Logs permissions (Zone Logs Read/Write or Account Logs Read/Write), and that general zone/account read access is insufficient.


rianvdm:
@Techsurfer1981 This is not correct. Logpush needs Logs:Write, and the doc already mentions that:

> Logs: Write - Grants read and write access to Logpull and Logpush, and read access to Instant Logs. Note that all Logpush API operations require Logs: Write permission because Logpush jobs contain sensitive information.

PR #31559, opened 2026-06-18
Add docs note about limitation on OTLP format.

JamieSinn:
### Summary

Cloudflare does not currently support the binary protobuf format for OTLP destinations, this is not well documented, and took a bit of figuring out when attempting to connect Worker traces to Dynatrace.

The only location that this is somewhat documented is here: https://developers.cloudflare.com/ai-gateway/observability/otel-integration/#common-otel-backends where it lists that the AI gateway supports the JSON or protobuf destination, but this doesn't highlight explicitly that the binary version of the protocol is not supported.

### Documentation checklist


irvinebroque (approved):
@nevikashah

PR #31562, opened 2026-06-18
flue: add generic code review alongside style-guide review

mvvmm:
Adds a generic engineering code review and a PR conventions check to the Flue bot. All three reviewers post to a single bot comment split into `### Code Review`, `### Conventions`, and `### Style Guide Review` sections.

## Architecture

All three reviewers run as **separate specialist workflows** (`code-review-specialist`, `conventions-specialist`, `style-guide-specialist`), each in their own Durable Object isolate with its own 128 MB memory budget. The orchestrator admits all three fire-and-forget; one failing degrades to an empty section rather than aborting the whole review.

### Code review specialist

- Reviews **all** changed text files (excluding lockfiles, generated output, binary/image files), capped at 20 files (largest-diff-first).
- Trusted TypeScript **pre-stages** the data before any model call: added lines are parsed from the diff patch (accurate new-file line numbers), and the full file content is fetched at the head SHA. Both are passed directly as `args.addedLines` and `args.fileContent`, eliminating the mandatory setup turns the model previously spent reading manifests and fetching files.
- `read_repo_file` (defaults to PR head SHA) and `search_repo` remain available as optional tools for cross-file lookups (callers, import sites, etc.).
- The root `AGENTS.md` is fetched from the PR **base** ref and injected as agent instructions so the reviewer has repo conventions without a checkout. (Base ref, not head, to prevent a PR from poisoning the reviewer's instructions.)
- Findings use a `critical` / `warning` / `suggestion` severity hierarchy with `CR-` IDs.

### Conventions specialist

- Checks PR-level concerns: whether the title and description explain the work, and whether the scope accurately reflects the changed files.
- Runs a single skill session (not per-file) — lightweight, fast.
- Findings use `CV-` IDs, `warning` severity only.

### Style guide specialist

- Reviews only `src/content/(docs|partials|changelog)/**.mdx`, capped at 20 files.
- Reads the diff and full file content from the staged workspace via the `code` tool.
- `warning` / `suggestion` severities only.

### Memory and timeout

- Each per-file agent session is **deleted immediately** after it finishes (`session.delete()`), bounding peak heap to ~concurrency live sessions rather than growing with the file count.
- Per-file **hard timeout** (10 min, env-overridable via `CODE_REVIEW_FILE_TIMEOUT_MS` / `STYLE_GUIDE_FILE_TIMEOUT_MS`): on timeout, the `CallHandle` is aborted first (settling the operation) and then the session is deleted, preventing zombie session leaks.
- Concurrency is env-overridable (`CODE_REVIEW_CONCURRENCY`, `STYLE_GUIDE_CONCURRENCY`) so local dev can run lower values without touching prod defaults.

### Other

- Reconciler handles each stream (`code`, `style`, `conventions`) independently against its prior findings + human comments; persists `{ code, style, conventions }` to R2.
- Incremental diff mode (SHA-pinned from last reviewed head to current head) when a prior review exists; full diff otherwise. Specialists self-heal incremental → full if the base SHA is gone.
- Cadence unchanged: automatic on open/synchronize/ready_for_review (up to 2-review cap), plus `/review`, `/full-review`, and `/ignore-review-limit` codeowner commands.
- Beta disclaimer rendered under the Code Review heading.

## Validated

`tsc --noEmit`, `flue build`, prettier. Verified end-to-end in `log` mode and `comment` mode against real PRs.


mvvmm:
**Findings not fixed (intentional):**

- **W4** — `env as Record<string, string>` cast in `getInternalHeaders`: pre-existing intentional pattern used consistently throughout the codebase.
- **W5** — `.catch(() => {})` on degraded R2 write in orchestrator: intentional — removing it would fail the orchestrator entirely, which is worse than the degraded write failing silently.
- **W15** (prior round) — line-agnostic finding ID hash: intentional trade-off, ID stability across commits is more valuable than uniqueness within a single run.
- **W10/W11** (second round) — `bin/clear-r2-pr-data.ts` issues: dev script, not production code.

kodster28 (approved):

PR #31578, opened 2026-06-19
Add Hypernode Hosting SSL certificate installation link

tdgroot:
### Summary

Our users are frequently using Cloudflare and could use a nudge into the right direction for setting up Cloudflare Origin CA.

Not sure where Cloudflare draws the line for adding vendors, but thought it was worth trying since AWS and GoDaddy are on the list as well. 

### Documentation checklist


baubuchon-cf:
Hi Timon,

Thanks for the PR, and for thinking of Hypernode's Cloudflare users.

I'm going to decline this one. That list is for guides that show how to install
an already-issued cert (like a Cloudflare Origin CA cert) on a given platform,
and the page it points to is really about buying SSL certs through Hypernode
rather than installing an Origin CA cert, so it isn't the right fit there.

Thanks again for contributing.

tdgroot:
Hi @baubuchon-cf thanks for your response.

Reason for declining makes sense! I linked specifically to a section of the page where we explain how to configure an already-issued cert, but I can see the confusion. Would you accept a PR that would target to a specific page with instructions for configuring an already-issued cert?

PR #31324, opened 2026-06-08
[DO] Document outbound connections keep Durable Objects alive

iglesiasbrandon:
### Summary

Documents the behavior change shipped in May 2026 where active outbound connections (TCP `connect()` and outbound WebSockets) now keep a Durable Object alive and prevent eviction for the duration of the connection, up to a 15-minute per-connection cap.

Before this change, the routing supervisor would evict a DO after 70–140 seconds of inactivity even if it had an open outbound connection — causing mid-stream evictions when, for example, streaming tokens from an LLM over a long-lived TCP socket.

**Changes:**

| File | Change |
| --- | --- |
| `concepts/durable-object-lifecycle.mdx` | Added outbound connections as a 5th hibernation-blocking condition; added a `:::note` block explaining the keepalive behavior, billing impact, 15-min cap, and that plain `fetch()` is excluded |
| `best-practices/websockets.mdx` | Clarified that outgoing WebSockets don't hibernate but *do* keep the DO alive — these were previously conflated |
| `platform/pricing.mdx` | Added billing caveat: an outbound connection keeps a DO in the billable non-hibernateable state even with no incoming requests |
| `runtime-apis/tcp-sockets.mdx` | Added Considerations bullet noting that a `connect()` socket opened inside a DO pins it for up to 15 minutes |
| `changelog/durable-objects/2026-06-19-...mdx` | New changelog entry |

**Key accuracy notes verified against the merged implementation:**
- 15-minute cap (`DEFAULT_HARD_LIMIT = 15 * kj::MINUTES`) is exact
- After the cap fires, the connection continues operating — only the DO pinning ends
- Plain `fetch()` is explicitly excluded (pass-through, no keepalive effect, even while streaming)
- `fetch()` upgraded to WebSocket via `Upgrade: websocket` IS covered

### Documentation checklist

- [x] Is there a [changelog](https://developers.cloudflare.com/changelog/) entry?
- [x] The change adheres to the [documentation style guide](https://developers.cloudflare.com/style-guide/).

PR #31593, opened 2026-06-19
[DO] Drop fetch/Hyperdrive exclusion paragraph from outbound connections changelog

iglesiasbrandon:
Drops the paragraph that described what the outbound connections keepalive does **not** cover.

Relates to #31324.


src/content/changelog/durable-objects/2026-06-19-outbound-connections-keep-dos-alive.mdx
threepointone, line 9:
```suggestion
Durable Objects now remain alive for the duration of active outbound connections created via [`connect()`](/workers/runtime-apis/tcp-sockets/) or an outbound WebSocket. Previously, a Durable Object would be evicted after 70-140 seconds of no incoming traffic, even if the object had an open outbound connection, which is a common pattern when streaming responses from a large language model (LLM).
```

threepointone: ^ this was meant to be an edit for that line, removing the TCP/WS reference at the end.

irvinebroque (approved):

joshthoward (approved):

PR #31609, opened 2026-06-21
WARP: document webkit2gtk3 dependency limitation on newer Fedora

### LINKED ISSUE #31057: WARP RPM fails to install — webkit2gtk3 dependency not found on Fedora 44
### Proposed changes

The WARP Linux installation documentation should include a note for Fedora 44 users about a dependency resolution failure caused by the RPM package requiring `webkit2gtk3`

### Subject Matter

When installing the WARP client RPM on Fedora, DNF fails with:    "nothing provides webkit2gtk3 needed by cloudflare-warp-[version]-1.x86_64"  Fedora uses a different package name for this library (`webkit2gtk4.0` or `webkitgtk` depending on the version). This is not documented anywhere in the WARP Linux setup guide and leaves Fedora users blocked with no documented workaround.

### Content Location

https://developers.cloudflare.com/cloudflare-one/team-and-resources/devices/cloudflare-one-client/download/#linux

### Additional information

**Affected package:** cloudflare-warp-2026.4.1350.0-1.x86_64.rpm
**Affected distro:** Fedora (tested on Fedora 44, x86_64)
**Package manager:** DNF

gords2:
Fixes #31057.

Users on newer Fedora releases (43/44) hit `nothing provides webkit2gtk3 needed by cloudflare-warp` when installing the RPM, with no explanation in the docs. The supported-versions table lists only Fedora 34/35; newer Fedora ships `webkit2gtk4.x` and no longer provides `webkit2gtk3`, which the `cloudflare-warp` RPM depends on — hence the failure.

Added a footnote on the Fedora entry (matching the existing `[^1]` RHEL/EPEL dependency footnote) documenting that the client UI requires `webkit2gtk3` and that Fedora 41+ does not provide it. This documents the limitation rather than recommending an unofficial workaround (the linked issue confirms there is no official fix yet).

### DIFF
--- a/src/content/partials/cloudflare-one/warp/system-requirements/linux.mdx
+++ b/src/content/partials/cloudflare-one/warp/system-requirements/linux.mdx
@@ -5,7 +5,7 @@
 
 |                       |                                                                               |
 | --------------------- | ----------------------------------------------------------------------------- |
-| **OS version**        | CentOS 8, RHEL 8, RHEL 9 [^1], Debian 12, Debian 13, Fedora 34, Fedora 35, Ubuntu 22.04 LTS, Ubuntu 24.04 LTS |
+| **OS version**        | CentOS 8, RHEL 8, RHEL 9 [^1], Debian 12, Debian 13, Fedora 34, Fedora 35 [^3], Ubuntu 22.04 LTS, Ubuntu 24.04 LTS |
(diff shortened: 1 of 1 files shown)


nikitacano (approved):

cf-rhett:
fyi -- this will be removed in the next week or so when we release support for 43/44 and deprecate 34/35 support

PR #31636, opened 2026-06-23
[Python] Remove all to_js usage from bindings examples

ryanking13:
### Summary

Removes remaining `pyodide.ffi.to_js` usage from workers bindings documentation. Python workers team has improved the SDK so that users does not have to manually convert Python objects into JS-serializable object before passing to RPC boundary.

### Documentation checklist


src/content/docs/workflows/python/bindings.mdx
ryanking13, line 1: This should be merged after https://github.com/cloudflare/workers-py/pull/138 is merged and released

irvinebroque (approved):

src/content/docs/workers/examples/cors-header-proxy.mdx
```diff
@@ -552,37 +549,38 @@ class Default(WorkerEntrypoint):
...
-            api_url2 = url.searchParams["apiurl"]
+            parsed = urlparse(request.url)
+            params = parse_qs(parsed.query)
+            api_url2 = params.get("apiurl", [None])[0]
 
             if not api_url2:
                 api_url2 = api_url
 
+            target = urlparse(api_url2)
             request = Request.new(api_url2, request)
```
ryanking13, line 560: @dom96 @hoodmane 

We don't have a corresponding method for this (cloning the Response with a new URL). I guess we should add it in our SDK

dom96: Request has a `clone` method, maybe we can give it an optional URL argument?

src/content/docs/workers/examples/modify-request-property.mdx
```diff
@@ -164,34 +161,34 @@ class Default(WorkerEntrypoint):
...
 
-        new_request.headers["X-Example"] =  "bar"
+        new_request.headers["X-Example"] = "bar"
         new_request.headers["Content-Type"] = "application/json"
 
         try:
             return await fetch(new_request)
         except Exception as e:
-            return Response.new({"error": str(e)}, status=500)
+            return Response(str(e), status=500)
```
dom96: I wonder if the idea here is to return json. Maybe you can use `Response.json` here?

dom96 (approved):
Nice, thanks for fixing these!

src/content/docs/workers/examples/modify-request-property.mdx
ryanking13: Yeah, makes sense.

PR #31638, opened 2026-06-23
[DNS] Document child zone delegation behavior for subdomain setup decommissioning

ngayerie:
## Summary

When a subdomain is set up as a child zone on Cloudflare, NS delegation is served automatically based on the child zone's active status — not based on NS records stored in the parent zone. Customers may expect that deleting NS records from the parent zone stops the delegation, but it doesn't. The child zone must be deleted to stop serving delegation.

This PR adds a **"Decommission a subdomain setup"** section to the [subdomain setup page](/dns/zone-setups/subdomain-setup/setup/) with clear steps.

## Changes

| File | Change |
|------|--------|
| `dns/zone-setups/subdomain-setup/setup/index.mdx` | New section: "Decommission a subdomain setup" with explanation and 4-step procedure |

## Stats

1 file changed, +13 insertions

Resolves [DEE-3626](https://jira.cfdata.org/browse/DEE-3626)

### DIFF
--- a/src/content/docs/dns/zone-setups/subdomain-setup/setup/index.mdx
+++ b/src/content/docs/dns/zone-setups/subdomain-setup/setup/index.mdx
@@ -11,7 +11,7 @@ head:
     content: Subdomain delegation and available setups
 ---
 
-import { DirectoryListing, GlossaryTooltip, Render } from "~/components";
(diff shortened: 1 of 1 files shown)


ngayerie:
Closing after chatting with @hannes-cf

PR #31639, opened 2026-06-23
[DNS] Remove Enterprise quota from DNS records documentation

ngayerie:
Removed Enterprise quota from the DNS records management documentation.


ngayerie:
@hannes-cf will raise a PR on his side to correct this

hannes-cf (approved):

PR #31655, opened 2026-06-23
Add wshim instrumentation and availability SLO

mvvmm:
## Summary

Adds Prometheus metric emission via wshim and an OpenSLO availability SLO definition for `developers.cloudflare.com`.

### Worker instrumentation
- **`wrangler.jsonc`**: Adds `internal_service_http` binding for wshim (auto-approved for internal CF tenant, no binding-grants PR needed)
- **`worker/worker-configuration.d.ts`**: Adds `WSHIM_SOCKET: Fetcher` and `WSHIM_TOKEN: string`
- **`worker/index.ts`**: Refactors `fetch()` to extract `handleFetch()` — ensures all request paths emit metrics (previously only the final asset-fetch path would have). Emits `cloudflare_docs_requests_total{status}` counter per request via `ctx.waitUntil()` — zero latency impact, errors are swallowed
- **`package.json`**: Adds `promjs` for Prometheus text format encoding (public npm package)

### SLO definition
- **`.cloudflare/health/cloudflare-docs-availability-slo.yaml`**: OpenSLO definition targeting 99.95% availability over 30 days, pointing at `cloudflare_docs_requests_total` in Thanos. Push to SLO Directory manually via `sloctl sync` after deploy.

## Before merging
1. `WSHIM_TOKEN` must be set as a Worker secret (`wrangler secret put WSHIM_TOKEN`)
2. Verify metrics appear in Grafana/Thanos after deploy: `cloudflare_docs_requests_total`
3. Once metrics are confirmed, run `sloctl sync .cloudflare/health/cloudflare-docs-availability-slo.yaml`

## Performance
Metric emission runs entirely in `ctx.waitUntil()` after the response is returned. No impact on TTFB or response latency. Wshim errors are caught and swallowed.

### DIFF
--- a/.cloudflare/health/cloudflare-docs-availability-slo.yaml
+++ b/.cloudflare/health/cloudflare-docs-availability-slo.yaml
@@ -1,0 +1,37 @@
+apiVersion: openslo/v1
+kind: SLO
+metadata:
+  name: cloudflare-docs-availability-slo
+  displayName: Cloudflare Docs Availability SLO
+spec:
+  service: component:default/cf-account-b54f07a6c269ecca2fa60f1ae4920c99
+  description: |
+    Measures the percentage of requests to developers.cloudflare.com
+    that succeed (non-5xx responses).
+  indicator:
+    metadata:
+      name: cloudflare-docs-availability-slo
+      displayName: Cloudflare Docs Availability Rate
+      annotations:
+        release-manager.cfdata.org/hmd: "false"
+        openslo.cfdata.org/sli-type: availability
+    spec:
+      ratioMetric:
+        counter: true
+        good:
+          metricSource:
+            metricSourceRef: thanos
+            spec:
+              query: cloudflare_docs_requests_total{status!~"5.*"}
+        total:
+          metricSource:
+            metricSourceRef: thanos
+            spec:
+              query: cloudflare_docs_requests_total
+  objectives:
+    - displayName: Availability > 99.95%
+      target: 0.9995
+  budgetingMethod: Occurrences
+  timeWindow:
+    - duration: 30d
+      isRolling: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -342,8 +342,8 @@ jobs:
 
           echo "branch_slug=$BRANCH_SLUG" >> "$GITHUB_OUTPUT"
 
-          pnpm exec wrangler deploy --dispatch-namespace preview-deployments --name $SHORT_SHA
-          pnpm exec wrangler deploy --dispatch-namespace preview-deployments --name $BRANCH_SLUG
+          pnpm exec wrangler deploy --dispatch-namespace preview-deployments --name $SHORT_SHA --env preview
+          pnpm exec wrangler deploy --dispatch-namespace preview-deployments --name $BRANCH_SLUG --env preview
 
       - name: Post preview URL on PR
         env:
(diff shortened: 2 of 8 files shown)


mvvmm:
## Review response

**Warning: Bare counter queries in SLO YAML** — False positive. The SLO Directory platform handles counter aggregation automatically when `counter: true` is set. Raw counter queries without `rate()` or `increase()` are the correct and documented approach for this platform.

**Warning: Counter label not declared** — False positive. `promjs` does not have a label declaration API. Labels are passed directly to `inc({ status: ... })` without pre-declaring them — this is the correct usage for this library. The reviewer appears to be conflating with the Go/Python Prometheus client pattern.

**Warning: Counter resets every request** — False positive. This is intentional and the correct pattern for wshim. Each Worker instance pushes its partial count (0→1) per request, and wshim aggregates across all instances. This is how the system is designed to work.

**Warning: Unhandled failures not instrumented** — Acknowledged but intentionally not addressed. `handleFetch` has comprehensive try/catch internally and is not expected to throw. Adding a wrapper for this edge case adds complexity without meaningful benefit.

**Suggestion: Silently swallowed errors** — Intentional. Metric emission is non-critical infrastructure. Logging wshim errors would add noise to production logs for a non-user-facing concern.

mvvmm:
## Review response (commit `c1cdd98`)

**Suggestion: Redundant catch-rethrow** — Already fixed in the subsequent commit (`2d61ea9`). The `catch (err) { throw err; }` block was removed to resolve a `no-useless-catch` ESLint error.

**Suggestion: HELP/TYPE metadata emitted per sample** — False positive for wshim. wshim is not a standard Prometheus scrape endpoint — it's an aggregating push gateway designed to receive per-request Prometheus text payloads from Worker instances. Including HELP/TYPE metadata with each push is the expected and correct usage for this system.

**Warning: Silent metric failures** — Intentional. Metric emission is non-critical infrastructure. Logging wshim errors or HTTP 4xx/5xx responses would add noise to production logs for a non-user-facing concern. If the token is misconfigured, the SLO data will simply show gaps, which is detectable in Grafana.

mvvmm:
## Review response

**Critical: Metric name mismatch** — This is a false positive. wshim automatically prefixes every emitted metric with the service name from the token. Our token was generated with `service=cloudflare_docs`, so a metric emitted as `requests_total` is stored in Prometheus as `cloudflare_docs_requests_total`.

Applying the suggested fix and emitting `cloudflare_docs_requests_total` would result in the metric being stored as `cloudflare_docs_cloudflare_docs_requests_total` — double-prefixed and broken.

The current code is correct as-is.

mvvmm:
/bonk please approve the pr, you're still requesting changes which is blocking me from merging. you need to approve the pr

kodster28 (approved):
dm'd you internally, but going ahead and approving broadly

.cloudflare/health/cloudflare-docs-availability-slo.yaml
```diff
@@ -0,0 +1,37 @@
...
+      annotations:
+        release-manager.cfdata.org/hmd: "false"
+        openslo.cfdata.org/sli-type: availability
+    spec:
+      ratioMetric:
+        counter: true
+        good:
+          metricSource:
+            metricSourceRef: thanos
+            spec:
```
danlapid, line 24: Instead of the flimsy approach with WSHIM you can use builtin runtime metrics instead with a way simpler config.
See https://backstage.cfdata.org/docs/default/component/release-manager/User_Reference/Builtin_Health_Signals/#worker-specific-built-in-health-signals

mvvmm: Perfect! I knew something like this had to exist. Thanks.

mvvmm:
Closing, moved SLO's internal.

PR #31683, opened 2026-06-25
[WAF] Improve managed rules troubleshooting: payload logging and malicious uploads

zeinjaber:
Resolves DEE-3647.

Adds two pieces of guidance to the WAF managed rules troubleshooting page that were missing from the page:

- **Payload logging** (false positives): tells readers to use payload logging to see which part of a request matched a rule, and to configure it ahead of time so the payload is captured for the next investigation. Links to the payload logging overview and the dashboard configuration page. Notes it is Enterprise-only.
- **Malicious uploads / binary content** (false positives, Additional recommendations): explains that managed rulesets inspect standard HTTP content, that binary uploads can cause false positives, and points to Malicious uploads detection for scanning file uploads.

### Intentionally not changed
The ticket also suggested adding WAF attack score guidance for false negatives and clarifying exception-vs-disable behavior. Both are already documented on this page (the *Troubleshoot false negatives* section already recommends WAF attack score, and the *Troubleshoot false positives* section already covers exceptions vs disabling rules), so they were left as-is to avoid duplication.

### Notes
- Docs-only change; no product/runtime changes.
- All internal links verified against existing pages.

### DIFF
--- a/src/content/docs/waf/managed-rules/troubleshooting.mdx
+++ b/src/content/docs/waf/managed-rules/troubleshooting.mdx
@@ -22,6 +22,8 @@ By default, WAF's managed rulesets are compatible with most websites and web app
 
 You can use [Security Events](/waf/analytics/security-events/) to help you identify what caused legitimate requests to get blocked. Add filters and adjust the report duration as needed.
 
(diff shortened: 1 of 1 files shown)


zeinjaber:
@elithrar @danielegm @hsaxenaCF - can you review this PR, please? Thanks

zeinjaber:
@elithrar @danielegm - could you please review this PR? Thanks.

xmflsct (approved):

PR #31714, opened 2026-06-26
Add Sematext OTLP endpoints to documentation

otisg:
### Summary

Added Sematext and its endpoints for traces and logs


otisg:
Re `/v1/traces` and `/v1/logs` missing from this PR for Sematext. We actually allow people to send traces/logs/metrics to just the root URI as well, so that's why I provided that endpoint. If that's a problem from Cloudflare's side I can change the PR to specify the full `/v1/traces` and `/v1/logs`.

otisg:
@irvinebroque Is the versioned URI path really required for this? If yes, I'll update the PR. I thought I would provide a simpler/shorter URL, but if you require the versioned URI please let me know.

irvinebroque (approved):

otisg:
Thanks @irvinebroque!

PR #31729, opened 2026-06-26
docs: update hyperdrive observability metrics documentation

ivoryibu:
Clarifies how Hyperdrive connection pool metrics are reported, to reduce confusion from customers who see "Open Connections" plateau below their configured limit and assume the product is throttling them. This is a recurring escalation path where the pooling behavior is healthy but the per-machine metric presentation looks broken.

- Add a "How Hyperdrive distributes your connection pool" subsection to **Pool size metrics** explaining that the pool is sharded across 5 machines per Cloudflare location, so the configured maximum is divided across them (for example, a limit of 100 gives each machine 20). This explains why `max.currentPoolSize` plateaus at the per-machine slice rather than the full limit even when healthy.
- Note that pools start empty and reuse connections, so Hyperdrive opens few origin connections — addressing customers who expect to see connection counts climb toward their full limit.
- Add a note clarifying that `waitingClients` is a point-in-time count, not a duration. A brief wait (typically milliseconds) registers as a spike, which customers misread as clients being stuck.
- State that pool size metrics are currently per-machine and not aggregated, and tell customers to sum per-location values to estimate total usage. Phrased as current behavior to leave room for a future aggregate ("All locations") view.
- Add context to the **dashboard** section noting the Pool connections chart shows per-machine values.
- Add a **GraphQL API** note that `hyperdrivePoolSizesAdaptiveGroups` reports per-machine values grouped by `coloCode`, and that summing across `coloCode` groups is the closest available estimate of total usage.

Source: recent connection-limit escalations and the near-term documentation actions in the pool metrics PRD. No data pipeline or schema changes are involved.

### DIFF
--- a/src/content/docs/hyperdrive/observability/metrics.mdx
+++ b/src/content/docs/hyperdrive/observability/metrics.mdx
@@ -45,21 +45,13 @@ The `hyperdrivePoolSizesAdaptiveGroups` dataset contains the following connectio
 | Peak open connections | `max.currentPoolSize`    | Peak number of connections open in the pool.                      |
 | Peak waiting clients  | `max.waitingClients`     | Peak number of clients waiting for a connection from the pool.    |
 
-#### How Hyperdrive distributes your connection pool
+Hyperdrive distributes your connection pool across multiple machines within each Cloudflare location. Pool size metrics report aggregate values across all pool shards for a given configuration. The dashboard and GraphQL API both reflect the total connection usage for each location.
 
-Hyperdrive shards your connection pool across five machines within each Cloudflare location. Your configured maximum is divided evenly across these machines. For example, a connection limit of 100 gives each machine a maximum of 20 connections (100 / 5).
-
-Each machine starts with an empty pool and reuses open connections. Hyperdrive opens few connections to your origin database, and each pool grows only as concurrent demand requires it.
-
-Pool size metrics currently report per-machine values for each location. They are not aggregated into a single total across all machines and locations. As a result, `max.currentPoolSize` plateaus at the per-machine slice (for example, 20) rather than your full configured maximum (for example, 100), even when the pool is operating normally.
+Connection contention appears as a sustained spike in waiting clients, or when open connections consistently approach the pool size maximum. If your open connections regularly approach this limit, consider [increasing your Hyperdrive connection limit](/hyperdrive/platform/limits/#request-a-limit-increase).
 
 :::note
-A spike in waiting clients does not mean clients are stuck waiting. The `waitingClients` metric is a point-in-time count of clients waiting for a connection at the moment each value is recorded. It does not measure how long a client waited, which is typically a few milliseconds.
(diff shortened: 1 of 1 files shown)


src/content/docs/hyperdrive/observability/metrics.mdx
```diff
@@ -36,16 +36,30 @@ The `volatile` cache status indicates the query contains a PostgreSQL function c
...
+| Avg. waiting clients  | `avg.waitingClients`     | Average number of clients waiting for a connection from the pool. |
+| Pool size maximum     | `max.maxPoolSize`        | Configured maximum size of the connection pool.                   |
+| Peak open connections | `max.currentPoolSize`    | Peak number of connections open in the pool.                      |
+| Peak waiting clients  | `max.waitingClients`     | Peak number of clients waiting for a connection from the pool.    |
 
-Connection contention appears as a spike in waiting clients, or when open connections consistently approach the pool size maximum. If your open connections regularly approach this limit, consider [increasing your Hyperdrive connection limit](/hyperdrive/platform/limits/#request-a-limit-increase).
+#### How Hyperdrive distributes your connection pool
+
+Hyperdrive shards your connection pool across five machines within each Cloudflare location. Your configured maximum is divided evenly across these machines. For example, a connection limit of 100 gives each machine a maximum of 20 connections (100 / 5).
```
vy-ton: We should not be so specify with 5 machines as that could change. "Hyperdrive distributes your connection pool across multiple locations..."

src/content/docs/hyperdrive/observability/metrics.mdx
```diff
@@ -36,16 +36,30 @@ The `volatile` cache status indicates the query contains a PostgreSQL function c
...
+
+Each machine starts with an empty pool and reuses open connections. Hyperdrive opens few connections to your origin database, and each pool grows only as concurrent demand requires it.
+
+Pool size metrics currently report per-machine values for each location. They are not aggregated into a single total across all machines and locations. As a result, `max.currentPoolSize` plateaus at the per-machine slice (for example, 20) rather than your full configured maximum (for example, 100), even when the pool is operating normally.
```
vy-ton: See metrics change in flight so we should update these docs to reflect that end state

src/content/docs/hyperdrive/observability/metrics.mdx
irvinebroque: big +1

irvinebroque: Why does a customer need to think about this at all?

src/content/docs/hyperdrive/observability/metrics.mdx
```diff
@@ -36,16 +36,30 @@ The `volatile` cache status indicates the query contains a PostgreSQL function c
...
+
+:::note
+A spike in waiting clients does not mean clients are stuck waiting. The `waitingClients` metric is a point-in-time count of clients waiting for a connection at the moment each value is recorded. It does not measure how long a client waited, which is typically a few milliseconds.
```
irvinebroque: I don't understand

irvinebroque (changes requested):

PR #31756, opened 2026-06-28
docs: add PowerShell Pages deployment cleanup example

### LINKED ISSUE #31753: Add Windows PowerShell alternative for "delete project with too many deployments" guide

The current documentation for deleting a project with too many deployments relies on a shell script workflow that works for Linux/macOS shells, but is not Windows-friendly as written. It also throws a "jq not found" error in WSL for me.

Please add examples in tabs to the guide:

- Shell script (Linux/macOS)
- PowerShell `.ps1` (Windows)

```ps1
$prod_id = ""
while ($true) {
    $json = npx wrangler pages deployment list --project-name <PROJECT_NAME> --json 2>$null
    $ids = ($json | ConvertFrom-Json) | Select-Object -ExpandProperty id
    $to_delete = $ids | Where-Object { $_ -ne $prod_id }
    if (-not $to_delete) {
        Write-Host "Done. Production: $prod_id"
        break
    }
    Write-Host "Deleting $(@($to_delete).Count) deployments..."
    foreach ($id in $to_delete) {
        $output = npx wrangler pages deployment delete $id --project-name <PROJECT_NAME> --force 2>&1
        if ($output -notmatch "Successfully deleted") {
            if ($output -match "active production deployment") {
                $prod_id = $id
            }
        }
    }
}
```

This would make the guide much more accessible for Windows users.

### Additional information

I _could_ make a PR, but I don't think I've got the time anytime soon. Maybe someone else is faster than me. If not, I will do it.

AlexMnrs:
## Summary
- add tabs for the Pages deployment cleanup script
- keep the existing macOS/Linux shell example
- add a Windows PowerShell version that uses `ConvertFrom-Json` instead of `jq`

Closes #31753

## Validation
- `git diff --check`

Note: I attempted `pnpm exec prettier --check src/content/docs/pages/platform/known-issues.mdx`, but it timed out locally after the dependency install did not complete within two minutes. I stopped the stale local processes and did not keep generated dependency files.

### DIFF
--- a/src/content/docs/pages/platform/known-issues.mdx
+++ b/src/content/docs/pages/platform/known-issues.mdx
@@ -9,7 +9,7 @@ products:
 
 ---
 
-import { DashButton } from "~/components";
+import { DashButton, Tabs, TabItem } from "~/components";
 
 Here are some known bugs and issues with Cloudflare Pages:
 
@@ -106,7 +106,10 @@ npx wrangler pages deployment delete <DEPLOYMENT_ID> --project-name <PROJECT_NAM
 
 Use the `--force` flag to skip the confirmation prompt and to force deletion of aliased deployments.
 
-To delete _all_ your deployments for a particular project name, you could run the following shell script:
+To delete _all_ your deployments for a particular project name, you could run one of the following scripts:
+
+<Tabs>
+	<TabItem label="macOS and Linux">
 
 ```sh
 prod_id=""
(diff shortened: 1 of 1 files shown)


MattieTK:
I’ve reworked this a bit along with the shell script at the same time to catch the production env, throw if we don’t detect it, improve parallel deletion and use project name as an argument.

AlexMnrs:
Thanks for tightening both examples, Mattie. The production-environment detection and clearer argument-based usage make the guide much stronger.

MattieTK:
Moved to #32273 — pushing to this branch requires write access to your fork, which I don't have, so I opened the continuation from a branch on the main repo instead. Closing this one.

PR #31788, opened 2026-06-29
Clarify Keyless SSL certificates and add DNS-only key server note

baubuchon-cf:
Improves the Keyless SSL "Public DNS" setup page:

- Adds a **Certificates used in Keyless SSL** section with a table distinguishing the **edge (Keyless SSL) certificate** (SAN = site hostnames) from the **key server authentication certificate** (SAN = key server hostname).
- Adds a caution: do not add the key server hostname to the public edge cert SAN — not required, and it leaks internal hostnames into Certificate Transparency logs.
- Adds a note that a key server hostname on a Cloudflare zone must use a **DNS-only (grey-cloud)** record, or the edge returns `NXDOMAIN` and the handshake fails.

### DIFF
--- a/src/content/docs/ssl/keyless-ssl/configuration/public-dns.mdx
+++ b/src/content/docs/ssl/keyless-ssl/configuration/public-dns.mdx
@@ -39,7 +39,7 @@ As a security measure, you should hide the hostname of your key server.
 
 :::note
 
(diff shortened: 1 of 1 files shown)


baubuchon-cf:
@elithrar -  mind giving this a CODEOWNER review? Clarifies the two Keyless certs (edge vs. key-server auth) and fixes a DNS note — proxied vs. missing records fail differently (edge-IP resolution vs. NXDOMAIN), not both as NXDOMAIN. Docs-only, bots green. Thanks!

lgarofalo (approved):

baubuchon-cf:
Hey @elithrar  — Leland reviewed and approved this one. Could you give it a product-owner approval when you get a chance? Thanks!

PR #31807, opened 2026-06-30
[Access] Clarify Okta group limits

kennyj42:
Clarifies Okta group limit behavior in the Cloudflare One Okta identity provider docs. This separates the tenant-level case where Cloudflare needs an Okta API token to retrieve more than 100 groups from Okta's OIDC token group-claim limitation, where a user token with more than 100 groups can omit policy-relevant groups.\n\nValidation:\n- pnpm run format:core:check\n- pnpm run check\n- git diff --check


jroyal:
lgtm

elithrar (approved):

PR #31812, opened 2026-06-30
docs(d1): add baseline migration pattern for existing databases

ivoryibu:
## Summary

Adds a new section to the D1 migrations page documenting how to adopt the migration system on an existing database that already has tables.

### Problem

The current migrations documentation assumes you are starting from scratch. Developers who have an existing D1 database with tables and want to start using `wrangler d1 migrations` have no guidance on how to create a baseline migration that captures the current schema without breaking anything.

### Changes

Adds an "Adopt migrations on an existing database" section to `src/content/docs/d1/reference/migrations.mdx` that covers:

- When a baseline migration is useful (teams, CI/CD, schema history) and when you can skip it (solo developer, single database)
- How to export the current schema with `sqlite_master`
- How to create a baseline migration using `CREATE TABLE IF NOT EXISTS` so it is safe to apply against the live database
- How to apply locally and remotely

### Context

Discovered during hands-on testing of D1 schema management workflows. The `CREATE TABLE IF NOT EXISTS` pattern ensures the migration records itself in `d1_migrations` without attempting to recreate tables that already exist.

### DIFF
--- a/src/content/docs/d1/reference/migrations.mdx
+++ b/src/content/docs/d1/reference/migrations.mdx
@@ -92,15 +92,15 @@ If your D1 database already has tables and you want to start using the migration
 
 A baseline migration records your existing schema as the first entry in your migration history, so every future migration builds on a known starting point. This lets anyone recreate the database from scratch — for example, when a new developer sets up a local database, when CI/CD spins up fresh test databases, or when you want a complete schema history in your git repository.
 
-The baseline is a suggestion, not a requirement. Migrations are useful even if you work solo, so that you can roll your schema forward and back over time. You only need a baseline if you want to reproduce the database from your migration files. If you never recreate the database and only ever apply new migrations to it, you can skip this step and start with your next migration.
+The baseline is a suggestion, not a requirement. Migrations are useful even if you work solo because they track schema changes over time. You only need a baseline if you want to reproduce the database from your migration files. If you never recreate the database and only ever apply new migrations to it, you can skip this step and start with your next migration.
 
 <Steps>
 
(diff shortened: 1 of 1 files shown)


src/content/docs/d1/reference/migrations.mdx
```diff
@@ -86,6 +86,49 @@ The pattern is a standard glob — `*` matches one path segment, `**` matches an
 
 `wrangler d1 migrations create` only writes top-level files inside `migrations_dir`, so if your `migrations_pattern` only matches nested files (as with the Drizzle layout), generate new migrations using your ORM's command (for example, `drizzle-kit generate`) instead.
 
+## Adopt migrations on an existing database
```
irvinebroque: is "adopt" the right verb here?

src/content/docs/d1/reference/migrations.mdx
```diff
@@ -86,6 +86,49 @@ The pattern is a standard glob — `*` matches one path segment, `**` matches an
...
+
+If your D1 database already has tables and you want to start using the migration system, create a baseline migration that captures your current schema.
+
+This is useful when you work on a team and a new developer needs to set up a local database from scratch, when you use CI/CD that creates fresh test databases, or when you want a complete schema history in your git repository. If you are a solo developer with one database and do not plan to recreate it, you can skip this step.
```
irvinebroque: I don't follow — if I am building an app just myself I still want to have migrations that i can roll forward and back? if i'm building a rails app i still use active record migrations for my own app: https://guides.rubyonrails.org/active_record_migrations.html

PR #31813, opened 2026-06-30
docs(d1): add ORM migration workflow and Drizzle integration warnings

ivoryibu:
## Summary

Documents the workflow for using ORMs (specifically Drizzle) with D1 migrations, and adds warnings about known gotchas that can cause data loss.

### Problem

The D1 docs list Drizzle as a community project and mention its nested migration layout, but do not document:

1. That `drizzle-kit migrate` does not work with D1 (you must use Wrangler to apply)
2. That `drizzle-kit generate` will generate `DROP TABLE` statements for tables not in your schema file — including Wrangler's own `d1_migrations` tracking table — which can cause data loss if applied blindly
3. The `tablesFilter` config option that prevents this
4. That Drizzle and Wrangler track migrations separately and don't communicate
5. When to use an ORM vs. raw SQL migrations

### Changes

**`src/content/docs/d1/reference/migrations.mdx`**
- Adds "Using an ORM for migrations" section explaining the generate-then-apply workflow
- Adds "When to use an ORM vs. raw SQL" guidance

**`src/content/docs/d1/reference/community-projects.mdx`**
- Adds a caution callout under the Drizzle ORM entry with the three specific gotchas and their fixes

### Context

Discovered during hands-on testing of D1 schema management. When using `drizzle-kit generate` against a D1 database, the generated migration attempted to drop the `d1_migrations` table and recreate the application table from scratch (instead of using `ALTER TABLE ADD COLUMN`). Both behaviors are dangerous if not caught during review.


irvinebroque (approved):
Glad we are documenting what is not working — important that we make this work better with drizzle

PR #31814, opened 2026-06-30
docs(d1): add rules of D1 best practices guide

ivoryibu:
## Summary

Rebased this PR onto the current `production` branch and revised the new **Rules of D1** best-practices guide under `src/content/docs/d1/best-practices/rules-of-d1.mdx`.

The guide now prioritizes D1 product guidance and addresses PR review feedback:

- Keeps the opening flow as **Overview → Limits → Scale out across many databases** and removes the separate **When to use D1** section while preserving alternative-storage guidance in the overview.
- Simplifies the overview language to **“D1 is a relational database.”**
- Reduces unnecessary hyphenation while keeping **key-value**.
- Tightens cautious sharding guidance: only shard with a stable shard key and fixed shard count, notes that D1 does not provide automatic resharding tools, and removes the misleading dynamic database example.
- Clarifies Workers for Platforms guidance around user Workers with isolated D1 bindings.
- Reworks indexing guidance with `EXPLAIN QUERY PLAN` examples before and after creating an index, and explains why adding indexes to large tables can stall a single threaded database.
- Switches `D1 Insights` wording to **D1 query insights**.
- Clarifies batch guidance and per query limits.
- Preserves the explicit engineering retry guidance: exponential backoff, full jitter, three-attempt maximum, and retry only retryable `D1_ERROR` codes from the documented error list.
- Adds chunked bulk mutation guidance, including indexing the column used to select chunks.
- Adds migration tooling guidance for hand written SQL versus ORM generated SQL, including generated migration review and `d1_migrations` risk.
- Removes the arbitrary table-count threshold from raw SQL migration tooling guidance.
- Adds guidance to run migrations and index creation outside Worker startup paths.
- Adds local before remote migration guidance with rollback behavior.
- Adds location hint, jurisdiction, and Sessions API/read replication guidance.
- Adds Data location to related resources.
- Addresses remaining style feedback: sentence case table headers, monospace `wrangler`, acronym expansion for object relational mapping (ORM) and create/read/update/delete (CRUD), spelled out small numbers, and read only `WITH` retry wording.

## Validation

- Rebased onto the latest `origin/production` during PR monitoring.
- Ran `git diff --check` successfully after the latest rebase.
- Ran `pnpm exec prettier --check src/content/docs/d1/best-practices/rules-of-d1.mdx` under Node 24 successfully.
- Ran `pnpm exec astro check` under Node 24; result: 0 errors, 0 warnings, existing repository hints only.

## Notes

A follow-up PR can add a dedicated Drizzle-specific D1 page as discussed.

### DIFF
--- a/src/content/docs/d1/best-practices/rules-of-d1.mdx
+++ b/src/content/docs/d1/best-practices/rules-of-d1.mdx
@@ -31,7 +31,7 @@ A single D1 database has fixed resources. You do not choose instance sizes, CPU,
 
 The first limits to design around are:
 
-| Limit                              | Workers Paid | Workers Free |
+| Limit                              | Workers paid | Workers free |
 | ---------------------------------- | ------------ | ------------ |
 | Databases per account              | 50,000       | 10           |
 | Maximum database size              | 10 GB        | 500 MB       |
@@ -264,7 +264,7 @@ Do this for every dynamic clause: `ORDER BY`, sort direction, optional filters, 
 
 ### Implement application level retry with exponential backoff and jitter for write queries
 
-D1 automatically retries `SELECT`, `EXPLAIN`, and `WITH` queries up to two times on transient errors. Write queries are never automatically retried. Your application must implement its own retry for idempotent write operations.
+D1 automatically retries read only queries up to two times on transient errors. Only queries containing solely `SELECT`, `EXPLAIN`, or read only `WITH` statements are retried. Any query that can write is never automatically retried. Your application must implement its own retry for idempotent write operations.
 
 Make writes idempotent before you retry them. For order creation, require a client generated idempotency key and enforce it with a `UNIQUE` constraint. If a transient error occurs after D1 commits the write but before the Worker receives the response, the retry returns the existing order instead of creating a duplicate. If a retry hits the unique constraint, treat that as a signal to fetch and return the existing record rather than issuing a second logical write.
 
@@ -391,7 +391,7 @@ An ORM such as Drizzle or Prisma works well when:
 - You want autocompletion in your editor when writing queries.
 - You are iterating rapidly on schema design and want to see what SQL each change produces before applying it.
 
-If you use an ORM, be aware that ORM migration generators cannot apply migrations to D1 directly. Generate the SQL with your ORM, then apply it with `wrangler`. Review all generated SQL before applying it, especially as described in [Review generated migrations before applying them](#review-generated-migrations-before-applying-them), because ORMs may generate destructive statements for tables they do not manage, including Wrangler's `d1_migrations` table.
+If you use an ORM, be aware that ORM migration generators cannot apply migrations to D1 directly. Generate the SQL with your ORM, then apply it with `wrangler`. Review all generated SQL before applying it, especially as described in [Review generated migrations before applying them](#review-generated-migrations-before-applying-them), because ORMs may generate destructive statements for tables they do not manage, including `wrangler`'s `d1_migrations` table.
 
 ### Apply migrations outside your Worker startup path
 
@@ -428,11 +428,11 @@ CREATE TABLE IF NOT EXISTS orders (
 );
 ```
 
-This pattern lets Wrangler record the migration without trying to recreate tables that already exist. After the baseline is applied, future migrations should contain only the incremental schema changes.
+This pattern lets `wrangler` record the migration without trying to recreate tables that already exist. After the baseline is applied, future migrations should contain only the incremental schema changes.
 
 ### Review generated migrations before applying them
 
-Never apply ORM generated SQL blindly. A migration generator only knows the schema model it was given. If Wrangler's `d1_migrations` table, manually created indexes, or existing production tables are missing from that model, generated SQL can include destructive statements you did not intend.
+Never apply ORM generated SQL blindly. A migration generator only knows the schema model it was given. If `wrangler`'s `d1_migrations` table, manually created indexes, or existing production tables are missing from that model, generated SQL can include destructive statements you did not intend.
 
 Bad migration review looks like this:
 
@@ -442,7 +442,7 @@ DROP TABLE d1_migrations;
 DROP TABLE order_events;
 ```
 
-Good migration review keeps Wrangler metadata intact and limits the migration to the intended schema change:
+`wrangler` metadata remains intact and limits the migration to the intended schema change:
 
 ```sql
 -- Good: only the intended application change remains.
@@ -451,7 +451,7 @@ CREATE INDEX IF NOT EXISTS idx_orders_archived_at ON orders(archived_at);
 PRAGMA optimize;
 ```
 
-If your ORM writes migrations in nested directories, configure `migrations_pattern` so Wrangler finds the files. For more detail, refer to [Migrations](/d1/reference/migrations/#nested-migration-layouts).
+If your ORM writes migrations in nested directories, configure `migrations_pattern` so `wrangler` finds the files. For more detail, refer to [Migrations](/d1/reference/migrations/#nested-migration-layouts).
 
 ### Defer foreign keys during migrations
 
--- a/src/content/docs/d1/best-practices/rules-of-d1.mdx
+++ b/src/content/docs/d1/best-practices/rules-of-d1.mdx
@@ -1,0 +1,329 @@
+---
+title: Rules of D1
+description: Best practices for building fast, reliable applications on D1, covering indexing, batching, retries, migrations, and scale.
+pcx_content_type: concept
+sidebar:
+  order: 2
+products:
+  - d1
+---
+
+import { TypeScriptExample } from "~/components";
+
+[D1](/d1/) is a serverless SQL database built on [SQLite](https://www.sqlite.org/). A single database runs against one primary instance and processes queries against it, so the cost of each query and the shape of your schema directly determine how your application performs and scales.
+
+This is a guidebook of rules for building fast, reliable applications on D1. Each rule explains the behavior behind it and shows the pattern to follow.
+
+## Query performance
+
+### Index every queried column
+
+A query that scans a whole table reads every row, which is slow and counts every scanned row toward your bill. An index turns a scan into a lookup. Create an index for any column you filter on in a `WHERE` clause, join on, or sort by in an `ORDER BY` clause.
+
+```sql
+-- Index a column used in predicates and joins
+CREATE INDEX IF NOT EXISTS idx_orders_customer_id ON orders(customer_id);
+```
+
+After you create or drop an index, run `PRAGMA optimize`. This runs `ANALYZE` on each table and updates the statistics the query planner uses to choose an index.
+
+```sql
+PRAGMA optimize;
+```
+
+Verify that a query uses the index you expect with `EXPLAIN QUERY PLAN`. Output containing `USING INDEX` confirms the index is used; a `SCAN` over the table means it is not.
+
+```sql
+EXPLAIN QUERY PLAN SELECT * FROM users WHERE email_address = ?;
+```
+
+For a multi-column index, a query uses the index only if it filters on all of the indexed columns, or on a left-prefix of them. An index on `(customer_id, transaction_date)` serves a query that filters on `customer_id`, or on both columns, but not one that filters on `transaction_date` alone.
+
+Tables with the default `INTEGER PRIMARY KEY` do not need a separate index for that column. Every other indexed column adds a write to the index on each `INSERT`, `UPDATE`, or `DELETE` that touches it. For read-heavy workloads, the reduction in rows read almost always outweighs this write amplification.
+
+Indexes are hard to add after the fact on a multi-gigabyte database, so define them up front where you can. For full detail on unique, partial, and multi-column indexes, refer to [Use indexes](/d1/best-practices/use-indexes/).
+
+### Batch related queries to reduce round-trips
+
+Each call to `run()` is a separate round-trip to your database. Calling `run()` in a loop multiplies network latency by the number of statements. [`batch()`](/d1/worker-api/d1-database/#batch) sends a set of statements in one round-trip, and runs them as a single transaction: if any statement fails, the entire sequence rolls back.
+
+<TypeScriptExample filename="index.ts">
+
+```ts
+const users = [{ name: "Alice" }, { name: "Bob" }, { name: "Carol" }];
+const stmt = env.DB.prepare("INSERT INTO users (name) VALUES (?)");
+
+// Bad: one round-trip per statement
+for (const user of users) {
+	await stmt.bind(user.name).run();
+}
+
+// Good: one round-trip, one transaction
+await env.DB.batch(users.map((user) => stmt.bind(user.name)));
+```
+
+</TypeScriptExample>
+
+The per-query limits apply to each statement inside a batch. For example, the 100 KB maximum statement length applies to every statement in the batch, not the batch as a whole.
+
+### Always use parameterized queries
+
+Never interpolate user input or runtime values into a SQL string. Bind them as parameters with [`bind()`](/d1/worker-api/prepared-statements/) instead. D1 follows SQLite's convention and supports ordered (`?NNN`) and anonymous (`?`) parameters.
+
+<TypeScriptExample filename="index.ts">
+
+```ts
+// Bad: string interpolation is open to SQL injection
+const bad = env.DB.prepare(`SELECT * FROM users WHERE email = '${email}'`);
+
+// Good: bind the value as a parameter
+const good = env.DB.prepare("SELECT * FROM users WHERE email = ?").bind(email);
+```
+
+</TypeScriptExample>
+
+Beyond preventing SQL injection, parameterized queries make your query metrics useful. D1 captures query strings for [query insights](/d1/observability/metrics-analytics/) but strips bound parameters, so every execution of the same query template aggregates together instead of appearing as thousands of distinct queries.
+
+## Reliability and error handling
+
+### Retry write queries from your application
+
+D1 automatically retries read-only queries up to two times on a retryable error. Only queries containing solely `SELECT`, `EXPLAIN`, or `WITH` are retried; any query that can write is never auto-retried. Your application must retry its own writes.
+
+Retry only idempotent writes, and only on errors that are safe to retry. Use exponential backoff with jitter so concurrent clients do not retry in lockstep.
+
+<TypeScriptExample filename="retry.ts">
+
+```ts
+// Retryable errors from the D1 error list
+function isRetryable(err: unknown): boolean {
+	const message = String(err);
+	return (
+		message.includes("Network connection lost") ||
+		message.includes("storage caused object to be reset") ||
+		message.includes("reset because its code was updated")
+	);
+}
+
+async function writeWithRetry(
+	stmt: D1PreparedStatement,
+	maxAttempts = 3,
+): Promise<D1Result> {
+	for (let attempt = 1; ; attempt++) {
+		try {
+			return await stmt.run();
+		} catch (err) {
+			if (attempt >= maxAttempts || !isRetryable(err)) throw err;
+			// Exponential backoff capped at 2s, with full jitter
+			const backoff = Math.min(50 * 2 ** attempt, 2000);
+			await new Promise((resolve) =>
+				setTimeout(resolve, Math.random() * backoff),
+			);
+		}
+	}
+}
+```
+
+</TypeScriptExample>
+
+For the complete set of error messages and which are safe to retry, refer to the [error list](/d1/observability/debug-d1/#error-list). Refer to [Retry queries](/d1/best-practices/retry-queries/) for more.
+
+### Chunk bulk mutations
+
+A single `UPDATE` or `DELETE` over hundreds of thousands of rows can exceed the database's CPU or memory limit and reset the instance, which surfaces as an internal error and loses the in-flight operation. Process large mutations in chunks of roughly 10,000 rows.
+
+<TypeScriptExample filename="cleanup.ts">
+
+```ts
+const CHUNK_SIZE = 10_000;
+let deleted = CHUNK_SIZE;
+
+// Delete in bounded chunks until none remain
+while (deleted >= CHUNK_SIZE) {
+	const result = await env.DB.prepare(
+		`DELETE FROM logs WHERE id IN (
+			SELECT id FROM logs WHERE created_at < ? LIMIT ?
+		)`,
+	)
+		.bind(cutoff, CHUNK_SIZE)
+		.run();
+
+	deleted = result.meta.changes ?? 0;
+}
+```
+
+</TypeScriptExample>
+
+The same applies to bulk imports. Partition import data into files of tens of megabytes rather than one large file. Refer to [Import and export data](/d1/best-practices/import-export-data/).
+
+## Schema and migrations
+
+### Use migrations for team and multi-environment workflows
+
+Migrations are not required for every schema change. If you are a solo developer with one database, running `ALTER TABLE` directly with `wrangler d1 execute` is valid and permanent.
+
+Migrations become valuable when you work on a team or have multiple environments (local dev, staging, production) that need identical schemas. A migration is a `.sql` file checked into git that records a schema change. Wrangler tracks which migrations have been applied in a `d1_migrations` table so they are never double-applied.
+
+Typical team workflow:
+
+1. Developer pulls latest code from git (which includes new `.sql` migration files from teammates).
+2. Developer runs `wrangler d1 migrations apply <DATABASE> --local` to update their local database.
+3. Developer creates a new migration with `wrangler d1 migrations create`, writes SQL, tests locally.
+4. Developer commits the migration file and pushes to git.
+5. CI/CD applies the migration to production with `wrangler d1 migrations apply <DATABASE> --remote`, then deploys the Worker.
+
+Migration files live in git, not in the database. The `d1_migrations` table is a checklist of which files have been applied in each environment. Refer to [Migrations](/d1/reference/migrations/) for the full guide.
+
+### Choose the right migration tooling
+
+You can write migration SQL by hand or use an ORM to generate it. Each approach has trade-offs.
+
+**Raw SQL with Wrangler** works well when:
+
+- You have 1-5 tables
+- Your queries are straightforward CRUD
+- You are the only developer
+- You are comfortable writing SQL
+
+**An ORM like Drizzle or Prisma** starts paying off when:
+
+- You have 10+ tables with foreign key relationships between them
+- You build complex queries with joins across 3+ tables, where type safety prevents you from accidentally joining on the wrong column or misspelling a field name
+- Multiple developers touch the schema, and the TypeScript definition becomes the single source of truth with compile errors catching mismatches immediately
+- You want autocompletion in your editor when writing queries
+- You are iterating rapidly on schema design and want to see what SQL each change produces without writing it yourself
+
+If you use an ORM, be aware that ORM migration generators cannot apply migrations to D1 directly. You generate the SQL with your ORM, then apply it with Wrangler. Review all generated SQL before applying — ORMs may generate destructive statements for tables they do not manage. Refer to the [Drizzle integration notes](/d1/reference/community-projects/#drizzle-orm) for specific gotchas.
+
+### Defer foreign keys during migrations, do not disable them
+
+D1 enforces foreign key constraints by default, equivalent to `PRAGMA foreign_keys = on` for every transaction. This is unlike plain SQLite, where enforcement is off by default. If you migrate a schema that was designed with enforcement off, existing violations surface immediately.
+
+Because D1 runs every query inside an implicit transaction, you cannot turn enforcement off with `PRAGMA foreign_keys = off`; D1 ignores it. A migration script copied from a SQLite dump that sets `foreign_keys = off` has no effect and can fail with constraint errors.
+
+To violate constraints temporarily within a migration, use `PRAGMA defer_foreign_keys = on`. This defers enforcement to the end of the current transaction instead of disabling it.
+
+```sql
+PRAGMA defer_foreign_keys = on;
+
+-- Run ALTER TABLE, data backfills, and other migration statements here.
+-- Resolve every constraint violation before the transaction ends.
+```
+
+Any violation left unresolved when the transaction commits fails with a `FOREIGN KEY constraint failed` error. For relationship definitions and the available actions (`CASCADE`, `RESTRICT`, `SET DEFAULT`, `SET NULL`, `NO ACTION`), refer to [Define foreign keys](/d1/sql-api/foreign-keys/).
+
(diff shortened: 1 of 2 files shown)


src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,329 @@
...
+
+Set a [location hint](/d1/configuration/data-location/) at creation time, matching the region closest to the workload that queries the database most.
+
+```sh
+npx wrangler d1 create my-database --location=weur
+```
+
+A location hint is a preference, not a guarantee, and write latency is always dominated by the primary's location even when read replicas exist. Refer to [Data location](/d1/configuration/data-location/) for the full list of hints and for jurisdiction constraints.
+
+### Use the Sessions API so you can add replicas later
```
irvinebroque: Getting started doesn't use the sessions API:

https://developers.cloudflare.com/d1/get-started/#write-queries-within-your-worker

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+[D1](/d1/) is Cloudflare's serverless SQL database. It is built on [SQLite](https://www.sqlite.org/), integrates directly with Workers, and gives applications a relational database without connection strings, connection pools, or database servers to manage. This is a guidebook on how to build more resilient, scalable, and correct D1 applications.
+
+## Overview
+
+D1 is the relational database in Cloudflare's developer platform. It gives Workers applications standard relational SQL, ACID transactions, structured querying, migrations, read replication, and automatic Time Travel backups without managing a database server.
+
+## When to use D1
+
+Use D1 for structured, relational data and complex SQL querying. Do not use D1 as a high-throughput key-value cache, blob store, or real-time coordination primitive.
```
rts-rob: Suggest changing the order here to match the bullets that follow or vice versa.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+Use D1 when you need:
+
+- **Relational data modeling** — Data that relies on tables, schemas, foreign keys, and `JOIN` operations, such as user profiles, ecommerce product catalogs, content management systems, and order records.
+- **Complex SQL queries** — Searching, filtering, sorting, paginating, and aggregating across multiple tables or rows.
```
rts-rob: nit: "pagination"

rts-rob: Probably also "aggregation" but that one is a question of style.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+Use alternative storage when you need:
+
+- **Workers KV for high-volume key-value reads** — Simple key-to-value lookups optimized for low-latency reads and flexible write propagation, such as session tokens, feature flags, and cached API responses.
+- **Durable Objects for real-time coordination** — In-memory state, long-lived WebSockets, or single-point serialization without SQL overhead, such as chat rooms, multiplayer lobbies, and live document editing.
+- **R2 for unstructured blob storage** — Large files, media assets, backups, and logs, such as images, PDFs, and video files.
+- **Hyperdrive for external databases** — Existing self-hosted or cloud-hosted PostgreSQL and MySQL databases that need connection pooling and query caching from Workers.
+
+D1 can serve high-traffic applications when the data model and access pattern fit. For example, a globally cached blog may rarely query D1 on the request path, so database placement matters less. If requests frequently query D1, choose a location hint close to the primary workload and use read replication for global reads.
```
rts-rob: Let's link to "location hints" here.

https://developers.cloudflare.com/d1/configuration/data-location/#provide-a-location-hint

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+| Limit                              | Workers Paid | Workers Free |
+| ---------------------------------- | ------------ | ------------ |
+| Databases per account              | 50,000       | 10           |
+| Maximum database size              | 10 GB        | 500 MB       |
+| Queries per Worker invocation      | 1,000        | 50           |
+| Maximum SQL query duration         | 30 seconds   | 30 seconds   |
+| SQL statements per query           | 100          | 100          |
+| Maximum bound parameters per query | 100          | 100          |
+
+A common failure mode for large D1 databases is an aggregation query that tries to scan an entire multi-gigabyte table into memory. Keep hot-path aggregations precomputed or bounded by indexed predicates. For the complete list, refer to [Limits](/d1/platform/limits/).
```
rts-rob: Would prefer "Precompute hot path aggregations or [use indexes](https://developers.cloudflare.com/d1/best-practices/use-indexes/) to avoid table scans."

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+  - d1
+---
+
+import { TypeScriptExample } from "~/components";
+
+[D1](/d1/) is Cloudflare's serverless SQL database. It is built on [SQLite](https://www.sqlite.org/), integrates directly with Workers, and gives applications a relational database without connection strings, connection pools, or database servers to manage. This is a guidebook on how to build more resilient, scalable, and correct D1 applications.
+
+## Overview
+
+D1 is the relational database in Cloudflare's developer platform. It gives Workers applications standard relational SQL, ACID transactions, structured querying, migrations, read replication, and automatic Time Travel backups without managing a database server.
```
alsuren: can we get away with merging these sections?

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+D1 is designed for horizontal scale-out across many smaller databases, not vertical scale-up of one large database. When a single database approaches its size or throughput limits, shard your data across multiple databases instead of growing one.
+
+A natural sharding boundary is one database per tenant: each user, organization, or project gets its own isolated database. This gives complete data isolation, per-tenant schema evolution, and a migration blast radius limited to a single tenant. You can create databases programmatically with the [REST API](/api/resources/d1/subresources/database/methods/create/) and set a location hint per tenant.
+
+Sharding has trade-offs. You cannot run SQL `JOIN` operations across D1 databases, and ACID transactions do not span database boundaries. Keep data that must be joined or updated atomically in the same database, or design explicit application-level workflows for cross-database operations.
+
+Attribute usage per database with the `rows_read` and `rows_written` fields that every query returns in its `meta` object. Apply per-query and per-tenant limits so one tenant cannot consume all of your account-level limits or inflate your bill.
```
rts-rob: This first sentence is confusing for me. I think it means "You can account for costs for each database using the `rows_read`..."?

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+Every indexed column adds at least one extra row write on `INSERT`, `UPDATE`, or `DELETE` because SQLite must update the index too. That tradeoff is usually worth it for read-heavy workloads because indexed point lookups avoid long scans. Add important indexes early; adding indexes after a database grows to multiple gigabytes is harder and slower.
+
+For more index patterns, refer to [Use indexes](/d1/best-practices/use-indexes/).
+
+### Batch related queries
+
+Do not call `run()` in a loop for related statements. Each separate call pays a round trip to D1. [`batch()`](/d1/worker-api/d1-database/#batch) sends related prepared statements in one round trip and runs them as a single transaction.
+
+Use `batch()` for multi-statement writes that should succeed or fail together, such as creating an order and its items.
```
rts-rob: Since this is a separate use case I would say "You should also use `batch()`..."

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+				"INSERT INTO order_items (id, order_id, sku, quantity) VALUES (?, ?, ?, ?)",
+			).bind(item.id, order.id, item.sku, item.quantity),
+		),
+	]);
+}
+```
+
+</TypeScriptExample>
+
+The per-query limits still apply to each statement in the batch. A batch reduces round trips; it does not make an unbounded write safe.
```
rts-rob: "...still apply *individually* to each..."

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+An ORM such as Drizzle or Prisma works well when:
+
+- You have 10 or more tables with foreign key relationships between them.
+- You build complex queries with joins across three or more tables, where type safety helps prevent joining on the wrong column or misspelling a field name.
+- Multiple developers touch the schema, and the TypeScript definition becomes the single source of truth with compile errors catching mismatches immediately.
+- You want autocompletion in your editor when writing queries.
+- You are iterating rapidly on schema design and want to see what SQL each change produces before applying it.
+
+If you use an ORM, be aware that ORM migration generators cannot apply migrations to D1 directly. Generate the SQL with your ORM, then apply it with `wrangler`. Review all generated SQL before applying it because ORMs may generate destructive statements for tables they do not manage, including Wrangler's `d1_migrations` table.
```
rts-rob: I think this last sentence deserves a little more clarity. What do I do if my ORM generates a migration for the `d1_migrations` table?

rts-rob: Now I see this comes up again later in "Review generated migrations before applyint them." In that case we should link to it here from "Review all generated SQL before applying it."

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+Set a location hint when you create the database if the workload that queries D1 is not near the place where you run the create command. Without a location hint, D1 creates the primary database instance close to the request that created the database, which may be your laptop, CI system, or control plane rather than your users.
+
+Choose the location closest to the primary write or query region. Cross-region requests add round-trip latency to every query that goes to the primary.
+
+```sh
+npx wrangler d1 create commerce-db --location=weur
+```
+
+A location hint is a preference, not a guarantee. D1 runs the database in the nearest available location to that preference. Read replicas can improve global read latency, but write latency is still dominated by the primary database location because writes go to the primary. For the full list of hints, refer to [Data location](/d1/configuration/data-location/#available-location-hints).
```
rts-rob: "A location hint is a preference, not a guarantee."

Yes, but a jurisdiction is a guarantee. If users need them they should be aware of them. We should discuss and link here.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+	return Response.json({ order, bookmark: session.getBookmark() });
+}
+```
+
+</TypeScriptExample>
+
+`withSession()` returns an object with the same query interface as the top-level database binding, so downstream query code stays familiar. If you may enable read replication later, designing new code around sessions can make that change a configuration update instead of an application rewrite. For more detail, refer to [Global read replication](/d1/best-practices/read-replication/).
+
+## Related resources
```
rts-rob, line 557: Suggest adding [Data location](https://developers.cloudflare.com/d1/configuration/data-location/)

rts-rob (changes requested):

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+A single D1 database has fixed resources. You do not choose instance sizes, CPU, or memory. The way you scale one database is by keeping queries short, indexed, bounded, and observable.
+
+The first limits to design around are:
+
+| Limit                              | Workers Paid | Workers Free |
+| ---------------------------------- | ------------ | ------------ |
+| Databases per account              | 50,000       | 10           |
+| Maximum database size              | 10 GB        | 500 MB       |
+| Queries per Worker invocation      | 1,000        | 50           |
```
FlorentCollin, line 38: This needs to be checked because the workers limits page indicates that the limit is 1000 per Worker invocation for internal services. I supposed that D1 is considered as an internal service.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+		// Bad: every user shares one hot database.
+		const overloaded = await env.DB.prepare(
+			"SELECT * FROM orders WHERE user_id = ? LIMIT 50",
+		)
+			.bind(userId)
+			.all();
+
+		// Good: route the request to the user's own D1 database.
+		const userDb = await getUserDatabase(env, userId);
```
FlorentCollin: Unfortunately, the complexity of this example is hidden behind the `getUserDatabase` function. As we don't have dynamic bindings, in this example each D1 database must be added to the bindings which isn't scalable beyond 10-20 databases. If I understood correctly, Worker for Platforms is the only way to ease the implementation in this case.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+		const { results, meta } = await userDb
+			.prepare("SELECT * FROM orders ORDER BY created_at DESC LIMIT 50")
+			.all();
+
+		return Response.json({
+			overloaded,
+			results,
+			rowsRead: meta.rows_read,
+			rowsWritten: meta.rows_written,
+		});
```
FlorentCollin: The response doesn't make a lot of sense with the bad and good examples in the same example.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+For platforms that deploy user Workers with per-customer D1 bindings, refer to [Workers for Platforms bindings](/cloudflare-for-platforms/workers-for-platforms/configuration/bindings/).
```
FlorentCollin: I think we should be more upfront about this. It is not a good experience to build the above without using Workers for Platforms.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+);
+```
+
+## Querying
+
+### Index every column you filter, join, or sort on
+
+D1 processes queries one at a time for each individual database. Your maximum throughput is directly related to query duration: if the average query takes `1 ms`, the database can process about `1,000` queries per second; if it takes `100 ms`, it can process about `10` queries per second.
+
+A table scan makes every matching request slower and reads rows you did not need. Create indexes for columns used in predicates, joins, and sort keys. For the schema above, `users.email`, `orders.user_id`, `order_items.order_id`, and `order_events.order_id` are hot-path columns.
```
FlorentCollin: I guessed the meaning behind `hot-path columns` but it doesn't really say what's important. I don't fully like my suggestion, so feel free to improve:

```suggestion
A table scan makes every matching request slower and reads rows you did not need. Create indexes for columns used in predicates, joins, and sort keys. For the schema above, `users.email`, `orders.user_id`, `order_items.order_id`, and `order_events.order_id` columns are good candidates because they are referencing foreign keys.
```

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+		// Bad: this becomes a table scan if users.email is not indexed.
+		const slowUser = await env.DB.prepare("SELECT * FROM users WHERE email = ?")
+			.bind(email)
+			.first();
+
+		// Good: create idx_users_email before this query is on the hot path.
+		const user = await env.DB.prepare("SELECT * FROM users WHERE email = ?")
+			.bind(email)
+			.first();
+
```
alsuren: Not convinced by this example. Bad and good are the same code. Could we show a query that performs badly without an index, and then the EXPLAIN output and then the index creation statement that would fix it and the fixed EXPLAIN output?

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+		// Bad: this becomes a table scan if users.email is not indexed.
+		const slowUser = await env.DB.prepare("SELECT * FROM users WHERE email = ?")
+			.bind(email)
+			.first();
+
+		// Good: create idx_users_email before this query is on the hot path.
+		const user = await env.DB.prepare("SELECT * FROM users WHERE email = ?")
+			.bind(email)
+			.first();
```
FlorentCollin: This example doesn't make sense, both the good and bad examples are the exact same query.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+		return Response.json({ user: user ?? slowUser });
+	},
+} satisfies ExportedHandler<Env>;
+```
+
+</TypeScriptExample>
+
+Every indexed column adds at least one extra row write on `INSERT`, `UPDATE`, or `DELETE` because SQLite must update the index too. That tradeoff is usually worth it for read-heavy workloads because indexed point lookups avoid long scans. Add important indexes early; adding indexes after a database grows to multiple gigabytes is harder and slower.
```
FlorentCollin: > ... harder and slower

This is vague, it would be better to explain that due to D1 being single threaded, adding an index can stall the database because it needs to create the index by doing a sequential scan of the entire table.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+<TypeScriptExample filename="index.ts">
+
+```ts
+export default {
+	async fetch(request: Request, env: Env): Promise<Response> {
+		const email = new URL(request.url).searchParams.get("email") ?? "";
+
+		// Bad: user input changes the SQL string itself.
+		const bad = await env.DB.prepare(
+			`SELECT * FROM users WHERE email = '${email}'`,
```
alsuren, line 212: Are we allowed to make a reference to https://xkcd.com/327/ here?

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+</TypeScriptExample>
+
+The per-query limits still apply to each statement in the batch. A batch reduces round trips; it does not make an unbounded write safe.
+
+### Bind every runtime value
+
+Never interpolate user input or runtime values directly into SQL strings. Use [`bind()`](/d1/worker-api/prepared-statements/) for values.
+
+This protects your application from SQL injection and makes query observability useful. D1 strips bound parameter values from query metrics, so all executions of the same query template aggregate together in [D1 Insights](/d1/observability/metrics-analytics/#query-insights). If you bake values into the SQL text, the metrics can fragment across many unique query strings.
```
FlorentCollin: With the insights project we are working on, I'm not sure that `D1 Insights` should be mentioned here

```suggestion
This protects your application from SQL injection and makes query observability useful. D1 strips bound parameter values from query metrics, so all executions of the same query template aggregate together in [D1 query insights](/d1/observability/metrics-analytics/#query-insights). If you bake values into the SQL text, the metrics can fragment across many unique query strings.
```

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+```ts
+function shouldRetryD1(error: unknown): boolean {
+	const message = String(error);
+	return (
+		message.includes("Network connection lost") ||
+		message.includes("storage caused object to be reset") ||
+		message.includes("reset because its code was updated") ||
+		message.includes("Cannot resolve D1 DB due to transient issue")
+	);
```
FlorentCollin, line 283: We need to do some work on this. With the JSRPC binding we will be able to return structured errors with properties like `.retryable`. This would make it much easier for the user to know which error is retryable.

This would be similar to the Durable Object errors https://developers.cloudflare.com/durable-objects/best-practices/error-handling/

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+Automatic retries handle retryable network-level or service-level failures for read-only queries. They do not decide whether a write is safe to repeat, and they do not resolve application-level outcomes such as unique constraint collisions. Your application owns the logical retry behavior for writes.
+
+Make writes idempotent before you retry them. For order creation, require a client-generated idempotency key and enforce it with a `UNIQUE` constraint. If a transient error occurs after D1 commits the write but before the Worker receives the response, the retry returns the existing order instead of creating a duplicate. If a retry hits the unique constraint, treat that as a signal to fetch and return the existing record rather than issuing a second logical write.
+
+Use exponential backoff with full jitter so clients do not retry in lockstep. Start around `50 ms`, cap around `2 s`, and keep the maximum attempt count small. Retry only errors marked retryable in the [D1 error list](/d1/observability/debug-d1/#error-list).
```
FlorentCollin: Is `full` jitter really our recommendation? Using full jitter makes it possible for the queries to be retried immediatly after failing.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+	const message = String(error);
+	return (
+		message.includes("Network connection lost") ||
+		message.includes("storage caused object to be reset") ||
+		message.includes("reset because its code was updated") ||
+		message.includes("Cannot resolve D1 DB due to transient issue")
+	);
+}
+
+async function writeWithRetry<T>(operation: () => Promise<T>): Promise<T> {
```
FlorentCollin: The function name could be a bit more generic in this case since the operation could be anything.

src/content/docs/d1/best-practices/rules-of-d1.mdx
alsuren: Do we have the ability to set a .retryable property on errors yet?

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+}
+```
+
+</TypeScriptExample>
+
+For more detail, refer to [Retry queries](/d1/best-practices/retry-queries/).
+
+### Chunk bulk mutations
+
+Never modify hundreds of thousands of rows in one query. A single `UPDATE` or `DELETE` over a large table can exceed CPU or memory limits, reset the underlying database object, and lose the in-flight operation.
```
FlorentCollin: Resetting the underlying DO also abort all the in-flight operations and not just one that tried to update or delete rows.
```suggestion
Never modify hundreds of thousands of rows in one query. A single `UPDATE` or `DELETE` over a large table can exceed CPU or memory limits, reset the underlying database object, and lose the in-flight operations.
```

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+	while (true) {
+		// Bad: one unbounded DELETE can scan and mutate too much at once.
+		// await env.DB.prepare("DELETE FROM order_events WHERE created_at < ?")
+		// 	.bind(cutoff)
+		// 	.run();
+
+		// Good: delete a bounded set of primary keys, then repeat.
+		const result = await env.DB.prepare(
+			`DELETE FROM order_events WHERE id IN (
+				SELECT id FROM order_events WHERE created_at < ? ORDER BY id LIMIT ?
```
FlorentCollin, line 353: Would it be worth stating that having an index on the `created_at` column is important here? Otherwise, this is going to do a full table scan.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+## Schema and migrations
+
+### Choose the right migration tooling
+
+You can write migration SQL by hand or use an object-relational mapping (ORM) tool to generate it. Each approach has trade-offs.
+
+Raw SQL with `wrangler` works well when:
+
+- You have one to five tables.
```
FlorentCollin: This seems arbitrary. Having a lot of tables doesn't make it harder to use wrangler migrations.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+
+### Capture a Time Travel bookmark before destructive migrations
+
+Time Travel is always on and has no additional cost. Before any migration that drops columns, drops tables, or performs a bulk `UPDATE` or `DELETE`, capture the current bookmark and store it with your deployment record.
+
+```sh
+npx wrangler d1 time-travel info DB
+```
+
+If the migration goes wrong, restore to the captured bookmark.
```
FlorentCollin, line 482: Precise that if the migration fails, D1 will automatically roll back the changes.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+	// Bad: after enabling replicas, an unconstrained read can reach a stale replica.
+	await env.DB.prepare(
+		"INSERT INTO orders (id, user_id, idempotency_key, status) VALUES (?, ?, ?, ?)",
+	)
+		.bind(input.id, input.userId, input.idempotencyKey, "pending")
+		.run();
+
+	return await env.DB.prepare("SELECT * FROM orders WHERE id = ?")
+		.bind(input.id)
+		.first<Order>();
```
FlorentCollin: This is partly wrong. In this case all queries will go to the primary because the example doesn't use the session API.

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,612 @@
...
+Use this for read-after-write flows, such as creating an order and then immediately fetching it.
+
+<TypeScriptExample filename="index.ts">
+
+```ts
+async function badCreateThenRead(
+	env: Env,
+	input: OrderInput,
+): Promise<Order | null> {
+	// Bad: after enabling replicas, an unconstrained read can reach a stale replica.
```
alsuren: This will cause all queries to go to the primary I think, so you don't get inconsistency, but you also don't get the benefit of read replicas at all?

Someone else please correct me.

src/content/docs/d1/best-practices/rules-of-d1.mdx
ivoryibu: replaced this example and added some WfP to highlight this ability

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -0,0 +1,568 @@
...
+
+</TypeScriptExample>
+
+For more detail, refer to [Retry queries](/d1/best-practices/retry-queries/).
+
+### Chunk bulk mutations
+
+Never modify hundreds of thousands of rows in one query. A single `UPDATE` or `DELETE` over a large table can exceed CPU or memory limits, reset the underlying database object, and lose the in flight operations.
+
+Process bulk changes in bounded chunks of about `10,000` rows. Use a stable key, commit each chunk, and continue until no rows remain. Create an index for the column that selects each chunk, such as `order_events.created_at` in the example below, so each chunk does not scan the full table.
```
simonha9, line 334: Is this number arbitrary? for example if each row holds a ton of data this might even be too large and make the isoalte exceed memory. Maybe just "Process bulk changes in bounded chunks"

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -10,320 +10,559 @@ products:
...
+
+-- Bad: the query planner must scan the table.
+-- SCAN users
+```
+
+Create the index, then run `PRAGMA optimize` so SQLite updates the statistics used by the query planner.
+
+```sql
+CREATE INDEX IF NOT EXISTS idx_users_email ON users(email);
 PRAGMA optimize;
```
alsuren, line 115: Given that a bunch of people are already getting into trouble doing migrations on worker startup, I wonder if this will tip people's LLMs into also running PRAGMA optimize on all worker startups as well (I've not checked how much that costs).

If this is best practice, it seems like the kind of thing that we should bake into `wrangler d1 migration apply`, and not mention in the docs?

src/content/docs/d1/best-practices/rules-of-d1.mdx
```diff
@@ -10,320 +10,559 @@ products:
...
+
+Only shard when you have a stable shard key and can choose a fixed number of shards up front. For example, you can bind a fixed set of databases such as `env.USERS_DB_0` through `env.USERS_DB_N_MINUS_1`, then route each request with `hash(userId) % N`. D1 does not provide automatic resharding tools. Changing the number of shards later requires an application level data migration and a careful cutover plan.
+
+Sharding has tradeoffs. You cannot run SQL `JOIN` operations across D1 databases, and ACID transactions do not span database boundaries. Keep data that must be joined or updated atomically in the same database, or design explicit application level workflows for cross database operations.
+
+Track costs for each database with the `rows_read` and `rows_written` fields that every query returns in its `meta` object. Apply per query and per tenant limits so one tenant cannot consume all of your account level limits or inflate your bill.
+
+If you are building a platform where user Workers need isolated D1 bindings, refer to [Workers for Platforms bindings](/cloudflare-for-platforms/workers-for-platforms/configuration/bindings/).
```
alsuren, line 57: * Softening the language a bit (e.g. "Scale out across many databases" sounds like an imperative thing that you *should* do).
* I didn't feel like explaining where to track per-tenant costs and apply limits because it's a bit niche and involved, so I removed it.
* I also read up on workers for platforms finally - it is just "deploy workers via an api"?

```suggestion
## Scaling out across many databases

D1 is designed for horizontal scale out across many smaller databases, not vertical scale up of one large database. Do not shard by default: sharding adds routing, migration, and operational complexity that D1 does not hide for you.

Only shard when you have a stable shard key and can choose a fixed number of shards up front. For example, you can bind a fixed set of databases such as `env.USERS_DB_0` through `env.USERS_DB_63`, then route each request with `hash(userId) % 64`. D1 does not provide automatic resharding tools. Changing the number of shards later requires an application level data migration and a careful cutover plan.

Sharding has tradeoffs. You cannot run SQL `JOIN` operations across D1 databases, and ACID transactions do not span database boundaries. Keep data that must be joined or updated atomically in the same database, or design explicit application level workflows for cross database operations.

If you need dynamic number of shards (e.g. one database per customer) then the [sqlite api in durable objects](/durable-objects/api/sqlite-storage-api/#sql-api) may be a good fit.

If you are building a platform where your users have their own Workers and each one wants its own isolated D1 database, refer to [Workers for Platforms bindings](/cloudflare-for-platforms/workers-for-platforms/configuration/bindings/).
```

PR #31826, opened 2026-07-01
Draft: Workers event notifications

yomna-shousha:
Draft reference page for Workers event notifications via Queues event subscriptions. Covers gradual deployment events, deployment/version lifecycle, routing, secrets/tags/settings, account-level events, and Workers for Platforms — with milestone 1 and milestone 2 breakdown.


yomna-shousha:
this will not be merged in

PR #31829, opened 2026-07-01
Unify binding ref in R2 docs

lkolbly:
### Summary

The docs say to create a binding named "MY_BINDING", this updates the code to match.

### Documentation checklist

### DIFF
--- a/src/content/docs/r2/api/workers/workers-api-usage.mdx
+++ b/src/content/docs/r2/api/workers/workers-api-usage.mdx
@@ -112,14 +112,14 @@ export default class extends WorkerEntrypoint<Env> {
 
     switch (request.method) {
       case "PUT": {
-        await this.env.R2.put(key, request.body, {
+        await this.env.MY_BUCKET.put(key, request.body, {
           onlyIf: request.headers,
           httpMetadata: request.headers,
         });
         return new Response(`Put ${key} successfully!`);
       }
       case "GET": {
-        const object = await this.env.R2.get(key, {
+        const object = await this.env.MY_BUCKET.get(key, {
           onlyIf: request.headers,
           range: request.headers,
         });
@@ -139,7 +139,7 @@ export default class extends WorkerEntrypoint<Env> {
         });
       }
       case "DELETE": {
-        await this.env.R2.delete(key);
+        await this.env.MY_BUCKET.delete(key);
         return new Response("Deleted!");
       }
       default:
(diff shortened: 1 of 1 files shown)


src/content/docs/r2/api/workers/workers-api-usage.mdx
```diff
@@ -162,14 +162,14 @@ export default {
 
     switch (request.method) {
       case "PUT": {
-        await this.env.R2.put(key, request.body, {
+        await this.env.MY_BUCKET.put(key, request.body, {
```
helloimalastair:
```suggestion
        await env.MY_BUCKET.put(key, request.body, {
```

src/content/docs/r2/api/workers/workers-api-usage.mdx
```diff
@@ -162,14 +162,14 @@ export default {
...
           onlyIf: request.headers,
           httpMetadata: request.headers,
         });
         return new Response(`Put ${key} successfully!`);
       }
       case "GET": {
-        const object = await this.env.R2.get(key, {
+        const object = await this.env.MY_BUCKET.get(key, {
```
helloimalastair:
```suggestion
        const object = await env.MY_BUCKET.get(key, {
```

src/content/docs/r2/api/workers/workers-api-usage.mdx
```diff
@@ -189,7 +189,7 @@ export default {
         });
       }
       case "DELETE": {
-        await this.env.R2.delete(key);
+        await this.env.MY_BUCKET.delete(key);
```
helloimalastair:
```suggestion
        await env.MY_BUCKET.delete(key);
```

helloimalastair (changes requested):
A few small changes, otherwise lgtm

src/content/docs/r2/api/workers/workers-api-usage.mdx
lkolbly: haha I made that mistake too, but this worker does it differently I guess and inherits from `WorkerEntrypoint<Env>` which I guess puts it on the instance:
```
export default class extends WorkerEntrypoint<Env> {
  async fetch(request: Request) {
```

helloimalastair: That's the TypeScript version though? In the JS version(this one), it is 
```ts
export default {
```

So I guess one solution here would be to make the JS version use `WorkerEntrypoint` too?

lkolbly: Ah yup, you're right, thanks!

helloimalastair (approved):
LGTM

PR #31843, opened 2026-07-01
fix: retry trailing-slash redirect when URL has a query string

helio-cf:
### Summary

The worker's trailing-slash retry ran `request.url.replace(/([^/])$/, "$1/")` against the full URL string. For requests with a query string, that regex appended the slash at the end of the query (e.g. `?test=whatever/`) rather than to the pathname, so a rule like `/foo/* /bar/:splat 301` was never matched and the request 404'd. 

**TLDR**: https://developers.cloudflare.com/cloudflare-one/policies/data-loss-prevention?x=y 404s but should redirect the same way as https://developers.cloudflare.com/cloudflare-one/policies/data-loss-prevention/?x=y

The retry now:

- Skips entirely when pathname already ends in `/` (avoids a redundant evaluator call on every non-redirected request).
- Appends `/` to the pathname only, so the retry actually forces a trailing slash on paths that carry a query string.
- Reuses the URL parsed at the top of `fetch` instead of allocating another one.

Adds two regression tests: one exercising a query string on a source path that already has a trailing slash (baseline for the library), and one on a source path that requires the trailing-slash retry to fire — the case the fix restores.

### DIFF
--- a/worker/index.ts
+++ b/worker/index.ts
@@ -165,16 +165,16 @@ export default class extends WorkerEntrypoint<Env> {
 				console.error("Could not evaluate redirects", error);
 			}
 
-			if (!pathname.endsWith("/")) {
+			// Markdown requests are handled by the first pass via the stripped
+			// base path, so no trailing-slash retry is needed here.
+			if (!pathname.endsWith("/") && !isMarkdownRequest) {
 				try {
 					const redirect = await redirectsEvaluator(
 						new Request(`${url.origin}${pathname}/${url.search}`, request),
 						this.env.ASSETS,
 					);
 					if (redirect) {
-						return isMarkdownRequest
-							? rewriteRedirectForMarkdown(redirect, url)
-							: redirect;
+						return redirect;
 					}
 				} catch (error) {
 					console.error(
--- a/worker/index.ts
+++ b/worker/index.ts
@@ -165,25 +165,23 @@ export default class extends WorkerEntrypoint<Env> {
 				console.error("Could not evaluate redirects", error);
 			}
 
-			try {
-				const forceTrailingSlashURL = new URL(
-					request.url.replace(/([^/])$/, "$1/"),
-					request.url,
-				);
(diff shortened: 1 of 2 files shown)


worker/index.ts
```diff
@@ -165,25 +165,23 @@ export default class extends WorkerEntrypoint<Env> {
...
-				);
-				const redirect = await redirectsEvaluator(
-					new Request(forceTrailingSlashURL, request),
-					this.env.ASSETS,
-				);
-				if (redirect) {
-					return isMarkdownRequest
-						? rewriteRedirectForMarkdown(redirect, url)
-						: redirect;
+			if (!pathname.endsWith("/")) {
```
helio-cf: pathname is not new, declared at around line 80 of this function

worker/index.ts
```diff
@@ -165,25 +165,23 @@ export default class extends WorkerEntrypoint<Env> {
...
+				try {
+					const redirect = await redirectsEvaluator(
+						new Request(`${url.origin}${pathname}/${url.search}`, request),
```
helio-cf, line 173: Image

joining these should be safe, pathname will always have a leading slash

helio-cf: image

worker/index.ts
```diff
@@ -165,25 +165,23 @@ export default class extends WorkerEntrypoint<Env> {
...
-					request.url,
-				);
-				const redirect = await redirectsEvaluator(
-					new Request(forceTrailingSlashURL, request),
-					this.env.ASSETS,
-				);
-				if (redirect) {
-					return isMarkdownRequest
-						? rewriteRedirectForMarkdown(redirect, url)
-						: redirect;
```
helio-cf, line 180: the new code does away with the regex but keeps the rest up until the ternary, i dont believe we need rewriteRedirectForMarkdown(redirect, url) at this point since it would have already been covered by the first redirect check above... or am i reading this wrong?

PR #31863, opened 2026-07-02
[Client-Side Security] Split allowlisted-domain violation troubleshooting into two causes

xmflszz:
### Summary

Splits the "I get rule violation reports for a domain I allowlisted" troubleshooting section into two sub-sections covering the two distinct causes:

- **Redirects to other domains**: the existing content, unchanged, describing redirects to domains not covered by the allow rule.
- **Ad-blocking browser extensions**: new content explaining that reports referencing user consent or behavioral tracking domains are often caused by ad-blocking extensions redirecting those requests rather than blocking them outright.

### DIFF
--- a/src/content/docs/client-side-security/troubleshooting.mdx
+++ b/src/content/docs/client-side-security/troubleshooting.mdx
@@ -51,6 +51,8 @@ You can safely ignore these warnings, since they are related to the reports that
 
 ## I get rule violation reports for a domain I allowlisted
 
+### Redirects to other domains
+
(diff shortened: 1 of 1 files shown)


xmflszz:
Fixed — reworded to correctly describe extensions blocking (often via a local redirect) requests to the tracking domain, rather than redirecting the browser toward it. The violation report still references the original requested domain, which is why it shows up as already allowlisted.

xmflsct (approved):

PR #31930, opened 2026-07-07
chore: add asamborski as CODEOWNER for Browser Isolation

asamborski:
### Summary

Adds @asamborski as a CODEOWNER for Browser Isolation docs and related partials.

- Added to `/src/content/docs/cloudflare-one/remote-browser-isolation/`
- Added new entries for two browser isolation partials that previously had no specific ownership:
  - `src/content/partials/cloudflare-one/troubleshooting/browser-isolation.mdx`
  - `src/content/partials/cloudflare-one/access/self-hosted-app/browser-isolation-non-443.mdx`


kodster28:
conflicts here, closing out and should re-open if needed

PR #32021, opened 2026-07-10
[Web Analytics/RUM] Updated beacon changelog & FAQ with latest release

ryantownsend:
The RUM beacon JavaScript injection has been update to add `type="module"` to exclude old browsers like Internet Explorer (they can't execute the script anyway, this just prevents noise in JavaScript exception trackers)


tkadlec (approved):
Ship it. 🚢

PR #32033, opened 2026-07-13
docs(client-side-security): add use case for blocking third-party tracking pixels

zeinjaber:
## What

Adds a **Use cases** section to the content security rules page with a subsection explaining how allow rules can block third-party tracking pixels and analytics tags.

## Why

Customers and support engineers repeatedly ask whether Client-Side Security can block third-party tracking pixels such as Facebook Pixel or Google Tags. The answer is yes — via an allow rule — but this is not documented anywhere on the rules page. Without guidance, customers either do not discover the capability or configure rules incorrectly.

## Files changed

- `src/content/docs/client-side-security/rules/index.mdx` — new `## Use cases` section with `### Block third-party tracking pixels` subsection (+8 lines)

## How to verify

Navigate to [/client-side-security/rules/](https://developers.cloudflare.com/client-side-security/rules/) and confirm the Use cases section appears between the Comparison table and Next steps.

Closes DEE-3703


zeinjaber:
Closing — content will be moved to the troubleshooting page instead of the rules overview.

PR #32034, opened 2026-07-13
docs(client-side-security): add troubleshooting entry for blocking third-party tracking pixels

zeinjaber:
## What

Adds a troubleshooting entry to the Client-Side Security troubleshooting page explaining how to use content security rules to block third-party tracking pixels and analytics tags.

## Why

Customers and support engineers repeatedly ask whether Client-Side Security can block third-party tracking pixels such as Facebook Pixel or Google Tags. The answer is yes — via an allow rule — but this is not documented. The troubleshooting page is the right location since customers encounter this question when investigating rule violations from third-party scripts.

## Files changed

- `src/content/docs/client-side-security/troubleshooting.mdx` — new section `## I want to block third-party tracking pixels or analytics tags` (+6 lines)

## How to verify

Navigate to [/client-side-security/troubleshooting/](https://developers.cloudflare.com/client-side-security/troubleshooting/) and confirm the new section appears after the ad-blocking extension section.

Closes DEE-3703

### DIFF
--- a/src/content/docs/client-side-security/troubleshooting.mdx
+++ b/src/content/docs/client-side-security/troubleshooting.mdx
@@ -69,6 +69,12 @@ These extensions commonly block requests to such domains, often by redirecting t
 
 Because only visitors with these extensions installed trigger this behavior, the violation reports are sparse compared to your site's overall traffic volume.
 
+## I want to block third-party tracking pixels or analytics tags
+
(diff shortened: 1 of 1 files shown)


xmflsct:
Question raised in DEE-3703

zeinjaber:
Closing — per feedback from the Page Shield team, this is a one-off compliance workaround rather than a standard use case. Not worth documenting.

PR #32069, opened 2026-07-14
[Realtime] Add FIPS 140-3 FAQ to TURN

renandincer:
Adds a FAQ item to the Cloudflare Realtime TURN page explaining that TURN supports FIPS 140-3 when encryption is used (such as TURN over TLS), and noting that TURN has no control over encryption of the data flowing under the TURN data layer.


renandincer:
Addressed the two style findings (removed filler "Note that" and switched to active voice) in 576a099.

Skipping two findings:
- **SG-4c96b64ab622** (trailing `?` in heading): every heading in this FAQ file is phrased as a question ending in `?`. Removing it from only this heading would be inconsistent with the rest of the page.
- **SG-0f4b4187b2a1** (`doesn't` contraction): the committed text already reads "does not", so there is nothing to change.

korinne (approved):

PR #32082, opened 2026-07-14
Add Cache Response Rules to Firewall role description

jayeve:
The Firewall role grants `response-cache-settings` permissions but the [roles reference page](https://developers.cloudflare.com/fundamentals/manage-members/roles/) only listed Cache Rules. This adds Cache Response Rules with a link to the existing docs page.

## Change

In the Account-scoped roles table, the Firewall role description changes from:

> Can edit WAF, IP Access rules, Zone Lockdown settings, and Cache Rules.

to:

> Can edit WAF, IP Access rules, Zone Lockdown settings, Cache Rules, and Cache Response Rules.


mbullock1986:
@ngayerie do you have approval to push this through?

ack-cf (approved):

KaydeeDee (approved):

PR #32121, opened 2026-07-16
feat: add /rebase command to cloudflare-docs-bot

mvvmm:
### Summary

Adds a `/rebase` slash command to the cloudflare-docs-bot (`.flue/`). The command is codeowner-only and posts status updates at the top of the bot's singleton comment throughout the operation.

**`/rebase`** — Rebases the PR branch against `production`. Stops with an explanation if:
- The PR targets a branch other than `production`
- The PR is from a fork (can't push)

On a clean rebase (no conflicts), posts a success status and triggers a `/full-review` automatically since the head SHA changed.

On conflict, attempts AI-assisted resolution:
- Identifies files changed on both sides since the merge base
- Fetches all three versions of each conflicting file (base, PR head, production head)
- Calls Workers AI (`kimi-k2.7-code`) with the file contents + PR intent to produce resolved versions and a confidence score
- **High confidence**: applies via Git Data API (blob → tree → commit → force-push), then triggers `/full-review`
- **Medium/low confidence**: stops and posts the reason why it couldn't resolve automatically

**Other changes:**
- `postOrUpdateComment` extracted from the two workflow files that duplicated it into `lib/code-review-render.ts` as a shared export
- `renderRebaseStatusUpdate()` added to the render lib — injects/replaces the rebase status block at the top of the bot comment, preserving existing review content below
- New `FlueRebaseWorkflow` Durable Object class added (wrangler.jsonc v9 migration)
- Commands table in the bot comment updated to list the new command

### Images
Screenshot 2026-07-17 at 5 29 40 PM


kodster28 (approved):

mvvmm:
**CR-0fc9cd37e19c** — skipping.

The design is intentional. `writePath` in the skill inputs is informational context for the agent (so it understands where its output will end up), not an output field. The agent always returns `path` (the PR-side conflict candidate path), and `applyResolution` maps that to the correct write destination via `conflictWritePathMap`. The skill already documents this: "For rename conflicts the system will map this to the correct write destination." No code is mismatched; the two fields serve different purposes.

mvvmm:
**CR-18ed11a65d12** — skipping for now.

Per-file confidence is a legitimate improvement but requires changing the output schema, `resolveConflictsWithAI`, and `applyResolution` in a non-trivial way. All-or-nothing is conservative and safe for v1: a medium-confidence file taints the whole batch, which errs on the side of caution rather than silently applying partial resolutions. Worth revisiting if we see cases where one ambiguous file repeatedly blocks otherwise clean resolutions.

mvvmm:
**CR-8513f9f82eee** — skipping.

`/rebaseWithConflicts` was intentionally removed in this PR. There is now a single `/rebase` command that always attempts AI-assisted conflict resolution — the two-command split was collapsed. The orchestrator, payload, and workflow all reflect this. The reviewer appears to be looking at a stale PR title; the description and code are already correct.

PR #30683, opened 2026-05-07
Add Spectrum troubleshooting docs for top support case drivers

stechedo:
## Summary

Adds a new Spectrum Troubleshooting page and improves existing Spectrum documentation to address the top 3 case-generating issues identified in the April 2026 Network Support-by-Request (SBR) case review (~190 cases analyzed).

**Estimated case deflection: 9-13 cases/month across these 3 issues.**

### New page
- **`/spectrum/reference/troubleshooting/`** — New troubleshooting page with structured symptom/cause/solution sections covering:
  - Health checks failing when IP Access Rules are enabled
  - DNS record conflict when creating Spectrum apps (same hostname as proxied record)
  - Protocol mismatch (origin receives HTTP instead of HTTPS)
  - Error 525 (SSL handshake failed) in Spectrum context
  - Common Spectrum event log status code patterns (444, 445, 497-499, 521, 522)

### Modified pages (5)
- **`/spectrum/reference/configuration-options/`** — Added warnings about IP Access Rules blocking health checks, Custom Rules not working with Spectrum, protocol upgrade not being supported, and double-proxy chains causing 525 errors. Expanded intro paragraph to clarify Spectrum's L4 pass-through behavior.
- **`/spectrum/reference/limitations/`** — Added new "IP access control" section and "Using the same hostname for HTTP proxy and Spectrum" section documenting the platform limitation and recommending split-hostname architecture (with Citrix Gateway use case).
- **`/spectrum/reference/logs/`** — Added note clarifying that Spectrum status codes are distinct from HTTP status codes, with cross-link to troubleshooting page.
- **`/load-balancing/additional-options/spectrum/`** — Added caution callout in the Monitors setup step warning about IP Access Rules blocking health checks.
- **`/dns/manage-dns-records/troubleshooting/records-with-same-name/`** — Added note explaining that DNS record conflicts also apply to Spectrum applications, with cross-link to Spectrum limitations.

## Motivation

These documentation gaps were identified by reviewing all April 2026 Network SBR cases. The top case-generating patterns were:
1. **IP Access Rules + Healthchecks** (~3-5 cases/month): Customers enable IP Access Rules, health checks fail silently. No docs explain the interaction.
2. **Same Hostname WAF + Spectrum** (~3-4 cases/month): Customers hit "DNS record already exists" error with no explanation that it's a hard platform limitation. Common in Citrix/VDI deployments.
[...]

PR #32137, opened 2026-07-17
[Spectrum] Document IP Access Rules blocking health checks

stechedo:
## Summary

Split out of #30683 at the reviewer's request — see [this comment](https://github.com/cloudflare/cloudflare-docs/pull/30683#issuecomment-4999486757):

> Could we just hold the IP Access Rules + health checks section? I need a bit more time to verify the mechanism, but I don't want to block everything else.

## Changes

- **`spectrum/reference/troubleshooting.mdx`** — New "Health checks fail when IP Access Rules are enabled" section (symptoms/cause/solution) covering the allowlist-model blocking Cloudflare's health check probes.
- **`spectrum/reference/configuration-options.mdx`** — Caution callout under IP Access rules warning about health check impact, plus a note that Custom Rules (WAF) do not apply to TCP/UDP Spectrum applications.
- **`load-balancing/additional-options/spectrum.mdx`** — Matching caution callout in the Monitors step of the Spectrum load balancing setup guide.

## Content area
Spectrum, Load Balancing

### DIFF
--- a/src/content/docs/load-balancing/additional-options/spectrum.mdx
+++ b/src/content/docs/load-balancing/additional-options/spectrum.mdx
@@ -54,6 +54,9 @@ The exact settings will vary depending on your use case. Refer to the following 
 8. On the **Monitors** page, define the settings presented and select **Next**.
    - Review the monitors attached to your pools.
    - If needed, you can attach an existing monitor or [create a new monitor](/load-balancing/monitors/create-monitor/#create-a-monitor).
+   :::caution[IP Access Rules and health checks]
+   If your Spectrum application has [IP Access Rules](/spectrum/reference/configuration-options/#ip-access-rules) enabled, Cloudflare's health check probes must be allowed by your rules. Otherwise, health checks will be blocked and all endpoints will be marked as unhealthy. To resolve this, add [Cloudflare IP addresses](https://www.cloudflare.com/ips/) to your IP Access Rules allowlist. Refer to [Spectrum Troubleshooting](/spectrum/reference/troubleshooting/#health-checks-fail-when-ip-access-rules-are-enabled) for details.
+   :::
 
 9. On the **Traffic Steering** page, choose an option for [Traffic steering](/load-balancing/understand-basics/traffic-steering/steering-policies/) and select **Next**.
 
--- a/src/content/docs/spectrum/reference/configuration-options.mdx
+++ b/src/content/docs/spectrum/reference/configuration-options.mdx
@@ -90,6 +90,14 @@ If [IP Access rules](/waf/tools/ip-access-rules/create/) are enabled for a Spect
 Network analytics data for Spectrum does not reflect the outcomes of IP Access rules. Instead, to verify whether traffic was allowed or blocked based on these rules, consult the Spectrum event logs.
 :::
 
+:::caution[Health checks and IP Access Rules]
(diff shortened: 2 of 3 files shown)


msimoncf:
Suggested fixes for IP Access Rules + health checks content

The current troubleshooting section says: "consider configuring health checks to use a TCP check type... and manage your access control at the origin level instead." That appears to be technically misleading. IP Access Rules on a Spectrum application evaluate every incoming TCP connection based on source IP, so switching from HTTP to TCP health checks does not bypass the block — the probe's TCP connection is still rejected before the health-check protocol is negotiated.

1. spectrum/reference/troubleshooting.mdx
Replace the current "Health checks fail when IP Access Rules are enabled" section with:
## Health checks fail when IP Access Rules are enabled

### Applies to

This issue applies when a Spectrum application is the target of a health check:

- A standalone Cloudflare [Health Check](/health-checks/) is configured against the Spectrum application, or
- A [Load Balancer monitor](/load-balancing/monitors/) uses the Spectrum application as an endpoint (the Spectrum app is the target of the Load Balancer).

This issue does **not** apply when the Spectrum application's **origin** is a Load Balancer hostname. In that architecture, the Load Balancer monitor probes the origin servers directly, not through the Spectrum application, so IP Access Rules on the Spectrum app do not affect the monitor.

### Symptoms

- Cloudflare Health Checks or Load Balancer monitors report the Spectrum endpoint as unhealthy.
- Removing IP Access Rules from the Spectrum application causes health checks to succeed.
- The Spectrum application works correctly for client traffic that is allowed by the IP Access Rules.

### Cause

When you turn on [IP Access Rules](/spectrum/reference/configuration-options/#ip-access-rules) on a Spectrum application, all incoming connections are evaluated against your rules based on source IP — including connections from Cloudflare's own health check probes. If your rules use an allowlist model (allow specific IPs, block all others), Cloudflare's health check source IPs will be blocked.

Switching the health check from HTTP to TCP does **not** bypass IP Access Rules, because the rules evaluate the incoming connection before the health-check protocol is negotiated.

### Solution

Add Cloudflare's IP ranges to your IP Access Rules allowlist. You can find the current list of Cloudflare IP addresses at [cloudflare.com/ips](https://www.cloudflare.com/ips/).

If you need to restrict client access but cannot allowlist Cloudflare IPs on the Spectrum application, move access control to your origin instead of using IP Access Rules on the Spectrum app. If IP Access Rules remain enabled without allowing Cloudflare probe IPs, health checks will continue to fail.
2. load-balancing/additional-options/spectrum.mdx
Update the caution callout in the Monitors step to clarify the architecture context:
   :::caution[IP Access Rules and health checks]
   If your Spectrum application is used as a Load Balancer endpoint and has [IP Access Rules](/spectrum/reference/configuration-options/#ip-access-rules) enabled, Cloudflare's health check probes must be allowed by your rules. Otherwise, health checks will be blocked and all endpoints will be marked as unhealthy. To resolve this, add [Cloudflare IP addresses](https://www.cloudflare.com/ips/) to your IP Access Rules allowlist. Refer to [Spectrum Troubleshooting](/spectrum/reference/troubleshooting/#health-checks-fail-when-ip-access-rules-are-enabled) for details.

   This does not apply when the Spectrum application's origin is a Load Balancer hostname. In that architecture, the monitor probes the origin directly and is not affected by IP Access Rules on the Spectrum app.
   :::
3. spectrum/reference/configuration-options.mdx
The caution callout here is fine as-is, but optionally add the same qualifier so the limitation is not overstated:
:::caution[Health checks and IP Access Rules]
When IP Access Rules are enabled with an allowlist model (allow specific IPs, block all others), health checks targeting the Spectrum application may be blocked, causing endpoints to appear unhealthy. This applies when the Spectrum application is the target of a Cloudflare Health Check or a Load Balancer monitor. To prevent this, add [Cloudflare IP addresses](https://www.cloudflare.com/ips/) to your allowlist. Refer to [Spectrum Troubleshooting](/spectrum/reference/troubleshooting/#health-checks-fail-when-ip-access-rules-are-enabled) for details.
:::
The Custom Rules note immediately below this can stay unchanged.

PR #32140, opened 2026-07-17
[Data Localization] Clarify regionalized IP binding CIDR granularity

mathew-cf:
### Summary

Clarifies the CIDR granularity rules for [Regionalized IP Bindings](https://developers.cloudflare.com/data-localization/regional-services/ip-bindings/) after field confusion about `409` conflicts when binding a BYOIP prefix to a region.

The previous guidance implied you could bind an entire prefix (for example, a whole `/24`) to a region. In practice a regional binding must be **more specific than the prefix**: the full prefix range is already in use by the prefix's own advertisement, so a create request at the prefix length returns `409 / 1108 "conflict: binding already exists for CIDR ..."`. Verified against live binding data (base `/24` binding + regional `/32` and `/31` sub-range bindings on the same prefix).

Changes:

- Rewrite **Choose which CIDR to bind** to state the binding must be a sub-range within the prefix (a `/25` down to a `/32` / `/128`), and show how to cover a whole prefix with multiple sub-ranges.
- Add a **Handle a conflict error** section explaining the two causes of the `409` (tried to bind the whole prefix; the CIDR is already bound) and how to resolve each.
- Update the create/list/update examples to use a sub-range (`203.0.113.0/25`) instead of the whole `/24`.

### Documentation checklist

### DIFF
--- a/src/content/docs/data-localization/regional-services/ip-bindings.mdx
+++ b/src/content/docs/data-localization/regional-services/ip-bindings.mdx
@@ -26,6 +26,20 @@ A prefix binding maps a CIDR (within a BYOIP prefix you own) to a [region key](#
 
 Bindings are managed through the Data Localization Suite API under `/accounts/{account_id}/dls/`.
 
+## Choose which CIDR to bind
+
+A binding covers a range of addresses within a prefix, not just a single address — you do **not** need to create one binding per IP address. The `cidr` you bind must:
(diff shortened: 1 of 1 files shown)


cferike (approved):
These changes make sense and clarify what the customer is expected to do. Thank you Mathew for making them :)

PR #32157, opened 2026-07-20
DNS: rewrite Pending Nameserver Update troubleshooting for the common case

ngayerie:
## What

Rewrite `/dns/zone-setups/troubleshooting/pending-nameservers/` around the actual most-common failure modes we see in support — parent-zone delegation mismatches, wrong Cloudflare nameserver pair, or the registrar not publishing the change. Stale DNSSEC DS records remain covered, but as one step (Step 4) rather than the article's opening frame.

## Why

The current article opens with DNSSEC as "the most common cause," but the Salesforce and Community evidence over the last 60 days is dominated by cases where the parent zone simply is not delegating to the assigned Cloudflare nameservers — the customer's registrar UI showed the right values, but the parent zone was still returning something else. Customers had no guidance in the docs on how to check what the parent zone actually publishes.

## New troubleshooting flow

1. Confirm the assigned Cloudflare nameservers in the dashboard.
2. Check what the parent zone actually publishes, with three options:
   - `dig +trace <domain> NS`
   - `nslookup -type=ns <domain>`
   - Web-based lookups via [digwebinterface.com](https://www.digwebinterface.com/) (enable the **Trace** option) and [whatsmydns.net](https://www.whatsmydns.net/)
   Followed by a result-driven action table (correct pair / wrong pair / other provider / none / mixed).
3. Verify the change was saved at the registrar — covers transfer/registrar-lock states, wrong-account edits, registrar propagation delay, and reseller setups.
4. Check for stale DNSSEC DS records (the previous article's content, kept and lightly refreshed).
5. What to attach to a support ticket if the zone is still Pending after Steps 1–4.

## Consolidation

This PR also adds three light cross-links so the rewritten article becomes the canonical troubleshooting entry point:

- **`dns/zone-setups/full-setup/troubleshooting.mdx`** — new callout at the top pointing to the pending-nameservers article.
- **`dns/zone-setups/full-setup/setup.mdx`** — extends the existing troubleshooting note in the nameserver-change step with a pointer to the pending-nameservers article.
- **`dns/zone-setups/partial-setup/setup.mdx`** — adds a note in the TXT-verification step for partial (CNAME) setups, since the same article covers the partial branch too.
- **`dns/faq.mdx`** — the existing "Where can I find my Cloudflare nameservers?" answer now links to the pending-nameservers article for parent-zone verification instead of duplicating dig examples.

## Files changed

- `src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx` (rewritten)
- `src/content/docs/dns/zone-setups/full-setup/troubleshooting.mdx`
- `src/content/docs/dns/zone-setups/full-setup/setup.mdx`
- `src/content/docs/dns/zone-setups/partial-setup/setup.mdx`
- `src/content/docs/dns/faq.mdx`

5 files changed, 137 insertions(+), 33 deletions(-).

DEE-3724


ngayerie:
### Bot review triage

**CR-50739f080263 — Foundation DNS nameserver count:** ✅ Agree, fixed.

Foundation DNS advanced nameservers assign three nameserver hostnames (one each in `foundationdns.com`, `foundationdns.net`, `foundationdns.org`) per [the docs](https://developers.cloudflare.com/dns/foundation-dns/advanced-nameservers/#nameservers-hosting-and-assignment). The initial version:
- Called it a "pair" in intro/wrap-up wording.
- Used "or" between the three TLDs in the table, implying one of three.
- Referred to "the two Cloudflare nameservers" further down.

Fixed in the follow-up commit by:
- Adding a **Number of nameservers** column to the Step 1 table (2 / 3 / 2 / varies).
- Reworking the Foundation DNS row to spell out that all three hostnames apply and all three must be set at the registrar.
- Replacing "the pair" / "the two Cloudflare nameservers" with wording that works for all setup types (full, Foundation DNS, Secondary, Custom NS).

ngayerie:
### Bot review triage (commit `10311d6`)

**CR-fed534616f69 — Misleading command explanation:** ✅ Agree, fixed.

**CR-f8c992b6d91e — Command does not query parent zone:** ✅ Agree, fixed.

Both findings are correct. `-debug` in `nslookup` toggles verbose response details; it does not force an authoritative answer from the parent zone. And `nslookup -type=ns example.com` with no server argument queries the system's resolver, which may serve a cached answer.

Fix in the follow-up commit:
- Kept `-debug` (it's still useful for seeing the authority section) but corrected the explanation to say it exposes full response details, not that it forces an authoritative response.
- Added a second `nslookup` example that targets a TLD nameserver directly (`a.gtld-servers.net` for `.com`), which is the true equivalent of `dig +trace`.
- Added a one-liner showing how to find the authoritative TLD nameservers for any extension (`dig NS com.`).

The `dig +trace` option (Option A) already covered the authoritative-parent-zone check; this rewrite makes Option B do the same for Windows / `nslookup` users.

src/content/docs/dns/zone-setups/partial-setup/setup.mdx
```diff
@@ -77,6 +77,10 @@ If you are adding a zone for the first time via API you can add it directly with
 
 <Render file="partial-setup-verification-record" product="dns" />
 
+:::note
+If your zone stays in **Pending Nameserver Update** status after adding the verification TXT record, confirm the record is served by your authoritative DNS provider (for example, with `dig TXT <YOUR_DOMAIN>` or a web-based tool such as [digwebinterface.com](https://www.digwebinterface.com/) or [whatsmydns.net](https://www.whatsmydns.net/)). For the full activation troubleshooting flow, refer to [Zone stuck in Pending Nameserver Update](/dns/zone-setups/troubleshooting/pending-nameservers/).
```
hannes-cf:
```suggestion
If your zone stays in **Pending Nameserver Update** status after adding the verification TXT record, confirm the record is served by your authoritative DNS provider (for example, with `dig TXT cloudflare-verify.<YOUR_DOMAIN>` or a web-based tool such as [digwebinterface.com](https://www.digwebinterface.com/) or [whatsmydns.net](https://www.whatsmydns.net/)). For the full activation troubleshooting flow, refer to [Zone stuck in Pending Nameserver Update](/dns/zone-setups/troubleshooting/pending-nameservers/).
```

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
-**To check for stale DS records:**
+:::note[Partial (CNAME) setup]
+If you use a [partial (CNAME) setup](/dns/zone-setups/partial-setup/), Cloudflare does not verify nameservers. Instead, it checks that the verification TXT record is present on your authoritative DNS provider. Refer to [Set up a partial zone](/dns/zone-setups/partial-setup/setup/) for details.
+:::
+
+For details on how zone status is evaluated, refer to [Zone status](/dns/zone-setups/reference/domain-status/).
+
+## 1. Confirm the assigned Cloudflare nameservers
+
+In the Cloudflare dashboard, open the domain and go to the **Overview** page. Copy the full list of nameservers Cloudflare has assigned to this zone. The number of nameservers and their hostname format depend on your setup:
```
hannes-cf, line 23:
```suggestion
In the Cloudflare dashboard, open the domain and go to the **Overview** page. Copy the full list of nameservers Cloudflare has assigned to this zone. The number of nameservers and their name format depend on your setup:
```

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
+If the parent zone does not return the correct Cloudflare nameservers, the registrar has not published your change. Common patterns:
+
+- The nameserver change was entered in the registrar UI but not saved or submitted.
+- The change was made on a different domain, on a subdomain, or in a different registrar account.
+- The domain is under a **Transfer**, **Registrar Lock**, or **Redemption** state that prevents nameserver changes. Complete or cancel the pending state first.
+- The registrar requires an additional confirmation step (email confirmation, admin approval, two-factor prompt).
+- The registrar publishes changes on a delay. Ask your registrar's support for their expected propagation window.
+- Your domain is at a reseller and the nameserver setting must be changed one level up. Refer to [Update your nameservers at your registrar](/dns/nameservers/update-nameservers/#specific-processes).
+
+If you use [Cloudflare Registrar](/registrar/), you can review and change the assigned nameservers from **Domain Registration** > select the domain > **Configuration**.
```
hannes-cf: I don't see this option for my domain I have registered through Cloudflare.

Can we actually configure this now or is this AI hallucinated?

ngayerie:
@hannes-cf thanks for the review — both suggestions applied in `c6bc045`:

- **`hostname format` → `name format`** — applied to the Step 1 intro sentence, the table column header, and the Custom nameservers row (`branded hostnames` → `branded names`) so the whole section is consistent.
- **`dig TXT cloudflare-verify.<YOUR_DOMAIN>`** — good catch, the verification record lives at the `cloudflare-verify.` subdomain, not the apex.

Note: I had to force-push to correct an earlier commit (`5df45da`) where `git add -A` in a sandbox with a stale working tree accidentally staged the deletion of a batch of unrelated `workflows/build/*` files. The force-push replaces that commit with a clean `c6bc045` that only touches the two intended files. PR file list should now show only the 5 DNS files.

ngayerie:
@hannes-cf you're right, that was a hallucination — Cloudflare Registrar doesn't let customers change the assigned nameservers from the Configuration UI. Removed the sentence entirely in `a7ca6e0`. The surrounding step already covers what to check at the registrar in general terms.

ngayerie:
### Bot review triage on rebased commit

**Correctness (accuracy warning):** ✅ Fixed.

The previous wording said "each Cloudflare zone gets its own set of nameservers, and the assignment is tied to your account." That contradicted the authoritative [Nameserver assignments](https://developers.cloudflare.com/dns/nameservers/nameserver-options/#assignment-method) page, which explains that Cloudflare favors *consistent* nameserver names across zones in the same account by default, and that different names only appear when there is a conflict (e.g., the same domain has been added to another account) or when the zone uses a different nameserver type. Reworded to reflect that.

**Style nits:** ✅ All addressed.
- `Option C - Web-based lookup` → `Option C - web-based lookup` (sentence case).
- Added a full introductory sentence with a colon before the "What to do based on the result" table.
- Split the semicolon-joined activation-check sentence into two.

**Style nits I did not apply:** two Oxford-comma findings on lines 11 and 36. Both lists already end with `, or` in the raw source — the commas are present, so I believe those were false positives.

ngayerie:
[AI] ### Bot review triage on rebased commit `9586261`

| Finding | Verdict | Action |
|---|---|---|
| **CR-ffbd5651287b** — partial-setup callout links to a page "about nameservers" | ❌ **Disagree** | The article title reflects the actual Cloudflare dashboard status name — **Pending Nameserver Update** — and per [Zone status](https://developers.cloudflare.com/dns/zone-setups/reference/domain-status/#pending), that same status name applies to full, partial, and secondary setups alike. The rewritten article also opens with an explicit partial-setup note explaining that Cloudflare checks the verification TXT record (not nameservers) for CNAME setups. Keeping the link as-is. |
| **SG-7479bf82e4f1** — `internet` → `Internet` | ✅ Agree | Fixed. |
| **SG-50329b560bc0** — passive → active voice | ✅ Agree | Fixed: `confirm the record is served by your authoritative DNS provider` → `confirm your authoritative DNS provider serves the record`. |

src/content/docs/dns/zone-setups/full-setup/troubleshooting.mdx
```diff
@@ -13,6 +13,10 @@ head:
 
 If you see unexpected results when [changing your nameservers](/dns/zone-setups/full-setup/setup/), review the following troubleshooting questions.
 
+:::note
+If your zone is still in **Pending Nameserver Update** status, refer to [Zone stuck in Pending Nameserver Update](/dns/zone-setups/troubleshooting/pending-nameservers/) for a step-by-step check of the delegation at your registrar (including `dig +trace`, `nslookup`, and web-based lookup tools).
```
svenr-cf: Do we need "(including ...)" here if that's already covered there anyway?

src/content/docs/dns/zone-setups/partial-setup/setup.mdx
```diff
@@ -77,6 +77,10 @@ If you are adding a zone for the first time via API you can add it directly with
 
 <Render file="partial-setup-verification-record" product="dns" />
 
+:::note
+If your zone stays in **Pending Nameserver Update** status after adding the verification TXT record, confirm your authoritative DNS provider serves the record (for example, with `dig TXT cloudflare-verify.<YOUR_DOMAIN>` or a web-based tool such as [digwebinterface.com](https://www.digwebinterface.com/) or [whatsmydns.net](https://www.whatsmydns.net/)). For the full activation troubleshooting flow, refer to [Zone stuck in Pending Nameserver Update](/dns/zone-setups/troubleshooting/pending-nameservers/).
```
svenr-cf, line 81: Not sure what our policy is (if any) on linking to third-party sites like this?

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
+nslookup -type=ns -debug example.com
+```
+
+By default, `nslookup` queries your system's configured resolver, which may return a cached answer. For a definitive check against the parent zone (equivalent to `dig +trace`), query a TLD nameserver directly by adding it as the last argument. For example, for a `.com` domain:
+
+```sh
+nslookup -type=ns -debug example.com a.gtld-servers.net
+```
+
+You can find the authoritative TLD nameservers for any extension with `dig NS com.` (replace `com.` with your own TLD).
```
svenr-cf: This section is mostly going to be interesting to people who don't have DiG (e.g., Windows users) so we should probably not put dig commands in here?

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
+| [Custom nameservers](/dns/nameservers/custom-nameservers/)                                                        | Varies                | Your own branded hostnames                                                                                                                                          |
+
+Whichever format applies, the exact values shown in your dashboard are the ones the parent zone must publish. Do not assume the assignment is the same as one you have used before on another domain or in another account — each Cloudflare zone gets its own set of nameservers, and the assignment is tied to your account. For details, refer to [Nameserver assignments](/dns/nameservers/nameserver-options/#assignment-method).
+
+## 2. Check what the parent zone actually publishes
+
+The registrar control panel shows what you *asked* the registrar to publish. It does not show what the parent zone (the TLD) is actually returning to the Internet. These can differ when a change was not saved, not yet propagated, applied to a different domain, or applied in a different registrar account.
+
+Use one of the following methods to query the parent zone directly.
```
svenr-cf, line 46: For A + B we should probably give some hint at these being terminal commands?

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
+The rest of this page walks through what to check, in order, and shows how to verify each item independently of your registrar's control panel.
 
-**To check for stale DS records:**
+:::note[Partial (CNAME) setup]
+If you use a [partial (CNAME) setup](/dns/zone-setups/partial-setup/), Cloudflare does not verify nameservers. Instead, it checks that the verification TXT record is present on your authoritative DNS provider. Refer to [Set up a partial zone](/dns/zone-setups/partial-setup/setup/) for details.
+:::
+
+For details on how zone status is evaluated, refer to [Zone status](/dns/zone-setups/reference/domain-status/).
+
+## 1. Confirm the assigned Cloudflare nameservers
```
svenr-cf, line 21: Typos (cloudlfare, cloudfare) are a common mistake too; should we add a note about that?

src/content/docs/dns/zone-setups/full-setup/troubleshooting.mdx
ngayerie: [AI] Good point — dropped the `(including ...)` parenthetical in `b8b4e025`. The linked article already lists the tools, no need to duplicate here.

src/content/docs/dns/zone-setups/partial-setup/setup.mdx
ngayerie: [AI] Fair — dropped the third-party links from this callout in `b8b4e025`. The callout now only shows the local `dig TXT cloudflare-verify.<YOUR_DOMAIN>` example and defers the tool selection to the linked troubleshooting article.

The main article `dns/zone-setups/troubleshooting/pending-nameservers.mdx` still references `digwebinterface.com` and `whatsmydns.net` as Option C for customers without local `dig`/`nslookup`. Happy to strip those too if that's the preferred policy — searching the repo, there are already ~15 links to `whatsmydns.net` across DNS docs (`/dns/faq/`, `/dns/troubleshooting/`, etc.), so I kept the pattern. LMK if you'd rather I remove them and I'll open a separate cleanup PR.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
+| Nameservers from a different provider.                             | The registrar has not published your change. Continue at Step 3.                                                                                                                                                                    |
+| No nameservers returned.                                           | The domain is not yet delegated. If it was just registered, wait for the parent TLD to propagate (up to 24 hours), then retest.                                                                                                     |
+| Cloudflare and other-provider nameservers together.                | Only valid if your setup uses [multi-provider DNS](/dns/nameservers/nameserver-options/#multi-provider-dns) or [Cloudflare as Secondary DNS](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/). Otherwise, remove the non-Cloudflare records at the registrar. |
```
svenr-cf: Please remove "or Cloudflare as Secondary DNS". Multi-provider is what counts, even if Secondary zones also enable that by default under most circumstances.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,141 @@
...
+4. Wait up to 24 hours for the removal to propagate through DNS caches.
 
-After the stale DS records are removed and expire from cache, your Cloudflare zone will activate automatically. You can then turn on DNSSEC in the Cloudflare dashboard if needed.
+After the stale DS records are removed and expire from cache, your Cloudflare zone will activate automatically. You can then [enable DNSSEC in Cloudflare](/dns/dnssec/) if you want to.
 
 For more information on DNSSEC configuration, refer to [Configure DNSSEC](/dns/dnssec/) and [Troubleshoot DNSSEC](/dns/dnssec/troubleshooting/).
+
+## 5. If the zone is still Pending
+
+If Steps 1-4 all check out, and the parent zone returns the correct Cloudflare nameservers, wait for Cloudflare's next activation check. Checks happen on an increasing interval. You can force one at any time from the **Overview** page or [trigger one via API](/api/resources/zones/subresources/activation_check/methods/trigger/).
```
svenr-cf: Not "at any time"; this endpoint is rate-limited and may return an error if already requested recently. In addition this check isn't immediate; it just puts the zone in a prioritized queue. We should emphasize the latter (it may still take a couple minutes to hours even after the endpoint reports success).

src/content/docs/dns/zone-setups/partial-setup/setup.mdx
ngayerie: [AI] Reverted the removal in the follow-up commit. After a quick audit, both `digwebinterface.com` and `whatsmydns.net` are already an established pattern across DNS docs — `dns/faq`, `dns/dnssec/dnssec-active-migration`, `dns/zone-setups/full-setup/setup`, `dns/zone-setups/troubleshooting/cannot-add-domain`, and a few others. `digwebinterface` is also listed on the canonical [Recommended third-party tools](https://developers.cloudflare.com/dns/reference/recommended-third-party-tools/) page. Keeping the links here for consistency with the rest of DNS docs.

Happy to open a separate PR if we want to formalize a repo-wide policy on third-party tool links — LMK.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Good point — dropped the `dig NS com.` line in `5c2193fb`. If someone is using `nslookup` because they don't have `dig`, telling them to use `dig` is contradictory. Replaced with a link to [IANA's root zone database](https://www.iana.org/domains/root/db) so readers can find the TLD nameservers for their own extension there.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Agree — both Options A and B now finish their intro paragraphs with `In a terminal, run:` (commit `5c2193fb`). Should be clearer for readers who aren't sure where the command should go.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Removed in `5c2193fb`. I dropped `or Cloudflare as Secondary DNS` from both places it appeared — the action-table row on line 91 that you flagged, and the Option A intro paragraph a bit earlier that had the same wording. Multi-provider DNS is the only one that matters for this check.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Fixed in `5c2193fb`. New wording is:

> You can request an earlier check from the **Overview** page or by triggering one via API. This endpoint is rate-limited and may return an error if you have requested a check recently. A successful request does not activate the zone immediately — it places your zone in a prioritized queue, and the check itself can still take a few minutes to a few hours to complete depending on DNS propagation.

LMK if you'd rather I use different wording for the timing range or the rate-limit language.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,147 @@
...
+After the stale DS records are removed and expire from cache, your Cloudflare zone will activate automatically. You can then [enable DNSSEC in Cloudflare](/dns/dnssec/) if you want to.
 
 For more information on DNSSEC configuration, refer to [Configure DNSSEC](/dns/dnssec/) and [Troubleshoot DNSSEC](/dns/dnssec/troubleshooting/).
+
+## 5. If the zone is still Pending
+
+If Steps 1-4 all check out, and the parent zone returns the correct Cloudflare nameservers, wait for Cloudflare's next activation check. Checks happen on an increasing interval.
+
+You can request an earlier check from the **Overview** page or by [triggering one via API](/api/resources/zones/subresources/activation_check/methods/trigger/). This endpoint is rate-limited and may return an error if you have requested a check recently. A successful request does not activate the zone immediately — it places your zone in a prioritized queue, and the check itself can still take a few minutes to a few hours to complete depending on DNS propagation.
```
matildeopbravo: Suggestion to avoid using the overloaded term "DNS Propagation" 
```suggestion
You can request an earlier check from the **Overview** page or by [triggering one via API](/api/resources/zones/subresources/activation_check/methods/trigger/). This endpoint is rate-limited and may return an error if you have requested a check recently. A successful request does not activate the zone immediately — it places your zone in a prioritized queue, and activation can take a few minutes to a few hours, depending both on when the recheck runs and on whether the nameserver change at your registrar has taken effect by then.
```

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,147 @@
...
+- The registrar publishes changes on a delay. Ask your registrar's support for their expected propagation window.
+- Your domain is at a reseller and the nameserver setting must be changed one level up. Refer to [Update your nameservers at your registrar](/dns/nameservers/update-nameservers/#specific-processes).
+
+After fixing the change at the registrar, re-run the check from Step 2.
+
+## 4. Check for stale DNSSEC DS records
+
+If Step 2 shows the correct Cloudflare nameservers at the parent zone but the zone is still Pending, check whether DNSSEC is still enabled from a previous DNS provider.
+
+DS records live at the registrar, not at the DNS provider, and they must be removed or updated when you move DNS providers. If they are not, resolvers will refuse the new nameservers.
```
matildeopbravo:
```suggestion
DS records live at the registrar, not at the DNS provider, and they must be removed or updated when you move DNS providers. If they are not, the DNSSEC chain of trust breaks and resolvers return SERVFAIL for your domain.
```

matildeopbravo (changes requested):

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,147 @@
...
+| ----------------------------------------------------------------------------------------------------------------- | --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Standard [full setup](/dns/zone-setups/full-setup/)                                                               | 2                     | `NAME.ns.cloudflare.com`                                                                                                                                            |
+| [Foundation DNS](/dns/foundation-dns/) with [advanced nameservers](/dns/foundation-dns/advanced-nameservers/#nameservers-hosting-and-assignment) | 3                     | One nameserver in each of `NAME.foundationdns.com`, `NAME.foundationdns.net`, and `NAME.foundationdns.org` — all three must be set at the registrar. |
```
matildeopbravo: Should probably specify that it's a color. In the nameserver docs we have `<color>.foundationdns.com`

matildeopbravo (changes requested):

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Good catch. Applied in the follow-up commit. Foundation DNS row now uses `<color>` to match `dns/nameservers/index.mdx` line 45. I also aligned the standard-nameserver rows with the same file's line 35 convention — `NAME` → `<proper_name>` for both `.ns.cloudflare.com` and `.secondary.cloudflare.com`.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Applied — replaced my vague `resolvers will refuse the new nameservers` with your concrete `the DNSSEC chain of trust breaks and resolvers return SERVFAIL for your domain`. Much clearer failure mode.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Applied verbatim. Good point on avoiding "DNS propagation" — the new phrasing splits the two independent factors (when the recheck runs vs. whether the registrar change has taken effect) explicitly.

matildeopbravo (commented):
_(superseded)_

matildeopbravo (commented):
_(superseded)_

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,147 @@
...
+description: Troubleshoot a Cloudflare zone that stays in Pending Nameserver Update status, including how to verify the delegation at your registrar and check for stale DNSSEC DS records.
 products:
   - dns
 ---
 
-If your nameservers are correctly set to Cloudflare but your zone remains in **Pending Nameserver Update** status, stale DNSSEC DS records at your registrar are the most common cause.
+A zone stays in **Pending Nameserver Update** when Cloudflare cannot confirm that your domain is delegated to the Cloudflare nameservers assigned to it.
 
-## Stale DNSSEC DS records
+The most common reasons are that the nameserver change was not fully published at the registrar, that the domain is not using the exact nameserver pair assigned to it, or that stale DNSSEC records at the registrar are blocking the delegation.
```
matildeopbravo: Foundation DNS is 3, so `pair` doesn't apply. Same fix in Option C.

```suggestion
The most common reasons are that the nameserver change was not fully published at the registrar, that the domain is not using the exact nameservers assigned to it, or that stale DNSSEC records at the registrar are blocking the delegation.
```

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,147 @@
...
+
+### Option A - `dig +trace`
+
+`dig +trace` follows the delegation from the root zone down, so its final output shows the nameservers that the parent zone is actually publishing for your domain. In a terminal, run:
+
+```sh
+dig +trace example.com NS
+```
+
+The last section of the output should return **only** the Cloudflare nameservers assigned to your zone (or, if you use [multi-provider DNS](/dns/nameservers/nameserver-options/#multi-provider-dns), it should include them alongside your other provider's nameservers).
```
matildeopbravo: `+trace`'s last block is the child's answer, not the parent's. Querying a TLD directly is better and matches the nslookup equivalent.

```suggestion
### Option A - `dig`

Query one of your TLD's nameservers directly to see what the parent zone publishes for your domain. For a `.com` domain:

```sh
dig NS example.com @a.gtld-servers.net
```

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
```diff
@@ -1,32 +1,147 @@
...
+| [Cloudflare as Secondary DNS](/dns/zone-setups/zone-transfers/cloudflare-as-secondary/)                            | 2                     | `<proper_name>.secondary.cloudflare.com`                                                                                                                                     |
+| [Custom nameservers](/dns/nameservers/custom-nameservers/)                                                        | Varies                | Your own branded names                                                                                                                                          |
+
+Whichever format applies, the exact values shown in your dashboard are the ones the parent zone must publish. Do not assume the assignment is the same as one you have used before on another domain or in another account — each Cloudflare zone gets its own set of nameservers, and the assignment is tied to your account. For details, refer to [Nameserver assignments](/dns/nameservers/nameserver-options/#assignment-method).
```
matildeopbravo:
```suggestion
Whichever format applies, the exact values shown in your dashboard are the ones the parent zone must publish. Do not assume the assignment is the same as one you have used before on another domain or in another account — each Cloudflare zone gets its own set of nameservers. For details, refer to [Nameserver assignments](/dns/nameservers/nameserver-options/#assignment-method).
```

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: I'm concerned that this is more complicated for people to find the "TLD's nameservers".. Not everyone will know where to search.. I feel that it's easier in most cases to use trace to see if there's anything wrong with the configuration

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Good catch — the intro sentence said `pair`, which doesn't fit Foundation DNS's 3-nameserver assignment. Applied your suggestion in `655d238b`.

On `Same fix in Option C` — Option C already uses `nameservers` (plural). The `pair`-adjacent wording lives in Options A and B (`only the Cloudflare nameservers assigned to your zone` and `nameservers other than the ones assigned to your Cloudflare zone`), both already generic. LMK if you were pointing at something else specifically.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Applied verbatim in `655d238b` — the domain-hijacking-prevention rationale is useful context I didn't know. Thanks for the explanation.

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
ngayerie: [AI] Update — found the Option C `pair` you flagged. Fixed in the follow-up commit: `The result must match the nameservers assigned in your Cloudflare dashboard.`

src/content/docs/dns/zone-setups/troubleshooting/pending-nameservers.mdx
matildeopbravo: Yeah, my reasoning was that you already do the same for `nslookup` below.
How about:
`dig +trace example.com NS +noall +authority +nodnssec` ?

ngayerie: [AI] Went with your `dig +trace example.com NS +noall +authority +nodnssec` suggestion in the follow-up commit. Also:

- Added a per-flag explanation (`+trace`, `+noall +authority`, `+nodnssec`) so a reader who is not familiar with `dig` understands what each option does.
- Added a caution callout under the command explaining that `+nodnssec` hides `DS` records, and pointing the reader to Step 4 to re-check without the flag (or with `dig DS example.com`) to catch stale DS records from a previous DNS provider.
- Small nit: changed `last section` to `last non-empty section` since `+noall +authority` can leave the final `dig` block empty for the target zone.

LMK if the per-flag descriptions read wrong to you.

matildeopbravo (approved):

PR #32178, opened 2026-07-20
ci: skip docs CI for flue-only PRs, add Flue CI workflow

mvvmm:
### Summary

Adds a `paths-ignore: [".flue/**"]` filter to `ci.yml` and a new `flue-ci.yml` workflow so that PRs touching only `.flue/` skip the full docs build (Astro, link validation, deploy preview) and run a lightweight Flue-specific check instead.

**`ci.yml`** — adds `paths-ignore: [".flue/**"]` to the trigger. The workflow is not started at all when every changed file is under `.flue/`.

**`flue-ci.yml`** — new workflow triggered on `paths: [".flue/**"]` with three jobs that mirror the required status check names:

| Job | What it does |
|---|---|
| Pre Build | `pnpm run typecheck` in `.flue/` |
| Build | `pnpm run build` in `.flue/` — compiles the Worker via `flue build` |
| Post Build | No-op placeholder for future unit tests / evals |

Since branch protection matches required checks by job name, a flue-only PR gets "Pre Build", "Build", and "Post Build" checks from `flue-ci.yml` instead of `ci.yml`, so merging is not blocked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     branches:
       - production
+    paths-ignore:
+      - ".flue/**"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/flue-ci.yml
+++ b/.github/workflows/flue-ci.yml
@@ -1,0 +1,72 @@
+name: Flue CI
+
+on:
+  pull_request:
+    branches:
+      - production
+    paths:
+      - ".flue/**"
+
+concurrency:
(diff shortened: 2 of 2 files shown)


mvvmm:
**CR-5e78317fda17** — this is addressed by the new `flue-ci.yml` workflow in this same PR. It has jobs named `Pre Build`, `Build`, and `Post Build` that run when `.flue/**` files change, satisfying any required status checks by name. The bot reviewed the two files independently rather than as a pair.

**CR-bcde13235f0f** — the `post-build` job is an intentional required-check stub, not a dead placeholder. Without it, the `Post Build` required status check is never reported for flue-only PRs, blocking merges. It will be replaced with real checks (unit tests, evals) when those are added.

kodster28 (approved):

PR #32179, opened 2026-07-20
[Fundamentals] Remove Workers references from Domain Admin RBAC docs

AdamBouhmad:
### Summary
Removes incorrect Workers references from the Domain Administrator and Domain Administrator Read Only role descriptions in the RBAC documentation


jhutchings1 (approved):
:shipit:

PR #32226, opened 2026-07-21
chore: migrate cloudflare-docs-bot to Flue 2.0

mvvmm:
### Summary

Migrates the `.flue/` PR-review bot (`cloudflare-docs-flue`) from Flue 0.11 to Flue 2.0.

Flue 2.0 replaces the 0.11 "workflow-per-Durable-Object + internal HTTP routing" model with **Cloudflare `WorkflowEntrypoint`s that drive per-agent Durable Objects**. Trusted TypeScript owns all control flow and every GitHub/R2 side effect; each AI step is a Flue agent that only reasons and returns structured data through a single Valibot-typed `submit_*` tool.

Behavior is preserved — the same code review / conventions / style-guide sections in one PR comment, the spam gate, the Dependabot path, the `/rebase` command, codeowner slash commands, and `log`/`comment` review modes.

Key changes:

- **Ingress** — one HMAC-verified `POST /webhooks/github`; pure, unit-tested `classifyWebhook` + `startReviewPipeline`. No internal HTTP routes or internal-auth.
- **Workflows** — `ReviewOrchestrator`, `IngestWorkflow` (spam gate), `DependabotReviewWorkflow`, `RebaseWorkflow`.
- **Agents (7)** — `code-review-file`, `style-guide-file`, `conventions-reviewer`, `reconcile-reviewer`, `spam-filter`, `dependabot-reviewer`, `rebase-conflict-resolver`; per-file fan-out is one Durable Object instance per changed file.
- **Build** — now `vite build` / `vite dev` (`@flue/cli` 2.0 dropped `build`/`dev`); `wrangler.jsonc` gains a `v10` DO migration that creates the 7 SQLite agent classes.
- **Roles** — re-homed via an explicit `useBotRole()` hook, since 2.0 has no role auto-discovery.
- Removes the 0.11 `workflows/` and superseded `lib/` modules.

Validated locally: `tsc` (0 errors), `vite build`, `wrangler deploy --dry-run`, and the Vitest suite all pass; the full INGEST spam-gate → review pipeline was exercised end to end in log mode. Not yet deployed — the first deploy applies the `v10` Durable Object migration.


kodster28 (approved):

mvvmm:
### Review Finding Skip Notes

**CR-9f4663d992a8 — Parallel Workflow steps:** Skipping. The Cloudflare Workflows docs explicitly state that "any JS control-flow primitive (if conditions, loops, `try...catch` blocks, promises, and more) can be used to manage steps inside the `run` method." `Promise.all` with uniquely-named `step.do` calls is a supported pattern — each step is identified by name and replayed by name, so concurrent creation does not affect determinism. The architecture intentionally designed these three specialists as concurrent steps.

**CR-0f9bb79578e3 — Possible unawaited async data writer:** Skipping. `useDataWriter` returns `(data) => void` per the Flue runtime type definitions (`writeMessageData(name: string, data: unknown): void`). The framework internally awaits the write chain before the response settles (per the source: "Ordered append chain for `useDataWriter` writes; awaited before the response settles"). There is no Promise to await.

mvvmm:
Skipping the following review findings:

**CR-0bd493aca627 — Force ref update race condition**
The code already checks `currentPr.head.sha !== pr.head.sha` at line 672 and throws if the branch moved. The residual TOCTOU window (between `getPullRequest` and `updateRef`) is inherent to the GitHub git refs API — it does not accept an `expected-current-SHA` parameter. Serializing `/rebase` invocations would require a concurrency-limiting mechanism (DO alarm/semaphore) for marginal benefit on a manual command that is rarely invoked concurrently. The existing check + throw is adequate.

**CR-66103a07cfeb — File size cap uses character count instead of bytes**
`FILE_CONTENT_MAX_BYTES = 32768` (32KB) is a rough guard, not a precision boundary. The files reviewed are MDX docs where multi-byte characters are rare. Worst case (all 4-byte chars) would be ~128KB, still well within model context limits. Renaming the constant to `FILE_CONTENT_MAX_CHARS` is cosmetic churn for no behavioral change. The real protection is the model context window, not this cap.

**CR-fd5562ccd03c — Non-constant-time secret comparison**
This is a dev-only internal endpoint (`/dev/review/:number`) behind `DOCS_FLUE_INTERNAL_TOKEN`, not a production auth path. The webhook HMAC verification (the real auth surface) already uses `timingSafeEqual`. Timing attacks on a dev convenience endpoint that is not deployed in production is disproportionate effort for minimal benefit.

PR #32232, opened 2026-07-22
[Cache] Clarify DYNAMIC vs BYPASS cache statuses

ngayerie:
## What / Why

Customers regularly confuse `CF-Cache-Status: DYNAMIC` and `CF-Cache-Status: BYPASS` — they look similar (both go to origin), but they signal fundamentally different stages of Cloudflare's cache decision:

- **DYNAMIC** — asset is determined ineligible for cache **at request time** (before contacting the origin)
- **BYPASS** — asset **is** eligible at request time, but the origin response makes it uncacheable **at response time**

The current [cache responses page](/cache/concepts/cache-responses/) does not make this distinction, and both wordings miss the case that trips customers up most often: a Cache Rule with `cache: true` on default-ineligible content (for example, HTML) will return `BYPASS` rather than `DYNAMIC` when the origin sends `no-store`/`no-cache`/etc.

This PR rewrites both sections to:

1. State the request-time vs response-time framing explicitly
2. Enumerate the concrete causes of each status
3. Cross-link the two sections so customers reading either can find the other case

## Context
ESCALATION-5657

### DIFF
--- a/src/content/docs/cache/concepts/cache-responses.mdx
+++ b/src/content/docs/cache/concepts/cache-responses.mdx
@@ -50,7 +50,17 @@ The resource was served from Cloudflare's cache but was expired. Cloudflare coul
 
 ## BYPASS
 
-The origin web server instructed Cloudflare to bypass cache via a `Cache-Control` header set to `no-cache`, `private`, or `max-age=0` even though Cloudflare originally preferred to cache the asset. BYPASS is returned when enabling [Origin Cache-Control](/cache/concepts/cache-control/). Cloudflare also sets BYPASS when your origin web server sends cookies in the response header. If the Request to your origin web server includes an `Authorization` header, in some cases the response will also be BYPASS. Refer to [Conditions](/cache/concepts/cache-control/#conditions) in the Origin Cache-Control behavior section for more details.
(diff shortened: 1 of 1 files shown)


ngayerie:
[AI] Bot review triage:

| ID | Verdict | Notes |
|---|---|---|
| CR-3ebab5cd0b19 | Agree | Narrowing the note to the `Cache-Control` directive case (the primary confusion reported). `Set-Cookie` under `Eligible for cache` is documented as MISS in `cache-behavior.mdx`, so the broader claim is overreaching. |
| CR-91ef6c1de053 | Agree | Dropping `max-age=0` from the BYPASS-causing list. `cache-control.mdx` documents that with Origin Cache Control enabled, `max-age=0` caches and revalidates rather than bypassing. |
| SG-6202f5c80816 | Agree | Removing `below`. |
| SG-9558dcc5a5d4 | Skip | The Oxford comma is already present in the source (`` `no-cache`, `no-store`, `private`, or `max-age=0` ``) — appears to be a false positive on backtick-wrapped list items. Also becomes moot after the CR-91ef6c1de053 fix shortens the list to three items. |
| SG-6a5c1e0bd07c | Agree | Changing `see` to `refer to`. |

Pushing the fixes next.

ack-cf (approved):

PR #32235, opened 2026-07-22
chore: require final newlines

svenr-cf:
### Summary

Omitting the final line terminator is not only incorrect but also causes unnecessary noise, warnings, and diffs.

https://stackoverflow.com/questions/729692

### Documentation checklist

n/a


kerolasa:
Looks good to me.

hannes-cf (approved):

jhutchings1 (approved):

PR #32258, opened 2026-07-22
chore: consolidate astro-config.ts into root astro.config.ts

mvvmm:
### Summary

Consolidates `src/astro-config.ts` into the root `astro.config.ts`. Previously `astro.config.ts` was a thin shell that imported `markdown`/`integrations`/`vite` exports from `src/astro-config.ts` — the only consumer of that file. This merges everything into a single config file at the root, which is where Astro expects it and where anyone reading the repo would look first.

No behavior change: contents are moved as-is, with relative import paths (`./src/plugins/satteri`, `./sitemap.serializer`, `./src/util/robots`, `./src/components.ts`) updated to resolve correctly from the root. Verified with `pnpm run check`, `pnpm run lint`, `pnpm run format:core:check`, and `pnpm run sync` (content sync + Nimbus integration load succeed).

Also updates the two places that referenced the old file path: `AGENTS.md`'s directory tree and a doc link in `style-guide/how-we-docs/links.mdx`.

### Documentation checklist


ahaywood (approved):

mvvmm:
Skipping CR-9c5dc3f64dd8 (Incorrect path normalization regex).

The finding claims line 74 is `.replace(/\\//, "")`, stripping the first path separator. The actual code is:

```ts
const here = (p: string) =>
	fileURLToPath(new URL(p, import.meta.url)).replace(/\/$/, "");
```

`/\/$/` is anchored to the end of the string, so it strips a trailing slash only — which is exactly the behavior the finding recommends as the fix. This is unchanged from the original `src/astro-config.ts` (just relocated by this PR), and `componentsBarrelId` resolves and matches correctly (verified via `pnpm run check`/`pnpm run sync`). No change needed.

PR #32267, opened 2026-07-23
fix: Include Render partial headings in page TOC

mvvmm:
### Summary

Consumes the pkg.pr.new Nimbus build from [cloudflare/nimbus#36](https://github.com/cloudflare/nimbus/pull/36), which adds `partialHeadings` support to `getDocsPageProps`.

Adds the cloudflare-docs-specific `resolvePartialId` resolver to `src/pages/[...slug].astro`, matching `Render.astro`'s `product ? `${product}/${file}` : file` convention. This lets Nimbus collect markdown headings from the same partials that render in-page via `<Render>`.

Adds `src/util/rendered-toc.ts`, a utility that detects when a page (or its transitive `<Render>` partials) uses `AnchorHeading` — a component that emits headings at runtime via `set:html`, making them invisible to compile-time heading extraction. When runtime headings are detected, the utility renders the page through an `AstroContainer` to scrape the actual `<h*>` elements from the rendered HTML. It also transitively resolves partial-to-partial `<Render>` references to determine which partials are "dynamic" (contain `AnchorHeading` directly or through a dependency chain).

Fixes literal markdown headings inside rendered partials missing from the page TOC. For example, `/bots/plans/pro/` now includes "Bot settings versus custom rules" (from `partials/bots/bot-settings-vs-custom-rules.mdx`) in its on-this-page TOC.

Does not fix `AnchorHeading`-generated headings; those are runtime component output and need a separate cloudflare-docs-specific follow-up if desired.

### Documentation checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,10 +157,5 @@
 			"version": ">=24.0.0",
 			"onFail": "error"
 		}
-	},
-	"pnpm": {
-		"overrides": {
-			"@cloudflare/nimbus-docs": "https://registry.npmjs.org/@cloudflare/nimbus-docs/-/nimbus-docs-0.7.1.tgz"
-		}
 	}
 }
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
 		"@astrojs/rss": "4.0.18",
 		"@astrojs/sitemap": "3.7.3",
 		"@base-ui/react": "1.5.0",
-		"@cloudflare/nimbus-docs": "^0.6.1",
+		"@cloudflare/nimbus-docs": "https://pkg.pr.new/@cloudflare/nimbus-docs@36",
 		"@cloudflare/vitest-pool-workers": "0.16.15",
 		"@cloudflare/workers-types": "4.20260615.1",
 		"@docsearch/css": "3.9.0",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
 		"@astrojs/rss": "4.0.18",
 		"@astrojs/sitemap": "3.7.3",
 		"@base-ui/react": "1.5.0",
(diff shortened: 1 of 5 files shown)


mvvmm:
## Triage: findings skipped

### CR-5d3e0f3f11de & CR-343fe494b089 — Race condition in lazy singleton init

**Skipping.** `??=` evaluates the RHS and assigns the result synchronously in the same tick. Both `computeDynamicPartials()` and `loadRenderers().then()` return Promises immediately (async functions / Promise-returning APIs). The assignment to the module-level variable happens before any `await` yields control. A second caller reaching the same line will see the already-assigned Promise, not `undefined`. There is no interleaving window in single-threaded JS — the nullish coalescing assignment is atomic with respect to the event loop.

### CR-c5a92a1bb422 — Direct tarball bypasses internal registry gateway

**Skipping.** The tarball URL is intentional and documented in the comment above the override. The internal registry gateway does not mirror `@cloudflare/nimbus-docs`, so a scope-level `.npmrc` rule (`@cloudflare:registry=...`) cannot be narrowed to a single package — it would route *all* `@cloudflare/*` packages to the public registry, breaking the gateway for every other Cloudflare dependency. The per-package tarball override is the only mechanism pnpm offers to bypass the scope rule for one package while leaving it intact for the rest.

mvvmm:
## Update: override removed

CR-0481180a08c6 and CR-c5a92a1bb422 are now moot — the entire `overrides` block for `@cloudflare/nimbus-docs` has been removed. The version in `package.json` is now the single source of truth, and the package resolves through the normal registry path like every other dependency.

ahaywood (approved):

PR #32275, opened 2026-07-23
Add CI Job indexing pages after build to an AI Search instance

ttmx:
### Summary

Adds a CI Job that sends the changed contents of pages to a worker which indexes them to an AI Search instance.

The job also stores a state cache in CI, in order to only send diffs of what has changed between runs.

The indexer script looks at the generated html pages from the build step and correlates them to the original markdown, indexing the markdown pages split by the headers.
In the case of generated pages without a markdown source, we add content transformers, which generate the content to be indexed from the same config files.
In the case there is no matching transformer or markdown document available, we index a transformation of the html (mostly stripping what isn't content and converting things to plaintext)

## IMPORTANT
Requires new secrets in the repository: (To be discussed internally)
- CF_ACCESS_CLIENT_ID: Access service-token client ID.
- CF_ACCESS_CLIENT_SECRET: Access service-token secret.

- AI_SEARCH_REINDEX_URL: Reindexing endpoint.

### Documentation checklist

- [ ] If a larger change - such as adding a new page- an issue has been opened in relation to any incorrect or out of date information that this PR fixes. -> Talked about internally, no issue
- [x] Files which have changed name or location have been allocated [redirects](https://developers.cloudflare.com/pages/configuration/redirects/#per-file). -> No file changed.


kodster28 (changes requested):
Can you remind me again, what are the benefits of doing this approach as opposed to a sitemap-based crawler (and using the native indexing pipeline)?

This is a _lot_ of surface to maintain and it feels like we get most of this "for free" if we were running native AI search indexing.

Am I missing something major here?

PR #32277, opened 2026-07-23
Restore current API schema pin

kennyj42:
## Summary

Restore the API schema pin to commit 29f8cda983dbd30c426fcc9d2767b2548409ac46. A later docs update reverted this value to an older schema commit.

## Testing


kodster28 (approved):

kodster28:
this has since been superseded, I think. re-open if still needed

PR #32279, opened 2026-07-23
Document per-server Gateway routing for MCP portals

kennyj42:
## Summary

- document how to route one MCP server through Cloudflare Gateway
- retain the existing portal-wide setup path
- clarify that portal-wide routing takes precedence over the server-level setting
- identify the secure_web_gateway API field and its default

## Testing

- checked heading structure and existing internal links


kennyj42:
Replaced by #33325, using the same commit from a branch in the direct repository.

PR #32280, opened 2026-07-23
Document Workers and Pages domains for MCP portals

kennyj42:
## Summary

- document workers.dev and pages.dev as supported MCP portal domains
- clarify that an active zone is no longer required when using either managed domain
- update the portal creation steps to match the dashboard domain selector

## Testing

- verified the workers.dev documentation link


kennyj42:
Replaced by #33324, using the same commit from a branch in the direct repository.

PR #32281, opened 2026-07-23
Document approval for infrastructure applications

kennyj42:
## Summary

- add infrastructure applications to the temporary authentication guide
- link purpose justification and approval setup from the infrastructure application workflow
- clarify that both controls are configured on an Allow policy

## Testing

- verified both internal policy links


kennyj42:
Replaced by #33323, using the same commit from a branch in the direct repository.

PR #32282, opened 2026-07-23
Document Access JIT request logs API

kennyj42:
## Summary

- explain that temporary authentication requests use a separate JIT log stream
- document account-wide filtering and pagination
- add examples for listing requests and retrieving chronological lifecycle events

## Testing

- used CURL instead of APIRequest because these endpoints are new
- checked the component import and endpoint fields against the current Access OpenAPI schema


kennyj42:
Replaced by #33322, using the same commit from a branch in the direct repository.

PR #32283, opened 2026-07-23
Document custom redirect schemes for managed OAuth

kennyj42:
## Summary

- remove the stale statement that managed OAuth redirect URIs must use HTTPS
- document private-use URI schemes for native clients such as Cursor
- clarify wildcard and exact-match rules for web and custom-scheme redirects

## Testing

- checked the rules against MCP-256 and the current Access validation behavior


kennyj42:
Replaced by #33321, using the same commit from a branch in the direct repository.

PR #32284, opened 2026-07-23
Document 90-day Cloudflare One Client session limit

kennyj42:
## Summary

Document the new 90-day maximum for the organization-level session duration used by Authenticate with Cloudflare One Client.

## Testing

- checked the value against AUTH-8922


kennyj42:
Replaced by #33320, using the same commit from a branch in the direct repository.

PR #32286, opened 2026-07-23
[Cloudflare One] Add SSH client routing troubleshooting

kennyj42:
## Summary

- add a client routing check to the Access for Infrastructure SSH troubleshooting flow
- show administrators how to verify the active device profile and Split Tunnel configuration
- call out identity selector mismatches and missing Access authentication logs as routing signals

## Validation

- git diff --check
- verified the Split Tunnels link target exists
- full Astro validation was not run because the installed Node.js 22 runtime does not meet the repository's Node.js 24 requirement


kennyj42:
@asamborski

kennyj42:
Replaced by #33319, using the same commit from a branch in the direct repository.

PR #32287, opened 2026-07-23
[Cloudflare One] Clarify Access mTLS plan availability

kennyj42:
## Summary

- document that Access mTLS requires a Zero Trust contract plan
- direct non-contract customers to service tokens for automated authentication
- distinguish Access Service Auth mTLS from zone-level mTLS, which has different plan availability

## Validation

- git diff --check
- verified both internal link targets exist
- full Astro validation was not run because the installed Node.js 22 runtime does not meet the repository's Node.js 24 requirement


kennyj42:
Replaced by #33318, using the same commit from a branch in the direct repository.

PR #32292, opened 2026-07-23
Test incremental static builds with astro + nimbus-docs previews

matthewp:
### Summary

This is testing a new incremental build feature in Astro and Nimbus. The goal is to see how if build times are improved with cached results.

**Do no merge**, this is only for testing the feature. If successful a new PR will come later after the feature is released.


matthewp:
This was just a test to show that incremental could bring gains to the docs. Closing this, will open a new PR when the feature is ready.

PR #32423, opened 2026-07-30
[Cache] Add troubleshooting page for uncached responses

ngayerie:
Adds `cache/troubleshooting/investigating-uncached-responses.mdx`, a header-driven troubleshooting page for URLs served from the origin when a `HIT` was expected. Uses `cf-cache-status` to branch into three sections:

- `DYNAMIC` — request not eligible for cache.
- `BYPASS` — origin response is not cacheable.
- Repeated `MISS` — cacheable but not in cache (cache key variance, eviction, low-traffic assets).

Also aligns `cache/concepts/cache-responses.mdx` with the [May 2026 BYPASS-status changelog](https://developers.cloudflare.com/changelog/post/2026-05-26-bypass-status-for-uncacheable-responses/):

- Tighten the `MISS` definition to reflect its eligible-only scope post-changelog.
- Add missing `BYPASS` causes the changelog highlighted: response exceeds cacheable size limit; `Vary: *`.
- Link `BYPASS` and `DYNAMIC` sections to the new troubleshooting page.

DEE-3742


zaidoon1:
note that since https://developers.cloudflare.com/changelog/post/2026-05-26-bypass-status-for-uncacheable-responses/ we no longer returned repeated MISS as that was a source of confusion. if we choose not to cache something, we will return BYPASS

ack-cf (approved):

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -26,7 +26,9 @@ The resource was found in Cloudflare's cache.
 
 ## MISS
 
-The resource was not found in Cloudflare's cache and was served from the origin web server.
+The response is eligible for cache but was not present in Cloudflare's cache at request time, so it was served from the origin web server. Responses that Cloudflare chooses not to cache return [`BYPASS`](#bypass) instead of `MISS`.
+
+If a URL keeps returning `MISS` when you expect a `HIT`, refer to [Investigate uncached responses](/cache/troubleshooting/investigating-uncached-responses/#repeated-miss--cacheable-but-not-in-cache) for the common causes.
```
zaidoon1: `MISS` -> `BYPASS`

ngayerie: [AI] Thanks — that's a real gap in the page as framed. The May 2026 changelog you flagged means most of the original "why MISS?" causes actually surface as `BYPASS` today. Two rounds of changes on this branch:

**Commit `f425926`** — broaden the page beyond MISS-only:

- Renamed to `investigating-uncached-responses.mdx` with a header-driven split: `DYNAMIC` (not eligible), `BYPASS` (origin not cacheable — size limit, `Cache-Control`, `Set-Cookie`, `Vary: *`, `Authorization`), and Repeated `MISS` (cache key variance, eviction). Development Mode and recent purges moved to a `Before you start` section.
- Aligned `cache-responses.mdx` with the same changelog: tightened the `MISS` definition to its eligible-only scope, added the two BYPASS causes the changelog highlighted but weren't listed (size limit and `Vary: *`).

**Commit `ec66b22`** — this round of feedback:

- Moved the troubleshooting pointer from the MISS section to the BYPASS section in `cache-responses.mdx`, per your `MISS -> BYPASS` comment. Post-changelog, BYPASS is where readers who "expected caching" land.
- Corrected the `Set-Cookie` behavior in the troubleshooting page. Per the [cache-behavior.mdx matrix](/cache/concepts/cache-behavior/#interaction-of-set-cookie-response-header-with-cache), `Set-Cookie` only produces `BYPASS` when Origin Cache Control is enabled — the two configurations that return `MISS` every time (Cache Rule *Eligible for cache* without Edge TTL, and legacy `Cache Level: Cache Everything` Page Rule) now sit under Repeated MISS with a fix.
- Tightened the HIT verification checklist (Cloudflare doesn't preserve `Set-Cookie` on cached responses).
- Removed the `Since May 2026` dated prose from the callout per the style guide; changelog link stays.

ngayerie:
Hi @zaidoon1 
Are you ok with the last updates on the PR here?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+- **The request method is not `GET` or `HEAD`.** Cloudflare only caches these two methods.
+- **[Development Mode](/cache/reference/development-mode/) is enabled on the zone.** Development Mode suspends cache for three hours and returns `DYNAMIC` for every response.
+
+Once the request is eligible, subsequent responses reflect the response-time decision (`HIT`, `MISS`, `BYPASS`, and so on).
+
+## BYPASS — origin response is not cacheable
+
+The request was eligible for cache, but the origin response or configuration prevented Cloudflare from storing it. Common causes:
+
+- **The response exceeds the [cacheable size limit](/cache/concepts/default-cache-behavior/#cacheable-size-limits) for your plan.** Serve large downloads or video from [R2](/r2/) with cache in front, or move to a plan with a higher limit.
```
zaidoon1: Moving an object to R2 does not raise the CDN cacheable file-size limit, so R2 with cache in front does not fix this `BYPASS` for an over-limit object. Could we recommend splitting the object or using a limit that supports its size, and describe R2 only as an origin or egress alternative?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+- **The origin returned `Cache-Control: no-store`, `no-cache`, or `private`, and [Origin Cache Control](/cache/concepts/cache-control/) is enabled.** Refer to [Understand `no-store` and `no-cache` directives](/cache/concepts/cache-control/#understand-no-store-and-no-cache-directives) and the [Conditions](/cache/concepts/cache-control/#conditions) table for the exact behavior. To cache the response anyway, configure a [Cache Rule](/cache/how-to/cache-rules/) with an explicit Edge TTL, or use [Cache Response Rules](/cache/how-to/cache-response-rules/) to rewrite the `Cache-Control` directive before cache.
```
zaidoon1: This reverses the `no-cache` behavior. With Origin Cache Control enabled, bare `no-cache` is stored and always revalidated. With it disabled, `no-cache`, `max-age=0`, and `s-maxage=0` are not cached. `no-store` is not cached in either mode. Could we split these directives by mode rather than grouping them under Origin Cache Control enabled? The same correction is needed in `cache-responses.mdx`.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+- **The origin returned a `Set-Cookie` header and [Origin Cache Control](/cache/concepts/cache-control/) is enabled.** Under the default cache level with Origin Cache Control disabled, Cloudflare strips `Set-Cookie` and caches the asset. When Origin Cache Control is enabled, Cloudflare preserves `Set-Cookie` and returns `BYPASS`. Refer to [Interaction of `Set-Cookie` response header with Cache](/cache/concepts/cache-behavior/#interaction-of-set-cookie-response-header-with-cache) for the full matrix — including two additional `Set-Cookie` configurations that return `MISS` instead (covered in [Repeated MISS](#repeated-miss--cacheable-but-not-in-cache) below).
+- **The origin returned `Vary: *`.** This value always bypasses cache, regardless of other configuration.
+- **The request included an `Authorization` header** and Cloudflare treated the response as private.
```
zaidoon1: `Authorization` does not by itself prevent caching when Origin Cache Control is disabled. When it is enabled, the response is cacheable only if `Cache-Control` includes `public`, `s-maxage`, or `must-revalidate`. Could we state these branches instead of saying the response was treated as private?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+Refer to [BYPASS](/cache/concepts/cache-responses/#bypass) for the reference definition of this status.
+
+## Repeated MISS — cacheable but not in cache
+
+`MISS` on the first request in each data center is expected — that request populates the cache. If the same URL keeps returning `MISS` across consecutive requests from the same client, one of the following is happening.
+
+### Cache key variance
+
+Cloudflare builds the cache key from the request scheme, host, path, and — depending on configuration — the query string, cookies, headers, and device type. Repeated `MISS` almost always means the key is different on every request. Common patterns:
```
zaidoon1: `Repeated MISS almost always means the key is different` is too strong and conflicts with the eviction section below. Could we describe cache-key variance as one common cause rather than the default diagnosis?

zaidoon1 (commented):
Here is some issues I found

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+- **Query parameters that change per request** — session IDs, timestamps, or marketing tags such as `utm_*`. By default, every unique query string is a separate cache entry. To ignore parameters that do not affect the response, configure [Cache Rules](/cache/how-to/cache-rules/) or [Cache Key Settings](/cache/how-to/cache-keys/#cache-key-settings) to sort or drop them.
```
zaidoon1: Sorting only canonicalizes parameter order; it does not ignore values that change per request. For session IDs, timestamps, or marketing parameters, the fix is to exclude those parameters or ignore the query string when safe. Could we reserve sorting for requests that differ only in parameter order?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+- **A [custom cache key](/cache/how-to/cache-keys/) includes a cookie or header with a value that changes per user.** The [Create custom cache keys](/cache/how-to/cache-keys/#create-custom-cache-keys) section warns that custom keys "may reduce your cache hit rate and result in cache sharding" — this is the same behavior.
+- **[Cache by device type](/cache/how-to/cache-rules/examples/cache-device-type/) classifies clients differently than expected**, particularly for bots or clients with unusual `User-Agent` values.
+- **[Vary in Cache Rules](/cache/how-to/cache-rules/settings/#vary) is configured for a high-cardinality header** — for example, an unnormalized `Accept-Language` value or a per-user header. Each variant is a separate cache entry.
```
zaidoon1: Vary configuration creates variants only when the origin response lists that header in `Vary`, and only `normalize` or `passthrough` select cached variants; `bypass` prevents storage. Could we add those conditions so this does not imply every configured Vary action creates separate cache entries?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+To confirm the key is the cause, send two identical requests and inspect the responses. If both return `MISS`, compare the query string, cookies, and any headers your Cache Rule or [Cache Reserve](/cache/advanced-configuration/cache-reserve/) configuration uses.
```
zaidoon1: Two identical `MISS` responses do not confirm cache-key variance. If the requests are identical and reach the same cache location, they compute the same key; the second `MISS` only shows that it did not find a stored response. Cache Reserve also does not add cache-key headers. Could we recommend checking the applied Cache Key and Vary settings in Trace and remove Cache Reserve from this comparison?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### `Set-Cookie` with a caching rule but no Edge TTL
+
+If the origin returns `Set-Cookie` and the response is marked cacheable — by a [Cache Rule](/cache/how-to/cache-rules/) with **Eligible for cache** set to *Yes*, or by a legacy `Cache Level: Cache Everything` [Page Rule](/rules/page-rules/) — Cloudflare preserves `Set-Cookie` but does not cache the asset. Every request returns `MISS`. Refer to [Interaction of `Set-Cookie` response header with Cache](/cache/concepts/cache-behavior/#interaction-of-set-cookie-response-header-with-cache).
```
zaidoon1: This is now `BYPASS`, not repeated `MISS`. When `Set-Cookie` prevents an otherwise eligible response from being stored, the response-time rejection returns `BYPASS`. The May 2026 changelog above moved uncacheable responses to `BYPASS`; the linked Set-Cookie matrix still carries the previous status. Could we remove this subsection from Repeated MISS and update line 48 accordingly?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+To cache the response, set an explicit Edge TTL on the Cache Rule using **Edge TTL → Ignore cache-control header and use this TTL** or **Status code TTL**. With an explicit Edge TTL, Cloudflare strips `Set-Cookie` and caches the asset.
+
+### Eviction and low-traffic assets
+
+Low-traffic assets can be evicted from the cache before the next request arrives. If two consecutive requests to the same URL from the same data center both return `MISS`, enable [Tiered Cache](/cache/how-to/tiered-cache/) or [Cache Reserve](/cache/advanced-configuration/cache-reserve/) to retain long-tail content longer.
+
+If you are testing from many different networks, each request reaches a different Cloudflare data center and produces its own `MISS`. Repeat the test from the same client and network to confirm the second request returns `HIT`.
```
zaidoon1: Client network and data center are not one-to-one: different networks can reach the same location, and the same network can be routed differently. Could we tell readers to compare the data center code in the `cf-ray` suffix instead of inferring location from the network?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## Confirm the response reaches cache
+
+After adjusting configuration, request the URL twice from the same client. On the second request, verify all of the following:
+
+- `cf-cache-status: HIT`.
+- An `Age` header appears and increases on subsequent requests.
```
zaidoon1: These all-of checks reject valid cached behavior. With Origin Cache Control enabled, `no-cache`, `max-age=0`, and `s-maxage=0` can be stored but revalidated on every request, returning `REVALIDATED` or `EXPIRED` instead of `HIT`. `Age` can also be absent on the first lower-tier `HIT`. Could we scope this checklist to a fresh positive-TTL response and add a revalidation case?

ngayerie:
[AI] Thanks for the detailed pass — all 10 comments applied across the troubleshooting page and `cache-responses.mdx`.

**Applied on the troubleshooting page:**

1. BYPASS size-limit bullet — R2 removed as a size-limit workaround; noted R2 is an origin alternative that does not raise the CDN cacheable size limit.
2. BYPASS Cache-Control — split by Origin Cache Control mode: `no-store`/`private` prevent caching in either mode; `no-cache`/`max-age=0`/`s-maxage=0` only produce BYPASS with Origin Cache Control disabled (default on Free/Pro/Business), with the enabled-mode outcome (`REVALIDATED`/`EXPIRED`) called out inline.
3. BYPASS Authorization — qualified to Origin Cache Control enabled, with the `public`/`s-maxage`/`must-revalidate` exception; clarified the default mode does not prevent caching.
4. Repeated MISS intro — softened to "if each real client request produces a different key" instead of "almost always", so it reads consistently with the eviction subsection.
5. Query-string bullet — distinguished excluding/ignoring (for changing values like session IDs or `utm_*`) from sorting (only for parameter-order differences).
6. Vary bullet — describes the origin-side `Vary` requirement plus per-action behavior (`normalize`/`passthrough` select variants, `bypass` prevents caching).
7. Confirmation — reworked to note that two identical requests share a key and recommend Rule Trace to inspect applied cache key and Vary; removed the Cache Reserve reference since it does not add cache-key headers.
8. Set-Cookie no-Edge-TTL — moved from Repeated MISS into the BYPASS Set-Cookie bullet with a fix path (set an explicit Edge TTL on the Cache Rule). Also flagged that the linked `cache-behavior.mdx` matrix still labels some cases `MISS` pre-changelog — happy to open a follow-up for that page.
9. Networks vs. data centers — replaced with `cf-ray` suffix comparison; noted different networks can reach the same data center.
10. HIT checklist — scoped to fresh positive-TTL; added the `REVALIDATED`/`EXPIRED` case for Origin Cache Control + `no-cache`; noted `Age` can be absent on the first lower-tier HIT (`CacheTieredFill=true`).

**Also on `cache-responses.mdx`** (same corrections needed, since the BYPASS bullets mirrored the outdated matrix): Cache-Control split by mode, Set-Cookie now points at the matrix for the specific configurations, and Authorization qualified to Origin Cache Control enabled + directive gate.

Two commits (`5a10c39` then `b7a2d51`); CI is re-running.

ngayerie:
[AI] Thanks @zaidoon1 — all 10 comments applied across the troubleshooting page, with the same corrections mirrored to `cache-responses.mdx` BYPASS bullets where they applied (Cache-Control, Set-Cookie, Authorization). Branch rebased on `production`. Summary:

1. R2 no longer presented as a size-limit workaround — clarified it is an origin alternative that does not raise the CDN cacheable size limit.
2. Cache-Control split by [Origin Cache Control](https://developers.cloudflare.com/cache/concepts/cache-control/) mode: `no-store`/`private` in either mode; `no-cache`/`max-age=0`/`s-maxage=0` only produce BYPASS with Origin Cache Control disabled (default on Free/Pro/Business). With Origin Cache Control enabled they produce `REVALIDATED`/`EXPIRED`.
3. Authorization qualified to Origin Cache Control enabled, with the `public`/`s-maxage`/`must-revalidate` exception; noted the default mode does not by itself prevent caching.
4. Repeated MISS intro softened to "if each real client request produces a different key" — no longer conflicts with the eviction subsection.
5. Query-string bullet distinguishes excluding/ignoring (for changing values like session IDs, `utm_*`) from sorting (only for parameter-order differences).
6. Vary bullet now describes the origin-side `Vary` requirement plus per-action behavior (`normalize`/`passthrough` select variants, `bypass` prevents caching).
7. Confirmation step reworked: two identical requests share a key and cannot reveal variance. Recommend Rule Trace to inspect the applied cache key and Vary action. Removed the Cache Reserve reference — it does not add cache-key headers.
8. Set-Cookie no-Edge-TTL moved from Repeated MISS into the BYPASS Set-Cookie bullet, with a fix path (set an explicit Edge TTL on the Cache Rule). Flagging that the linked [cache-behavior.mdx `Set-Cookie` matrix](https://developers.cloudflare.com/cache/concepts/cache-behavior/#interaction-of-set-cookie-response-header-with-cache) still labels some cases `MISS` pre-changelog — happy to open a follow-up for that page.
9. Networks vs. data centers replaced with `cf-ray` suffix comparison; noted different networks can reach the same data center.
10. HIT checklist scoped to fresh positive-TTL responses; added the `REVALIDATED`/`EXPIRED` case for Origin Cache Control + `no-cache`; noted `Age` can be absent on the first lower-tier HIT (`CacheTieredFill=true`).

ngayerie:
Hi @zaidoon1 
I've updated the PR following your comments, thanks!
Would it be possible for you to review again please?

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+- **The response exceeds the [cacheable size limit](/cache/concepts/default-cache-behavior/#cacheable-size-limits) for your plan.** Split the object into smaller assets, or move to a plan with a higher limit. [R2](/r2/) is an origin storage alternative — it does not raise the CDN cacheable size limit.
+- **The origin returned `Cache-Control: no-store` or `private`.** These directives prevent caching in either [Origin Cache Control](/cache/concepts/cache-control/) mode.
+- **The origin returned `Cache-Control: no-cache`, `max-age=0`, or `s-maxage=0`, and [Origin Cache Control](/cache/concepts/cache-control/) is disabled** (the default for Free, Pro, and Business plans). With Origin Cache Control enabled, these directives cause Cloudflare to cache and revalidate the response instead, producing [`REVALIDATED`](/cache/concepts/cache-responses/#revalidated) or [`EXPIRED`](/cache/concepts/cache-responses/#expired). Refer to [Understand `no-store` and `no-cache` directives](/cache/concepts/cache-control/#understand-no-store-and-no-cache-directives) and the [Conditions](/cache/concepts/cache-control/#conditions) table.
```
zaidoon1: The plan defaults are backwards here. Origin Cache Control is on by default for Free, Pro and Business, and off by default for Enterprise.

So the "Origin Cache Control is disabled" branch isn't the common case, it's the Enterprise one. Same fix needed on the `Authorization` bullet below, and on the matching bullets in `cache-responses.mdx`.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,87 @@
...
+- **The origin returned a `Set-Cookie` header** under a configuration that prevents caching. `Set-Cookie` returns `BYPASS` when [Origin Cache Control](/cache/concepts/cache-control/) is enabled, when a [Cache Rule](/cache/how-to/cache-rules/) marks the response **Eligible for cache** without an explicit Edge TTL, or when a legacy `Cache Level: Cache Everything` [Page Rule](/rules/page-rules/) matches without an explicit Edge TTL. Under the default cache level with Origin Cache Control disabled, Cloudflare strips `Set-Cookie` and caches the asset. Refer to [Interaction of `Set-Cookie` response header with Cache](/cache/concepts/cache-behavior/#interaction-of-set-cookie-response-header-with-cache) for the full matrix. To cache the response, set an explicit Edge TTL on the Cache Rule using **Edge TTL → Ignore cache-control header and use this TTL** or **Status code TTL**.
```
zaidoon1: The three configurations listed aren't right.

With Origin Cache Control on, `Set-Cookie` stops the response being cached at any cache level, including the default one on a plain static file. As written this reads like the default cache level is safe under OCC, and it isn't.

The Cache Everything without an Edge TTL case is the Origin Cache Control **off** rule, not an OCC-on one. That's the only configuration where `Set-Cookie` on its own blocks caching with OCC off.

Two ways a `Set-Cookie` response does get cached that aren't covered: `Cache-Control: private="Set-Cookie"` or `no-cache="Set-Cookie"`, where the header is dropped and the rest is cached, and a Cache Rule that strips `Set-Cookie`, since the header is already gone by the time the cache decision happens.

Probably easier to reframe as "a `Set-Cookie` response is not cached unless something explicitly allows it" and then list the ways to allow it, rather than enumerating the configurations that block it.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,87 @@
...
+- **[Development Mode](/cache/reference/development-mode/) is enabled on the zone.** Development Mode suspends cache for three hours and returns `DYNAMIC` for every response.
+
+Once the request is eligible, subsequent responses reflect the response-time decision (`HIT`, `MISS`, `BYPASS`, and so on).
+
+## BYPASS — origin response is not cacheable
+
+The request was eligible for cache, but the origin response or configuration prevented Cloudflare from storing it. Common causes:
+
+- **The response exceeds the [cacheable size limit](/cache/concepts/default-cache-behavior/#cacheable-size-limits) for your plan.** Split the object into smaller assets, or move to a plan with a higher limit. [R2](/r2/) is an origin storage alternative — it does not raise the CDN cacheable size limit.
+- **The origin returned `Cache-Control: no-store` or `private`.** These directives prevent caching in either [Origin Cache Control](/cache/concepts/cache-control/) mode.
```
zaidoon1: Two exceptions worth calling out, since this reads as absolute.

`private` with field names, like `private="Set-Cookie"`, is still cacheable with Origin Cache Control on. Only the headers it names get dropped. Bare `private` is the form that blocks caching in both modes.

And an Edge TTL that ignores origin cache-control overrides both directives, so `no-store` plus an Edge TTL does get cached. You already give that caveat on the `Set-Cookie` bullet, it applies here too.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+If your requests reach different Cloudflare data centers, each produces its own first-request `MISS`. Compare the data center code — the last three characters of the `cf-ray` header — to confirm two responses came from the same data center. Different client networks can still reach the same data center, so a network change does not guarantee a different location.
+
+## Confirm the response reaches cache
+
+After adjusting configuration, request the URL twice from the same client and verify the expected outcome for your configuration:
+
+- **Fresh, positive Edge TTL:** `cf-cache-status: HIT` and an `Age` header that increases on subsequent requests. `Age` may be absent on the first lower-tier request that populates the local cache from a tiered cache fill (`CacheTieredFill=true`).
+- **Origin returns `Cache-Control: no-cache` or `must-revalidate` with [Origin Cache Control](/cache/concepts/cache-control/) enabled:** `cf-cache-status: REVALIDATED` when the origin confirms the cached copy is unchanged, or [`EXPIRED`](/cache/concepts/cache-responses/#expired) when the origin returns new content. Both indicate the response is cached.
```
zaidoon1: `must-revalidate` on its own doesn't behave this way. With no `max-age` or `s-maxage` there's no freshness lifetime to expire, so the asset picks up the normal TTL and you get `HIT` while it's fresh. All `must-revalidate` changes is that we won't serve it stale once it does expire.

`no-cache` is the one that gives you `REVALIDATED` or `EXPIRED` on every request. Worth dropping `must-revalidate` from this bullet, or splitting it out with its own description.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,87 @@
...
+- **Any cached response:** no `Set-Cookie` header on the response.
```
zaidoon1: This doesn't hold up as a check. On the `MISS` that fills the cache we still pass the origin's `Set-Cookie` downstream, it only gets stripped from the stored copy. And a `BYPASS` forwards it untouched, so someone testing a URL that's bypassing because of `Set-Cookie` will see the header and have no way to tell that apart from the failure this bullet is meant to catch.

Suggest scoping it to responses actually served from cache (`HIT`, `REVALIDATED`, `STALE`, `UPDATING`), or dropping it.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+Refer to [BYPASS](/cache/concepts/cache-responses/#bypass) for the reference definition of this status.
+
+## Repeated MISS — cacheable but not in cache
+
+`MISS` on the first request in each data center is expected — that request populates the cache. If the same URL keeps returning `MISS` across consecutive requests, one of the following is happening.
+
+### Cache key variance
+
+Cloudflare builds the cache key from the request scheme, host, path, and — depending on configuration — the query string, cookies, headers, and device type. If each real client request produces a different key, the cache never sees a repeat and every request is a `MISS`. Common patterns:
```
zaidoon1: Two problems in this sentence.

The query string isn't configuration-dependent, it's part of the key by default. The bullet directly below says exactly that ("By default, every unique query string is a separate cache entry"), so the two contradict each other. The bullet is the correct one.

And "scheme" here is how Cloudflare connects to your origin, not the scheme the client used. A zone with a single origin scheme serves HTTP and HTTPS requests from the same cache entry. Worth stating, because people do assume the client scheme splits the cache.

zaidoon1 (commented):
Went through this against how caching actually behaves today. Most of it holds up. The Origin Cache Control plan defaults are backwards though, and that knocks on to a couple of the other bullets.

ngayerie:
[AI] Thanks @zaidoon1 — all 6 comments applied in `e6f3b49`, with the OCC-default corrections mirrored to `cache-responses.mdx` where the same bullets appeared.

1. `no-store`/`private` bullet — added the two exceptions: `Cache-Control: private="<header>"` field-form remains cacheable (only the named headers are dropped), and a [Cache Rule](https://developers.cloudflare.com/cache/how-to/cache-rules/) with an Edge TTL that ignores origin cache-control overrides both directives.
2. Origin Cache Control defaults corrected everywhere they appeared — enabled by default on Free/Pro/Business, disabled by default on Enterprise. Applies to the `no-cache`/`max-age=0`/`s-maxage=0` bullet, the `Authorization` bullet, and their mirrors in `cache-responses.mdx`.
3. `Set-Cookie` bullet — reframed as "not cached unless something explicitly allows it". With OCC enabled (FPB default) `Set-Cookie` blocks caching at every cache level. `Cache Level: Cache Everything` without an Edge TTL is now correctly attributed to the OCC-off path. Added the two allowances that were missing: `Cache-Control: private="Set-Cookie"`/`no-cache="Set-Cookie"` from origin, and stripping `Set-Cookie` with a [Response Header Modification Transform Rule](https://developers.cloudflare.com/rules/transform/response-header-modification/) before the cache decision.
4. Cache-key intro — query string is now stated as part of the default key rather than "depending on configuration". Added the origin-scheme clarification: the scheme in the key is the scheme Cloudflare uses to reach the origin, so a zone with a single origin scheme serves HTTP and HTTPS clients from the same cache entry.
5. HIT confirmation `no-cache`/`must-revalidate` bullet — dropped `must-revalidate`. `no-cache` is the directive that produces `REVALIDATED`/`EXPIRED` per request; `must-revalidate` only prevents stale content after the freshness TTL expires.
6. HIT confirmation Set-Cookie bullet — dropped. Since `Set-Cookie` is forwarded on `MISS` and `BYPASS` too, its absence isn't a reliable HIT indicator.

CI is re-running.

drcaramelsyrup (approved):

zaidoon1 (approved):

PR #32437, opened 2026-07-30
[CF One] Rename "WARP ingress IP" heading to "Cloudflare One Client ingress IP"

cf-rhett:
### Summary

The **WARP ingress IP** heading on the Cloudflare One Client firewall page reads as though the section covers the consumer WARP client, but it documents the ingress IPs and UDP ports used by the Cloudflare One Client. Renamed it to **Cloudflare One Client ingress IP**, which also matches the phrasing already used on the [switch organizations](https://developers.cloudflare.com/cloudflare-one/team-and-resources/devices/cloudflare-one-client/deployment/mdm-deployment/switch-organizations/) page.

The `#warp-ingress-ip` anchor is deliberately preserved via an explicit heading ID (`## Cloudflare One Client ingress IP {/* warp-ingress-ip */}`). This is the section customers are pointed at to build firewall allowlists, so it is widely deep-linked from support material and internal runbooks, and URL fragments cannot be redirected. All eight inbound links keep working untouched.

Link text was updated in the same pass. On the firewall page itself the surrounding sentence already names the client, so those read just "ingress IPs" rather than repeating the product name; off-page references (a table cell in the diagnostic logs page, a troubleshooting step, and the Cloudflare WAN partial) use the full name since they have no surrounding context.

### Documentation checklist

- [x] Files which have changed name or location have been allocated [redirects](https://developers.cloudflare.com/pages/configuration/redirects/#per-file). No redirect needed — no page moved, and the heading anchor was kept stable rather than renamed.


jiulingz (approved):

cf-rhett:
Skipping **SG-170db579f45f**. The link at `diagnostic-logs.mdx:120` already includes the required trailing slash before the fragment: `/deployment/firewall/#warp-ingress-ip`. The proposed replacement is identical to the current link, so there is no code change to make.

PR #32517, opened 2026-08-04
[Cloudflare for SaaS] Correct DCV renewal method guidance

cferike:
### Summary

Clarifies how Cloudflare selects DCV for Custom Hostname certificate issuance and renewal:

- The selected method applies to initial issuance.
- Eligible non-wildcard hostnames may renew through automatic HTTP validation even when TXT was selected initially.
- Wildcard hostnames and hostnames using Delegated DCV remain TXT-based.
- Manual validation should use the current `ssl.method` and `ssl.validation_records`.

Also corrects PATCH examples to omit `ssl` for metadata-only changes and preserve the complete current `ssl.settings` object when changing TLS settings or certificate authorities. This avoids unintended setting resets or validation-method changes.

### Testing

- `pnpm run format:check`
- `pnpm run check`
- `pnpm run build`

### Documentation checklist


baubuchon-cf (approved):
Verified this against current production behavior before approving. Cross-checked the renewal-eligibility claims against the shipped renewal logic and the RCA for SFDC-02317381 (Sept 8) — current production behavior matches this PR: eligible non-wildcard TXT hostnames transition to HTTP on renewal; wildcard and Delegated DCV hostnames stay on TXT.
Approving.

PR #32543, opened 2026-08-05
AI Gateway: clarify /compat is required for Dynamic Routing despite deprecation notice

kai-cloudflare:
### Summary

Documentation fix. The AI Gateway docs currently contradict themselves on how to call a dynamic route, and the contradiction actively sends readers down a path that cannot work.

[Unified API (OpenAI compat)](https://developers.cloudflare.com/ai-gateway/usage/chat-completion/) opens with a blanket `Deprecated` callout telling readers to use the REST API instead. But dynamic routes are invoked *only* through that same `/compat/chat/completions` endpoint — every example on [Using a dynamic route](https://developers.cloudflare.com/ai-gateway/features/dynamic-routing/usage/) hits `/compat`, and `src/content/docs/ai-gateway/usage/rest-api.mdx` contains zero references to dynamic routing. So for anyone using Dynamic Routing, "use the REST API instead" is not just unhelpful, it is wrong, and there is nothing on the REST API page to correct the impression.

This came out of a real customer engagement (Amazon Bedrock on dynamic routes) where the deprecation notice led to time lost building against an endpoint that does not support the feature.

Changes:

1. **`ai-gateway/usage/chat-completion.mdx`** — scope the deprecation to standard single-model chat completions, and add a `note` stating that dynamic routes (`dynamic/{route}`) are invoked through this endpoint and that the REST API does not currently cover dynamic routing. Links to the dynamic routing usage page for examples.
2. **`ai-gateway/features/dynamic-routing/index.mdx`** — name the endpoint explicitly (`/compat/chat/completions`) in step 5 of Getting Started, link to the usage page, and add a `note` pre-empting the deprecation badge the reader is about to hit.
3. **`ai-gateway/features/dynamic-routing/usage.mdx`** — the page description read "using the OpenAI SDK or REST API", but none of its three examples use the REST API (they are OpenAI SDK, direct HTTP, and the Workers AI binding). Corrected, since it repeats the same false claim.

No behaviour or product change, wording only. Happy to adjust the callout titles if maintainers prefer different phrasing — the substance is that `/compat` needs a documented carve-out for dynamic routing.

I left `sidebar.badge: Deprecated` on `chat-completion.mdx` alone, since the nav slot only fits one word and I did not want to make that call unilaterally. Flagging it as the remaining piece of the same confusion, if you want it changed.

### Documentation checklist

### DIFF
--- a/src/content/docs/ai-gateway/features/dynamic-routing/index.mdx
+++ b/src/content/docs/ai-gateway/features/dynamic-routing/index.mdx
@@ -60,4 +60,10 @@ Ensure your gateway has [authentication](/ai-gateway/configuration/authenticatio
    - Click **Save** to save the state. You can always roll back to earlier versions from **Versions**.
    - Deploy the version to make it live.
 5. Call the route from your code.
-   - Use the [OpenAI compatible](/ai-gateway/usage/chat-completion/) endpoint, and use the route name in place of the model, for example, `dynamic/support`.
+   - Use the [OpenAI compatible](/ai-gateway/usage/chat-completion/) endpoint (`/compat/chat/completions`), and use the route name in place of the model, for example, `dynamic/support`. See [Using a dynamic route](/ai-gateway/features/dynamic-routing/usage/) for examples.
+
+:::note
+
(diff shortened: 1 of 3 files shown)


superhighfives (approved):
Looks good to me! Thanks @kai-cloudflare. 👍

superhighfives:
@kai-cloudflare do you mind rebasing and I'll get this merged?

PR #32554, opened 2026-08-05
Switch to Cloudflare Images & fix rolldown OOM

mvvmm:
> **Note:** This was originally a copy of #32438 by @ericclemmons.

### Summary

This repo times out at 30 minutes on Workers Builds due to issues with 8 GB of memory (vs. GitHub's 16 GB).

Until that's resolved, there are a few fixes that I found were able to get ~11m builds:

- [x] https://github.com/rolldown/rolldown/pull/10554 is the root issue, causing memory pressure ~5 minutes in and hanging until the timeout at 30 minutes.
- [x] Switch to Cloudflare Images (to avoid spending 20+ minutes on Sharp image compression at build time)
  https://docs.astro.build/en/guides/integration-guide/cloudflare/#imageservice
- [x] Revert https://github.com/cloudflare/cloudflare-docs/pull/31192, because Workers Builds expects the default cache location at `node_modules/.astro` - done in #32587
- [x] Validate Cloudflare Images works as a replacement to `sharp` (which reduced both build times & memory)
- [x] Add Image Resizing subrequest bypass to `worker/index.ts` and `worker/index.preview.ts` — the `@astrojs/cloudflare` image service uses Cloudflare Image Resizing at runtime, which makes subrequests to fetch source images from `/_astro/`. Without a bypass in the Worker entrypoints, these subrequests loop back through the Worker and return 307 redirects instead of the original asset. The bypass checks for the `image-resizing` token in the `Via` header and passes matching `/_astro/` requests directly to `env.ASSETS`.

#### Before

Screenshot 2026-07-30 at 5 06 47 PM

#### After (cold cache)

Screenshot 2026-07-30 at 5 08 08 PM

#### After (warm cache)

Screenshot 2026-07-30 at 5 07 24 PM

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -281,10 +281,7 @@ export default defineConfig({
 	markdown,
 	image: {
 		service: {
-			entrypoint: "astro/assets/services/sharp",
-			config: {
-				limitInputPixels: false,
-			},
+			entrypoint: "@astrojs/cloudflare/image-service",
 		},
 	},
 	server: {
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
 		"@actions/github": "9.1.1",
 		"@apidevtools/swagger-parser": "12.1.0",
 		"@astrojs/check": "0.9.10",
+		"@astrojs/cloudflare": "14.1.7",
 		"@astrojs/markdown-remark": "7.2.1",
 		"@astrojs/markdown-satteri": "0.3.4",
 		"@astrojs/mdx": "^7.0.4",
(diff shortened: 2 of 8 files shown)


mvvmm:
Skipping **CR-453e5fe5f2e4** as originally filed — the `Via: image-resizing` passthrough has been removed entirely from both workers. Image Resizing subrequests now reach ASSETS directly because we reverted `run_worker_first: false` in an earlier commit. No scoped bypass is needed.

mvvmm:
Correction to my previous comment: **CR-453e5fe5f2e4** is being addressed — the `Via: image-resizing` bypass is being scoped to `/_astro/` paths only, so preview-specific routes (robots.txt, sitemap, openapi.json, etc.) are unaffected.

ahaywood (approved):

PR #32555, opened 2026-08-05
Document containers rollout none option

skepticfx:
### Summary

Documents `--containers-rollout=none` in the Wrangler `deploy` command reference. This option deploys Worker changes without building or updating associated Containers, which is useful when the container code has not changed.

This documents existing Wrangler behavior, so no changelog entry is needed.

### Screenshots (optional)

Not applicable; this is a command reference update.

### Documentation checklist

- [x] The change adheres to the [[documentation style guide](https://developers.cloudflare.com/style-guide/)](https://developers.cloudflare.com/style-guide/).


gpanders (approved):

nikitassharma (approved):

mattjohnsonpint (approved):

irvinebroque:
we can't konw this in some way automatically?

skepticfx:
> we can't konw this in some way automatically?

Are you saying this is already automatically generated from wrangler? If so we should indeed do that instead. Let me look at what is in Wrangler's cli docs.

irvinebroque:
No I mean why does there need to be a flag?

skepticfx:
> No I mean why does there need to be a flag?

Sometimes users just want to iterate on their worker code, regardless of the container changes. This allows them to do that by not combining Container rollouts with the Worker updates within the same project.

Also, some customers use the Containers rollout API from their deploy systems to have more granular control over their container rollouts.

irvinebroque (approved):
I get it now. What Nafeez said we should be clear about, that you don’t need to do this

PR #32569, opened 2026-08-06
[Cache] Document CDN-Cache-Control as a BYPASS cause

ngayerie:
Two small follow-ups to #32423.

**What**

- Add a bullet to the BYPASS lists in `cache-responses.mdx` and `investigating-uncached-responses.mdx` documenting `Cloudflare-CDN-Cache-Control` and `CDN-Cache-Control`, with the precedence chain `Cloudflare-CDN-Cache-Control` > `CDN-Cache-Control` > `Cache-Control`. Uses a concrete example so the reader sees that a cacheable `Cache-Control` alongside a non-cacheable `CDN-Cache-Control` still produces `BYPASS`.
- Rewrite the `Age` callout in `cache-responses.mdx` and the corresponding HIT-verification bullet in the troubleshooting page. Both currently mention `CacheTieredFill=true`, which is an `http_requests` Logpush field, not an HTTP header — readers cannot observe it on a response.

**Why**

Neither BYPASS list mentioned `Cloudflare-CDN-Cache-Control` or `CDN-Cache-Control`. When both `Cache-Control` and `CDN-Cache-Control` are set, `CDN-Cache-Control` wins per [CDN-Cache-Control](https://developers.cloudflare.com/cache/concepts/cdn-cache-control/#header-precedence), so a reader debugging a `BYPASS` while only looking at `Cache-Control` has no signal from these pages that another header is in play.

@zaidoon1 for review.


src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -15,7 +15,7 @@ The `CF-Cache-Status` header output indicates whether a resource is cached or no
 :::note[`Age` response header]
 The `Age` response header is a header returned from cache that specifies the time in seconds that an asset has been in Cloudflare's cache. This value resets if the asset is revalidated, purged, or evicted and then re-cached.
```
foreseaz: As we're making this note more clear, I suggest updating this paragraph too, e.g. `Cloudflare's cache` is vague.

```suggestion
The `Age` response header is returned when Cloudflare serves a response from a local cache object on the data center handling the request. It is the number of seconds since that object was admitted or last revalidated at that cache. It resets if the asset is revalidated, purged, or evicted and then re-cached.
```

ngayerie: [AI] Thanks @foreseaz — applied in `7ebf4751`. Per-data-center scope is more accurate, and it also lines up with the second paragraph of the callout (which describes Tiered Cache behavior between lower and upper tiers).

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -15,7 +15,7 @@ The `CF-Cache-Status` header output indicates whether a resource is cached or no
...
 
-The `Age` header is only present for responses served from the cache. It will not appear on a cache MISS, dynamic traffic, the first request that populates the lower tier HIT from tiered cache `CacheTieredFill=true` or any responses that did not originate from the cache (for example, responses generated by a Worker that bypassed the cache).
+The `Age` header is only present for responses served from the cache. It will not appear on a cache MISS, dynamic traffic, the first request that populates a lower-tier data center's cache from an upper tier via [Tiered Cache](/cache/how-to/tiered-cache/), or any responses that did not originate from the cache (for example, responses generated by a Worker that bypassed the cache).
```
foreseaz: LGTM on clarifying the tiered-cache case.
Optional follow-up: this note still underspecifies when Age appears. We only set it for pure local cache serves (HIT / STALE / UPDATING) — not REVALIDATED, EXPIRED, or BYPASS.

Suggest update:
```suggestion
`Age` is typically present for `HIT`, `STALE`, and `UPDATING`.
It is not present for:
- `MISS`, `EXPIRED`, `REVALIDATED`, `BYPASS`, `DYNAMIC`, and `NONE`/`UNKNOWN`
- the first request that populates a lower-tier data center's cache from an upper tier via [Tiered Cache](/cache/how-to/tiered-cache/) (even if `CF-Cache-Status` is `HIT`)
- responses that never hit cache (for example, a Worker that bypassed the cache)

After that lower-tier data center is populated, later local hits include `Age`.
```

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -55,6 +55,7 @@ Cloudflare considered the asset eligible for cache at request time — either be
 Common reasons the origin response is treated as not cacheable include:
 
 * The response exceeds the [maximum cacheable file size](/cache/concepts/default-cache-behavior/#cacheable-size-limits) for your plan.
+* The origin returned a `Cloudflare-CDN-Cache-Control` or `CDN-Cache-Control` header with a non-cacheable directive. Cloudflare evaluates these headers ahead of `Cache-Control`, in the precedence `Cloudflare-CDN-Cache-Control` > `CDN-Cache-Control` > `Cache-Control`. For example, an origin returning both `Cache-Control: public, max-age=3600` and `CDN-Cache-Control: no-store` produces `BYPASS`. Inspect every response header, not just `Cache-Control`. Refer to [CDN-Cache-Control](/cache/concepts/cdn-cache-control/) for the precedence rules.
```
foreseaz: good call, maybe we want to mention `Surrogate-Control` too.

```
Cloudflare-CDN-CC > CDN-CC > Surrogate-Control > Cache-Control
```

cc @zaidoon1

src/content/docs/cache/concepts/cache-responses.mdx
foreseaz: this suggestion needs another look from @zaidoon1 or other folks.

foreseaz:
Both changes LGTM. Left some nits to make the existing change more clear, thanks for the effort @ngayerie

ngayerie:
Hi @mbullock1986 @zaidoon1 
Would it be possible for you to review/approve this PR?
Thanks!

zaidoon1 (commented):
Checked the CDN-Cache-Control and `Age` claims against actual behavior. The BYPASS bullet's core point is right, but a few details need tightening. Comments inline.

src/content/docs/cache/concepts/cache-responses.mdx
zaidoon1: Worth naming the directives. `no-store` and bare `private` are what produce BYPASS here. `no-cache`, `max-age=0`, and `s-maxage=0` in these headers don't: you get MISS on the first request, then REVALIDATED or EXPIRED.

The bullet two below lists those same three directives as BYPASS causes, so someone with `CDN-Cache-Control: no-cache` reads both and comes away wrong. Its Origin Cache Control condition doesn't apply to these headers either, since they follow Origin Cache Control semantics whether or not the zone has the setting on.

zaidoon1: The example reads as unconditional. An Edge TTL that ignores origin cache-control overrides `CDN-Cache-Control: no-store` the same way it overrides `Cache-Control: no-store`, so with that rule in place these two headers give MISS then HIT. The `no-store` bullet right below already carries that exception, and the CDN-Cache-Control page says Edge Cache TTL wins.

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -13,9 +13,9 @@ head:
 The `CF-Cache-Status` header output indicates whether a resource is cached or not. To investigate cache responses returned by this header, use services like [Redbot](https://redbot.org/), [webpagetest.org](http://www.webpagetest.org/), or a visual tool like [Cloudflare Optics plugin](https://chromewebstore.google.com/detail/cloudflare-optics/mdjgbjnbdnhneejmmaabmccfehigbjbe).
 
 :::note[`Age` response header]
-The `Age` response header is a header returned from cache that specifies the time in seconds that an asset has been in Cloudflare's cache. This value resets if the asset is revalidated, purged, or evicted and then re-cached.
+The `Age` response header is returned when Cloudflare serves a response from a local cache object on the data center handling the request. It is the number of seconds since that object was admitted or last revalidated at that cache. It resets if the asset is revalidated, purged, or evicted and then re-cached.
```
zaidoon1: "at that cache" doesn't hold with Tiered Cache. A HIT from a data center that filled seconds ago can still come back with a large `Age`, because the value tracks the upper tier's copy rather than the local one. Filled 10 seconds ago from an upper tier that had held it two hours, and you get `Age: 7200`.

The wording this replaces ("the time in seconds that an asset has been in Cloudflare's cache") was closer. I'd keep it network-wide and drop "at that cache".

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -13,9 +13,9 @@ head:
...
 
-The `Age` header is only present for responses served from the cache. It will not appear on a cache MISS, dynamic traffic, the first request that populates the lower tier HIT from tiered cache `CacheTieredFill=true` or any responses that did not originate from the cache (for example, responses generated by a Worker that bypassed the cache).
+The `Age` header is only present for responses served from the cache. It will not appear on a cache MISS, dynamic traffic, the first request that populates a lower-tier data center's cache from an upper tier via [Tiered Cache](/cache/how-to/tiered-cache/), or any responses that did not originate from the cache (for example, responses generated by a Worker that bypassed the cache).
```
zaidoon1: Too strong in both directions.

`Age` is absent on EXPIRED and REVALIDATED when the request actually revalidated against the origin, and those still count as served from cache. It's present on HIT, STALE, and UPDATING.

Other direction: an `Age` set by the origin still reaches the client on uncacheable responses. Origin sends `Age: 42`, status comes back DYNAMIC or BYPASS, client sees `Age: 42`. So "will not appear on a cache MISS" only holds when the origin didn't send one.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -40,6 +40,7 @@ Once the request is eligible, subsequent responses reflect the response-time dec
 The request was eligible for cache, but the origin response or configuration prevented Cloudflare from storing it. Common causes:
 
 - **The response exceeds the [cacheable size limit](/cache/concepts/default-cache-behavior/#cacheable-size-limits) for your plan.** Split the object into smaller assets, or move to a plan with a higher limit. [R2](/r2/) is an origin storage alternative — it does not raise the CDN cacheable size limit.
+- **The origin returned a `Cloudflare-CDN-Cache-Control` or `CDN-Cache-Control` header with a non-cacheable directive.** Cloudflare evaluates these headers ahead of `Cache-Control`, in the precedence `Cloudflare-CDN-Cache-Control` > `CDN-Cache-Control` > `Cache-Control`. For example, an origin returning both `Cache-Control: public, max-age=3600` and `CDN-Cache-Control: no-store` produces `BYPASS`. Inspect every response header, not just `Cache-Control`. Refer to [CDN-Cache-Control](/cache/concepts/cdn-cache-control/) for the precedence rules.
```
zaidoon1: "Inspect every response header" won't surface `Cloudflare-CDN-Cache-Control`. Cloudflare doesn't return that one to the client, so it's invisible from the response side. That's the useful bit for a troubleshooting page: if you see BYPASS with no `CDN-Cache-Control` in the response, check what the origin actually sent, or have the origin drop the header and retest.

Same two points I left on the `cache-responses.mdx` bullet apply here.

ngayerie:
[AI] Thanks @zaidoon1 — all 5 comments applied in `26b189cb`.

**`cache-responses.mdx`**

1. Age callout scope — dropped `at that cache`, kept it network-wide, and added the Tiered Cache clarification (`HIT` served locally can carry an `Age` inherited from an upper tier, so the value can be older than the last local fill).
2. Age status enumeration — rewritten by `cf-cache-status`: `Age` is set on `HIT`, `STALE`, `UPDATING`. It is absent on `MISS`, on `EXPIRED` and `REVALIDATED` when the request revalidated against the origin, and on `DYNAMIC`, `BYPASS`, `NONE`/`UNKNOWN`. Origin-set `Age` on an uncacheable response is proxied to the client unchanged.
4. `CDN-Cache-Control` BYPASS bullet — named the directives (only `no-store` and bare `private` produce `BYPASS`; `no-cache`/`max-age=0`/`s-maxage=0` produce `MISS` then `REVALIDATED`/`EXPIRED`), dropped the Origin Cache Control condition (these headers follow OCC-style semantics regardless of the zone setting).
5. `CDN-Cache-Control` BYPASS bullet — added the [Edge Cache TTL](https://developers.cloudflare.com/cache/how-to/cache-rules/settings/#edge-ttl) override, same as `Cache-Control: no-store`.

**`investigating-uncached-responses.mdx`**

3. `CDN-Cache-Control` BYPASS bullet — mirrored the same corrections, and reframed the troubleshooting angle: noted that Cloudflare does not forward `Cloudflare-CDN-Cache-Control` to the client, so if a reader sees `BYPASS` without a visible `CDN-Cache-Control` header the next step is checking what the origin actually sent (or dropping the header at origin and retesting), not "inspect every response header".

Also reconciled the HIT verification bullet on the same page with your Age scope correction — a fresh local fill from an upper tier can already report a large `Age`, so "may be absent" was wrong.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -81,7 +82,7 @@ If your requests reach different Cloudflare data centers, each produces its own
 
 After adjusting configuration, request the URL twice from the same client and verify the expected outcome for your configuration:
 
-- **Fresh, positive Edge TTL:** `cf-cache-status: HIT` and an `Age` header that increases on subsequent requests. `Age` may be absent on the first lower-tier request that populates the local cache from a tiered cache fill (`CacheTieredFill=true`).
+- **Fresh, positive Edge TTL:** `cf-cache-status: HIT` and an `Age` header. With [Tiered Cache](/cache/how-to/tiered-cache/), the `Age` value reflects the object's age in Cloudflare's network-wide cache — a fresh local fill from an upper tier can already report a large `Age` inherited from that upper tier.
```
zaidoon1: The clause you dropped here was right. On the request that fills a lower tier from an upper tier, the client gets `HIT` with no `Age` at all. The `HIT` reflects the upper tier, and the local data center hasn't served it from cache yet, so there's no age to report.

As written, the checklist tells people to expect an `Age` header on exactly the request where it won't be there. The new note about the value being large is good, it just describes the request after. Both belong here: absent on the fill, then present and possibly already large.

I'd keep "that increases on subsequent requests" too, since that's the signal someone is actually checking.

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -13,9 +13,9 @@ head:
 The `CF-Cache-Status` header output indicates whether a resource is cached or not. To investigate cache responses returned by this header, use services like [Redbot](https://redbot.org/), [webpagetest.org](http://www.webpagetest.org/), or a visual tool like [Cloudflare Optics plugin](https://chromewebstore.google.com/detail/cloudflare-optics/mdjgbjnbdnhneejmmaabmccfehigbjbe).
 
 :::note[`Age` response header]
-The `Age` response header is a header returned from cache that specifies the time in seconds that an asset has been in Cloudflare's cache. This value resets if the asset is revalidated, purged, or evicted and then re-cached.
+The `Age` response header is returned when Cloudflare serves a response from cache. It is the number of seconds an asset has been in Cloudflare's cache since it was admitted or last revalidated, and it resets on revalidation, purge, or eviction. With [Tiered Cache](/cache/how-to/tiered-cache/), `Age` reflects the object's age in Cloudflare's network-wide cache — a `HIT` served locally can carry an `Age` inherited from an upper tier, so the value can be older than the last local fill.
```
zaidoon1: "resets on revalidation" doesn't quite hold now that the value is network-wide. With Tiered Cache, a lower tier revalidating against an upper tier that still has the object picks the upper tier's age back up rather than resetting to zero. The sentence after covers it loosely, but the reset rule is stated first and without qualification.

"revalidated against the origin" would fix it, and it matches how you've phrased the paragraph below.

src/content/docs/cache/concepts/cache-responses.mdx
```diff
@@ -55,6 +55,7 @@ Cloudflare considered the asset eligible for cache at request time — either be
 Common reasons the origin response is treated as not cacheable include:
 
 * The response exceeds the [maximum cacheable file size](/cache/concepts/default-cache-behavior/#cacheable-size-limits) for your plan.
+* The origin returned `no-store` or bare `private` in a `Cloudflare-CDN-Cache-Control` or `CDN-Cache-Control` header. Cloudflare evaluates these headers ahead of `Cache-Control`, in the precedence `Cloudflare-CDN-Cache-Control` > `CDN-Cache-Control` > `Cache-Control` — so an origin returning `Cache-Control: public, max-age=3600` together with `CDN-Cache-Control: no-store` produces `BYPASS`. A [Cache Rule](/cache/how-to/cache-rules/) with an [Edge Cache TTL](/cache/how-to/cache-rules/settings/#edge-ttl) setting that ignores origin cache-control overrides these directives, same as it does for `Cache-Control: no-store`. `no-cache`, `max-age=0`, or `s-maxage=0` in these headers do not produce `BYPASS`; they produce `MISS` on the first request, then [`REVALIDATED`](#revalidated) or [`EXPIRED`](#expired). Refer to [CDN-Cache-Control](/cache/concepts/cdn-cache-control/) for the precedence rules.
```
zaidoon1: Semicolon here. The style guide says avoid them, and you split one out earlier in this PR. Two sentences reads fine: "...do not produce `BYPASS`. They produce `MISS` on the first request, then..."

Same one is in the matching bullet on the troubleshooting page.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -40,6 +40,7 @@ Once the request is eligible, subsequent responses reflect the response-time dec
...
+- **The origin returned `no-store` or bare `private` in a `Cloudflare-CDN-Cache-Control` or `CDN-Cache-Control` header.** Cloudflare evaluates these headers ahead of `Cache-Control`, in the precedence `Cloudflare-CDN-Cache-Control` > `CDN-Cache-Control` > `Cache-Control`. An origin returning `Cache-Control: public, max-age=3600` together with `CDN-Cache-Control: no-store` produces `BYPASS`. Cloudflare does not forward `Cloudflare-CDN-Cache-Control` to the client, so it is not visible in the response — if you see `BYPASS` without a `CDN-Cache-Control` header, check what the origin actually sent, or have the origin drop the header and retest. A [Cache Rule](/cache/how-to/cache-rules/) with an [Edge Cache TTL](/cache/how-to/cache-rules/settings/#edge-ttl) setting that ignores origin cache-control overrides both directives. `no-cache`, `max-age=0`, or `s-maxage=0` in these headers do not produce `BYPASS`; they produce `MISS` on the first request, then [`REVALIDATED`](/cache/concepts/cache-responses/#revalidated) or [`EXPIRED`](/cache/concepts/cache-responses/#expired). Refer to [CDN-Cache-Control](/cache/concepts/cdn-cache-control/) for the precedence rules.
```
zaidoon1: Optional. This bullet is six sentences where its neighbours are one or two, and it's carrying four points: the directives, the precedence, the visibility caveat, and the Edge TTL override. All worth saying, but it's a wall on a page people scan while debugging. Pulling the `no-cache`/`max-age=0`/`s-maxage=0` clarification into its own bullet would help, since that's the one telling the reader to stop looking here.

zaidoon1 (commented):
Latest addresses everything I raised. One clause got lost in the rewrite that I'd put back, plus a couple of small things.

ngayerie:
[AI] Thanks @zaidoon1 — all 4 comments applied in `f1675c05`.

1. **Age reset rule** — qualified as `revalidation against the origin`, matching the phrasing on the paragraph below. A lower tier revalidating against an upper tier that still has the object picks up the upper tier's age, so the unqualified `revalidation` was wrong for Tiered Cache.
2. **Semicolon on `cache-responses.mdx` line 58** — split into two sentences.
3. **Six-sentence bullet on the troubleshooting page** — pulled the `no-cache`/`max-age=0`/`s-maxage=0` clarification into its own bullet. It's the clause that tells the reader to look under [Repeated MISS](https://developers.cloudflare.com/cache/troubleshooting/investigating-uncached-responses/#repeated-miss--cacheable-but-not-in-cache) instead, so it earns the separate bullet. That split also removed the matching semicolon on the troubleshooting page.
4. **HIT verify `Age` bullet** — restored the absent-on-fill case: on the request that fills a lower-tier data center from an upper tier, the client gets `HIT` with no `Age` at all, because the local data center has not served it from cache yet. Subsequent requests include `Age`, and with Tiered Cache the value can already be large — kept the "increases on subsequent requests" signal you called out.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -40,6 +40,8 @@ Once the request is eligible, subsequent responses reflect the response-time dec
...
+- **`no-cache`, `max-age=0`, or `s-maxage=0` in these headers do not produce `BYPASS`.** They produce `MISS` on the first request, then [`REVALIDATED`](/cache/concepts/cache-responses/#revalidated) or [`EXPIRED`](/cache/concepts/cache-responses/#expired) — check under [Repeated MISS](#repeated-miss--cacheable-but-not-in-cache) instead.
```
zaidoon1: The pointer sends people the wrong way. Repeated MISS is about cache key variance and eviction. Someone who lands here sees MISS once and then REVALIDATED on every request after, so that section has nothing for them and they'll go dig through cache key config for a response that's caching fine.

"Confirm the response reaches cache" is the right target, its second bullet already covers the REVALIDATED and EXPIRED outcome. Or drop the pointer, since the bullet already says what to expect.

zaidoon1: "these headers" only resolves if you've read the bullet above, which won't happen for someone scanning. Worth naming them here. Helps twice over, since a "not this" item sitting in a list of BYPASS causes is easy to skim wrong.

zaidoon1 (commented):
Two small ones on the new bullet, neither blocking.

ngayerie:
[AI] Thanks @zaidoon1 — both applied in `6afd6b1b`.

- **`"these headers"` reference** — replaced with the explicit header names at the start of the bullet, so it stands alone for a scanner reading only the bolded openings.
- **Pointer to `Repeated MISS`** — dropped. That section covers cache key variance and eviction, neither of which matches the `MISS` → `REVALIDATED`/`EXPIRED` pattern this bullet describes, so it was sending readers the wrong way. Kept the sequence description — [Confirm the response reaches cache](https://developers.cloudflare.com/cache/troubleshooting/investigating-uncached-responses/#confirm-the-response-reaches-cache) already covers verification with those same two statuses.

src/content/docs/cache/concepts/cache-responses.mdx
ngayerie: Hi @mbullock1986 , @zaidoon1 
What do you think? Should we mention Surrogate-Control in this page?

mbullock1986 (approved):
based on Zaidoon's comments being fixed

ngayerie:
Hi @zaidoon1 
Happy for me to merge?

zaidoon1 (approved):

PR #32571, opened 2026-08-06
Update Sandbox SDK docs covering proxy & environment variables

aron-cf:
### Summary

Updates Sandbox SDK docs to replace the outdated proxy framework guide with a note that points readers to the current outbound traffic guide. Adds `rpc` as a valid `SANDBOX_TRANSPORT` mode and updates related Sandbox transport and security guidance so RPC is described as the current persistent-connection option while WebSocket remains deprecated.

### Documentation checklist


src/content/docs/sandbox/guides/proxy-requests.mdx
```diff
@@ -1,279 +1,15 @@
 ---
```
ghostwriternr, line 1: Why not just delete this guide, and IF needed (don't think we do imo), configure a redirect from the deleted route to the outbound traffic guide?

ghostwriternr (approved):

src/content/docs/sandbox/guides/proxy-requests.mdx
aron-cf: Fair

PR #32579, opened 2026-08-06
[Cloudflare for SaaS] Clarify O2O works with full and partial DNS setups

damiete:
## What

Clarify that orange-to-orange (O2O) applies when a customer zone has a proxied CNAME to a SaaS provider on Cloudflare, including partial (CNAME) DNS setups where Cloudflare is not the authoritative DNS provider.

## Why

The O2O how-it-works page currently reads as if Cloudflare must be authoritative DNS ("If your authoritative DNS is Cloudflare, you have the option to proxy…"). Customers on partial setup with SaaS providers already use O2O; the docs should match that behavior and call out that the proxied CNAME in the customer zone is what enables O2O.

## Changes

- **How O2O works:** prerequisites, full vs partial setup, CNAME target requirements, note when adding your own zone later enables O2O
- **Hostname priority:** callout distinguishing single-zone hostname priority from O2O; Scenario 1 no longer requires "authoritative DNS"
- **Partial setup:** short cross-link section for SaaS / O2O
- **Product compatibility:** one line that full and partial setups are supported

Provider-specific steps remain in the existing provider guides.

## Pages

- `/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/how-it-works/`
- `/ssl/reference/certificate-and-hostname-priority/`
- `/dns/zone-setups/partial-setup/setup/`
- `/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/product-compatibility/`

### DIFF
--- a/src/content/docs/dns/zone-setups/partial-setup/setup.mdx
+++ b/src/content/docs/dns/zone-setups/partial-setup/setup.mdx
@@ -108,3 +108,16 @@ If your zone stays in **Pending Nameserver Update** status after adding the veri
 ## Other record types
 
 If you are preparing a conversion from CNAME setup (partial) to primary setup (full), or if you have a more specific use case, you can use the [Create DNS Record](/api/resources/dns/subresources/records/methods/create/) API endpoint to create DNS records of any supported type.
+
+
+---
+
+## Use with a SaaS provider (O2O)
+
+If you use a partial setup and a SaaS platform that runs on Cloudflare, you can send traffic through your Cloudflare zone and then the provider's zone. That path is called [orange-to-orange (O2O)](/cloudflare-for-platforms/cloudflare-for-saas/saas-customers/how-it-works/).
+
+1. In your Cloudflare zone, create a [proxied](/dns/proxy-status/) CNAME for the hostname that targets the name your SaaS provider specifies.
+2. At your authoritative DNS provider, keep DNS that delivers visitors to Cloudflare for that hostname (for example `{hostname}.cdn.cloudflare.net`, or the provider's Cloudflare hostname when their guide uses that pattern).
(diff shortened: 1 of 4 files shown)


src/content/docs/dns/zone-setups/partial-setup/setup.mdx
```diff
@@ -108,3 +108,16 @@ If your zone stays in **Pending Nameserver Update** status after adding the veri
 ## Other record types
 
 If you are preparing a conversion from CNAME setup (partial) to primary setup (full), or if you have a more specific use case, you can use the [Create DNS Record](/api/resources/dns/subresources/records/methods/create/) API endpoint to create DNS records of any supported type.
+
+
+---
+
+## Use with a SaaS provider (O2O)
```
hannes-cf, line 115: I am not convinced that these steps belong into the setup section for partial zones. It literally just says: You can use also use O2O on partial zones and then explains generically how to add proxy a certain name on a partial zone, There is nothing special for that with respect to O2O.

hannes-cf (changes requested):
Please drop the changes on `/dns/zone-setups/partial-setup/setup.mdx`.

baubuchon-cf (approved):
Content looks accurate to me. The old wording made it sound like Cloudflare had to be your authoritative DNS for O2O to work, and that's not right.. good catch fixing that.

Agree with hannes-cf on partial-setup/setup.mdx though. That section just repeats what's already on the how-it-works page. I'd drop it or cut it to a one-line link, then it's good to merge.

Approving with that one ask.

baubuchon-cf (changes requested):
Changing to request changes to align with hannes-cf's feedback. The how-it-works.mdx changes look great — those should stay. Please remove only the new section in dns/zone-setups/partial-setup/setup.mdx (the "Use with a SaaS provider (O2O)" section that hannes flagged). Once that file is reverted, I'll re-approve immediately.

PR #32583, opened 2026-08-06
[Directory] Eliminate directory-entry lookup noise from non-product routes

mvvmm:
## What changed

Added `getDirectoryEntryBySection()` in `src/util/directory.ts` — a shared helper that resolves directory entries by URL section (first path segment) using a pre-built map from `entry.data.entry.url`. Preserves dots so `1.1.1.1` resolves correctly, and skips known non-product routes (`videos`, `glossary`, `plans`, `directory`, `sponsorships`, `resources`, `404`).

Replaced 6 scattered `getEntry("directory", section.replaceAll(".", ""))` call sites across 4 files with the shared helper:
- `src/pages/[...slug].astro`
- `src/components/PageHead.astro`
- `src/components/cf/ResourcesBySelector.astro`
- `src/util/sidebar.ts` (3 locations)

## Root causes

- `replaceAll(".", "")` corrupted `1.1.1.1` into `1111`, which has no directory entry (85 warnings)
- Non-product routes like `/videos/`, `/resources/`, `/glossary/` were looked up as directory entries (37 warnings)
- The lookup was duplicated across 4 files with no shared guard

## Results

| Metric | Before | After |
|--------|--------|-------|
| `Entry directory → ... was not found.` | 122 | 0 |
| Pages built | 8,802 | 8,802 |

## Checklist

- [x] `pnpm run check` — 0 errors
- [x] `pnpm run build` — 8,802 pages, no new warnings
- [x] `pnpm exec prettier` — clean

### DIFF
--- a/src/components/PageHead.astro
+++ b/src/components/PageHead.astro
@@ -5,7 +5,6 @@
  */
 import { existsSync } from "node:fs";
 import { join } from "node:path";
-import { getEntry } from "astro:content";
 import { differenceInCalendarDays } from "date-fns";
 import { config } from "virtual:nimbus/config";
 import {
@@ -22,6 +21,7 @@ import {
 	resolveSocialImagePath,
 } from "~/util/page-head";
 import { stripMarkdownDescription } from "~/util/description";
+import { getDirectoryEntryBySection } from "~/util/directory";
 
 const PRODUCTION_GENERATOR = "Nimbus v0.2.2";
 
@@ -86,11 +86,11 @@ const versionAlternates =
 			}))
 		: [];
 
-// Section slug (first path segment, dots stripped) — matches production.
-const currentSection = pathname.split("/")[1]?.replaceAll(".", "") ?? "";
+// Section slug (first path segment) — matches production.
+const currentSection = pathname.split("/")[1] ?? "";
 const sectionProduct =
 	currentSection && currentSection !== "changelog"
-		? await getEntry("directory", currentSection)
+		? await getDirectoryEntryBySection(currentSection)
 		: undefined;
 const product = meta?.product ?? sectionProduct?.data.entry?.title;
 // productGroup (which honours meta.productGroup) drives the pcx_content_group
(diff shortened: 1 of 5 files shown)


kodster28 (approved):

mvvmm:
## Review Finding Skip Notes

### CR-72f9630fd5e5 — Directory helper can return the wrong product for shared first segments

**Skipping — finding is factually incorrect.**

The concern is that `access.yaml` (URL `/cloudflare-one/...`) sorts before `cloudflare-one.yaml` and would map the `cloudflare-one` section to the Access entry. However, line 19 of `directory.ts` runs unconditionally:

```js
directoryBySection.set(entry.id, entry);
```

When `cloudflare-one.yaml` (ID: `cloudflare-one`) is processed, it **overwrites** the `cloudflare-one` key that was previously set by `access.yaml`'s URL segment. The final map value is the correct `Cloudflare One` entry.

Verified all 20 shared URL segments (`cloudflare-one`, `rules`, `fundamentals`, `waf`, `ssl`, `cache`, `realtime`, `dns`, `workers`, `speed`, `cloudflare-for-platforms`, `analytics`, `stream`, `smart-shield`, `security-center`, `security`, `logs`, `load-balancing`, `cloudflare-wan`, `bots`) — every one has a matching directory entry by ID, so the ID-key overwrite always produces the correct result.

### CR-70324f735b22 — Incomplete non-product route guard

**Skipping — finding is factually incorrect on both counts.**

1. The `getEntry` fallback was already removed in commit `d85d86c`. The current code is `return directoryBySection.get(section)` — no async lookup path remains.

2. `style-guide` has a directory entry (`src/content/directory/style-guide.yaml`), so `getDirectoryEntryBySection("style-guide")` returns it from the map. It does not return `undefined` and does not produce lookup noise. Checked all docs top-level sections — every one has a corresponding directory entry. The `NON_PRODUCT_SECTIONS` set only needs to cover routes that lack directory entries (`404`, `videos`, `glossary`, `plans`, `directory`, `sponsorships`, `resources`), which it already does.

PR #32638, opened 2026-08-10
[MCP Portals] Document auth_credentials shape for bearer auth

scaraclette:
Callers using the API or Terraform to create MCP servers with `auth_type: bearer` currently see only *"provide the token in `auth_credentials`"*, which is ambiguous about whether the field expects a raw token or a JSON-encoded object.

This adds a subsection under "Create an MCP server" documenting the two accepted forms with copy-pasteable examples:
- Raw bearer token (sent as `Authorization: Bearer <token>`)
- JSON-encoded `{"headers": {...}}` for custom or multiple headers


scaraclette:
Skipping **SG-bf3c40a3ba19** (bullet list length, line 563).

The two forms are parallel structural alternatives (raw token vs JSON header map), each paired with its own code example. Rewriting as prose would either collapse the visual separation between the two forms, or require intermediate headings to distinguish them, which is heavier than the current shape. Bulleted parallel examples with code blocks is a shape used elsewhere in this file, and keeping the bullets here optimizes for scannability — a caller looking up which form to use can see both options at a glance.

Addressing the other three findings (SG-9e86248c66a9, SG-d97ff2d31832, SG-467ddd1342f5) in the next push.

asamborski (approved):

PR #32645, opened 2026-08-10
Clarify initial API token permissions

dinasaur404:
### Summary

Clarify which initial API token permission is required to:

- grant users access to an account as members
- create API tokens

### Documentation

Updates the **Generating the initial token** section of the Create tokens via API guide.

### DIFF
--- a/src/content/docs/fundamentals/api/how-to/create-via-api.mdx
+++ b/src/content/docs/fundamentals/api/how-to/create-via-api.mdx
@@ -20,7 +20,10 @@ The API Token Template [**Create additional tokens**](/fundamentals/api/referenc
 
 ## Generating the initial token
 
-Before you can create tokens via the API, you need to [generate the initial token](/fundamentals/api/get-started/create-token/) via the Cloudflare dashboard.
+Before you can create tokens via the API, you need to [generate the initial token](/fundamentals/api/get-started/create-token/) via the Cloudflare dashboard. Select the permission required for the task you want the token to perform:
+
+- To grant users access to an account as members, select **Account** > **Account Settings** > **Edit**.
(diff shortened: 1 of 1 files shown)


dinasaur404:
Review triage:

- **CR-1b2f8ca97e8c — Disagree.** This PR intentionally documents both supported initial-token permission paths. Account membership management is relevant when the caller uses the API to grant users access as account members; **Account Settings > Edit** is required for that operation. **Account API Tokens > Edit** is required when the caller creates API tokens. I will preserve both cases but clarify that the permission depends on the API operation.
- **SG-5d8fac8e1b78 — Agree.** I will rewrite the two-item list as prose.

dinasaur404:
Review triage for **CR-f68d1213e9fa**: **Partially agree.** The guidance should state that the **Create additional tokens** template grants **Account > Account API Tokens > Edit**, rather than implying that permission can be selected manually. I will also scope the recommendation against extra permissions specifically to tokens created from that template. I disagree that the member-management case should be removed: documenting the separate **Account > Account Settings > Edit** path is the purpose of this change, and it applies when the caller uses the API to grant users account access. I will make the two operations and their separate token requirements explicit.

dinasaur404:
Review triage for **CR-9d1dbdae068e**: **Agree.** The page supports both user-owned and account-owned token creation, but the current wording conflates their permissions. I will distinguish the three operations explicitly: **Account Settings > Edit** for member access, **Account API Tokens > Edit** for account-owned token creation, and the **Create additional tokens** template (**User > API Tokens > Edit**) for user-owned token creation. This keeps the requested account-scoped guidance while making the existing note and template reference consistent.

AdamBouhmad (approved):

PR #32656, opened 2026-08-11
Add Workers storage product decision tree

ivoryibu:
## Summary

- Adds a Mermaid decision tree to the Workers storage-options page.
- Routes readers through product tradeoffs for Hyperdrive, R2/Pipelines, Analytics Engine, Vectorize, Queues, KV, D1, and Durable Objects.
- Updates the SQL database options copy to clarify D1 vs. SQLite-backed Durable Objects: D1 for relational application data; Durable Objects for per-object state that needs coordination, serialization, or WebSockets.

## Test plan

- `prettier --check src/content/docs/workers/platform/storage-options.mdx`
- `git diff --check`
- `astro check --minimumFailingSeverity=hint`
- `tsc --noEmit -p ./worker/tsconfig.json`

### DIFF
--- a/src/content/docs/workers/platform/storage-options.mdx
+++ b/src/content/docs/workers/platform/storage-options.mdx
@@ -17,11 +17,83 @@ This guide describes the storage & database products available as part of Cloudf
 
 ## Choose a storage product
 
-The following table maps our storage & database products to common industry terms as well as recommended use-cases:
+The following table maps Cloudflare storage and database products to common industry terms and recommended use cases. Use the table to compare products, then use the decision tree to choose the first product to evaluate.
 
 <Render file="storage-products-table" product="workers" />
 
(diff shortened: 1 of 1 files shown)


irvinebroque:
diagram seems complicated?

MattieTK:
(this is in my review queue and I thought I'd drop some thoughts).

I don't think this diagram is readable. It's far too tall with too many nodes, and it's not helped by the fact that the mermaid integration in docs results in it being popped out in a panel. People are just not going to engage with it.

I made a different example that uses a more responsive decision tree through a new Astro component to do the same thing https://docs-workers-storage-decision-ladder.preview.developers.cloudflare.com/workers/platform/storage-options/ (https://github.com/cloudflare/cloudflare-docs/pull/32671). You're welcome to use/not use it as you wish!

PR #32760, opened 2026-08-14
docs(workers-for-platforms): per-tenant observability with Workers Logs

Ankcorn:
## Why

The Workers for Platforms observability page didn't explain how platforms give each end user an isolated view of their own Worker's logs and metrics. It also described Tail Worker behaviour that no longer matches the runtime.

## What

- Lead the **Logs** section with Workers Logs as the recommended starting point — a fully managed logging platform (storage, indexing, queryable API) with no database, Logpush destination, or Tail Worker to operate.
- Document enabling Workers Logs per user Worker via the `observability` block in upload metadata, and how the platform controls enablement/sampling.
- Add the per-tenant query pattern: scope every Observability API query to the user Worker's `$metadata.service` in the `cloudflare-workers` dataset, applied server-side so one tenant can't read another's telemetry.
- Add a GraphQL example for per-user metrics via `workersInvocationsAdaptive` filtered by `scriptName` (works without Workers Logs enabled).
- Correct the Tail Worker section: logs are collected by attaching a Tail Worker directly to the user Worker — the previous claim about a dispatch-Worker tail auto-collecting for the whole namespace no longer matches runtime behaviour.

### DIFF
--- a/src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
+++ b/src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
@@ -10,7 +10,38 @@ Workers for Platforms provides you with logs and analytics that can be used to s
 
 ## Logs
 
-Learn how to access logs with Workers for Platforms.
+There are a few ways to access logs with Workers for Platforms. For most platforms, [Workers Logs](#workers-logs) is the best starting point: it is a fully managed logging platform, so you get storage, indexing, and a queryable API without building or operating a logging pipeline of your own. Reach for Logpush or Tail Workers when you need to forward logs to your own destination.
+
+### Workers Logs
+
+[Workers Logs](/workers/observability/logs/workers-logs/) automatically stores logs from your user Workers in your Cloudflare account, where you can query them with the [Query Builder](/workers/observability/query-builder/) or the [Workers Observability API](/api/resources/workers/subresources/observability/). Because it is fully managed, you can surface a per-user view of logs in your own dashboard without standing up a database, configuring a Logpush destination, or deploying a Tail Worker.
+
+Enable Workers Logs on a user Worker by including the `observability` setting in the [metadata](/cloudflare-for-platforms/workers-for-platforms/reference/metadata/) when you [upload the script](/api/resources/workers_for_platforms/subresources/dispatch/subresources/namespaces/subresources/scripts/subresources/content/methods/update/) to your dispatch namespace:
+
+```json
+{
+	"main_module": "index.js",
+	"observability": {
+		"enabled": true,
+		"logs": {
+			"enabled": true,
+			"invocation_logs": true,
+			"head_sampling_rate": 1
+		}
+	}
+}
+```
+
+Because you control the upload metadata, you decide whether logging is enabled for each user Worker. Set `head_sampling_rate` to a value between 0 and 1 to log a percentage of requests, or set `observability.enabled` to `false` to turn off collection.
+
+#### Query logs for a single user Worker
+
+Workers Logs are stored in the `cloudflare-workers` dataset. When you query the [Workers Observability API](/api/resources/workers/subresources/observability/) on behalf of an end user, scope every query to the specific user Worker so that one tenant cannot read another tenant's telemetry. Filter on the script name your platform assigned to the user Worker in the dispatch namespace:
+
+- `dataset` — `cloudflare-workers`, the dataset that stores Workers Logs.
+- `$metadata.service` — the user Worker's script name.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
```diff
@@ -10,7 +10,38 @@ Workers for Platforms provides you with logs and analytics that can be used to s
 
 ## Logs
 
-Learn how to access logs with Workers for Platforms.
+There are a few ways to access logs with Workers for Platforms. For most platforms, [Workers Logs](#workers-logs) is the best starting point: it is a fully managed logging platform, so you get storage, indexing, and a queryable API without building or operating a logging pipeline of your own. Reach for Logpush or Tail Workers when you need to forward logs to your own destination.
```
irvinebroque: Here and elsewhere in the docs — I think the "Workers Logs" _versus_ Logpush framing here is confusing, because you can export both logs and traces like this:

https://developers.cloudflare.com/workers/observability/exporting-opentelemetry-data/

But when we say "reach for logpush" — that sends people towards:

https://developers.cloudflare.com/workers/observability/logs/logpush/

Which is different

Can DM about — but are we solving this?

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
```diff
@@ -31,7 +62,7 @@ A [Tail Worker](/workers/observability/logs/tail-workers/) receives information
 
 Use [Tail Workers](/workers/observability/logs/tail-workers/) instead of Logpush if you want granular control over formatting before logs are sent to their destination to receive [diagnostics channel events](/workers/runtime-apis/nodejs/diagnostics-channel), or if you want logs delivered in real-time.
 
-Adding a Tail Worker to  your dispatch Worker collects logs for both the dispatch Worker and for any user Workers in the dispatch namespace. Logs are automatically collected for all new Workers added to a dispatch namespace. To enable logging for an individual user Worker rather than an entire dispatch namespace, add the [Tail Worker configuration](/workers/observability/logs/tail-workers/#configure-tail-workers) directly to the user Worker.
+To collect logs from a user Worker, add the [Tail Worker configuration](/workers/observability/logs/tail-workers/#configure-tail-workers) directly to that user Worker.
```
irvinebroque, line 66: Is this actually the path we recommend?

I would expect the path here to involve something like

- setting up logpush job
- showing you how you can filter out first-party dispatch Worker logs, to get only user workers
- showing you how you can filter only logs for a given user worker


Shipping off logs every invocation of a Tail Worker is not gonna work for most people's o11y pipelines, just can't handle the load

irvinebroque (commented):
@dinasaur404 @nevikashah

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
Ankcorn: No the recommended path is to use workers observability

The doc prioritises them

1. use workers observability for fully managed pipeline and storage
2. use logpush - let's catch up here :)
3. use a tailworker - if you want granular control over formatting before logs are sent to their destination to receive diagnostics channel events, or if you want logs delivered in real-time.

irvinebroque: But above we say:

> There are a few ways to access logs with Workers for Platforms. For most platforms, [Workers Logs](#workers-logs) is the best starting point: it is a fully managed logging platform, so you get storage, indexing, and a queryable API without building or operating a logging pipeline of your own. Reach for Logpush or Tail Workers when you need to forward logs to your own destination.

Which makes it sound like in order to push logs to your own destination you need to use Tail Workers or do something other than using Workers Observability

I don't understand why I can't enable observability (logs) like I can for tracing and export to 3rd party destination — and instead need to use tail workers or use a _different_ flavor of logpush

@nevikashah

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
rianvdm: Yes we are on the path of solving this. Logpush jobs and OTel Destinations will get folded into a single "Export" section in the Dash. Part of Birthday Week, can share docs if needed.

irvinebroque: great!

src/content/docs/cloudflare-for-platforms/workers-for-platforms/configuration/observability.mdx
Ankcorn: ah gotcha - I see the confusion. The word "destination" is overloaded here. This has nothing to do with the destination feature we shipped last year with the traces launch. I will rephrase that

irvinebroque (approved):

Issue #7019, opened 2022-12-13
Destination IP Selector does not exist


iceman3k:
### Which Cloudflare product does this pertain to?

Zero Trust

### Existing documentation URL(s)

https://developers.cloudflare.com/cloudflare-one/tutorials/zero-trust-network-access

### Section that requires update

Network Configuration

### What needs to change?

Destination IP Selector does not exist. Please add as selection.

### How should it change?

Selector Destination IP does not exist. Please correct or create

### Additional information

![Screenshot_20221213_064451](https://user-images.githubusercontent.com/16768871/207342792-e057141a-ea6f-4d41-8f80-7cbb9531d1e7.png)
![selector_selection](https://user-images.githubusercontent.com/16768871/207345057-9b36f654-398f-40b8-a481-aff3b6bd2cd2.jpg)

iceman3k:
Found it.. my bad

PR #32777, opened 2026-08-14
[Workers] Document two event spec compliance compatibility flags

### LINKED ISSUE #6022: EventTarget is broken and not aligned with spec
Spec: https://dom.spec.whatwg.org/#interface-eventtarget
Used e.g. in WebSocket

extra built-in wrappers around `on*` handling in `EventTarget` directly are not aligned with the spec and break usage of anything subclassing `EventTarget`, as adding `on*` handlers in subclasses causes double firing and out of order.

```js
const EventsTest1_AC = {
  async test(ctrl, env, ctx) {
    console.log('# Should print: a c')
    const x = new EventTarget()
    x.addEventListener('hit', () => console.log('a'))
    x.onhit = () => console.log('[SHOULD NOT FIRE]')
    x.addEventListener('hit', () => console.log('c'))
    x.dispatchEvent(new Event('hit'))
    await new Promise((accept) => setTimeout(accept, 100))
  },
}

const EventsTest2_ABC = {
  async test(ctrl, env, ctx) {
    console.log('# Should print: a b c')
    const x = new EventTarget()

    for (const type of ['hit']) {
      let current
      Object.defineProperty(x, `on${type}`, {
        get: () => current,
        set(value) {
          if (current) this.removeEventListener(type, current)
          current = value
          if (current) this.addEventListener(type, current)
        },
      })
    }

    x.addEventListener('hit', () => console.log('a'))
    x.onhit = () => console.log('b')
    x.addEventListener('hit', () => console.log('c'))
    x.dispatchEvent(new Event('hit'))

    await new Promise((accept) => setTimeout(accept, 100))
  },
}

if (globalThis.process?.versions?.node || globalThis.window) {
  await EventsTest1_AC.test()
  console.log()
  await EventsTest2_ABC.test()
}

export { EventsTest1_AC, EventsTest2_ABC }
```

on workerd, prints:
```
# Should print: a c
[SHOULD NOT FIRE]
a
c

# Should print: a b c
b
a
b
c
```

on Node.js, Chrome, Safari, Firefox, Deno and Bun prints:
```
# Should print: a c
a
c

# Should print: a b c
a
b
c
```

rexxars:
### Summary

Documents two new compatibility flags added to the Workers runtime in cloudflare/workerd#7019. Both fix places where our event classes deviate from the DOM and HTML standards.

- `spec_compliant_event_handler_attributes`: `on<type>` handlers, such as `WebSocket.onmessage` and `AbortSignal.onabort`, fire in the order they were assigned rather than ahead of every listener, and fire once rather than twice for a class that implements them by registering a listener. Fixes cloudflare/workerd#6022.
- `spec_compliant_message_event_origin`: `MessageEvent.origin` reports an empty string rather than `null` when there is no origin, and a `WebSocket` opened from a URL reports the origin of that URL.

Each page covers what changes, a before and after example, and the disable flag.

Neither has an `enable_date` yet, so both pages omit that field and use the date they were written as `sort_date`. The flags are opt-in until the runtime PR settles on dates, and I will update this PR with them at the same time as the flag definitions. This goes up alongside the runtime change because `docs/api-updates.md` in workerd asks for the documentation to land before the enable date.

The runtime PR carries a third fix, making `isTrusted` false on events constructed in JavaScript, which has no compatibility flag after review feedback that it did not need one. There is no page for it here, since these pages document flags.

No changelog entry for the two flags, since nothing changes for an existing Worker until they have enable dates. The `isTrusted` fix does take effect when the runtime PR merges, so tell me if you would like a changelog entry for that one and I will add it.

### Documentation checklist

### DIFF
--- a/src/content/compatibility-flags/spec-compliant-event-handler-attributes.md
+++ b/src/content/compatibility-flags/spec-compliant-event-handler-attributes.md
@@ -1,0 +1,46 @@
+---
+_build:
+  publishResources: false
+  render: never
+  list: never
+
+name: "Spec compliant `on<type>` event handlers"
+sort_date: "2026-08-14"
+enable_flag: "spec_compliant_event_handler_attributes"
+disable_flag: "no_spec_compliant_event_handler_attributes"
+---
+
+`on<type>` event handler properties, such as `WebSocket.onmessage` and `AbortSignal.onabort`, behave the way the [DOM](https://dom.spec.whatwg.org/#interface-eventtarget) and [HTML](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-attributes) standards describe them.
+
+Previously, `EventTarget` looked for an `on<type>` property on the object every time an event was dispatched and called it before any listener added with `addEventListener()`. That had two consequences: an `on<type>` handler always ran first rather than in the order it was assigned, and a class that extends `EventTarget` and implements an `on<type>` accessor using `addEventListener()` had its handler called twice per event.
+
+With this flag enabled, `EventTarget` no longer looks for `on<type>` properties. The interfaces that the standards give event handler attributes to (`AbortSignal`, `EventSource`, `MessagePort` and `WebSocket`) implement them as accessors that register an ordinary listener, so:
+
+- The handler runs in the order it was assigned, in sequence with listeners added using `addEventListener()`.
+- Assigning a different handler replaces the value without changing its position in the listener list.
+- Assigning `null` removes it. Assigning any other non-object is treated as `null`. A non-callable object is stored and returned by the property, but never called.
+- `this` inside the handler is the object the handler was assigned to.
+- Returning `false` from the handler cancels the event, per the [event handler processing algorithm](https://html.spec.whatwg.org/multipage/webappapis.html#the-event-handler-processing-algorithm). Any other return value is ignored. Without this flag, returning `true` cancels the event and `false` is ignored. That is not standard behavior. The DOM standard ignores the return value of a listener added with `addEventListener()` entirely, and the Workers runtime applies its own `true` cancels rule to those listeners as well. This flag changes only `on<type>` handlers, so listeners added with `addEventListener()` keep the existing behavior.
+
+```js
+const controller = new AbortController();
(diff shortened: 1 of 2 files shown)


rexxars:
### CR-1127b6e4aab4 (Incorrect addEventListener return-value behavior)

Disagree on the factual claim, but rewording anyway, because the sentence is ambiguous enough that a reader could reach the same conclusion.

The claim is that return-true-cancels "applies only to the old `on<type>` dispatch path, not to `addEventListener()` listeners." In the Workers runtime it applies to both, and always has:

- There is no separate dispatch path for `on<type>`. In `EventTarget::dispatchEventImpl()` (`src/workerd/api/basics.c++`), the rule is applied to every JavaScript handler in the callbacks vector, and that vector is built from the handler table `addEventListener()` writes to. The legacy `on<type>` property lookup contributes one additional entry to the same vector.
- This predates cloudflare/workerd#7019. On `main`, `if (handle->IsTrue()) event->preventDefault();` sits in the single `JavaScriptHandler` case of that loop, under the comment "Returning true is the same as calling preventDefault() on the event."
- #7019 adds a test for it. `listenerReturningTrueStillCancels` registers a listener with `addEventListener()` that returns `true`, dispatches a cancelable event, and asserts `dispatchEvent()` returns `false` and `defaultPrevented` is `true`. It passes with the flag both enabled and disabled.

The DOM standard part of the finding is correct: a listener's return value is ignored, and never cancels. That is exactly why #7019 leaves `addEventListener()` listeners alone and changes only `on<type>` handlers. There is no standard behavior to move listeners to, so changing them would be a much wider break with nothing to align to.

So the docs sentence was describing the runtime's rule, not the standard's. Rewording it to say that outright, since "which is the rule the Workers runtime applies" reads as a claim about expected behavior:

> Without this flag, returning `true` cancels the event and `false` is ignored. That is not standard behavior. The DOM standard ignores the return value of a listener added with `addEventListener()` entirely, and the Workers runtime applies its own `true` cancels rule to those listeners as well. This flag changes only `on<type>` handlers, so listeners added with `addEventListener()` keep the existing behavior.

Pushing that now.

PR #32781, opened 2026-08-14
[Workers] Document ctx.access limitations

MattieTK:
### Summary

Documents the current `ctx.access` limitations for Worker-to-Worker HTTP and RPC calls and Workers with Static Assets. Adds cross-references from the relevant Workers pages and notes that the Cloudflare Vite plugin can generate Static Assets configuration for frameworks such as TanStack Start.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/configuration/cloudflare-access.mdx
+++ b/src/content/docs/workers/configuration/cloudflare-access.mdx
@@ -281,9 +281,11 @@ export default {
 ### `ctx.access` limitations
 
 :::note
-`ctx.access` applies only to the Worker invocation authenticated by Access. It does not currently propagate through Worker-to-Worker calls. This includes [Service Binding](/workers/runtime-apis/bindings/service-bindings/) HTTP requests and remote procedure call (RPC) invocations. The downstream Worker does not receive the caller's Access context.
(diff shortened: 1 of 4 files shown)


src/content/docs/workers/configuration/cloudflare-access.mdx
```diff
@@ -278,6 +278,16 @@ export default {
 
 </TypeScriptExample>
 
+### `ctx.access` limitations
+
+:::note
+`ctx.access` applies only to the Worker invocation authenticated by Access. It does not currently propagate through Worker-to-Worker calls. This includes [Service Binding](/workers/runtime-apis/bindings/service-bindings/) HTTP requests and remote procedure call (RPC) invocations. The downstream Worker does not receive the caller's Access context.
```
BSFishy: "does not currently propagate" implies to me that this may change in the future, however we have no plans for that. propagating the context has security implications that could lead to bypass in some configurations

BSFishy: it might also be valuable to mention that a `fetch()` subrequest with the proper service token headers or `CF_Authorization` cookie would get the appropriate ctx.access object since Access would be running again

MattieTK: Addressed in an update

MattieTK: Removed in an update

irvinebroque (approved):

PR #32857, opened 2026-08-19
chore: partially migrate deployment to cf

MattieTK:
### Summary

Partially migrates the Cloudflare Docs Worker from Wrangler to the new `cf` CLI. Production now builds through the Cloudflare Vite plugin using `cloudflare.config.ts` and deploys the generated Build Output with `cf deploy --prebuilt`.

Wrangler remains only where equivalent high-level `cf` integration is not yet available: Workers test configuration, Worker type generation, and PR preview deployment packaging. The preview workflow currently relies on:

```sh
wrangler deploy \
  --config wrangler.preview.json \
  --dispatch-namespace preview-deployments \
  --name "<script-name>"
```

`cf dispatch-namespaces scripts update <script-name>` supports low-level Workers for Platforms uploads, but `cf deploy` does not accept a dispatch namespace or script name. Equivalent support should preserve `cf deploy`'s Build Output, static asset, binding, compatibility, and observability handling while targeting the named namespace script.


kodster28 (changes requested):
Just putting a placeholder here, definitely should discuss before we'd get close to merging any of this.

PR #32906, opened 2026-08-20
NOJIRA: correct flex purge batch limits from 30 to 100

foreseaz:
Hostname, tag, and prefix how-to pages still said 30 items per request. The API default was raised to 100 in CACHE-12090, and the dashboard already shows 100.

- Update hostname, tag, and prefix pages to 100 items per request
- Point prefix limits at the shared availability table so the 30 vs 100 contradiction does not come back

### DIFF
--- a/src/content/docs/cache/how-to/purge-cache/purge_by_prefix.mdx
+++ b/src/content/docs/cache/how-to/purge-cache/purge_by_prefix.mdx
@@ -59,7 +59,7 @@ Depending on whether the upper tier has the resource or not, and whether the end
 There are several limitations regarding purge by prefix:
 
 * Path separators are limited to 31 for a prefix `(example.com/a/b/c/d/e/f/g/h/i/j/k/l/m…)`.
-* Purge requests are limited to 100 prefixes per request. For the combined hostname, tag, and prefix limit, refer to [Availability and limits](/cache/how-to/purge-cache/#availability-and-limits).
+* Purge requests are limited to 100 prefixes per request.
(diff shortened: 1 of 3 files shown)


drcaramelsyrup (approved):

src/content/docs/cache/how-to/purge-cache/purge_by_prefix.mdx
```diff
@@ -59,7 +59,7 @@ Depending on whether the upper tier has the resource or not, and whether the end
 There are several limitations regarding purge by prefix:
 
 * Path separators are limited to 31 for a prefix `(example.com/a/b/c/d/e/f/g/h/i/j/k/l/m…)`.
-* Purge requests are limited to 30 prefixes per request.
+* Purge requests are limited to 100 prefixes per request. For the combined hostname, tag, and prefix limit, refer to [Availability and limits](/cache/how-to/purge-cache/#availability-and-limits).
```
zaidoon1: The linked section doesn't state a combined limit. It has a "Max operations per request" row, but nothing there defines an operation or says the budget is shared across hostnames, tags, and prefixes, so this sends readers looking for something that isn't on the page. The page also already links to that same section further up.

The 100 applies per key type. Hostnames, tags, and prefixes can go in the same request and each is checked against its own limit, so there's no combined figure to point at. I'd drop the second sentence.

foreseaz: Thanks Zaidoon — dropped that sentence. 100 is per key type, not a combined budget.

zaidoon1 (approved):

PR #32923, opened 2026-08-21
ci: remove agent evals from Flue CI

mvvmm:
### Summary

Removes the `evals` job from `.github/workflows/flue-ci.yml`. Will come back with a better solution here soon. For now, these are getting in the way.

Local eval commands are unchanged:
- `pnpm run flue:evals` (starts dev server + runs evals + tears down)
- `pnpm --dir .flue run evals` (run against an already-running server)
- `pnpm --dir .flue run evals:json` (same, with JSON output)

Run evals locally before merging changes to agent behavior.


iruygloriani-dotcom:
Commit

kodster28 (approved):

PR #32947, opened 2026-08-24
Fixes python examples links and shortens ASGI code.

dom96:
We've recently renamed the directories in the python-workers-examples repo (in https://github.com/cloudflare/python-workers-examples/pull/95), so this renames them in the docs too. Also changes the examples to use the new asgi syntax.

### DIFF
--- a/src/content/docs/workers/languages/python/packages/fastapi.mdx
+++ b/src/content/docs/workers/languages/python/packages/fastapi.mdx
@@ -35,12 +35,8 @@ app = FastAPI()
 def read_root():
     return {"Hello": "World"}
 
-import asgi
-from workers import WorkerEntrypoint
-
-class Default(WorkerEntrypoint):
-    async def fetch(self, request):
-        return await asgi.fetch(app, request, self.env)
+from workers import asgi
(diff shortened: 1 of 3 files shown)


src/content/docs/workers/languages/python/packages/fastapi.mdx
```diff
@@ -36,11 +36,7 @@ def read_root():
     return {"Hello": "World"}
 
 import asgi
```
ryanking13: let's update this to `from workers import asgi` as well so that we can import wsgi and asgi from same workers namespace.

src/content/docs/workers/languages/python/packages/fastapi.mdx
```diff
@@ -127,16 +123,13 @@ dev = [
 Then write your worker:
 
 ```python title="src/worker.py"
-from workers import WorkerEntrypoint
 from fastapi import FastAPI, Request
 from fastapi.responses import Response
 import asgi
```
ryanking13: here too

ryanking13 (approved):

PR #32962, opened 2026-08-24
[R2] Update limits.mdx

luishcardoso:
Hello team, 

### Summary

Can we link  the docs here for further info on error 429?

https://developers.cloudflare.com/support/troubleshooting/http-status-codes/4xx-client-error/error-429/#429-too-many-requests

Thanks.

### DIFF
--- a/src/content/docs/r2/platform/limits.mdx
+++ b/src/content/docs/r2/platform/limits.mdx
@@ -30,7 +30,7 @@ import { Render } from "~/components";
 
 [^4]: The max upload size is 5 MiB less than 5 GiB, so 4.995 GiB.
 
-[^5]: Concurrent writes to the same object name (key) at a higher rate return HTTP 429 (rate limited) responses.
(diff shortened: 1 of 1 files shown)


rdimaio (commented):
Not sure we should do this, as we do not normally link every HTTP error-code mention. The current R2 docs contain no links to the generic HTTP status-code pages; references to `401`, `403`, `404`, `412`, and `429` are unlinked or point to R2-specific guidance such as the [R2 error-code reference](https://developers.cloudflare.com/r2/api/error-codes/). This matches the broader documentation: existing `429` references are unlinked in [Workflows](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/workflows/reference/limits.mdx#L52), [AI Gateway](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/ai-gateway/features/rate-limiting.mdx#L25), and [WAF](https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/waf/reference/legacy/old-rate-limiting/index.mdx#L35). 

I would be okay with doing this if we had a wider overall convention about it, but I'm worried that it might look inconsistent if we only sporadically link to error codes

PR #32975, opened 2026-08-24
[Durable Objects] Document shutdown connection errors

iglesiasbrandon:
### Summary

Two error messages that Durable Objects returns when an instance shuts down were not documented. This adds an entry for each to the troubleshooting page:

- `Connection closed: this Durable Object instance is no longer active. Reconnect or retry the request.`
- `Durable Object connection closed because the object was reset.`

The troubleshooting guidance explains automatic retries for eligible `DurableObjectStub.fetch()` calls, manual recovery for RPC methods, and WebSocket reconnection without treating internal reset text as a browser close reason.

While documenting them, this corrects existing shutdown guidance that was inaccurate or too narrow:

| Page                                | Change                                                                                                                                                                      |
| ----------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `concepts/durable-object-lifecycle` | Clarifies that instance replacement closes WebSockets and clients must reconnect. It also distinguishes instance replacement from hibernation, which keeps connections open. |
| `best-practices/websockets`         | Expands the caution beyond code deployments to cover any instance replacement.                                                                                              |

Also removes two stray Markdown code fences at the end of `best-practices/websockets.mdx`.

### Documentation checklist


src/content/docs/durable-objects/best-practices/websockets.mdx
```diff
@@ -715,9 +715,11 @@ Configure your Wrangler file with a Durable Object [binding](/durable-objects/ge
 
 A full example is available in [Build a WebSocket server](/durable-objects/examples/websocket-server/).
 
-:::caution[WebSocket disconnection on deploy]
+:::caution[WebSocket disconnection when a Durable Object restarts]
 
-Code updates disconnect all WebSockets. Deploying a new version restarts every Durable Object, which disconnects any existing connections.
```
joshthoward: I would phrase this as code updates disconnect TCP connections including WebSocket connections (direct outbound TCP connections are supported and inbound and outbound HTTP connections also use TCP).

I would also avoid using new verbs unnecessarily. DOs do not restart, they reset or are hibernated. These usually mean the same thing, but those are the two terms we have used interchangeably since before my time. 

I would not document "Including hibernating ones" because we have an MR about to be merged fixing that bug. I assume it will be released on the same timeframe as this docs change.

src/content/docs/durable-objects/concepts/durable-object-lifecycle.mdx
```diff
@@ -94,9 +94,13 @@ Durable Objects will occasionally shut down and objects are restarted, which wil
 When a Durable Object is shut down, the object instance is automatically restarted and new requests are routed to the new instance. In-flight requests are handled as follows:
 
 - **HTTP & RPC requests**: In-flight requests are allowed to finish if they do not access a Durable Object's storage.  If a request attempts to access a Durable Object's storage, it will be stopped immediately and return an error to maintain Durable Objects global uniqueness property. When the Worker runtime system is being updated, in-flight requests have up to 30 seconds to complete.
-- **WebSocket connections**: WebSocket requests are terminated automatically during shutdown. This is so that the new instance can take over the connection as soon as possible.
```
joshthoward: I like the former version better. I would update to "WebSocket requests are terminated automatically during hibernation"

src/content/docs/durable-objects/concepts/durable-object-lifecycle.mdx
```diff
@@ -94,9 +94,13 @@ Durable Objects will occasionally shut down and objects are restarted, which wil
...
+- **WebSocket connections**: When an object restarts, all of its WebSocket connections close and clients must reconnect. This includes connections that were [hibernating](/durable-objects/best-practices/websockets/#durable-objects-hibernation-websocket-api). Evicting an idle object into hibernation does not close its connections.
 - **Other invocations (email, cron)**: Other invocations are treated similarly to HTTP requests.
 
+Closing WebSocket connections promptly allows every client to reconnect to the new instance together. If connections stayed open, some clients would remain attached to the previous instance while others connected to the new one.
+
+Code that uses a [`DurableObjectStub`](/durable-objects/api/stub/) after a shutdown receives `Connection closed: this Durable Object instance is no longer active. Reconnect or retry the request.` A WebSocket closed by a shutdown reports `Durable Object connection closed because the object was reset.` Refer to [Troubleshooting](/durable-objects/observability/troubleshooting/) to handle both.
```
joshthoward: We wanted to reword `Connection closed: this Durable Object instance is no longer active. Reconnect or retry the request.` given customer feedback.

src/content/docs/durable-objects/best-practices/websockets.mdx
```diff
@@ -715,9 +715,13 @@ Configure your Wrangler file with a Durable Object [binding](/durable-objects/ge
 
 A full example is available in [Build a WebSocket server](/durable-objects/examples/websocket-server/).
 
-:::caution[WebSocket disconnection on deploy]
+:::caution[WebSocket disconnection when an instance is replaced]
```
joshthoward, line 718: I would title as just "WebSocket disconnection" the causes are many and can be expanded upon below.

src/content/docs/durable-objects/best-practices/websockets.mdx
```diff
@@ -715,9 +715,13 @@ Configure your Wrangler file with a Durable Object [binding](/durable-objects/ge
...
 
-Code updates disconnect all WebSockets. Deploying a new version restarts every Durable Object, which disconnects any existing connections.
+Cloudflare occasionally replaces a Durable Object instance. For example, this happens when you deploy new Worker code.
```
joshthoward, line 720: The fix for customer code deploys is almost out, so this is no longer a good example. I would actually remove the text below. 

I am really struggling to think through how to describe this to customers. On the happy path we handle maintenance and code deploys with zero-downtime. The only known issue as of this quarter's end is that a runtime code update will disconnect hibernatable WebSockets. This is something that we will fix next, but after that the only real situations that customers have to deal with is something like a hardware failure or a Durable Object that never hibernates and so we inevitably have to terminate connections. These are real tail events that we should find a way to document.

Sorry for the overall unhelpful review comment.

src/content/docs/durable-objects/concepts/durable-object-lifecycle.mdx
```diff
@@ -84,20 +84,26 @@ A Durable Object incurs charges only when it is **actively running in-memory**,
...
 - Workers runtime decisions on where to host objects
 
-When a Durable Object is shut down, the object instance is automatically restarted and new requests are routed to the new instance. In-flight requests are handled as follows:
+When a Durable Object instance shuts down, Cloudflare routes new requests to a new instance. In-flight requests are handled as follows:
 
 - **HTTP & RPC requests**: In-flight requests are allowed to finish if they do not access a Durable Object's storage.  If a request attempts to access a Durable Object's storage, it will be stopped immediately and return an error to maintain Durable Objects global uniqueness property. When the Worker runtime system is being updated, in-flight requests have up to 30 seconds to complete.
-- **WebSocket connections**: WebSocket requests are terminated automatically during shutdown. This is so that the new instance can take over the connection as soon as possible.
+- **WebSocket connections**: All WebSocket connections to the instance close, including [hibernating connections](/durable-objects/best-practices/websockets/#durable-objects-hibernation-websocket-api). Clients must reconnect.
```
joshthoward, line 99: All non-hibernatable WebSocket connections will be closed.

src/content/docs/durable-objects/concepts/durable-object-lifecycle.mdx
```diff
@@ -84,20 +84,26 @@ A Durable Object incurs charges only when it is **actively running in-memory**,
...
 - **Other invocations (email, cron)**: Other invocations are treated similarly to HTTP requests.
 
-It is important to ensure that any services using Durable Objects are designed to handle the possibility of a Durable Object being shut down.
+Hibernation itself does not close WebSocket connections. Cloudflare does not transfer existing connections to a new instance.
```
joshthoward, line 102: I like the version before better.

joshthoward (changes requested):

src/content/docs/durable-objects/best-practices/websockets.mdx
iglesiasbrandon: updated

src/content/docs/durable-objects/best-practices/websockets.mdx
iglesiasbrandon: i dropped the deploy example and updated. please let me know if you think this works now.

PR #32984, opened 2026-08-25
[Python] Document Django support in Python workers

ryanking13:
### Summary

This adds a new page under Python Workers packages section explaining how users can use Django framework in Python workers.

### Documentation checklist

### DIFF
--- a/src/content/docs/workers/languages/python/packages/django.mdx
+++ b/src/content/docs/workers/languages/python/packages/django.mdx
@@ -152,7 +152,9 @@ def my_view():
 
 To use D1 as a database backend, first setup your D1 database in Wrangler:
 
-```jsonc title="wrangler.jsonc"
+<WranglerConfig>
+
+```jsonc
 {
   "d1_databases": [
     {
@@ -163,6 +165,10 @@ To use D1 as a database backend, first setup your D1 database in Wrangler:
   ]
 }
 ```
+
+</WranglerConfig>
+
+
 
 Then, configure the backend in your Django settings:
 
@@ -216,7 +222,9 @@ so this environment variable is required to enable database operations.
 
 To use Durable Objects as a database backend, first setup your Durable Objects binding in Wrangler:
 
-```jsonc title="wrangler.jsonc"
+<WranglerConfig>
+
+```jsonc
 {
   "durable_objects": {
     "bindings": [
@@ -235,6 +243,8 @@ To use Durable Objects as a database backend, first setup your Durable Objects b
 }
 ```
 
+</WranglerConfig>
+
 Then, configure the backend in your Django settings:
 
 ```python title="src/app/settings.py"
@@ -252,7 +262,7 @@ import os
 
 from django.core.wsgi import get_wsgi_application
 from django_cf.db.backends.do.storage import set_storage
-from workers import WorkerEntrypoint, wsgi
+from workers import WorkerEntrypoint, DurableObject, wsgi
 
 # your Django settings module
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "app.settings")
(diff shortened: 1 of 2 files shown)


src/content/docs/workers/languages/python/packages/django.mdx
```diff
@@ -0,0 +1,282 @@
...
+When you are accessing any database backend in Django, you must set this environment variable
+to allow database operations to work in Python Workers.
+
+By default, Django does not allow running database operations inside the event loop,
+to prevent multiple event loops from accessing the same database connection simultaneously,
+which can lead to data corruption.
+
+Cloudflare database backends (D1, Durable Objects) are designed to work with event loops,
+so this environment variable is required to enable database operations.
+
```
ryanking13: Unfortunately, this is discouraged in Django but we need this to access database backends.

I think this should be okay for D1, DO backends, which is designed to work with Cloudflare workers (handling concurrent requests at the same time), but for other conventional database drivers are not guaranteed to async-safe, we need to carefully check if they are working properly.

src/content/docs/workers/languages/python/packages/django.mdx
```diff
@@ -0,0 +1,299 @@
...
+# your Django settings module
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "app.settings")
+
+application = get_asgi_application()
+
+
+class Default(WorkerEntrypoint):
+    async def fetch(self, request):
+        return await asgi.fetch(application, request, self.env)
```
dom96:
```suggestion
Default = asgi.entrypoint(app)
```

src/content/docs/workers/languages/python/packages/django.mdx
```diff
@@ -0,0 +1,299 @@
...
+Both drive the synchronous Django ORM, so serve your application through the WSGI path when you use them.
+
+:::caution[Transaction support]
+
+The D1 and Durable Object backends do not support transactions.
+
+```python
+from django.db import transaction
+
+# This will be a no-op, and rollback will not work
```
dom96, line 135: Should this actually be an error?

src/content/docs/workers/languages/python/packages/django.mdx
```diff
@@ -0,0 +1,299 @@
...
+```python title="src/index.py"
+import os
+
+from django.core.wsgi import get_wsgi_application
+from workers import WorkerEntrypoint, wsgi
+
+# your Django settings module
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "app.settings")
+
+# If you are using wsgi, you need to set this to allow database operations.
```
dom96: Does this only apply to wsgi? If so, should we discourage heavily using wsgi with D1/DO backends? perhaps even disallow it forcefully? That would mean we can avoid recommending DJANGO_ALLOW_ASYNC_UNSAFE.

dom96 (approved):

src/content/docs/workers/languages/python/packages/django.mdx
ryanking13: It applies to both wsgi and asgi. Since Python workers has a top-level event loop, django always complains about there is a running event loop.

src/content/docs/workers/languages/python/packages/django.mdx
ryanking13: Good point. It seems like django internally uses atomic / transactions when ORMs include multi queries, so raising an error by default would probably block users from using some ORM patterns.

So I guess we can keep this as-is and provide an opt-in option to raise on transaction.

irvinebroque (commented):
seems deserving of a changelog!

ryanking13:
@irvinebroque sure, done!

irvinebroque (approved):

PR #33036, opened 2026-08-26
Add @cjolowicz as CODEOWNER for log-explorer docs and changelog

cjolowicz:
Add @cjolowicz (Log Explorer engineering manager) as a CODEOWNER for the log-explorer documentation and changelog paths, so the engineering team can review doc updates.


cjolowicz:
## CR-88514bc3c66a — Conflicting duplicate CODEOWNERS pattern

Agreed. Lines 98 and 107 both assign `/src/content/changelog/log-explorer/` with different owner lists. GitHub CODEOWNERS uses the last matching pattern, so line 98 was already dead before this PR — `@dcpena` on that line was never effective.

**Fix:** Removing the dead duplicate (line 98). Line 107 is the effective entry and already includes `@cjolowicz`. Not adding `@dcpena` to the consolidated entry since it was already overridden by the pre-existing duplicate and this PR is scoped to log-explorer only.

kodster28 (approved):

PR #33064, opened 2026-08-27
[Cache] Document selected header matching

foreseaz:
### Summary

Clarifies that selected values for custom and restricted cache-key headers use case-sensitive substring matching. Includes examples of a comma-separated header value, a longer value, and a case difference.

### Documentation checklist

### DIFF
--- a/src/content/docs/cache/how-to/cache-keys.mdx
+++ b/src/content/docs/cache/how-to/cache-keys.mdx
@@ -110,6 +110,10 @@ In the **Include headers and selected values** section, you can add header names
   * `referer`
   * `user-agent`
 
+:::caution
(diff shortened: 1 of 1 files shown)


src/content/docs/cache/how-to/cache-keys.mdx
```diff
@@ -110,6 +110,10 @@ In the **Include headers and selected values** section, you can add header names
   * `referer`
   * `user-agent`
 
+:::caution
+For restricted headers, selected values use partial matching. For example, `text/plain` matches an `Accept` header value of `text/plain, foo/bar`, as well as `text/plaintext`. Choose selected values carefully, because they can match unintended header values.
```
zaidoon1: This also applies to custom headers, and the substring match is case-sensitive.

```suggestion
For custom and restricted headers, selected values use case-sensitive substring matching. For example, `text/plain` matches an `Accept` header value of `text/plain, foo/bar`. It can also match the unintended value `text/plaintext`, but not `Text/Plain`.
```

zaidoon1 (approved):

PR #33073, opened 2026-08-27
docs: False Positive Abuse Reports

shanecloudflare:
### Suggested Change

Add a section on handling false positive abuse reports, including steps to take when a legitimate website is flagged as phishing. This could include information on how to submit a false positive report, how to create a WAF custom rule to allow traffic, and best practices for narrowing the scope of skip rules to prevent potential abuse.

### DIFF
--- a/src/content/docs/bots/concepts/feedback-loop.mdx
+++ b/src/content/docs/bots/concepts/feedback-loop.mdx
@@ -211,6 +211,21 @@ If you are not certain if some traffic received an incorrect score, keep this tr
 
 We appreciate any comments you wish to leave in the description field that might help our team better understand these requests in the context of typical traffic to your domain.
 
+## Handling False Positive Abuse Reports
+
+If your legitimate website is flagged as phishing, it can cause issues with user access and trust. To handle false positive abuse reports, follow these steps:
+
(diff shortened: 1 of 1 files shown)


src/content/docs/bots/concepts/feedback-loop.mdx
```diff
@@ -211,6 +211,21 @@ If you are not certain if some traffic received an incorrect score, keep this tr
 
 We appreciate any comments you wish to leave in the description field that might help our team better understand these requests in the context of typical traffic to your domain.
 
+## Handling False Positive Abuse Reports
+
+If your legitimate website is flagged as phishing, it can cause issues with user access and trust. To handle false positive abuse reports, follow these steps:
```
migueldemoura, line 216: This is in the bots section, but talks about phishing?

migueldemoura (changes requested):

src/content/docs/bots/concepts/feedback-loop.mdx
zeinjaber: Agree with @migueldemoura!

We already have dev docs for phishing abuse reports:

- https://developers.cloudflare.com/fundamentals/reference/report-abuse/complaint-types/#phishing
- https://developers.cloudflare.com/fundamentals/reference/report-abuse/submit-report/

PR #33093, opened 2026-08-28
Document raw Containers API alongside Container class

thomasgauvin:
## Summary

Documents two ways to interact with Containers:

- Recommends the Durable Object Container API (`ctx.container`) for direct lifecycle control.
- Keeps the higher-level `Container` class as an optional convenience layer.
- Adds an API comparison and container lifecycle diagram.
- Moves both API references into the Containers docs.
- Adds raw API variants to examples, listed before `Container` class variants.
- Adds Wrangler configuration documentation.
- Updates navigation, links, and redirects.

## Testing

- `pnpm run check`
- Redirect validation
- Type-checked the new raw API examples against Workers types
- `git diff --check`


src/content/docs/containers/concepts/architecture.mdx
```diff
@@ -17,7 +17,38 @@ times when scaling up the number of concurrent container instances.
...
+        Running["Running<br/>Not ready"]
+        Ready["Ready<br/>Accepting traffic"]
+        Stopping["Stopping<br/>Stop requested"]
+        StoppedAgain["Stopped"]
+
+        Stopped --> Starting --> Running --> Ready --> Stopping --> StoppedAgain
+    end
+```
+
+A Container can only be accessed through its Durable Object. A Worker sends a request to the Durable Object, which accesses the Container through `ctx.container`.
```
benbuzbee, line 43: "Durable Object" could link to (https://developers.cloudflare.com/durable-objects/) so people & agents starting here can learn about the Durable Object execution model

irvinebroque (approved):
Love this! 🚀

thomasgauvin:
Review triage for findings on 285ad23:

- CR-9183433ea624: Keep both the navigation stub and the 301 redirect. The stub supplies the Durable Objects sidebar link to the new Containers API page; the redirect handles direct requests to the old URL. The stub is not intended as a separate content destination.
- SG-f074c72134c1 and SG-0e8254639803: These blocks are method signatures, not executable Worker examples. They now use txt fences instead of raw ts fences. TypeScriptExample generates JavaScript and its parser rejects standalone method signatures without bodies. Executable examples use TypeScriptExample.

The other findings are fixed and validated in the isolated local PR worktree. No commit or push has been made yet.

src/content/docs/containers/api/index.mdx
```diff
@@ -0,0 +1,156 @@
...
+
+interface Env {}
+
+export class MyContainer extends DurableObject<Env> {
+	private ready: Promise<void> | undefined;
+
+	constructor(ctx: DurableObjectState, env: Env) {
+		super(ctx, env);
+		ctx.blockConcurrencyWhile(() =>
+			ctx.container!.setInactivityTimeout(10 * 60 * 1000),
```
gabivlj: This example could use a `container: Container;` property that gets init in the constructor, and then users dont have to use `!` operator when accessing ctx.container.

src/content/docs/containers/concepts/architecture.mdx
```diff
@@ -91,9 +120,9 @@ should be built for the `linux/amd64` architecture, and should stay within
 
 ### Container shutdown
 
-The Container class sets [`sleepAfter`](/containers/reference/container-class/#sleepafter) to 10 minutes by default. Its default [`onActivityExpired()`](/containers/reference/container-class/#onactivityexpired) implementation signals the container to stop after that period without activity. You can change the duration or override the hook.
+With the Durable Object Container API, call [`setInactivityTimeout()`](/containers/api/durable-object-container/#setinactivitytimeout) to let the runtime stop an inactive container. You can also stop a container with [`signal()`](/containers/api/durable-object-container/#signal) or [`destroy()`](/containers/api/durable-object-container/#destroy).
```
gabivlj: There will be questions around `setinactivitytimeout`. In general, we should explain that setInactivityTimeout triggers when the Durable Object goes to sleep, use it so the container can be kept around while the DO is inactive (which is when requests do not come in).

src/content/docs/containers/examples/container-backend.mdx
```diff
@@ -138,18 +145,111 @@ Your Worker needs to be able to both serve static assets and route requests to t
...
+	ASSETS: Fetcher;
+	BACKEND: DurableObjectNamespace<Backend>;
+}
+
+export class Backend extends DurableObject<Env> {
+	private ready: Promise<void> | undefined;
+
+	constructor(ctx: DurableObjectState, env: Env) {
+		super(ctx, env);
+		ctx.blockConcurrencyWhile(() =>
```
gabivlj: In these examples I do not think we shouldbe using blockConcurrencyWhile.

src/content/docs/containers/examples/container-backend.mdx
```diff
@@ -138,18 +145,111 @@ Your Worker needs to be able to both serve static assets and route requests to t
...
+		url.host = "container";
+		const forwarded = new Request(url, request);
+		forwarded.headers.delete("host");
+		return container.getTcpPort(8080).fetch(forwarded);
+	}
+
+	private async startAndWaitForPort(): Promise<void> {
+		const container = this.ctx.container!;
+		if (!container.running) {
+			container.start();
```
gabivlj: Do we have any examples with monitor()? We should add an incentive for LLMs to use monitor() to track when the container exits (and track if there is any errors.

gabivlj:
We should add an example somewhere for LLMs to read that showcase how to keep the container "always running", and add an RPC in the DO that exposes a destroy() + destry alarm.

src/content/docs/containers/examples/env-vars-and-secrets.mdx
```diff
@@ -109,13 +119,52 @@ added to `env`.
...
+		super(ctx, env);
+
+		ctx.blockConcurrencyWhile(async () => {
+			if (!ctx.container!.running) {
+				ctx.container!.start({
+					image: env.CONTAINER_IMAGE,
+					enableInternet: true,
+					env: {
+						ENV_VAR: env.ENV_VAR,
+						WORKER_SECRET: env.WORKER_SECRET,
```
gabivlj: No blockConcurrencyWhile, I recommend just exposing a start() method on this DO as it does not seem to do anything.

In general, let's discourage starting in the constructor for LLMs

src/content/docs/containers/examples/env-vars-and-secrets.mdx
```diff
@@ -109,13 +119,52 @@ added to `env`.
...
+}
+
+export class MyContainer extends DurableObject<Env> {
+	constructor(ctx: DurableObjectState, env: Env) {
+		super(ctx, env);
+
+		ctx.blockConcurrencyWhile(async () => {
+			if (!ctx.container!.running) {
+				ctx.container!.start({
+					image: env.CONTAINER_IMAGE,
```
gabivlj: ctx.container.images?

src/content/docs/containers/examples/env-vars-and-secrets.mdx
```diff
@@ -127,16 +176,91 @@ export class MyContainer extends Container {
...
+	SECRET_STORE: SecretsStoreSecret;
+	WORKER_SECRET: string;
+}
+
+export class MyContainer extends DurableObject {
+	launch(image: string, env: Record<string, string>): void {
+		if (this.ctx.container!.running) {
+			throw new Error("Container is already running");
+		}
+		this.ctx.container!.start({ image, enableInternet: true, env });
```
gabivlj: This is a great example, let's add monitor() too?

PR #33117, opened 2026-08-30
Added new working Wrangler configurations in breakpoints.mdx

echoAwoo:
Unclear why previously supplied configuration does not work, or when it should, no clear information about context is present. The new Launch Wrangler always works for breakpoints, in Wrangler, under Vite, and Astro.

node attach Wrangler configuration appears to have environment restrictions in place that aren't clarified in documentation. node-terminal launched Wrangler configuration appears to always work, regardless of the aforementioned unknown environmental restriction.

Added small Launch Wrangler configuration to relevant documentation page to cover edge cases.


karelyvalenzuela879-lang (commented):
p207964148t20260830171412

irvinebroque:
@MattieTK

irvinebroque (approved):

PR #33123, opened 2026-08-31
[Tunnel] Add HTTPS origin troubleshooting

hugo-vicente11:
### Summary

Adds guidance for troubleshooting Cloudflare Tunnel HTTPS origins and enables local Markdown previews.

### Screenshots (optional)
image
image
image

### Documentation checklist

- [x] The change adheres to the [documentation style guide](https://developers.cloudflare.com/style-guide/).
- [ ] An issue has been opened for this new page.

PR #33125, opened 2026-08-31
fix: stop expand dialog from shrinking tall Mermaid diagrams

mvvmm:
### Summary

Fixes the expand-to-dialog view of Mermaid diagrams so taller-than-wide diagrams render at their natural size instead of shrinking dramatically.

The dialog forced the SVG into a wide, viewport-height-bounded box (`width: 100%` plus `max-height: calc(92vh - 120px)`). With the default `preserveAspectRatio: xMidYMid meet`, a diagram whose viewBox is taller than wide (e.g. a `flowchart TD` decision tree) letterboxed into that box and rendered at a small fraction of its natural scale — nodes were legible on the page but appeared scattered and tiny inside the dialog, in a mostly empty canvas.

The dialog body already scrolls (`overflow: auto`), so the height cap was removed and the clone now keeps Mermaid's natural-width `max-width` cap for taller-than-wide diagrams. Wide diagrams keep the previous stretch-to-fill behavior.

| Diagram shape | Before | After |
| --- | --- | --- |
| Taller than wide (`flowchart TD` trees, long sequence diagrams) | letterboxed at ~0.26x in a 92vh-bounded box | natural size (1x), body scrolls |
| Wider than tall (most `flowchart LR` diagrams) | stretched to fill dialog width | unchanged |

Measured on the affected dialog path: a 792x1574 viewBox diagram rendered at meet-scale 0.261 before this change and 1.0 after. Found while reviewing #33123, whose new decision-tree diagram triggered it; the behavior predates and is unrelated to the recent Mermaid version bump.

### Documentation checklist

### DIFF
--- a/src/scripts/mermaid.client.ts
+++ b/src/scripts/mermaid.client.ts
@@ -70,7 +70,14 @@ function openDiagram(container: HTMLElement) {
 
 	const svg = clone.querySelector("svg");
 	if (svg) {
-		svg.removeAttribute("style");
+		// A taller-than-wide diagram must keep its natural-width cap, otherwise
+		// the dialog stretches it to full width and the content shrinks to
+		// letterbox inside the viewport-bounded body. Wide diagrams stretch.
+		const [, , vbW = 0, vbH = 0] = (svg.getAttribute("viewBox") ?? "")
+			.trim()
+			.split(/[\s,]+/)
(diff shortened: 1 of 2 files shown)


mvvmm:
closing, fix was to revert

PR #33128, opened 2026-08-31
Warn against public IPs for Managed Networks

kennyj42:
Warns that using a public IP for a Managed Networks TLS endpoint bypasses Gateway controls. Recommends a private endpoint so its traffic stays scoped to the managed network.

Internal source: https://chat.google.com/room/AAAAUdh_23M/pDKksyodKQg


kennyj42:
@kkrum what do you think about this update? Looks to line up based on your recommendation in that chat

kennyj42:
Replaced by #33316, using the same commit from a branch in the direct repository.

PR #33132, opened 2026-08-31
ssl: add redirect loop caution to Flexible mode doc

ngayerie:
Add a `:::caution` block to the Flexible mode page warning that origins forcing HTTPS will cause redirect loops with Flexible mode, and cross-link the Enforce HTTPS connections guide from the `ERR_TOO_MANY_REDIRECTS` troubleshooting page.

DEE-3621


baubuchon-cf (approved):
Approved. The Flexible-mode redirect-loop warning is technically accurate and clearly points users toward Full or Full (strict), the ERR_TOO_MANY_REDIRECTS troubleshooting guide, and edge-level HTTPS enforcement. Internal SSL/TLS guidance corroborates the behavior and recommendation.

PR #33148, opened 2026-08-31
Clarify VNET availability propagation

kennyj42:
Documents how to request access to the gated VNET availability beta and explains the device propagation window. It also clarifies when users need to reconnect and what happens when they select a VNET.

Internal sources:

- https://chat.google.com/room/AAAAUdh_23M/7efFiUAFqFE
- https://chat.google.com/room/AAAAUdh_23M/GCGguYGxwu8
- https://chat.google.com/room/AAAAUdh_23M/Kjq1-JwwTxA


kennyj42:
@tojens-ietf / @cf-rhett is it worth calling this out as a beta still or will it ship soon for all?

cf-rhett:
> @tojens-ietf / @cf-rhett is it worth calling this out as a beta still or will it ship soon for all?

Probably a month or two before this is fully GA. Work is underway and I will get the team to remove all beta refs when this goes live

kennyj42:
ah yea, fair. I'll close this!

PR #33168, opened 2026-09-01
[Rules] Document response body inspection

mbullock1986:
## Summary

- document which Cloudflare features inspect or modify response bodies
- explain how inspection can affect streaming and time to first byte
- add a troubleshooting workflow using path-scoped Configuration Rules
- link the new guidance from the Response Body Buffering setting

## Tests

- `pnpm run check` (passes)
- `pnpm run format:core:check` (passes)
- `pnpm run build` (blocked during prerendering by the existing Astro dependency error: `cookie` does not export `parseCookie`)

### DIFF
--- a/src/content/docs/rules/configuration-rules/response-body-inspection.mdx
+++ b/src/content/docs/rules/configuration-rules/response-body-inspection.mdx
@@ -1,0 +1,64 @@
+---
+title: Response body inspection
+pcx_content_type: concept
+description: Understand when Cloudflare inspects response bodies and how inspection can affect streaming responses.
+products:
+  - rules
+sidebar:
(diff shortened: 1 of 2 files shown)


irvinebroque (approved):
Thank you! 

@mhart i think you have had to debug lot of these

mhart:
> Thank you!
> 
> @mhart i think you have had to debug lot of these

Yes I have! In many cases in the past Content-Encoding was an issue - different encodings can affect streaming due to how they buffer. However I *think* these should be all resolved at the moment. Also some clients impose a minimum number of bytes before streaming will occur (eg I think Safari is 1024 bytes).

PR #33188, opened 2026-09-02
[Cache] Document Origin Range Requests API

mbullock1986:
### Summary

- Documents the `origin_range_requests` Cache Rules API setting and its `on`, `off`, and `default` modes.
- Adds complete Rulesets API examples for turning on the feature and overriding Cloudflare defaults without changing cache eligibility.
- Adds a customer-facing range request behavior reference covering client responses, origin request alignment, origin response requirements, cache admission, file-size limits, and Cache Reserve compatibility.
- Updates the default cache behavior page to link to the detailed reference.
- Adds a Cache changelog entry for the new API control.

Related: RM-5257, CACHE-5646

### Screenshots (optional)

Not applicable.

### Documentation checklist

- [x] Is there a [changelog](https://developers.cloudflare.com/changelog/) entry ([guidelines](https://developers.cloudflare.com/style-guide/documentation-content-strategy/content-types/changelog/))?


src/content/changelog/cache/2026-09-02-origin-range-requests-rulesets-api.mdx
```diff
@@ -0,0 +1,29 @@
...
+products:
+  - cache
+date: 2026-09-02
+---
+
+The Rulesets API now supports Origin Range Requests in Cache Rules. This setting lets Cloudflare request only the needed portions of large files when clients send HTTP `Range` requests. It can reduce origin bandwidth and begin serving partial content without downloading the full file first.
+
+Set `origin_range_requests.mode` to `on`, `off`, or `default` for any traffic matched by a Cache Rule.
+
+To opt out of Cloudflare's default Origin Range Requests behavior, set the mode to `off`. The following rule turns off Origin Range Requests for all traffic without changing its cache eligibility:
```
ack-cf: it's on by default?

src/content/docs/cache/how-to/cache-rules/settings.mdx
```diff
@@ -365,6 +365,49 @@ Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#exa
 
 </Details>
 
+#### Origin range requests
+
+Origin Range Requests let Cloudflare request only the needed portions of a file from your origin. Use this setting for large files when clients request content with the HTTP `Range` header. This can reduce origin bandwidth and begin serving partial content without downloading the full file first.
+
+Your origin must support HTTP range requests. The response must also meet Cloudflare's normal [cache eligibility requirements](/cache/concepts/default-cache-behavior/).
```
gumpt: Let's also include that the Content-Encoding of the response can also disable ranging. Origin must honor `Accept-Encoding: Identity` as well.

gumpt (approved):

src/content/changelog/cache/2026-09-02-origin-range-requests-rulesets-api.mdx
```diff
@@ -0,0 +1,29 @@
+---
+title: Configure Origin Range Requests with the Rulesets API
+description: Reduce origin bandwidth for large files by controlling Origin Range Requests through the Rulesets API.
+products:
+  - cache
+date: 2026-09-02
+---
+
+The Rulesets API now supports Origin Range Requests in Cache Rules. This setting lets Cloudflare request only the needed portions of large files when clients send HTTP `Range` requests. It can reduce origin bandwidth and begin serving partial content without downloading the full file first.
```
zaidoon1: `only the needed portions` is too strong. Cloudflare can expand a client range to cache-aligned boundaries and split one client request into several single-range origin requests. The `Range` seen by the origin can therefore be wider than or different from the client's header. Please correct this here and in the settings intro because it affects origin bandwidth, request counts, logs, rate limits, and request signing.

src/content/changelog/cache/2026-09-02-origin-range-requests-rulesets-api.mdx
```diff
@@ -0,0 +1,29 @@
...
+  "action": "set_cache_settings",
+  "action_parameters": {
+    "origin_range_requests": {
+      "mode": "off"
+    }
+  }
+}
+```
+
+Your origin must support HTTP range requests. Normal cache eligibility requirements still apply.
```
zaidoon1: `must support` is too absolute. If the origin ignores `Range` and returns a complete `200`, Cloudflare can use it and may synthesize the client range, but it has to download the whole object. After Cloudflare starts composing a response from partial data, later origin requests must keep returning compatible range responses or the client response can end early. Please document both cases here and in the detailed settings.

src/content/docs/cache/how-to/cache-rules/settings.mdx
zaidoon1: This conflates use of the feature with cache admission. `origin_range_requests` does not make an otherwise ineligible request cacheable. An eligible `GET` can still use origin range fetching when the response is uncacheable, but the bytes are not stored. The maximum cacheable file size is checked against the complete object size from `Content-Range`, not the returned range size, so a small range does not bypass the limit. Please spell out this distinction here and correct the changelog summary too.

src/content/docs/cache/how-to/cache-rules/settings.mdx
```diff
@@ -365,6 +365,49 @@ Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#exa
...
+
+<Details header="API information">
+
+API configuration object name: `"origin_range_requests"`.
+
+The `mode` property accepts these values:
+
+| Value     | Behavior                                                                                                           |
+| --------- | ------------------------------------------------------------------------------------------------------------------ |
+| `on`      | Uses origin range requests for matching requests.                                                                  |
```
zaidoon1: Please add the existing Cache Reserve limitation here: Origin Range Requests are not supported with Cache Reserve. Link to `/cache/advanced-configuration/cache-reserve/#limits` so customers do not expect these settings to work together.

src/content/docs/cache/how-to/cache-rules/settings.mdx
```diff
@@ -365,6 +365,49 @@ Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#exa
...
+| `off`     | Does not send range requests to the origin. Cloudflare can still serve ranges from a complete cached file.         |
```
zaidoon1: `off` only disables range requests generated by Origin Range Requests. Cloudflare can still forward the original `Range` header from the client when a Cache Rule sets **Cache eligibility** to **Bypass cache**, or when the URL is not cache eligible under default cache behavior, such as an HTML page that no Cache Rule makes eligible for cache. On a cacheable miss, Cloudflare may instead remove `Range` and fetch the full object. Please update this row so it does not promise that the origin will never receive `Range`, and keep the cached-file sentence.

src/content/docs/cache/how-to/cache-rules/settings.mdx
```diff
@@ -365,6 +365,49 @@ Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#exa
...
+| `default` | Uses Cloudflare's default file extension list for Origin Range Requests. Omitting this setting has the same effect. |
```
zaidoon1: `default` does not currently enable Origin Range Requests for any extension. It makes no change today, and omitting the setting also makes no change. Not sure how we want to make it clear that in the future, it will be but not right now

src/content/docs/cache/how-to/cache-rules/settings.mdx
```diff
@@ -365,6 +365,49 @@ Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#exa
...
+
+```json title="Turn off default Origin Range Requests"
+"action_parameters": {
+  "origin_range_requests": {
+    "mode": "off"
+  }
+}
+```
+
+Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#example-turn-off-default-origin-range-requests) for a complete opt-out example.
```
zaidoon1, line 411: This fragment has no target. The `<Details>` block for the example does not have an `id`. Add `id="example-turn-off-default-origin-range-requests"` to the target or link to the existing `#example-requests` anchor.

src/content/docs/cache/how-to/cache-rules/settings.mdx
```diff
@@ -365,6 +365,49 @@ Refer to [Create a cache rule via API](/cache/how-to/cache-rules/create-api/#exa
...
+
+</Details>
```
zaidoon1, line 413: Please add a linked customer behavior reference. The settings table is not enough for customers to predict the response. Use the status names as well as the numbers, and cover these cases:

- **Which requests use it.** Cloudflare can combine several cache reads or origin fetches only for a `GET` that is eligible for caching when the request arrives. The response may later prove uncacheable and not be stored. `HEAD` and other methods do not use this path. A `HEAD` with `Range` can return `200 OK`, `206 Partial Content`, `304 Not Modified`, or `416 Range Not Satisfiable`, depending on the range and conditions. It never has a body. A `206` includes `Content-Range`, and `If-Range` is ignored for `HEAD`.
- **No `Range` header.** A successful, unconditional `GET` returns the complete file as `200 OK`. On a cold miss, the first origin request also has no `Range`. If only part of the file is cached, Cloudflare may request the missing bytes from the origin while still returning one complete `200` response to the client.
- **One valid range.** For a successful response, `bytes=100-199`, `bytes=100-`, and `bytes=-100` return `206 Partial Content`. Cloudflare shortens an end past the end of the file. A suffix larger than the file returns the whole file as `206`.
- **When Cloudflare returns `416`.** `bytes=`, `bytes=-0`, inverted bounds, a valid range starting at or past the end of the file, and a valid range against an empty uncompressed file return `416 Range Not Satisfiable`. Overlapping or out of order ranges also return `416`. A `416` generated by Cloudflare has an empty body and `Content-Range: bytes */<file-size>`.
- **When Cloudflare ignores `Range`.** Cloudflare processes the request as a non-range request when the syntax cannot be parsed, the unit is not `bytes`, a number is too large to parse, or the header contains more than 300 ranges. This normally returns the complete file as `200`, but conditional requests can return `304` and origin errors still apply.
- **Multiple ranges.** Two through exactly 300 valid, ascending, non-overlapping ranges return a multipart `206`. Cloudflare omits individual ranges that cannot be satisfied. The response remains multipart if only one range survives. If none survive, Cloudflare returns `416`.
- **Conditional `GET` requests.** `If-None-Match` and `If-Modified-Since` take precedence and can return `304 Not Modified`. Otherwise, an `If-Range` value that exactly matches the cached `ETag` or `Last-Modified` value keeps the requested range behavior. A mismatch ignores `Range` and uses the normal full-response path.
- **Decompression.** If Cloudflare must decompress a complete encoded response, it ignores `Range` and sends the complete uncompressed body as `200`. This applies to cached responses and responses just received from the origin. A ranged request for an empty encoded response returns an empty `200`, not `416`.
- **When the origin does not follow the request.** If the origin ignores a generated `Range` and returns a complete `200`, Cloudflare can use it but must download the whole file. It may then return the requested range to the client. If the origin returns an encoded `206` despite being asked for uncompressed content, Cloudflare does not cache it. Before the client response starts, Cloudflare may send that one `206` without caching. After the response starts, Cloudflare must end it early, so the client receives fewer bytes than promised. Other incompatible responses during a multi-request response can also end it early.

zaidoon1 (commented):
I left inline notes for the customer-visible behavior and origin requirements that still need coverage.

src/content/docs/cache/how-to/cache-rules/settings.mdx
zaidoon1: Agreed. Please make the full origin contract explicit. Cloudflare adds `Accept-Encoding: identity` to every origin request generated by Origin Range Requests, including a full-file `GET` handled by the feature. This does not disable normal client-facing compression. A cacheable `206` must be unencoded, include a concrete `Content-Range` whose start matches the `Range` Cloudflare sent and whose end is the requested end or EOF, whichever comes first, include the complete size, include a numeric `Content-Length` equal to the returned interval length, and omit `Transfer-Encoding`. `ETag` is optional, but while Cloudflare assembles the same cached representation, its presence and exact value must stay consistent across fills, as must the total size. Violations can prevent caching or end a composed response early.

mbullock1986: Addressed in 1e7c928e72. The new range request behavior reference documents `Accept-Encoding: identity` on every generated origin request, the complete cacheable `206 Partial Content` contract, validator and size consistency, and the encoded/incompatible response behavior.

src/content/changelog/cache/2026-09-02-origin-range-requests-rulesets-api.mdx
mbullock1986: Addressed in 1e7c928e72. The settings and changelog now describe cache-aligned ranges, including expansion and multiple single-range origin requests. The new reference also calls out implications for logs, request limits, rate limits, and request signing.

src/content/changelog/cache/2026-09-02-origin-range-requests-rulesets-api.mdx
mbullock1986: Addressed in 1e7c928e72. The absolute origin-support requirement is removed. Both the settings and changelog describe the complete `200 OK` fallback, while the new reference explains when incompatible or encoded partial responses prevent caching or end a composed response early.

src/content/docs/cache/how-to/cache-rules/settings.mdx
mbullock1986: Addressed in 1e7c928e72. The docs now separate origin range fetching from cache admission, state that an uncacheable response is not stored, and explain that the maximum cacheable file size applies to the complete size from `Content-Range`.

src/content/docs/cache/how-to/cache-rules/settings.mdx
mbullock1986: Addressed in 1e7c928e72. Both the settings page and the new range request reference state that Origin Range Requests are not supported with Cache Reserve and link directly to the Cache Reserve limits.

src/content/docs/cache/how-to/cache-rules/settings.mdx
mbullock1986: Addressed in 1e7c928e72. The target `Details` component now has `id="example-turn-off-default-origin-range-requests"`, matching the link fragment.

src/content/docs/cache/how-to/cache-rules/settings.mdx
mbullock1986: Addressed in 1e7c928e72. The `off` documentation now says it disables only ranges generated by Origin Range Requests. It covers complete cached-file ranges, complete-object fetches on cacheable misses, and forwarding the original client `Range` when caching is bypassed or ineligible.

src/content/docs/cache/how-to/cache-rules/settings.mdx
mbullock1986: Addressed in 1e7c928e72. `default` now means no Cache Rule override and explicitly states that no file extensions are currently turned on by the platform default. The opt-out example is framed as overriding Cloudflare default behavior rather than disabling active default extensions.

src/content/docs/cache/how-to/cache-rules/settings.mdx
mbullock1986: Addressed in 1e7c928e72 with a dedicated `/cache/reference/range-requests/` page linked from the setting, changelog, and default cache behavior. It covers eligible methods, `HEAD`, complete responses, single and multipart ranges, `416 Range Not Satisfiable`, ignored ranges, conditional requests, decompression, origin request alignment, the full partial-response contract, incompatible responses, cache admission, file-size limits, and Cache Reserve.

mbullock1986:
@zaidoon1 Thanks for the detailed review. I addressed every thread in the squashed commit and added a dedicated customer-facing range request behavior reference covering the full client and origin contract. All review threads are resolved and CI is green. Ready for your re-review.

src/content/docs/cache/reference/range-requests.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+## Request eligibility
+
+Cloudflare uses the partial-file path only for a `GET` request that is eligible for caching when the request arrives. The origin response may later prove uncacheable. In that case, Cloudflare can still use origin range fetching, but does not store the response bytes.
+
+`HEAD` requests and other methods do not use this path.
+
+### HEAD requests
+
+A `HEAD` request containing `Range` can return `200 OK`, `206 Partial Content`, `304 Not Modified`, or `416 Range Not Satisfiable`. A `HEAD` response never has a body. A `206 Partial Content` response includes `Content-Range`, and Cloudflare ignores `If-Range` for `HEAD` requests.
```
zaidoon1: `If-Range` is not always ignored for `HEAD`. When Cloudflare serves a complete cached file without Origin Range Requests, a matching validator can keep a `206 Partial Content` response, while a mismatch returns `200 OK`. Please qualify this sentence.

src/content/docs/cache/reference/range-requests.mdx
```diff
@@ -0,0 +1,111 @@
...
+### Unsatisfiable ranges
+
+Cloudflare returns `416 Range Not Satisfiable` for:
+
+- An empty range, such as `bytes=`
+- A zero-length suffix, such as `bytes=-0`
+- Inverted range bounds
+- A range starting at or beyond the end of the file
+- A range against an empty, uncompressed file
+- Overlapping or out-of-order ranges
```
zaidoon1: Overlapping or out-of-order ranges do not always return `416 Range Not Satisfiable`. When Cloudflare serves a complete cached file without Origin Range Requests, it ignores the range and normally returns the full file as `200 OK`. Please qualify this bullet.

src/content/docs/cache/reference/range-requests.mdx
```diff
@@ -0,0 +1,111 @@
...
+A Cloudflare-generated `416 Range Not Satisfiable` response has an empty body. It includes `Content-Range: bytes */<FILE_SIZE>`.
+
+### Ignored Range headers
+
+Cloudflare processes the request as a non-range request when:
+
+- The range syntax cannot be parsed
+- The range unit is not `bytes`
+- A number is too large to parse
+- The header contains more than 300 ranges
```
zaidoon1: When Cloudflare serves a complete cached file without Origin Range Requests, a header containing exactly 300 ranges is also ignored. Please qualify this limit instead of stating `more than 300` as universal.

src/content/docs/cache/reference/range-requests.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+This normally returns the complete file as `200 OK`. Conditional requests can return `304 Not Modified`, and origin errors still apply.
+
+### Multiple ranges
+
+Between two and 300 valid, ascending, non-overlapping ranges return a multipart `206 Partial Content` response. Cloudflare omits individual ranges that cannot be satisfied. The response remains multipart if only one range remains. If no ranges remain, Cloudflare returns `416 Range Not Satisfiable`.
```
zaidoon1: When Cloudflare serves a complete cached file without Origin Range Requests, one remaining satisfiable range produces a single-range `206 Partial Content` response rather than multipart. Please qualify this paragraph.

ack-cf (approved):

src/content/changelog/cache/2026-09-02-origin-range-requests-rulesets-api.mdx
zaidoon1: eventually, default = on for some specific file extensions but that is not happening now

drcaramelsyrup (approved):

src/content/docs/cache/reference/range-requests.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+For Cloudflare to cache an origin `206 Partial Content` response, the response must meet these requirements:
+
+| Response field | Requirement |
+| --- | --- |
+| `Content-Encoding` | The response must be unencoded. |
+| `Content-Range` | Include a concrete range. Its start must match Cloudflare's request. Its end must match the requested end or the end of the file, whichever comes first. Include the complete file size. |
+| `Content-Length` | Include a numeric value equal to the returned interval length. |
+| `Transfer-Encoding` | Omit this header. |
+| `ETag` | Optional. If present, its presence and exact value must remain consistent across fills for the same cached representation. |
```
drcaramelsyrup:
```suggestion
| `ETag` | Optional. If present, its presence and exact value must remain consistent across responses for the same cached representation. |
```

src/content/docs/cache/reference/range-requests.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+The complete file size must also remain consistent across fills. Violations can prevent caching or end a composed response early.
```
drcaramelsyrup:
```suggestion
The complete file size must also remain consistent across responses. Violations can prevent caching or end responses served to the client early.
```

zaidoon1 (approved):

PR #33190, opened 2026-09-02
[SSL] Link to alternative CT search tool

rellem:
### Summary

crt.sh is a great resource, but it seems to have server issues quite often so a backup link could be useful.

Disclosure: I'm the founder of CertObserver.

### Documentation checklist


baubuchon-cf:
Hey rellem, thanks for the PR and for calling out the CertObserver connection up front.. I appreciate that.

Going to close this one though. This isn't a reflection on CertObserver, we just don't want to start adding vendor-specific links to our docs on a case by case basis.

If crt.sh downtime turns into a real recurring problem, we can revisit with more general wording. Thanks for contributing.

rellem:
I understand. Thanks for letting me know!

PR #33218, opened 2026-09-04
[SSL] document backup certificate race condition known issue

ngayerie:
Add a troubleshooting section to the backup certificates page documenting a known race condition where a backup certificate can be deleted immediately after issuance.

DEE-3829

When Universal SSL is disabled and re-enabled in quick succession, certificate processing jobs are not guaranteed to run in submission order. This can cause a newly issued backup certificate to be deleted before it becomes active. The fix documents the issue and the workaround (wait between toggling, use Advanced Certificate Manager for uninterrupted coverage).

### DIFF
--- a/src/content/docs/ssl/edge-certificates/backup-certificates.mdx
+++ b/src/content/docs/ssl/edge-certificates/backup-certificates.mdx
@@ -27,3 +27,21 @@ For additional details, refer to the [introductory blog post](https://blog.cloud
 ## Opt out
 
 Enterprise customers can request to opt out of backup certificates by opening a support case. Opting out removes the backup-certificate redundancy for your domain.
+
+## Troubleshooting
+
+### Backup certificate deleted after turning Universal SSL back on
+
+After you turn off and quickly turn Universal SSL back on, your domain may end up without a backup certificate.
+
+When Universal SSL is toggled off and on in quick succession, certificate processing jobs are not guaranteed to run in the order they were submitted. This race condition can cause a newly issued backup certificate to be deleted before it becomes active.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/ssl/edge-certificates/backup-certificates.mdx
```diff
@@ -23,3 +23,21 @@ For additional details, refer to the [introductory blog post](https://blog.cloud
...
+
+### Backup certificate deleted after turning Universal SSL back on
+
+After you turn off and quickly turn Universal SSL back on, your domain may end up without a backup certificate.
+
+When Universal SSL is toggled off and on in quick succession, certificate processing jobs are not guaranteed to run in the order they were submitted. This race condition can cause a newly issued backup certificate to be deleted before it becomes active.
+
+To recover your backup certificate:
+
+1. Turn Universal SSL off again.
```
baubuchon-cf, line 41: Can we move the TLS continuity requirement before the recovery steps? Turning off Universal SSL can cause new TLS connections to fail unless another valid certificate is already active. “Ordering” an Advanced certificate does not guarantee it is active, and Custom certificates are also supported.

Suggested text:

Before turning off Universal SSL, make sure another valid Advanced or Custom certificate is active and covers the required hostnames. Otherwise, new TLS connections will fail.

Please also confirm that “several minutes” is the validated cooldown from DEE-3829.

baubuchon-cf (approved):
LGTM. Verified the race condition and recovery steps against the underlying escalation — matches the confirmed root cause. ACM workaround checks out too.

PR #33225, opened 2026-09-04
[SSL] document Enterprise opt-out process for backup certificates

ngayerie:
Adds an `## Opt out` section to the backup certificates page explaining that Enterprise customers need to contact Cloudflare Support to disable backup certificates.

DEE-3827

### DIFF
--- a/src/content/docs/ssl/edge-certificates/backup-certificates.mdx
+++ b/src/content/docs/ssl/edge-certificates/backup-certificates.mdx
@@ -23,3 +23,7 @@ For additional details, refer to the [introductory blog post](https://blog.cloud
 | - | - | - | - | - |
 | Availability | Yes | Yes | Yes | Yes |
 | Can opt out? | No | No | No | Yes |
+
+## Opt out
+
+Enterprise customers can request to opt out of backup certificates by opening a support case. Opting out removes the backup-certificate redundancy for your domain.


src/content/docs/ssl/edge-certificates/backup-certificates.mdx
```diff
@@ -19,7 +19,11 @@ For additional details, refer to the [introductory blog post](https://blog.cloud
...
-|  | Free | Pro | Business | Enterprise |
-| - | - | - | - | - |
-| Availability | Yes | Yes | Yes | Yes |
-| Can opt out? | No | No | No | Yes |
+|              | Free | Pro | Business | Enterprise |
+| ------------ | ---- | --- | -------- | ---------- |
+| Availability | Yes  | Yes | Yes      | Yes        |
+| Can opt out? | No   | No  | No       | Yes        |
+
+## Opt out
```
baubuchon-cf, line 27: This addition makes sense. I do think we should tighten the wording a bit. “Contact Cloudflare Support” feels broader than what we probably mean here, especially since the linked Support page is generic and says Support cannot make configuration or account changes on a customer’s behalf. If the intended path is to open a support case to request opt-out, I’d suggest saying that directly. For example: “Enterprise customers can request backup certificate opt-out by opening a support case.” We could also add a short note that opting out removes backup-certificate redundancy.

ngayerie: [AI] Good point — I've tightened the wording to say "request to opt out by opening a support case" and added a note that opting out removes backup-certificate redundancy. Updated in commit 508ebd4.

baubuchon-cf (approved):
LGTM. Tightened wording addresses the earlier concern about the generic "Contact Support" phrasing and link. Enterprise-only opt-out and the redundancy note both check out. Approving.

PR #33285, opened 2026-09-07
[DNS] document Google Workspace DNS record setup

ngayerie:
Add a new how-to page covering the DNS records needed to use a Cloudflare-managed domain with Google Workspace: domain ownership verification (TXT), MX records, SPF, DKIM, DMARC, and a multi-domain zone-file import tip. Includes callouts for the Email Routing MX conflict and the "domain already in use" Google-side error.

DEE-3619

### DIFF
--- a/src/content/docs/dns/manage-dns-records/how-to/set-up-google-workspace.mdx
+++ b/src/content/docs/dns/manage-dns-records/how-to/set-up-google-workspace.mdx
@@ -1,0 +1,156 @@
+---
+pcx_content_type: how-to
+description: Add the DNS records required to verify and use your domain with Google Workspace.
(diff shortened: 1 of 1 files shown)


ngayerie:
[AI] Addressing CR-3e5909df36f7: the reviewer is correct — MX and TXT records in Cloudflare are always DNS-only and cannot be proxied, so the note was based on a false premise. Fixing the page to state this factually and replacing the faulty troubleshooting item with a real cause (propagation delay).

ngayerie:
Hi @hannes-cf 
Would it be possible for you to review this PR?
Thanks!

chreo (approved):

PR #33292, opened 2026-09-08
[Secondary DNS] Proxy traffic hidden primary

tomislavcf:
Hidden Primary is only recommended model when proxying traffic, therefore for full traffic coverage through Cloudflare (required for Spectrum, WAF, CDN), use a hidden primary configuration where your primary DNS server is not listed at the registrar. If both your primary and Cloudflare nameservers are listed at the registrar, resolvers will split queries between them and ~50% of traffic will bypass Cloudflare.

### DIFF
--- a/src/content/docs/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic.mdx
+++ b/src/content/docs/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic.mdx
@@ -23,6 +23,12 @@ Only A, AAAA, and CNAME records can be proxied.
 
 :::
 
+:::note
+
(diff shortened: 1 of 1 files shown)


src/content/docs/dns/zone-setups/zone-transfers/cloudflare-as-secondary/proxy-traffic.mdx
```diff
@@ -23,6 +23,12 @@ Only A, AAAA, and CNAME records can be proxied.
 
 :::
 
+:::note
+
+For full traffic coverage through Cloudflare (required for Spectrum, WAF, CDN), use a hidden primary configuration where your primary DNS server is not listed at the registrar. If both your primary and Cloudflare nameservers are listed at the registrar, resolvers will split queries between them and ~50% of traffic will bypass Cloudflare.
```
chreo, line 28: The note is good but should be improved. This does not seem distinct enough for me with regards to what the doc page is about.

chreo (changes requested):

PR #33297, opened 2026-09-08
[CF4SaaS] Add Always Online compatibility note for O2O setups

mbullock1986:
Adds Always Online to the O2O product compatibility table, noting that it does not trigger on the eyeball zone and is not recommended on the SaaS provider zone because the eyeball zone may cache Internet Archive responses.


baubuchon-cf (approved):
LGTM on content — verified the Always Online trigger behavior against the product docs (activates only on unreachable origin, and the SaaS provider zone is always network-reachable from the eyeball zone in O2O) and the caching-caveat framing matches the existing pattern used for the Cache row in this table. Row is correctly alphabetized between Access and API Shield.

Rebase is in and the branch is up to date. The Build job failure is a transient GitHub Actions artifact-finalize error (403 from an intermediary, unrelated to this diff) — re-running that job should clear it. Approving.

PR #33306, opened 2026-09-08
[Cache] Add missing purge edge-case documentation

ngayerie:
Add three missing edge cases to the cache purge documentation.

DEE-1922

**index.mdx** — note that a 200 response from the purge API does not confirm eviction. The API returns 200 even when the targeted URL was not in cache, which is a frequent source of confusion in support cases.

**purge-by-single-file.mdx** — new "Redirect responses" subsection explaining that purging a URL that returns a 301/302 removes the cached redirect response, not the destination resource. Customers need to purge the final destination URL.

**purge-everything.mdx** — promote the existing performance warning to a `:::caution` callout and recommend targeted purge methods first to reduce origin load.

### DIFF
--- a/src/content/docs/cache/how-to/purge-cache/index.mdx
+++ b/src/content/docs/cache/how-to/purge-cache/index.mdx
@@ -17,7 +17,7 @@ Cloudflare's Instant Purge ensures that updates to your content are reflected im
 To purge cached content using the Cloudflare API, refer to [Purge Cached Content](/api/resources/cache/methods/purge/).
 
 :::note
-A successful purge request returns `HTTP 200`. This indicates that Cloudflare received the request — it does not confirm that Cloudflare cached the targeted content or evicted any content. To verify a purge, request the asset after purging and confirm that `CF-Cache-Status` is no longer `HIT`.
+A successful purge request returns `HTTP 200`. This indicates that Cloudflare received the request — it does not confirm that Cloudflare cached the targeted content or evicted any content. To verify a purge, request the asset after purging and confirm that [`CF-Cache-Status`](/cache/concepts/cache-responses/) is no longer `HIT`.
 :::
 
 :::note
--- a/src/content/docs/cache/how-to/purge-cache/index.mdx
+++ b/src/content/docs/cache/how-to/purge-cache/index.mdx
@@ -17,7 +17,7 @@ Cloudflare's Instant Purge ensures that updates to your content are reflected im
 To purge cached content using the Cloudflare API, refer to [Purge Cached Content](/api/resources/cache/methods/purge/).
 
 :::note
-A successful purge request returns `HTTP 200`. This indicates the request was received — it does not confirm that the targeted content was cached or that any content was evicted. To verify a purge, request the asset after purging and confirm that `CF-Cache-Status` is no longer `HIT`.
+A successful purge request returns `HTTP 200`. This indicates that Cloudflare received the request — it does not confirm that Cloudflare cached the targeted content or evicted any content. To verify a purge, request the asset after purging and confirm that `CF-Cache-Status` is no longer `HIT`.
 :::
 
 :::note
--- a/src/content/docs/cache/how-to/purge-cache/index.mdx
+++ b/src/content/docs/cache/how-to/purge-cache/index.mdx
@@ -17,7 +17,7 @@ Cloudflare's Instant Purge ensures that updates to your content are reflected im
 To purge cached content using the Cloudflare API, refer to [Purge Cached Content](/api/resources/cache/methods/purge/).
 
 :::note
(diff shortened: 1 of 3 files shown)


ngayerie:
[AI] Addressing CR-83b35cfca815, SG-75a49e9ab491, and SG-406b148f414d in the next commit.

Note on CR-83b35cfca815: agreed that the current `curl -v … | grep -i location` is incomplete for redirect chains. Not using the bot's suggested `curl -sIL` because `-I` sends a HEAD request — we moved away from `-I` specifically to avoid unexpected behavior differences from GET. Using `curl -Ls -o /dev/null -w "%{url_effective}\n"` instead: follows all redirects with GET, outputs the final URL directly, no grep ambiguity.

ngayerie:
[AI] Addressing both findings in the next commit.

**CR-0ff315b2aa28:** The text/command mismatch is real — fixing by rewording the surrounding text to match the command, rather than changing the command. The `curl -Ls -o /dev/null -w "%{url_effective}"` approach (follows all redirects with GET, outputs the final destination URL) is more useful for purging than printing the raw `Location` header, which may be relative and only covers the first hop. Not using the bot's suggestion of `curl -sI | grep location` because `-I` sends a HEAD request, which was the original issue.

**SG-f41779aeedad:** Applying the bot's suggested active voice rewrite verbatim.

src/content/docs/cache/how-to/purge-cache/index.mdx
```diff
@@ -16,6 +16,10 @@ Cloudflare's Instant Purge ensures that updates to your content are reflected im
 
 To purge cached content using the Cloudflare API, refer to [Purge Cached Content](/api/resources/cache/methods/purge/).
 
+:::note
+A successful purge request returns `HTTP 200`. This indicates that Cloudflare received the request — it does not confirm that Cloudflare cached the targeted content or evicted any content. To verify a purge, request the asset after purging and confirm that `CF-Cache-Status` is no longer `HIT`.
+:::
```
mbullock1986, line 21: Is there a link how to get this / headers?

src/content/docs/cache/how-to/purge-cache/purge-everything.mdx
```diff
@@ -13,7 +13,13 @@ import { DashButton } from "~/components"
 
 To maintain optimal site performance, Cloudflare strongly recommends using single-file (by URL) purging instead of a complete cache purge.
 
-Purging everything instantly clears all resources from your CDN cache in all Cloudflare data centers. Each new request for a purged resource returns to your origin server to validate the resource. If Cloudflare cannot validate the resource, Cloudflare fetches the latest version from the origin server and replaces the cached version. When a site with heavy traffic contains a lot of assets, requests to your origin server can increase substantially and result in slow site performance.
+Purging everything instantly clears all resources from your CDN cache in all Cloudflare data centers. Each new request for a purged resource returns to your origin server to validate the resource. If Cloudflare cannot validate the resource, Cloudflare fetches the latest version from the origin server and replaces the cached version.
```
mbullock1986: unsure what the last sentence means?

src/content/docs/cache/how-to/purge-cache/purge-everything.mdx
```diff
@@ -13,7 +13,13 @@ import { DashButton } from "~/components"
...
+
+:::caution
+When you purge everything, all cached content for your zone is removed at once. Every subsequent request must be served from your origin until the cache is repopulated. On high-traffic sites with many assets, this can cause a large spike in origin requests and may significantly slow down your site or overload your origin server.
```
mbullock1986, line 19: can we add a link details on Tiered Caching to help here.

ngayerie: [AI] Added in f2b050e: "If your zone uses [Tiered Cache](/cache/how-to/tiered-cache/), cache repopulation also propagates across cache tiers, which can further increase origin load during this period."

src/content/docs/cache/how-to/purge-cache/index.mdx
ngayerie: [AI] Added a link to [`CF-Cache-Status`](/cache/concepts/cache-responses/) in f2b050e so readers can find the header documentation and understand how to inspect it.

src/content/docs/cache/how-to/purge-cache/purge-everything.mdx
ngayerie: [AI] This is pre-existing text, not part of our change. You're right that "validate the resource" is ambiguous. Rewritten in f2b050e to: "If the cached version is no longer valid, Cloudflare fetches the latest version from your origin server and caches it."

mbullock1986 (approved):

PR #33130, opened 2026-08-31
Document BYOIP dedicated egress assignment

kennyj42:
Clarifies the setup required after general BYOIP onboarding. Customers must ask their account team to assign the address to their Zero Trust organization before it appears in egress policy options.

Internal source: https://chat.google.com/room/AAAAUdh_23M/4rFswCpeNko

PR #33315, opened 2026-09-08
Document BYOIP dedicated egress assignment

kennyj42:
Clarifies the setup required after general BYOIP onboarding. Customers must ask their account team to assign the address to their Zero Trust organization before it appears in egress policy options.

Internal source: https://chat.google.com/room/AAAAUdh_23M/4rFswCpeNko


kennyj42:
Supersedes #33130. This replacement uses a branch in the direct repository so the full docs workflow is available.

kodster28 (approved):

PR #33316, opened 2026-09-08
Warn against public IPs for Managed Networks

kennyj42:
Warns that using a public IP for a Managed Networks TLS endpoint bypasses Gateway controls. Recommends a private endpoint so its traffic stays scoped to the managed network.

Internal source: https://chat.google.com/room/AAAAUdh_23M/pDKksyodKQg


kennyj42:
Supersedes #33128. This replacement uses a branch in the direct repository so the full docs workflow is available.

kodster28 (approved):

PR #33323, opened 2026-09-08
Document approval for infrastructure applications

kennyj42:
## Summary

- add infrastructure applications to the temporary authentication guide
- link purpose justification and approval setup from the infrastructure application workflow
- clarify that both controls are configured on an Allow policy

## Testing

- verified both internal policy links


kennyj42:
Supersedes #32281. This replacement uses a branch in the direct repository so the full docs workflow is available.

kodster28 (approved):

PR #33324, opened 2026-09-08
Document Workers and Pages domains for MCP portals

kennyj42:
## Summary

- document workers.dev and pages.dev as supported MCP portal domains
- clarify that an active zone is no longer required when using either managed domain
- update the portal creation steps to match the dashboard domain selector

## Testing

- verified the workers.dev documentation link


kennyj42:
Supersedes #32280. This replacement uses a branch in the direct repository so the full docs workflow is available.

irvinebroque (approved):

PR #33362, opened 2026-09-10
[Workers] Request redirect mode: only follow and manual are supported

stayinalive181:
### Summary

Correction to the Workers `Request` runtime API page (`src/content/docs/workers/runtime-apis/request.mdx`).

The page currently lists `error` as a valid redirect mode in two places: the `redirect` option of `RequestInit` and the read-only `redirect` property. Workers does not implement that mode. workerd rejects it while parsing the `RequestInit` dictionary, in [`src/workerd/api/http.c++`](https://github.com/cloudflare/workerd/blob/94e65452649f6772fe5e79dc7cf6fa3d26f3decc/src/workerd/api/http.c%2B%2B#L482-L487):

```
TypeError: Invalid redirect value, must be one of "follow" or "manual" ("error" won't be implemented since it does not make sense at the edge; use "manual" and check the response status code).
```

Because the exception is thrown when the `Request` is constructed, `fetch(url, { redirect: "error" })` fails before any request is sent, for every URL. Reproduced with `wrangler dev` and on deployed Workers; the identical failure for all destinations makes it look like a network problem, which is how we lost a day to it before finding the message above.

Both descriptions now say `follow` or `manual`, and explain that `error` throws a `TypeError` at construction and that `manual` plus a status-code check is the equivalent. A stray double space on the first line was removed at the same time.

### Documentation checklist

- [x] No changelog entry: this corrects the description of existing behaviour rather than announcing a change.
- [x] No files were renamed or moved, so no redirects are needed.


irvinebroque:
@stayinalive181 i am curious what you were building / what you were trying to make work when you ran into this

irvinebroque (commented):
@jasnell for review

stayinalive181:
Thanks for taking a look. We run a control plane on Workers that registers GPU inference nodes and delivers webhooks. Those outbound calls used `fetch(url, { redirect: "error" })`, since we never want to follow a redirect there. Every call threw before a request was sent, for every destination (`https://example.com` included), so it looked like an outbound networking problem rather than an invalid argument, and it took us about a day to get to the TypeError. The Request page listing `error` as a valid mode is what made us trust the option. We now use `redirect: "manual"` and check the status code, as the workerd error message suggests.

PR #33386, opened 2026-09-10
ci: run preview deploys concurrently

zeke:
This PR makes the two `wrangler deploy` calls in the `Deploy Preview` job's "Deploy to Cloudflare Workers" step run concurrently instead of sequentially.

Each PR preview is deployed twice, once under the short commit SHA and once under the branch-slug name, to the same dispatch namespace. These deploys are independent of each other, but currently run back-to-back in a single shell step, roughly doubling the time of that step (observed ~5m18s across two deploys on a recent run).

This backgrounds both `wrangler deploy` invocations and waits on each, so the step fails the same way as before if either deploy fails, but the wall-clock time should be close to that of a single deploy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,8 +285,16 @@ jobs:
           echo "branch_slug=${BRANCH_SLUG}" >> "$GITHUB_OUTPUT"
           echo "short_sha=${SHORT_SHA}" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


mvvmm:
I'm fine w the review finding.

mvvmm (approved):

zeke:
Thanks for tending to this one, @mvvmm 🙌🏼

PR #33414, opened 2026-09-14
docs(ai-gateway): make the analytics counting basis explicit (requests vs attempts vs cache hits)

### LINKED ISSUE #32983: Semantics question: in AI Gateway analytics, is a retried provider call one request or two — and do cache hits count?
**Background.** AI Gateway logs every proxied request with status, token usage, cost, and duration, and can retry timed-out provider calls (request-handling docs), fall back across providers (Universal Endpoint), and serve responses from cache.

**Question or gap.** Scenario: 1 client prompt → provider 429/timeout → gateway retry succeeds → 2 provider-side log entries, 1 completion delivered. Later, the identical prompt is served from cache with no provider call. Across those three events, what does "requests" mean in Analytics — client requests, provider attempts, or log rows? For Costs and billing, are cached hits and failed attempts included in request counts or in token/cost totals? A team reporting "we served N requests for $X" gets materially different numbers depending on the answer (here: 3, 2, or 1 requests), and the docs don't seem to define the unit per metric.

**Proposed direction.** A short docs paragraph defining the counting unit per analytics metric (client request vs. provider attempt vs. cache hit), or an explicit dimension separating attempts from client requests, would make the numbers auditable.

**Non-goal.** Not pitching anything; we hit this while defining internal usage metering and want Cloudflare's intended semantics.

roy-tong:
## Summary

Adds a **Counting basis** section to the AI Gateway analytics reference, defining what the Requests / Tokens / Cost metrics count per event shape: plain request, gateway retry, cross-provider fallback, and cache hit with zero provider attempts.

Follows the direction agreed in #32983:

- four-case table mapping each event shape to log rows, provider attempts, the Requests metric, and token/cost attribution;
- two audit rules: attempts and requests are separate grains (reliability reads attempts, demand reads requests), and **a metric whose counting basis is not stated should be treated as not yet defined**, never assumed;
- current built-in behavior is labeled as observed from the outside and pending maintainer confirmation wherever the public docs did not previously define it, so reviewers can evaluate concrete wording rather than intent.

## Notes for reviewers

- The intent is definitional, not behavioral: no product behavior is described as changed.
- Wording for rows marked observed is deliberately conservative and should be confirmed against the backend; anything maintainers can confirm can drop the observed qualifier.

Closes #32983


shawnbure (commented):
Thanks for turning the issue into a concrete reviewable artifact. The request/attempt separation and the four event shapes are the right structure.

One evidence-boundary change is still needed before this can be treated as an auditable definition: the table currently states exact behavior as fact, while the PR description says those behaviors are observed or pending maintainer confirmation. Please add a **Definition status** column (for example, documented / backend-confirmed / observed / pending confirmation) and apply it per row or per cell.

The highest-risk cells are:

- retry: “2” log rows and “1 request”;
- fallback: one row per provider attempt and “1 request”;
- retry/fallback token and cost attribution;
- cache hit: “1” log row, counted as a served request, and zero cost contribution.

Until maintainers confirm those units, I would phrase them as proposed/observed and avoid “sum of what providers billed,” which can imply a billing guarantee. A safer interim form is “displayed attribution is pending backend confirmation; do not use for billing or enforcement.”

That change would make the UNPROVABLE rule operational instead of leaving it only in prose.

roy-tong:
Pushed 16712cd per your review: added a Definition status column (observed / pending confirmation, with backend-confirmed reserved for maintainers), removed the billing-guarantee phrasing - cost cells now say displayed cost attribution, and the rules state that displayed attribution must not be used for billing or enforcement until backend-confirmed. The not-yet-defined rule now names the statuses instead of leaving them in prose. Also applied the docs-bot wording fixes. The risky cells you called out (retry rows/requests, fallback attribution, cache-hit cost) are all marked pending confirmation.

shawnbure (commented):
The Definition status column and removal of the provider-billing guarantee address the evidence boundary in my September 14 review. Thank you for making those changes in 16712cd.

One remaining presentation risk: the pending-confirmation rows still read as exact definitions when the table is excerpted without the rules below it. Please label those cell values as proposed (or put a short “proposed counting basis; not backend-confirmed” sentence immediately before the table). That will preserve the boundary when someone quotes the table in a report. I agree with the bot's small wording correction: merge the final bullet into the second, or change “Two rules” to “Three rules.”

The request/attempt distinction is useful; this review does not verify the proposed backend behavior. Maintainer confirmation remains the prerequisite for treating the rows as a product definition or using their cost attribution for billing or enforcement.

shawnbure (commented):
The new pre-table boundary resolves the excerpting risk from my September 15 review. I no longer have an evidence-boundary concern with this revision: the counting basis is explicitly proposed, cost attribution is not positioned for billing or enforcement, and backend confirmation remains the prerequisite for a product definition. The remaining wording items are editorial and are already captured by the docs bot.

PR #33419, opened 2026-09-14
Add note about profiler preventing eviction of DO

dom96:
Just a little note about a new way that a DO can be kept from eviction.


irvinebroque (commented):
how does the reader know what is meant by an "active profiling session"?

is this the right home for this content?

danlapid:
> how does the reader know what is meant by an "active profiling session"?
> 
> is this the right home for this content?

We just need to add a link to the profiling docs once we have them, to be fair we can add this note in the same PR as the profiling docs.

I think it's a fitting home to put a short comment here because it affects the lifecycle of DOs which is something we want to make more explicit and clear.

shrima-cf (approved):

iglesiasbrandon (approved):

PR #33422, opened 2026-09-14
cache: document Tiered Cache eligibility timing and upper-tier origin-path latency

stechedo:
## What

Documents two Tiered Cache behaviors that customer tickets repeatedly surface, from a real escalation (InvestCloud custom hostname, intermittent asset latency from EWR):

1. **Request-time vs response-time bypass.** Tiered Cache activates for cache-eligible requests when they arrive. Origin response headers (`Cache-Control: no-store` / bare `private`) act on the response, after the upper-tier fetch has already happened, so origin-header bypass produces `BYPASS` without stopping the upper-tier fetch. Only a request-time Cache Rule **Bypass cache** prevents tiering.
2. **Upper-tier selection ignores upper-tier→origin latency.** Generic Global maps lower tiers to fixed regional upper tiers, statically; only Smart Tiered Cache uses in-house origin-latency data. A congested provider on the upper-tier→origin path slows tiered misses without affecting cache hits, and the fix is a network-path change, not disabling Tiered Cache.

## Changes

- `cache/how-to/tiered-cache.mdx`
  - Generic Global section: static regional lower→upper mapping, no origin-latency probing (Smart does), requests tier within region only.
  - New "Tiered Cache and cache eligibility" section: activation timing, BYPASS-after-upper-tier-fetch behavior, request-time Bypass cache rule.
- `cache/troubleshooting/investigating-tiered-cache-latency.mdx` (new page): intermittent slow-on-MISS-only diagnosis, MTR comparison from upper-tier colo to origin IP vs a working colo, when to route the provider-path finding to network engineering, why disabling Tiered Cache is not the fix, upper-tier placement options (Smart topology, cloud region hint, custom topology).
- `cache/troubleshooting/investigating-uncached-responses.mdx`: `BYPASS` section gains the upper-tier-fetch note; Related resources link the new pages.

Companion internal work already shipped: CSUP wiki runbook updates (Cache Troubleshooting Guide Tiered Cache section; Declaring an Incident anti-criteria) and Support Seal guidance MR (tiered-cache latency journey + escalation judgment guard).


ngayerie:
Hi @mbullock1986 , @zaidoon1 
Could you please review this doc PR?
This is following up incident INCIDENT-9679, to avoid such escalation at first place.

src/content/docs/cache/how-to/tiered-cache.mdx
```diff
@@ -33,6 +33,8 @@ Cloudflare allows you to select your cache topology so that you have control ove
 
 Generic Global topology allows for all of Cloudflare’s global data centers to serve as a network of upper-tiers. This topology may help reduce the long tail latencies for far-away visitors.
 
+Generic Global maps each lower-tier data center to fixed upper tiers within its region, and requests only tier within their region: a request served in Africa is not tiered to an upper tier in North America. The mapping is static, and Generic Global does not use latency data between the upper tier and your origin. If the path from your origin to the assigned upper tier is congested, cache misses tiered through it are slower without affecting cache hits. To select upper tiers using in-house origin-latency data, use [Smart Tiered Cache](#smart-tiered-cache). To confirm whether a request used Tiered Cache, refer to the `CacheTieredFill` field in your [http_requests](/logs/logpush/logpush-job/datasets/zone/http_requests/#cachetieredfill) logs.
```
zaidoon1, line 36: The sentence directly above this still says all of Cloudflare's global data centers serve as a network of upper tiers. With this paragraph added right after it, the page says two different things about Generic Global's scope two lines apart — worth reworking the older sentence rather than leaving both.

src/content/docs/cache/how-to/tiered-cache.mdx
```diff
@@ -45,6 +47,12 @@ This can help improve performance for **Smart** and **Custom Tiered Cache** topo
 
 Custom Tiered cache allows Enterprise customers to work with their account team to set a custom topology that fits your specific needs, for instance you have close upper tiers or you have a unique traffic pattern. If you want a custom topology, please engage your account team.
 
+## Tiered Cache and cache eligibility
+
+Tiered Cache activates for requests that are eligible for cache when they arrive. Origin response headers act later, on the response: an origin returning `Cache-Control: no-store` or bare `private` produces a [`BYPASS`](/cache/troubleshooting/investigating-uncached-responses/#bypass--origin-response-is-not-cacheable) status, but the upper-tier fetch has already happened. Origin-header bypass does not prevent the upper-tier fetch. To stop requests from being tiered, prevent their eligibility at request time with a [Cache Rule](/cache/how-to/cache-rules/settings/#bypass-cache) set to **Bypass cache**.
```
zaidoon1, line 52: A legacy `Cache Level: Bypass` Page Rule or Configuration Rule stops eligibility the same way a Cache Rule does, and [Investigate uncached responses](/cache/troubleshooting/investigating-uncached-responses/) already lists them together. Worth widening this to "a request-time bypass rule" so Cache Rules don't read as the only route.

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
...
+description: Diagnose intermittent slowness on requests that Cloudflare fetches from an upper-tier data center through Tiered Cache.
+sidebar:
+  order: 3
+---
+
+Some requests for the same URL are slow and others are fast. The slow requests miss in the lower-tier data center and fetch through an upper-tier data center, while requests served from the local cache are fast. Tiered Cache activation is per request, which is why the slowness appears intermittent.
+
+## Before you start
+
+Capture the `cf-cache-status` value and the Ray ID for both a slow and a fast request. The Ray ID suffix identifies the data center that served the request (`cf-ray: 3ff26f9cc66f5a4d-EWR`). The slow requests are typically `MISS` or `EXPIRED` in a data center that fetches through an upper tier; the fast requests are `HIT`.
```
zaidoon1, line 15: This heuristic doesn't hold once Tiered Cache is in play. When the upper tier has the asset, the status the client sees reflects that hit, so `HIT` doesn't mean the request was served from the local data center's cache — it may well have crossed a data center hop and still been fast.

So `cf-cache-status` on its own won't separate local hits from tiered fetches; `CacheTieredFill` is what does. Suggest leaning on that field here rather than on the HIT/MISS split.

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+To confirm whether a request was served through Tiered Cache, check the `CacheTieredFill` field in your [http_requests](/logs/logpush/logpush-job/datasets/zone/http_requests/#cachetieredfill) logs.
+
+## The slow requests tier through an upper tier with a congested origin path
+
+Tiered Cache maps each lower-tier data center to upper-tier data centers per your topology. The mapping is static, and it does not use real-time latency between the upper tier and your origin (only [Smart Tiered Cache](/cache/how-to/tiered-cache/#smart-tiered-cache) uses origin-latency data). When the path from the upper tier to your origin is congested, for example because a transit provider on that path is saturated, tiered cache misses take longer without affecting cache hits.
```
zaidoon1, line 21: Step 1 below asks the reader to test "the upper-tier data center", but a lower tier doesn't have a single upper tier — which one a request uses can differ between URLs and change over time. Worth saying that here, otherwise a comparison run against one asset gets read as describing the whole zone.

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+1. Run an MTR from the upper-tier data center to the origin IP, and compare with a data center whose requests are fast.
```
zaidoon1, line 23: A customer can't run an MTR from a Cloudflare data center, and nothing earlier on the page tells them which upper tier their request used — the Ray ID suffix is the data center that served them, not the upper tier. As written this step isn't something the reader can carry out.

Suggested reframe: capture Ray IDs and `cf-cache-status` for a fast and a slow request, then open a support case asking for the upper-tier path to the origin to be checked. An MTR from their own host to the origin is still worth including as the customer-side half.

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
...
+2. Compare the paths. Different providers or different path latencies between the two data centers and the origin explain the difference.
+3. Route the network-path finding to our network team with the MTR evidence: the congested provider path, the origin IP and ASN, and the working comparison. The fix is a network-path change, such as routing that prefix over a different provider.
```
zaidoon1, line 25: "Route the network-path finding to our network team" is internal process — the reader here is the customer, and they have no way to do this (or to produce the MTR evidence it depends on). Suggest turning it into what to include in the support case, and noting that Cloudflare can look at rerouting the origin path from there.

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+Do not disable Tiered Cache as the fix for a congested provider path. Disabling it only sends origin fetches from every lower-tier data center over uncoordinated paths; routing around the congestion is what fixes it.
+
+## Tiered Cache activates on requests the origin header-bypasses
+
+An origin response header such as `Cache-Control: no-store` or bare `private` takes effect on the response, after Cloudflare has already fetched the asset through the upper tier. The response status is [`BYPASS`](/cache/concepts/cache-responses/#bypass), but origin traffic still concentrated through the upper tier. To prevent requests from being tiered at all, stop their cache eligibility at request time with a [Cache Rule](/cache/how-to/cache-rules/settings/#bypass-cache) set to **Bypass cache**. Refer to [Tiered Cache and cache eligibility](/cache/how-to/tiered-cache/#tiered-cache-and-cache-eligibility).
+
+## The customer wants a closer or faster upper tier
```
zaidoon1, line 33: This heading, and "the topology to recommend" below it, are written for someone advising a customer rather than for the customer reading the page. Something like "Get a closer or faster upper tier" with "use this when the assigned upper tier is far from your origin" keeps the same voice as the rest of the page.

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
...
+3. Route the network-path finding to our network team with the MTR evidence: the congested provider path, the origin IP and ASN, and the working comparison. The fix is a network-path change, such as routing that prefix over a different provider.
+
+Do not disable Tiered Cache as the fix for a congested provider path. Disabling it only sends origin fetches from every lower-tier data center over uncoordinated paths; routing around the congestion is what fixes it.
+
+## Tiered Cache activates on requests the origin header-bypasses
+
+An origin response header such as `Cache-Control: no-store` or bare `private` takes effect on the response, after Cloudflare has already fetched the asset through the upper tier. The response status is [`BYPASS`](/cache/concepts/cache-responses/#bypass), but origin traffic still concentrated through the upper tier. To prevent requests from being tiered at all, stop their cache eligibility at request time with a [Cache Rule](/cache/how-to/cache-rules/settings/#bypass-cache) set to **Bypass cache**. Refer to [Tiered Cache and cache eligibility](/cache/how-to/tiered-cache/#tiered-cache-and-cache-eligibility).
```
zaidoon1, line 31: Worth a "by default" — [Investigate uncached responses](/cache/troubleshooting/investigating-uncached-responses/) documents two exceptions for these same directives (`private="<header>"` with field names stays cacheable, and an Edge TTL rule set to ignore origin cache-control overrides both). Otherwise the two pages disagree for anyone using them.

src/content/docs/cache/troubleshooting/investigating-uncached-responses.mdx
```diff
@@ -43,14 +43,16 @@ The request was eligible for cache, but the origin response or configuration pre
...
 - **The origin returned `Cache-Control: no-store` or bare `private`.** These directives block caching in either [Origin Cache Control](/cache/concepts/cache-control/) mode by default. Two exceptions: `Cache-Control: private="<header>"` with field names remains cacheable — Cloudflare drops only the named headers. And a [Cache Rule](/cache/how-to/cache-rules/) with an Edge TTL that ignores origin cache-control (**Edge TTL → Ignore cache-control header and use this TTL** or **Status code TTL**) overrides both directives, so a response with `no-store` plus that Edge TTL setting is cached.
+- **Tiered Cache already fetched the asset through an upper tier.** If the request was eligible for cache when it arrived, Tiered Cache may have activated and fetched the asset from an upper tier before the origin response headers took effect. The response status is `BYPASS`, but the origin fetch still concentrated through the upper tier. Origin-header bypass does not prevent this; a request-time [Cache Rule](/cache/how-to/cache-rules/settings/#bypass-cache) with **Bypass cache** does. Refer to [Tiered Cache and cache eligibility](/cache/how-to/tiered-cache/#tiered-cache-and-cache-eligibility).
```
zaidoon1, line 46: This bullet is a consequence rather than a cause — every other item in the list answers why the response wasn't cacheable, while this one is true regardless of which of them applied. It reads better as a short note after the list.

Also worth mirroring it in the `BYPASS` section of [Cache responses](/cache/concepts/cache-responses/), since the two lists have tracked each other until now.

zaidoon1 (commented):
thanks for working on this! I've left some comments for things we should fix before this lands

src/content/docs/cache/how-to/tiered-cache.mdx
foreseaz: same feedback with @zaidoon1 

I think we should replace the existing Generic Global description rather than append this paragraph. Generic Global does not use a fixed lower-tier-to-upper-tier mapping: it selects from a regional pool of upper-tier candidates using a consistent hash of the cache key. “Static” and the Africa/North America example are therefore misleading, since cache regions are Cloudflare-defined and are not a customer-visible promise about continental boundaries.

Could we replace both paragraphs with:




```suggestion
Generic Global Tiered Cache organizes Cloudflare’s network into regional upper-tier pools. Cloudflare designs these pools to balance geographic scope, proximity to visitor-serving locations, and the distribution of suitably provisioned cache capacity. Each pool contains a set of eligible data centers that can serve as upper tiers.

For requests that use Tiered Cache, Generic Global uses a consistent hash of the cache key to select an upper-tier data center from the relevant regional pool. This normally directs requests with the same cache key in a region to the same upper-tier cache, reducing duplicate cache fills and improving cache efficiency.

Cloudflare updates the candidate pool as data-center availability, capacity, and network topology evolve. When the candidate pool changes, some cache keys may be reassigned.

Unlike Smart Tiered Cache, Generic Global does not use latency measurements between a particular origin and candidate upper-tier data centers to make this selection. Instead, it emphasizes regional locality and shared-cache efficiency. Smart Tiered Cache can use origin-specific performance and latency data to select upper tiers when that data is available.

To confirm whether a request used Tiered Cache, refer to the `CacheTieredFill` field in your [http_requests](/logs/logpush/logpush-job/datasets/zone/http_requests/#cachetieredfill) logs.
```

src/content/docs/cache/how-to/tiered-cache.mdx
foreseaz: same above, here we could give a general intro to the pointer of more detailed "Tiered Cache and cache eligibility" discussion in trouble shooting.

```suggestion
Tiered Cache applies only to requests eligible for cache at request time. Origin-response cacheability is evaluated later, so an eligible request can be routed through an upper tier before Cloudflare treats the origin response as uncacheable. For details and request-time bypass options, refer to [Investigate uncached responses](/cache/troubleshooting/investigating-uncached-responses/).
```

src/content/docs/cache/troubleshooting/investigating-tiered-cache-latency.mdx
```diff
@@ -0,0 +1,43 @@
+---
```
foreseaz, line 1: I think this page should target a narrower, customer-actionable question: whether slow request samples correlate with Tiered Cache usage or a particular upper-tier candidate, and what evidence to collect before contacting Support.

The HTTP Requests dataset already exposes `CacheTieredFill`, `EdgeColoCode`, `UpperTierColoID`, cache status, and timing fields. Could we build the workflow around comparing slow and fast request records with those fields instead of asking customers to run an MTR? We should be explicit that `OriginResponseDurationMs` includes Tiered Cache, Argo, connection, and origin time, so these fields establish correlation rather than isolate a network hop or prove congestion.

So in short, agree with what @zaidoon1 's feedbacks below, and to me, this page needs rework as whole based on what current customer's confusion.

ngayerie:
[AI] This PR has been superseded by #33587, which addresses the engineering review comments from @zaidoon1 and @foreseaz:

- **Generic Global description rewritten**: the "static fixed mapping" characterization was incorrect. The new text describes the consistent-hash selection from a regional pool of upper-tier data centers, with no per-origin latency data (versus Smart Tiered Cache which does use latency data).
- **Troubleshooting page fully reworked**: the MTR-based workflow has been replaced with a Log Explorer–based workflow using `CacheTieredFill`, `EdgeColoCode`, `UpperTierColoID`, and `OriginResponseDurationMs` fields, with a note that `OriginResponseDurationMs` measures correlation across all hops, not path isolation.
- **`investigating-uncached-responses.mdx`** updated with a Tiered Cache + BYPASS timing note and related-resources links.

Reviewer pings (@zaidoon1, @foreseaz) have been requested on #33587. Feel free to close this PR.

ngayerie:
New PR raised, addressing the initial request + all the comments
https://github.com/cloudflare/cloudflare-docs/pull/33587

PR #33433, opened 2026-09-15
chore: bump nimbus version to 0.14.1

MohamedH1998:
### Summary

Updates Nimbus Docs from 0.11.0 to 0.14.1 and applies the corresponding integration migrations. This adds the Nimbus migration baseline, moves partial ID resolution into the generated Markdown configuration, updates the docs route to the current page-props API, and ignores the generated coordinates manifest.

### Documentation checklist


MohamedH1998:
Skipping the warning because the reported regression is not present in Nimbus 0.14.1.

`getDocsPageProps` now accepts only the `Astro` global, and its public API states that headings from `<Render file="..." />` partials are recursively merged at build time. The 0.14.1 implementation builds prepared heading records with `mergePartialHeadings(...)` and uses `markdown.partialResolver` to map Cloudflare's `product/file` IDs. The generated `/r2/reference/wrangler-commands/` page also confirms that headings from `partials/workers/wrangler-commands/r2.mdx` are present in the TOC.

Restoring the removed `partialHeadings` argument would therefore be type-invalid and would duplicate the replacement mechanism.

diogoascarneiro (approved):

PR #33439, opened 2026-09-15
[DNS] Add shadowed record warnings changelog

matildeopbravo:
## Summary

Adds a DNS changelog entry describing shadowed record warnings across zones and the opt-in API metadata for identifying delegations and glue records.

## Documentation checklist

- [x] Changelog entry
- [x] Style guide adherence

## Validation

- Prettier check passed for the changelog entry.
- `pnpm run check` could not run because `@astrojs/markdown-satteri` is missing from the local dependencies.


src/content/changelog/dns/2026-09-14-shadowed-record-warnings.mdx
```diff
@@ -0,0 +1,11 @@
+---
+title: Shadowed record warnings are now available for all zones
+description: Find DNS records shadowed by a delegation.
+products:
+  - dns
+date: 2026-09-14
+---
+
+Cloudflare now displays warnings for shadowed records in all zones. A record is shadowed when a subdomain delegation gives authority for its name, or a name below it, to another set of nameservers. The record remains present, but your zone is not authoritative for it. These warnings help you find records that may no longer resolve from the expected zone.
```
hannes-cf:
```suggestion
Cloudflare now displays warnings for shadowed records in all zones. A record is shadowed when a subdomain delegation gives authority for its name, or a name below it, to another set of nameservers. The record remains present, but your zone is not authoritative for it thus Cloudflare will not respond with it to matching DNS queries. These warnings help you find records that may no longer resolve from the expected zone.
```

hannes-cf (approved):

PR #33473, opened 2026-09-16
[Workers] Note that node:crypto randomUUIDv7 is not supported

### LINKED ISSUE #32892: No support for node:crypto.randomUUIDv7() but documentation suggests it is

https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/

The page linked above makes this statement: _All node:crypto APIs are fully supported in Workers with the following exceptions:_ However, workerd does not support `node:crypto.randomUUIDv7()` yet. Either this fact should be added to the list of NOT supported features or support should be added (preferable). Understandably, adding the support is not the same level of effort as updating the documentation to say it is not supported, but it would be helpful to include the support.

### Additional information

- The UUIDv7 functionality was added to `node:crypto` in v26.1. 
- My `nodejs_compat` flag is set and my `compatibility_date` is "2026-08-15", so I don't believe I'm missing anything - it's simply not supported yet. Here is the error produced:

` ✘ [ERROR] service core:user:app-api: Uncaught SyntaxError: The requested module 'node:crypto' does not provide an export named 'randomUUIDv7'`

Denimworld12:
### Summary

Fixes #32892.

[`node:crypto`](https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/) states that all `node:crypto` APIs are fully supported in Workers with a listed set of exceptions. `randomUUIDv7` — added to `node:crypto` in Node.js v26.1 — is not implemented in workerd and was not in that list, so the page implied support that does not exist.

Confirmed in workerd: `src/node/internal/crypto_random.ts` exports `randomUUID` and has no `randomUUIDv7`. Importing it fails at startup:

```
✘ [ERROR] service core:user:app-api: Uncaught SyntaxError: The requested module 'node:crypto' does not provide an export named 'randomUUIDv7'
```

Adds it to the exceptions list, matching the phrasing of the existing `argon2` entry.

If support is added to workerd later, this line should be removed again.

### Documentation checklist


Denimworld12:
Adding the verification behind this change, in case it saves a reviewer the setup.

**Reproduced on the runtime.** Minimal Worker, `nodejs_compat`, `compatibility_date` `2026-08-15`, wrangler 4.132.0:

```js
import { randomUUIDv7 } from "node:crypto";
export default { fetch() { return new Response(randomUUIDv7()); } };
```

```
✘ [ERROR] service core:user:uuidv7-probe: Uncaught SyntaxError:
  The requested module 'node:crypto' does not provide an export named 'randomUUIDv7'
✘ The Workers runtime failed to start.
```

**Control, same config** — only the import changed to `randomUUID`:

```
v4 works: 88309a3a-435e-4d0c-9e79-6807f0d0a43f
```

So it isn't a missing flag or a stale compat date; v4 is present and v7 specifically is not.

**In workerd:** `src/node/internal/crypto_random.ts` exports `randomUUID` and no `randomUUIDv7`; `src/node/crypto.ts` imports and re-exports `randomUUID` only, and its support checklist has `[x] crypto.randomUUID([options])` with no v7 entry. A code search for `randomUUIDv7` across the repo returns nothing, while `randomUUID` returns six hits.

**One thing worth flagging:** `wrangler deploy --dry-run` **passes** for the failing Worker. esbuild treats `node:crypto` as external, so bundling never resolves the named export — the error only appears once the runtime loads the module. A build-level check would report this page as accurate.

As noted in the description, this line should be removed again if `randomUUIDv7` lands in workerd.

PR #33555, opened 2026-09-21
[Hyperdrive] Document Postgres.js prepare: false pitfall

ivoryibu:
### Summary

Adds a caution to the shared Hyperdrive driver recommendation note against setting `prepare: false` on a Postgres.js client. The note renders wherever Postgres.js is listed as supported: the [supported drivers](https://developers.cloudflare.com/hyperdrive/examples/connect-to-postgres/) page and the [Postgres.js](https://developers.cloudflare.com/hyperdrive/examples/connect-to-postgres/postgres-drivers-and-libraries/postgres-js/) and node-postgres driver pages.

With prepared statements disabled, Postgres.js sends additional protocol messages to discover parameter types before each query, which Hyperdrive's transaction pooling mode does not reliably support. This can cause individual queries to hang or fail intermittently while the connection pool otherwise appears healthy — a failure mode that is hard to attribute to the driver configuration. The caution directs readers to node-postgres (`pg`), which is already the recommended driver.

Prompted by a community report of intermittent query hangs with Postgres.js `prepare: false` behind Hyperdrive.

Possible follow-up: the [Drizzle ORM page](https://developers.cloudflare.com/hyperdrive/examples/connect-to-postgres/postgres-drivers-and-libraries/drizzle-orm/) note says node-postgres and Postgres.js are both "supported and compatible" — worth qualifying there too, since Drizzle with the Postgres.js driver may not use named prepared statements by default.

### Documentation checklist


rzazadebilal7:
[HKEX-EPS_20260827_12302447_0.pdf](https://github.com/user-attachments/files/32465435/HKEX-EPS_20260827_12302447_0.pdf)

rzazadebilal7:
Mene kömək edin onu düzəldə ideal iwlesin

PR #33583, opened 2026-09-22
Improve Cloudflare One auto labelling

mvvmm:
Improve Cloudflare One auto labelling.


mvvmm:
Skipping behavior changes for `CR-a4f83230387e` and `CR-612fa130749c`.

The empty-`newLabels` guard is intentional. Product labels have no provenance, so when an edited body contains no recognized Docs links the action cannot distinguish labels it previously added from labels applied manually or by other automation. Removing every `product:*` label would silently undo manual triage, which the previous review identified as unsafe. The follow-up documents this conservative behavior in code; stale automatic labels may remain until label provenance is tracked.

MohamedH1998 (approved):

PR #33595, opened 2026-09-22
feat: cf cli components

MohamedH1998:
### Documentation checklist


MohamedH1998:
Skipping these findings after validating them against the current implementation:

- `CR-b5a93a57dfe9`: `args` intentionally defines the Wrangler invocation, while `cfCommand` provides its complete reviewed alternative. Rejecting or appending `args` would break valid callsites.
- `CR-739c4caeb984`: `BaseLayout.astro` already includes inline pre-paint CSS that hides the inactive variant, with preference restoration running in `<head>`.
- `CR-6e8454ee2aba`: Direct loads have no server-rendered CLI preference attribute, so stored-preference resolution is not skipped.
- `CR-fa1b52b2778a`: `PackageManagerTabs` is private and only receives normalized output from `getTabs()`, which guarantees the expected command shape.
- `CR-4fa6019a90b1`: The root preference is initialized by `CliPreferenceRestore` using the same override → stored → route-default precedence used after navigation.
- `CR-2fb3db653983`: Command metadata comes from pinned executable dependencies, not user input. Those packages already execute at build time, so this is not a meaningful sanitization boundary.
- `CR-cd48836ab9a2`: Removing the small Astro template duplication would require a new private component and add more abstraction than it removes.
- `CR-7043c1e2891a`: Eliminating this small Tabs duplication similarly requires another component or broader abstraction without a behavioral benefit.

mvvmm (approved):

PR #33606, opened 2026-09-22
feat: add changelog date check to docs bot

mvvmm:
### Summary

Adds a non-blocking changelog date check to the docs bot (`.flue/`) so PRs that add changelog entries dated in the past get a single ⚠️ comment, without failing CI.

- **Nightly cron**: sweeps every open PR and reconciles the warning comment (create/update while a stale entry exists, delete once the date is current or the file is removed). Runs in the existing `scheduled()` handler before the draft-stale sweep, with independent error isolation.
- **Webhooks**: `pull_request` events (`opened`/`reopened`/`synchronize`/`ready_for_review`/`closed`) run the same reconciliation inline before the regular review routing — additive, never replaces it. `closed` removes the comment.
- **Scope**: only files the PR introduces (status `added`/`copied`) under `src/content/changelog/**` with a frontmatter `date:`. An entry dated before the current UTC day is stale; future dates pass (`publish_future_dated_entry` is a supported feature); unparsable dates are skipped.
- **Design**: deterministic trusted TypeScript — no agent/model calls, no R2 state (the singleton comment is located by an HTML marker each run), no new bindings or migrations.

Example comment:

```markdown
⚠️ @author This pull request adds a changelog entry dated in the past:

- `src/content/changelog/workers/2026-09-14-my-entry.mdx` — dated **2026-09-14**, 8 days old

Typically, changelog entries should be dated the day they merge. This is not blocking — if the date is unintentional, please update it.
```


mvvmm:
Triage notes for the review findings — two items declined/skipped, rest fixed in an upcoming push:

- **CR-47efc861db01** (duplicated API work in scheduled handler) — skipping. One extra token mint plus one extra paginated open-PR listing per *nightly cron* is negligible, and the suggested refactor would couple the changelog sweep and the draft-stale sweep (shared token/list changes `runDraftStaleSweep`'s signature and collapses their independent failure isolation — a token/listing failure in one would now abort the other).

- **CR-e5636379803a** (unbounded inline latency on webhook hot path) — partially addressed: the per-file content fetches in `collectStaleChangelogEntries` are parallelized with `Promise.all` instead of run sequentially, bounding the per-file latency. Declining the suggestion to move the check into the durable INGEST path: it would complicate `closed`-event comment cleanup and draft-PR coverage (INGEST doesn't run for either), and realistic PRs add 1–3 changelog files, so inline cost stays well within GitHub's delivery timeout.

ahaywood (approved):

